PR #3105, opened 2019-06-11
Increase fidelity of error logging for PPMS service issues

kfrz:
## Description of change
PPMS is occasionally experiencing high latency and failed requests. This resulted in a lot of [sentry events](http://sentry.vetsgov-internal/vets-gov/platform-api-production/issues/59689) which weren't very helpful in debugging>

* Add spec for GatewayTimeout case
* Add spec for status 500 case
* Refactor PPMSParser to handle 500 case
* Add PPMS code to exceptions.en.yml

Co-Authored-By: null <SMLuthi@users.noreply.github.com>

## Testing done
Specs written, local debugging

## Testing planned
CI

#### Unique to this PR

- [x] FacilityLocator requests should raise an BackendServiceException if they result in an error
- [x] Tests in place for errored cases

#### Applies to all PRs

- [x] Appropriate logging


lib/common/client/middleware/response/ppms_parser.rb
```diff
@@ -13,10 +13,16 @@ def on_complete(env)
...
-            # flag Not Found errors as success so the error doesn't bubble up
-            if hash['error'] && hash['error']['message'] && hash['error']['message'] =~ /No Providers found/
+            msg = hash.dig('error', 'message')
+
+            case msg
+            when /No Providers found/ # flag Not Found errors as success so the error doesn't bubble up
               env[:status] = 200
               return []
+            when /An error has occurred/ # PPMS has encountered an internal error
+              hash['error']['code'] = nil if hash['error']['code'].blank? # Set code so it can match exceptions.en.yml
```
kfrz: Revisit this in the AM to make sure it makes sense :)

lib/common/client/middleware/response/ppms_parser.rb
```diff
@@ -13,10 +13,16 @@ def on_complete(env)
...
+              hash['error']['detail'] = hash['error']['message']
+              return hash['error']
             end
```
lihanli: if msg doesn't match either case isn't [this error](http://sentry.vetsgov-internal/vets-gov/platform-api-production/issues/59689/) still going to occur where `body` is `nil`?

SMLuthi: I agree that we need to account for a potential issue where there is no body (however unlikely it is... since even in this case there WAS a body but it was just a stringified hash error message) that we return, at the very least, an empty hash so it doesnt fail on trying to index in the `raise_error` file.

kfrz: Yeah - if we're getting an error with an empty body that means PPMS is sending unexpected payloads. Would have to cross reference their API docs to cover any additional error cases, I imagine.

kfrz: @lihanli @SMLuthi I think it'd be more appropriate to handle the "empty body" case in a separate body of work, as it relates to other services/controllers. As discussed in the eng. mtg. today w/ @saneshark brought up better handling of "non-coded" errors, this would fall into that category.

Which, I think, would be best handled in a response middleware with accompanying shared_context examples for any 3rd party service class

saneshark: I think in the case of the empty / nil body, just making sure that the 500 response received can map to an appropriate response that can be properly handled is maybe satisfactory. A 500 error might be a 502 or 503. I'm certain we're handling this sort of thing elsewhere (presumably in Rx and SM APIs since those are most mature at this point), could take a look there and make sure we're handling things consistently?

kfrz: Yeah, like we have the VA900 code as a catch-all [here](https://github.com/department-of-veterans-affairs/vets-api/blob/master/config/locales/exceptions.en.yml#L174)

kfrz: @saneshark to be clear, this PR will make it such that `500` responses will raise a `BackendServiceException` with the `PPMS_500` key to identify. If a different `50x` status comes back we would see a `VA900` code which should prompt us to introduce further specific error handling for this endpoint.

This will help us, for the time being, to identify more precisely how often PPMS is returning this error.

--- 

In a future work effort we will investigate solutions to further streamline nil response checking, i.e. in a middleware layer.

We do have [`gids_errors`](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/common/client/middleware/response/gids_errors.rb) that has a similar pattern, though the errors body there is pluralized.

I think there may be a way to extract the task of checking for JSON errors into its own `json_errors` middleware, but given the inconsistencies between providers, it may be cleaner to follow a `{service}_errors.rb` middleware pattern as we're doing already with [`mhv_errors.rb`](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/common/client/middleware/response/mhv_errors.rb).

config/locales/exceptions.en.yml
```diff
@@ -445,6 +445,12 @@ en:
         code: 'PAGERDUTY_429'
         detail: "Exceeded PagerDuty's API rate limit"
         status: 429
+      PPMS_500:
+        <<: *external_defaults
+        title: Unexpected service error
+        code: 'PPMS_500'
+        detail: 'An unexpected error has occurred'
+        status: 500
```
saneshark: is a 500 exception from downstream necessarily a 500 exception? I'm wondering if this ought to be 502 or 503?

kfrz: Good nit, yes this should probably send a `502 - Bad Gateway` as it matches: 

> The server, while acting as a gateway or proxy, received an invalid response from an inbound server it accessed while attempting to fulfill the request.

config/locales/exceptions.en.yml
```diff
@@ -450,6 +450,12 @@ en:
         code: 'PAGERDUTY_429'
         detail: "Exceeded PagerDuty's API rate limit"
         status: 429
+      PPMS_502:
+        <<: *external_defaults
+        title: Bad Gateway
+        code: 'PPMS_502'
+        detail: 'PPMS returned an invalid response'
+        status: 502
```
ncksllvn: It looks like we're considering `408, `504`, and `503` to be status codes associated with timeouts on the FE, but we'll need to add `502` - https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/applications/facility-locator/components/ResultsList.jsx#L21. I can do that once this is merged.

ncksllvn:
@saneshark sorry, missed your question -

> do you have bandwidth to test this? It might be a little tricky to trigger these specific errors and see what the return response is on the FE, but I'd be satisfied with you comparing error handling logic on the FE against this to make sure all exceptions are handled properly.

If this is going to be in API-Dev soon, I should be able to fully test this. But as long as the status codes align between FE and API, we should be safe.

config/locales/exceptions.en.yml
ncksllvn: Here's that FE PR adding `502` to the list of status codes that will trigger the FE's messaging for timeouts, https://github.com/department-of-veterans-affairs/vets-website/pull/10284.

saneshark: at some point it would be nice if we have a comprehensive fallback for those common status codes at a more macro level, ideally where the FE shows the message returned when it doesn't need / have more granular handling for the specific code returned.

In particular i'm usually concerned with VA900 codes. Those are generic exceptions (usually ones we've never seen before or haven't had a chance to introduce more granular / special handling on the BE).

saneshark: Also 502 is not necessarily a Timeout.

ncksllvn: Ah gotcha, should it just fall through to the catch-all generic error message then?

saneshark: I'll leave it to Keifer to more explain some of the context around these messages. Frankly from the perspective of the user the generic catch-all probably makes sense in most cases -- there will be ample opportunity to introduce more granular FE handling I think in the future when this gets handed off to VSA.

saneshark: I think eventually some GA related feedback about latency from a FE perspective would be great too #wishlist

kfrz: We landed on `502` because from the perspective of making a request to vets-api, the PPMS response comes back as a `500` but has no further details to pass on to the user.

I think the generic error message makes most sense here. Further improvements around VA900/general codes are on VSP/Triage roadmap for the next few sprints

PR #3107, opened 2019-06-11
Handle bing errors kr

edmkitty:
In the v1 nearby facilities endpoint, handle errors from Bing more gracefully

https://github.com/department-of-veterans-affairs/vets-contrib/issues/2285


app/models/nearby_facility.rb
```diff
@@ -48,6 +59,32 @@ def make_linestring(polygon)
...
+                   .try(:[], 'resources')
+                   .try(:first)
+                   .try(:[], 'polygons')
+                   .try(:first)
+                   .try(:[], 'coordinates')
+                   .try(:first)
+    end
+
+    def handle_bing_errors(body, headers)
+      response_body = JSON.parse(body)
```
nfasulo: We are now parsing `response.body` twice, which would normally make me want to refactor a little. Except in this case we know this isn't going to be a long-lived implementation.

app/models/nearby_facility.rb
```diff
@@ -48,6 +59,32 @@ def make_linestring(polygon)
       polygon.map { |point| "#{point[1]} #{point[0]}" }.join(',')
     end
 
+    def parse_isochrone(response_json)
+      response_json.try(:[], 'resourceSets')
```
nfasulo: Is it possible to write this  `try` chain with safe navigation operators (`&.`)? Not asking you to change it (though it makes me slightly sad that it takes up this many lines); only asking because I tried a couple permutations but didn't get the syntax right.

lib/common/exceptions/external/bing_service_error.rb
```diff
@@ -0,0 +1,25 @@
...
+      def initialize(error_messages)
+        @error_messages = { detail: error_messages }
+      end
+
+      def errors
+        Array(SerializableError.new(i18n_data.merge(@error_messages)))
+      end
+
+      def status_code
+        500
```
nfasulo: Does this need to be a `500` error? All the Bing errors we have managed to run into could be classified as `503`, right?

edmkitty: Maybe let's table that level of granularity for later

app/models/nearby_facility.rb
edmkitty: Possibly. There is some weirdness with the safe navigation operator if something in the chain returns true/false, but I will test it

lib/common/exceptions/external/bing_service_error.rb
rtravitz: We have a ticket out there for figuring out how we communicate intermittent service issues to end users.

PR #3109, opened 2019-06-12
Fix facilities default pagination size

rtravitz:
## Description of change
Fixes [#2288](https://github.com/department-of-veterans-affairs/vets-contrib/issues/2288)

Responses from the facilities api index and nearby routes currently defaults to returning 30 items when a `per_page` param is not specified. Our documentation for these endpoints states that the default is 20 items, and `per_page` methods on `BaseFacility` and `NearbyFacility` indicate that 20 was indeed intended.

The problem is that the `WillPaginate` object from the `will_paginate` gem has a default `per_page` of 30, which we never actually override. To solve it, this PR optionally uses the `per_page` value from `BaseFacility` or `NearbyFacility` when none is provided by the params.

A previous example of this being necessary can be [seen in a facilities controller](https://github.com/department-of-veterans-affairs/vets-api/blob/b85eb618d004c1734bfcd81f381738975e461da7/app/controllers/v0/facilities/va_controller.rb#L20) under the main `/app` directory.

## Testing done
Added two tests asserting that the `per_page` values in `BaseFacility` and `NearbyFacility` are honored.

#### Unique to this PR

- [x] Fix facilities api pagination to match up with published documentation

#### Applies to all PRs


modules/va_facilities/app/controllers/va_facilities/v0/facilities_controller.rb
```diff
@@ -34,7 +34,8 @@ def all
       end
 
       def index
-        resource = BaseFacility.query(params).paginate(page: params[:page], per_page: params[:per_page])
+        resource = BaseFacility.query(params).paginate(page: params[:page],
+                                                       per_page: params[:per_page] || BaseFacility.per_page)
```
nfasulo: Do we want to go ahead and start using `max_per_page` when it is smaller than `params[:per_page]`? We could also just create another ticket if you are ready to move on.

rtravitz: Will go with a followup ticket so we can move on to the next ticket with this!

PR #3113, opened 2019-06-17
docker cache bundled gems

kreek:
## Description of change
Caches bundle install by moving it out of the docker build process and into an entry point script. Adding a new gem will no longer require reinstalling all our gems (which takes some time; looking at you nokogiri 👀) 

## Testing done
Added a new gem locally in both existing and fresh docker installs.

## Testing planned
Confirm CI builds still run

#### Unique to this PR

- [ ] Docker does not need to rebuild all gems when the Gemfile changes

#### Applies to all PRs

- [ ] Appropriate logging

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,8 +20,13 @@ poppler-utils && \
 freshclam
 
 WORKDIR $APP_PATH
-ADD Gemfile $APP_PATH
-ADD Gemfile.lock $APP_PATH
 ADD . /src/vets-api
 
-RUN bundle install
+COPY ./docker-entrypoint.sh /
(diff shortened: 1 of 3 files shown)


kreek:
@wyattwalter and @b00klegger this is for local development. Is there a way to cache gems for Jenkins build. Another question is do we have a VA dockerhub account? Can we get images pre-built up to bundle stage of the Docker file?

annaswims:
what about [bundle package](https://bundler.io/v1.6/bundle_package.html)?  That would protect us in cases when rubygems is down.  One catch might be that `sidekiq-pro` and `sidekiq-ent` shouldn't be committed publicly.

PR #3119, opened 2019-06-20
Creating ITF expiration validation for 526

bastosmichael:
## Description of change
Solves the issue for ticket https://github.com/department-of-veterans-affairs/vets-contrib/issues/2363

## Testing done
- RSpec corrections made

## Testing planned
- Will test in dev-api.va.gov

- We are using header timestamps to check whether a claim's Intent to File is still valid (one year).

#### Applies to all PRs

- [ ] Appropriate logging

### DIFF
--- a/modules/claims_api/app/controllers/claims_api/base_form_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/base_form_controller.rb
@@ -26,7 +26,7 @@ module ClaimsApi
           errors: [
             {
               status: 422,
-              details: 'Intent to File Expiration Date has expired, please resubmit ITF.'
+              details: 'Intent to File Expiration Date has expired, resubmit ITF.'
             }
           ]
         }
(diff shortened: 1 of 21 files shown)


modules/claims_api/app/controllers/claims_api/base_form_controller.rb
```diff
@@ -20,6 +20,24 @@ def schema
 
     private
 
+    def verification_itf_expiration
+      unless itf_service.get_active('compensation')['intent_to_file'].expiration_date > Time.now.utc
```
charleystran: @bastosmichael what happens here if there is no current ITF for the target veteran?

bastosmichael: I went ahead and made the change to ensure it also fails if intent to file doesn't exist

PR #3121, opened 2019-06-21
Add ip address and user agent to logs

saneshark:
## Description of change
We need incoming ip-address added to logs in addition to request_id, this adds it to logs when invoked. Another PR will ensure we log every controller action and sidekiq job that is applicable.

I've also taken the opportunity to move around some of the configuration for logging into production.rb so that it's easier to switch from :color to :json when testing the logging locally.

https://github.com/department-of-veterans-affairs/vets.gov-team/issues/19003

## Testing done
Here is a screen shot of what sidekiq would output:
Screen Shot 2019-06-21 at 10 49 25 AM

incoming requests would behave similarly assuming Thread.current is properly set.

## Testing planned
Just need to verify that production / cloudwatch log are in the proper :json format.

#### Unique to this PR
Added ip address and user agent to log.

#### Applies to all PRs

- [x] Appropriate logging

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -48,6 +48,8 @@ Rails.application.configure do
   # config.action_view.raise_on_missing_translations = true
   ConfigHelper.setup_action_mailer(config)
 
-  # Change this to :json to simulate production
-  config.rails_semantic_logger.format = :color
+  config.rails_semantic_logger.semantic   = false
+  config.rails_semantic_logger.started    = true
+  config.rails_semantic_logger.processing = true
+  config.rails_semantic_logger.rendered   = true
 end
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -48,8 +48,6 @@ Rails.application.configure do
   # config.action_view.raise_on_missing_translations = true
   ConfigHelper.setup_action_mailer(config)
 
-  config.rails_semantic_logger.semantic   = false
-  config.rails_semantic_logger.started    = true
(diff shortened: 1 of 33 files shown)


config/environments/development.rb
```diff
@@ -48,8 +48,6 @@
   # config.action_view.raise_on_missing_translations = true
   ConfigHelper.setup_action_mailer(config)
 
-  config.rails_semantic_logger.semantic   = false
-  config.rails_semantic_logger.started    = true
-  config.rails_semantic_logger.processing = true
-  config.rails_semantic_logger.rendered   = true
+  # Change this to :json to simulate production
```
annaswims: is this just while you're working or is it meant as a permanent change?

saneshark: I out the comment there so someone knows that it's easy to toggle if they want. For some reason ENV var didn't work even with (ENV['LOGGER FORMAT'] || 'color).to_sym

annaswims:
I'd much rather not change the log format in dev.  Can we make a way to toggle for the rare ctimes when we actually want the prod-like log?

Before: 
Screen Shot 2019-06-24 at 9 48 35 AM

After:
Screen Shot 2019-06-24 at 9 50 44 AM

saneshark:
@annaswims there is no change made to dev as a part of this PR. It should be in exactly the same human readable format. If someone wants to play around with json on dev, they will need to change :color to :json on their local and make sure not to commit those changes.

PR #3125, opened 2019-06-25
Starter VSP engine template

kreek:
## Description of change
Adds a module generator that creates a Rails engine starter app that wires up a service backed endpoint. To run the generator `rails va new  {module_name}`. Given a name of `foo` it will generate/modify the following files:

```
      create  modules/foo/app/controllers/foo/v0/foo_controller.rb
      create  modules/foo/app/controllers/foo/application_controller.rb
      create  modules/foo/app/models/foo/foo.rb
      create  modules/foo/app/serializers/foo/foo_serializer.rb
      create  modules/foo/app/services/foo/configuration.rb
      create  modules/foo/app/services/foo/service.rb
      create  modules/foo/lib/foo/engine.rb
      create  modules/foo/lib/foo/version.rb
      create  modules/foo/lib/tasks/foo_tasks.rake
      create  modules/foo/lib/foo.rb
      create  modules/foo/bin/rails
       chmod  modules/foo/bin/rails
      create  modules/foo/config/routes.rb
      create  modules/foo/foo.gemspec
      create  modules/foo/Rakefile
      create  modules/foo/Gemfile
     gemfile  foo
       route  mount Foo::Engine, at: '/foo'
      append  config/settings.yml
         run  bundle install from "."
```

## Testing done
Local module generation

## Testing planned
N/A

#### Unique to this PR
- [x] Generator generates a hello world endpoint.

#### Applies to all PRs

- [x] Appropriate logging


lib/generators/module/templates/Gemfile.erb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+source 'https://rubygems.org'
+git_source(:github) { |repo| "https://github.com/#{repo}.git" }
```
jholton: I don't think we need this line if these Engines are going to remain part of the `vets-api` codebase

lib/generators/module/USAGE
```diff
@@ -0,0 +1,8 @@
+Description:
+    Explain the generator
```
jholton: Should you fill out this file with basically the content in this PR's description?

lib/generators/module/module_generator.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def create_config
+    path = "modules/#{file_name}"
+    template 'bin/rails.erb', File.join(path, 'bin', 'rails')
+    chmod File.join(path, 'bin', 'rails'), 0o755
+    template 'config/routes.rb.erb', File.join(path, 'config', 'routes.rb')
+    template 'gemspec.erb', File.join(path, "#{file_name}.gemspec")
+    template 'Rakefile.erb', File.join(path, 'Rakefile')
+    template 'Gemfile.erb', File.join(path, 'Gemfile')
```
jholton: I don't see the `README` anywhere.  Should that be included here (or somewhere else in this file)?

lib/generators/module/module_generator.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  # rubocop:disable Rails/Output
+  def install
+    gem file_name, path: File.join('modules', file_name)
+    route "mount #{file_name.capitalize}::Engine, at: '/#{file_name}'"
+    append_to_file 'config/settings.yml', "\n#{file_name}:\n  url: 'https://api.va.gov'"
```
jholton: Can we append the `gem` in the Gemfile?
`append_to_file 'Gemfile', "\ngem '#{file_name}', path: 'modules/#{file_name}'"`

Furthermore, how hard is it to place this in the Gemfile with the other Engines at the top?

lib/generators/module/templates/README.md
```diff
@@ -0,0 +1,28 @@
...
+Short description and motivation.
+
+## Usage
+How to use my plugin.
+
+## Installation
+Add this line to your application's Gemfile:
+
+```ruby
+gem 'vsp'
```
jholton: Can you rename this file `README.md.erb` and insert the actual name of the gem here with the path? 
e.g. `gem 'appeals_api', path: 'modules/appeals_api'`

lib/generators/module/templates/README.md
```diff
@@ -0,0 +1,28 @@
...
+```
+
+And then execute:
+```bash
+$ bundle
+```
+
+Or install it yourself as:
+```bash
+$ gem install vsp
```
jholton: Also replace `vsp` here with actual gem name if possible

lib/generators/module/templates/README.md
```diff
@@ -0,0 +1,28 @@
...
+```
+
+## Contributing
+Contribution directions go here.
```
jholton: Do we have some standard text that can go here?
Since all these Engines live in the same `vets-api` codebase, should all of these Engine README sections be identical?  Or will they vary by team?

lib/generators/module/templates/README.md
```diff
@@ -0,0 +1,28 @@
+# Vsp
```
jholton: Ours is `README.md`.  All the other existing Engines have a `README.rdoc`.  Do we want to standardize on one?

lib/generators/module/templates/Rakefile.erb
```diff
@@ -0,0 +1,34 @@
...
+RDoc::Task.new(:rdoc) do |rdoc|
+  rdoc.rdoc_dir = 'rdoc'
+  rdoc.title    = '<%= class_name %>'
+  rdoc.options << '--line-numbers'
+  rdoc.rdoc_files.include('README.md')
+  rdoc.rdoc_files.include('lib/**/*.rb')
+end
+
+APP_RAKEFILE = File.expand_path('test/dummy/Rakefile', __dir__)
+load 'rails/tasks/engine.rake'
```
jholton: We're not generating a `dummy` app, so the two lines above can go away.

spec/support/vcr_cassettes/vsp/get_message.yml
```diff
@@ -0,0 +1,57 @@
+---
```
jholton: Should this file be included?

jholton:
I know we are still reaching final concensus on documentation efforts, but all the other Engines have a `README.yml` file that is their Open API spec.  Do we want to auto-generate one?

Gemfile
```diff
@@ -34,6 +35,8 @@ gem 'clam_scan'
 gem 'config'
 gem 'connect_vbms', git: 'https://github.com/department-of-veterans-affairs/connect_vbms.git', branch: 'master'
 gem 'date_validator'
+gem 'dry-struct'
+gem 'dry-types'
```
jholton: I think we also should include 'fast_jsonapi' here in `Gemfile` too.
If the user doesn't have it already installed, this error occurs:
Screen Shot 2019-07-11 at 3 44 41 PM

kreek: 👍weird it's defined in the gemspec file, I would've thought it gets installed when the bundle runs, but I'll add it in to the main Gemfile

kreek:
Found a random bug on this which I think has to do with Rails dev mode reloading classes. The first time you hit the endpoint it works, but subsequent requests throw (for an app named 'foo') `uninitialized constant Foo::Foo::Service` where it's trying to add a second namespace (because the code is called from within a class also namespaced to Foo). Can fix it by changing all refs to `::Foo::Service` but that doesn't seem like the ideal solution.

lib/common/models/resource.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'dry-types'
+require 'dry-struct'
+
+module Types
+  include Dry::Types.module
```
jholton: Noticed a deprecation warning when hitting the `hello world` endpoint:
```
[dry-types] Dry::Types.module is deprecated and will be removed in the next major version
Use Dry.Types() instead. Beware, it exports strict types by default, for old behavior use Dry.Types(default: :nominal). See more options in the changelog
```

jholton:
> Found a random bug on this which I think has to do with Rails dev mode reloading classes. The first time you hit the endpoint it works, but subsequent requests throw (for an app named 'foo') `uninitialized constant Foo::Foo::Service` where it's trying to add a second namespace (because the code is called from within a class also namespaced to Foo). Can fix it by changing all refs to `::Foo::Service` but that doesn't seem like the ideal solution.

This [blog post](https://urbanautomaton.com/blog/2013/08/27/rails-autoloading-hell/) explains this autoloading quirk well.

Both the `controller` and the `service` have the same namespace (`Foo`), so instead of referencing it as `Foo::Service` in the controller, it should just be `Service`.

lib/generators/module/templates/gemspec.erb
```diff
@@ -0,0 +1,34 @@
...
+  spec.name        = '<%= class_name.downcase %>'
+  spec.version     = <%= class_name %>::VERSION
+  spec.authors     = ['<%= `git config user.name`.squish %>']
+  spec.email       = ['<%= `git config user.email`.squish %>']
+  spec.homepage    = 'https://api.va.gov'
+  spec.summary     = 'An api.va.gov module'
+  spec.description = 'This module was auto-generated please update this description'
+  spec.license     = 'MIT'
+
+  # Prevent pushing this gem to RubyGems.org. To allow pushes either set the 'allowed_push_host'
```
jholton: Should we remove this section?

kreek:
@annaswims generated specs for the generated code or more specs for the generator itself?

Gemfile
```diff
@@ -34,8 +35,11 @@ gem 'clam_scan'
 gem 'config'
 gem 'connect_vbms', git: 'https://github.com/department-of-veterans-affairs/connect_vbms.git', branch: 'master'
 gem 'date_validator'
+gem 'dry-struct'
```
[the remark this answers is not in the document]
kreek: @jholton is ahead of the game 😃https://github.com/department-of-veterans-affairs/va.gov-team/issues/466

annaswims:
specs for the generated code.

PR #3126, opened 2019-06-25
Add draft drive time bands endpoint spec to v1

nfasulo:
Adding specification for `/drive_time_bands` to v1 of Facilities API. I included language in the description that indicates that is only a draft and the endpoint is not actually available, which will need to be removed once there is an actual endpoint.

Resolves department-of-veterans-affairs/vets-contrib/issues/2384.


modules/va_facilities/README.v1.yml
```diff
@@ -233,6 +233,265 @@ paths:
...
+      description: |
+        *This is a draft specification and not an available endpoint.*
+        
+        Retrieve the polygons used to determine in which drive time band veterans fall into for specified facilities. 
+
+        One can retrieve multiple facilities by id using this endpoint by making a
+        request with a comma-separated list of ids like `?ids=id1,id2`. When requesting facilities
+        in bulk by `id`, the API will return as many ids as it can find matches for and omit any
+        ids where there is no match. It will not return an HTTP error code if it is unable to match
+        a requested `id`. Clients may supply ids up to the limit a their HTTP client enforces for
```
rtravitz: "a their HTTP client"

modules/va_facilities/README.v1.yml
```diff
@@ -233,6 +233,265 @@ paths:
...
+        URI path lengths -- usually 2,048 characters.
+
+        Alternatively, one can retrieve a single facility using /drive_time_bands/{id}.
```
rtravitz: Backticks around route?

modules/va_facilities/README.v1.yml
```diff
@@ -568,6 +814,168 @@ components:
...
+                  type: array
+                  items:
+                    $ref: "#/components/schemas/DriveTimeBand"
+    DriveTimeBand:
+      description: a drive time band
+      type: object
+      properties:
+        name:
+          type: string
+          example: "drive_time_band_30"
```
rtravitz: Does it add extra clarity to have a name string when number/unit convey roughly the same information?

modules/va_facilities/README.v1.yml
```diff
@@ -233,6 +233,265 @@ paths:
...
+            application/json:
+              schema:
+                $ref: "#/components/schemas/AuthorizationError"
+        '406':
+          description: Requested format unacceptable
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/APIError"
+  /drive_time_bands/{id}:
```
rtravitz: Should we point them to `/drive_time_bands` for more information? Will someone be confused if they open this endpoint first? Same question for the `/all` variant, too.

modules/va_facilities/README.v1.yml
nfasulo: I think this is needed mostly for the first element in the array which is the lat/long of the facility itself. `{ value: 0, unit: "minutes" }` doesn't really tell you what that element is.

modules/va_facilities/README.v1.yml
nfasulo: Fixed this in v0 as well.

nfasulo:
> Left a few comments. My main question is: should we provide all of the facility information for this endpoint? That seems more the purpose of the `/facilities` endpoint. In this case would it make sense to just give back id, name, location, and the polygon?

@rtravitz I agree that we don't need to provide all of the attributes of a facility for either this endpoint or nearby. I have been relying the assumption that the array of required properties in the Facility schema which includes that only `name`, `facility_type`, `lat`, and `lng` means that we aren't promising that a returned facilty will have any attributes other than those. That way we can use the same schema for the main endpoint and these new ones and not have to determine exactly which of the attributes we consider non-essential for any particular endpoint.

Maybe I'm misreading the meaning of `required` or maybe that is just too cute of a use of it and not fair to consumers in its vagueness. Either way if I am open to creating a separate schema for indicating what an abbreviated facility object would look like if folks think that makes more sense.

rtravitz:
> I have been relying the assumption that the array of required properties in the Facility schema which includes that only `name`, `facility_type`, `lat`, and `lng` means that we aren't promising that a returned facilty will have any attributes other than those. That way we can use the same schema for the main endpoint and these new ones and not have to determine exactly which of the attributes we consider non-essential for any particular endpoint.

@nfasulo Oh, this completely changes my understanding of this spec! While it's great (and right) that the facility schema should be reusable, the subtlety around required was not apparent to me at all. I looked at the generated sample response and assumed that, that indicated which fields were coming back.

> Maybe I'm misreading the meaning of `required` or maybe that is just too cute of a use of it and not fair to consumers in its vagueness. Either way if I am open to creating a separate schema for indicating what an abbreviated facility object would look like if folks think that makes more sense.

While we open ourselves up to accidental inconsistencies and more work by having multiple schemas, I think the clarity tradeoff for the end users is worth it. At least as long as we have Swagger UI rendering the Open API spec in the way that it does!

rtravitz:
It looks like there are still non-required attributes that come back with `FacilityBasic`. Should we return address and services for this pared down response? 

Also in general, do we have any way for users to filter out non-required response fields? Or we're just saying that we reserve the option to not return those fields at various points?

nfasulo:
> It looks like there are still non-required attributes that come back with `FacilityBasic`. Should we return address and services for this pared down response?

I have been thinking of it in terms of what data would be useful if you needed to make a map of facilities. I figure address for displaying in a card and services for possible filtering. I've been living with this assumption since we started working planning `/nearby`. Maybe it does make more sense to go completely minimalist and then this pared down response can be used by any facility resource that is exposing data outside of the main facility dataset. That offloads more work onto consumers, but it is probably better to make the less assumptions about their use cases.

> Also in general, do we have any way for users to filter out non-required response fields? Or we're just saying that we reserve the option to not return those fields at various points?

No, users can't filter which fields they are going to get returned. I take it more to mean: we don't own this facility data so we can't guarantee for its completeness, caveat emptor.

rtravitz:
> Maybe it does make more sense to go completely minimalist and then this pared down response can be used by any facility resource that is exposing data outside of the main facility dataset. That offloads more work onto consumers, but it is probably better to make the less assumptions about their use cases.

🚀 I like this approach a lot. Then we just provide lighter utility endpoints and drive back to `/facilities` for the full data.

davem-va:
> > Maybe it does make more sense to go completely minimalist and then this pared down response can be used by any facility resource that is exposing data outside of the main facility dataset. That offloads more work onto consumers, but it is probably better to make the less assumptions about their use cases.
> 
> 🚀 I like this approach a lot. Then we just provide lighter utility endpoints and drive back to `/facilities` for the full data.

Yes, I agree with this approach.  Thanks all.

modules/va_facilities/README.v1.yml
```diff
@@ -726,11 +726,22 @@ paths:
...
-      summary: Retrieve a specific facility by ID
+      summary: Retrieve polygons of the drive time bands for a facility by ID
       description: |
         *This is a draft specification and not an available endpoint.*
+        
+        Retrieve the polygons used to determine in which drive time band veterans fall into for a specified facility. 
+
+        Alternatively, one can retrieve drive time bands for all facilities using `/facilities/all/drive_time_bands`.
+
+        The `ids` parameter is *required* to query this API. Requests without one or the other will return `400 Bad Request`.
```
rtravitz: The latest commit says it removes the ids query from the spec, but I'm not seeing that reflected here. Is that correct?

modules/va_facilities/README.v1.yml
```diff
@@ -726,11 +726,22 @@ paths:
...
+
+        Results of this operation are paginated. JSON responses include pagination information
+        in the standard JSON API "links" and "meta" elements. GeoJSON responses include pagination
+        information in the "Link" header.
 
-        See `/drive_time_bands` for more information.
+        The facility information returned by `/drive_time_bands` is a small subset of the information returned by `/facilities`.
+        Please use `/facilities` to detailed information.
```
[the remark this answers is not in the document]
nfasulo: I'm trying to tell the consumer to make a request to `/facilities`to get detailed information for a facility. What if I change this to:
>Please use `/facilities` to request more detailed information on facilities.

modules/va_facilities/README.v1.yml
nfasulo: I don't know how I managed to not remove this route.

modules/va_facilities/README.v1.yml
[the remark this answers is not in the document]
rtravitz: Either/or! Just a minor comment about what seems like a missing word and not any problem with what it communicates.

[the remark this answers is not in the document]
nfasulo: Ugh, yeah, that's gives you an idea of my reading comprehension this morning. I'm just not actually understanding.

PR #3129, opened 2019-06-27
Reference URL fixes in API docs

jeff2d2:
## Description of change
Fixes https://github.com/department-of-veterans-affairs/vets-contrib/issues/2527
Noticed that all of our API docs that contain a `Reference` section with the `Raw Open API Spec` URLs have links for *dev* instead of *prod*. Also some were `http` instead of `https`.

- [x] All reference URLs use `https`
- [x] All reference URLs point to `prod`

#### Applies to all PRs

- [x] Appropriate logging

### DIFF
--- a/modules/appeals_api/README.yml
+++ b/modules/appeals_api/README.yml
@@ -32,7 +32,7 @@ info:
 
       ## Reference
 
-      Raw Open API Spec: http://dev-api.va.gov/services/appeals/docs/v0/api
+      Raw Open API Spec: http://api.va.gov/services/appeals/docs/v0/api
 
   termsOfService: ''
   contact:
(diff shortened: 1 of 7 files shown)


jeff2d2:
I noticed that Facilities is the only one that uses links instead of displaying the URL as text. Would that be preferred? It's the only one with multiple references, so maybe that's why it's that way and the others aren't?

dhall:
this means that viewing these from dev-developer.va.gov the reference will point to prod .. but I think that makes more sense then prod pointing to dev. I don't feel strongly that we need to find a way to make them environment aware so dev points to dev and prod points to prod.. all that to say -- thanks, this looks good to me :)

PR #3133, opened 2019-06-28
Claims api healthcheck

charleystran:
## Description of change
This change is to add a health check endpoint to the V0 and V1 claims endpoints. These will get added into the blackbox exporter process.

## Testing done
Rspec and browser testing

- [x]  An endpoint for V0 is available to report on health
- [x]  An endpoint for V1 is available to report on health

#### Applies to all PRs
- [x] Appropriate logging
- [x] https://github.com/department-of-veterans-affairs/vets-contrib/issues/2530


modules/claims_api/app/models/claims_api/evss_claim.rb
```diff
@@ -56,5 +56,44 @@ def supporting_documents
         []
       end
     end
+
+    def self.services_are_healthy?
+      # TODO: we should add check for Okta and SAML Proxies being up as well
```
dvogel: This comment seems to lead us toward an unsustainable future because neither Okta nor the SAML proxy are direct dependencies of the API. They are definitely related, because you need an access token, which would come from interactions with both of those services. However, if we have every API concerned with auxillary dependencies, notifications will become a tangled mess. The practical impact of a notification would be less clear. For example, the API could be reported as out of service but still be able to serve requests carrying an access token that was generated before either component went down, so long as the access token had already been verified before okta went down.

modules/claims_api/app/models/claims_api/evss_claim.rb
```diff
@@ -56,5 +56,44 @@ def supporting_documents
...
+        errors: [
+          {
+            title: 'ClaimsAPI Unavailable',
+            detail: 'ClaimsAPI is currently unavailable.',
+            code: '503',
+            status: '503'
+          }
+        ]
+      }.to_json
+    end
```
dvogel: The `healthy_service_response` and `unhealthy_service_response` methods feel out of place in this model. They are specific to the needs of the `healthcheck` controller action. They don't have any fundamental relationship to anything else in the model(s) and aren't theoretically re-usable. Moving them to the controller or a serializer seems more appropriate to me.

modules/claims_api/app/models/claims_api/evss_claim.rb
charleystran: @dvogel You are correct that once an initial token is created, we may not need to communicate with the SAML proxy or Okta, unless a refresh token is being used and it is time for a refresh. The reason we added this comment is to be able to mark our services as down (and potentially degraded on the status page) when a service we depend on is down. However it brings up another question on whether the SAML proxy will respond gracefully when Okta cannot be reached. @dvogel do you know how that would be handled?

edmangimelli:
NOFM

modules/claims_api/app/models/claims_api/evss_claim.rb
mjlumetta: @charleystran the SAML proxy will still be reachable if Okta cannot be reached, but the OAuth proxy will not

modules/claims_api/app/models/claims_api/evss_claim.rb
```diff
@@ -56,5 +56,44 @@ def supporting_documents
...
+
+    def self.healthy_service_response
+      {
+        data:  {
+          id: 'claims_healthcheck',
+          type: 'claims_healthcheck',
+          attributes: {
+            healthy: true,
+            date: Time.zone.now.to_formatted_s(:iso8601)
+          }
```
mjlumetta: should these attributes (every key in `data`, actually) also be in the `unhealthy_service_response`? they still seem applicable

charleystran: Hey @mjlumetta they do not seem to be present on the error response of the appeals healthcheck error response. Which was our starting point. @bastosmichael and @dvogel what do you think? I am open to adding them.

modules/claims_api/app/models/claims_api/evss_claim.rb
charleystran: @mjlumetta I am not 100% sure what you are voting for in this case. Our position was that since the saml proxy and okta are dependencies of V1 of the claims API, we should add them into our overall healthchecks. We would do this separately as 2 different checks.

mjlumetta: no position on that! i was trying to answer your question about whether the SAML proxy will respond gracefully if Okta is down (yes). but i'm adding the clarification that the OAuth proxy, which lives in the same repo as the SAML proxy, will not be up if Okta is down. does that help?

modules/claims_api/app/models/claims_api/evss_claim.rb
mjlumetta: yeah, that makes sense if that's what you're working from, and i know pretty much nothing about the specific context, so i defer to whatever you think is best. i guess my position on this is this (and it's like, a tiny comment): if you're returning both of these JSON structures from the same endpoint, then i don't get the value of `data.attributes.healthy`; when would you ever check it? the healthy vs unhealthy check for a consumer would be satisfied by the 200 vs 503 response and `res.data` vs `res.errors` as the top level key, so a `healthy` boolean nested that deep doesn't add anything. similarly, why have an `id`/`type` field in only one fork of the response? either i need that info to process at some point, in which case i might need it for healthy or unhealthy use cases, or i don't (bc i know that i'm calling the healthcheck endpoint, so i can just decide how to identify that as a consumer). 

it's very possible that there are specific reasons for these choices and that's totally cool if there are. but it's difficult to understand what this means in context intuitively because there's a lot of divergence between responses.

modules/claims_api/app/models/claims_api/evss_claim.rb
charleystran: @mjlumetta what does gracefully mean in your words? I since its hard to force the error, I am looking to see what kind of response we can expect if Okta is down. If Okta is down, what is the response code we would receive?

mjlumetta: yeah, so... not sure what endpoint you're going to use as a healthcheck or what request you'll make. but the SAML proxy doesn't depend on Okta in and of itself. it does **redirect** to Okta, so i think you would want to test for a 302 rather than a 200 if you're going through the flow. or if there is a health check endpoint already that i don't know about/forgot about that can return 200, i would assume that would be unaffected.

modules/claims_api/app/models/claims_api/evss_claim.rb
charleystran: @mjlumetta  the data structure was put in place to mimic the appeals healthcheck which we don't control. Which a passthrough to a VA endpoint. The goal was consistency, but there is nothing really tying this code to it other than having the healthchecks be consistent The ID/Type are in place for json api compliance.

modules/claims_api/app/models/claims_api/evss_claim.rb
charleystran: these were moved out of the model and actually the check was moved out of the individual controllers since both v0 and v1 provide the same checks as of now, there was no need to duplicate the logic.

modules/claims_api/app/models/claims_api/evss_claim.rb
charleystran: I commented on the wrong item, but essentially these were moved out into the controller.

bastosmichael:
I wanted to start the conversation based on @dvogel's comment from https://github.com/department-of-veterans-affairs/vets-api/pull/3133#discussion_r299616572 but I don't see a point for a healthcheck unless you at least test hard dependencies. In the case of MVI or EVSS, those are being tested because without MVI neither Claims or the SAML proxy will work because we use it to validate the existence of the veteran and without EVSS we simply wouldn't have the data necessary to give the end user since Claims is merely an EVSS pass through. There is essentially nothing that our API gives that isn't required by those two end points so testing their up time in my mind was the best approach to doing the healthcheck since if vets-api is down, then the healthcheck is down anyhow. The SAML proxy and Okta was more of a concern that with Okta being down our system wouldn't be able to decode the JWT tokens and or the end consumer wouldn't be able to generate refresh tokens so in a sense the if Okta was down the user couldn't really still make submissions and or requests to Claims if mostly for the 15 minute window before a token expired which is why I brought it up to Charlie to take those into consideration.

modules/claims_api/app/models/claims_api/evss_claim.rb
mjlumetta: cool, that makes sense! i mean, as long as there's a reason. thanks for looking into it, Charley

PR #3137, opened 2019-07-01
Move PPIU Direct Deposit confirmation email to a background job

jholton:
## Description of change
Move the PPIU Direct Deposit confirmation email to a background job so that any mail failures don't disrupt the overall Direct Deposit update request

## Testing done
specs, local testing

## Testing planned
dev (or staging) to verify delivery via a Sidekiq job

#### Unique to this PR

- [x] Move delivery of Direct Deposit email to a sidekiq job

#### Applies to all PRs

- [x] Appropriate logging
- ~~[ ] Swagger docs have been updated, if applicable~~
- ~~[ ] Provide which alerts would indicate a problem with this functionality (if applicable)~~


app/workers/direct_deposit_email_job.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class DirectDepositEmailJob
+  include Sidekiq::Worker
+  sidekiq_options expires_in: 1.day
```
jholton: My thinking here is that if the email isn't successfully sent within 24 hours, don't keep trying.
If I updated my info, but received a confirmation email 4 days later, I might not associate it with my update 4 days earlier.  I'd be confused and worried that a nefarious actor had updated it after me.

PR #3139, opened 2019-07-01
18950 dont retry ep codes

SMLuthi:
## Description of change

EVSS has confirmed that, on a form526 submission, if an `EP Code is not valid` service error is raised we should not be retrying them because no amount waiting will allow the submission to go through successfully. Currently all upstream service errors force the submission job to retry until timeout or success so this case needs to have an exception in the logic.

## Testing done

spec tests.

## Testing planned

Attempt on staging... but hard to reproduce.

#### Unique to this PR

- [x] Create a non-retry exception for `EP is not valid` error

#### Applies to all PRs

- [x] Appropriate logging

### DIFF
--- a/app/workers/evss/disability_compensation_form/submit_form526.rb
+++ b/app/workers/evss/disability_compensation_form/submit_form526.rb
@@ -54,19 +54,26 @@ module EVSS
         raise NotImplementedError, 'Subclass of SubmitForm526 must implement #service'
       end
 
-      # Retries any errors caused by an upstream service from EVSS being unavailable or
-      # if getting a "PIF in use" error
+      # Retries any errors caused by an upstream service from EVSS being unavailable (unless the
+      # external serivce is caused by an invalid EP code) or if getting a "PIF in use" error
       # Otherwise it marks it as non-retryable and stops the job
       #
       # @param error [ErrorClass] An exception object of type {EVSS::DisabilityCompensationForm::ServiceException}
       #
       def retry_form526_error_handler!(error)
-        if error.key == 'evss.external_service_unavailable' ||
+        if (error.key == 'evss.external_service_unavailable' && ep_code_valid(error)) ||
            error.key == 'evss.disability_compensation_form.pif_in_use'
           retryable_error_handler(error)
         else
           non_retryable_error_handler(error)
         end
+      end
+
+      def ep_code_valid(error)
+        error.messages.each do |message|
+          return false if message['text'].include?('EP Code is not valid')
+        end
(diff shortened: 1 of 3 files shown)


app/workers/evss/disability_compensation_form/submit_form526.rb
```diff
@@ -54,20 +54,27 @@ def service(_auth_headers)
...
         else
           non_retryable_error_handler(error)
         end
       end
+
+      def ep_code_valid(error)
+        error.messages.each do |message|
+          return false if message['text'].include?('EP Code is not valid')
+        end
+        true
```
dogweather: Can these four lines can be replaced with this simpler one-liner?

```ruby
error.messages.none? { |msg| msg['text'].include?('EP Code is not valid') }
```

app/workers/evss/disability_compensation_form/submit_form526.rb
```diff
@@ -54,20 +54,27 @@ def service(_auth_headers)
         raise NotImplementedError, 'Subclass of SubmitForm526 must implement #service'
       end
 
-      # Retries any errors caused by an upstream service from EVSS being unavailable or
-      # if getting a "PIF in use" error
+      # Retries any errors caused by an upstream service from EVSS being unavailable (unless the
+      # external serivce is caused by an invalid EP code) or if getting a "PIF in use" error
       # Otherwise it marks it as non-retryable and stops the job
       #
       # @param error [ErrorClass] An exception object of type {EVSS::DisabilityCompensationForm::ServiceException}
```
dogweather: I think this could be simplified to:

```ruby
# @param error [EVSS::DisabilityCompensationForm::ServiceException]
```

spec/jobs/evss/disability_compensation_form/submit_form526_increase_only_spec.rb
```diff
@@ -103,6 +103,17 @@
...
     end
 
+    context 'with an upstream service error for EP code not valid' do
+      it 'sets the transaction to "non_retryable_error"' do
+        VCR.use_cassette('evss/disability_compensation_form/submit_500_with_ep_not_valid') do
+          subject.perform_async(submission.id)
+          expect_any_instance_of(EVSS::DisabilityCompensationForm::Metrics).to receive(:increment_non_retryable).once
+          described_class.drain
+          expect(Form526JobStatus.last.status).to eq Form526JobStatus::STATUS[:non_retryable_error]
+        end
```
dogweather: Nice!

PR #3140, opened 2019-07-01
18719 undefined method dig error

SMLuthi:
## Description of change

Sentry has been logging issues for when theres a `get_letters` request to EVSS. Some service errors returned are coming back as string and not json bodies. Normally, we would handle this as a default BackendServiceException but there is a `dig` method call being made on the error body that is a string causing cascading failures. This has been refactored to avoid this problem in the future.

## Testing done

spec tests

#### Unique to this PR

- [x] Move the `log_edipi` logic into the `handle_error` functionality (since it does the same checks anyways, including a check that makes sure the var is a Hash).

#### Applies to all PRs

- [x] Appropriate logging

### DIFF
--- a/lib/evss/letters/service.rb
+++ b/lib/evss/letters/service.rb
@@ -50,9 +50,12 @@ module EVSS
 
       def handle_error(error)
         if error.is_a?(Common::Client::Errors::ClientError) && error.status != 403 && error.body.is_a?(Hash)
-          log_edipi if invalid_address_error?(e)
-          save_error_details(error)
-          raise EVSS::Letters::ServiceException, error.body
+          begin
+            log_edipi if invalid_address_error?(error)
+          ensure
+            save_error_details(error)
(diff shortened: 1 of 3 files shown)


lib/evss/letters/service.rb
```diff
@@ -66,8 +66,7 @@ def log_edipi
       end
 
       def invalid_address_error?(error)
-        return false unless error.is_a?(Common::Client::Errors::ClientError)
-        error&.body&.dig('messages')&.any? { |m| m['key'].include? INVALID_ADDRESS_ERROR }
+        error.body.dig('messages')&.any? { |m| m['key'].include? INVALID_ADDRESS_ERROR }
```
lihanli: if body is a string isn't dig still going to fail?

SMLuthi: Yea, but the call made to `invalid_address_error?` is within an error check that makes sure its a `Hash`

PR #3145, opened 2019-07-05
Flipper demo

annaswims:
## Description of change
This PR is not intended to be merged, its a proof-of concept for a feature flipper gem called [flipper](https://github.com/jnunemaker/flipper).  

Flipper is build with a number of optional gems that add features.  
### Gates
we can gate a feature 
### data store
Flipper can be [backed with](https://github.com/jnunemaker/flipper/blob/master/docs/Adapters.md) a variety of adapters, but postgresql or redis seem like the most likely one we'd use.  Flipper also supports caching in memory, for better performance. 

### UI
Flipper has a [UI](https://github.com/jnunemaker/flipper/blob/master/docs/ui/README.md) that we could use, though it's also possible to toggle/configure features programatically.  We can secure the API and UI however we wish. 
Screen Shot 2019-07-05 at 11 37 10 AM

### API
Should it become desirable in the future, we can make calls to flipper to configure flipper or evaluate a feature's availability with the api, though I think initially we can wrap the flipper evaluations into the user's list of services.

edit: I added a proof-of-concept endpoint that will return a boolean (within the context of the user, if the user is authenticated) . http://localhost:3000/v0/feature_flags/another_toggle,flipper_demo 
`{"another_toggle":true,"flipper_demo":false}`


annaswims:
To run locally
* `bundle install
* `bundle exec rake db:migrate`
* View feature toggles at http://localhost:3000/flipper/features 

To verify that the feature toggle is working look at the services list on the user endpoint. (only flipper_demo is configured with flipper at this point.)
Screen Shot 2019-07-05 at 1 48 32 PM

rianfowler:
This looks good. Few questions: 

- Does flipper have a js client? 
- Does it support stickiness e.g. if we want to a/b test a feature, then a specific user should get the same variant every time they refresh until they clear their cookies, cache, etc?
- Does it have tracking e.g. can I see how many users have gotten a specific variant? 
- Does it have feature gating in its admin e.g. can we limit who has access to a specific feature flag based on their login?

rianfowler:
One more question: it's not clear to me how the feature flag information is provided to the client. Does this provide a web service that `vets-website` could hit and what does the response look like if it does. 

One of the issues I had with [unleash](https://github.com/Unleash/unleash) is that the server only provided a description of the feature flag and it's enabled state and the _client_ has to interpret that into a behavior. e.g. if you set a percentage based roll out, it would only tell you the ratio and the client has to implement the randomization / final determination of the flag state. It also only provides server side clients (including nodjs which we could theoretically copy into a browser client).

annaswims:
Great questions, @rianfowler 

* Does flipper have a js client?
No, there's no js client.  I'm imagining we would add an endpoint on the be to be consumed by the FE, or roll it into the user endpoint like we do for the list of services available to a user. 

* Does it support stickiness e.g. if we want to a/b test a feature, then a specific user should get the same variant every time they refresh until they clear their cookies, cache, etc?
The "percentage of logged in users" uses an algorithm that will select the same x% of users, as long as the percentage is not decreased. But, otherwise we'd have to keep track of it separately.  There is a method we could call to enable the feature for  each uuid though.

* Does it have tracking e.g. can I see how many users have gotten a specific variant?
We get some metrics through statsD/grafana, but we'd likely want to add it to Google Analytics also.

* Does it have feature gating in its admin e.g. can we limit who has access to a specific feature flag based on their login?
No.  We can configure access to the feature flag with a method, but there's no way to limit access to specific flags out of the box. We can prevent users from adding/deleting feature flags through the UI though.

kreek:
LGTM, I think there was one feature flipper that had both Ruby and JS clients? (I might be thinking of Launch Darkly) For open source lib like this I think an endpoint consumed by the FE is a good solution.

annaswims:
closing demo branch.

PR #2905, opened 2019-03-21
Status page: Creates Vets-API external service statuses endpoint

hpjaj:
## Description of change
**User Story:** As a FE I want an endpoint that I can hit to get the current status of our external services so that I can know the current state of our services.

## Testing done

Automated spec coverage

## Testing planned

Testing in staging

#### Unique to this PR

- [x] Creates a new `GET /v0/backend_statuses` endpoint to fetch current external service statuses
- [x] Swagger docs and specs
- [x] Spec coverage
- [x] Extends `:raise_error` middleware to recognize more custom exception keys

#### Applies to all PRs

- [x] Appropriate logging
- [x] Swagger docs have been updated, if applicable
- [x] Provide link to originating GitHub issue, or connected to it via ZenHub
- [x] Does not contain any sensitive information (i.e. PII/credentials/internal URLs/etc., in logging, hardcoded, or in specs)
- [x] Provide which alerts would indicate a problem with this functionality (if applicable)

#### New `/v0/backend_statuses` endpoint swagger docs

image

#### New `/v0/backend_statuses` endpoint swagger response model details

image

PR #3155, opened 2019-07-12
Add short name to backend status

annaswims:
## Description of change
Add an attribute service_id to the backend_status.  That way we can can find the status using the same key as the `maintenance_windows` endpoint.

Also removed authentication requirement from `backend_statuses` . I'm not clear why it would be an authenticated resource if its intended to be used in a status page. Much of the information is already available to unauthenticated users in `maintenance_windows`

Not all services have [mapping](https://github.com/department-of-veterans-affairs/devops/blob/e8441ce11de217858f79be719f494d9f00b67bcb/ansible/deployment/config/vets-api/prod-settings.local.yml.j2#L232) at this time, but all those that poll for windows currently are. 

## Testing done
I created a personal api token in pager duty and added it to  (My Profile> user Settings > Create API token)  and added it to `settings.local.yml`
````
maintenance:
  pagerduty_api_token: MY TOKEN
````

and viewed http://localhost:3000/v0/backend_statuses

#### Applies to all PRs

- [x] Appropriate logging

### CODE REVIEW

**Comment by rtluu** [comment]
@gunsch, @batemapf a question came up in relation to whether or not the backend statuses need to be hidden behind authentication. 

The general sentiment from the tools team is that it does not have to be gated behind login, but wanted to pose to you both for consideration. As a side note, the maintenance windows are publicly accessible for our downtime banners, which is in part why this information seems to be kosher for public access.

**Comment by gunsch** [comment]
@rtluu My understanding was that there's no reason to hide this --- that while the intended audience is "VA.gov stakeholders" rather than "the general public", there's also no reason to make it auth-protected.

**Comment by batemapf** [comment]
Concur w/ @gunsch.

**Comment by annaswims** [comment]
@jholton can you give this one a look when you have some time?

**Comment by annaswims** [comment]
oops, I see that merging master caused a spec to fail,  I'll take a look before you need to review.

File: app/controllers/v0/backend_statuses_controller.rb [hunk 1]
```diff
@@ -2,8 +2,10 @@
 
 module V0
   class BackendStatusesController < ApplicationController
-    skip_before_action :authenticate, only: [:show]
+    skip_before_action :authenticate, only: %i[index show]
 
+    # Note: this endpoint is somewhat misleading.  Index gets data from PagerDuty and
+    # show only looks at GI bill scheduled downtime (and gets no data from PagerDuty)
```
jholton: [Here's](https://github.com/department-of-veterans-affairs/vets-api/pull/2905#issuecomment-475626457) how we ended up in this place.  Agreed that `show` should be refactored in the future to exist more cohesively with `index` functionality.

app/controllers/v0/backend_statuses_controller.rb
```diff
@@ -2,8 +2,10 @@
 
 module V0
   class BackendStatusesController < ApplicationController
-    skip_before_action :authenticate, only: [:show]
+    skip_before_action :authenticate, only: %i[index show]
```
jholton: Should we remove `only: %i[index show]` altogether?

PR #3156, opened 2019-07-15
Change mvi log level from error to info

omgitsbillryan:
### Context
https://github.com/department-of-veterans-affairs/va.gov-team/issues/271
In an effort to make our Sentry Alerting less noisy, we've decided to change the logging levels of connection problems to backend services from `ERROR` to `INFO`.

### Description
This changes two places where we log MVI connection problems as `ERROR` to `INFO`. [Sentry Alerted us](https://dsva.slack.com/archives/CJTDG22NM/p1563082714001000) of [`MVI find_profile connection failed.`](http://sentry.vfs.va.gov/vets-gov/platform-api-production/issues/46783)

## Testing done
Local rspecs.

## Testing planned
None.

#### Applies to all PRs

- [x] Appropriate logging

### DIFF
--- a/lib/mvi/service.rb
+++ b/lib/mvi/service.rb
@@ -51,10 +51,10 @@ module MVI
       end
     rescue Breakers::OutageException => e
       Raven.extra_context(breakers_error_message: e.message)
-      log_console_and_sentry('MVI find_profile connection failed.', :info)
+      log_console_and_sentry('MVI find_profile connection failed.', :warn)
       mvi_profile_exception_response_for('MVI_503', e)
     rescue Faraday::ConnectionFailed => e
-      log_console_and_sentry("MVI find_profile connection failed: #{e.message}", :info)
+      log_console_and_sentry("MVI find_profile connection failed: #{e.message}", :warn)
       mvi_profile_exception_response_for('MVI_504', e)
     rescue Common::Client::Errors::ClientError, Common::Exceptions::GatewayTimeout => e
       log_console_and_sentry("MVI find_profile error: #{e.message}", :error)
--- a/lib/mvi/service.rb
+++ b/lib/mvi/service.rb
@@ -51,10 +51,10 @@ module MVI
       end
     rescue Breakers::OutageException => e
       Raven.extra_context(breakers_error_message: e.message)
-      log_console_and_sentry('MVI find_profile connection failed.', :error)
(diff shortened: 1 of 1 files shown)


lib/mvi/service.rb
```diff
@@ -51,10 +51,10 @@ def find_profile(user_identity)
...
       Raven.extra_context(breakers_error_message: e.message)
-      log_console_and_sentry('MVI find_profile connection failed.', :error)
+      log_console_and_sentry('MVI find_profile connection failed.', :info)
       mvi_profile_exception_response_for('MVI_503', e)
     rescue Faraday::ConnectionFailed => e
-      log_console_and_sentry("MVI find_profile connection failed: #{e.message}", :error)
+      log_console_and_sentry("MVI find_profile connection failed: #{e.message}", :info)
       mvi_profile_exception_response_for('MVI_504', e)
     rescue Common::Client::Errors::ClientError, Common::Exceptions::GatewayTimeout => e
       log_console_and_sentry("MVI find_profile error: #{e.message}", :error)
```
annaswims: whatever is decided, this should be the same level as the others I would think.

omgitsbillryan:
Should also mention that this is already monitored in grafana:
http://grafana.vfs.va.gov/dashboard/db/mvi?orgId=1&from=1562984568497&to=1563127926693

And has an alerting rule in Prometheus + PagerDuty:
https://dsva.pagerduty.com/incidents/P6C5CQB

lib/mvi/service.rb
omgitsbillryan: I tend to agree and had originally done a slight refactor to make `Common::Exceptions::GatewayTimeout` as `INFO` as well (but not `Common::Client::Errors::ClientError`). 

However, I then [searched the error string](http://sentry.vetsgov-internal/vets-gov/platform-api-production/?query=is%3Aunresolved+MVI+find_profile+error) and I'm not confident that I _wouldnt_ want to be alerted on at least some of the following:
image

I also can't figure out a sure-fire way of determining which of these error messages maps to which exception type other then using my gut.

kfrz: `:warn` makes more sense to me :+1:

omgitsbillryan:
For auditing, this connects: https://github.com/department-of-veterans-affairs/va.gov-team/issues/577

PR #3158, opened 2019-07-17
VBADocuments Submission Status Caching

ATeal:
## Description of change
In order to make the reports endpoint function quicker, deal with timeouts, and various other reasons we decided rather than hit central mail on every request to reports, we would instead asynchronously sync that status of submissions and return the current value. Allowing the consumer to manually update individual status if an immediate update is necessary.

although closed, resolves 100% https://github.com/department-of-veterans-affairs/vets-contrib/issues/930
resolves https://github.com/department-of-veterans-affairs/vets-contrib/issues/2676
resolves https://github.com/department-of-veterans-affairs/vets-contrib/issues/2578

## Testing done
- rspec

#### Unique to this PR
- change in ideology in how we handle statuses for payloads

#### Applies to all PRs

- [x] Appropriate logging

### DIFF
--- a/modules/vba_documents/app/controllers/vba_documents/v0/reports_controller.rb
+++ b/modules/vba_documents/app/controllers/vba_documents/v0/reports_controller.rb
@@ -12,7 +12,7 @@ module VBADocuments
       ID_PARAM = 'ids'
 
       def create
-        statuses = VBADocuments::UploadSubmission.refresh_and_get_statuses!(params[ID_PARAM])
+        statuses = VBADocuments::UploadSubmission.where(guid: params[ID_PARAM])
         render json: with_spoofed(statuses),
                each_serializer: VBADocuments::UploadSerializer
       end
--- a/modules/vba_documents/app/controllers/vba_documents/v0/reports_controller.rb
+++ b/modules/vba_documents/app/controllers/vba_documents/v0/reports_controller.rb
@@ -8,6 +8,7 @@ module VBADocuments
       skip_before_action(:authenticate)
       before_action :validate_params
 
+      MAX_REPORT_SIZE = 1000
       ID_PARAM = 'ids'
 
       def create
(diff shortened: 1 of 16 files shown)


modules/vba_documents/app/controllers/vba_documents/v0/reports_controller.rb
```diff
@@ -8,17 +8,23 @@ class ReportsController < ApplicationController
...
       before_action :validate_params
 
-      MAX_REPORT_SIZE = 100
+      MAX_REPORT_SIZE = 1000
       ID_PARAM = 'ids'
 
       def create
-        statuses = VBADocuments::UploadSubmission.refresh_and_get_statuses!(params[ID_PARAM])
-        render json: statuses,
+        statuses = VBADocuments::UploadSubmission.where(guid: params[ID_PARAM])
```
bastosmichael: we still probably want to set a limit here so that they can't put something like 10K ids and the payload returned becomes to much for us to serve.

modules/vba_documents/app/controllers/vba_documents/v1/reports_controller.rb
```diff
@@ -8,17 +8,23 @@ class ReportsController < ApplicationController
...
 
-      MAX_REPORT_SIZE = 100
+      MAX_REPORT_SIZE = 1000
       ID_PARAM = 'ids'
 
       def create
-        statuses = VBADocuments::UploadSubmission.refresh_and_get_statuses!(params[ID_PARAM])
-        render json: statuses,
+        statuses = VBADocuments::UploadSubmission.where(guid: params[ID_PARAM])
+        render json: with_spoofed(statuses),
```
dhall: help me understand the spoofed/fake status approach.. just curious

ATeal: So in the past we've added a not found status for any guids requested that we didn't have in the db.

Whats interesting is our inconsistency in how we handle expired payloads.... if you hit `/uploads/#{guid}` with an expired guid, in v0 it gives back the fake status with not found for the code, in v1 we straight up raise not found (giving a json api error not found). In reports though in this implementation and the previous, we actually return the record with a status of expired... I'll add that to the ticket for improving handling of expired payloads. 

~~ Just realized that while making sure I was correct in the implementation here ~~

PR #3166, opened 2019-07-22
Form 526 validation endpoint

charleystran:
## Description of change
This change adds the 526 validation endpoint. As well it calls the EVSS validation as part of the 526 creation process. 

## Testing done
Rspec specs

- [x] The validation endpoint is available for v0 and v1
- [x] The create 526 endpoint calls the validation logic

#### Applies to all PRs

- [x] Appropriate logging


modules/claims_api/app/controllers/claims_api/v0/forms/disability_compensation_controller.rb
```diff
@@ -9,9 +9,10 @@ module Forms
       class DisabilityCompensationController < BaseFormController
         FORM_NUMBER = '526'
         before_action :verification_itf_expiration, only: [:submit_form_526]
+        before_action :validate_526_payload, only: [:submit_form_526]
         skip_before_action(:authenticate)
         skip_before_action(:verify_power_of_attorney)
-        skip_before_action :validate_json_schema, only: [:upload_supporting_documents]
+        skip_before_action :validate_json_schema, only: %i[upload_supporting_documents]
```
ATeal: some inconsistency, this one uses `%i[` the other just uses `[`

modules/claims_api/app/controllers/claims_api/v0/forms/disability_compensation_controller.rb
```diff
@@ -42,14 +43,19 @@ def upload_supporting_documents
         end
 
         def validate_form_526
-          validate_form526(form_attributes)
-          render json: { data: 'success' }
+          valid_service = EVSS::DisabilityCompensationForm::ServiceAllClaim.new(auth_headers)
+          validation_result = valid_service.validate_form526(form_attributes.to_json)
+          render json: validation_result
```
ATeal: I would say eliminate the assignment here and just do `render json: valid_service.validate...`

modules/claims_api/app/controllers/claims_api/v0/forms/disability_compensation_controller.rb
```diff
@@ -42,14 +43,19 @@ def upload_supporting_documents
...
+          render json: { errors: format_errors(e.details) }, status: :unprocessable_entity
         end
 
         private
 
-        def validate_form526(form_data)
-          service(auth_headers).validate_form526(form_data)
+        def format_errors(errors)
+          errors.map do |error|
+            { status: 422, detail: "#{error['key']} #{error['detail']}", source: nil }
```
ATeal: is there anypoint to doing `source: nil` could it be `source: error['key']`?

modules/claims_api/app/controllers/claims_api/base_form_controller.rb
```diff
@@ -52,5 +52,12 @@ def form_attributes
 
       payload_attributes
     end
+
+    def validate_526_payload
+      valid_service = EVSS::DisabilityCompensationForm::ServiceAllClaim.new(auth_headers)
```
ATeal: Maybe name this just service? `valid_service` is kind of a misnomer, or go full bore and call it `validate_service`

lib/evss/disability_compensation_form/service.rb
```diff
@@ -59,12 +59,40 @@ def validate_form526(form_content)
...
+            # when evss returns invalid, the middleware will catch the error and it will be rescued
+            # by the caller. This is the fallback response for an unknown error response
+            error_response
+          end
         end
       end
 
       private
 
+      def valid_response
```
ATeal: do we typically handle the conversion to json-api formatted responses in the service? Seems like something we handle at the controller level

modules/claims_api/app/controllers/claims_api/v0/forms/disability_compensation_controller.rb
charleystran: That or possibly external? What do you think?

modules/claims_api/app/controllers/claims_api/v0/forms/disability_compensation_controller.rb
charleystran: The only reason for the assignment is actually for Rubocop line length

modules/claims_api/app/controllers/claims_api/v0/forms/disability_compensation_controller.rb
charleystran: This is because our rubocop rules require %i if there is more than one item, so I updated the single item arrays to also use %i

ATeal: yeah that sounds good to me!

modules/claims_api/app/controllers/claims_api/base_form_controller.rb
```diff
@@ -52,5 +52,29 @@ def form_attributes
...
     end
+
+    def validate_526_payload
+      service = EVSS::DisabilityCompensationForm::ServiceAllClaim.new(auth_headers)
+      service.validate_form526(form_attributes.to_json)
+    rescue EVSS::ErrorMiddleware::EVSSError => e
+      render json: { errors: format_errors(e.details) }, status: :unprocessable_entity
+    end
+
+    def valid_526_response
```
ATeal: I think this is fine for now considering we only have one form, but we may need one more layer of inheritence given this is 526 specific stuff, or break it out into modules and include the base form. Definitely not a now concern but something to think about in the future

PR #3171, opened 2019-07-24
Update state cemeteries

rtravitz:
## Description of change

This PR updates the file that provides State Cemetery information to the Facilities API. The changes will be picked up in a job that run every 24 hours. It came from [#2447](https://app.zenhub.com/workspaces/developervagov-5ac272fff41bfa50c0e46c48/issues/department-of-veterans-affairs/vets-contrib/2547).

## Testing done

Ran job locally to confirm that the changes in the file did not cause any issues.

## Testing planned

Will check lower environments a day after merge. If this causes a problem, alerts should trigger from Prometheus into the #ao-alerts channel in Lighthouse Slack.

#### Unique to this PR

- [x] The cems.xml is updated from the source

#### Applies to all PRs


rtravitz:
> I'm concerned that this is wiping out several lat/long entries .. this will cause issues for us. I recall this being an issue with the source data.. I'll see if I can dig up any history

Thanks for this catch! Wondering if we can push the group maintaining that sheet to update the source data so we don't need to make manual updates.

nfasulo:
@dhall I think this is what you are looking for: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10687#issuecomment-406614139

dhall:
@rtravitz  you are fast.. that's exactly what I was looking for!

nfasulo:
This would also be a reason not to blindly update this XML file on a schedule. Although maybe we can reach out and get this data added into the source.

rtravitz:
If I fix the regressions we have, would that be sufficient to get this update in  @dhall @nfasulo? And @davem-va here's where we're at. I can pull you a list of updates that would need to be made if we ever want to move to a programmatic solution 🚀

davem-va:
@rtravitz Can you please get me a list of updates we'd require to enable an automated/programmatic update solution?  Thanks.

rtravitz:
The four in this comment need to be made: https://github.com/department-of-veterans-affairs/vets.gov-team/issues/10687#issuecomment-406614139

And then here's a spreadsheet of all the other state cemeteries that need lat/long at some point: https://lighthouseva.slack.com/archives/CEXC8GV2S/p1564005671052700

davem-va:
Reached out to NCA data owner for all required changes.

rtravitz:
Okay, this update fixes the regressions. There are some new cemeteries currently under construction that don't have a lat/long, but those don't get pulled into our database. They're thrown out in the daily reload job.

lib/facilities/cemetery_data/cems.xml
```diff
@@ -302,7 +303,7 @@ contact2=""
 phone="707-944-4600"
 fax=""
 email=""
-lat="38.390938"
+lat=" 38.390938"
```
dhall: leading space.. not sure if that's a problem or not

rtravitz: Whoops. Fixing.

PR #3174, opened 2019-07-24
Update documentation for Facilities v1

nfasulo:
Making a couple of changes to Facilities v1 documentation:
- ~Update /nearby response so that it returns the `FacilityBasic` object instead of `Facility`~
- Remove draft drive time endpoints
- Remove draft facilities v1 endpoints


nfasulo:
@rtravitz oof, I didn't uncomment Facility and FacilityFeature after I reverted /nearby to using them.

PR #3181, opened 2019-07-26
Create example mobile schema

rtravitz:
## Description of change

This is just a draft pull request for us to discuss the diffed changes identifying mobile facilities in the Open API schema. It is not intended to be merged.

### DIFF
--- a/modules/va_facilities/README.v1.yml
+++ b/modules/va_facilities/README.v1.yml
@@ -39,7 +39,9 @@ info:
 
     ### Mobile Facilities
 
-    VA has a set of mobile facilities that move regularly within a region. If a facility comes back from this API with `"mobile": "true"`, the latitude and longitude should be considered inaccurate.
+    VA has a set of mobile facilities that move regularly within a region. 
+    If a facility comes back from this API with `"mobile": "true"`, the latitude/longitude and address should be considered inaccurate.
+    To get the exact current location, end consumers should contact the mobile facility directly.
 
     ## Reference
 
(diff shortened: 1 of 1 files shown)


modules/va_facilities/README.v1.yml
```diff
@@ -37,6 +37,10 @@ info:
     - The list of available services in the `services` element is only applicable to VA health and
       benefits facilities.
 
+    ### Mobile Facilities
+
+    VA has a set of mobile facilities that move regularly within a region. If a facility comes back from this API with `"mobile": "true"`, the latitude and longitude should be considered inaccurate.
```
nfasulo: I think we need to mention that the address should be considered inaccurate as well.

nfasulo: Maybe we should also say that the appropriate way to get the current address is the contact the mobile facility directly.

rtravitz:
> This makes me wonder if mobile facilities should be excluded from nearby and drive_time bands. If we can't trust the location of the facility are we doing more harm than good in telling someone that it is close to a veteran?

I agree with this. We should raise it back to the team!

davem-va:
I know from enough DST discussions, that the DST team will very much want the mobile sites (and includes them today). 

I agree that they may be less useful than the regular sites for drive time assessment but given DST and that the CCE API returns the facilities meeting drive time criteria, the application has the means to include/excluded, assuming we add the "mobile" field to each facility in the /nearby endpoint and the CCE API.

PR #3182, opened 2019-07-26
Consume SSOe assertion

bshyong:
## Description of change
Adds class to consume attributes from SSOe SAML response and map to existing VA API user and user identity models for VA API application session.

These changes will merge to feature branch `ssoe-prototype`

## Testing planned
- [x] Unit test coverage for SSOe attributes class

- [x] Attributes verified against SSOe documentation
- [x] Attributes verified against SSOe integration env
- [x] Unit tests added for SSOe attributes class
 
 
Connected to issue https://github.com/department-of-veterans-affairs/va-sso/issues/4


bshyong:
@hinzed1127 this branch is passing specs now—take a look when you get a chance. There are 5 files touched in this PR; 2 relating to specs and 3 relating to work for consuming SSOe attributes and comments

lib/saml/user.rb
```diff
@@ -85,6 +88,8 @@ def user_attributes_class
         SAML::UserAttributes::DSLogon
       when 'multifactor', 'dslogon_loa3', 'myhealthevet_loa3', LOA::IDME_LOA3, LOA::IDME_LOA1
         SAML::UserAttributes::IdMe
+      when 'ssoe'
+        SAML::UserAttributes::SSOe
```
hinzed1127: does this take care of the `TODO` above this?

bshyong: 👍

bshyong:
I think we have adequate documentation in the va-sso repo. The intent behind the TODOs are easily searchable placeholders where I expect we will need to make changes.

After the IAM integration is set up, we would go through the TODOs and remove them as we finalize and test the integraiton

PR #3191, opened 2019-08-02
separate test db and production db

lihanli:
## Description of change
separate test database and prod database so that we can run specs when ssh'd into staging
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/19381

## Testing planned
test in staging

#### Applies to all PRs

- [x] Appropriate logging


config/database.yml
```diff
@@ -55,7 +55,7 @@ development:
 # re-generated from your development database when you run "rake".
 # Do not set this db to the same as development or production.
 test:
-  url: <%= Settings.database_url %>
+  url: <%= Settings.test_database_url || Settings.database_url %>
```
jholton: Now that you've defined `Settings.test_database_url` in our settings, should this just be
```suggestion
  url: <%= Settings.test_database_url %>
```
?

PR #3196, opened 2019-08-07
Trial flipper controller

annaswims:
## Description of change
Add and endpoint that will give the current state of the feature toggle, within the context of that user, if the user is logged in.

http://localhost:3000/v0/feature_toggles?features=foo,another_toggle

````JSON
{
   "data":{
      "type":"feature_toggles",
      "features":[
         {
            "name":"foo",
            "value":false
         },
         {
            "name":"another_toggle",
            "value":true
         }
      ]
   }
}
````

## Testing done
specs

## Testing planned
Test with FE, fix Swagger docs

#### Unique to this PR
If you wish to toggle the features or add new features go to http://localhost:3000/flipper (protected with username: SUPER_SECRET_USERNAME
  password: SUPER_SECRET_PASSWORD)

Before this goes to production, I'll make a PR  to add real passwords in settings.yml
#### Applies to all PRs

- [x] Appropriate logging

### CODE REVIEW

**Comment by annaswims** [comment]
swagger docsScreen Shot 2019-08-07 at 1 15 45 AM

**Comment by annaswims** [comment]
Ah, Interesting!  We use postgresql, so it's not a [reserved word](https://www.postgresql.org/docs/10/sql-keywords-appendix.html).  It

File: app/models/user.rb [hunk 1]
```diff
@@ -273,6 +273,10 @@ def power_of_attorney
     EVSS::CommonService.get_current_info[:poa]
   end
 
+  def flipper_id
```
jholton: I don't see this used anywhere.  Is this used internally by Flipper?

annaswims: yeah, I'll ad a comment. It can be used to enable a feature for individual users, 
Screen Shot 2019-08-07 at 11 43 25 AM

but it's also used in the algorithm for enabling a feature for x% of users

PR #3198, opened 2019-08-07
Scaffolding out ssoe metadata endpoint

hinzed1127:
Mostly just duplication and renaming of existing code to `v2` vars and pointing to a new `metadata _v2` endpoint

### DIFF
--- a/app/controllers/saml_controller.rb
+++ b/app/controllers/saml_controller.rb
@@ -1,10 +1,15 @@
 # frozen_string_literal: true
 
 class SamlController < ApplicationController
-  skip_before_action :authenticate, only: [:metadata]
+  skip_before_action :authenticate, only: [:metadata, :metadata_v2]
 
   def metadata
     meta = OneLogin::RubySaml::Metadata.new
-    render xml: meta.generate(saml_settings), content_type: 'application/xml'
+    render xml: meta.generate(SAML::SettingsService.saml_settings), content_type: 'application/xml'
+  end
+
+  def metadata_v2
+    meta = OneLogin::RubySaml::Metadata.new
+    render xml: meta.generate(SAML::SettingsServiceV2.saml_settings), content_type: 'application/xml'
   end
 end
--- a/app/controllers/saml_controller.rb
+++ b/app/controllers/saml_controller.rb
@@ -5,11 +5,6 @@ class SamlController < ApplicationController
 
   def metadata
     meta = OneLogin::RubySaml::Metadata.new
-    render xml: meta.generate(SAML::SettingsService.saml_settings), content_type: 'application/xml'
-  end
-
-  def metadata_v2
-    meta = OneLogin::RubySaml::Metadata.new
-    render xml: meta.generate(SAML::SettingsServiceV2.saml_settings), content_type: 'application/xml'
+    render xml: meta.generate(saml_settings), content_type: 'application/xml'
   end
 end
--- a/app/controllers/saml_controller.rb
+++ b/app/controllers/saml_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class SamlController < ApplicationController
-  skip_before_action :authenticate, only: [:metadata, :metadata_v2]
+  skip_before_action :authenticate, only: [:metadata]
 
   def metadata
     meta = OneLogin::RubySaml::Metadata.new
--- a/app/controllers/v0/sessions_controller.rb
+++ b/app/controllers/v0/sessions_controller.rb
@@ -66,6 +66,24 @@ module V0
       stats(:total)
     end
 
+    def saml_callback_v2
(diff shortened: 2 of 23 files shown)


app/controllers/v0/sessions_controller.rb
```diff
@@ -66,6 +66,24 @@ def saml_callback
       stats(:total)
     end
 
+    def saml_callback_v2
```
hinzed1127: initially just added this but given the amount of logic related to saml in here, we probably just wanna create a duplicate sessions controller too?

patrickvinograd: Yes, let's actually create a sessions_controller under `app/controllers/v1/sessions_controller`

app/controllers/saml_controller.rb
```diff
@@ -1,10 +1,15 @@
...
-  skip_before_action :authenticate, only: [:metadata]
+  skip_before_action :authenticate, only: [:metadata, :metadata_v2]
 
   def metadata
     meta = OneLogin::RubySaml::Metadata.new
-    render xml: meta.generate(saml_settings), content_type: 'application/xml'
+    render xml: meta.generate(SAML::SettingsService.saml_settings), content_type: 'application/xml'
+  end
+
+  def metadata_v2
```
patrickvinograd: let's fold this method into the new sessions_controller class, I don't see a strong reason for a separate controller just for the metadata, and that will clear. the way to eventually remove some of. the saml_settings stuff from application_controller where it doesn't really belong.

lib/saml/settings_service_v2.rb
```diff
@@ -0,0 +1,115 @@
...
+
+require 'memoist'
+require 'sentry_logging'
+require 'saml/health_status_v2'
+
+module SAML
+  # This class is responsible for putting together a complete ruby-saml
+  # SETTINGS object, meaning, our static SP settings + the IDP settings
+  # which must be fetched once and only once via IDP metadata.
+  module SettingsServiceV2
```
patrickvinograd: Can you rename this class to "SsoeSettingsService" so the use case is clear.

config/settings.yml
```diff
@@ -13,6 +13,15 @@ saml:
   callback_url: http://localhost:3000/auth/saml/callback
   metadata_url: https://api.idmelabs.com/saml/metadata/provider
 
+# Settings for new SAML authentication with SSOe
+saml_v2:
```
patrickvinograd: Likewise change this to 'saml_ssoe'

config/routes.rb
```diff
@@ -5,6 +5,7 @@
   match '/services/*path', to: 'application#cors_preflight', via: [:options]
 
   get '/saml/metadata', to: 'saml#metadata'
+  get '/saml/metadata_v2', to: 'saml#metadata_v2'
```
patrickvinograd: For routes, I was thinking clean up the paths somewhat to  align with everything under a v1/sessions_controller:

```
/v1/sessions/metadata
/v1/sessions/callback
/v1/sessions/logout
/v1/sessions/:type:/new
```

patrickvinograd:
On the right track, just making some naming changes and suggestions for even more code isolation. I think a separate controller will make it easier to see what code we can eliminate in the future based on getting a more consistent SAML payload from SSOe. Once you incorporate feedback, I think we'll want to create PR to merge your changes into the ssoe_prototype branch instead of master.

lib/saml/health_status.rb
```diff
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 
-require 'saml/settings_service'
```
bshyong: I think this line is still necessary because we are referencing `SettingsService` below. Without the require we would need `SAML::SettingsService`

lib/saml/ssoe_health_status.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+require 'sentry_logging'
+
```
bshyong: Need `require 'saml/ssoe_health_status'` here

hinzed1127: isn't that this file?

lib/saml/health_status.rb
hinzed1127: right now `SAML::SettingsService` requires `health_status`, so it's implicitly present. Including the deleted line was causing issues via a circular dependency

hinzed1127: see [this commit](https://github.com/department-of-veterans-affairs/vets-api/pull/3198/commits/79c86d10214bf795cbbc65429446c13195c5a907). for some reason it wasn't causing issues with the existing files, but adding in the new SSOe classes was introducing it somehow

lib/saml/ssoe_health_status.rb
bshyong: ah I meant ssoe_settings_service, but if it's the same issue with circular dependencies as `health_status` then it should be good to leave it

PR #3201, opened 2019-08-09
Fixing production error being covered up by mocks

bastosmichael:
## Description of change
Seeing if this change fixes an error ocuring in production that all of our mocks and even staging masks from ticket https://github.com/department-of-veterans-affairs/vets-contrib/issues/2859

- Makes sure that the VsoSearch doesn't return a 500 in production

#### Applies to all PRs

- [x] Appropriate logging


dhall:
Can you add a test for this case?

PR #3208, opened 2019-08-13
Log 526 validation failures and validation types to statsd

ATeal:
## Description of change
In order to gain insight into how validations are working we're logging some stats in statsd. This should allow us to set up an alert if validation fails get above a threshold and the key tag should allow us to aggregate and see what validations are failing the most.

resolves https://github.com/department-of-veterans-affairs/vets-contrib/issues/2762

#### Applies to all PRs

- [x] Appropriate logging


modules/claims_api/app/controllers/claims_api/base_form_controller.rb
```diff
@@ -76,5 +80,14 @@ def format_526_errors(errors)
         { status: 422, detail: "#{error['key']} #{error['detail']}", source: error['key'] }
       end
     end
+
+    def track_526_validation_errors(errors)
```
bastosmichael: can you include a spec test for this? I can see someone trying to change this in the future.

modules/claims_api/spec/requests/v1/disability_compensation_request_spec.rb
```diff
@@ -117,6 +117,15 @@
         end
       end
 
+      it 'increment counters for statsd' do
+        with_okta_user(scopes) do |auth_header|
+          VCR.use_cassette('evss/disability_compensation_form/form_526_invalid_validation') do
+            expect(StatsD).to receive(:increment).exactly(6).times
```
annaswims: This test seems fragile and would fail if someone added another metric and only 3 of the `StatsD.increment` calls are part of the newly added metrics.  Is there anything we can do to make the test more specific?  (it's possible it's too big of a hassle)

modules/claims_api/app/controllers/claims_api/base_form_controller.rb
```diff
@@ -76,5 +80,14 @@ def format_526_errors(errors)
...
+      StatsD.increment STATSD_VALIDATION_FAIL_KEY
+
+      errors.each do |error|
+        key = error['key'].gsub(/\[(.*?)\]/, '')
```
annaswims: Are we confident that brackets are the only thing that aren't in this whitelist? https://docs.datadoghq.com/tagging/#defining-tags

ATeal: It's actually less about the whitelist and more about making the tags less unique. 

The errors for array type objects include the index where the error occurred, this is to get rid of that index.

modules/claims_api/spec/requests/v1/disability_compensation_request_spec.rb
ATeal: I'll change this to `at_least(:once)` seems way better thanks

ATeal: resolved here: https://github.com/department-of-veterans-affairs/vets-api/pull/3208/commits/914ebffce8c62366c3bc6d6cc8b2571761834bdc

PR #3210, opened 2019-08-13
SSOe prototype initial work

hinzed1127:
Feature branch combining 2 things:
- Exposing a new saml metadata endpoint via `v1/sessions/metadata`
- @bshyong 's work setting expanding `lib/saml` to consume an assertion from SSOe


app/controllers/v1/sessions_controller.rb
```diff
@@ -0,0 +1,192 @@
+# frozen_string_literal: true
```
annaswims: I'm not clear why this big copy-paste is happening.

patrickvinograd: One of the goals of this approach is that we offload some of the complexity of all the credential/LOA variations onto SSOe to some extent, i.e. we should get one format of SAML assertion and it should already include the information from MVI (since SSOe is tightly integrated with MVI). I think there is some simplification we can do both here and in the lib/saml code.

So if  this goes to production, we'd be able to drop the v0 controller and probably some of the user_identity subclasses, and having the option to do that by just deleting classes rather than intermixing a bunch of conditional logic seemed preferable to me.  It's possible we can extract some truly common stuff out into a concern though.

annaswims: Ooh, I like getting MVI data from SSOe in theory, but does that mean that we'll be more heavily reliant on MVI in order to log in? We did some work a few months ago to make sure people could still log in while MVI was down, because it has problems so often.
Screen Shot 2019-08-14 at 10 57 04 AM

patrickvinograd: That's a legitimate concern, and might be a point against  this approach relative to our other options. I know SSOe has a mode where e.g. if somebody comes in with a LOA1 credential, they also don't do an MVI lookup and just pass through the credential provider traits and no MVI data. I don't know if they have any degraded mode just for MVI being down. I will ask though.

bshyong: @patrickvinograd I think they do have a degraded mode which they will indicate with the attribute `va_eauth_csponly` TRUE, meaning there was no MVI data available

patrickvinograd:
@omgitsbillryan The problem with the existing settings_service and health_status is that they are implemented as static/singleton classes. So couldn't just instantiate new version with alternate settings. I agree there's probably a better pattern to use though.

app/controllers/v1/sessions_controller.rb
patrickvinograd: Confirmed that if MVI is unreachable, SSOe sets the `csponly` attribute to true and we can effect the same kind of conditional logic.

PR #3213, opened 2019-08-15
Art.validate page size

ATeal:
## Description of change
There have been issues with large page sizes and downstream systems. This is the beginning of solving that issue by validating page size from pdfinfo

resolves https://github.com/department-of-veterans-affairs/vets-contrib/issues/2846

## Testing done
- rspec

#### Applies to all PRs

- [x] Appropriate logging


modules/vba_documents/app/swagger/vba_documents/document_upload/status_code_description.md
```diff
@@ -7,5 +7,6 @@ Unambiguous status code. Only present if status = "error"
 * `DOC105` - Invalid or unknown id
 * `DOC106` - File size limit exceeded. Each document may be a maximum of 100MB.
 * `DOC107` - Empty payload.
+* `DOC108` - Maxium dimensions exceeded. Height and width must be less than 21 in x 21 in.
```
bastosmichael: Was this ever decided on as a breaking change or not? Before we merge we need to confirm with Andrew on whether or not VetPro and Vetrespec are okay with this.

lib/pdf_info.rb
```diff
@@ -42,8 +42,30 @@ def pages
...
+      in_pts = page_size
+      {
+        height: pts_to_inches(in_pts[:height]),
+        width: pts_to_inches(in_pts[:width])
+      }
+    end
+
     private
 
+    def pts_to_inches(dimension)
```
bastosmichael: Can you describe this method a little better?

modules/vba_documents/lib/vba_documents/upload_error.rb
```diff
@@ -13,6 +13,7 @@ class UploadError < StandardError
     DOC105 = 'Invalid or unknown id'
     DOC106 = 'Maximum document size exceeded. Limit is 100MB per document'
     DOC107 = 'Empty payload'
+    DOC108 = 'Maximum page size exceeded'
```
bastosmichael: We should probably do like DOC106 and tell them what the size is in the error response.

modules/vba_documents/app/swagger/vba_documents/document_upload/status_code_description.md
ATeal: Yes in the ticket they mentioned the communication

ATeal: it's not a breaking change

modules/vba_documents/lib/vba_documents/upload_error.rb
ATeal: fair enough I'll add it there.

PR #3215, opened 2019-08-16
Vet360 telephone text permissions

annaswims:
## Description of change
In order for the Bylight team to complete their notifications (text message) project they will need to update the value of text_message_perm_ind in Vets 360.

in addition we are updating the relevant swagger docs to include previously implemented attributes
## Testing done
updated specs

## Testing planned
verify on staging

#### Applies to all PRs

- [x] Appropriate logging


annaswims:
I did some testing in their swagger docs.  I started testing at https://int.vet360.va.gov:7005/contact-information-hub/swagger-ui.html#/TelephoneBio-v1 but that environment seemed to have errors related to an MVI connection so I switched to https://qa.vet360.va.gov:7005/contact-information-hub/swagger-ui.html#/TelephoneBio-v1.  qa.vet360.va.gov is the server that our staging environment connects to.   

We'll need to ensure a value is sent for `textMessagePermInd` each put/post or the value will be falsy on vets360.

annaswims:
I did some more testing in vets360 qa swagger environment and it looks like sending a nil textMessagePermInd value has the same behavior as not setting it at all (it's false-y)

elwoodva:
@annaswims Does this need to be updated with master branch?

annaswims:
@elwoodva  there's no harm in updating it now, but I generally don't bother until it's gone through review.  It'll have to be updated again if anybody else merges.

PR #3218, opened 2019-08-16
change services list from GIS to access to care

whuang85:
## Description of change
https://github.com/department-of-veterans-affairs/vets-contrib/issues/2825

## Testing done
Ran ruby specs,
Compared a limited set of data from /facilities/all

#### Applies to all PRs

- [x] Appropriate logging


dvogel:
The linked ticket says:

> Most of the PrimaryCare and MentalHealthCare data can also come from the Access to Care Wait Time Data. At the writing of this ticket, the source of the rest of the primarycare, mentalhealthcare, and dentalservices data is yet to be determined.

If this merges now, does this mean that the data for which the source is TBD will be missing from the API? Should this be implemented as an option, allowing us to switch between the older method and the newer method? That approach feels safer to me. Is there a reason it's not viable or unwise?

nfasulo:
@dvogel as the move to the new datasources is rather large, we have a separate branch `team-z/facilities-datasource-switchover` that we are merging all of our changes into. Once we are ready to merge that branch into `master` we are going to do a final comparison of the old and new responses from `facilities/all` to verify that the data conforms to expectations.

The language in the ticket is a little out of date, here is a brief update:

- It has been determined that we are going to own DentalServices for the time being and the work done to pull that data is captured in another ticket.

- The older data source for PrimaryCare and MentalHealthCare was hand-maintained and quit being updated at some point. Here is what Kat found when comparing the old and new sources: 
  > Primary mismatch count: 30 (97.5% accuracy)
  > Mental mismatch count: 123 (89.7% accuracy)

More context can be found here: https://lighthouseva.slack.com/archives/CEXC8GV2S/p1564506301009700

spec/lib/facilities/vha_facility_spec.rb
```diff
@@ -1,9 +1,14 @@
...
 
 require 'rails_helper'
+require 'facilities/bulk_json_client'
+
 module Facilities
   RSpec.describe VHAFacility do
-    before(:each) { BaseFacility.validate_on_load = false }
+    before(:each) do
+      BaseFacility.validate_on_load = false
+      # Facilities::AccessDataDownload.new.perform
```
edmkitty: Was this commented line meant to be taken out?

PR #3220, opened 2019-08-19
Add Checks for LOA header for V0 of claims api

charleystran:
## Description of change
This pull requests adds a requirement on an LOA header for v0 of the claims api. 

## Testing done
rspec tests

- [x] LOA is required, validated, and errors are properly responded to

#### Applies to all PRs

- [x] Appropriate logging


modules/claims_api/app/controllers/claims_api/application_controller.rb
```diff
@@ -77,5 +84,15 @@ def poa_request?
       headers_to_check = ['HTTP_X_VA_SSN', 'HTTP_X_VA_Consumer-Username', 'HTTP_X_VA_Birth_Date']
       (request.headers.to_h.keys & headers_to_check).length.positive?
     end
+
+    def verify_loa
+      loa_value = header(key = 'X-VA-LOA') ? header(key) : raise_missing_header(key)
```
ATeal: wouldn't the key be nil for the true/false paths? or is key coming from somewhere else?

ATeal: also couldn't you just call `header_loa` here? does that exception it throws not cause the 401 below?

charleystran: I refactored to use the header_loa method

modules/claims_api/app/models/claims_api/veteran.rb
```diff
@@ -68,6 +68,10 @@ def authn_context
       'authn'
     end
 
+    def set_edipi(edipi_header)
```
bastosmichael: This isn't being called anywhere, why are we adding this method back?

charleystran: could be a merge issue. I will check it out

charleystran: yea, it was added back in the merge. I removed it as well as another redundant method. I also added back the loa assignment that got lost in the shuffle as well.

modules/claims_api/app/controllers/claims_api/application_controller.rb
```diff
@@ -43,7 +62,11 @@ def veteran_from_headers(with_gender: false)
         va_profile: ClaimsApi::Veteran.build_profile(header('X-VA-Birth-Date')),
         last_signed_in: Time.now.utc
       )
-      vet.loa = @current_user.loa if @current_user
+      vet.loa = if @current_user
+                  @current_user.loa
+                else
+                  vet.loa = { current: header('X-VA-LOA'), highest: header('X-VA-LOA') }
```
bastosmichael: we need to ensure that if we fall back to X-VA-LOA and it's nil that we fall back to 0 so that it fails due to permissions instead of 500 because it's a nil value.

charleystran: Thats actually taken care of by an order of operations. Before target veteran is set, we validate that the header is present and the level is 3. I dont love it because some one could call this method outside of the current work flow, but then all the required headers could fail.

bastosmichael: :+1:

PR #3222, opened 2019-08-19
added is_textable and is_text_permitted to vets360 telephone model

annaswims:
as well as updated swagger docs to include those attributes along with several other telephone attributes that were missing.

(@thilton-oddball doesn't have write access to vets-api so we're collaborating this way)
#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging


app/swagger/schemas/vet360/telephone.rb
```diff
@@ -8,67 +8,137 @@ class Telephone
...
                    pattern: ::Vet360::Models::Telephone::VALID_AREA_CODE_REGEX.inspect
+          property :country_code,
+                   type: :string,
+                   enum: ['1'],
+                   example: '1'
+          # property :created_at,
+          #          type: :string,
+          #          format: 'date-time',
+          #          example: '2018-04-21T20:09:50Z'
+          property :effective_end_date,
```
annaswims: I think that we want to take this param out of the documentation because it's always removed or overridden by our code.

app/swagger/schemas/vet360/telephone.rb
```diff
@@ -8,67 +8,137 @@ class Telephone
...
+                   type: %i[string null],
+                   format: 'date-time',
+                   example: '2018-04-21T20:09:50Z'
+          property :effective_start_date,
```
annaswims: I know that it's not a widely used attribute, but I think a few of these could benefit from a `description` attribute.  I'm thinking you could grab exactly what the vets360 documentation that I emailed you says.

annaswims:
looks good to me, I'm going to merge it and run tests with my branch.

PR #3223, opened 2019-08-19
Quick fix for calling edipi

bastosmichael:
## Description of change

This is what actually ended up working on staging to call EDIPI properly ticket https://github.com/department-of-veterans-affairs/vets-contrib/issues/2859

- Properly call EDIPI for Veteran model

### DIFF
--- a/modules/claims_api/app/models/claims_api/veteran.rb
+++ b/modules/claims_api/app/models/claims_api/veteran.rb
@@ -25,7 +25,7 @@ module ClaimsApi
     delegate :birls_id, to: :mvi, allow_nil: true
     delegate :participant_id, to: :mvi, allow_nil: true
 
-    alias :dslogon_edipi :edipi
+    alias dslogon_edipi edipi
 
     def birth_date
       va_profile[:birth_date]
--- a/modules/claims_api/app/models/claims_api/veteran.rb
+++ b/modules/claims_api/app/models/claims_api/veteran.rb
@@ -25,7 +25,7 @@ module ClaimsApi
     delegate :birls_id, to: :mvi, allow_nil: true
     delegate :participant_id, to: :mvi, allow_nil: true
 
-    alias :dslogon_edipi
+    alias :dslogon_edipi :edipi
 
     def birth_date
       va_profile[:birth_date]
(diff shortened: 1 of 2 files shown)


modules/claims_api/app/models/claims_api/veteran.rb
```diff
@@ -25,6 +25,8 @@ class Veteran
     delegate :birls_id, to: :mvi, allow_nil: true
     delegate :participant_id, to: :mvi, allow_nil: true
 
+    alias :dslogon_edipi
```
charleystran: I think you need to alias this to edipi or else it will always be niil

charleystran: We should also add a test to make sure they are equal

bastosmichael: The alias is still required because the MVI model calls it when it's doing the Mvi.for_user(self)

modules/claims_api/app/models/claims_api/veteran.rb
```diff
@@ -65,8 +67,8 @@ def authn_context
       'authn'
     end
 
-    def set_edipi(edipi_header)
-      self.edipi = edipi_header.presence || mvi.profile&.edipi
+    def edipi(edipi_header = nil)
```
charleystran: We shouldn't have edipi as an attribute and a method. We should either use the old set_edipi method to set the attribute or remove it as a model attribute.

bastosmichael: I think the reason it was failing before is that it wasn't ever actually setting the self.edipi which is why it kept coming back nil even with your changes. Referencing the @edipi ensured that the attribute value was filled in and making it a method ensured reference. Making it a both an attribute and a method means it'll always catch it no matter where it's referenced and from all of my testing works in ensuring that the EDIPI is never nil.

charleystran: Can you add a spec that shows that? I am not following how the method doesnt cancel out the attribute.

bastosmichael: so I've tried it every other way and we seem to need both the attributes_attr and the similar named method for assignment. Hear me out here on the why. When you called set_edipi from the self.from_headers method, loa was never set yet so your fall back for mvi.profiles&.edipi would never actually get called because mvi would come back with nothing because loa 3 wasn't set yet. So by having a fall back method that checks the edipi_header.presence and if that still doesn't exist fall back to mvi.profile&.edipi ensures that even it doesn't load via the initial headers, if it's called at any point later it will reference the edipi in the mvi after loa is set. You need both the attributes set and the method as a fall back because you don't know at which stage the edipi method/variable will be called.

PR #3226, opened 2019-08-20
config key/certificate onboarding instruction updates

LindseySaari:
To set up configurations when using localhost, the team repo certificate and key will need to be copied and pasted from the team repo to the files listed in these instructions. A blank key and certificate no longer works. This is an important step for onboarding to properly set up configurations. 

## Description of change
This change provides instructions for on boarding personnel in vets-api to set up the proper configurations. Without the correct key and certificates, configuration breaks when testing in a local environment

## Testing done
After updating the key and certificate configurations, SSO was tested locally

#### Unique to this PR
- [x] Replacing/Adding instructions to native.md for onboarding

#### Applies to all PRs

- [ ] Appropriate logging


docs/setup/native.md
```diff
@@ -69,6 +69,14 @@ the `vets-api` directory, run `./bin/setup-osx && source ~/.bash_profile && cd
      commands will provide basic functionality, such as for running unit tests:
    - `touch ./config/certs/vetsgov-localhost.crt`
    - `touch ./config/certs/vetsgov-localhost.key`
+   
+   - *NOTE:* using `touch` to create blank cert and key files no longer works.
```
jholton: How about instead of adding another note, we just delete the first note?

.rubocop.yml
```diff
@@ -119,7 +119,6 @@ Style/RescueStandardError:
     - "lib/mvi/messages/find_profile_message_helpers.rb"
     - "lib/mvi/responses/profile_parser.rb"
     - "lib/saml/settings_service.rb"
-    - "lib/saml/ssoe_settings_service.rb"
```
johnpaulashenfelter: Does this file just not exist anymore?

LindseySaari: Looks like it exists

PR #3228, opened 2019-08-21
add feature descriptions

annaswims:
## Description of change
Add details into UI alongside feature flag names in  [Flipper ui](http://localhost:3000/flipper/features)

Also, enable features as they're initialized in development. 

## Testing done
Screen Shot 2019-08-21 at 6 15 51 PM

#### Unique to this PR
This is the only erb within vets-api.

#### Applies to all PRs

- [ ] Appropriate logging

### DIFF
--- a/config/features.yml
+++ b/config/features.yml
@@ -1,0 +1,6 @@
+---
+features:
+  facilityLocatorShowCommunityCares:
+    description: >
+      This is a feature flag description.
+    actor: User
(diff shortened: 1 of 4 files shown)


annaswims:
@thilton-oddball It's a bit of rails magic. 
 We're only overriding a single body of a page from the flipper_ui gem (it's a [rails engine](https://guides.rubyonrails.org/engines.html)), and we're still using the[ layout from the gem]( https://github.com/jnunemaker/flipper/blob/master/lib/flipper/ui/views/layout.erb) and it the [css that is part of the gem](https://github.com/jnunemaker/flipper/blob/master/lib/flipper/ui/public/css/application.css)

PR #3229, opened 2019-08-22
Art.refactor claims

ATeal:
#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging

### DIFF
--- a/modules/claims_api/app/controllers/claims_api/concerns/request_logging.rb
+++ b/modules/claims_api/app/controllers/claims_api/concerns/request_logging.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+module ClaimsApi
+  module RequestLogging
+    extend ActiveSupport::Concern
+    # rubocop:disable Metrics/BlockLength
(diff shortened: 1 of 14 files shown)


modules/claims_api/app/controllers/claims_api/concerns/request_logging.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      def log_response(additional_fields = {})
+        logged_info = {
+          'consumer' => consumer,
+          'va_user' => requesting_va_user
+        }.merge(additional_fields)
+        Rails.logger.info('Claims App Response', logged_info)
+      end
+
+      def consumer
```
charleystran: Do you think this is the right class to have the header methods? If we need to use the LOA header, I don't think the Request logging concern would be the right place. Maybe we should have module specifically for header logic?

ATeal: mmm good idea, I'll look into that

PR #3230, opened 2019-08-22
Add dental services source for facilities

nfasulo:
Adds a CSV file that includes the 225 VHA facilities that currently have DentalServices listed in ArcGIS so that we can add this service into these facilities once we move to GIS from ArcGIS. This CSV is filled with `unique_id`, `facility_type`, and a flag indicating that `DentalService` was included in the `services.health` array from my local copy of the table `base_facilities`.

The `psql` command to create this CSV:
`\copy (SELECT unique_id, facility_type, 'y' As DentalServices FROM base_facilities WHERE facility_type = 'va_health_facility' AND services->>'health' ilike '%Dental%') TO './lib/facilities/dental_services.csv' WITH CSV`

## Testing
- verified that the number of records matched @edmkitty's comparison from 7/31


edmkitty:
If every row in this list is a facility that has dental services, then I'm not sure there is much point in having a column whose value will be "y" for every row. I think that column should be removed.

rtravitz:
Agree with @edmkitty's thought about the Dental Services column. Other than that, it looks good to me.

nfasulo:
I included the DentalServices column to be ridiculously explicit as to what the data means, but I can take it out. You are right that it doesn't actually make the work any easier.

PR #3231, opened 2019-08-22
Make SAML initializer safe to missing SSOe cert/key files

patrickvinograd:
## Description of change
Makes SAML initializer safe to missing key/cert files, for SSOe variants of those files. In response to issue where we were not injecting those files on worker instances. We will also make a change to inject those files, but this makes vets-api server/worker more robust.

## Testing done
Verified locally with dangling path references in settings.

#### Applies to all PRs

- [ ] Appropriate logging


patrickvinograd:
Note we will also push a PR to devops to inject keys on the dev/staging worker instances, but since we have not provisioned production keys from VA PKI team, I want to deploy this first to make sure all environments are robust to missing keys.

PR #3244, opened 2019-08-28
Provider verification for contract testing with Pact

U-DON:
## Description of change
* Verifies the pact for `POST /health_care_application` created by `vets-website` in department-of-veterans-affairs/vets-website#10970.
* Integrates with a Pact broker instance to publish results.

Refer to the [full overview](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/teams/vsp/teams/tools/2019-11-13-consumer-driven-contract-testing-with-pact.md) for instructions to run the broker and generate the pact from the FE.

## Testing done
Using the Docker setup, I ran: 

```docker-compose run vets-api rake pact:verify```

Note that I had to make the following additions/changes to my `config/settings.local.yml`. This is just a copy of some of the environment variables already set in `docker-compose.yml`, and I'm not sure why they're necessary. I'm also not sure how this will all have to change when we try to incorporate the Pact flow into CI.

```
redis:
  host: redis
  port: 6379

database_url: postgis://postgres:password@postgres:5432/vets_api_development?pool=4
```

## Testing planned
If all looks good, the next step would be to incorporate the Pact flow into CI.

#### Unique to this PR
- [ ] There should be a Rake task that verifies the pacts generated by vets-website and published to the broker.
- [ ] The pact verification results should be published to a broker.

#### Applies to all PRs

- [ ] Appropriate logging

### CODE REVIEW

File: spec/service_consumers/pact_helper.rb [hunk 1]
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+require 'pact/provider/rspec'
+
+Pact.service_provider "VA.gov API" do
+  honours_pact_with 'VA.gov' do
```
U-DON: Might change from considering VA.gov as a single consumer to each app being individual consumers.

spec/service_consumers/pact_helper.rb
```diff
@@ -0,0 +1,27 @@
...
+Pact.service_provider 'VA.gov API' do
+  # This example points to a local file, however, on a real project with a continuous
+  # integration box, you would use a [Pact Broker](https://github.com/pact-foundation/pact_broker) or publish your pacts as artifacts,
+  # and point the pact_uri to the pact published by the last successful build.
+  # honours_pact_with 'VA.gov' do
+  #   pact_uri 'tmp/va.gov-va.gov_api.json'
+  # end
+
+  app_version provider_version
+  publish_verification_results true
```
U-DON: The broker needs a version for both the consumer and provider to properly tabulate the results. Provider verson can be anything we want, but commit hashes seemed like a solid way of setting the version for the broker. I'm still not sure of the exact deploy strategy we want to enforce, though, in terms of when we can tell a consumer that it is good to deploy based on the version of the provider that passed verification.

The flag to turn on the publishing of verification results can be tied to deploy environment if we wanted.

spec/service_consumers/pact_helper.rb
```diff
@@ -0,0 +1,27 @@
...
+# The easiest way to do this is to include the build number (or a SHA) in your version.
+provider_version = ENV['GIT_COMMIT'] || `git rev-parse --verify HEAD`
+# publish_flag = ENV['PUBLISH_VERIFICATION_RESULTS'] == "true"
+
+Pact.service_provider 'VA.gov API' do
+  # This example points to a local file, however, on a real project with a continuous
+  # integration box, you would use a [Pact Broker](https://github.com/pact-foundation/pact_broker) or publish your pacts as artifacts,
+  # and point the pact_uri to the pact published by the last successful build.
+  # honours_pact_with 'VA.gov' do
+  #   pact_uri 'tmp/va.gov-va.gov_api.json'
```
U-DON: If having trouble doing verification through a broker, or if we decide we don't really need a broker (though it seems to be recommended), we can also share pacts between consumers and the API through a central location. For my initial pass at using Pact, I simply copied the pact generated from `vets-website` to the `tmp` folder here in `vets-api`.

PR #3250, opened 2019-08-29
Adding Attorneys and Claims Agents to VSR Model

bastosmichael:
## Description of change
In reference to ticket https://github.com/department-of-veterans-affairs/vets-contrib/issues/2924 where we needed to add Attorneys and Claim Agents to our Veteran Service Rep table.

## Testing done
- Ran locally and was able to pull in all the data we wanted.
- Updated Rspec test and VCR Cassette

## Testing planned
- Will run this in staging and the production

- Attorneys and Claim Assistants can have access to their own Power of Attorneys in the Veteran Service Rep models.

#### Applies to all PRs

- [x] Appropriate logging


modules/veteran/app/workers/veteran/base_reloader.rb
```diff
@@ -0,0 +1,38 @@
...
+                                                                   first_name: hash['First Name'],
+                                                                   last_name: hash['Last Name'])
+      rep.poa_codes ||= []
+      rep.user_types ||= []
+      rep.poa_codes << hash['POA Code'].gsub!(/\W/, '')
+      rep.phone = hash['Phone']
+      rep
+    end
+
+    def fetch_data(action)
```
dhall: you may have already spent time cleaning this up but wondering if it could be tidied up a bit.. is excel -> html -> csv -> hash the best approach or the first that worked?

modules/veteran/app/workers/veteran/base_reloader.rb
```diff
@@ -0,0 +1,38 @@
...
+      page = Faraday.new(url: BASE_URL).post(action, id: 'frmExcelList', name: 'frmExcelList').body
+      doc = Nokogiri::HTML(page)
+      content = CSV.generate(headers: true) do |csv|
+        doc.xpath('//table/tr').each do |row|
+          tarray = []
+          row.xpath('td').each do |cell|
```
dhall: can this just be done with .map

modules/veteran/app/workers/veteran/vso_reloader.rb
```diff
@@ -3,54 +3,58 @@
...
 module Veteran
-  class VsoReloader
-    include Sidekiq::Worker
-    BASE_URL = 'https://www.va.gov/ogc/apps/accreditation/'
-
+  class VsoReloader < BaseReloader
     def perform
       array_of_organizations = reload_representatives
-      Veteran::Service::Organization.import(array_of_organizations, on_duplicate_key_ignore: true)
+      Veteran::Service::Representative.where.not(representative_id: array_of_organizations).destroy_all
```
dhall: this is a rare case for me where I think a comment would be helpful. is this a de-duping operation? what is intended to be destroyed?

modules/veteran/app/workers/veteran/vso_reloader.rb
```diff
@@ -3,54 +3,58 @@
...
     end
 
     private
 
     def reload_representatives
-      data = fetch_data('orgsexcellist.asp')
-      data.each do |hash|
-        find_or_create_representative(hash)
+      vso_reps = []
+      vso_orgs = fetch_data('orgsexcellist.asp').map do |hash|
```
dhall: `hash` is a weak variable name.. can you use something that better reveals intention?

modules/veteran/app/workers/veteran/vso_reloader.rb
charleystran: I agree a comment would help context. I believe the reason is we are not informed of removals to the list so we clean the slate and start over each time

modules/veteran/spec/workers/veteran/vso_reloader_spec.rb
```diff
@@ -15,8 +15,8 @@
     it 'should reload data from pulldown' do
       VCR.use_cassette('veteran/ogc_poa_data') do
         Veteran::VsoReloader.new.perform
-        expect(Veteran::Service::Representative.count).to eq 152
-        expect(Veteran::Service::Organization.count).to eq 3
+        expect(Veteran::Service::Representative.count).to eq 23_361
+        expect(Veteran::Service::Organization.count).to eq 91
```
charleystran: Does are test cassette really have 23k records? I think that it to many and will make the test too slow. Previously with 9k records it was a minute just for the single test. Can we modify the test cassette to not import the whole data set?

modules/veteran/app/workers/veteran/base_reloader.rb
bastosmichael: The data actually comes in as XML so it never really enters the excell, html or csv formats, hash in my opinion is the cleanest implementation to get it into before persisting to the DB.

modules/veteran/app/workers/veteran/base_reloader.rb
bastosmichael: Yes it can

modules/veteran/app/workers/veteran/vso_reloader.rb
bastosmichael: So the assumption is made early on that every time the list is updated, if anyone has been removed from the list that may mean they are no longer an authorized VSO from OGC so the decision was made to remove them from our list accordingly. If they make it back on the list in the future that's fine but since we are using OGC as the source of truth here then it makes the most sense to remove anyone not in the incoming list. I'll add a comment to address this question.

modules/veteran/app/workers/veteran/vso_reloader.rb
bastosmichael: I'll change it to something else like vso...

modules/veteran/spec/workers/veteran/vso_reloader_spec.rb
bastosmichael: I'll see if I can strip down the VCR in some way.

modules/veteran/app/workers/veteran/base_reloader.rb
```diff
@@ -0,0 +1,38 @@
...
+    include Sidekiq::Worker
+    BASE_URL = 'https://www.va.gov/ogc/apps/accreditation/'
+
+    private
+
+    def find_or_initialize(hash)
+      rep = Veteran::Service::Representative.find_or_initialize_by(representative_id: hash['Registration Num'],
+                                                                   first_name: hash['First Name'],
+                                                                   last_name: hash['Last Name'])
+      rep.poa_codes ||= []
```
charleystran: Instead of an OR Equals, we could just set this as a DB default, then we dont need initial the empty arrays

bastosmichael: Was trying to avoid a column change like this but I agree that it's the better approach.

modules/veteran/app/workers/veteran/vso_reloader.rb
erluti: `hash` is used in a few other places in here

modules/veteran/app/workers/veteran/base_reloader.rb
ATeal: Can't help but agree here, transforming the data through even 3 different formats feels weird, Seems like you could just go XML -> hashes.

I know you say that it does enter html or csv formats, but is that not what the block that starts on 26 doing? Then you map over the contents of the CSV calling `to_h` to get your array of hashes

modules/veteran/app/workers/veteran/vso_reloader.rb
```diff
@@ -3,54 +3,58 @@
...
-      array_of_organizations = data.map do |h|
-        { poa: h['POA'], name: h['Organization Name'], phone: h['Org Phone'], state: h['Org State'] }
-      end.uniq.compact
-      array_of_organizations
+      vso_reps + attorneys + claim_agents
+    end
+
+    def find_or_create_attorneys(hash)
+      rep = find_or_initialize(hash)
+      rep.user_types << 'attorney'
```
ATeal: You should probably do a check if that `user_type` exists no? If you find the record multiple times you'll continuously add 'attorney'` to the list

modules/veteran/app/workers/veteran/vso_reloader.rb
```diff
@@ -3,54 +3,58 @@
...
+      rep.save
     end
 
-    def find_or_create_representative(hash)
-      rep = Veteran::Service::Representative.find_or_initialize_by(representative_id: hash['Registration Num'])
+    def find_or_create_claim_agents(hash)
+      rep = find_or_initialize(hash)
+      rep.user_types << 'claim_agents'
```
ATeal: same as above

modules/veteran/app/workers/veteran/vso_reloader.rb
```diff
@@ -3,54 +3,58 @@
...
+      rep = Veteran::Service::Representative.find_or_initialize_by(representative_id: hash['Registration Num'],
+                                                                   first_name: hash['Representative'].split(' ').second,
+                                                                   last_name: hash['Representative'].split(',').first)
       rep.poa_codes ||= []
       rep.poa_codes << hash['POA'].gsub!(/\W/, '')
-      rep.first_name = hash['Representative'].split(' ').second
-      rep.last_name = hash['Representative'].split(',').first
       rep.phone = hash['Org Phone']
+      rep.user_types ||= []
+      rep.user_types << 'veteran_service_officer'
```
ATeal: same issue as above, wouldn't this continuously add `veteran_service_officer` if you found them again

modules/veteran/app/workers/veteran/base_reloader.rb
bastosmichael: So no I took another look and I was wrong, It comes in as HTML and I can convert it straight to Hash for persistence but it's never in xml or excel, the CSV parsing was just the quickest way to do it at the time but I'll modify it to go straight from HTML to Hash in this case.

bastosmichael: Done

modules/veteran/app/workers/veteran/base_reloader.rb
bastosmichael: Fixed

modules/veteran/app/workers/veteran/vso_reloader.rb
bastosmichael: Comment added.

modules/veteran/app/workers/veteran/base_reloader.rb
```diff
@@ -23,16 +23,10 @@ def find_or_initialize(hash)
...
-        doc.xpath('//table/tr').each do |row|
-          tarray = []
-          row.xpath('td').each do |cell|
-            tarray << cell.text.scrub
-          end
-          csv << tarray
-        end
+      headers = doc.xpath('//table/tr').first.children.children.map {|header| header.text }
+      doc.xpath('//table/tr').map do |row|
+        Hash[headers.zip(row.children.children.map {|cell| cell.text})]
```
dhall: you dropped .scrub ... was that intentional?

modules/veteran/app/workers/veteran/vso_reloader.rb
bastosmichael: Done on all accounts.

modules/veteran/app/workers/veteran/base_reloader.rb
bastosmichael: Change made to migration

modules/veteran/app/workers/veteran/vso_reloader.rb
bastosmichael: Checking now to make sure it doesn't do double entry.

modules/veteran/app/workers/veteran/vso_reloader.rb
bastosmichael: Checking now to make sure it doesn't do double entry.

modules/veteran/app/workers/veteran/vso_reloader.rb
bastosmichael: Checking now to make sure it doesn't do double entry.

modules/veteran/app/workers/veteran/base_reloader.rb
bastosmichael: Yes, scrub was removing the \n and \r left overs that were left over in the string, actually going through the children.children removes all of that and leaves you only with the cell text itself.

charleystran: Scrub wasn't removing the newlines, it was removing invalid byte sequences. I believe it was a badly encoded apostrophe. We probably need to add it back

modules/veteran/spec/workers/veteran/vso_reloader_spec.rb
bastosmichael: So all of the attempts I made at modifying the cassette to reduce the payload down even further were met with one error after another below (even when I tried to recalculate the payload hash after modifying it etc.

```1) Veteran::VsoReloader importer should reload data from pulldown
     Failure/Error: Veteran::VsoReloader.new.perform
     
     VCR::Errors::UnhandledHTTPRequestError:
     
     
       ================================================================================
       An HTTP request has been made that VCR does not know how to handle:
         POST https://www.va.gov/ogc/apps/accreditation/caexcellist.asp
     
       VCR is currently using the following cassette:
         - /home/michael/oddball/vets-api/spec/support/vcr_cassettes/veteran/ogc_poa_data.yml
           - :record => :once
           - :match_requests_on => [:method, :uri]
     
       Under the current configuration VCR can not find a suitable HTTP interaction
       to replay and is prevented from recording new requests. There are a few ways
       you can deal with this:
     
         * If you're surprised VCR is raising this error
           and want insight about how VCR attempted to handle the request,
           you can use the debug_logger configuration option to log more details [1].
         * You can use the :new_episodes record mode to allow VCR to
           record this new request to the existing cassette [2].
         * If you want VCR to ignore this request (and others like it), you can
           set an `ignore_request` callback [3].
         * The current record mode (:once) does not allow new requests to be recorded
           to a previously recorded cassette. You can delete the cassette file and re-run
           your tests to allow the cassette to be recorded with this request [4].
         * The cassette contains 1 HTTP interaction that has not been
           played back. If your request is non-deterministic, you may need to
           change your :match_requests_on cassette option to be more lenient
           or use a custom request matcher to allow it to match [5].
     
       [1] https://www.relishapp.com/vcr/vcr/v/3-0-3/docs/configuration/debug-logging
       [2] https://www.relishapp.com/vcr/vcr/v/3-0-3/docs/record-modes/new-episodes
       [3] https://www.relishapp.com/vcr/vcr/v/3-0-3/docs/configuration/ignore-request
       [4] https://www.relishapp.com/vcr/vcr/v/3-0-3/docs/record-modes/once
       [5] https://www.relishapp.com/vcr/vcr/v/3-0-3/docs/request-matching
       ================================================================================
     # /home/michael/.rvm/gems/ruby-2.4.5/gems/vcr-3.0.3/lib/vcr/request_handler.rb:97:in `on_unhandled_request'
     # /home/michael/.rvm/gems/ruby-2.4.5/gems/vcr-3.0.3/lib/vcr/library_hooks/webmock.rb:129:in `on_unhandled_request'
     # /home/michael/.rvm/gems/ruby-2.4.5/gems/vcr-3.0.3/lib/vcr/request_handler.rb:24:in `handle'
     # /home/michael/.rvm/gems/ruby-2.4.5/gems/vcr-3.0.3/lib/vcr/library_hooks/webmock.rb:144:in `block in <module:WebMock>'
     # /home/michael/.rvm/gems/ruby-2.4.5/gems/webmock-3.5.1/lib/webmock/stub_registry.rb:28:in `block in register_global_stub'
     # /home/michael/.rvm/gems/ruby-2.4.5/gems/webmock-3.5.1/lib/webmock/request_pattern.rb:40:in `matches?'
     # /home/michael/.rvm/gems/ruby-2.4.5/gems/webmock-3.5.1/lib/webmock/stub_registry.rb:58:in `block in request_stub_for'
     # /home/michael/.rvm/gems/ruby-2.4.5/gems/webmock-3.5.1/lib/webmock/stub_registry.rb:57:in `each'
     # /home/michael/.rvm/gems/ruby-2.4.5/gems/webmock-3.5.1/lib/webmock/stub_registry.rb:57:in `detect'
     # /home/michael/.rvm/gems/ruby-2.4.5/gems/webmock-3.5.1/lib/webmock/stub_registry.rb:57:in `request_stub_for'
     # /home/michael/.rvm/gems/ruby-2.4.5/gems/webmock-3.5.1/lib/webmock/stub_registry.rb:50:in `response_for_request'
     # /home/michael/.rvm/gems/ruby-2.4.5/gems/webmock-3.5.1/lib/webmock/http_lib_adapters/net_http.rb:79:in `request'
     # /home/michael/.rvm/gems/ruby-2.4.5/gems/faraday-0.9.2/lib/faraday/adapter/net_http.rb:82:in `perform_request'
     # /home/michael/.rvm/gems/ruby-2.4.5/gems/faraday-0.9.2/lib/faraday/adapter/net_http.rb:40:in `block in call'
     # /home/michael/.rvm/gems/ruby-2.4.5/gems/faraday-0.9.2/lib/faraday/adapter/net_http.rb:87:in `with_net_http_connection'
     # /home/michael/.rvm/gems/ruby-2.4.5/gems/faraday-0.9.2/lib/faraday/adapter/net_http.rb:32:in `call'
     # /home/michael/.rvm/gems/ruby-2.4.5/gems/faraday-0.9.2/lib/faraday/request/url_encoded.rb:15:in `call'
     # /home/michael/.rvm/gems/ruby-2.4.5/gems/faraday-0.9.2/lib/faraday/rack_builder.rb:139:in `build_response'
     # /home/michael/.rvm/gems/ruby-2.4.5/gems/faraday-0.9.2/lib/faraday/connection.rb:377:in `run_request'
     # /home/michael/.rvm/gems/ruby-2.4.5/gems/faraday-0.9.2/lib/faraday/connection.rb:177:in `post'
     # ./modules/veteran/app/workers/veteran/base_reloader.rb:23:in `fetch_data'
     # ./modules/veteran/app/workers/veteran/vso_reloader.rb:45:in `reload_claim_agents'
     # ./modules/veteran/app/workers/veteran/vso_reloader.rb:16:in `reload_representatives'
     # ./modules/veteran/app/workers/veteran/vso_reloader.rb:8:in `perform'
     # ./modules/veteran/spec/workers/veteran/vso_reloader_spec.rb:17:in `block (4 levels) in <top (required)>'
     # /home/michael/.rvm/gems/ruby-2.4.5/gems/vcr-3.0.3/lib/vcr/util/variable_args_block_caller.rb:9:in `call_block'
     # /home/michael/.rvm/gems/ruby-2.4.5/gems/vcr-3.0.3/lib/vcr.rb:189:in `use_cassette'
     # ./modules/veteran/spec/workers/veteran/vso_reloader_spec.rb:16:in `block (3 levels) in <top (required)>'```

If you have another way to do it let me know but otherwise I've got no way so far of pulling off a file reduction.

File: modules/veteran/app/workers/veteran/base_reloader.rb [hunk 11]
```diff
@@ -0,0 +1,32 @@
...
+      rep.phone = hash_object['Phone']
+      rep
+    end
+
+    def fetch_data(action)
+      page = Faraday.new(url: BASE_URL).post(action, id: 'frmExcelList', name: 'frmExcelList').body
+      doc = Nokogiri::HTML(page)
+      headers = doc.xpath('//table/tr').first.children.map { |child| child.children.text.scrub }
+      doc.xpath('//table/tr').map do |line|
+        row = line.children.map { |child| child.children.text.scrub }
```
dhall: during your refactoring you were able to drop .scrub without causing a test to fail... I saw the comment that we need this, I'd recommend a failing test so that it can't be dropped in the future

bastosmichael: Will add it to another PR then just to close this one out for now since it's been open for longer than desired and that test may take longer to figure out because of the nature of the test payload, in essence the scrub was first removed because the data set I was using didn't special characters in the OGC data that can some times be there and sometimes not it just depends on when the data was pulled. Defaulting to scrubbing the data makes the most sense by default at the moment.

PR #3257, opened 2019-08-30
SAML Okta hook

hinzed1127:
Things working:
- Consumes and parses out the Okta hook request successfully

Things left to finish:
- I'm still getting this error from the MVI lookup `Performed MVI Query -- Exception: Common::Exceptions::ValidationErrors: Validation error -- { :uuid => nil, :authn_context => nil }` . I saw some [slack conversation](https://adhoc.slack.com/archives/C5SR3DCDR/p1567036903001100) around the uuid, although I'm not sure what's getting lost in translation with the authn_context, which I'm grabbing off the request below.

### DIFF
--- a/modules/openid_auth/app/controllers/openid_auth/v0/okta_controller.rb
+++ b/modules/openid_auth/app/controllers/openid_auth/v0/okta_controller.rb
@@ -1,0 +1,101 @@
+# frozen_string_literal: true
+
+require_dependency 'openid_auth/application_controller'
+
+module OpenidAuth
+  module V0
+    class OktaController < ApplicationController
+      skip_before_action :authenticate
+
+      def initialize
+        @okta_response = {
+          "commands": [
+            {
+              "type": 'com.okta.assertion.patch',
+              "value": []
+            }
+          ]
+        }
+      end
+
+      def parse_user_attributes(json_request)
+        okta_attributes = json_request.dig(:data, :assertion, :claims)
+        user_attributes = {}
+        okta_attributes.each do |k, v|
+          user_attributes[k.to_sym] = v[:attributeValues][0][:value]
+        end
+        user_attributes
+      end
+
+      def fetch_mvi_profile(user_attributes)
+        user_identity = OpenidUserIdentity.new(
+          birth_date: user_attributes[:dob],
+          dslogon_edipi: user_attributes[:dslogon_edipi],
+          email: user_attributes[:user_email],
+          first_name: user_attributes[:first_name],
+          gender: user_attributes[:gender]&.chars&.first&.upcase,
+          last_name: user_attributes[:last_name],
+          loa:
+          {
+            current: user_attributes[:level_of_assurance].to_i,
+            highest: user_attributes[:level_of_assurance].to_i
+          },
+          mhv_icn: user_attributes[:mhv_icn],
+          ssn: user_attributes[:ssn],
+          uuid: user_attributes[:idp_uuid]
+        )
+
+        mvi_response = MVI::Service.new.find_profile(user_identity)
+
+        raise mvi_response.error if mvi_response.error # TODO: add error logging
+        mvi_response[:profile]
+      end
+
+      def okta_callback
+        user_attributes = parse_user_attributes(params)
+        mvi_profile = fetch_mvi_profile(user_attributes)
+
+        create_replace_stanza('dslogon_edipi', mvi_profile[:edipi])
+        create_add_stanza('icn_with_aaid', mvi_profile[:icn_with_aaid])
+
+        render json: @okta_response
+      end
+
+      private
+
+      def create_replace_stanza(path_variable, value)
+        replace_stanza = {
+          "op": 'replace',
+          "path": "/claims/#{path_variable}/attributeValues/0/value",
+          "value": value
+        }
+
+        @okta_response[:commands][0][:value].push(replace_stanza)
+      end
+
+      def create_add_stanza(path_variable, value)
+        # TODO: add futher args to specify name_format, type
+
+        add_stanza = {
+          "op": 'add',
+          "path": "/claims/#{path_variable}",
+          "value": {
+            "attributes": {
+              "NameFormat": 'urn:oasis:names:tc:SAML:2.0:attrname-format:basic'
+            },
+            "attributeValues": [
+              {
+                "attributes": {
+                  "xsi:type": 'xs:string'
+                },
+                "value": value
+              }
+            ]
+          }
+        }
+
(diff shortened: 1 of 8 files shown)


app/controllers/application_controller.rb
```diff
@@ -55,7 +55,7 @@ def clear_saved_form(form_id)
 
   # returns a Bad Request if the incoming host header is unsafe.
   def block_unknown_hosts
-    return if controller_name == 'example'
+    return if controller_name.in?(%w[example okta])
```
hinzed1127: workaround while this is still coming via localtunnel. Will remove prior to merging and once Kong is set up

app/controllers/okta_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  def fetch_mvi_profile(json_request)
+    # TODO: pull out info from Okta request for querying MVI. Put into helper func
+    okta_attributes = json_request.dig(:data, :assertion, :claims)
+    user_attributes = Hash.new
+    okta_attributes.each do |k, v|
+      user_attributes[k] = v[:attributeValues][0][:value]
+    end
+    user_attributes["authn_context"] = json_request[:data][:assertion][:authentication][:authnContext][:authnContextClassRef]
```
hinzed1127: I don't think this is getting passed through, after closer examination. 

Validation error is coming from [here](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/mvi/service.rb#L133). We're sending the `mhv_icn`, so something is making the `.valid?` check return false

app/controllers/okta_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+    
+    binding.pry
+    user_identity = OpenidAuth::V0::MviUsersController.new.build_identity_from_attributes(user_attributes)
```
hinzed1127: this needs to get moved, but I was still having trouble getting this method to run after moving this controller inside the `OpenIdAuth` module

modules/openid_auth/config/routes.rb
```diff
@@ -7,6 +7,7 @@
     get 'validation', to: 'validation#index'
     get 'mvi-user', to: 'mvi_users#show'
     post 'mvi-user', to: 'mvi_users#search'
+    post 'okta', to: 'okta#okta_callback'
```
hinzed1127: was able to have a hook hit this route, but still running into the issues mentioned above

app/controllers/okta_controller.rb
patrickvinograd: I would  just  not try to re-use the MVIUsersController method. Copy what code you need to build a UserIdentity object and pass it to the MVI lookup method. If there's enough overlap, we can extract it out to a controller concern, that's the more proper way of sharing code vs. instantiating a controller and invoking a method on it.

modules/openid_auth/app/controllers/openid_auth/v0/okta_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+        okta_attributes = json_request.dig(:data, :assertion, :claims)
+        user_attributes = Hash.new
+        okta_attributes.each do |k, v|
+          user_attributes[k.to_sym] = v[:attributeValues][0][:value]
+        end
+        user_attributes
+      end
+
+      def fetch_mvi_profile(user_attributes)    
+        user_identity = OpenidUserIdentity.new(
```
hinzed1127: @patrickvinograd per your suggestion, creating the new user identity directly. I think the icn is the only thing technically required to get what we need in the current flow, but not sure what other attributes we want to include, if any

modules/openid_auth/app/controllers/openid_auth/v0/okta_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+
+        render json: {
+          "commands": [
+            {
+              "type": "com.okta.assertion.patch",
+              "value": [
+                {
+                  "op": "replace",
+                  "path": "/claims/dslogon_edipi/attributeValues/0/value",
+                  "value": mvi_profile[:edipi]
```
hinzed1127: this and the added value are arbitrary. just wanted to verify things were getting modified by Okta correctly and then update to what we want to pass through

modules/openid_auth/app/controllers/openid_auth/v0/okta_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+        mvi_response[:profile]
+      end
+
+      def okta_callback
+        user_attributes = parse_user_attributes(params)
+        mvi_profile = fetch_mvi_profile(user_attributes)
+
+        render json: {
+          "commands": [
+            { "mvi_profile": mvi_profile },
```
hinzed1127: is this getting processed further downstream? I thought anything that doesn't conform to the response Okta expects will get ignored

bshyong: leftover from local testing—will remove!

modules/openid_auth/app/controllers/openid_auth/v0/okta_controller.rb
```diff
@@ -0,0 +1,86 @@
+# frozen_string_literal: true
+
+require_dependency 'openid_auth/application_controller'
+
+module OpenidAuth
+  module V0
+    class OktaController < ApplicationController
+      skip_before_action :authenticate
+
+      def metadata
```
patrickvinograd: This looks like dead code, don't need a SAML metadata endpoint in the Okta hook.

modules/openid_auth/app/controllers/openid_auth/v0/okta_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+
+        raise mvi_response.error if mvi_response.error # TODO: add error logging
+        mvi_response[:profile]
+      end
+
+      def okta_callback
+        user_attributes = parse_user_attributes(params)
+        mvi_profile = fetch_mvi_profile(user_attributes)
+
+        render json: {
```
patrickvinograd: The way I would break  this up is, imagine we  were going to update more than just these two elements. Like 10 elements. You'd probably make a method like `create_replace_stanza(path, value)` that returned the first object in `commands` and a method like `create_add_stanza(path, value)` that returned the second object. Maybe the `create_add_stanza` would take an optional `nameformat` or `type` argument if any of those vary. 

Then have the `okta_callback` method call those in turn with the edipi and icn_with_aaid paths/values.

bshyong:
Setting up new PR for cleaner history:

PR #2991, opened 2019-04-23
Add endpoint for vso rep search

charleystran:
## Description of change
This change is to add an endpoint for the saml proxy to call which will say when a rep exists from OGC

## Testing done
rspec

## Acceptance Criteria (Definition of Done)
- [x] when first name and last name match request, the rep and poa code are returned
- [x] When no rep is found, a proper errors is returned

#### Applies to all PRs

- [x] Appropriate logging
- [x] Swagger docs have been updated, if applicable
- [x] https://github.com/department-of-veterans-affairs/vets-contrib/issues/1823
- [x] Does not contain any sensitive information (i.e. PII/credentials/internal URLs/etc., in logging, hardcoded, or in specs)
- [x] Provide which alerts would indicate a problem with this functionality (if applicable)

PR #3258, opened 2019-08-30
VHA Facilities from GIS

rtravitz:
## Description of change
This PR addresses [#2991](https://github.com/department-of-veterans-affairs/vets-contrib/issues/2991). It is part of a [larger epic](https://app.zenhub.com/workspaces/developervagov-5ac272fff41bfa50c0e46c48/issues/department-of-veterans-affairs/vets-contrib/2373) to change datasources for VHA facilities.

In order to run this code, you will need a new set of recordings that currently live on the `rt/gis-service-mocks` [branch in vets-api-mockdata](https://github.com/department-of-veterans-affairs/vets-api-mockdata/tree/rt/gis-service-mocks). 

Besides recording new mocks and VCR casettes for the tests, most of the changes for this datasource switch are in the new `gis_configuration.rb`, `gis_metadata_configuration.rb`, `gis_client.rb`, and `gis_metadata_client.rb`. Those clients and configurations are mostly duplicates of existing clients and configs for ArcGIS with a couple of notable exceptions. First, they specify different URLs. Second, they make use of Betamocks middleware if it's enabled. (Note: the default config has it enabled when developing locally).

Other notable changes:
- There are significant changes to the attributes map in `vha_facility.rb` because GIS has different keys than ArcGIS in its responses.
- There are changes to the regex used to determine what type of facility should be used to pull the attribute map. Another line of regex seemed easier than trying to shoehorn a very different URL scheme into the existing regex.
- Last updated times are handled differently when coming from GIS because GIS provides them as a unix timestamp instead of a string like ArcGIS did.
- A large number of keys were deleted from the expected list in `vha_facility.rb` because those keys do not exist in GIS. Most of those are for services that we will not determine from GIS.

## Testing done
- I re-recorded VCR cassettes that simulate data that will come from the new GIS backing service. After changing to those recordings, I made sure the existing tests passed to maintain current functionality. The only tests that were modified were those related to services, which will no longer come from GIS.
- I made sure that data can be pulled to a local database for all facility types from a clean start, both with and without Access to Care data already in Redis. I then checked to make sure that, that data could be returned correctly from the vets-api consumer-facing endpoint.

## Testing planned
- We will need to test these changes eventually from an environment like DEV that has VA network access and will not use the mocks I have created in the vets-api-mockdata repo.

#### Unique to this PR

- [x] VHA facilities from GIS and inserted into base_facilities

#### Applies to all PRs

- [x] Appropriate logging


lib/common/client/middleware/response/facility_parser.rb
```diff
@@ -29,7 +31,7 @@ def build_facility_attributes(location, mapping)
             facility.merge!(make_address_mappings(location, mapping))
             facility.merge!(make_benefits_mappings(location, mapping, facility['unique_id'])) if mapping['benefits']
             facility.merge!(make_hours_mappings(location, mapping))
-            facility.merge!(make_service_mappings(location, mapping)) if mapping['services']
+            facility.merge!(make_service_mappings(location, mapping, facility['unique_id'])) if mapping['services']
```
nfasulo: I realize this isn't something you introduced, but I wonder if this would be clearer if `make_service_mappings` and/or the `services` map was was named `health_services`. It's probably just me but I keep forgetting the benefits services are handled separately. Maybe renaming to `make_benefits_mappings` and/or the `benefits` map to `benefits_services` would make sense too.

config/settings/development.yml
```diff
@@ -15,3 +15,6 @@ betamocks:
 
 account:
   enabled: true
+
+locators:
+  mock_gis: true
```
nfasulo: This setting and the reference to it in the GIS client configurations are all that is required to make `https://dev-api.va.gov` use the mock data, correct?

lib/common/client/middleware/response/facility_parser.rb
nfasulo: I know you didn't introduce it and maybe it's just me, but I keep forgetting that `make_service_mappings` is only for health services. What do you think of renaming `make_service_mappings` -> `make_health_service_mappings` and `make_benefits_mappings` -> `make_benefits_service_mappings`? And/or we could rename the service mappings themselves (e.g. `mapping['services']` -> `mapping['health_services']`), but I think changing the methods would be enough.

lib/facilities/vha_facility.rb
```diff
@@ -6,35 +6,28 @@ class << self
...
-          (attrs['Extension']).blank? || (attrs['Extension']).zero?
-        result
-      end
-
       def zip_plus_four(attrs)
-        zip = attrs['Zip']
-        zip << "-#{attrs['Zip4']}" unless attrs['Zip4'].to_s.strip.empty?
+        zip = attrs['zip']
+        zip4_is_empty = attrs['Zip4'].to_s.strip.empty? || attrs['Zip4'].to_s.strip == '0000'
+        zip << "-#{attrs['Zip4']}" unless zip4_is_empty
```
nfasulo: Love that `zip` is all lowercase but `Zip4` isn't. :grimacing:

lib/facilities/vha_facility.rb
```diff
@@ -44,35 +37,30 @@ def satisfaction_data(attrs)
...
+        datum = FacilityWaitTime.find(attrs['Sta_No'].upcase)
         if datum.present?
           datum.metrics.each { |k, v| result[k.to_s] = v }
           result['effective_date'] = to_date(datum.source_updated)
         end
         result
       end
 
-      def arcgis_type
-        'VHA_Facilities'
```
nfasulo: I do like that you just moved this into `pull_source_data`. No need to pollute the class.

lib/facilities/vha_facility.rb
```diff
@@ -115,23 +103,10 @@ def outpatient_surgical_specialty_list
...
-           ComplementaryAlternativeMed DiagnosticServices ImagingAndRadiology LabServices
-           EmergencyDept EyeCare OutpatientMHCare OutpatientSpecMHCare VocationalAssistance
-           OutpatientMedicalSpecialty AllergyAndImmunology CardiologyCareServices UrgentCare
-           DermatologyCareServices Diabetes Dialysis Endocrinology Gastroenterology
-           InfectiousDisease InternalMedicine Nephrology Neurology Oncology
-           PulmonaryRespiratoryDisease Rheumatology SleepMedicine OutpatientSurgicalSpecialty
-           CardiacSurgery ColoRectalSurgery
-           GeneralSurgery Gynecology Neurosurgery Orthopedics PainManagement
-           PlasticSurgery Podiatry
-           ThoracicSurgery Urology VascularSurgery Rehabilitation WellnessAndPreventativeCare]
```
nfasulo: I didn't realize how many unused fields we had mapped for validation purposes.

lib/common/client/middleware/response/facility_parser.rb
edmkitty: +1 to this

rtravitz:
>     * There will be a PR for the mocked data repo at some point, right?
Yes! Just wanted to make sure this was on the right track before making the official PR. I'll add that now.

>     * Does `spec/support/vcr_cassettes/facilities/va/vha_facilities.yml` include all VHA facilities at the time of its pull or just a subset?
Yes, it includes all of the VHA facilities at the time of the pull.

>     * Can be deleted now `spec/support/vcr_cassettes/facilities/va/vha_facilities_limit_results.yml`? Is there any value to keeping it around?
It makes the tests run much faster. It's valuable to not parse the whole list when the assertions are just about one facility.

config/settings/development.yml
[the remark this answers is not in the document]
rtravitz: Yep, the `mock_gis` flag determines whether betamocks will be used for the `gis_configuration.rb` and `gis_metadata_configuration.rb`. My understanding is that having it set to true in `development.yml` should cause those clients to using mocks in a local environment.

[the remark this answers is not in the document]
rtravitz: I'm now realizing I need to double-check if the DEV environment also uses this settings file. Because we want DEV to pull from the real source.

lib/common/client/middleware/response/facility_parser.rb
rtravitz: This is a good suggestion. This whole thing is a bit tangled and tough to follow. Might be good to take a longer look at how these fields are merged together at some point.

lib/facilities/vha_facility.rb
rtravitz: The validation also only subtracts one way. If we include a field in this mapping, we fail the validation if it doesn't exist in the object that we receive. But we don't fail if the object starts adding additional fields.

PR #2692, opened 2018-12-27
Goodbye Token Based Authentication

saneshark:
## Description of change
Removes token based authentication. The only way to sign-in to ApplicationController controllers will be with a session cookie.

## Testing done
Had to rewrite nearly every spec.

## Testing planned
Some thorough regression testing would be ideal. Additionally verifying with leads from other teams to help review and test on staging when ready.

#### Unique to this PR
Introduces a mechanism for signing in specs that will be shared across all specs. No more reinventing the wheel 50 different ways.

#### Applies to all PRs

- [x] Appropriate logging
- [x] Swagger docs have been updated, if applicable
- [x] Provide link to originating GitHub issue, or connected to it via ZenHub
- [x] Does not contain any sensitive information (i.e. PII/credentials/internal URLs/etc., in logging, hardcoded, or in specs)
- [x] Provide which alerts would indicate a problem with this functionality (if applicable)

PR #3265, opened 2019-09-04
Rt/facilities websites

rtravitz:
## Description of change

This PR finishes [#2692](https://github.com/department-of-veterans-affairs/vets-contrib/issues/2692). It makes it so that we pull website urls for VHA facilities from a static CSV. Other facility types continue pulling their website urls from the original locations.

## Testing done
- Added tests for the website url service
- Added a test for where the new website urls integrate with the rest of the vha facility data pull
- Confirmed that the website urls appear in Postgres after running the Sidekiq job to pull all facility data

#### Unique to this PR

- [x] VHA facilities have website URLs that come from the website URLs spreadsheet maintained by Lighthouse.

#### Applies to all PRs
- [x] Appropriate logging

### DIFF
--- a/app/services/facilities/website_url_service.rb
+++ b/app/services/facilities/website_url_service.rb
@@ -4,13 +4,6 @@ require 'csv'
 
 module Facilities
   class WebsiteUrlService
-    FACILITY_TYPES = {
-      'va_benefits_facility' => 'VBA',
-      'va_health_facility' => 'VHA',
-      'va_cemetery' => 'NCA',
-      'vet_center' => 'VC'
-    }.freeze
-
     def initialize
       source = Rails.root.join('lib', 'facilities', 'website_data', 'websites.csv')
       station_websites = CSV.read(source, headers: true)
--- a/app/services/facilities/website_url_service.rb
+++ b/app/services/facilities/website_url_service.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+module Facilities
+  class WebsiteUrlService
+    def initialize
+      source = Rails.root.join('lib', 'facilities', 'website_data', 'websites.csv')
+      station_websites = CSV.read(source, headers: true).to_h
+      @websites = map_websites_to_stations(station_websites)
+    end
+
+    def find_for_station(id)
+      entry = @websites[id]
+      entry.nil? ? '' : entry['Website_URL']
+    end
+
+    private
+
+    def map_websites_to_stations(station_websites)
+      station_websites.each_with_object({}) do |station, hash|
+        station_id = station['StationNum']
+        hash[station_id] = station['Website_URL']
+      end
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


app/services/facilities/website_url_service.rb
```diff
@@ -0,0 +1,35 @@
...
+    }.freeze
+
+    def initialize
+      source = Rails.root.join('lib', 'facilities', 'website_data', 'websites.csv')
+      station_websites = CSV.read(source, headers: true)
+      @websites = map_websites_to_stations(station_websites)
+    end
+
+    def find_for_station(id, type)
+      unique_id = "#{id}_#{FACILITY_TYPES[type]}"
```
nfasulo:
```suggestion
      unique_id = "#{id}_#{BaseFacility.PREFIX_MAP[type].upcase}"
```
We could pull the types from `BaseFacility` and eliminate the map at the top of this file.

app/services/facilities/website_url_service.rb
```diff
@@ -0,0 +1,35 @@
...
+      entry = @websites[unique_id]
+      entry.nil? ? '' : entry
```
nfasulo: It looks like we currently return `"website": null` when there isn't one, not an empty string, so I think we can drop this line and modify the corresponding test.

app/services/facilities/website_url_service.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+
+    private
+
+    def map_websites_to_stations(station_websites)
+      station_websites.each_with_object({}) do |station, hash|
+        unique_id = "#{station['StationNum']}_#{station['Org']}"
```
nfasulo: I don't feel strongly about this, but I wonder if it is worth flipping the order of the fields in `unique_id` so that it matches the id we create in the serializer and send in the response. :man_shrugging:

app/services/facilities/website_url_service.rb
rtravitz: Pulled it from `BaseFacility` 👍

app/services/facilities/website_url_service.rb
rtravitz: Updated to return nil instead of an empty string!

app/services/facilities/website_url_service.rb
rtravitz: Flipped it and reversed it ◀️

app/services/facilities/website_url_service.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+module Facilities
+  class WebsiteUrlService
+    FACILITY_TYPES = {
```
jholton: This is no longer needed, right?

rtravitz: Correct, it got re-added by accident. Thanks @jholton!

PR #3269, opened 2019-09-06
Kr/get all vha services from redis

edmkitty:
## Description of change
https://app.zenhub.com/workspaces/developervagov-5ac272fff41bfa50c0e46c48/issues/department-of-veterans-affairs/vets-contrib/2989

In the past, the services for a facility were acquired from an Arcgis endpoint and Access to Care. We are no longer using the Arcgis endpoint, and we will now be receiving services data from Access to Care. 

Changes in this PR:
- DentalServices is retrieved from a hard coded list through Redis
- The `services_from_gis` method in FacilityParser is no longer being used, so I renamed it. Its purpose is now to collect health services from both wait time data and the redis dental services. 
- The OnlineScheduling flag is no longer available, so it was taken out. We were not using this anyway.
- I refactored the Facility Parser class to include another class (TempFacility). Since `mapping` and `location` were being passed around so often, it was an indication that we were trying to capture state, and the best way to do this is through a class. Suggestions on this new class name are welcome.

### DIFF
--- a/lib/common/client/middleware/response/facility_parser.rb
+++ b/lib/common/client/middleware/response/facility_parser.rb
@@ -18,7 +18,9 @@ module Common
             path_test = env.url.path.match(/(FacilitySitePoint_\w{3})/) if path_test.nil?
             path_part = path_test[1]
             facility_map = facility_klass(path_part).attribute_map
-            env.body['features'].map { |location_data| TempFacility.new(location_data, facility_map).build_facility_attributes }
+            env.body['features'].map do |location_data|
+              TempFacility.new(location_data, facility_map).build_facility_attributes
+            end
           end
 
           def facility_klass(path_part)
@@ -32,11 +34,11 @@ module Common
               @location = location_data
               @mapping = facility_map
 
-              @facility = { 
+              @facility = {
                 'address' => {},
                 'services' => {},
                 'lat' => @location['geometry']['y'],
-                'long' => @location['geometry']['x'] 
+                'long' => @location['geometry']['x']
               }
             end
 
@@ -47,24 +49,24 @@ module Common
               @facility['services']['benefits'] = map_benefits_services if @mapping['benefits']
               @facility['hours'] = make_hours_mappings
               @facility['services'] = map_health_services if @mapping['services']
-              
+
               @facility
             end
 
             def make_direct_mappings
-              %w[unique_id name classification website].each_with_object({}) do |name, attributes|
+              %w[unique_id name classification website].each_with_object({}) do |name, _attributes|
                 @facility[name] = strip(@location['attributes'][@mapping[name]])
               end
             end
 
             def make_complex_mappings
-              %w[access feedback phone].each_with_object({}) do |name, attributes|
+              %w[access feedback phone].each_with_object({}) do |name, _attributes|
                 @facility[name] = complex_mapping(name)
               end
             end
 
             def make_address_mappings
-              { 
+              {
                 'physical' => complex_mapping('physical'),
                 'mailing' => complex_mapping('mailing')
               }
@@ -79,11 +81,11 @@ module Common
 
             def calculate_standard_benefits
               cleaned_benefits = clean_benefits(complex_mapping('benefits'))
-              cleaned_benefits << 'Pensions' if has_pensions?
+              cleaned_benefits << 'Pensions' if pensions?
               cleaned_benefits
             end
 
-            def has_pensions?
+            def pensions?
               BaseFacility::PENSION_LOCATIONS.include?(@facility['unique_id'])
             end
 
@@ -111,7 +113,9 @@ module Common
             def complex_mapping(attr_name)
               attrs = @location['attributes']
               item = @mapping[attr_name]
+
               return {} unless item
+
               item.each_with_object({}) do |(key, value), hash|
                 hash[key] = value.respond_to?(:call) ? value.call(attrs) : strip(attrs[value])
               end
@@ -129,9 +133,7 @@ module Common
               id = @facility['unique_id'].upcase
               facility_wait_time = FacilityWaitTime.find(id)
 
-              if facility_wait_time&.source_updated.present?
-                Date.strptime(facility_wait_time&.source_updated).iso8601
-              end
+              Date.strptime(facility_wait_time&.source_updated).iso8601 if facility_wait_time&.source_updated.present?
             end
 
             def collect_health_services
@@ -158,14 +160,11 @@ module Common
 
             def dental_services(facility_id)
               services = []
-              if FacilityDentalService.exists?(facility_id)
-                services << { 'sl1' => ['DentalServices'], 'sl2' => [] }
-              end
+              services << { 'sl1' => ['DentalServices'], 'sl2' => [] } if FacilityDentalService.exists?(facility_id)
 
               services
             end
           end
-
         end
       end
     end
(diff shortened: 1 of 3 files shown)


lib/common/client/middleware/response/facility_parser.rb
```diff
@@ -18,129 +18,152 @@ def parse_body(env)
...
+            BaseFacility::PATHMAP[path_part]
           end
 
-          def make_direct_mappings(location, mapping)
-            %w[unique_id name classification website].each_with_object({}) do |name, attributes|
-              attributes[name] = strip(location['attributes'][mapping[name]])
-            end
-          end
+          class TempFacility
+            attr_accessor :facility, :location, :mapping
```
omgitsbillryan: The `TempFacility.facility` abstraction doesn't read very well. Can the classname or attribute name, be reworked to not collide with itself?

lib/common/client/middleware/response/facility_parser.rb
```diff
@@ -18,129 +18,152 @@ def parse_body(env)
...
-                         'long' => location['geometry']['x'] }
-            facility.merge!(make_direct_mappings(location, mapping))
-            facility.merge!(make_complex_mappings(location, mapping))
-            facility.merge!(make_address_mappings(location, mapping))
-            facility.merge!(map_benefits_services(location, mapping, facility['unique_id'])) if mapping['benefits']
-            facility.merge!(make_hours_mappings(location, mapping))
-            facility.merge!(map_health_services(location, mapping, facility['unique_id'])) if mapping['services']
-            facility
+          def facility_klass(path_part)
+            BaseFacility::PATHMAP[path_part]
```
omgitsbillryan: Does `facility_klass` get used anywhere else? Can we just get ride of this method and use `BaseFacility::PATHMAP[path_part]` directly inline?

lib/common/client/middleware/response/facility_parser.rb
```diff
@@ -18,129 +18,152 @@ def parse_body(env)
...
+              services << { 'sl1' => ['UrgentCare'], 'sl2' => [] } if facility_wait_time&.urgent_care&.any?
+              Facilities::AccessDataDownload::WT_KEY_MAP.each_value do |service|
+                services << { 'sl1' => [service.camelize], 'sl2' => [] } if metric_keys.include?(service)
+              end
+              services
+            end
+
+            def dental_services(facility_id)
+              services = []
+              services << { 'sl1' => ['DentalServices'], 'sl2' => [] } if FacilityDentalService.exists?(facility_id)
```
omgitsbillryan: What is `FacilityDentalService `? I can't find the definition.

spec/lib/facilities/vha_facility_spec.rb
```diff
@@ -93,7 +94,7 @@ module Facilities
               f2_health = f2_services['health']
 
               expect(f1_services.keys).to match(%w[last_updated health other])
-              expect(f1_services['last_updated']).to eq('2019-07-09')
+              expect(f1_services['last_updated']).to eq('2017-03-31')
```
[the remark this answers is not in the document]
edmkitty: This is a bit of a long story, I'll try to summarize. 
On the `master` branch, facilities pulls data from an arcgis endpoint that we are moving away from. The arcgis endpoint includes facility data, services data, mental health data, etc. The new gis endpoint we are using only includes facility data, not services data. On the epic branch, we updated the VCR to use data from gis but because the gis endpoint did not actually contain services data, the `last_updated` attribute was temporarily set to the date the facility data itself was updated. In this branch, the `last_updated` attribute is being set to data from Access to Care, i.e. the actual `last_updated` date for services. You can see where the old code was taken out in the removal of the `services_date` method.

lib/common/client/middleware/response/facility_parser.rb
edmkitty: Agreed! What do you think of `facility_hash` or maybe `temp_hash`? (My dream is to use the VHAFacility model object itself instead of this TempFacility class, but that refactor is too deep for this PR.)

edmkitty: I mean, the facility subclasses

lib/common/client/middleware/response/facility_parser.rb
edmkitty: We can get rid of it, but I like the `facility_klass` method because it adds context to this code, and an explanation for what it's for. Do you feel strongly that I get rid of it?

edmkitty:
> > DentalServices is retrieved from a hard coded list through Redis
> 
> ☝️ Can you explain a bit more about how this works?

Sure! We have a hard coded list of facilities in a csv that offer DentalServices. It is stored here: https://github.com/department-of-veterans-affairs/vets-api/tree/team-z/facilities-datasource-switchover/lib/facilities/dental_service_data

We have a sidekiq job that runs to update the Redis store with this data, in the event that we need to update it. The sidekiq job is here: https://github.com/department-of-veterans-affairs/vets-api/blob/team-z/facilities-datasource-switchover/app/workers/facilities/dental_service_reload_job.rb

Although this sequence will update the redis store, it does not update the facilities objects themselves. That is done during the access_data_download_job inside the `facility_parser.rb`. This is the feature change in this PR, done here: https://github.com/department-of-veterans-affairs/vets-api/pull/3269/files#diff-2241651a93f63c7f8072a02292cd4c76R161

lib/common/client/middleware/response/facility_parser.rb
omgitsbillryan: So a "facility" is :
```
'address'
'services'
'lat'
'long'
```
?
That looks a lot like what I would call a "location" (an attribute that already exists). What about simply flattening out `TempFacility? e.g.:
```
TempFacility
  - location
  - mapping
  - address
  - services
   - geo_location
    - lat
    - long
```
I'm not that familiar with what data this service is communicating, so I value your judgement over mine. My only real objection is the `Facility.facility` conflict.

edmkitty: The `facility` noted here has many more attributes than the ones you listed. Further confounding this, it is meant to represent multiple subtypes of facility, so there is no "one size fits all" object. The `facility` is meant to represent our format of facility, and the `location` is how the facility data is coming through from the GIS endpoint. The `facility_parser` translates the facility from gis and adds other things to the facility object, such as services. I think this warrants a heavier refactor at a later time. But I will rename the facility attribute for now.

lib/common/client/middleware/response/facility_parser.rb
omgitsbillryan: If I felt strongly, I'd deny the review ❌ . Otherwise, consider all feedback as suggestions that you can use or not use.

edmkitty: Ok, thank you!

lib/common/client/middleware/response/facility_parser.rb
nfasulo: Would things be clearer if we renamed `location` -> `src_facility` and `facility` -> `dest_facility`?

lib/common/client/middleware/response/facility_parser.rb
```diff
@@ -18,129 +18,152 @@ def parse_body(env)
...
+            BaseFacility::PATHMAP[path_part]
           end
 
-          def make_direct_mappings(location, mapping)
-            %w[unique_id name classification website].each_with_object({}) do |name, attributes|
-              attributes[name] = strip(location['attributes'][mapping[name]])
-            end
-          end
+          class TempFacility
+            attr_accessor :temp_hash, :location, :mapping
```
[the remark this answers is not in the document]
edmkitty: Good point, we do not

PR #1558, opened 2017-11-28
GIT_REVISION added to log output

jkassemi:
Provide the ability to filter logs by git revision to enable faster debugging. Previously we'd need to find the server IP addresses from AWS and deployment logs, then filter log events by IPs of those
servers.

Action item from: https://github.com/department-of-veterans-affairs/vets.gov-team/pull/5679/

PR #3272, opened 2019-09-06
Update installation instructions

johnpaulashenfelter:
## Description of change 

Update the installation instructions for manually installing the app

 ## Testing done 
[ ] Re-run all commands step-by-step on local machine

## Testing planned
[ ] Run through final instructions on another machine
[ ] Check on Ubuntu

Another machine can be boostrapped using these instructions

#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging
- [ ] Provide link to originating GitHub issue, or connected to it via ZenHub #1558

### DIFF
--- a/docs/setup/native.md
+++ b/docs/setup/native.md
@@ -97,7 +97,7 @@ the `vets-api` directory, run `./bin/setup-osx && source ~/.bash_profile && cd
        - `touch ./config/certs/vetsgov-localhost.crt`
        - `touch ./config/certs/vetsgov-localhost.key`
 1. Create dev database: `bundle exec rake db:setup`
-1. Go to the file `config/settings/development.yml` in your local vets-api. Switch the commented out lines pertaining to the cache_dir: uncomment out line 14 (what you use for running the app via Rails), and comment out line 15 (what you use for running the app via Docker).
+1. Go to the file `config/settings/development.yml` in your local vets-api. Switch the commented out lines pertaining to the cache_dir: uncomment out line 12 (what you use for running the app via Rails), and comment out line 13 (what you use for running the app via Docker).
 1. Make sure you have the [vets-api-mockdata](https://github.com/department-of-veterans-affairs/vets-api-mockdata) repo locally installed
 
 
--- a/docs/setup/native.md
+++ b/docs/setup/native.md
@@ -1,6 +1,3 @@
-## Native Setup
-
-If you can run docker, use the [Base Setup](/README.md#base-setup) instructions. Reach out to the #devops slack channel if you need to use these.
 
 ## Developer Setup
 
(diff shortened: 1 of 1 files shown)


johnpaulashenfelter:
I've got a lot of basic questions:

* Do we have a (semi-?)consistent format for branch names?
* Do we squash and merge (looks like it from the dropdown above)
* Do we delete old branches or does CI handle
* Not that it matters much for this one, but where is staging?

docs/setup/native.md
```diff
@@ -97,7 +97,7 @@ the `vets-api` directory, run `./bin/setup-osx && source ~/.bash_profile && cd
        - `touch ./config/certs/vetsgov-localhost.crt`
        - `touch ./config/certs/vetsgov-localhost.key`
 1. Create dev database: `bundle exec rake db:setup`
-1. Go to the file `config/settings/development.yml` in your local vets-api. Switch the commented out lines pertaining to the cache_dir: uncomment out line 14 (what you use for running the app via Rails), and comment out line 15 (what you use for running the app via Docker).
```
johnpaulashenfelter: Seems like since `config/settings/development.yml` is under source control and this is suggesting changing a source controlled file that maybe this setting should be overridden in `config/settings.local.yml`?

omgitsbillryan: This line of doc does not read well 😬 .

I agree with your opinion that this should be overridden in `settings.local.yml`. I double-checked the doc and confirmed that settings in `settings.local.yml` will override `settings/development.yml`.

[Source](https://github.com/railsconfig/config#accessing-the-settings-object)

I took a poll a while back to see how many people were using native V.S. docker for local `vets-api` testing, and it was overwhelmingly native.

jholton:
>     * Do we have a (semi-?)consistent format for branch names?

No

>     * Do we squash and merge (looks like it from the dropdown above)

Yes

>     * Do we delete old branches or does CI handle

We are responsible for deleting our own branches.

>     * Not that it matters much for this one, but where is staging?

Deployed environments:  dev.va.gov, staging.va.gov, www.va.gov (prod)

docs/setup/native.md
jholton: +1 for changing this text to recommend overriding settings in `settings.local.yml`.

docs/setup/native.md
```diff
@@ -59,7 +59,8 @@ the `vets-api` directory, run `./bin/setup-osx && source ~/.bash_profile && cd
     - `echo "DatabaseMirror database.clamav.net" > freshclam.conf`
     - `freshclam -v`
 1. Install [pdftk](https://www.pdflabs.com/tools/pdftk-the-pdf-toolkit/pdftk_server-2.02-mac_osx-10.11-setup.pkg)
-1. Install gem dependencies: `cd vets-api; bundle install`
+1. Install gem dependencies: `cd vets-api; bundle install --without production`
```
johnpaulashenfelter: This turns out to break `bundle outdated` which doesn't have a way to skip certain gems that I can find.

johnpaulashenfelter: Blocked it until I remove this. The local settings is useful, but this turns out to cause other problems with bundling.

jholton:
Looks like the numbered items are rendering weird:
image

PR #3273, opened 2019-09-09
add in lat and lng to nearby endpoint

whuang85:
allow lat lng query for nearby and verify it's either the address info or the lat and lng info to query the endpoint

## Description of change
This PR adds the ability to query the nearby facilities using lat and lng along with the current address parameters.  It checks if you are only passing in either the lat and lng fields or the address fields. More info from this [ticket](https://github.com/department-of-veterans-affairs/vets-contrib/issues/2955)

## Testing done
unit tests added and hitting the endpoint locally

#### Unique to this PR

- [x] Need to fix cyclomatic complexity

#### Applies to all PRs

- [x] Appropriate logging


app/models/nearby_facility.rb
```diff
@@ -82,6 +87,26 @@ def empty_resource_set?(response_body)
...
+        return ''
+      end
+    end
+
+    def location_type(params)
+      REQUIRED_PARAMS.each do |key, value|
+        return key if (value - params.keys.map(&:to_sym)).empty?
+      end
+      nil
+    end
```
edmangimelli: `find` might be a bit more idiomatic
```
    def location_type(params)
      REQUIRED_PARAMS.each do |key, value|
        break key if (value - params.keys.map(&:to_sym)).empty?
      end
    end
```

edmangimelli: `find` might be a bit more idiomatic
```
    def location_type(params)
      REQUIRED_PARAMS.find do |key, value|
        break key if (value - params.keys.map(&:to_sym)).empty?
      end
    end
```

dvogel: Avoiding opaque variable names and relying on the principle of one-concept-per-line, I'd write it like this:

```
    def location_type(params)
      obs_fields = params.keys.map(&:to_sym)                           # only do this once
      REQUIRED_PARAMS.find do |loc_type, req_field_names|              # more meaningful variable names
        any_missing_fields = (req_field_names - obs_fields).empty?     # give this list a meaningful name
        any_missing_fields and loc_type                                # rely on the find method false-y semantics
      end
    end
```

(comments just for this PR)

app/models/nearby_facility.rb
```diff
@@ -82,6 +86,26 @@ def empty_resource_set?(response_body)
...
+      values = REQUIRED_PARAMS[type].map { |k| params[k].to_s }
+      case type
+      when :address
+        return values.join(' ')
+      when :lat_lng
+        return values.join(',')
+      else
+        return ''
+      end
+    end
```
dvogel: It seems like the API interface has burrowed it's way down into the model layer. I could see a cleaner separation of concerns between the controller and the model if the model provided explicit methods for each type of query. Extending that idea, since each method would have it's own interface, the parameter requirements could be enforced by using kwargs.

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
```diff
@@ -51,32 +52,38 @@ def validate_params
...
       end
 
-      def validate_drive_time
-        Integer(params[:drive_time]) if params[:drive_time]
-      rescue ArgumentError
-        raise Common::Exceptions::InvalidFieldValue.new('drive_time', params[:drive_time])
+      def ambiguous(address_difference, lat_lng_difference, address_params, lat_lng_params)
+        address_difference.empty? || lat_lng_difference.empty? ||
+          (address_difference != address_params && lat_lng_difference != lat_lng_params)
       end
```
dvogel: Maybe rename this to `ambiguous?` (with question mark) to communicate the boolean predicate nature of the method?

app/models/nearby_facility.rb
edmangimelli: @dvogel will your use of `find` return what he wants? you can break from the block and return anything you want, but, if you let the block naturally return (if that makes sense), it will return an array of [key, value] (in this scenario (find on a hash)) 
```
{a: 1, b: 2}.find { |k,v| v % 2 == 0} # [:b, 2]


{a: 1, b: 2}.find { |k,v| break "#{v} is even!" if v % 2 == 0} # "2 is even!"
```
maybe i'm misunderstanding

dvogel: @edmangimelli you're right, I overlooked that bit. Using the first 3 comments as my rationale,  I would still write the last line of the block as:

```
any_missing_fields and break loc_type
```

Or, as applied to your example:

```
{a: 1, b: 2}.find { |k,v| (v % 2 == 0) and break "#{v} is even!" }
```

app/models/nearby_facility.rb
whuang85: @dvogel hopefully you can point me in the right direction.  I'm having some trouble using kwargs.. I refactored the NearbyFacility code to have two methods now (one for lat/lng, other for address) and I am using metaprogramming to call the appropriate method in NearbyController but it doesn't work when I converted the two NearbyFacility query methods to use kwargs.

I know the first problem I ran into was that it complained that the controllers params isn't a hash so I used

`resource = query_method.call(params.permit!.to_hash)`

but I seem to now run into a different issue and not sure what it would be.

modules/va_facilities/README.v1.yml
```diff
@@ -448,31 +449,49 @@ paths:
           in: query
           description: |
             Street address of the location from which drive time will be calculated.
-          required: true
+          required: false
```
edmkitty: Why are these changing to false? Aren't they required?

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
```diff
@@ -18,13 +18,25 @@ class NearbyController < ApplicationController
...
+        lat_lng: %i[lat lng].freeze
+      }.freeze
+      QUERY_INFO = {
+        address: NearbyFacility.method(:query),
+        lat_lng: NearbyFacility.method(:query_by_lat_lng)
+      }.freeze
+
       MISSING_PARAMS_ERR =
-        'Must supply street_address, city, state, and zip to query nearby facilities.'
+        'Must supply street_address, city, state, and zip or lat, and lng to query nearby facilities.'
```
nfasulo:
```suggestion
        'Must supply street_address, city, state, and zip or lat and lng to query nearby facilities.'
```

modules/va_facilities/lib/va_facilities/param_validators.rb
```diff
@@ -34,6 +34,31 @@ def validate_type_and_services_known
       raise Common::Exceptions::InvalidFieldValue.new('services', unknown) unless unknown.empty?
     end
 
+    def validate_street_address
```
nfasulo: We have been using `ParamValidators` for shared validation methods and leaving validation used by just one controller in that controller. If we are going to move all validators into here, I would suggest pulling them out of `FacilitiyController` as well.

whuang85: oh ok, yea I wasn't sure why it was there in the first place since other validators were in that class.  I'm happy to move all the facility related param validators to the param_validators class if that's fine with you.

modules/va_facilities/README.v1.yml
whuang85: I was looking at v0 facilities openapi spec and it seemed to me that when having an either or situation that all the required params for each group had the required removed and was just documented in the endpoint description so I followed suit.

The other alternative is to have `street_address`, `city`, `state`, `zip` and `lat` and `lng` all required.  Not sure which option is more confusing

modules/va_facilities/app/controllers/va_facilities/v0/facilities_controller.rb
```diff
@@ -100,15 +100,6 @@ def verify_float(param)
         raise Common::Exceptions::InvalidFieldValue.new(param.to_s, params[param])
       end
 
-      def validate_bbox
```
edmkitty: If this method is being removed, why is it still being called in the `validate_params` method?

edmkitty: For that matter, why is this method being removed?

modules/va_facilities/README.v1.yml
edmkitty: Ah, got it.

modules/va_facilities/lib/va_facilities/param_validators.rb
```diff
@@ -34,8 +41,102 @@ def validate_type_and_services_known
...
+      params[:bbox]&.each { |x| Float(x) }
+    rescue ArgumentError
+      raise Common::Exceptions::InvalidFieldValue.new('bbox', params[:bbox])
+    end
+
+    def validate_a_param_exists(require_one_param)
+      lat_and_long = params.key?(:lat) && params.key?(:long)
+
+      if !lat_and_long && require_one_param.none? { |param| params.key? param }
+        require_one_param.each do |param|
```
jholton: At this point (via the conditional) we've established that none of the required_params exist.
`unless params.key? param` (the line below) is redundant.
Thus, the exception is going to be raised on the first value of `require_one_param`.
So, this whole method can be reduced to:
```
def validate_a_param_exists(require_one_param)
  lat_and_long = params.key?(:lat) && params.key?(:long)
  if !lat_and_long && require_one_param.none? { |param| params.key? param }
    raise Common::Exceptions::ParameterMissing.new(require_one_param.first.to_s, detail: MISSING_FACILITIES_PARAMS_ERR)
  end
end
```
Also, I don't think there is a spec testing this failure path.  Can we get a spec testing the path where the error is raised?
```
it 'fails validation when none of the required params are present' do
      @dummy_class.params = { blah: '75075' }
      expect{@dummy_class.validate_a_param_exists(REQUIRE_ONE_PARAM)}.to raise_error
    end
```

modules/va_facilities/lib/va_facilities/param_validators.rb
```diff
@@ -34,8 +41,102 @@ def validate_type_and_services_known
...
+    def validate_required_nearby_params(required_params)
+      param_keys = params.keys.map(&:to_sym)
+      address_params = required_params[:address]
+      lat_lng_params = required_params[:lat_lng]
+      address_difference = (address_params - param_keys)
+      lat_lng_difference = (lat_lng_params - param_keys)
+
+      unless valid_params?(address_difference, lat_lng_difference)
+        if ambiguous?(address_difference, lat_lng_difference, address_params, lat_lng_params)
+          raise Common::Exceptions::ParameterMissing.new(detail: AMBIGUOUS_PARAMS_ERR)
```
jholton: This line can never be reached.
`valid_params?` method eliminates any `address_difference` and `lat_lng_difference` combination from getting through that would allow `ambiguous?` to return `true`.

(There is no spec testing this path. That is, there is no spec testing the case when `AMBIGUOUS_PARAMS_ERR` would be raised.)

whuang85: Here's the scenario where it fails for being ambiguous

https://github.com/department-of-veterans-affairs/vets-api/blob/add_lat_and_lng_wh/modules/va_facilities/spec/validators/param_validators_spec.rb#L462

modules/va_facilities/lib/va_facilities/param_validators.rb
```diff
@@ -34,8 +41,104 @@ def validate_type_and_services_known
...
+      required_params = required_one_param - lat_lng
+      !lat_lng_exists && required_params.none? { |param| params.key? param }
+    end
+
+    def valid_params?(difference1, difference2)
+      (difference1.empty? && difference2.any?) || (difference2.empty? && difference1.any?)
+    end
+
+    def ambiguous?(address_difference, lat_lng_difference, address_params, lat_lng_params)
+      address_difference.empty? || lat_lng_difference.empty? ||
```
jholton: If I'm understanding the logic correctly, the `ambiguous?` method only works right now because it is wrapped in the `unless valid_params?` conditional.  The following change would make it work independently of that conditional.

Is this the logic?:
It is ambiguous If (both parameter sets are empty) OR if (both parameter sets have values present but are incomplete)

```suggestion
      (address_difference.empty? && lat_lng_difference.empty?) ||
```

whuang85: oops you are correct, changed it

whuang85:
I'm going to wait for that code climate config change to be pulled into master (https://github.com/department-of-veterans-affairs/vets-api/pull/3307) so I can ignore those 2 code climate issues.

PR #3285, opened 2019-09-10
require users to have multifactor and loa3 for ppiu access

lihanli:
## Description of change
https://github.com/department-of-veterans-affairs/vets.gov-team/issues/19647#issuecomment-529999662

## Testing done
automated tests

## Testing planned
test in staging

#### Applies to all PRs

- [x] Appropriate logging


app/controllers/v0/ppiu_controller.rb
```diff
@@ -3,6 +3,7 @@
 module V0
   class PPIUController < ApplicationController
     before_action { authorize :evss, :access? }
+    before_action { authorize :ppiu, :access? }
```
jholton: Can we get a spec showing that a 403 is returned if the user does not have ppiu access?
Currently, if you remove this line, all the specs still pass.

lihanli:
@jholton @kfrz ok i made the changes

lihanli:
@kfrz could i get another review on this? thanks

spec/request/ppiu_request_spec.rb
```diff
@@ -8,6 +8,17 @@
   let(:user) { create(:user, :mhv) }
   before(:each) { sign_in(user) }
 
+  def self.test_unauthorized(verb)
+    context 'with an unauthorized user' do
+      let(:user) { create(:user, :loa3) }
```
omgitsbillryan: So you're trying to test if a user is `loa3` & not multifactor? Would this ever actually occur in the wild?

I think this is only passing because you create a `:mhv` user (which is `multifactor = false`) only to later overwrite that user with an `:lo3` user, which surprisingly to me, doesn't specify a `mutlifactor`. Seems like some weird indirection to get the behavior you want.

lihanli: a loa3 user does not always have multifactor https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/models/user.rb#L161
the mhv user is `multifactor = true`

i wrote the spec for a loa3 user that doesn't have multifactor getting a 403, that seems reasonable to me, what do you think i should change or add to the spec?

omgitsbillryan: Whoops, I misunderstood. I didn't realized `:user` defaulted `multifactor` to `false`. This LGTM.

PR #3289, opened 2019-09-11
Add ruby version to Gemfile

johnpaulashenfelter:
## Description of change

Bundler supports checking a Ruby version so `bundle install` fails fast on the wrong Ruby version.

Feature was added in https://bundler.io/v1.12/gemfile_ruby.html

## Testing done

Added the line, bundled, then changed my ruby and system ruby with rvm. If versions didn't match, `bundle` failed.

## Testing planned

Checking deploy on staging. Not sure exaclty how that pipeline works

#### Unique to this PR

Verify it works with whatever ruby manager is on staging/production

#### Applies to all PRs

- [x] Appropriate logging

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,8 @@
 
 source 'https://rubygems.org'
 
+ruby '2.4.5'
+
 # Modules
 gem 'appeals_api', path: 'modules/appeals_api'
 gem 'claims_api', path: 'modules/claims_api'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -782,5 +782,8 @@ DEPENDENCIES
   yard
   zero_downtime_migrations
 
+RUBY VERSION
+   ruby 2.4.5p335
+
 BUNDLED WITH
    1.17.3


johnpaulashenfelter:
And `rvm` at least should respect the value in the `Gemfile` as well, even without a .ruby-version

https://rvm.io/workflow/projects#project-file-gemfile

dvogel:
This is specifically requiring mri 2.4.5 Why that specific version? Why not `'~> 2.4.5'` or even `'>= 2.4, < 3.0'`?

johnpaulashenfelter:
@dvogel I don't know a lot about how deployment works yet (eg are the servers running RVM/rbenv? single version? a docker-based version?) so not sure about the ramifications of opening up the version to a fuzzier range.

The `.ruby-version` file has a specific version for devs, so this PR just duplicates that by specifying the Ruby version we're already using in a canonical location where we also already specify all the other versions of Rubygems we're using. It potentially lets us remove `.ruby-version` altogether. 

I'd 👎 anything like < 3 because there that opens the door to a host of potential version conflict/surprise issues:

* an upgrade to ruby being released somewhere between dev => staging => production that puts a different version on production that was tested
* suprise bugs in official Ruby release (eg last week's https://bugs.ruby-lang.org/issues/16136)
* differences in available Ruby versions that are the "latest" on `apt`/etc vs rvm/rbenv/etc vs Docker/yard leading to differences between environments and even individual developer machines.

jholton:
> This is specifically requiring mri 2.4.5 Why that specific version? Why not `'~> 2.4.5'` or even `'>= 2.4, < 3.0'`?

The deployed environments are built from Docker images.  The Ruby installation is part of the Docker image.  When we upgraded to 2.4.5 was the latest available when we did our upgrade to 2.4.  
The plan was to move to some ruby version manager for further upgrades (2.5, 2.6).  
That was right before the VSP contract started, so that hasn't happened yet.
Hopefully we'll get the further Ruby upgrades prioritized soon.

johnpaulashenfelter:
@jholton what docker image do we point at? a custom one? Now I'm just curious 😄

jholton:
[This one.](https://github.com/department-of-veterans-affairs/vets-api/blob/09da1cfe0500e09a30462d6df79acde4efaddde5/Dockerfile#L1)

johnpaulashenfelter:
@jholton Awesome! That means if the Docker version changes the underlying Ruby, then you'll not be able to start the app (that's a good thing) because it won't match the Gemfile ruby. If you're paranoid you can even do things like lock the patchlevel (https://bundler.io/v1.12/gemfile_ruby.html).

Right now, if the `Dockerfile` was updated to Ruby 2.6, our app would happily deploy running on it (which I guess is sort of ok since hopefully the timing would put it through staging first and tests passing are tests passing) but none of the folks with the native install would know it since `.ruby-version` doesn't care what the Dockerfile says. The gem file ties it all together so *everything* matches before the app boots.

dogweather:
I like the strategy. Personally, I'm used to locking down the Ruby version because it's not semantically versioned. Plus, changes can be breaking: The latest version has a bad String bug, and it's a good idea to hold back until a fix comes out.

wyattwalter:
@jholton @johnpaulashenfelter Sorry, I'm late to the party here. The deployment is not on Docker today, the version is controlled here: https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/build/roles/vets-api/meta/main.yml#L3 It's pinned to latest 2.4 in Amazon's yum repos currently. Looks like the builds are still succeeding, but only because the version hasn't changed.

PR #3291, opened 2019-09-12
add profileShowDirectDeposit release toggle

rianfowler:
## Description of change
- adds direct deposit release toggle to flipper array 

## Testing planned

- will validate in staging today 

#### Applies to all PRs

- [ ] Appropriate logging


config/initializers/flipper.rb
```diff
@@ -43,7 +43,7 @@
       # default feautures to enabled for development and test only
       Flipper.enable(feature) if Rails.env.development? || Rails.env.test?
     end
-  rescue ActiveRecord::NoDatabaseError, ActiveRecord::StatementInvalid
```
annaswims: catching the all of the db errors pre-migrations was a game of whack-a-mole

config/features.yml
```diff
@@ -8,3 +8,6 @@ features:
     description: >
       On https://www.va.gov/find-locations/ enable veterans to search for Community care by showing that option in the "Search for" box.
       This toggle is owned by Rian
+  profile_show_direct_deposit:
```
rianfowler: perfect thank you!

config/initializers/flipper.rb
kfrz: Maybe add a `Raven.tags_context(feature: 'flipper')` so we can easily find all flipper related errors in sentry?

annaswims: All of the errors were when I tried to run a rake task like `rake db:migrate` or `rake db:setup` before the flipper before the flipper tables existed.  So, I don't expect any of the errors to be in prod.

config/features.yml
annaswims: One thing I'm wondering about now is if we want to have a singular toggle for "enable the API endpoint for direct deposit" and "show it in the UI".  If we had reason to shut it off we might want to make sure the backend also won't accept requests to update the banking info.

rianfowler: I think separate toggles work for short lived toggles (release toggles) but and that long term toggles (ops toggle) would cover it both systems. 

We also have the option to poll the toggle values and live update ui.

rianfowler:
I think this is ready for final approval.

PR #2837, opened 2019-02-25
Fix spelling error in service_history.yml

jeff2d2:
Fixes error in https://github.com/department-of-veterans-affairs/vets-contrib/issues/1391
"toekn" -> "token"

## Description of change
Fix misspelling in Veteran Verification/Service History endpoint API docs ("toekn" -> "token")

## Testing done
- [x] Checked yml preview

## Testing planned
None

## Acceptance Criteria (Definition of Done)
- [x] Spelling error fixed

#### Unique to this PR
None

#### Applies to all PRs

- [x] Appropriate logging
- [x] Swagger docs have been updated, if applicable
- [x] Provide link to originating GitHub issue, or connected to it via ZenHub (above)
- [x] Does not contain any sensitive information (i.e. PII/credentials/internal URLs/etc., in logging, hardcoded, or in specs)
- [x] Provide which alerts would indicate a problem with this functionality (if applicable)

PR #3293, opened 2019-09-12
Add mobile flag to Facilities API

rtravitz:
## Description of change
This PR finishes [#2837](https://github.com/department-of-veterans-affairs/vets-contrib/issues/2837)

It adds a flag that shows whether a facility is mobile. For now, only VHA facilities will show `true` or `false` for the field. Other types of facilities will remain `nil` in Ruby or `null` in the database or the json that returns to the end consumer. This is because ArcGIS continues to back non-VHA facilities and does not include information about mobile. 

#### These are the major changes:
- Adds a `mobile` column to the `base_facilities` table through a new database migration
- Changes the FacilityParser and VHAFacility to insert the mobile field into the database
- Changes the FacilitySerializer, GeoSerializer, and CsvSerializer to allow the mobile field to return to the end consumer
- Updates the documentation with language [figured out in a previous ticket](https://github.com/department-of-veterans-affairs/vets-contrib/issues/2729) to show this mobile field change

#### Other changes
- Deletes unneeded fixtures from the `va_facilities` module. All of our facilities tests actually use the fixtures at the top level. No tests fail after this change.
- Updates existing fixtures with default values for `mobile`.

## Testing done
- Adds a test to `vha_facility_spec.rb` asserting that `mobile` can go into and out of the database correctly
- Adds tests to `facilities_request_spec.rb` asserting that the `mobile` flag returns correctly to end consumers for the various facility types
- Verified that responses come back correctly for json/geojson/csv

#### Unique to this PR
- [x] Make database migration for adding "mobile" column to the base_facility model
- [x] Update serializers for json, geojson, and text/csv to return this column
- [x] Add mobile flag to facilities api

#### Applies to all PRs

- [x] Appropriate logging


spec/factories/nca_facilities.rb
```diff
@@ -121,6 +124,7 @@
     facility_type 'va_cemetery'
     classification 'Rural'
     website nil
+    mobile nil
```
johnpaulashenfelter: Why would all of these be nil? Isn't a cemetery clearly not mobile (`false`) or is this a limitation of the data?

modules/va_facilities/spec/requests/facilities_request_spec.rb
```diff
@@ -370,4 +370,35 @@ def query_params(url)
...
   end
+
+  context 'mobile flag' do
+    it 'responds with a boolean mobile flag for VHA facilities' do
+      create :vha_648A4
+      get base_query_path + '/vha_648A4', params: nil, headers: accept_json
+      expect(response).to be_successful
+      expect(response.body).to be_a(String)
+      json = JSON.parse(response.body)
+      expect(json['data']['attributes']['mobile']).to eq(false)
```
johnpaulashenfelter: If you want to check if the key *exists*, maybe `expect(json['data']['attributes']).to have_key('mobile')`

This is checking for a value which doesn't seem to be important for the spec.

spec/factories/nca_facilities.rb
```diff
@@ -159,6 +163,7 @@
     facility_type  'va_cemetery'
     classification 'National Cemetery'
     website 'https://www.cem.va.gov/cems/nchp/cypresshills.asp'
+    mobile nil
```
johnpaulashenfelter: Is nil a default? Seems like maybe it should be in the factory? And that raises the question if it should be a default in the db.

modules/va_facilities/spec/requests/facilities_request_spec.rb
edmkitty: Considering that the value can be nil and shouldn't be nil for vha facilities, I think it's important to check the value here. But it does seem important to check for the key as well.

spec/factories/nca_facilities.rb
edmkitty: We don't have a data source to confirm this yet, which is why we're filling in nil for mobile in the other facility types. Although a mobile cemetery does seem highly unlikely...

modules/va_facilities/spec/requests/facilities_request_spec.rb
johnpaulashenfelter: @edmkitty awesome -- then the 2nd check is would be that its' not nil. That describes the context a lot better (it exists, is not nil) -- this implies false is a somehow expected.

spec/factories/nca_facilities.rb
edmkitty: +1 to this. If we can have mobile default to `nil` in the db somehow then maybe we can avoid all these changes in the factories

modules/va_facilities/spec/requests/facilities_request_spec.rb
edmkitty: This is a good point. I like that method better. 👍

johnpaulashenfelter:
@rtravitz I went ahead and did a PR off this to fix the deprecations so its not dumped on your plate. https://github.com/department-of-veterans-affairs/vets-api/pull/3298. Left it as a draft. If you would pull that in, I'd appreciate it!

spec/factories/nca_facilities.rb
rtravitz: The DB does default to nil, conveniently! I was just trying to communicate intent by adding nil explicitly to these fixtures. But it seems like it causes more confusion than not including it at all. I can take them back out!

rtravitz:
> @rtravitz I went ahead and did a PR off this to fix the deprecations so its not dumped on your plate. #3298. Left it as a draft. If you would pull that in, I'd appreciate it!

@johnpaulashenfelter We're in a situation where we're pulling things into an epic branch right now because we need a bunch of things to all get merged to master at once as part of a datasource switchover. Because of that I won't pull in that PR directly, but I'll make the same changes you have in the draft. This epic branch should go to master in the next couple of weeks. Is that timeline okay for the deprecations?

PR #3294, opened 2019-09-12
SSOe - support unencrypted responses

bshyong:
## Description of change
API throws a 500 when parsing unencrypted SAML assertions. The root cause is because `decrypted_document` does not exist. The `ruby-saml` library does not appear to set the `decrypted document` variable for unencrypted responses; `document` should be used instead.

Reference: https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/response.rb#L68

Additionally, the AuthnContext for SSOe is temporarily hardcoded for testing purposes.

## Context
During integration testing with IAM, errors occurred during the parsing of the received SAML assertion. Initial assumption was that errors were due to an unexpected AuthnContext value. Further investigation indicated that the error was related to another issue relating to the use of unencrypted SAML assertions.

Screen Shot 2019-09-12 at 11 00 54 AM
http://sentry.vfs.va.gov/vets-gov/platform-api-development/issues/72108/?query=is%3Aunresolved%20authn

## Testing done
Local

## Testing planned
Testing on VA Network

- [x] Resolve SAML parsing issue
- [x] Updated Specs
- [x] Temporarily hardcoded AuthnContext for SSOe
- [x] Appropriate logging


lib/saml/responses/base.rb
```diff
@@ -62,7 +62,8 @@ def map_message_to_error(error_message)
       end
 
       def authn_context_text
-        REXML::XPath.first(decrypted_document, '//saml:AuthnContextClassRef')&.text
+        response_doc = assertion_encrypted? ? decrypted_document : document
```
bshyong: This is the main change—`authn_context` below is left alone because it is used broadly for statsD metrics and several other methods. The implicit assumption is that assertions received by VA.gov will always been encrypted (there is PII in the assertions)

lib/saml/responses/base.rb
```diff
@@ -62,7 +62,8 @@ def map_message_to_error(error_message)
...
+        REXML::XPath.first(response_doc, '//saml:AuthnContextClassRef')&.text
       end
 
       def authn_context
```
patrickvinograd: is `decrypted_document` truthy or falsy in the case we ran into with testing? Because if the latter, then won't  this never even get to calling `authn_context_text`?

If this feels to tricky to unwind, we can also consider just asking Prashant to enable encrypted assertions. It might make development a little tougher,  but. we have the ability to attach a. debugger and get the decrypted assertion if we need it.  Your call on that one.

bshyong: @patrickvinograd `decrypted_document` is nil in the case we ran into. In `/lib/saml/user.rb`, we call `authn_context_text` directly, which is why we got the 500—it will always get called when the response is parsed.

https://github.com/department-of-veterans-affairs/vets-api/blob/35b2d98c278c65132f4c9c9654b40e3bf912bae2/lib/saml/user.rb#L76

A little confusing but in the user object (`lib/saml/user.rb`), `authn_context` calls `saml_response.authn_context_text`, and in the _response object_ (`lib/saml/responses/base.rb`)  calling authn_context will return `SAML::User::UNKNOWN_AUTHN_CONTEXT` if the assertion is not encrypted.

I left the `authn_context` method on the SAML response class untouched because there are existing session controller specs that explicitly check for `SAML::User::UNKNOWN_AUTHN_CONTEXT` if decrypted_document is falsey, whereas for the purposes of our test and the use case in `lib/saml/user.rb`, it is fine for `authn_context` to return `nil`

patrickvinograd: Yeesh, yeah, I did not follow that from the git diff :) 👍

lib/saml/user.rb
```diff
@@ -79,7 +80,17 @@ def authn_context
       raise
     end
 
+    # For future use; can check if issuer matches /eauth\.va\.gov/ for SSOe
+    def issuer
+      saml_response.issuers[0]
+    rescue StandardError
```
annaswims: rubocop was just upgraded in master and now wants the just `rescue` with `StandardError` being implicit

lib/saml/user.rb
```diff
@@ -20,7 +20,8 @@ class User
       'dslogon_multifactor' => { loa_current: nil, sign_in: { service_name: 'dslogon' } },
       'dslogon_loa3' => { loa_current: '3', sign_in: { service_name: 'dslogon' } },
       'myhealthevet' => { loa_current: nil, sign_in: { service_name: 'myhealthevet' } },
-      'ssoe' => { loa_current: nil, sign_in: { service_name: 'ssoe' } },
+      # SSOe context
+      'urn:oasis:names:tc:SAML:2.0:ac:classes:Password' => { loa_current: nil, sign_in: { service_name: 'ssoe' } },
```
annaswims: `urn:oasis:names:tc:SAML:2.0:ac:classes:Password` would make a good consttant

spec/support/saml/response_builder.rb
```diff
@@ -62,6 +62,7 @@ def build_saml_response(authn_context:, account_type:, level_of_assurance:, mult
         multifactor: multifactor
       )
       saml_response = SAML::Responses::Login.new(document_partial(authn_context).to_s)
+      allow(saml_response).to receive(:assertion_encrypted?).and_return(true)
```
annaswims: do you want to return false for SSOe?

bshyong: @annaswims I set this to true in this spec support method because we implicitly assume that we are always dealing with encrypted responses—setting this to true forces the tests that mock out a SAML response to always check `decrypted document`

PR #3295, opened 2019-09-13
small mvi rspec cleanup

omgitsbillryan:
Followup to https://github.com/department-of-veterans-affairs/vets-api/pull/3247

I'm still a bit baffled on this one. I'm guessing a config changed somewhere in the middle of testing, but this code seems to be not needed...

### DIFF
--- a/spec/lib/mvi/service_spec.rb
+++ b/spec/lib/mvi/service_spec.rb
@@ -281,9 +281,6 @@ describe MVI::Service do
       let(:base_path) { MVI::Configuration.instance.base_path }
       it 'should raise a service error', :aggregate_failures do
         allow_any_instance_of(Faraday::Connection).to receive(:post).and_raise(Faraday::TimeoutError)
-        expect(Rails.logger).to receive(:warn).with(
-          'Could not load MVI SSL cert: No such file or directory @ rb_sysopen - /fake/client/cert/path'
-        )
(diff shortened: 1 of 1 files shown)


spec/lib/mvi/service_spec.rb
```diff
@@ -281,9 +281,6 @@
       let(:base_path) { MVI::Configuration.instance.base_path }
       it 'should raise a service error', :aggregate_failures do
         allow_any_instance_of(Faraday::Connection).to receive(:post).and_raise(Faraday::TimeoutError)
-        expect(Rails.logger).to receive(:warn).with(
```
johnpaulashenfelter: Is this mostly checking if the certificate was checked?

I've got very little context too, but this is expecting to fail to load a certificate after a certificate AND to do a gateway timeout? Seems like two pretty different things. 

If the test on the certificate check was necessary, i certainly wouldn't rely on the `Rails.logger` to log a message -- I'd expect to get an exception from whatever is doing the checking.

omgitsbillryan: > Is this mostly checking if the certificate was checked?

No. It was a hack just to get a spec to pass, after we changed a Rails.logger level from `:error` to `:warn' - context: #3247. I'm not convinced that it ever caused the spec to pass in the first place but I still can't explain why...

PR #3079, opened 2019-05-30
Update Raven with more context

edmkitty:
The original ticket was to investigate Pundit errors and see if they were returning useful information to the user. The Pundit errors do return useful info and a 403 status, which means they are functioning as expected. The problem here then, was that Sentry was improperly tagging the level as "error" when it should be "info." This PR updates the way we do Sentry logging to include more context, including error responses and status code. It also attempts to tag all 400-level responses as "info." 

https://github.com/department-of-veterans-affairs/vets-contrib/issues/2048

PR #3297, opened 2019-09-13
509 evss backend sentry error to info

LindseySaari:
## Description of change
This change reduces Sentry noise by logging the EVSSBackendServiceError with a level of info instead of error, no longer creating a sentry error alert. After discussion, triggering a sentry alert is not necessary.

## Testing done
Spec tests run & pass
Sentry logging remains the same for other error logging

- [x] Spec tests passing
- [x] Sentry alerts are no longer recording EVSSBackendServiceError with a level of error
- [x] The EVSSBackendServiceError is an info-level event

#### Applies to all PRs
- [x] Appropriate logging


jholton:
I missed the [first PR](https://github.com/department-of-veterans-affairs/vets-api/pull/3079/files) that this one builds off of.  I would have said this then:

I think should move logic that changes an error's 'level' into a Sentry [processor](https://github.com/department-of-veterans-affairs/vets-api/tree/master/lib/sentry/processor).  We already have one that sets certain exceptions to `:warn`.  Now we have multiple places where 'levels' are set. 

We should keep this type of logic in one place (processors, IMO).  Otherwise we are asking for a situation in the future where `sentry_logging` sets a level and then a processor sets it to something else.

lib/sentry_logging.rb
```diff
@@ -65,10 +65,15 @@ def non_nil_hash?(h)
...
-    va_exception_errors.present? && va_exception_errors.detect { |h| client_error_status?(h[:status]) }.present?
+    va_exception_errors.present? &&
+      va_exception_errors.detect { |h| client_error_status?(h[:status]) || evss_503?(h[:code], h[:status]) }.present?
   end
 
   def client_error_status?(status)
     (400..499).cover?(status.to_i)
   end
+
+  def evss_503?(code, status)
```
kfrz: I think this might be better located in the `EVSS::ErrorMiddleware` class, perhaps on [line 37](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/evss/error_middleware.rb#L37). Reason being, this change is specific to the EVSS service, and we already have a conditional branch in the linked class handling 503 responses, at which point we could change the error level.

kfrz:
@jholton just FYI - we are going to merge this change for now to supress the particular EVSS 503 response, but we will be working to refactor our overall error handling in order to leverage a sentry processor/middleware pattern.

PR #3300, opened 2019-09-13
Upgrade to FactoryBot v5

johnpaulashenfelter:
## Description of change

The factorybot v5 upgrade is assisted by first moving to 4.11, handling all the deprecation warnings for factories, and then moving to v5.x.

## Testing done

- [x] Verified the specs still pass
## Testing planned

- [x] Verify on staging
- [x] All specs still pass

#### Applies to all PRs

- [x] Appropriate logging


spec/controllers/v0/sessions_controller_spec.rb
```diff
@@ -367,7 +364,7 @@ def expect_logger_msg(level, msg)
                    'mhvCorrelationId' => loa3_user.mhv_correlation_id,
                    'signIn' => { 'serviceName' => 'idme' },
                    'credential_used' => 'id_me',
-                   'expirationTime' => expire_at.iso8601(0))
+                   'expirationTime' => 30.minutes.from_now.iso8601(0))
```
johnpaulashenfelter: These *could* be wrapped in a `Timecop.freeze` block to get back to being _correct_ about this. The `iso8601(0)` function basically strips off seconds though, so unless it takes ~1 minute for this single spec to run, or we hit a ~1s run from :59 to :00 seconds (next minute) we'd be fine.

If we do need to wrap it, I would recommend/plan to use `travel_to` Rails helper instead

johnpaulashenfelter:
Don't let the number of changed files scare you --- the factories were all automatically corrected to move to the dynamic attributes `{ foo }` that FactoryBot 5.0 uses.

There's one spec (copied into a 2nd file verbatim) that failed thanks to a time/Timecop issue and is fixed as well. Turns out there was a false negative test so now this runs correctly for MFA checking the timestamps on SAML login 😴 .....

Gemfile
```diff
@@ -146,7 +146,8 @@ group :development, :test do
   gem 'rack-test', require: 'rack/test'
   gem 'rack-vcr'
   gem 'rspec-rails', '~> 3.5'
-  gem 'rubocop', '~> 0.52.1', require: false
```
annaswims: Rubocop upgrade deserves its own PR, but fortunately @thilton-oddball is on it https://app.zenhub.com/workspaces/vsp-5cedc9cce6e3335dc5a49fc4/issues/department-of-veterans-affairs/va.gov-team/1457

johnpaulashenfelter: @annaswims understood. Had to upgrade to use the auto conversion in `rubocop-rspec`. Pulling that gem out and resetting back to original before this is done. Good catch

dogweather:
FYI, we might need to adjust the Code Climate config for the new rubocop version.

spec/controllers/v0/sessions_controller_spec.rb
kfrz: I'd prefer to see it wrapped with `freeze_time { }` (`rails-5.2` alias of `travel_to Time.now`). It would probably be fine either way but might cause a headache to track down in the future. 

That said I think we should open a track of work to remove the basically deprecated `Timecop` dependency altogether. So I did that [here](https://github.com/department-of-veterans-affairs/va.gov-team/issues/2007)

johnpaulashenfelter: Yeah -- i did a quick pass with `freeze_time` and started down the "add it to the Rspec.config or just include here or what..." 

I would be HAPPY to help refactor this.

Gemfile
johnpaulashenfelter: Removed

spec/models/notification_spec.rb
```diff
@@ -5,7 +5,7 @@
 RSpec.describe Notification, type: :model do
   describe 'validations' do
     it 'has a valid factory' do
-      notification = build :notification
+      notification = create(:notification)
```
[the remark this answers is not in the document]
johnpaulashenfelter: I want to know too. I checked the built-in `FactoryBot.lint` command, which i'd like to migrate to, and by *default* it uses `create`. Building when you have dependent objects (that might themselves have dependencies) can get weird.

johnpaulashenfelter:
@thilton-oddball and @department-of-veterans-affairs/va-api-engineers 

Didn't think about it, but there's a rubocop issue with our config and how FactoryBot suggests handling factory attributes.

TL;DR The gem expects `{...}` for attributes which means multiple lines for hashes, which violates `Style/BlockDelimiters` in Rubocop. Do any of us really care?

This means we get something like this

```
FactoryBot.define do
  factory :vha_648A4, class: Facilities::VHAFacility do
    unique_id { '648A4' }
    name { 'Portland VA Medical Center-Vancouver' }
    facility_type { 'va_health_facility' }
    classification { 'VA Medical Center (VAMC)' }
    website { 'http://www.portland.va.gov/locations/vancouver.asp' }
    lat { 45.6394162600001 }
    long { -122.65528736 }
    address {
      { 'mailing' => {},
        'physical' => {
          'zip' => '98661-3753',
          'city' => 'Vancouver',
          'state' => 'WA',
          'address_1' => '1601 East 4th Plain Boulevard',
          'address_2' => nil,
          'address_3' => nil
        } }
    }
```

which to me reads a lot better than

```
FactoryBot.define do
  factory :vha_648A4, class: Facilities::VHAFacility do
    unique_id { '648A4' }
    name { 'Portland VA Medical Center-Vancouver' }
    facility_type { 'va_health_facility' }
    classification { 'VA Medical Center (VAMC)' }
    website { 'http://www.portland.va.gov/locations/vancouver.asp' }
    lat { 45.6394162600001 }
    long { -122.65528736 }
    address do
      { 'mailing' => {},
        'physical' => {
          'zip' => '98661-3753',
          'city' => 'Vancouver',
          'state' => 'WA',
          'address_1' => '1601 East 4th Plain Boulevard',
          'address_2' => nil,
          'address_3' => nil
        } 
      end
    }
```
*if* that even works. 

This whole `{..}` thing is the whole big breaking change in FactoryBot 5.x when they deprecated static attributes (https://github.com/thoughtbot/factory_bot/blob/master/NEWS.md#4110-august-15-2018) so there's not a big choice here. 

I think hash in hash in hash makes sense in our Javascript/JSON-centric world and app. Any objections to the change in `.rubocop.yml` to ignore factory files? for the `Style/BlockDelimiters` cop? Normally i'd pull it into a separate PR but this seems to go together.

kfrz:
@johnpaulashenfelter  I think the second example in your comment would read as:

```rb

FactoryBot.define do
  factory :vha_648A4, class: Facilities::VHAFacility do
    unique_id { '648A4' }
    address do
      { 
        'mailing' => {},
        'physical' => {
          'zip' => '98661-3753',
          'city' => 'Vancouver',
          'state' => 'WA',
          'address_1' => '1601 East 4th Plain Boulevard',
          'address_2' => nil,
          'address_3' => nil
        } 
      }
    end
  end
end
```

As a Rubyist I definitely prefer the `do ... end` multi-line block syntax, and the second example reads clearer and more idiomatic to me when written as above.

Unfortunately it seems like a lot of lift to *enforce* my preferred rule, and if I'm understanding correctly leaving the existing rule will fail linting for all the factories.... so I think your change is acceptable.

.rubocop.yml
```diff
@@ -147,3 +147,8 @@ Rails/HasManyOrHasOneDependent:
     - "app/models/saved_claim/education_benefits.rb"
     - "app/models/saved_claim.rb"
     - "app/models/terms_and_conditions.rb"
+
+Style/BlockDelimiters:
+  Exclude:
+    - "spec/factories/*.rb"
+    - "modules/*/spec/factories/*.rb"
```
omgitsbillryan: I think a concise "why"-style comment is warranted here. It could save a future engineer who is cleaning up exemptions some time.

johnpaulashenfelter: Fair enough. I think we should probably annotate *everything* in the rubocop.yml. And use an RFC doc to clarify it so there's a mechanism for changing that's not super-adHoc

johnpaulashenfelter:
@kfrz Not that its a surprise, but the `do/end` syntax does work as expected. Not sure its worth the effort to change that by had -- there are ~80 files :/

PR #3304, opened 2019-09-17
Direct deposit sentry filter

lihanli:
## Description of change
https://github.com/department-of-veterans-affairs/va.gov-team/issues/1733
filter request bodies from the `PPIUController`

## Testing done
tested on staging
http://sentry.vfs.va.gov/vets-gov/platform-api-staging/issues/60838/?query=is%3Aunresolved

#### Applies to all PRs

- [x] Appropriate logging


spec/lib/sentry/processor/filter_request_body_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  def self.expect_filter(filtered)
+    it "should#{filtered ? '' : "n't"} filter the request body" do
+      expect(result['request']['data']).public_send(
+        filtered ? 'to' : 'to_not',
+        eq(Sentry::Processor::PIISanitizer::FILTER_MASK)
+      )
+    end
+  end
+
+  context 'with data from ppiu' do
```
johnpaulashenfelter: `with data from a controller in FILTERED_CONTROLLER`?

spec/lib/sentry/processor/filter_request_body_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  context 'with data from another controller' do
+    before do
+      data['tags']['controller_name'] = 'health_care_applications'
+    end
+
+    expect_filter(false)
+  end
+
+  context 'with no request body' do
```
johnpaulashenfelter: `with no data key in the request hash`?

spec/lib/sentry/processor/filter_request_body_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      )
+    end
+  end
+
+  context 'with data from ppiu' do
+    expect_filter(true)
+  end
+
+  context 'with data from another controller' do
+    before do
```
johnpaulashenfelter: There's only one example here -- let's just set the data

spec/lib/sentry/processor/filter_request_body_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      data['tags']['controller_name'] = 'health_care_applications'
+    end
+
+    expect_filter(false)
+  end
+
+  context 'with no request body' do
+    before do
+      data['request'].delete('data')
```
johnpaulashenfelter: The  test setup here would be a lot clearer to inline.

lib/sentry/processor/filter_request_body.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Sentry
+  module Processor
+    class FilterRequestBody < Raven::Processor
+      FILTERED_CONTROLLERS = %w[ppiu].freeze
+
+      def process(data)
```
johnpaulashenfelter: Our other processors (eg `pii_sanitizer`) have a public `process` method and a private `sanitize` method that implements the details. It makes sense to me to keep that consistent format.

spec/lib/sentry/processor/filter_request_body_spec.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Sentry::Processor::FilterRequestBody do
+  let(:client) { double('client') }
+  let(:processor) { Sentry::Processor::FilterRequestBody.new(client) }
+  let(:result) { processor.process(data) }
+
+  let(:data) do
```
johnpaulashenfelter: data is not the best name here since you're also focused on the `data` key in the `request`

config/initializers/raven.rb
```diff
@@ -8,6 +8,7 @@
   config.processors << Sentry::Processor::EmailSanitizer
   config.processors << Sentry::Processor::PIISanitizer
   config.processors << Sentry::Processor::LogAsWarning
+  config.processors << Sentry::Processor::FilterRequestBody
```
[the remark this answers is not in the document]
lihanli: not sure i understand your question, `['request']['data']` is where sentry stores the request body. every sentry log not from sidekiq includes a request key, and every request that has a request body includes the `['request']['data']` key

lib/sentry/processor/filter_request_body.rb
```diff
@@ -0,0 +1,21 @@
...
+        stringified_data = data.deep_stringify_keys
```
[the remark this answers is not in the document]
lihanli: yes, because data can have symbol keys or string keys

spec/lib/sentry/processor/filter_request_body_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      'tags' => { 'controller_name' => 'ppiu', 'sign_in_method' => { 'service_name' => 'idme', 'acct_type' => nil } },
+      'request' =>
+       {
+         'data' =>
+          "{\n  \"account_type\": \"Checking\"}"
+       }
+    }
+  end
+
+  def self.expect_filter(filtered)
```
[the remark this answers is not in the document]
lihanli: i don't think it's hard to understand this function and it reduces duplicate code

spec/lib/sentry/processor/filter_request_body_spec.rb
lihanli: i don't think it's a bad idea to have a context with only one example, rspec itself has many specs like that https://github.com/rspec/rspec-core/blob/master/spec/rspec/core/formatters_spec.rb#L78

config/initializers/raven.rb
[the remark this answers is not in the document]
lihanli: to clarify, `['request']` is the not the request body and it has a key named `data`, `['request']['data']` is the full request body

lihanli:
@johnpaulashenfelter hey i made some changes, could i get another review on this? thanks

spec/lib/sentry/processor/filter_request_body_spec.rb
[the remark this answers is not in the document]
johnpaulashenfelter: Reducing duplicate code isn't the primary goal of a test -- understanding what happens in the test is. Having to mentally parse a function that takes a param and returns "should" and "shouldn't" is a lot more mental baggage to deal with than writing out "should" and "shouldn't". Anyone new coming in has to add the extra hassle of figuring that out -- doesn't seem worth it to make it harder for future devs/support and the alternative is clearer.

spec/lib/sentry/processor/filter_request_body_spec.rb
johnpaulashenfelter: These two specs have done line that requires a fair bit of parsing to understand what's going on. I'd like to see specs that are clear. I can live with the context. I don't like that the whole spec relies on mentally untangling a lot of other code when the actual spec only is a few lines long.

config/initializers/raven.rb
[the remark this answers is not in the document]
johnpaulashenfelter: So this is sanitizing the data in the request, right? Is there a clearer name to communicate what it does? PIIBodySanitizer? 

Not directly an issue here, but seems like PIISanitizer should handle this too. But I get you're focusing on a single controller in this instance. I do think we'd likely want it in all controllers but I'm not positive.

johnpaulashenfelter: Should this come before the `LogAsWarning` to make sure anything that needs sanitized occurs before the log is written? I haven't done a deep dive into `Sentry::Processor` but the `<<` makes me think its an array-ish object that order can matter.

johnpaulashenfelter:
@lihanli I did a draft PR with a quick refactor. See what you think #3329

config/initializers/raven.rb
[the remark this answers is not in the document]
lihanli: this is filtering all request bodies for specific controllers regardless of whether or not there is PII in them, unlike PIISanitizer which filters only if certain PII keys are present. making the name similar to PIISanitizer sounds like it might cause confusion about what this class is doing.

lihanli: LogAsWarning doesn't write the log, it sets the log level to warning instead of error

PR #3305, opened 2019-09-17
1457 upgrade rubocop

thilton-oddball:
## Description of change
Upgraded Rubocop to version 0.74.0 and corrected all offenses that it discovered.


annaswims:
we'll have to tell codeclimate to use the same version of rubocop.  We can talk about that today.

lib/saml/ssoe_health_status.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
```
annaswims: did this code sneak in here as part of a bad merge?

johnpaulashenfelter:
@thilton-oddball You're going to get some potential surprises on Rubocop runs on CodeClimate: https://docs.codeclimate.com/docs/rubocop

Run into this before IRL. We _probably_ should move to a `.codeclimate.yml` config file so that can be controlled better. Its also probably useful since I'm a little unsure how settings/privileges work on the open-source CodeClimate projects. /cc @annaswims probably knows more about how admins/etc are configured.

.rubocop.yml
```diff
@@ -99,6 +103,7 @@ Metrics/BlockLength:
     - 'app/controllers/v0/apidocs_controller.rb'
 
 # Don't worry about ambiguous blocks in RSpec
+# Still recommended to disable specs in accordance to this lint rule 8/2019
```
annaswims: can you take a look through this PR and make sure that all changes are yours?  for example, this comment came from @LindseySaari's branch, right?

annaswims:
@thilton-oddball This PR is big enough now.  Can we get this to a place where rubocop and specs pass (with todo's) so that we can deliver something?

app/controllers/v0/evss_claims_controller.rb
```diff
@@ -25,8 +26,9 @@ def show
     def request_decision
       claim = EVSSClaim.for_user(current_user).find_by(evss_id: params[:id])
       raise Common::Exceptions::RecordNotFound, params[:id] unless claim
+
       jid = service.request_decision(claim)
-      claim.update_attributes(requested_decision: true)
```
annaswims: can we revert all of the ` update_attributes` changes?  It would be a change in behavior that could possibly cause bugs so I'd rather not see it as part of a huge PR like this one.

PR #3102, opened 2019-06-07
[wip] Integration tests

lihanli:
## Description of change

integration tests, anonymous application works when ssh'd into staging
how to run in staging:
1. clone shallow vets-website
2. clone shallow vagov-content
3. install node
4. install chrome and chromedriver
5. run `npm run watch` in vets-website
6. run feature

## Testing done
tested in staging

#### Applies to all PRs

- [x] Appropriate logging
- [x] Swagger docs have been updated, if applicable
- [x] Provide link to originating GitHub issue, or connected to it via ZenHub
- [x] Does not contain any sensitive information (i.e. PII/credentials/internal URLs/etc., in logging, hardcoded, or in specs)
- [x] Provide which alerts would indicate a problem with this functionality (if applicable)

PR #3307, opened 2019-09-17
commit codeclimate config

annaswims:
Commit codeclimate config to .yml so that all who can write to the repo can modify config. 

I took codeclimate defaults + current plugin settings (though I did remove some unnecessary patterns from the `exclude_patterns`)

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,47 @@
+version: "2"         # required to adjust maintainability checks
+checks:
+  argument-count:
+    config:
+      threshold: 4
+  complex-logic:
+    config:
+      threshold: 4
+  file-lines:
+    config:
+      threshold: 250
+  method-complexity:
+    config:
+      threshold: 5
+  method-count:
+    config:
+      threshold: 20
+  method-lines:
+    config:
+      threshold: 25
+  nested-control-flow:
+    config:
+      threshold: 4
+  return-statements:
+    config:
+      threshold: 4
+  similar-code:
+    config:
+      threshold: # language-specific defaults. an override will affect all languages.
+  identical-code:
+    config:
+      threshold: # language-specific defaults. an override will affect all languages.
+plugins:
+  rubocop:
+    enabled: true
+    channel: rubocop-0-52
+  brakeman:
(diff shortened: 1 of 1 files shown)


.codeclimate.yml
```diff
@@ -0,0 +1,47 @@
...
+    config:
+      threshold: # language-specific defaults. an override will affect all languages.
+  identical-code:
+    config:
+      threshold: # language-specific defaults. an override will affect all languages.
+plugins:
+  rubocop:
+    enabled: true
+    channel: rubocop-0-52
+  brakeman:
```
johnpaulashenfelter: Does brakeman have a version/channel? We've got a warning on the newest version when I tried running it locally last week.

.codeclimate.yml
```diff
@@ -0,0 +1,47 @@
...
+    enabled: true
+  bundler-audit:
+    enabled: true
+exclude_patterns:
+  - 'config/'
```
johnpaulashenfelter: Should we add `docs/`?

.codeclimate.yml
```diff
@@ -0,0 +1,47 @@
...
+  - 'db/'
+  - 'features/'
```
johnpaulashenfelter: This isn't part of our structure

.codeclimate.yml
```diff
@@ -0,0 +1,47 @@
...
+  - 'script/'
```
johnpaulashenfelter: This isn't part of our structure

.codeclimate.yml
annaswims: nope. doesn't look like it supports a channel param https://docs.codeclimate.com/docs/brakeman

.codeclimate.yml
annaswims: it was part of the defaults, but I'll take it out.  At first I was thinking this WIP was using it https://github.com/department-of-veterans-affairs/vets-api/pull/3102

annaswims:
I intend to come back to the config soon to alter defaults, but this at least gets the config to a state where anyone can change/update it.

PR #3308, opened 2019-09-17
Kr/add mental health etl download job

edmkitty:
## Description of change
This change is to add all the code for adding the mental health data to the vha facilities. Previously, the facilities api received facility data along with mental health data (and other facility data) from a public arcgis endpoint. We are changing the data source. The new data source for mental health data will be a database on the SQL52 server behind the VA network. The mock database for development is in a different PR. 

This change includes the following pieces:
- A Redis store for mental health data
- A SQL52 client with the intention of being reused for other SQL52 databases
- A MentalHealth client that inherits from the SQL52 client. It specifies a database. 
- A download job from the SQL52 server to the Redis store
- A change to the vha_facility model to add data from Redis to the model in Postgres

Relevant ticket: 
https://github.com/department-of-veterans-affairs/vets-contrib/issues/2992


lib/facilities/sql52_client.rb
```diff
@@ -0,0 +1,25 @@
...
+        username: Settings.sql_52.username,
+        password: Settings.sql_52.password,
+        host: Settings.sql_52.hostname,
+        port: Settings.sql_52.port,
+        database: database_name
+      }
+    end
+
+    def database_name
+      # fill this out in the subclasses
```
johnpaulashenfelter: Maybe an error would work better than a comment.

`fail NotImplementedError, "A Facilities::SQL52Client requires a database_name`

lib/facilities/mental_health_client.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    def database_name
+      Settings.sql_52.facilities_mental_health.database_name
+    end
+
+    def mental_health_table
+      Settings.sql_52.facilites_mental_health.table_name
+    end
+
+    def order_by_column
```
johnpaulashenfelter: Not sure making this constant a private method really improves the code -- simpler just to inline it on L12

lib/facilities/mental_health_client.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+
+    private
+
+    def select_all_mental_sql_query
+      "SELECT * FROM #{mental_health_table} ORDER BY #{order_by_column} ASC"
+    end
+
+    def database_name
+      Settings.sql_52.facilities_mental_health.database_name
```
johnpaulashenfelter: Will this (and the table name) change from environment to environment?

Gemfile
```diff
@@ -88,6 +88,7 @@ gem 'sidekiq-instrument'
 gem 'staccato'
 gem 'statsd-instrument'
 gem 'swagger-blocks'
+gem 'tiny_tds', '~> 0.7.0'
```
johnpaulashenfelter: Locking this to such an old version seems like it needs an explanation. The newest is 2.1.2 and there are a lot of changes around TDS client, openssl, etc.

spec/lib/facilities/vha_facility_spec.rb
```diff
@@ -64,6 +64,38 @@
         end
       end
 
+      context 'mental health data' do
+        before(:each) do
+          FacilityMentalHealth.keys.map { |k| FacilityMentalHealth.delete(k) }
```
nfasulo: This cleanup is handled for you, isn't it?

spec/jobs/facilities/mental_health_reload_job_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+        },
+        {
+          'StationNumber' => '202A', 'MHPhone' => '321-123-1234', 'Extension' => '0002', 'Modified' => '2019-09-06T13:00:00.000'
+        }
+      ]
+    end
+  end
+
+  before(:each) do
+    FacilityMentalHealth.keys.map { |k| FacilityMentalHealth.delete(k) }
```
nfasulo: This cleanup is handled for you, isn't it?

edmkitty: Is it? I wasn't sure, which is why I added this.

lib/facilities/mental_health_client.rb
edmkitty: No, it will be the same in every environment.

Gemfile
edmkitty: Took out tinytds completely. We will be storing the data on the vets api server temporarily while we find out a solution for the sql52 server connection.

config/settings.yml
```diff
@@ -392,3 +392,14 @@ bing:
...
+
+oit_lighthouse2:
+  sql52:
+    hostname: vhacdwdwhsql52.vha.med.va.gov
+    username: fake_username
+    password: fake_password
+    port: 1433
+    facilities_mental_health:
+      database_name: OIT_Lighthouse2
+      schema_name: A01
```
nfasulo:
```suggestion
      schema_name: App
```
Learned this morning that the app utility account only has permissions to the `App` schema.

config/settings.yml
```diff
@@ -392,3 +392,14 @@ bing:
...
+      table_name: OMHSP_PERC_Share__DOEx__FieldDataEntry_MHPhone
```
nfasulo:
```suggestion
      table_name: VHA_Mental_Health_Contact_Info
```
This is my proposed name for the view created for the app utility account to query (it is filled with the data in the table `OMHSP_PERC_Share__DOEx__FieldDataEntry_MHPhone`, but I thought we might give the view a more readable name.

PR #3180, opened 2019-07-26
Reorder education form routing rules

dmccurdybah:
## Description of change
Reorder education form routing rules to prevent 1995 STEM from being sent to western RO for Philippines facilities

## Testing done
Unit tests updated and QA reviewed

## Testing planned
N/A

#### Unique to this PR

- [x] All 1995 STEM submissions are sent to the Buffalo RO

#### Applies to all PRs
- [x] Appropriate logging
- [x] Swagger docs have been updated, if applicable
- [x] Provide link to originating GitHub issue, or connected to it via ZenHub
- [x] Does not contain any sensitive information (i.e. PII/credentials/internal URLs/etc., in logging, hardcoded, or in specs)
- [x] Provide which alerts would indicate a problem with this functionality (if applicable)

PR #3314, opened 2019-09-18
Filter out Vet Centers and Mobile Vet Centers

rtravitz:
## Description of change

This PR addresses part of [#3180](https://github.com/department-of-veterans-affairs/vets-contrib/issues/3180). The second part is addressed in a [PR to the vets-api-mockdata repo](https://github.com/department-of-veterans-affairs/vets-api-mockdata/pull/93).

After previous work in vets-api PR #3258, we ended up with duplicated Vet Centers and Mobile Vet Centers in our cache. Vet Centers will continue to come from ArcGIS and need to be excluded from the GIS pull.

## Testing done

- Updated cassettes for existing specs covering this functionality

#### Unique to this PR

- [x] Vet Centers are excluded from the GIS pull and are not cached in Postgres
- [x] Update GIS mockdata to exclude Vet Centers

#### Applies to all PRs
- [x] Appropriate logging


rtravitz:
> I was thinking about a future move of the source of vet centers to GIS and that it will require a refactor of how that `where` clause is built. It would be nice if there was a test we could add now to save ourselves from accidentally including non-VHA facilities in the VHA_facilities class. I'm not sure that we have a good way to do that with how the request to GIS is made and mocked though. 🤔

I did some hand-wringing over this when making the switch. Like you mention, we present the stubs from VCR almost at the spot where the `where` clause is built. The most I can think is adding a test that asserts that nothing in our cassettes is a Vet Center, for now. But then we're testing out test data, and I'm uncertain whether that's worthwhile.

PR #3316, opened 2019-09-18
Update Betamocks docs to mention that order matters

rtravitz:
#### Quick Explanation
Updating a doc to mention that order can matter when applying Betamocks middleware.

#### Extra context
When mocking a service with Betamocks recently, I was confused about how to make the middleware only record what came from the backing service and not the transformations applied by other reponse middleware in the chain. @kreek mentioned that order can matter when applying the Betamocks middleware.


docs/setup/betamocks.md
```diff
@@ -26,8 +26,9 @@ without being connect to the VA VPN. By default all users have been mocked for M
 only M. Webb (vets.gov.user+228@gmail.com) will work for the other services unless their mock data has been added.
 
 ## Mocking a Service
-1. Add the Betamocks middleware to the service to be mocked. It should
-be the first response middleware listed in the connection block.
+1. Add the Betamocks middleware to the service to be mocked. Previously, it was recommended that it should be
```
kfrz: I'd opt to leave out the historical approach in a setup-type document, IMO it leads to more confusion.

```suggestion
If a service class implements response middleware, it is important to consider the order in which the middleware is stacked. For further details, refer to the [Faraday API documentation](https://www.rubydoc.info/gems/faraday#Advanced_middleware_usage). 

In the following example, Betamocks will only record the raw response from the backing service, and will not record any transformations applied by the `::FacilityParser` or `::FacilityValidator` middlewares.  

```ruby
def connection
  Faraday.new(base_path, headers: base_request_headers, request: request_options) do |conn|
    conn.use :breakers
    conn.request :json
    
    conn.response :raise_error, error_prefix: service_name
    conn.response :facility_parser
    conn.response :facility_validator
    conn.response :betamocks if Settings.locators.mock_gis
    
    conn.adapter Faraday.default_adapter
  end
end
```
```

**Comment by rtravitz** [comment]
> This has been a point of confusion in the past, I think it best to point to the Faraday docs and add the example given. Good catch.
> 
> Kind of a nitpick, 👉 I like to see third-person objective tense in documentation as much as humanly possible.

All makes sense to me @kfrz. Thanks for writing up changes!

File: docs/setup/betamocks.md [hunk 2]
```diff
@@ -42,6 +60,25 @@ def connection
 end
 ```
 
+However, if you have custom response middleware that changes regularly, it might be worth placing your Betamocks
```
kfrz: Can remove lines 63-81 of this diff

docs/setup/betamocks.md
```diff
@@ -26,8 +26,26 @@ without being connect to the VA VPN. By default all users have been mocked for M
...
+    
+    conn.response :raise_error, error_prefix: service_name
+    conn.response :facility_parser
+    conn.response :facility_validator
+    conn.response :betamocks if Settings.locators.mock_gis
+    
+    conn.adapter Faraday.default_adapter
+  end
+end
+the first response middleware listed in the connection block.
```
kfrz: extra line snuck in here too

docs/setup/betamocks.md
rtravitz: Hey @kfrz, is this diff out of date? I don't have those lines anymore in the most recent commit.

kfrz: yep I just pushed the update, sorry for the double-take, thought I could ninja it :)

PR #3317, opened 2019-09-18
Upgrading Rubocop to 0.74.0 and adding in rubocop-rails gem.

thilton-oddball:
## Description of change
Upgraded Rubocop to 0.74.0 and added in rubocop-rails gem.


.rubocop_todo.yml
```diff
@@ -0,0 +1,273 @@
+# This configuration was generated by
```
kfrz: Is this required by Codeclimate, or is it otherwise necessary to add this to the repository VCS? I'm not outright against it,  just want to make sure we know how it's used.

annaswims: It's meant to be transitional. Rubocop wants to change ALL THE THINGS.  So, rather than doing that in a single PR, we're going to only upgrade rubocop in this pr, then address the new rules in smaller batches. If we decide to ignore rules in the long term we'll add them to .rubocop.yml.

annaswims: *except for the couple tiny things that rubocop is complaining about, desptie the todo. Those are happening in this pr.

PR #3320, opened 2019-09-19
Flush stdout to display error output in jenkins

LindseySaari:
## Description of change
Jenkins is currently swallowing rubocop lint and migration failure console output. Developers need to see this information for CI debugging purposes. This change will flush (sync) stdout and force error output to appear immediately. Jenkins by default pushes console error output to a buffer (for some indeterminate amount of time) and usually holds the output until enough has
accumulated. In our case, our exit status in combination with buffer issues was causing error output to be swallowed. By flushing stdout, it forces the output to appear immediately instead of being held in the buffer.

## For reviewers: ****
The only relevant changes are these two commits here, the rest were for Jenkins testing purposes to purposely break the build and verify stdout flush was working properly. 
https://github.com/department-of-veterans-affairs/vets-api/pull/3320/commits/2f47d9f5bb60090a4ee18bc938405bd35dc7c446

## Testing planned
Investigation around the makefile and linting 
Testing a broken rubocop rule and broken migration to evaluate that jenkins console error output matches what displays during local testing

- [x] Jenkins displays correct error output for failures in rubocop and migrations
- [x] Local CI error output matches expected jenkins output

#### Applies to all PRs
- [x] Appropriate logging

Below are images of the newly improved Jenkins console error output for rubocop and migrations: 

Screen Shot 2019-09-20 at 9 23 04 AM
Screen Shot 2019-09-20 at 10 27 34 AM
Screen Shot 2019-09-20 at 10 27 40 AM


rakelib/security.rake
```diff
@@ -15,7 +15,7 @@ task :security do
   puts 'running bundle-audit to check for insecure dependencies...'
   exit!(1) unless ShellCommand.run('bundle-audit update')
   # TODO: remove --ignore CVE-2019-5420 after updating to rails 5.2
-  audit_result = ShellCommand.run('bundle-audit check --ignore CVE-2019-5420')
+  audit_result = ShellCommand.run('bundle-audit check')
```
johnpaulashenfelter: Thanks for doing this! Would love to see it as a separate commit

lib/saml/health_status.rb
```diff
@@ -46,7 +46,7 @@ def idp_certs_valid?
 
         begin
           signing_cert    = SettingsService.merged_saml_settings&.idp_cert_multi&.dig(:signing)&.first
-          encryption_cert = SettingsService.merged_saml_settings&.idp_cert_multi&.dig(:encryption)&.first
+          encryption_cert = SettingsService.merged_saml_settings&.idp_cert_multi.dig(:encryption).first
```
johnpaulashenfelter: Nice change while you're here. Also seems like a separate fixup PR/commit to me.

LindseySaari: Hey there, the purpose of this rubocop change was just for testing. The goal was to break the jenkins build to verify that stdout flush was working properly 😸

rakelib/security.rake
kfrz: Agreed this is good but for a different tiny-PR

kfrz: Can also remove the `#TODO` while you're at it

kfrz:
This looks fine after cherry picking unrelated change out. :rocket:

lib/saml/health_status.rb
johnpaulashenfelter: Gotcha. But... don't `&.` and `dig` accomplish similar things? And thus seems weird the way this is written to me at least.

In any case, glad its not part of the PR 😄

rakelib/security.rake
LindseySaari: Done ✔️

PR #3327, opened 2019-09-23
mapping classification for VHA

whuang85:
## Description of change
This provide a generic mechanism to transform the classification id into a human readable string

Closes ticket https://github.com/department-of-veterans-affairs/vets-contrib/issues/3181

## Testing done
Updated the VCR tests to make sure VHA classification was changed from a number to a human readable string and the other facilities tests to ensure those did not change.

## Testing planned
Testing was done via VCR tests and hitting the endpoint manually to verify numbers were no longer found in the "classification" section of VHA facilities

#### Unique to this PR
N/A

#### Applies to all PRs

- [X] Appropriate logging


lib/common/client/middleware/response/facility_parser.rb
```diff
@@ -162,6 +168,12 @@ def dental_services(facility_id)
 
               services
             end
+
+            def get_mapper(key)
+              @mapping['mappers']&.find do |mapping_key, mapper|
+                break mapper if mapping_key.eql?(key)
```
edmkitty: no need to break when using `.find`. Just this should work 
```suggestion
                mapping_key.eql?(key)
```

lib/facilities/vha_facility.rb
```diff
@@ -76,10 +76,17 @@ def attribute_map
           'feedback' => { 'health' => method(:satisfaction_data) },
           'services' => health_services,
           'mobile' => 'Mobile',
-          'mapped_fields' => mapped_fields_list
+          'mapped_fields' => mapped_fields_list,
+          'mappers' => { 'classification' => method(:classification_mapping) }
```
edmkitty: This feels convoluted. Why do we need this extra `mappers` field?

lib/common/client/middleware/response/facility_parser.rb
```diff
@@ -63,6 +63,12 @@ def make_complex_mappings
               end
             end
 
+            def map_classification
+              mapper = get_mapper('classification')
+              src_value = @src_facility['attributes'][@mapping['classification']]
+              mapper.nil? ? src_value : mapper.call(src_value, @src_facility)
```
edmkitty: Maybe we could check the facility type here instead? For example, if the facility type is vha, then call the `classification_mapping` method

lib/facilities/vha_facility.rb
nfasulo: Instead of adding `mappers`, you should be able to update the value of `classification` above to use `classification_mapping`.

lib/common/client/middleware/response/facility_parser.rb
```diff
@@ -63,6 +63,12 @@ def make_complex_mappings
               end
             end
 
+            def map_classification
```
nfasulo: Can `classification` this just be included in the array for `make_complex_mappings`? Maybe not because it is only complex for one facility type, but if that is the case I think you should be able to use `complex_mapping` for VHA instead of adding the method `get_mapper`.

lib/common/client/middleware/response/facility_parser.rb
whuang85: > Maybe we could check the facility type here instead? For example, if the facility type is vha, then call the `classification_mapping` method

I was trying to keep it generic so there were no facility type checks in the parser but I can add it to simplify that logic

lib/common/client/middleware/response/facility_parser.rb
whuang85: > no need to break when using `.find`. Just this should work

Without the break, I believe it returns an array so it would return [{'classification' => my_method }]

PR #3328, opened 2019-09-23
Adding PDF Generation Disabled to schema

bastosmichael:
## Description of change

Fixes for ticket https://github.com/department-of-veterans-affairs/vets-contrib/issues/3096

## Testing done
- Rspec testing

## Testing planned
Testing in development

- Adding `"autoCestPDFGenerationDisabled":false` as default

#### Applies to all PRs

- [x] Appropriate logging


lib/evss/disability_compensation_form/data_translation.rb
```diff
@@ -19,6 +19,7 @@ def initialize(user, form_content, _)
       #
       def translate
         form['claimantCertification'] = true
+        form['autoCestPDFGenerationDisabled'] ||= false
```
johnpaulashenfelter: Not your fault, but the double negative (...disabled is false, which means enabled) was a little hard to understand but makes sense in the context of what's at the other end of the API. You want it to NOT DISABLE autogeneration by default which is the current behavior.

modules/claims_api/spec/fixtures/form_526_evss.json
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+{"form526":{"veteran":{"currentlyVAEmployee":false,"currentMailingAddress":{"city":"Portland","country":"USA","zipFirstFive":"12345","addressLine1":"1234 Couch Street","addressLine2":"Apt. 22","type":"DOMESTIC","zipLastFour":"6789","state":"OR"}},"applicationExpirationDate":"2015-08-28T19:53:45+00:00","serviceInformation":{"servicePeriods":[{"activeDutyEndDate":"1990-01-02","serviceBranch":"Air Force","activeDutyBeginDate":"1980-02-05"},{"activeDutyEndDate":"1999-01-01","serviceBranch":"Air Force Reserve","activeDutyBeginDate":"1990-04-05"}]},"disabilities":[{"ratedDisabilityId":"1100583","diagnosticCode":9999,"disabilityActionType":"NEW","name":"PTSD (post traumatic stress disorder)","secondaryDisabilities":[{"name":"PTSD personal trauma","disabilityActionType":"SECONDARY","serviceRelevance":"Caused by a service-connected disability\\nLengthy description"}]}],"claimantCertification":true,"autoCestPDFGenerationDisabled":false},"form526_uploads":[],"form4142":null,"form0781":null,"form8940":null}
```
johnpaulashenfelter: Just curious -- why is this JSON flattened? Makes it really hard to see the diff compared to a pretty version.

modules/claims_api/spec/fixtures/form_526_json_api.json
```diff
@@ -32,6 +32,7 @@
         }
       ],
       "standardClaim": false,
+      "autoCestPDFGenerationDisabled": false,
```
johnpaulashenfelter: THIS one _is_ a lot easier to diff.

modules/claims_api/spec/fixtures/form_526_evss.json
johnpaulashenfelter: Why is this one not pretty-formatted like the other fixtures? Makes the diff *really* hard to find.

spec/support/disability_compensation_form/saved_claim_with_4142.json
```diff
@@ -3,4 +3,4 @@
...
...et2\":\"1B\",\"city\":\"Baltimore\",\"state\":\"MD\",\"country\":\"USA\",\"postalCode\":\"21200-1111\"}},{\"providerFacilityName\":\"provider 4\",\"treatmentDateRange\":[{\"from\":\"1980-4-1\",\"to\":\"1985-4-1\"},{\"from\":\"1986-4-1\",\"to\":\"1987-4-1\"}],\"providerFacilityAddress\":{\"street\":\"101 Main Street\",\"street2\":\"1B\",\"city\":\"Baltimore\",\"state\":\"MD\",\"country\":\"USA\",\"postalCode\":\"21200-1111\"}},{\"providerFacilityName\":\"provider 5\",\"treatmentDateRange\":[{\"from\":\"1980-5-1\",\"to\":\"1985-5-1\"},{\"from\":\"1986-5-1\",\"to\":\"1987-5-1\"}],\"providerFacilityAddress\":{\"street\":\"102 Main Street\",\"street2\":\"1B\",\"city\":\"Baltimore\",\"state\":\"MD\",\"country\":\"USA\",\"postalCode\":\"21200-1111\"}}]},\"overflowText\":\"VA Form 21-4142/4142a has been completed by the applicant and sent to the PMR contractor for processing in accordance with M21-1 III.iii.1.D.2.\",\"claimantCertification\":true,\"autoCestPDFGenerationDisabled\":false,\"applicationExpirationDate\":\"2015-08-28T19:53:45+00:00\",\"directDeposit\":{\"accountType\":\"CHECKING\",\"accountNumber\":\"9876543211234\",\"routingNumber\":\"042102115\",\"bankName\":\"Comerica\"}}}" }
```
johnpaulashenfelter: Ditto the question about why this is flat

lib/evss/disability_compensation_form/data_translation.rb
bastosmichael: Yes, because the generation happens down stream there really isn't a way that I can think of to really test this on the application itself. We don't even get a response back from EVSS on when this occurs so difficult to even use a mock in the test when returning a response etc. The `||=` is intended to trigger only when form['autoCestPDFGenerationDisabled'] is nil or not included which should never happen with the json api validation but good to catch none the less.

modules/claims_api/spec/fixtures/form_526_json_api.json
bastosmichael: Agreed

spec/support/disability_compensation_form/saved_claim_with_4142.json
bastosmichael: This one is out of the scope of our modules we are responsible for so I'm more hesitant to modify this one especially since it appears to be string inside of json etc.

bastosmichael:
@johnpaulashenfelter Thanks for the review, because the generation happens down stream there really isn't a way that I can think of to really test this on the application itself. We don't even get a response back from EVSS on when this occurs so difficult to even use a mock in the test when returning a response etc. The `|| false` you mentioned is actuall `||= false` and is intended to trigger only when form['autoCestPDFGenerationDisabled'] is nil or not included which should never happen with the json api validation but good to catch none the less.

modules/claims_api/spec/fixtures/form_526_evss.json
bastosmichael: Unflattened

PR #3329, opened 2019-09-23
Refactor the specs for clarity

johnpaulashenfelter:
## Description of change

The specs now are each independent with clearer setup, exercise, and verification phases.

## Testing done

Refactors specs so continuing to pass is the only real test

## Testing planned

None.

- [ ] Specs continue to pass
 
#### Applies to all PRs

- [ ] Appropriate logging

### DIFF
--- a/spec/lib/sentry/processor/filter_request_body_spec.rb
+++ b/spec/lib/sentry/processor/filter_request_body_spec.rb
@@ -10,14 +10,14 @@ RSpec.describe Sentry::Processor::FilterRequestBody do
     end
 
     it 'filters PII found in a FILTERED_CONTROLLER' do
-      sentry_request = create_sentry_request(controller: 'ppiu')
+      sentry_request = create_sentry_request_with_pii(controller: 'ppiu')
       result = @processor.process(sentry_request)
 
       expect(result['request']['data']).to eql(Sentry::Processor::PIISanitizer::FILTER_MASK)
     end
 
     it 'ignores any contoller not specified in FILTERED_CONTROLLER' do
-      sentry_request = create_sentry_request(controller: 'another_controller')
+      sentry_request = create_sentry_request_with_pii(controller: 'another_controller')
 
       result = @processor.process(sentry_request)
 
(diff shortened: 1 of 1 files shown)


spec/lib/sentry/processor/filter_request_body_spec.rb
```diff
@@ -3,47 +3,54 @@
 require 'rails_helper'
 
 RSpec.describe Sentry::Processor::FilterRequestBody do
-  let(:client) { double('client') }
-  let(:processor) { Sentry::Processor::FilterRequestBody.new(client) }
-  let(:result) { processor.process(sentry_data) }
+  context 'with PII in the [:request][:data] hash' do
+    before(:each) do
+      client = double('client')
+      @processor = Sentry::Processor::FilterRequestBody.new(client)
```
johnpaulashenfelter: I'm tempted to just inline `double('client')` here since it doesn't affect anything

spec/lib/sentry/processor/filter_request_body_spec.rb
```diff
@@ -3,47 +3,54 @@
...
 
-    expect_filter(false)
+  def create_sentry_request_with_pii(controller: 'default')
+    {
+      'tags' => {
+        'controller_name' => controller,
+        'sign_in_method' => { 'service_name' => 'idme', 'acct_type' => nil }
+      },
+      'request' => {
+        'data' => "{\n  \"account_type\": \"Checking\"}"
```
johnpaulashenfelter: Is the whitespace (`\n`) important?

spec/lib/sentry/processor/filter_request_body_spec.rb
kfrz: Only argument against I can think of is if someone didn't notice and duplicated that could be avoided by leaving it at top of block.

spec/lib/sentry/processor/filter_request_body_spec.rb
kfrz: I'm not sure if this request body is strict or where it was generated/written, but ultimately I'd expect the processor to process the filtering regardless of excess `\n`

PR #3330, opened 2019-09-23
Rubocop offenses

thilton-oddball:
## Description of change
Cleaning up the remaining Rubocop offenses.


.rubocop.yml
```diff
@@ -139,11 +136,75 @@ Style/DateTime:
...
   Exclude:
     - "app/models/education_benefits_claim.rb"
     - "app/models/saved_claim/education_benefits.rb"
     - "app/models/saved_claim.rb"
     - "app/models/terms_and_conditions.rb"
+
+# Using .empty? with the safe navigation operator seems to be the correct call in this situation
+Lint/SafeNavigationWithEmpty:
+  Exclude:
+    - 'lib/hca/configuration.rb'
```
jholton: Let's get rid of this exclusion by changing `lib/hca/configuration.rb` to use `present?` instead of `empty?`
```
    CERT_STORE = if Settings.hca.ca.present?
                   cert_store(Settings.hca.ca)
                 else
                   nil
                 end
```

.rubocop.yml
```diff
@@ -139,11 +136,75 @@ Style/DateTime:
...
+
+# The suggestions here don't seem to make sense in most cases. Skip for now.
+Naming/MemoizedInstanceVariableName:
+  Enabled: false
+
+# The suggestions here appear that they would break intended behavior. Skip for now.
+Naming/RescuedExceptionsVariableName:
```
jholton: How can this break intended behavior?
Don't we just need to rename the rescued exception variable to `e` in each case to satisfy this cop?

.rubocop.yml
```diff
@@ -139,11 +136,75 @@ Style/DateTime:
...
+# Skip for now.
+Style/AccessModifierDeclarations:
+  Exclude:
+    - 'lib/sm/client.rb'
+    - 'modules/va_facilities/lib/va_facilities/api_serialization.rb'
+    - 'modules/va_facilities/lib/va_facilities/api_serialization_v1.rb'
+
+# The use of variables such as n to represent number, x and y for x-axis and y-axis, el for element, 
+# v for value, i for index, and e for error are all standard abbreviations and should be excluded.
+Naming/UncommunicativeMethodParamName:
```
jholton: I agree with this.  So, should we completely disable this cop instead of just excluding the files listed below?

.rubocop.yml
dogweather:
```ruby
Settings.hca.ca.present? ? cert_store(Settings.hca.ca) : nil
```

.rubocop.yml
```diff
@@ -139,11 +136,75 @@ Style/DateTime:
...
+    - 'lib/emis/responses/response.rb'
+    - 'lib/mvi/responses/profile_parser.rb'
+    - 'lib/mvi/service.rb'
+    - 'lib/pdf_fill/hash_converter.rb'
+    - 'lib/sentry_logging.rb'
+    - 'modules/vba_documents/app/workers/vba_documents/upload_processor.rb'
+    - 'rakelib/mvi.rake'
+
+# These do not seem to actually be using Kernel#open so should be excluded.
+Security/Open:
```
jholton: They are using `Kernel#open`.  When you just call `open` (instead of something like `File.open` it is being called on `Kernel`.
Glancing at the files listed below, I believe you can fix them by changing `open` calls to `File.open` in each file.

thilton-oddball: It looks like I was putting too much trust in SublimeText. It was listing the 'open' being used as a method coming from line 72 of evss_claim_base_serializer.rb

.rubocop.yml
thilton-oddball: If, in general, we are fine with any use of a less than three letter variable then I would agree that we should just disable this cop. I feel like when a less than three letter variable is used it generally is pretty obvious what it is intended to represent.

.rubocop.yml
thilton-oddball: This was a total mistake on my part.

.rubocop.yml
```diff
@@ -125,7 +125,6 @@ Style/FormatStringToken:
   Exclude:
     - "rakelib/mvi.rake"
 
-
 # Disabling this rule for now. We should consider changing.
```
jholton: Can we remove this comment?  We're not disabling it.  We're just excluding one file (which has a comment/explanation in line)

thilton-oddball: Yeah, there is also a comment on the cop just above it as well where they mention disabling it. I'll remove from both.

PR #3331, opened 2019-09-23
Temporarily set loa_highest to loa_current

bshyong:
## Description of change
Temporarily sets `loa_highest` to `loa_current` to facilitate SSOe testing with IAM team

## Testing done
Local testing

## Testing planned
Testing in staging 

Temporarily sets `loa_highest` to `loa_current` to facilitate SSOe testing with IAM team


patrickvinograd:
This is failing a test case.

johnpaulashenfelter:
@bshyong is this solely for testing (as in will never get merged)? Would love to see a blocked tag on it (and would prefer to see it as a draft PR, but github doesn't give you a way to convert back to that).

patrickvinograd:
The overall SSOe integration feature is exposed in dev only. But it needs to get merged to master to unblock additional testing, and thus is not a draft PR. 

This logic is temporary because we are awaiting some additional attributes to  be mapped into the SAML  payload from the SSOe team, at which point we will adjust the logic to account for the additional incoming data.

johnpaulashenfelter:
@bshyong @patrickvinograd oh! this is for the work in the v1 sessions controller?

bshyong:
@johnpaulashenfelter the V1 sessions controller work relates to Okta which is a separate stream of work. This PR touches code that will only get triggered when testing our specific login flow

patrickvinograd:
@johnpaulashenfelter Yes. The ssoe user attributes class is only used in the v1 session controller code  path.

PR #3332, opened 2019-09-24
WiP - Add danger-ruby to CI

omgitsbillryan:
## Description of change
Adds danger-ruby to CI.

Update, connects: https://github.com/department-of-veterans-affairs/va.gov-team/issues/1934

## Testing done
Jenkins

#### Unique to this PR
Adds danger-ruby to CI

#### Applies to all PRs

- [x] Appropriate logging


johnpaulashenfelter:
@omgitsbillryan I am a fan of `danger` but I'm also a little worried about the combination of `danger-ruby` and `overcommit` and things on both Jenkins and code climate. Lots to keep in sync. Is the plan to replace overcommit? We've also got some rake tasks hanging around for linting and gosh knows what else.

I would love to see some things get *removed/replaced* as part of adding this. I think CI _is_ the right place for most of this, though it all should be runnable local as well so devs can fix it before CI.

I think configuration sync is going to be pretty interesting. We already have code climate needing to track the local rubocop so versions match. Odds are we need that kind of sync with other lint tools?

TL;DR Love the tool, but would love to _remove_ or *consolidate* the plethora of tools we have to make things simpler.

omgitsbillryan:
Agree completely - its a lint tool cacophony!

I updated to the OP with the originating [gh issue](https://github.com/department-of-veterans-affairs/va.gov-team/issues/1934), specifically, I want to achieve:

1. Warn developers that their PR's are too long
2. Warn developers when they create a migration at the same time as modifying code

[I asked Anna if this may be achieved with codeclimate, and the answer was no](https://dsva.slack.com/archives/C0MQ281DJ/p1568905349027700). I also tried a jenkins DSL / groovy solution for this, and it _may_ have been possible, but would have severely muddied up our `jenkinsfile`.

I don't intend on `Danger` being a linter (we have _lots_ of that already), but instead a safeguard for common PR pitfalls that are specific to `va.gov` policies (I think some of these policies are written down in markdown files somewhere....). Since there's nothing that gives us this functionality already, I don't think any of our tools can be easily replaced.... 🤔

Gemfile
```diff
@@ -35,6 +35,7 @@ gem 'carrierwave-aws'
 gem 'clam_scan'
 gem 'config'
 gem 'connect_vbms', git: 'https://github.com/department-of-veterans-affairs/connect_vbms.git', branch: 'master'
+gem 'danger'
```
dhall: Could the dependency go in test/development or is it needed in all environments for the build process?

omgitsbillryan:
Closing this PR since it's really messy with commits. Started a new PR with more or less this same code here: https://github.com/department-of-veterans-affairs/vets-api/pull/3431

PR #3333, opened 2019-09-24
[WIP] #1173 Encrypted PDF Validator

camerontesterman:
#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging

### DIFF
--- a/lib/evss/documents_service.rb
+++ b/lib/evss/documents_service.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'evss/base_service'
+require 'pdf-reader'
 
 module EVSS
   class DocumentsService < BaseService
@@ -10,6 +11,7 @@ module EVSS
     DEFAULT_TIMEOUT = 180 # seconds
 
     def upload(file_body, document_data)
+      verify_readability_of file_body
       headers = { 'Content-Type' => 'application/octet-stream' }
       post 'queuedDocumentUploadService/ajaxUploadFile', file_body, headers do |req|
         req.params['systemName'] = SYSTEM_NAME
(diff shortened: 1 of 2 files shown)


lib/evss/documents_service.rb
```diff
@@ -10,6 +11,7 @@ class DocumentsService < BaseService
     DEFAULT_TIMEOUT = 180 # seconds
 
     def upload(file_body, document_data)
+      verify_readability_of file_body
```
johnpaulashenfelter: I like the method name!

lib/evss/documents_service.rb
```diff
@@ -10,6 +11,7 @@ class DocumentsService < BaseService
...
       headers = { 'Content-Type' => 'application/octet-stream' }
```
johnpaulashenfelter: Why is this `application/octet-stream` instead of `application/pdf`?

lib/evss/documents_service.rb
```diff
@@ -26,5 +28,12 @@ def upload(file_body, document_data)
     def self.breakers_service
       BaseService.create_breakers_service(name: 'EVSS/Documents', url: BASE_URL)
     end
+
+    private
+
+    verify_readability_of(file_body)
+      file = PDF::Reader.new(file_body)
```
johnpaulashenfelter: I don't know much about the performance of PDF::Reader but wondering if checking *every* doc is the right choice -- if the header says "application/pdf" can you trust it and then check?

lib/evss/documents_service.rb
johnpaulashenfelter: Ah -- because its encrypted?

PR #3338, opened 2019-09-25
Move the "new" method call to the end of download chain

edmkitty:
Move the call for `new` to the end of chain of methods that download and update facility objects. It needs to be moved because we want an accurate fingerprint on the attributes of the object.

### DIFF
--- a/lib/facilities/vha_facility.rb
+++ b/lib/facilities/vha_facility.rb
@@ -6,11 +6,15 @@ module Facilities
       attr_writer :validate_on_load
 
       def pull_source_data
+        get_all_the_facilities_data.map(&method(:new))
+      end
+
+      def get_all_the_facilities_data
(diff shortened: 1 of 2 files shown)


lib/facilities/vha_facility.rb
```diff
@@ -6,30 +6,32 @@ class << self
...
+        resp = Facilities::GisClient.new.get_all_facilities(gis_type, sort_field, max_record_count)
         resp_with_websites = add_websites(resp)
         add_mental_health(resp_with_websites)
       end
 
       def add_websites(facilities)
         service = Facilities::WebsiteUrlService.new
         facilities.map do |fac|
-          fac.website = service.find_for_station(fac.unique_id, fac.facility_type)
+          fac['website'] = service.find_for_station(fac['unique_id'], sti_name)
```
nfasulo: What is `sti_name`?

PR #3339, opened 2019-09-25
VHA Facilities datasource switchover

whuang85:
## Description of change
This closes: https://github.com/department-of-veterans-affairs/vets-contrib/issues/2373

This PR is the culmination of work done for the epic Move to new VHA data source (department-of-veterans-affairs/vets-contrib/issues/2373). It moves the source for VHA Facilities to a GIS server that is internal to the VA network, along with separate sources for websites, mental health phone numbers, and facilities with dental services. It also moves the source for facilities with primary care and mental health care services to Access to Care.

## Testing done
Comparison of original and new data sources
Comparison of `facilities/all` geoJSON response from master and this branch

## Testing planned
Once merged we will run comparisons for each environment to verify that deployment was successful

#### Unique to this PR
- [x] New datasources are being used for VHA facilities
- [x] Structure of facilities responses continues to match OAS contract

#### Applies to all PRs

- [x] Appropriate logging


app/models/base_facility.rb
```diff
@@ -11,12 +11,6 @@ class BaseFacility < ApplicationRecord
 
   YES = 'YES'
 
-  APPROVED_SERVICES = %w[
-    MentalHealthCare
-    PrimaryCare
-    DentalServices
-  ].freeze
-
   HOURS_STANDARD_MAP = DateTime::DAYNAMES.each_with_object({}) { |d, h| h[d] = d }
```
johnpaulashenfelter: Since this is a constant and unlikely to change, I'd expect to see it written out so its clear what the constant is, instead of having to mentally (or literally) run the code. This is *not* data that's going to change so generating it seems like extra complexity.

app/workers/facilities/dental_service_reload_job.rb
```diff
@@ -0,0 +1,63 @@
...
+                  local_updated: Time.now.utc.iso8601 }
+        obj = model.find(k)
+        if obj
+          obj.update(attrs)
+        else
+          model.create(attrs)
+        end
+      end
+
+      logger.info "Updated facility dental service cache for #{facilities.size} facilities"
```
johnpaulashenfelter: Are these log messages necessary in production?

app/workers/facilities/mental_health_reload_job.rb
```diff
@@ -0,0 +1,66 @@
...
+    def fetch_mental_health_data
+      mental_file = Rails.root.join('lib', 'facilities', 'mental_health_data', 'mental_health_phone_numbers.csv')
+      CSV.read(mental_file, headers: true)
+    rescue => e
+      raise MentalHealthDownloadError, "Failed to download mental health data: #{e.cause}"
+    end
+
+    def update_cache(model, records)
+      records.each do |r|
+        ext = if r['Extension'] == '0' || r['Extension'] == 'NULL'
```
johnpaulashenfelter: Extracting this into a private(?) method would let you name it for better understanding -- I don't know a ton of context here but assume its something like `filter_records_with_bad_extensions` -- and would address the code climate issue as well. 

Without checking the original PR, maybe you could reject all of the bad ones in one method and then loop over the remaining ones unless you already combined that due to other suggestions 😄

docs/setup/facilities_locator.md
```diff
@@ -4,3 +4,31 @@ For the current maps.va.gov endpoint, you will need to add the VA internal root
...
 `<HOMEBREW_DIR>/etc/openssl/cert.pem`.
+
+#### SQL52
+
+When running facilities sidekiq jobs, mental health phone numbers depend on
+a database called SQL52 that is behind the VA network. To get around this
+locally you can run your own dockerized SQL52 container locally. The following
+instructions will help you run SQL52 locally:
+
+1. `docker pull dsva/vets-api-sql52-stubbed:latest`
```
johnpaulashenfelter: Thanks for making this optional instead of part of the build. I recognize that change 👍

app/models/base_facility.rb
annaswims: HOURS_STANDARD_MAP evaluates to `{"Sunday"=>"Sunday", "Monday"=>"Monday", "Tuesday"=>"Tuesday", "Wednesday"=>"Wednesday", "Thursday"=>"Thursday", "Friday"=>"Friday", "Saturday"=>"Saturday"}`, which seems off to me.

lib/facilities/gis_configuration.rb
```diff
@@ -0,0 +1,36 @@
...
+require 'common/client/middleware/response/facility_parser'
+
+module Facilities
+  class GisConfiguration < Common::Client::Configuration::REST
+    def base_path
+      Settings.locators.gis_base_path
+    end
+
+    def service_name
+      'FL'
```
johnpaulashenfelter: Seems a little strain this isn't a constant, but not a huge issue.

lib/facilities/gis_metadata_configuration.rb
```diff
@@ -0,0 +1,34 @@
...
+    def service_name
+      'FL'
+    end
+
+    def connection
+      Faraday.new(base_path, headers: base_request_headers, request: request_options) do |conn|
+        conn.use :breakers
+        conn.request :json
+
+        # Uncomment this if you want curl command equivalent or response output to log
```
johnpaulashenfelter: Seems like all of this should be removed instead of commented out. Or toggled?

modules/va_facilities/spec/requests/facilities_request_spec.rb
```diff
@@ -370,4 +370,68 @@ def query_params(url)
       expect(response).to have_http_status(:unprocessable_entity)
     end
   end
+
+  context 'mobile flag' do
+    it 'responds with a boolean mobile flag for VHA facilities' do
```
johnpaulashenfelter: Minor, but the rest of the specs show a clearer spacing between setup/exercise/verify steps (next one is a good example)

spec/lib/facilities/vha_facility_spec.rb
```diff
@@ -1,89 +1,171 @@
...
-              expect(facility.services['last_updated']).to eq('2018-02-09')
-              expect(facility.services['health'].size).to eq(2)
-              expect(facility.services['health'].first.keys).to eq(%w[sl1 sl2])
-              expect(facility.services['health'].first.values).to eq([['MentalHealthCare'], []])
-              expect(facility.services['health'].second.keys).to eq(%w[sl1 sl2])
-              expect(facility.services['health'].second.values).to eq([['PrimaryCare'], []])
-              expect(facility.services['other']).to be_empty
-            end
+        let(:wait_time_data) do
+          fixture_file_name = "#{::Rails.root}/spec/fixtures/facility_access/wait_time_data.json"
```
johnpaulashenfelter: Did you really need `::Rails.root` here? That is surprising.

spec/lib/facilities/vha_facility_spec.rb
```diff
@@ -1,89 +1,171 @@
...
           end
         end
+      end
 
-        it 'should include zip +4 when available' do
-          VCR.use_cassette('facilities/va/vha_facilities_limit_results') do
-            expect(facility_2.address['physical']['zip']).to eq('04330-6796')
+      context 'services' do
+        let(:satisfaction_data) do
+          fixture_file_name = "#{::Rails.root}/spec/fixtures/facility_access/satisfaction_data.json"
```
johnpaulashenfelter: Maybe `Rails.root.join`?

spec/lib/facilities/vha_facility_spec.rb
johnpaulashenfelter: Ditto -- `Rails.root.join` work here instead?

spec/support/vcr_cassettes/facilities/va/vha_facilities_limit_results.yml
```diff
@@ -2,7 +2,7 @@
 http_interactions:
```
johnpaulashenfelter: How were you able to rerecord these?!? So happy!

app/workers/facilities/dental_service_reload_job.rb
edmkitty: Nah. I'll take them out

modules/va_facilities/README.v1.yml
```diff
@@ -37,6 +37,18 @@ info:
     - The list of available services in the `services` element is only applicable to VA health and
       benefits facilities.
 
+    ### Mobile Facilities
```
kfrz: Why is there a README.yml and a README.v1.yml? Do we need both?

modules/va_facilities/spec/requests/facilities_request_spec.rb
kfrz: +1, readability is crucial

modules/va_facilities/spec/requests/facilities_request_spec.rb
```diff
@@ -370,4 +370,68 @@ def query_params(url)
...
+      vba = json['data'][1]
+      vc = json['data'][2]
+
+      expect(nca['mobile']).to be_nil
+      expect(vba['mobile']).to be_nil
+      expect(vc['mobile']).to be_nil
+    end
+  end
+
+  context 'active_status flag' do
```
kfrz: When describing a context, start its description with "when" or "with".

```suggestion
  context 'with active_status flag' do
```

modules/va_facilities/spec/requests/facilities_request_spec.rb
```diff
@@ -370,4 +370,68 @@ def query_params(url)
       expect(response).to have_http_status(:unprocessable_entity)
     end
   end
+
+  context 'mobile flag' do
```
kfrz: When describing a context, start its description with "when" or "with".

```suggestion
  context 'with mobile flag' do
```

modules/va_facilities/spec/requests/facilities_request_spec.rb
```diff
@@ -370,4 +370,68 @@ def query_params(url)
...
+    it 'responds with a boolean mobile flag for VHA facilities' do
+      create :vha_648A4
+      get base_query_path + '/vha_648A4', params: nil, headers: accept_json
+      expect(response).to be_successful
+      expect(response.body).to be_a(String)
+      json = JSON.parse(response.body)
+      expect(json['data']['attributes']['mobile']).to eq(false)
+      expect(json['data']['attributes']['mobile']).to_not be_nil
```
kfrz: If the expectation on the preceding line passes, why would this line ever fail?

spec/jobs/facilities/dental_service_job_reload_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    allow_any_instance_of(
+      Facilities::DentalServiceReloadJob
+    ).to receive(:fetch_dental_service_data).and_return(CSV.parse(dental_service_data, headers: true))
+
+    Facilities::DentalServiceReloadJob.new.perform
+    expect(FacilityDentalService.find('436GH')).to be_nil
+  end
+
+  context 'on error' do
```
kfrz: When describing a context, start its description with "when" or "with".

```suggestion
  context 'when encountering an error' do
```

spec/jobs/facilities/dental_service_job_reload_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    before do
```
kfrz: I think you can do something like this:

```rb
require 'lib/sentry_logging_spec_helper'

before do
  let(:exception) { Facilities::DentalServiceError }
end

it_behaves_like 'a sentry logger'
```

Though the shared example might not be robust enough to just drop in here, yet

spec/jobs/facilities/mental_health_reload_job_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+    ).to receive(:fetch_mental_health_data).and_return(CSV::Table.new(mental_health_data))
+    Facilities::MentalHealthReloadJob.new.perform
+
+    expect(FacilityMentalHealth.find('101A')).to_not be_nil
+    expect(FacilityMentalHealth.find('202A')).to_not be_nil
+    expect(FacilityMentalHealth.find('101A').mh_ext).to eq(nil)
+    expect(FacilityMentalHealth.find('202A').mh_ext).to eq(nil)
+  end
+
+  context 'on error' do
```
kfrz: When describing a context, start its description with "when" or "with".

```suggestion
  context 'when encountering an error' do
```

spec/lib/facilities/vha_facility_spec.rb
```diff
@@ -1,89 +1,171 @@
...
+      end
+
+      it 'should get the correct classification name' do
+        VCR.use_cassette('facilities/va/vha_facilities_limit_results') do
+          expect(facility.classification).to eq('Other Outpatient Services (OOS)')
+          expect(facility_2.classification).to eq('VA Medical Center (VAMC)')
+        end
+      end
+
+      context 'mental health data' do
```
kfrz: When describing a context, start its description with "when" or "with".

```suggestion
  context 'with mental health data' do
```

spec/lib/facilities/vha_facility_spec.rb
```diff
@@ -1,89 +1,171 @@
...
+            expect(facility.phone['mental_health_clinic']).to eq('407-123-1234')
+            expect(facility_2.phone['mental_health_clinic']).to eq('321-987-6543 x 0002')
           end
         end
+      end
 
-        it 'should include zip +4 when available' do
-          VCR.use_cassette('facilities/va/vha_facilities_limit_results') do
-            expect(facility_2.address['physical']['zip']).to eq('04330-6796')
+      context 'services' do
```
kfrz: When describing a context, start its description with "when" or "with".

```suggestion
  context 'with services' do
```

config/settings.yml
```diff
@@ -390,3 +392,14 @@ bing:
 flipper:
   username: SUPER_SECRET_USERNAME
   password: SUPER_SECRET_PASSWORD
+
+oit_lighthouse2:
+  sql52:
+    hostname: vhacdwdwhsql52.vha.med.va.gov
```
edmkitty: Should we take this out? They're not actually being used, and I don't think they will be

config/settings/development.yml
```diff
@@ -15,3 +15,6 @@ betamocks:
 
 account:
   enabled: true
+
+locators:
+  mock_gis: true
```
edmkitty: Should this be here?

docs/setup/facilities_locator.md
```diff
@@ -4,3 +4,31 @@ For the current maps.va.gov endpoint, you will need to add the VA internal root
...
+1. `docker run -e ACCEPT_EULA=Y -e SA_PASSWORD=password1! -p 1433:1433 -d dsva/vets-api-sql52-stubbed:latest`
+1. In `config/settings.local.yml`, you will also need to set these variables
+
```
edmkitty: We may need to overhaul this, considering the code doesn't use a SQL52 client at all anymore

app/workers/facilities/mental_health_reload_job.rb
```diff
@@ -0,0 +1,66 @@
...
+    def remove_invalid(model, record_keys)
+      invalid = model.keys - record_keys
+      invalid.each { |x| model.delete(x) }
+      logger.info "Removed #{invalid.size} obsolete entries from cache"
+    end
+
+    def update_mental_health_data
+      records = fetch_mental_health_data
+      update_cache(FacilityMentalHealth, records)
+      remove_invalid(FacilityMentalHealth, records.map { |r| r['StationNumber'] })
```
jholton: The method definition isn't shared with another class, so you can just reference `FacilityMentalHealth` inside the methods instead of passing it as a parameter.

app/workers/facilities/dental_service_reload_job.rb
```diff
@@ -0,0 +1,63 @@
...
+    def parse_dental_service_data(records)
+      records.map { |r| r['unique_id'] }
+    end
+
+    def update_dental_service_data
+      dental_file = Rails.root.join('lib', 'facilities', 'dental_service_data', 'dental_services.csv')
+      records = fetch_dental_service_data(dental_file)
+      facilities = parse_dental_service_data(records)
+      update_cache(FacilityDentalService, facilities)
+      invalidate_removed(FacilityDentalService, facilities)
```
jholton: The method definition isn't shared with another class, so you can just reference `FacilityDentalService` inside the methods instead of passing it as a parameter.

app/workers/facilities/dental_service_reload_job.rb
```diff
@@ -0,0 +1,63 @@
...
+    def update_cache(model, facilities)
+      facilities.each do |k|
+        attrs = { station_number: k,
+                  local_updated: Time.now.utc.iso8601 }
+        obj = model.find(k)
+        if obj
+          obj.update(attrs)
+        else
+          model.create(attrs)
+        end
```
jholton: Looking at `Common::RedisStore`, I think this (lines 28 - 33) can be reduced to
```
obj = model.find_or_build(k)
obj.update(attrs)
```

app/workers/facilities/mental_health_reload_job.rb
```diff
@@ -0,0 +1,66 @@
...
+          local_updated: Time.now.utc.iso8601
+        }
+
+        obj = model.find(r['StationNumber'])
+
+        if obj
+          obj.update(attrs)
+        else
+          model.create(attrs)
+        end
```
jholton: Similar to my other comment, looking at `Common::RedisStore`, I think this (lines 38 - 44) can be reduced to
```
obj = model.find_or_build(r['StationNumber'])
obj.update(attrs)
```

lib/common/client/middleware/response/facility_parser.rb
```diff
@@ -14,130 +14,164 @@ def on_complete(env)
...
-          def services_from_wait_time_data(facility_id)
-            facility_wait_time = FacilityWaitTime.find(facility_id)
-            metric_keys = facility_wait_time&.metrics&.keys || []
-            services = []
-            services << { 'sl1' => ['EmergencyCare'], 'sl2' => [] } if facility_wait_time&.emergency_care&.any?
-            services << { 'sl1' => ['UrgentCare'], 'sl2' => [] } if facility_wait_time&.urgent_care&.any?
-            (Facilities::AccessDataDownload::WT_KEY_MAP.values - %w[primary_care mental_health]).each do |service|
-              services << { 'sl1' => [service.camelize], 'sl2' => [] } if metric_keys.include?(service)
+            def strip(value)
+              value.respond_to?(:strip) ? value.strip : value
```
jholton: Nitpicky, but I think this (and the few other similar conditionals (e.g. line 128 above)) is a good use case for `try`.
Thoughts?
```suggestion
              value.try(:strip) || value
```

lib/facilities/gis_metadata_configuration.rb
edmkitty: Could you explain "toggle"? Do you mean the activerecord method?

spec/support/vcr_cassettes/facilities/va/vha_facilities_limit_results.yml
edmkitty: @rtravitz experienced much pain

app/models/base_facility.rb
```diff
@@ -119,7 +113,7 @@ def suggested(facility_types, name_part)
   PATHMAP = { 'NCA_Facilities' => Facilities::NCAFacility,
               'VBA_Facilities' => Facilities::VBAFacility,
               'VHA_VetCenters' => Facilities::VCFacility,
-              'VHA_Facilities' => Facilities::VHAFacility }.freeze
+              'FacilitySitePoint_VHA' => Facilities::VHAFacility }.freeze
```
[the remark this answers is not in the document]
edmkitty: It is not up to us. This is part of the GIS url.

modules/va_facilities/spec/requests/facilities_request_spec.rb
edmkitty: This is a good point. I will reverse the expectations

modules/va_facilities/README.v1.yml
rtravitz: We do need both. They provide the content for different versions of the API documentation that you can select in the dropdown here: https://developer.va.gov/explore/facilities/docs/facilities

lib/common/client/middleware/response/facility_parser.rb
edmkitty: `try` is not appropriate here. Depending on the type of object (for example, a Hash), `value` may or may not respond to `try`.

dogweather: Interesting - shouldn't every object respond to `#try`?

app/workers/facilities/dental_service_reload_job.rb
dogweather: Maybe leave in but change to `debug`?

app/models/base_facility.rb
rtravitz: It's odd for sure—but accurate. That same map is used across facility types, and this is the most common format. So certain types override it locally, but we can let this base transformation run on the other types without needing to add a bunch of conditionals to our parser. 

Also, changing the code that evaluates the map into a frozen hash!

johnpaulashenfelter: @annaswims if you search the code, it gets relooped over and downcased in one of the three places that uses it https://github.com/department-of-veterans-affairs/vets-api/blob/eeb0b2b9ee14ab377b251de45033693a2f1e7483/lib/facilities/vc_facility.rb#L35

app/workers/facilities/dental_service_reload_job.rb
johnpaulashenfelter: Looks like production level is `info` based on `production.rb` though I'm not sure what else happens in Jenkins. In any case, debug works ok if its necessary

lib/facilities/gis_client.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+require 'common/client/base'
+
+module Facilities
+  # Core class responsible for api interface operations
+  class GisClient < Common::Client::Base
+    configuration Facilities::GisConfiguration
+
+    def get_all_facilities(facility_type, order_field, max_record_count)
```
jholton: Can we get some specs for this?
We should test edge cases for infinite loops, or unexpected results.
If `max_record_count` is records `per_page`, right?
And `query_count` is essentially a `page`, right?
Here's a scenario:
`max_record_count` is 10
`query_count` is 2 
total number of items in the external service is 20
Therefore we have all the records back but the `break` conditional has not been met
What happens when the loop runs again when `query_count` is 3?
What does that do to `data_collector`?  It's hard to picture because I don't know what `response.body` returns.

lib/facilities/gis_metadata_configuration.rb
johnpaulashenfelter: @edmkitty no -- I mean you've got commented out code here that ... might need to be used at some point? Wondering if there's either some kind of toggle value (environment, log level, etc) that would indicate this output should be injected. Or more likely this is something that was useful during dev but can now be removed

lib/facilities/gis_configuration.rb
jholton: True, but this follows the standard that is used in all of our configuration classes.

spec/jobs/facilities/dental_service_job_reload_spec.rb
edmkitty: I changed the spec to have this in another branch, but learned that those group specs will test the exception itself, but not the code calling the exception. Changing this test reduced the code coverage and made simplecov unhappy. So I am leaving it how it is for now

lib/common/client/middleware/response/facility_parser.rb
edmkitty: Yup, you're right. I had tested it in ruby and not rails. Didn't realize it was part of the rails framework

lib/facilities/vha_facility.rb
```diff
@@ -6,36 +6,62 @@ class << self
...
+        get_all_the_facilities_data.map(&method(:new))
+      end
+
+      def get_all_the_facilities_data
+        gis_type = 'FacilitySitePoint_VHA'
+        sort_field = 'Sta_No'
+        metadata = Facilities::GisMetadataClient.new.get_metadata(gis_type)
         max_record_count = metadata['maxRecordCount']
-        Facilities::Client.new.get_all_facilities(arcgis_type, sort_field, max_record_count).map(&method(:new))
+        resp = Facilities::GisClient.new.get_all_facilities(gis_type, sort_field, max_record_count)
```
jholton: Are other classes going to use the `GisClient` and `GisMetadataClient`?  
If not, you can hard code `gis_type` and `sort_field` in the `GisClient` and `GisMetadataClient` classes instead of passing them as parrameters.

edmkitty: Yes, probably.

nfasulo: Yes, in the near future we will begin using `GisClient` to pull Vet Centers.

PR #3343, opened 2019-09-26
Kr/refactor logging and constants

edmkitty:
A PR to make changes to the epic branch based on the PR for the epic branch into master. PR-ception.


app/workers/facilities/dental_service_reload_job.rb
```diff
@@ -21,25 +18,18 @@ def fetch_dental_service_data(file)
       raise DentalServiceError, "Failed to read CSV file: #{file} - caused by: #{e.cause}"
     end
 
-    def update_cache(model, facilities)
+    def update_cache(facilities)
       facilities.each do |k|
```
dogweather: Could we have a more descriptive variable for `k`? Coming to this new, I'm not sure what the intent is, or what its type is.

app/workers/facilities/dental_service_reload_job.rb
```diff
@@ -21,25 +18,18 @@ def fetch_dental_service_data(file)
...
-      logger.info "Updated facility dental service cache for #{facilities.size} facilities"
     end
 
-    def invalidate_removed(model, facility_keys)
-      invalidate = model.keys - facility_keys
-      invalidate.each { |x| model.delete(x) }
-      logger.info "Removed #{invalidate.size} obsolete entries from cache"
+    def invalidate_removed(facility_keys)
+      invalidate = FacilityDentalService.keys - facility_keys
+      invalidate.each { |x| FacilityDentalService.delete(x) }
```
dogweather: Maybe rename `invalidate` to `removed_keys`? (If that's actually what this code means - I can't quite tell from the names.) Or, if it mean invalid keys, then could we name it that? I.e., the plural noun, `invalid_keys`

dogweather: Ah hah, from the deleted debug message (!) I'm inferring that this is doing the opposite of what the function name implied to me: `invalidate_removed(facility_keys)`. I.e., I read that to mean, "Invalidate the removed facility keys". But it's actually, "Remove the invalid facility keys" (?) If that's true, I suggest we rename the function as well.

app/workers/facilities/mental_health_reload_job.rb
```diff
@@ -35,26 +31,24 @@ def update_cache(model, records)
...
         }
 
-        obj = model.find(r['StationNumber'])
-
-        if obj
-          obj.update(attrs)
-        else
-          model.create(attrs)
-        end
+        obj = FacilityMentalHealth.find_or_build(r['StationNumber'])
```
dogweather: Do we know anything more specific about the kinds of `obj` to give it a more meaningful name? E.g., just guessing, `station`.

app/workers/facilities/mental_health_reload_job.rb
```diff
@@ -19,13 +19,9 @@ def fetch_mental_health_data
...
 
-    def update_cache(model, records)
+    def update_cache(records)
       records.each do |r|
-        ext = if r['Extension'] == '0' || r['Extension'] == 'NULL'
-                nil
-              else
-                r['Extension']
-              end
+        ext = valid_extension?(r['Extension']) ? r['Extension'] : nil
```
johnpaulashenfelter: Ok, now I think this is harder to understand.

I thought you were excluding all the records where the extension was '0' or 'NULL' but you're doing a *default* of `nil`. Its a lot harder to read like this for sure. How about something like

```
        attrs = {
          station_number: r['StationNumber'],
          mh_phone: r['MHPhone'],
          mh_ext: ensure_valid_extension(r['Extension']),
          modified: r['Modified'],
          local_updated: Time.now.utc.iso8601
        }

....

def ensure_valid_extension(extension:)
  if extension == '0' || extension == 'NULL'
                nil
      else
               extension
      end
end
```

dogweather: Interesting. Thinking deeper about it, I think the code's true function is converting from one encoding to another. That is, encoding how to denote "no extension extension":

* In the source datastore, "none" is represented by the strings `"0"` or `"NULL"`.
* In the destination datastore, "none" is represented by the object `nil`.

And so I'd vote for making this explicit:

```ruby
EXTENSION_MAPPINGS = {
  '0' => nil,
  'NULL => nil
}

# ...

def convert_extension(ext)
  return ext unless EXTENSION_MAPPINGS.key?(ext)

  EXTENSION_MAPPINGS[ext]
end
```

johnpaulashenfelter: Similar idea and does the same job of moving the logic out of the way and into a method (which was the original issue w/ the CodeClimate warning).

I think the `if/else` is way more braindead (a plus) and thus easier to understand when someone reads this a year from now but the map is pretty readable too and  makes a lot of sense, esp for a more complex mapping for sure @dogweather

edmkitty: I'm gonna go with the if statement. I like the idea of confining all the logic to one method

PR #3344, opened 2019-09-26
Remove sql52 docker info

whuang85:
## Description of change
Removes SQL52 docker information as that is not needed anymore.

### DIFF
--- a/docs/setup/facilities_locator.md
+++ b/docs/setup/facilities_locator.md
@@ -7,6 +7,7 @@ done by appending the exported/downloaded certificate to
 
 #### Retrieving Mental Health Phone Numbers
 
-To retrieve mental health phone numbers, please run this rake job:
+To retrieve mental health phone numbers, run the `jobs:pull_facility_mental_health_phone` rake task
+E.g.,
 
 `bundle exec rake jobs:pull_facility_mental_health_phone`
--- a/docs/setup/facilities_locator.md
+++ b/docs/setup/facilities_locator.md
@@ -5,30 +5,8 @@ CA certificate to your trusted certificates. With `homebrew` this is typically
 done by appending the exported/downloaded certificate to
 `<HOMEBREW_DIR>/etc/openssl/cert.pem`.
 
-#### SQL52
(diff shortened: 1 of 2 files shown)


docs/setup/facilities_locator.md
```diff
@@ -5,30 +5,8 @@ CA certificate to your trusted certificates. With `homebrew` this is typically
...
-      database_name: SQL52
-      schema_name: A01
-      table_name: OMHSP_PERC_Share__DOEx__FieldDataEntry_MHPhone
-```
-
-If you would like to know more about how this image was put together, please
-go here: https://github.com/department-of-veterans-affairs/vets-api-sql52-stubbed-docker
+To retrieve mental health phone numbers, please run this rake job:
 
+`bundle exec rake jobs:pull_facility_mental_health_phone`
```
dogweather: I suggest leaving out "please" because it's not quite accurate. And we can be more specific about the rake _task_:

----

To retrieve mental health phone numbers, run the `jobs:pull_facility_mental_health_phone` rake task. E.g.,

```bash
bundle exec rake jobs:pull_facility_mental_health_phone
```

spec/lib/facilities/vha_facility_spec.rb
```diff
@@ -116,14 +116,14 @@
...
-          fixture_file_name = ::Rails.root.join("/spec/fixtures/facility_access/satisfaction_data.json")
+          fixture_file_name = ::Rails.root.join('/spec/fixtures/facility_access/satisfaction_data.json')
           File.open(fixture_file_name, 'rb') do |f|
             JSON.parse(f.read)
           end
         end
 
         let(:wait_time_data) do
-          fixture_file_name = ::Rails.root.join("/spec/fixtures/facility_access/wait_time_data.json")
+          fixture_file_name = ::Rails.root.join('/spec/fixtures/facility_access/wait_time_data.json')
```
dogweather: Is it necessary to prepend the scope resolution operator? Could we get rid of the noise and use `Rails.root...` instead of `::Rails.root...`?

PR #3346, opened 2019-09-27
cbt/1137-validate_pdf.rb fix

camerontesterman:
## Description of change
[Link to originating issue](https://github.com/department-of-veterans-affairs/va.gov-team/issues/1173)

Changes introduced in this pull request fix the pdf validator that is used to validate pdfs atached as supporting evidence form attachments. Currently, if you try to attach a PDF that is encrypted, the validator will not properly flag it as encrypted. I removed a return statement, as a 500 error should be returned if the file is not an actual pdf file.

## Testing done
This has been tested locally against my own local dev environment. I tested it with three dummy PDFs, a correct sample.pdf, a jpg that's extension was changed to pdf, and an encrypted pdf. Only the sample.pdf (the valid one) was able to be uploaded, the others resulted in 500 errors.

Below are three screenshots displaying that malformed or encrypted documents are no longer accepted.

Screen Shot 2019-09-27 at 3 22 44 PM
Screen Shot 2019-09-27 at 3 23 04 PM
Screen Shot 2019-09-27 at 3 23 14 PM

Only acceptable PDFs are allowed to uploaded. Malformed and encrypted PDFs are denied.

#### Applies to all PRs

- [x] Appropriate logging


app/uploaders/validate_pdf.rb
```diff
@@ -11,10 +13,9 @@ def validate_pdf(file)
     temp_file = file.tempfile
     return unless temp_file.readpartial(4) == '%PDF'
 
-    PDF::Reader.new(temp_file).info
-  rescue PDF::Reader::MalformedPDFError
-    raise CarrierWave::UploadError, 'PDF is missing an end of file marker'
-  rescue PDF::Reader::EncryptedPDFError
```
SMLuthi: Is there a reason why this wasnt working to begin with?

kfrz: :heavy_plus_sign: to understanding the original failure w/ `PDF` library

Also if we're replacing the dependency w/ `origami`, can the original library be removed/replaced entirely?

app/uploaders/validate_pdf.rb
```diff
@@ -11,10 +13,9 @@ def validate_pdf(file)
...
-    raise CarrierWave::UploadError, 'PDF is encrypted'
+    pdf = Origami::PDF.read temp_file, decrypt: false
+    raise CarrierWave::UploadError, 'PDF is encrypted' if pdf.encrypted?
+  rescue Origami::InvalidPDFError
+    raise CarrierWave::UploadError, 'PDF is invalid'
```
SMLuthi: Can we get more detailed on the responses here? More specific error handling could result in easier debugging in the future.

camerontesterman: I've updated these to be a bit more descriptive.

camerontesterman:
> Left a few comments but overall the change is straightforward.
> 
> I also want to make sure that this doesnt impact any other services on the FE. It looks like a fairly straight drop in replacement... though im not familiar with `origami`. we are removing a specific error response message: `PDF is missing an end of file marker`, so I hope thats not expected anywhere in another service.
> 
> Also, providing some tests to capture encrypted pdfs would be good.

Tests have been updated.

app/uploaders/validate_pdf.rb
camerontesterman: So the cause of why it wouldn't trigger the error is still unclear to me. Testing locally, if you run `PDF::Reader.new("encrypted.pdf").info` it would trigger successfully within `irb` - so something about the upload process is causing it to not trigger. As demonstrated in my screenshots though, my fix does indeed catch the PDFs that are encrypted or malformed.

camerontesterman: @kfrz just doing some checking within the codebase, `PDF::Reader` can't be removed at this time, it's used in some other spots.

SMLuthi: Hm, I think we should look more into why PDF::Reader is catching encrypted files locally but not when put into actual practice... this is a little disconcerting. Especially if it means we wouldnt need to add, yet, another gem to do work that a pre-existing gem should already be handling. However, I wont let that keep this from being merged in.

PR #3349, opened 2019-09-27
1995s - Spool submission report conditional recipients

dmccurdybah:
## Description of change
- Updated logic for conditional recipients to support 1995s submissions. 
- Improved test coverage.
- Updated spool submission report recipient list for staging

https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/vets.gov-team/19741

## Testing done
Tested locally and QA review

## Testing planned
N/A

#### Unique to this PR
- [x] 1995s form type triggers conditional STEM recipients 

#### Applies to all PRs

- [x] Appropriate logging


spec/jobs/education_form/create_spool_submissions_report_spec.rb
```diff
@@ -64,6 +64,26 @@ def perform
           expect(File.read(filename)).to eq(csv_string)
         end
 
+        it 'should check for stem submissions' do
```
jholton: These are testing the results of `#create_csv_array`, so you can move them under the 
> describe '#create_csv_array' do

section and you can remove the `perform` method calls.  They're not necessary.

I also noticed an indentation issue that you didn't create.
`describe '#perform'` should be at the same level of indentation as `describe '#create_csv_array'`.
Currently it's nested inside it.
Can you fix that too please while you're in here?

dmccurdybah: Thanks, done!

app/workers/education_form/create_spool_submissions_report.rb
```diff
@@ -39,11 +39,12 @@ def create_csv_array
     end
 
     def check_claim_for_stem(education_benefits_claim, parsed_form, stem_exists)
-      if stem_exists || (education_benefits_claim.form_type == '1995' && parsed_form['isEdithNourseRogersScholarship'])
-        return true
-      end
+      return true if stem_exists
+
+      # Remove when 1995s form type is in prod
```
johnpaulashenfelter: Is this working like a feature flag? Notes like this are pretty fragile... can you mark it with a TODO which will show up in some of our tools?

spec/factories/education_benefits_claims.rb
```diff
@@ -7,11 +7,17 @@
     end
   end
 
-  %w[1990e 1990n].each do |form_type|
+  %w[1990e 1990n 1995s].each do |form_type|
     factory :"education_benefits_claim_#{form_type}", class: 'EducationBenefitsClaim' do
```
johnpaulashenfelter: Before this gets even more opaque, let's break these out into three factories. 

It seems weird that they all inherit from the same class but they're different factories. Are there really underlying traits here `education_benefits_claim trait: form_1990e`? Building factory *definitions* programmatically makes it harder to lint, let alone understand.

spec/jobs/education_form/create_spool_submissions_report_spec.rb
```diff
@@ -35,40 +36,57 @@
...
-          end
+    describe '#perform' do
+      let(:filename) { "tmp/spool_reports/#{time.to_date}.csv" }
+
+      before do
+        expect(FeatureFlipper).to receive(:send_edu_report_email?).once.and_return(true)
+      end
+
+      after do
+        File.delete(filename)
```
johnpaulashenfelter: You're deleting here, but the create is in a `let` -- wouldn't it make just as much sense to put it in the `before`?

Do you need this file for every example in this describe block? If not it would be a lot clearer to include it only in the example(s) that need it.

Al `before(:all)` would let you create it once, followed by an `after(:all)` to clean it up, all inside this `describe` block. That can get hairy when mixed with (:each) but in this case you're saving a lot of file creation steps too.

dmccurdybah: Both examples create a spool file, so deleting it in the `after` makes sense. The `let` here is only the name of the file.

spec/factories/education_benefits_claims.rb
dmccurdybah: Sounds good!

app/workers/education_form/create_spool_submissions_report.rb
dmccurdybah: Done!

spec/jobs/education_form/create_spool_submissions_report_spec.rb
dmccurdybah: @johnpaulashenfelter have you had a chance to look at this?  Thanks!

app/workers/education_form/create_spool_submissions_report.rb
```diff
@@ -39,11 +39,12 @@ def create_csv_array
     end
 
     def check_claim_for_stem(education_benefits_claim, parsed_form, stem_exists)
-      if stem_exists || (education_benefits_claim.form_type == '1995' && parsed_form['isEdithNourseRogersScholarship'])
-        return true
-      end
+      return true if stem_exists
+
+      # TODO: Remove when 1995s form type is in prod
```
kfrz: I'd rather see this as a ticket on your board,  `# TODO` comments are not as useful because they're not tracked

spec/jobs/education_form/create_spool_submissions_report_spec.rb
johnpaulashenfelter: 👍

app/workers/education_form/create_spool_submissions_report.rb
dmccurdybah: Thanks, we will definitely do that!

PR #3352, opened 2019-09-27
Gis client testing

whuang85:
## Description of change
Add some GIS Client testing

#### Applies to all PRs

- [x] Appropriate logging


spec/lib/facilities/gis_client_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+require 'rails_helper'
+RSpec.describe Facilities::GisClient do
+  RSpec::Matchers.define :has_order_and_offset do |order, offset|
+    match do |params|
+      params[:orderByFields].eql?(order) && params[:resultOffset].eql?(offset)
+    end
+  end
+
+  context 'params' do
```
rtravitz: This strikes me as a `describe` instead of a `context` and should maybe say `'get_all_facilities'`

spec/lib/facilities/gis_client_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    match do |params|
+      params[:orderByFields].eql?(order) && params[:resultOffset].eql?(offset)
+    end
+  end
+
+  describe 'get_all_facilities' do
+    subject { described_class.new }
+
+    it 'passes the correct offset to the query when looping twice' do
+      r1 = double(Faraday::Response.new)
```
nfasulo: I would rename these "r" variables to something more identifiable like `faraday_response_offset_0`. You can also declare each just once if you move them above of the describe block like `let(:faraday_response_offset_0)`.

spec/lib/facilities/gis_client_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      r2 = double(Faraday::Response.new)
+
+      allow_any_instance_of(Faraday::Connection)
```
nfasulo: You can declare `allow_any_instance_of(Faraday::Connection)` for each offset just once if you move them into a `before(:each)` block.

spec/lib/facilities/gis_client_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+        Faraday::Connection
+      ).to receive(:get).with(anything, has_order_and_offset('field', 10)).and_return(faraday_response_offset_10)
+
+      allow_any_instance_of(
+        Faraday::Connection
+      ).to receive(:get).with(anything, has_order_and_offset('field', 20)).and_return(faraday_response_offset_20)
+    end
+
+    it 'passes the correct offset to the query when looping twice' do
+      allow(faraday_response_offset_0).to receive(:env).and_return(double(body: [*1..12]))
```
jholton: You're passing a `max_record_count` of 10, but then mocking that the response returns 12 results.  
Is that a possibility?

whuang85: I believe so as the GIS_Client only passes an offset to the server to make a request and does not pass a maximum number of records to retrieve.  The `max_record_count` only comes into play when looking at a particular response from the GIS server and at that point is only looking at the length of the body.

edmkitty: If we are assuming the server returns "maxRecordCount" and it should be the "per page" variable, then we should also assume the first expectation would return 10 records. We can investigate the server's behavior more in depth later.

jholton:
Thanks for the unit tests.  It would be nice to see some VCR cassettes so engineers looking at this in the future have an idea what the actual responses look like.

nfasulo:
@jholton Good point about using cassettes. `facility_location_download_job_spec.rb` and `vha_facility_spec.rb` both use cassettes of responses from the GIS server. We are going to be pulling a second Vet Centers from this server in the near future and we can add tests for the GisClient that utilitize cassettes as we make that change.

jholton:
> Thanks for the unit tests. It would be nice to see some VCR cassettes so engineers looking at this in the future have an idea what the actual responses look like.

And I'm realizing that because `GisClient` is used by `VHAFacility` that there are VCR cassettes for this used in `vha_facility_spec`

PR #3356, opened 2019-09-30
SSOe loa current

bshyong:
## Description of change
Context: We had originally mapped `va_eauth_credentialassurancelevel` to `loa_current`. This mapping should be changed to `loa_highest` instead. Additionally, this PR disables the MHV SSO cookie when signing in through the SSOe flow as the cookie is not needed in that scenario.

This change is needed to unblock further testing of SSOe flow with IAM team.

## Summary:
- Refactoring and changes to support inference of loa_current attribute
- Mapping of `va_eauth_credentialassurancelevel` to `loa_highest`
- Changes to specs for test coverage
- Disable SSO cookie in SSOe flow to resolve error: http://sentry.vfs.va.gov/vets-gov/platform-api-development/issues/75532/?query=is%3Aunresolved

## Testing done
Local tests

## Testing planned
Test in staging/dev environment

- [x] Appropriate logging


lib/saml/user.rb
```diff
@@ -81,27 +81,22 @@ def authn_context
       raise
     end
 
-    # For future use; can check if issuer matches /eauth\.va\.gov/ for SSOe
     def issuer
-      saml_response.issuers[0]
-    rescue
-      Raven.tags_context(controller_name: 'sessions', sign_in_method: 'not-signed-in:error')
-      raise
+      saml_response.issuer_text
```
patrickvinograd: I'm a little concerned that we dropped the `rescue` which presumably was added for a reason.

lib/saml/user.rb
```diff
@@ -81,27 +81,22 @@ def authn_context
...
     end
 
-    # TODO: validate that the AuthN attribute name for SSOe case is 'ssoe'
     # SSOe Issuer value is https://int.eauth.va.gov/FIM/sps/saml20fedCSP/saml20
     # SSOe AuthnContext currently set to urn:oasis:names:tc:SAML:2.0:ac:classes:Password
     def user_attributes_class
+      return SAML::UserAttributes::SSOe if issuer&.match(/eauth\.va\.gov/)
```
patrickvinograd: 👍

lib/saml/user.rb
patrickvinograd: Oh, I see this is a new method and the rescue was copied from the authn_context method...I guess authn_context is checked early during parsing but would not be present on a SAML 'login failed' response. But I don't know if issuer would/would not be present or if we'd have the same issue if we read issuer on a long failed event.

patrickvinograd: I guess worst case `issuer` would be nil and the `user_attribute_class` method would fall through.

bshyong: I'll add this back—I think the original intention behind this for error tracking within Sentry, since we re-raise the exception within the rescue

lib/saml/user.rb
```diff
@@ -21,7 +21,7 @@ class User
       'dslogon_loa3' => { loa_current: '3', sign_in: { service_name: 'dslogon' } },
       'myhealthevet' => { loa_current: nil, sign_in: { service_name: 'myhealthevet' } },
       # SSOe context
-      'urn:oasis:names:tc:SAML:2.0:ac:classes:Password' => { loa_current: nil, sign_in: { service_name: 'ssoe' } },
+      'urn:oasis:names:tc:SAML:2.0:ac:classes:Password' => { loa_current: '1', sign_in: { service_name: 'ssoe' } },
```
johnpaulashenfelter: This seems like a pretty significant change, but i think this whole section (SSOe) is new so guessing the SAML responses are in flux?

bshyong: The SAML responses are relatively set, but we are debugging and making adjustments throughout the testing process to get unblocked. In particular, they currently do not pass a custom AuthnContext, and only pass `urn:oasis:names:tc:SAML:2.0:ac:classes:Password`.  There are active development plans on their end to pass through the custom AuthNContext.

lib/saml/user_attributes/ssoe.rb
```diff
@@ -63,29 +63,47 @@ def dslogon_edipi
...
+      end
 
-      # TODO: This should be the ID.me highest level of assurance attained;
-      # VA IAM team to get this integrated and propagated from ID.me
-      # double check attribute name after VA IAM finalizes
-      # TODO: reset this attribute to `level_of_assurance` after IAM adds this
+      # This is the ID.me highest level of assurance attained
       def loa_highest
-        loa_current
+        attributes['va_eauth_credentialassurancelevel']&.to_i
```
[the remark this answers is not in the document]
bshyong: That's right

PR #3357, opened 2019-09-30
Upgrade brakeman gem

johnpaulashenfelter:
When the brakeman gem was updated, there was one old/obsolete brakeman.ignore entry that was removed

There is a new brakeman issue that is now in the ignore file. Basically, its encouraging us to move from the older Rails <~ 4.2 default of storing cookies as marshalled objects to using JSON.

This led me down a small rabbit hole into looking at the new options for encrypting Rails cookies introduced in Rails ~> 5.2 that might replace our custom cookie encryptor and further simplify the app. https://blog.bigbinary.com/2018/06/26/rails-5-2-uses-aes-256-gcm-authenticated-encryption-as-default-cipher-for-encrypting-messages.html

## Description of change

Gem was updated, ignore file was updated to match, allowing any existing issues that can't be clearly remediated to be skipped.

## Testing done

Brakeman runs without additional warnings
## Testing planned

Brakeman runs without additional warnings on staging

#### Applies to all PRs

- [x] Appropriate logging

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -184,7 +184,7 @@ GEM
       thread_safe (~> 0.3, >= 0.3.1)
     benchmark-ips (2.7.2)
     bindex (0.8.1)
-    brakeman (4.4.0)
+    brakeman (4.6.1)
     breakers (0.2.4)
       faraday (>= 0.7.4, < 0.10)
       multi_json (~> 1.0)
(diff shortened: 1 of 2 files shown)


johnpaulashenfelter:
I also learned today that the defaults for `brakeman` from the CLI on a local machine and the params used on CI are different `brakeman --run-all-checks`. Apparently the default command doesn't run everything.

PR #3364, opened 2019-10-02
Add detail to User Profile swagger and remove redundant 'autn_context' attr

annaswims:
## Description of change

Prompted by https://github.com/department-of-veterans-affairs/va.gov-team/issues/1798, update documentation related to user profile.  Also, `authn_context` is the same value as `sign_in.service_name` and should be removed.
## Testing done
specs

#### Applies to all PRs

- [x] Appropriate logging


app/swagger/schemas/user_internal_services.rb
```diff
@@ -33,18 +33,34 @@ class UserInternalServices
...
-              property :last_name, type: :string
-              property :birth_date, type: :string
-              property :gender, type: :string
-              property :zip, type: :string
-              property :last_signed_in, type: :string
+              property :first_name, type: :string, example: 'Abigail'
+              property :middle_name, type: :string, example: 'Jane'
+              property :last_name, type: :string, example: 'Brown'
+              property :birth_date, type: :string, example: '1900-01-01'
+              property :gender, type: :string, example: 'F'
```
johnpaulashenfelter: Thanks for the examples! 👍

spec/request/swagger_spec.rb
```diff
@@ -26,7 +26,7 @@
   subject { Apivore::SwaggerChecker.instance_for('/v0/apidocs.json') }
 
   let(:rubysaml_settings) { build(:rubysaml_settings) }
-  let(:mhv_user) { build(:user, :mhv) }
+  let(:mhv_user) { build(:user, :mhv, middle_name: 'Bob') }
```
johnpaulashenfelter: I'm not completely clear why this needed to be added in this PR -- seems unexpected so therefore 🤔 

Did something change in the factory?

johnpaulashenfelter: Answered my own question -- there's a new field (middle name) in the swagger docs above that I missed. Carry on!

@annaswims Would it make sense to put middle name in the factory itself?

Feels like putting it in the `let` here makes it something I should care about looking at for the spec here (why is middle_name set? did I miss something? why is it the special attribute compared to everything else). 

Of course adding it to the factory might have other ramifications?

app/services/users/profile.rb
```diff
@@ -71,9 +71,7 @@ def profile
         loa: user.loa,
         multifactor: user.multifactor,
         verified: user.loa3?,
-        sign_in: user.identity.sign_in,
-        # FIXME: this wont be necessary after FE makes appropriate changes
```
johnpaulashenfelter: Yay having the FIXME *and* removing it by fixing it!

spec/request/swagger_spec.rb
annaswims: Honestly, I'd consider it a bit of a cheat. I documented the `middle_name` attribute, but the test was failing because `middle_name` was `null` when I said it was a string.  Perhaps we want to accept `null` as an acceptable value in these tests when the docs say it's a string.

I was afraid if I changed the factories to have middle names that I'd fail all sorts of other tests.

app/swagger/schemas/user_internal_services.rb
```diff
@@ -33,18 +33,34 @@ class UserInternalServices
...
-                property :service_name, type: :string
+                property :service_name,
+                         type: :string,
+                         enum: %w[myhealthevet dslogon idme],
+                         example: 'myhealthevet',
+                         description: ''
+                property :account_type,
+                         enum: %w[Basic Premium],
+                         example: 'Basic',
+                         description: 'myhealthevet account_types: Basic, Premium. dslogon account account_types: 1-3'
```
U-DON: I forget now, but does this mean `account_type` can have values of 1-3 for `service_name: dslogon`? Should those values be in the enum as well?

annaswims:
@johnpaulashenfelter Can you give this one an approval?  @U-DON approved, but he's not a BE codeowner.

PR #3366, opened 2019-10-03
PPIU error logging

lihanli:
## Description of change
https://github.com/department-of-veterans-affairs/va.gov-team/issues/2074#issuecomment-537355433
log requests and responses to the ppiu api as requested by the EVSS team

## Testing done
automated tests
tested in staging

#### Applies to all PRs

- [x] Appropriate logging


lib/evss/ppiu/service.rb
```diff
@@ -28,18 +28,30 @@ def get_payment_information
...
         with_monitoring_and_error_handling do
-          raw_response = perform(:post, 'paymentInformation', request_body(pay_info), headers)
+          raw_response = perform(:post, 'paymentInformation', body, headers)
           PaymentInformationResponse.new(raw_response.status, raw_response)
         end
       end
 
       private
 
+      def save_req_body(req_body)
```
jholton: Something like `sanitize_req_body` seems like a more accurate name

lib/evss/logged_service_exception.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module EVSS
+  class LoggedServiceException < EVSS::ServiceException
+    def initialize(res_body, user, req_body = nil)
+      @res_body = res_body
+      @user = user
+      @req_body = req_body
+
+      log_exception
```
annaswims: I think I'd prefer seeing `res_body, user, req_body` passed to log_exception as arguments rather than instance variables.  At first I thought they might be used elsewhere because they were instance variables.

PR #3371, opened 2019-10-03
Update README.v1.yml

edmkitty:
DO NOT MERGE UNTIL THIS TICKET IS COMPLETE: https://github.com/department-of-veterans-affairs/vets-contrib/issues/2954

Changes in this PR:
- Updates the /nearby endpoint meta attribute in the README.v1.yml with drive time band info


edmkitty:
@nfasulo I don't know how much value it would add to mention an API that nobody has access to. Perhaps it's worth mentioning our own drive_time_bands endpoint once that gets added?

modules/va_facilities/README.v1.yml
```diff
@@ -449,6 +449,11 @@ paths:
         in the standard JSON API "links" and "meta" elements. GeoJSON responses include pagination
         information in the "Link" header.
 
+        Also in the "meta" element is information about the drive time bands returned for the facilities
+        in the response. The values of drive_time_band_min and drive_time_band_max are in minutes.
```
nfasulo: I'm trying to make this language more clear, but realize this I might be going overboard. Does this make sense?
> Also in the "meta" element is information about the drive time band that the requested location fell within for each the facility included in the response.

edmkitty: Yes, thank you!

jeff2d2:
Yeah, this is cool. Seems like a much better UX having what can/cannot go in here be well defined.

jeff2d2:
@nfasulo is updating this to remove `type` as a request param, as `/nearby` will only return VHA (health) facilities. He'll also be restructuring the response to not include the main facility object as part of it (it was determined that the user can call `/facilities/{ids}` for this info and therefore it isn't necessary to dup that here).

@jeff2d2 will update description language to more closely match what `/nearby` does.

modules/va_facilities/README.v1.yml
```diff
@@ -434,7 +414,7 @@ paths:
     get:
       tags:
         - facilities
-      summary: Query facilities based on drive time from a specified location
+      summary: Query which VA health (VHA) facilities are nearby, given a specified location and desired drive time
       description: |
         Retrieve all facilities contained within an isochrone (a polygon composed of the area
```
rtravitz: This still seems to imply that facilities are being retrieved based on a user-centered isochrone. We are now retrieving all facilities where the passed in location intersects within a facility's isochrone of duration less than the passed in drive time—which is a mouthful. Would it be better to say something more general like: "Retrieve all facilities reachable by driving within the specified time period."

modules/va_facilities/README.v1.yml
```diff
@@ -445,9 +425,13 @@ paths:
...
 
-        Results of this operation are paginated. JSON responses include pagination information
-        in the standard JSON API "links" and "meta" elements. GeoJSON responses include pagination
-        information in the "Link" header.
+        Results of this operation are paginated. Responses include pagination information
+        in the standard JSON API "links" and "meta" elements.
+
+        Also in the "meta" element is information about the drive time band that the requested location 
+        fell within for each the facility included in the response. The values of drive_time_band_min and 
+        drive_time_band_max are in minutes. For example, a facility returned with a matched drive_time_band_min
```
rtravitz: I don't think the implementation is currently set up for `drive_time_band_min`. Instead it is `drivetime_band_min` where `drivetime` is one word. The classes reflect that too with names like `DrivetimeBand` and the Postgres table name `drivetime_bands`. While that might not be correct, it is currently consistent. Should we make a final canonical call on whether it's `drive time` or `drivetime`?

modules/va_facilities/README.v1.yml
```diff
@@ -445,9 +425,13 @@ paths:
...
-        information in the "Link" header.
+        Results of this operation are paginated. Responses include pagination information
+        in the standard JSON API "links" and "meta" elements.
+
+        Also in the "meta" element is information about the drive time band that the requested location 
+        fell within for each the facility included in the response. The values of drive_time_band_min and 
+        drive_time_band_max are in minutes. For example, a facility returned with a matched drive_time_band_min 
+        of 10 and drive_time_band_max of 20 is a 10 to 20 minute drive from the requested location.
 
         The facility information returned by `/nearby` is a small subset of the information returned by `/facilities`.
```
rtravitz: I wouldn't even call the id link a "small subset of the information." Would it be better to say something like: "This endpoint does not return facility information. To retrieve details for facilities found by `/nearby` follow the relationship links and check documentation for `/facilities`."

modules/va_facilities/README.v1.yml
```diff
@@ -890,337 +884,383 @@ components:
...
+            - drive_time_band_min
+            - drive_time_band_max
           properties:
-            name:
+            drive_time_band_min:
               type: string
-              example: Washington VA Medical Center
-            facility_type:
+              example: 10
+            drive_time_band_max:
```
rtravitz: Same question as above about whether to go with `drivetime` or `drive_time`

modules/va_facilities/README.v1.yml
jeff2d2: For this, I was more going for what the user could expect to get out of this endpoint vs. how it functions on the inside. Although the polygons are now fixed to the facility itself instead of created on-the-fly based on the provided location, I'm not sure it matters to a user either way?

modules/va_facilities/README.v1.yml
edmkitty: drivetime is good, would prefer to be consistent with existing code

modules/va_facilities/README.v1.yml
rtravitz: If the user doesn't care about the underlying implementation, then I stick with the suggestion to just explain the final product: "Retrieve all facilities reachable by driving within the specified time period."

jeff2d2:
A decision was made to change `drive_time_band_min` and `drive_time_band_max` to `min_time` and `max_time`.

PR #3376, opened 2019-10-04
move common raise_backend_exception method to base service class

LindseySaari:
## Description of change
This change extracts duplicate methods to a single method definition in the Common::Client::Base class for services. To follow best practices, the common code was moved to the base service class.

## Testing done
Test suite
Manual review of all service classes referencing raise_backend_exception
Review of service class inheritance from base service class

- [x] Re-run of test suite
- [x] Investigation of current tests to verify that the exception raising case is still covered (Common::Exceptions::BackendServiceException)

#### Applies to all PRs

- [ ] Appropriate logging


lib/evss/service.rb
```diff
@@ -65,25 +65,16 @@ def handle_error(error)
 
       case error
       when Faraday::ParsingError
-        raise_backend_exception('EVSS502', self.class)
+        raise_backend_exception('EVSS502', "EVSS::#{self.class}")
```
jholton: The `EVSS::` prefix isn't necessary.  This results in `EVSS::EVSS::Service` being sent.

lib/common/client/base.rb
```diff
@@ -54,6 +54,15 @@ def self.configuration(configuration = nil)
         @configuration ||= configuration.instance
       end
 
+      def raise_backend_exception(key, source, error = nil)
+        raise Common::Exceptions::BackendServiceException.new(
+          key,
+          { source: source.to_s },
+          error&.status,
```
johnpaulashenfelter: Just curious, why would error not have a status or body?

lib/common/client/base.rb
```diff
@@ -54,6 +54,15 @@ def self.configuration(configuration = nil)
         @configuration ||= configuration.instance
       end
 
+      def raise_backend_exception(key, source, error = nil)
```
johnpaulashenfelter: I don't think there's a real plan around it, but how do we feel about keyword arguments ? Nothing you have to do here @LindseySaari just saw the method and was like -- oh, keywords.

lib/common/client/base.rb
johnpaulashenfelter: Ah -- probably just from extracting the method.

lib/common/client/base.rb
LindseySaari: Are you referring to using an options hash instead or basically running into issues where arguments with the same name as keywords can be dangerous in terms of being overridden?

johnpaulashenfelter: @LindseySaari I mean literal keyword arguments

`def raise_backend_exception(key:, source:, error: nil)`

this has nothing to do directly with your PR! This is just reviewing one method and focusing on one method made me wonder about moving towards keywords. Nothing actionable to do

LindseySaari: Right, I completely agree!

lib/evss/service.rb
LindseySaari: Fixing this 😄

PR #3377, opened 2019-10-04
Update drivetime_bands table fields

nfasulo:
## Description of change
Adding min and max values to the drive time bands instead of a single value. This will ensure that we are resilient to future changes to the size of the bands. This table is not yet in use.


db/migrate/20190920134906_create_drivetime_bands.rb
```diff
@@ -2,7 +2,8 @@ class CreateDrivetimeBands < ActiveRecord::Migration[5.2]
   def change
     create_table :drivetime_bands do |t|
       t.string :name
-      t.integer :value
+      t.integer :min
+      t.integer :max
```
edmkitty: Should we make a separate migration for this update?

nfasulo:
Thanks, @johnpaulashenfelter. Will correct.

PR #3378, opened 2019-10-04
Remove unused octokit gem

johnpaulashenfelter:
## Description of change

Was going to update the octokit gem (connects to Github) but found its no longer used in any active code.

Searched for any instance of `octokit` with no relevant results, though I found an old VCR cassette that had used the gem which is also no longer used.

## Testing done

Ran the specs to verify 
No specs break.

#### Applies to all PRs

- [ ] Appropriate logging


Gemfile.lock
```diff
@@ -393,8 +393,6 @@ GEM
     notiffany (0.1.3)
       nenv (~> 0.1)
       shellany (~> 0.0)
-    octokit (4.7.0)
```
johnpaulashenfelter: Removed one gem, now two less overall since sawyer isn't used by anything else. Win/win

PR #3382, opened 2019-10-04
19740: STEM: Update 1995s descriptor in the YTD reports

elder-joseph:
## Description of change
As a developer, I need to update the report name for STEM1995 in the YTD report so that it is distinguished from real form names.

https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/vets.gov-team/19740

## Testing done
Manual tests and `make ci` pass locally

#### Unique to this PR
- [x] The 1995 STEM column header text "1995s" is replaced with the following: “1995 STEM”

#### Applies to all PRs
- [x] Appropriate logging


app/workers/education_form/create_daily_fiscal_year_to_date_report.rb
```diff
@@ -14,7 +14,11 @@ class CreateDailyFiscalYearToDateReport
     }.freeze
 
     FORM_TYPE_HEADERS = EducationBenefitsClaim::FORM_TYPES.map do |form_type|
-      ["22-#{form_type}", '', '']
+      if form_type == '1995s'
```
johnpaulashenfelter: Seems like this would be clearer if you just materialize the constant -- literally type it in -- instead of dynamically generating it here *and* in the `app/workers/education_form/create_daily_year_to_date_report.rb`.

This seems like a UI deviation in the CSV file from all the other similar examples, which runs the risk of this `if` turning into something more complex with each additional change for similar reasons.

Since you're generating it in two different places, I'm wondering if you can make it a constant in a single place

```
FORM_TYPE_HEADERS=[
"22-1990", "", "", 
"22-1995", "", "", 
...
"22-1995 STEM", "", ""
]
```

And if you have to generate them, let's see about generating them in a single place.

johnpaulashenfelter: Did the spreadsheet/csv here just become three columns ("22-1995 STEM", "", "") wider?

elder-joseph: Hi @johnpaulashenfelter,

The spreadsheet/csv still has the same amount of columns, the only change was to what was being output to the report, in this case the header name which I was setting using the `if` statement check in those two places as you mentioned, since it differed from the form type slightly ("1995 STEM" vs "1995s").

I agree performing that check in two places was not ideal. I took your advice and created `FORM_HEADERS` inside of `EducationBenefitsClaim` as a way to decouple the form header from the form type, in case we ever do find ourselves in a situation where we need to output a different header than what was being generated dynamically.

I pushed these changes in another commit.

app/workers/education_form/create_daily_fiscal_year_to_date_report.rb
```diff
@@ -13,8 +13,8 @@ class CreateDailyFiscalYearToDateReport
       daily_processed: 0
     }.freeze
 
-    FORM_TYPE_HEADERS = EducationBenefitsClaim::FORM_TYPES.map do |form_type|
-      ["22-#{form_type}", '', '']
+    FORM_TYPE_HEADERS = EducationBenefitsClaim::FORM_HEADERS.map do |form_header|
+      [form_header.to_s, '', '']
```
johnpaulashenfelter: Aren't they already a string?

app/models/education_benefits_claim.rb
```diff
@@ -3,7 +3,9 @@
 require 'attr_encrypted'
 class EducationBenefitsClaim < ApplicationRecord
   FORM_TYPES = %w[1990 1995 1990e 5490 5495 1990n 0993 0994 1995s].freeze
-
+  FORM_HEADERS = [
+    '22-1990', '22-1995', '22-1990e', '22-5490', '22-5495', '22-1990n', '22-0993', '22-0994', '22-1995 STEM'
```
johnpaulashenfelter: Definitely makes labeling clearer.

Not asking for it, but this is getting pretty close to a real data structure where there's an ordered set of form types and form headers. These two arrays are next to each other so its pretty quick to debug if they get mismatched but there's also the option to do something like

```
FORMS=
[
  {code: '1990', header: '22-1990'},
...
  {code: '1995s', header: '22-1995 STEM'},
]
```

That would let you be even more specific and clear in the loops

`FORM_TYPE_HEADERS = EducationBenefitsClaim::FORMS.map { |form|       [form[:header], '', ''] }` 

or something similar. 

But that means rewriting everything that already uses `FORM_TYPES`

In any case now the headers are separate from the types which is a plus 👍

app/workers/education_form/create_daily_year_to_date_report.rb
```diff
@@ -11,8 +11,8 @@ class CreateDailyYearToDateReport
       daily_processed: 0
     }.freeze
 
-    FORM_TYPE_HEADERS = EducationBenefitsClaim::FORM_TYPES.map do |form_type|
-      ["22-#{form_type}", '', '']
+    FORM_TYPE_HEADERS = EducationBenefitsClaim::FORM_HEADERS.map do |form_header|
+      [form_header.to_s, '', '']
```
johnpaulashenfelter: If these are the only two places you use `FORM_TYPE_HEADERS` you can still hardcode _the whole thing_

`'22-1990', '', '', '22-1995', , '', ''...` and avoid any need for someone to figure out that the headers actually look like.

PR #3395, opened 2019-10-07
add base64 support to multipart parser

charleystran:
## Description of change
This change allows the benefits intake api to allow for the file contents to be base64 encoded on upload to S3

## Testing done
rspec and local tests

## Testing planned
testing on staging

- [x] a base64 encoded pdf can be uploaded and processed properly

#### Applies to all PRs

- [x] Appropriate logging
- [x] Part of https://github.com/department-of-veterans-affairs/vets-contrib/issues/1990 


modules/vba_documents/lib/vba_documents/multipart_parser.rb
```diff
@@ -27,6 +34,23 @@ def self.parse(infile)
       end
     end
 
+    def self.base64_encoded(infile)
+      content = File.read(infile)
+      content.start_with?('data:application/pdf;base64,')
```
bastosmichael: Are we sure this is how it'll always start or does it make sense to see if it includes just `base64` somewhere in the payload instead? Wouldn't want it to skip because it didn't find this exact match etc.

charleystran: I think this is the spec we should require. technically the base64 data is after the , and we strip out this info out before decoding it. So we need to know what to strip out and I think it would be best to have that be a hard requirement vs looser rules. However I am open to other opinions

bastosmichael: As long as we remember to include this in the documentation when announcing this to the world then I'm good with it.

PR #3399, opened 2019-10-07
Remove call to set_sso_cookie

bshyong:
## Description of change
The MHV SSO cookie is not used in the SSOe flow, and the call to set the cookie will throw an error. This PR conditionally skips the call because the cookie is irrelevant for SSOe purposes.

## Testing done
Local

## Testing planned
Staging

- [x] Appropriate logging


patrickvinograd:
This is an iterative change needed as part of debugging SSOe authentication. The complete flow can't be debugged locally because we don't have a VPN-enabled local development environment, and SSOe's dev environment is VPN-only.

bshyong:
@annaswims thanks for the context here—given this logic it seems it may be better to leave this cookie in place and simply add the mapping for SSOe into `sso_cookie_sign_credential_used`?

patrickvinograd:
Ultimately we do want to stop setting the shared cookie - I think the answer is that this controller. concern is not really shared code between these implementations anymore. But I see how this will cause the user session to immediately get reset, mistakenly thinking that MHV reset the cookie. So Ben's suggestion is an okay interim step.

PR #3400, opened 2019-10-07
Kr/create mock drive time band data

edmkitty:
Changes in this PR: 
- Adds a client for the PSSG DrivetimeBand endpoint
- Adds betamocks configuration for the endpoint
- Adds drive time base path to locators in settings.yml

Also see the update in vets-api-mockdata: 
https://github.com/department-of-veterans-affairs/vets-api-mockdata/pull/98

Fulfills this ticket:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/3102


rtravitz:
I am having trouble getting it to use the new mocks, even though I can find them at the correct path on my machine. Investigating more... 🔍

config/betamocks/services_config.yml
```diff
@@ -464,3 +464,16 @@
...
       :path: "/server/rest/services/VA/FacilitySitePoint_VHA/FeatureServer/0"
       :file_path: "gis/vha/metadata"
+
+#PSSG
+- :name: 'PSSG'
+  :base_uri: <%= "#{URI(Settings.locators.drive_time_band_base_path).host}:#{URI(Settings.locators.drive_time_band_base_path).port}" %>
+  :endpoints:
+    - :method: :get
+      :path: "/arcgis2/rest/services/Portal/MonthlyVAST_TTB/FeatureServer/0/*"
+      :file_path: "pssg/drive_time_bands"
```
rtravitz: Figured out why I couldn't get a match! This should be `pssg/drivetime_bands` to match up with the structure in the `vets-api-mockdata` repo.

edmkitty: d'oh

lib/facilities/drivetime_band_client.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+require 'common/client/base'
+
+module Facilities
+  # Core class responsible for api interface operations
+  class DrivetimeBandClient < Common::Client::Base
+    configuration Facilities::DrivetimeBandConfiguration
+
+    def get_all_drivetime_bands(offset, limit)
```
rtravitz: Should we still call this `get_all_drivetime_bands` considering that it now takes an offset and a limit? Maybe it should just be `get_drivetime_bands`? Is there some other piece that will call this in the future and iterate through all of them?

lib/facilities/drivetime_band_configuration.rb
```diff
@@ -0,0 +1,28 @@
...
+require 'common/client/middleware/response/raise_error'
+
+module Facilities
+  class DrivetimeBandConfiguration < Common::Client::Configuration::REST
+    def base_path
+      Settings.locators.drive_time_band_base_path
+    end
+
+    def service_name
+      'FL'
```
rtravitz: Is this still part of FL, which I assume stands for Facility Locator? Is that service name for anything related to facilities, or just for things that serve the actual Facility Locator feature on va.gov?

lib/facilities/drivetime_band_client.rb
edmkitty: Good catch! I will rename it. My half-baked plan was to give the future sidekiq download job the responsibility of iterating through all the bands.

lib/facilities/drivetime_band_configuration.rb
edmkitty: I'm not sure about the meaning or all the places it's used, but it is at least used for labels in error handling. For example, here: https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/common/client/base.rb#L113

PR #3404, opened 2019-10-08
GIDS - Institution Program Search

dmccurdybah:
#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging


annaswims:
depends on https://github.com/department-of-veterans-affairs/gibct-data-service/pull/452 so we'll review that first.

app/swagger/requests/gibct/institution_programs.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
```
annaswims: Please add examples to your swagger docs.  This would be a lot more clear if realistic examples were used instead of `"string`
Screen Shot 2019-10-21 at 5 24 12 PM

app/swagger/schemas/gibct/institution_programs.rb
```diff
@@ -0,0 +1,105 @@
...
+      class InstitutionPrograms
+        include Swagger::Blocks
+
+        swagger_schema :GibctInstitutionProgramsAutocomplete do
+          key :required, %i[data meta links]
+
+          property :data, type: :array, minItems: 0, uniqueItems: true do
+            items do
+              property :id, type: :integer
+              property :value, type: :string
```
annaswims: I'd like to see example values here.

app/controllers/v0/gi/institution_programs_controller.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module V0
+  module GI
+    class InstitutionProgramsController < GIController
+      def autocomplete
+        render json: client.get_institution_program_autocomplete_suggestions(scrubbed_params)
```
omgitsbillryan: I'm concerned by the following:
- This is a synchronous request -`vets-api` will hold the request until the gids service responds. 
- Autocomplete endpoints tend to get called frequently

Even if the gids service is very responsive and FE devs use a sensible bounce interval for calling this endpoint, we should try to code defensively to protect `vets-api`. Have you made any considerations for performance?

Perhaps look into using [`cache_aside`](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/common/models/concerns/cache_aside.rb).

jholton: This is definitely an existing issue that has [wreaked havoc](https://github.com/department-of-veterans-affairs/vets.gov-team/blob/master/Postmortems/2019-06-10-GI-Latency.md) before, but was never addressed.

dmccurdybah: The institution_program autocomplete will be used by future stories in vets-website.  We also create stories for updating the new institution_program and existing institution autocompletes to use `cache_aside` in a future story. That's definitely an important existing issue that'd be great to fix.

app/swagger/schemas/gibct/institution_programs.rb
```diff
@@ -0,0 +1,105 @@
...
+        swagger_schema :GibctInstitutionProgramsSearchMeta do
+          key :type, :object
+          key :required, %i[version count facets]
+
+          property :version, '$ref': :GibctVersion
+
+          property :count, type: :integer
+          property :facets, type: :object do
+            key :required, %i[type state country]
+            property :type, type: :object do
```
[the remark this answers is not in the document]
dmccurdybah: Can we handle this in a follow up story? The existing GIDS services are also not well documented, and it seems like we could have one story to bring that up to spec without holding up approval for this PR (especially since it's not currently being used by vets-website in prod)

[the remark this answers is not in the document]
annaswims: these sort of things tend to be forgotten.  I'll approve if you link to an issue.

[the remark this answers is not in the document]
dmccurdybah: @annaswims can you take a look at this so we can get this in?  Thanks!

PR #3406, opened 2019-10-08
Apply Formation CSS to Flipper admin site in deployed environments.

U-DON:
## Description of change
Rails is disabled from static file hosting in deployed environments, so the Flipper UI looks very broken. As a quick fix, I'm pulling in the stylesheet from design.va.gov. It's still not fully styled since the site's CORS policy is blocking the loading of fonts and icons, but it's mostly there.


annaswims:
We have a pesky intermittently failing test that JP is looking into so I restarted your tests for you.

PR #3407, opened 2019-10-08
Rubocop rspec fixes round 9

thilton-oddball:
## Description of change
Continuation of fixing Rubocop Rspec offenses.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -167,3 +167,7 @@ Style/BlockDelimiters:
   Exclude:
     - "spec/factories/*.rb"
     - "modules/*/spec/factories/*.rb"
+
+# With specs in several different places, this should remain disabled unless or until we put all specs under the main specs directory.
+RSpec/FilePath:
+  Enabled: false
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -64,13 +64,6 @@ RSpec/ExampleLength:
 RSpec/ExpectInHook:
   Enabled: false
 
-# Offense count: 222
-# Cop supports --auto-correct.
-# Configuration parameters: EnforcedStyle.
-# SupportedStyles: implicit, each, example
-# RSpec/HookArgument:
-#   Enabled: false
-
 # Offense count: 254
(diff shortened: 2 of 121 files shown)


.rubocop_todo.yml
```diff
@@ -64,17 +64,12 @@ RSpec/ExampleLength:
...
-RSpec/FilePath:
-  Enabled: false
-
 # Offense count: 222
 # Cop supports --auto-correct.
 # Configuration parameters: EnforcedStyle.
 # SupportedStyles: implicit, each, example
-RSpec/HookArgument:
-  Enabled: false
+# RSpec/HookArgument:
```
jholton: Did you mean to just comment this out, or should we remove it?

thilton-oddball: No, I meant to delete them.

.rubocop.yml
```diff
@@ -167,3 +167,7 @@ Style/BlockDelimiters:
   Exclude:
     - "spec/factories/*.rb"
     - "modules/*/spec/factories/*.rb"
+
+# With specs in several different places, this should remain disabled unless or until we put all specs under the main specs directory.
+RSpec/FilePath:
+  Enabled: false
```
jholton: If you run this cop, you'll find 3 different types of failures:
* all specs in `modules` directory*
* violations due to files not following standard Ruby/Rails naming conventions (i.e. the directory path to the file does not match the class/module name)
* spec files whose naming/directory path does not match up with the class it is testing


I'd like to see you take a shot (in a different PR) of fixing any of the last type of violations I mentioned.

I'm not sure of an easy solution for making specs in `modules` compliant.
You can exclude them in your run:
```
RSpec/FilePath:
  Exclude:
    - 'modules/**/*'
```

As far as my second point, it's probably something we should look into, but I'm wary of moving core files around just to satisfy Rubocop

To sum it up: I think disabling this is fine for now, but I'd like to see a PR with any fixes that can be made by 
changing the config to
```
RSpec/FilePath:
  Exclude:
    - 'modules/**/*'
```
and moving/renaming files in the `spec/` directory only.

PR #3409, opened 2019-10-09
convert address to lat lng

whuang85:
## Description of change
Change nearby facility address method to get the lat lng of the address and call the lat lng method instead of getting the isochrones with address.

https://github.com/department-of-veterans-affairs/vets-contrib/issues/3317

## Testing done
Unit tests were updated

#### Applies to all PRs

- [x] Appropriate logging


app/models/nearby_facility.rb
```diff
@@ -36,13 +42,25 @@ def request_isochrone(waypoint, params)
...
 
-      response = Faraday.get "#{Settings.bing.base_api_url}/Isochrones", query
+      response = Faraday.get "#{Settings.bing.base_api_url}/Routes/Isochrones", query
       response_body = JSON.parse(response.body)
       handle_bing_errors(response_body, response.headers)
 
       response_body
     end
 
+    def request_location(address)
```
edmkitty: Could you add a quick test for this in `nearby_facility_spec.rb` please?

app/models/nearby_facility.rb
```diff
@@ -67,6 +85,15 @@ def make_linestring(polygon)
       polygon.map { |point| "#{point[1]} #{point[0]}" }.join(',')
     end
 
+    def parse_location(response_json)
+      response_json.try(:[], 'resourceSets')
```
johnpaulashenfelter: Wouldn't `dig('resourceSets', 'resources'....)` work here as well? I think that might be a lot more readable

config/settings.yml
```diff
@@ -386,7 +386,7 @@ search:
   url: https://search.usa.gov/api/v2
 
 bing:
-  base_api_url: 'https://dev.virtualearth.net/REST/v1/Routes'
+  base_api_url: 'https://dev.virtualearth.net/REST/v1'
```
johnpaulashenfelter: Is there a corresponding change that needs to be made in production/staging config?

app/models/nearby_facility.rb
```diff
@@ -36,13 +42,25 @@ def request_isochrone(waypoint, params)
         key: Settings.bing.key
       }
 
-      response = Faraday.get "#{Settings.bing.base_api_url}/Isochrones", query
+      response = Faraday.get "#{Settings.bing.base_api_url}/Routes/Isochrones", query
```
johnpaulashenfelter: You moved `/Routes` here from the settings.yml file (which makes sense to me) -- but if production/staging doesn't have a corresponding URL then this will turn into `/Routes/Routes`

app/models/nearby_facility.rb
```diff
@@ -10,9 +10,15 @@ class << self
     def query(street_address: '', city: '', state: '', zip: '', **params)
       return NearbyFacility.none unless [street_address, city, state, zip].all?(&:present?)
 
-      waypoint = "#{street_address} #{city} #{state} #{zip}"
-      isochrone_response = request_isochrone(waypoint, params)
-      get_facilities_in_isochrone(params, isochrone_response)
```
johnpaulashenfelter: Are these methods still necessary?

app/models/nearby_facility.rb
edmkitty: Does that work when there are arrays involved?

config/settings.yml
edmkitty: The 'dev' is referring to the bing api available to devs, not our dev environment. This link is used in all environments.

app/models/nearby_facility.rb
whuang85: not anymore for the method that takes in an address as we now want to call an additional bing api to get the lat/lng and call the existing get nearby facilities by lat/lng method

app/models/nearby_facility.rb
whuang85: thanks, will check that

app/models/nearby_facility.rb
johnpaulashenfelter: Awesome. Love seeing code removed

app/models/nearby_facility.rb
johnpaulashenfelter: @whuang85 maybe `dig(:resourceSets).first.dig(:resources).first.dig(:point, :coordinates)`? Same idea with the `first`s you have here.

config/settings.yml
johnpaulashenfelter: @edmkitty Ah -- i wasn't clear about what change i meant. 

The `base_api_url` changed here in the settings.yml file, but there are different ways that the `bing.base_api_url` are set in staging and/or production settings files/env that I wanted to make sure got the same change. /cc @whuang85

johnpaulashenfelter: Also @whuang85 dealt with this in the comment below when the `/Routes` got moved

PR #3410, opened 2019-10-09
Timecop fixes in specs

jholton:
## Description of change
I found a number of `Timecop.freeze` calls that did not have corresponding `Timecop.return` calls.  
This was causing spec failures on certain runs.
Example:  
`rspec --seed 41882` resulted in [this expectation](https://github.com/department-of-veterans-affairs/vets-api/blob/a58358f8e2d57b242b6d04a2891c58d268f052c9/spec/controllers/v1/sessions_controller_spec.rb#L363) in `sessions_controller_spec` (wherein two timestamps were expected to be different) to fail because [this use of `Timecop.freeze`](https://github.com/department-of-veterans-affairs/vets-api/blob/a58358f8e2d57b242b6d04a2891c58d268f052c9/spec/request/user_request_spec.rb#L225) in 
`user_request_spec` was never unfrozen.

I also removed a couple of redundant method call stubs.

## Testing done
specs

## Testing planned
n/a

#### Unique to this PR

- [x] Add missing `Timecop.return` calls to specs
- [x] remove unnecessary `before` block

#### Applies to all PRs

- [ ] ~~Appropriate logging~~

### DIFF
--- a/modules/vba_documents/spec/jobs/upload_processor_spec.rb
+++ b/modules/vba_documents/spec/jobs/upload_processor_spec.rb
@@ -355,6 +355,7 @@ RSpec.describe VBADocuments::UploadProcessor, type: :job do
         Timecop.freeze(Time.zone.now)
         described_class.new.perform(upload.guid)
         expect(described_class.jobs.last['at']).to eq(30.minutes.from_now.to_f)
+        Timecop.return
       end
     end
 
--- a/spec/controllers/v0/sessions_controller_spec.rb
+++ b/spec/controllers/v0/sessions_controller_spec.rb
@@ -307,10 +307,6 @@ RSpec.describe V0::SessionsController, type: :controller do
     end
 
     describe 'POST saml_callback' do
-      before(:each) do
-        allow(SAML::User).to receive(:new).and_return(saml_user)
-      end
-
       around(:each) do |example|
         Settings.sso.cookie_enabled = true
         example.run
--- a/spec/controllers/v1/sessions_controller_spec.rb
+++ b/spec/controllers/v1/sessions_controller_spec.rb
@@ -307,10 +307,6 @@ RSpec.describe V1::SessionsController, type: :controller do
     end
 
     describe 'POST saml_callback' do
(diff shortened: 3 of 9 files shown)


spec/controllers/v1/sessions_controller_spec.rb
```diff
@@ -307,10 +307,6 @@ def expect_logger_msg(level, msg)
     end
 
     describe 'POST saml_callback' do
-      before(:each) do
-        allow(SAML::User).to receive(:new).and_return(saml_user)
-      end
```
jholton: [This is redundant](https://github.com/department-of-veterans-affairs/vets-api/blob/a58358f8e2d57b242b6d04a2891c58d268f052c9/spec/controllers/v1/sessions_controller_spec.rb#L153).

spec/controllers/v0/sessions_controller_spec.rb
```diff
@@ -307,10 +307,6 @@ def expect_logger_msg(level, msg)
     end
 
     describe 'POST saml_callback' do
-      before(:each) do
-        allow(SAML::User).to receive(:new).and_return(saml_user)
-      end
```
jholton: [This is redundant.](https://github.com/department-of-veterans-affairs/vets-api/blob/a58358f8e2d57b242b6d04a2891c58d268f052c9/spec/controllers/v0/sessions_controller_spec.rb#L153)

jholton:
Perhaps we should use [`Timecop.safe_mode`](https://github.com/travisjeffery/timecop#timecopsafe_mode)

modules/vba_documents/spec/jobs/upload_processor_spec.rb
```diff
@@ -355,6 +355,7 @@
         Timecop.freeze(Time.zone.now)
         described_class.new.perform(upload.guid)
         expect(described_class.jobs.last['at']).to eq(30.minutes.from_now.to_f)
+        Timecop.return
```
johnpaulashenfelter: Nice find.

PR #3411, opened 2019-10-09
Add logout URL for SSOe flow

bshyong:
## Description of change
SSOe has a logout mechansim that is not based on SAML single logout. 

This PR adds logout routes and handling for the SSOe logout flow
- adds new route `v1/sessions/ssoe_slo/new` for SSOe logout
- adds callback route `/v1/sessions/ssoe_logout` as logout callback route
- adds handling for triggering logout flow with SSOe

Resolves https://github.com/department-of-veterans-affairs/va-sso/issues/32

This is an iterative change needed as part of debugging SSOe authentication. The complete flow can't be debugged locally because we don't have a VPN-enabled local development environment, and SSOe's dev environment is VPN-only.

## Testing done
n/a

## Testing planned
dev/staging

- [x] Appropriate logging

### DIFF
--- a/app/controllers/v1/sessions_controller.rb
+++ b/app/controllers/v1/sessions_controller.rb
@@ -30,7 +30,7 @@ module V1
         reset_session
       elsif type == 'ssoe_slo'
         Rails.logger.info('SSOe: LOGOUT', sso_logging_info)
-        redirect_to(url_service.send(:ssoe_slo_url)) && return
+        reset_session
       end
       # clientId must be added at the end or the URL will be invalid for users using various "Do not track"
       # extensions with their browser.
@@ -38,7 +38,6 @@ module V1
     end
 
     def ssoe_slo_callback
-      reset_session
       redirect_to url_service.base_redirect_url
     end
 
--- a/app/controllers/v1/sessions_controller.rb
+++ b/app/controllers/v1/sessions_controller.rb
@@ -31,7 +31,10 @@ module V1
         reset_session
       elsif type == 'ssoe_slo'
         Rails.logger.info('SSOe: LOGOUT', sso_logging_info)
-        reset_session
+if type == 'slo' || type == 'ssoe_slo'
+  Rails.logger.info("LOGOUT of type #{type}", sso_logging_info)
+  reset_session
+end
       end
       # clientId must be added at the end or the URL will be invalid for users using various "Do not track"
       # extensions with their browser.
--- a/app/controllers/v1/sessions_controller.rb
+++ b/app/controllers/v1/sessions_controller.rb
@@ -26,7 +26,7 @@ module V1
       StatsD.increment(STATSD_SSO_NEW_KEY, tags: ["context:#{type}"])
       url = url_service.send("#{type}_url")
 
-      if type == 'slo' || type == 'ssoe_slo'
+      if %w[slo ssoe_slo].include?(type)
         Rails.logger.info("LOGOUT of type #{type}", sso_logging_info)
         reset_session
       end
--- a/app/controllers/v1/sessions_controller.rb
+++ b/app/controllers/v1/sessions_controller.rb
@@ -7,7 +7,7 @@ require 'saml/responses/logout'
 
 module V1
   class SessionsController < ApplicationController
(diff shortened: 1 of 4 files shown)


lib/saml/url_service.rb
```diff
@@ -14,6 +14,8 @@ class URLService
       'http://127.0.0.1:3000' => { base_redirect: 'http://127.0.0.1:3001' }
     }.freeze
 
+    SSOE_LOGOUT_URL = 'https://int.eauth.va.gov/pkmslogout'
```
omgitsbillryan: Probably deserves to be a `Settings.yml` config instead of hardcoding it in `url_service.rb`.

app/controllers/v1/sessions_controller.rb
```diff
@@ -24,15 +24,24 @@ def new
 
       StatsD.increment(STATSD_SSO_NEW_KEY, tags: ["context:#{type}"])
       url = url_service.send("#{type}_url")
+
       if type == 'slo'
         Rails.logger.info('SSO: LOGOUT', sso_logging_info)
         reset_session
+      elsif type == 'ssoe_slo'
+        Rails.logger.info('SSOe: LOGOUT', sso_logging_info)
+        redirect_to(url_service.send(:ssoe_slo_url)) && return
```
omgitsbillryan: Couldn't you let the `redirect_to` on [line 37](https://github.com/department-of-veterans-affairs/vets-api/pull/3411/files#diff-fe28872274afc489d12314e50a3f3b4cR37) happen? Also, `url` will already contain your ssoe url so it seemes redundant to call `url_service.send(:ssoe_slo_url)` again.

bshyong: 👍 I had originally named the method something else in url service but makes sense to streamline this now that the type parameter and method name are consistent

app/controllers/v1/sessions_controller.rb
```diff
@@ -24,15 +24,23 @@ def new
...
+      elsif type == 'ssoe_slo'
+        Rails.logger.info('SSOe: LOGOUT', sso_logging_info)
+        reset_session
       end
       # clientId must be added at the end or the URL will be invalid for users using various "Do not track"
       # extensions with their browser.
       redirect_to params[:client_id].present? ? url + "&clientId=#{params[:client_id]}" : url
     end
 
+    def ssoe_slo_callback
```
bshyong: This is the URL we can have SSOe redirect to after logout is called. This method will allow us to have control over where the user ultimately lands after logout.

annaswims: Could you just make a change here? https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/saml/url_service.rb#L57

bshyong: I think we still need this method because it is tied to a callback url that SSOe will call after they successfully trigger logout. https://github.com/department-of-veterans-affairs/vets-api/blob/972f0d7dd46478b88316a34811b186e67336a608/config/routes.rb#L20

Right now it just calls the redirect URL in url_service and we can make changes later on to the URL either here or in the URL service if needed.

app/controllers/v1/sessions_controller.rb
```diff
@@ -25,15 +25,23 @@ def new
 
       StatsD.increment(STATSD_SSO_NEW_KEY, tags: ["context:#{type}"])
       url = url_service.send("#{type}_url")
+
       if type == 'slo'
         Rails.logger.info('SSO: LOGOUT', sso_logging_info)
         reset_session
+      elsif type == 'ssoe_slo'
+        Rails.logger.info('SSOe: LOGOUT', sso_logging_info)
+        reset_session
```
omgitsbillryan:
```suggestion
if type == 'slo' || type == 'ssoe_slo'
  Rails.logger.info("LOGOUT of type #{type}", sso_logging_info)
  reset_session
end
```

I think we can DRY this up a bit.

bshyong: you're right—I think the logs are just used for debugging so it should be fine to change the message and DRY here. Thanks for the suggestion

PR #3423, opened 2019-10-11
VAOS Module and User Service

kreek:
## Description of change
Creates the VAOS module and the user service which is used to retrieve a JWT token that can be used for calls to other VAOS services

## Testing done
unit tests

## Testing planned
staging integration tests

#### Unique to this PR
- [ ] VOAS rails engine module
- [ ] VOAS service configuration
- [ ] VOAS user service

#### Applies to all PRs

- [ ] Appropriate logging


config/initializers/inflections.rb
```diff
@@ -7,9 +7,11 @@
   inflect.acronym 'EVSS'
   inflect.acronym 'GI'
   inflect.acronym 'IHub'
+  inflect.acronym 'JWT'
```
johnpaulashenfelter: Why to do you to inflect this? Its just a module name below, right?

modules/vaos/.rubocop.yml
```diff
@@ -0,0 +1,14 @@
+inherit_from:
+  - ../../.rubocop.yml
+
+AllCops:
+  Exclude:
+    - 'bin/*'
+    - bin/rails
+    - db/migrate/*.rb
+    - 'spec/dummy/**/*'
```
johnpaulashenfelter: Necessary?

modules/vaos/.rubocop.yml
```diff
@@ -0,0 +1,14 @@
...
+    - spec/dummy/db/schema.rb
+
+Metrics/BlockLength:
+  Exclude:
+    - spec/**/*.rb
```
johnpaulashenfelter: This is already in the included `rubocop.yml` -- if that doesn't work when you're in the module, maybe we need to hoist `modules/*/spec/**/*.rb` up to that file to avoid including hit in each?

modules/vaos/app/services/vaos/configuration.rb
```diff
@@ -0,0 +1,31 @@
...
+
+        faraday.response :betamocks if mock_enabled?
+        faraday.response :snakecase
+        faraday.response :json, content_type: /\bjson$/
+        faraday.adapter Faraday.default_adapter
+      end
+    end
+
+    def mock_enabled?
+      [true, 'true'].include?(Settings.va_mobile.mock)
```
johnpaulashenfelter: Everywhere else we use a more idiomatic pattern `Settings.va_mobile.mock || false` or something similar

modules/vaos/app/services/vaos/jwt.rb
```diff
@@ -0,0 +1,39 @@
...
+      {
+        authenticated: true,
+        sub: @user.icn,
+        idType: 'ICN',
+        iss: 'gov.va.vaos',
+        firstName: @user.first_name,
+        lastName: @user.last_name,
+        authenticationAuthority: 'gov.va.iam.ssoe.v1',
+        jti: SecureRandom.uuid,
+        nbf: now.utc.to_i - 60,
```
johnpaulashenfelter: Is this `1.minute.ago`?

modules/vaos/app/services/vaos/jwt.rb
```diff
@@ -0,0 +1,39 @@
...
+        exp: now.utc.to_i + 3600,
```
johnpaulashenfelter: Is this `1.hour.from_now`?

modules/vaos/lib/tasks/vaos_tasks.rake
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
```
johnpaulashenfelter: Is this a placeholder?

modules/vaos/lib/vaos.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+require 'vaos/engine'
+
+module VAOS
+  # Your code goes here...
```
johnpaulashenfelter: Is this a placeholder?

modules/vaos/spec/services/configuration_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  describe '#connection' do
+    it 'returns a connection' do
+      expect(VAOS::Configuration.instance.connection).not_to be_nil
+    end
+  end
+
+  describe '#mock_enabled?' do
+    context 'when Settings.va_mobile.mock is true' do
```
johnpaulashenfelter: The `before` blocks here are each used by a single example. Seems like inlining them into the example would make each example clear and independent.

modules/vaos/spec/services/jwt_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+
+require 'rails_helper'
+
+describe VAOS::JWT do
+  subject { VAOS::JWT.new(user) }
+
+  let(:user) { build(:user, :mhv) }
+
+  describe '.new' do
+    it 'creates a VAOS::JWT instance' do
```
johnpaulashenfelter: This spec doesn't seem to have a real purpose. The subject is `VAOS::JWT.new(user)` -- so I don't see this being very valuable. 

Maybe remove this and inline the subject into the `#token` definition?

modules/vaos/vaos.gemspec
```diff
@@ -0,0 +1,22 @@
...
+
+# Describe your gem and declare its dependencies:
+Gem::Specification.new do |spec|
+  spec.name        = 'vaos'
+  spec.version     = VAOS::VERSION
+  spec.authors     = ['Alastair Dawson']
+  spec.email       = ['alastair.j.dawson@gmail.com']
+  spec.homepage    = 'https://api.va.gov'
+  spec.summary     = 'An api.va.gov module'
+  spec.description = 'This module was auto-generated please update this description'
```
johnpaulashenfelter: Please update 😄

modules/vaos/app/services/vaos/configuration.rb
kreek: did it this to only allow the boolean true or the string 'true', e.g. "fals" is truthy. It's also in the generator for modules, although I have to admit I'm biased because I wrote it 😄 
https://github.com/department-of-veterans-affairs/vets-api/blob/71b87a85784cc236710717dc823d8b6ab2266a28/lib/generators/module/templates/app/services/configuration.rb.erb#L27

modules/vaos/lib/tasks/vaos_tasks.rake
kreek: yes, part of the module generator, which is copied from the rails engine generator. It could be removed though.

modules/vaos/lib/vaos.rb
kreek: yes, also from the generator, `require 'vaos/engine'` is needed but I'll remove the empty module

modules/vaos/spec/services/configuration_spec.rb
kreek: It would be more concise but I believe the context block is for the state the spec is running in. In this case flipping the mock setting. Then the it block is for confirming the expectation. Ideally in one line but obviously that's not always possible 😄

modules/vaos/spec/services/jwt_spec.rb
kreek: This was a relic of TDD'ing the class, and I agree doesn't make much sense now, I'll remove it

modules/vaos/app/services/vaos/configuration.rb
saneshark: I think I would agree with Alastair here that this is technically more safe. Since any value other than `false` or `nil` would evaluate as a truthy. 

Might want to make changes elsewhere that are doing `Settings.va_mobile.mock || false`

modules/vaos/app/services/vaos/user_service.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    def create_session(user)
+      url = '/users/v2/session?processRules=true'
+      token = VAOS::JWT.new(user).token
+      response = perform(:post, url, token, 'Content-Type' => 'text/plain', 'Referer' => 'https://api.va.gov')
+      if response&.body
+        SessionStore.new(user_uuid: user.uuid, token: response.body).save
+        response.body
+      end
+    end
```
saneshark: this doesn't appear to handle exceptions; if the the service is down or unavailable will this be invoked indefinitely?

kreek: Ya the Faraday error gets caught by the common client and that error bubbles up to the controller that was trying to use the token. Then the application controller error handling would return a json error response. Breakers also tracks metrics and, as you mentioned, would kick in if enough calls failed. In a later PR we can add VAOS specific errors to exceptions.en.yml and could add additional metrics.

modules/vaos/README.rdoc
```diff
@@ -0,0 +1,10 @@
+= Vaos
+Rails engine module for the VAOS (VAMF) integration.
+
+== Installation
+Ensure the following line is in the root project's Gemfile:
+
+  gem 'vaos', path: 'modules/vaos'
+
+== License
+This module is open source under the terms of the [MIT License](https://opensource.org/licenses/MIT).
```
annaswims: vets-api has a license [CC0 1.0 Universal Summary](https://github.com/department-of-veterans-affairs/vets-api/blob/master/LICENSE.md) and I don't think we want a separate license for gems.

modules/vaos/Rakefile
```diff
@@ -0,0 +1,31 @@
...
+  rdoc.rdoc_files.include('lib/**/*.rb')
+end
+
+load 'rails/tasks/statistics.rake'
+
+require 'bundler/gem_tasks'
+
+require 'rake/testtask'
+
+Rake::TestTask.new(:test) do |t|
```
annaswims: is this generator cruft?

modules/vaos/app/services/vaos/jwt.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
```
annaswims: Is this service document somewhere yet?

modules/vaos/Rakefile
kreek: yes from the va module generator, but originally comes from the rails engine generator, but we can remove both since we use rspec

modules/vaos/app/services/vaos/jwt.rb
kreek: There's not yet any publicly accessible documentation for VAOS as a whole. The VAOS docs are all in a confluence wiki. This User Service is only for getting a JWT token to connect to the other VAOS endpoints. In follow-up PRs all the endpoints will be feature flagged and won't be turned on in prod until the app is complete. At that time we'll have public documentation in place.

PR #3426, opened 2019-10-14
[1781] Higher Level Review show & intake_status Endpoints

camerontesterman:
## Originating Issues
[1781 - Create Higher Level Review Retrieve details endpoint](https://github.com/department-of-veterans-affairs/va.gov-team/issues/1781)
[1777 - Create Higher Level Review Retrieve intake status endpoint](https://github.com/department-of-veterans-affairs/va.gov-team/issues/1771)

## Description of change
Adds two new routes, and methods for handling them, for the appeals controller. The two new routes added are a GET higher level review route, which shows detailed information about a particular higher level review. This leverages the DecisionReview service to call the caseflow API to retrieve data about the specified higher level review. A UUID must be supplied in the request to retrieve the details about the review.

The other route and method added shows the intake status of a particular review. Like the other addition mentioned above, it also utilizes the Decision Review service to make the appropriate call to the caseflow API and likewise requires a UUID supplied in the request path in order to retrieve the status.

## Testing done
Has been tested locally and confirmed using the VCR cassettes that were recorded when the service (DecisionReview) was created.

## Testing planned
None at this time (Dec 17, 2019)

The two routes work as expected, returning the appropriate data when supplied a valid UUID.

#### Applies to all PRs

- [x] Appropriate logging


spec/request/appeals_request_spec.rb
```diff
@@ -2,13 +2,75 @@
...
 
+  uuid = '1234567890'
+  ssn = '111223333'
+
+  %w[loa1 loa3].each do |user|
+    define_method("with_#{user}_user") do |options*|
+      using_ssn = options[:without_ssn] ? nil : ssn
+      let(:user) { FactoryBot.create(:user, user.to_sym, ssn: using_ssn) }
+    end
+  end
```
kfrz: This is hard to understand. It seems you could just define two `shared_contexts` which would read easier.

```
RSpec.shared_context "with_loa1_user" do
  let(:user) { ... }
end
```

Also, what does the `|options*|` splat usage do? I've never seen it used like so and can't find docs. Doesn't seem like you're passing it anywhere

app/swagger/requests/appeals.rb
```diff
@@ -59,6 +59,56 @@ class Appeals
...
+            key :description, 'Bad Gateway: the upstream appeals app returned an invalid response (500+)'
+            schema do
+              key :'$ref', :Errors
+            end
+          end
+        end
+      end
+
+      swagger_path 'v0/appeals/higher_level_review/intake_status/{intake_id}' do
+        operation :get do
```
kfrz: Should this be expanded?

app/controllers/v0/appeals_controller.rb
```diff
@@ -12,5 +12,21 @@ def index
...
+
+    def intake_status
+      intake_status = service.get_higher_level_reviews_intake_status params[:uuid]
+      render json: intake_status
+    end
+
+    private
+
+    def service
+      DecisionReview::Service.new
```
kfrz: Might we define the `DecisionReview` service namespaced under `Appeals` so as to share configuration? It connects to the same backend service (Caseflow) right?

camerontesterman: I think this suggestion makes sense. Wouldn't be a bad idea to raise this with Silvio in his PR, where the service integration work is defined. [PR 3439](https://github.com/department-of-veterans-affairs/vets-api/pull/3439)

camerontesterman:
> Is this still a WIP? If not ready for review please use the draft PR feature.

It isn't ready. Was unaware of that feature, but will use in the future - thanks for letting me know.

spec/request/appeals_request_spec.rb
camerontesterman: I converted this over to use `shared_context`, appreciate the suggestion. The splat wasn't necessary, it has been removed.

app/swagger/requests/appeals.rb
camerontesterman: This has been expanded.

spec/controllers/v0/higher_level_review_controller_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe HigherLevelReviewController, type: :controller do
+  describe
```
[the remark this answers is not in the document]
camerontesterman: Methods were added to appeals controller instead of separate higher level review controller, this file has been removed.

lib/decision_review/schemas/review.json
```diff
@@ -193,7 +193,7 @@
                         "enum": ["RequestIssue"]
                       },
                       "id": {
-                        "type": "string"
+                        "type": ["string", "integer"]
```
SMLuthi: Why was this changed to include `integer`s? The response from HLR designates this as only ever being a `string` var.

camerontesterman: Fixed

johnpaulashenfelter: @camerontesterman @SMLuthi I don't have the context of the actual endpoint, but it doesn't look like it _was_ fixed

camerontesterman: @johnpaulashenfelter ok that is weird.. look at this commit https://github.com/department-of-veterans-affairs/vets-api/pull/3426/commits/c8166bdb8c3b0c380bcd6f4b9250747a0f5ee56f

camerontesterman: I did fix it and push the change... I'm not sure why it isn't reflecting that.

johnpaulashenfelter: That is totally weird. I checked this branch locally and it is 100% fine.

modules/appeals_api/config/routes.rb
```diff
@@ -4,7 +4,13 @@
   match '/v0/*path', to: 'application#cors_preflight', via: [:options]
 
   namespace :v0, defaults: { format: 'json' } do
-    resources :appeals, only: [:index]
+    resources :appeals, only: [:index] do
+      collection do
+        get 'higher_level_reviews/:uuid', to: 'appeals#show_higher_level_review'
+        get 'intake_statuses/:intake_id', to: 'appeals#show_intake_status'
+      end
```
johnpaulashenfelter: I believe it would have been more RESTy to do something similar to

```
    resources :appeals, only: [:index] do
        resources 'higher_level_reviews/:uuid', only: :show
        resources 'intake_statuses/:intake_id',  only: :show
```

but talked to @camerontesterman and they decided against it earlier on. Seems like a big ask to change at this point.

modules/appeals_api/app/controllers/appeals_api/v0/appeals_controller.rb
```diff
@@ -20,6 +20,16 @@ def index
         )
       end
 
+      def show_higher_level_review
+        higher_level_review = review_service.get_higher_level_reviews(params[:uuid])
```
kfrz: This is an anti-pattern in conventional Rails controller. We should consider opening a separate controller that maps more appropriately to this resource - i.e. `/v0/higher_level_review` 

Perhaps y'all have an argument otherwise - but I don't see why this can't be it's own resource

modules/appeals_api/config/routes.rb
```diff
@@ -4,7 +4,13 @@
   match '/v0/*path', to: 'application#cors_preflight', via: [:options]
 
   namespace :v0, defaults: { format: 'json' } do
-    resources :appeals, only: [:index]
+    resources :appeals, only: [:index] do
+      collection do
+        get 'higher_level_reviews/:uuid', to: 'appeals#show_higher_level_review'
```
kfrz: See comment above about separate resources

kfrz:
:joy_cat: timing ! @SMLuthi could you please update this PR with a comment explaining your rationale for *not* adding a more RESTful resource route? Mostly for historical documentation purposes but also because I may not be seeing a compelling Railsy argument in favor of the current code

PR #3433, opened 2019-10-15
Moving remaining TODOs over to .rubocop.yml

thilton-oddball:
## Description of change
Final cleanup moving remaining TODOs over to .rubocop.yml


.rubocop.yml
```diff
@@ -200,3 +198,120 @@ RSpec/DescribeClass:
     - 'spec/lib/bb/client_spec.rb'
     - 'spec/lib/gi/client_spec.rb'
     - 'spec/lib/sm/client/folders_spec.rb'
+
+# Determined to be too benign and/or numerous to justify changing
```
johnpaulashenfelter: This is fine

.rubocop.yml
```diff
@@ -200,3 +198,120 @@ RSpec/DescribeClass:
...
+
+# This seems to generate a lot of false positives
+RSpec/EmptyExampleGroup:
+  Enabled: false
+
+# This seems to generate a lot of false positives
+RSpec/LetSetup:
+  Enabled: false
+
+# Might be worth revisiting at some point
```
johnpaulashenfelter: These should stay in the todo I think. We can copy the .rubocop.yml between files so we can share a config that's consistent (!!!) and including local repo info makes that harder. The other things *are* todo, just not right now ;)

thilton-oddball: Do you mean keep everything below line 258 in the todo?

annaswims: I was the one who suggest we move them to the .rubocop.yml .   The rubocop rspec rules are all new and at some point it just didn't seem worth the effort to fix all of the violations.  I don't feel that strongly, but it seemed like we might just grandfather these.  I thought we'd move them out of the todo, since we don't really intend to do them any time soon. 

I'm going to review the list myself to be a second opinion on whether they're worth the effort though.

PR #3438, opened 2019-10-17
create pssg download job

whuang85:
## Description of change
Creates the PSSG download job

https://github.com/department-of-veterans-affairs/vets-contrib/issues/3103

## Testing done
Created rspec for testing how it would interact with data coming from the drive time band client

#### Applies to all PRs

- [x] Appropriate logging


app/workers/facilities/pssg_download.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    def extract_name(attributes)
+      name = attributes.dig('Name')
+      name.partition(':').first.strip!
+    end
+
+    def extract_vha_id(attributes)
+      'vha_' + extract_name(attributes)
+    end
```
omgitsbillryan: You could get rid of `def extract_vha_id` and reduce the calls to `extract_name` from 2 down to 1 with something like this:
```ruby
name = extract_name(attributes)
vha_id = 'vha_' + name
#...
drive_time_band = facility.drivetime_bands.find_or_initialize_by(vha_facility_id: name)
```

whuang85: Thanks, I added some exception handling but not sure if it's pragmatically correct in ruby

config/sidekiq_scheduler.yml
```diff
@@ -104,6 +104,12 @@ FacilityLocationBulkUpdateVHA:
   args: ['vha']
   description: "Download and store facility location data"
 
+FacilityLocationVHADriveTime:
+  cron: "55 4 * * * America/New_York"
+  class: Facilities::PSSGDownload
+  args: ['vha']
+  description: "Download and store facility location data"
```
edmkitty: Update this to include "drive time band"

PR #3443, opened 2019-10-17
Page size validation for claims api submissions

ATeal:
## Description of change
To meet our end vendor's requirements we should validate so that we don't submit anything that can't be processed.

resolves https://github.com/department-of-veterans-affairs/vets-contrib/issues/3320

## Testing done
- local submissions

#### Applies to all PRs

- [x] Appropriate logging


modules/claims_api/app/controllers/claims_api/concerns/page_size_validation.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module ClaimsApi
+  module PageSizeValidation
+    extend ActiveSupport::Concern
+    included do
+      def validate_documents_page_size
+        render json: { errors: document_errors }, status: 422 unless document_errors.empty?
```
omgitsbillryan: I'm probably OOL - it used to be that you had to raise an exception to return a non-200 status code that would adhere to our error JSON format. I'm not / wasn't very fond of that way of doing things so this is still good to me.

PR #3445, opened 2019-10-17
Exclude some files from danger PR size check

omgitsbillryan:
## Description of change
Connects: https://github.com/department-of-veterans-affairs/va.gov-team/issues/2547

Excludes the following from the PR size check that `Danger-ruby` performs:
```
['Gemfile.lock', '.json', 'spec/fixtures/', '.txt']
```

## Why?
The danger bot is warning In PR's where large test files (`.json`, fixtures, `.txt`) are either added or deleted. These test files can be _thousands_ of LoC, and are not typically considered cumbersome to review.

## Testing done
Local testing

## Testing planned
None

#### Applies to all PRs

- [x] Appropriate logging


Dangerfile
```diff
@@ -1,6 +1,14 @@
 # Warn if a pull request is too big
 MAX_PR_SIZE = 250
-if git.lines_of_code > MAX_PR_SIZE
+EXCLUSIONS = ['Gemfile.lock', '.json', 'spec/fixtures/', '.txt']
```
jholton: Should we add `.yml` files?

omgitsbillryan: My first thought is no because things like `Settings.yml` & `docker-compose.yml` _are_ things I'd want to include. However, you made me realize that our VCR cassettes are YAML 😱 . I'll add an exception for our cassettes as a middle ground.

PR #3446, opened 2019-10-18
New API endpoint for EVSS findRatingInfoPID Service

kathleencrawford:
## Description of change
New API endpoint for the EVSS findRatingInfoPID service, which returns information about a veteran's rated disabilities and total combined disability rating.

The new endpoint is: "/v0/disability_compensation_form/find_rating_info_pid"

Original ticket here: https://github.com/department-of-veterans-affairs/va.gov-team/issues/2604

Betamocks recordings have been created and a PR has been submitted here: https://github.com/department-of-veterans-affairs/vets-api-mockdata/pull/107

## Testing done
- [x] Local 

## Testing planned
Staging and FE integration tests.

#### Unique to this PR

_N/A_

#### Applies to all PRs

- [ ] Appropriate logging


app/serializers/rating_info_serializer.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class RatingInfoSerializer < ActiveModel::Serializer
+  attribute :user_percent_of_disability
+
+  def id
+    nil
```
johnpaulashenfelter: The JSON output always has a null id? Is that because of the structure of JSON that's expected?

config/routes.rb
```diff
@@ -41,6 +41,7 @@
       get 'submission_status/:job_id', to: 'disability_compensation_forms#submission_status', as: 'submission_status'
       get 'user_submissions'
       get 'suggested_conditions'
+      post 'find_rating_info_pid'
```
johnpaulashenfelter: This whole resource is pretty non-RESTy, so makes sense that its just another url, but wondering if something more RESTy would work:

```
resource :disability_compensation_form, only: [] do
  resource :rating, only: :show
end
```

if that makes sense semantically.

lib/evss/disability_compensation_form/configuration_rating_info.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module EVSS
+  module DisabilityCompensationForm
+    # Configuration for the findRatingInfoPID service
+    #
+    class ConfigurationRatingInfo < EVSS::DisabilityCompensationForm::Configuration
+      API_VERSION = Settings.evss.versions.common
+
+      # :nocov:
```
johnpaulashenfelter: This should get executed when you've got real specs, right? Anything we can do to remove the coverage exemption would make this less worrisome

spec/request/rating_info_request_spec.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'rating info', type: :request do
+  include SchemaMatchers
+
+  let(:user) { build(:disabilities_compensation_user) }
+  let(:headers) { { 'CONTENT_TYPE' => 'application/json' } }
```
johnpaulashenfelter: Didn't expect you to need this -- but might be in all the code -- I'll take a look.

Maybe a more descriptive name? `let(:request_json_content)....`

lib/evss/disability_compensation_form/configuration_rating_info.rb
charleystran: I would echo JP's thought. If we have a :nocov we should have a comment explaining why. The next dev to see this might wrap other code in a :nocov without understanding the why

kathleencrawford: Thank you @johnpaulashenfelter and @charleystran, I meant to take that out.

app/serializers/rating_info_serializer.rb
johnpaulashenfelter: JSON::API requirement

config/routes.rb
```diff
@@ -41,6 +41,7 @@
       get 'submission_status/:job_id', to: 'disability_compensation_forms#submission_status', as: 'submission_status'
       get 'user_submissions'
       get 'suggested_conditions'
+      get 'find_rating_info_pid'
```
johnpaulashenfelter: i'd order these alphabetically if there's not any importance in their order

spec/request/rating_info_request_spec.rb
kathleencrawford: Thanks, you were right - I didn't need to post so was able to remove the headers

config/routes.rb
kathleencrawford: Done!

spec/request/rating_info_request_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    sign_in_as(user)
+  end
+
+  describe 'GET /v0/disability_compensation_form/find_rating_info_pid' do
+    context 'with a valid 200 evss response' do
+      it 'matches the rating info schema' do
+        VCR.use_cassette('evss/disability_compensation_form/find_rating_info_pid') do
+          get '/v0/disability_compensation_form/find_rating_info_pid'
+          expect(response).to have_http_status(:ok)
+          # expect(response).to match_response_schema('rating_info_response')
```
johnpaulashenfelter: Might want to remove this

spec/lib/evss/disability_compensation_form/service_rating_info_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+  let(:user) { build(:disabilities_compensation_user) }
+
+  describe '#get_rating_info' do
+    context 'with a valid evss response' do
+      it 'returns a rating info response object' do
+        VCR.use_cassette('evss/disability_compensation_form/find_rating_info_pid') do
+          response = subject.get_rating_info
+          expect(response).to be_ok
+          expect(response).to be_an EVSS::DisabilityCompensationForm::RatingInfoResponse
+          expect(response.user_percent_of_disability).to be_an String
```
johnpaulashenfelter: i think `be_a` is an alias so it reads easier

lib/evss/disability_compensation_form/configuration_rating_info.rb
```diff
@@ -0,0 +1,22 @@
...
+module EVSS
+  module DisabilityCompensationForm
+    # Configuration for the findRatingInfoPID service
+    #
+    class ConfigurationRatingInfo < EVSS::DisabilityCompensationForm::Configuration
+      API_VERSION = Settings.evss.versions.common
+
+      def base_path
+        "#{Settings.evss.url}/wss-common-services-web-#{API_VERSION}/rest/ratingInfoService/#{API_VERSION}"
+      end
```
omgitsbillryan: This configuration already lives in the [evss/common_service.rb](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/evss/common_service.rb#L8). Looks to me like you're really creating a new endpoint in an existing service rather than a brand new service altogether.

Can you add to or extend the `evss/common_service` and corresponding configuration rather than duplicate?

spec/request/rating_info_request_spec.rb
kathleencrawford: Thanks, done.

spec/lib/evss/disability_compensation_form/service_rating_info_spec.rb
kathleencrawford: Good catch!

app/swagger/schemas/form526/find_rating_info_pid.rb
```diff
@@ -0,0 +1,17 @@
...
+module Swagger
+  module Schemas
+    module Form526
+      class FindRatingInfoPid
+        include Swagger::Blocks
+
+        swagger_schema :RatingInfo do
+          property :disability_decision_type_name, type: :string, example: 'Service Connected'
+          property :service_connected_combined_degree, type: :string, example: '90'
+          property :user_percent_of_disability, type: :string, example: '90'
```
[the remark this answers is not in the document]
kathleencrawford: @omgitsbillryan I had the same conversation with one of our FE developers and thought it should be an integer as well.  I hesitated because if the user does not have a rating, "user_percent_of disability" gets passed as an empty string, instead of zero (same for "service_connected_combined_degree").  Let me know if you'd like to discuss more, as I'm not sure what I should do in this case.  Thanks!

lib/evss/disability_compensation_form/configuration_rating_info.rb
kathleencrawford: Thank you for pointing that out...I've gone ahead and updated this to use common_service instead of a new service.

spec/lib/evss/common_service_spec.rb
```diff
@@ -32,4 +32,17 @@
...
   end
+
+  describe '#get_rating_info' do
+    context 'with a valid evss response' do
+      it 'returns a rating info response object' do
+        VCR.use_cassette('evss/disability_compensation_form/find_rating_info_pid') do
+          response = subject.get_rating_info
+          expect(response).to be_ok
+          expect(response).to be_an EVSS::DisabilityCompensationForm::RatingInfoResponse
+          expect(response.user_percent_of_disability).to eq nil
```
johnpaulashenfelter: Does the factory user have a nil value for `user_percent_of_disability`?

Not sure you actually need this line since you already specified the response -- it seems like expecting a nil here _could_ lead to false positives if you get a nil for that value for other reasons.

spec/request/rating_info_request_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  let(:user) { build(:disabilities_compensation_user) }
+
+  before do
+    sign_in_as(user)
+  end
+
+  describe 'GET /v0/disability_compensation_form/find_rating_info_pid' do
+    context 'with a valid 200 evss response' do
+      it 'matches the rating info schema' do
```
johnpaulashenfelter: This isn't matching a schema -- is a line missing?

spec/request/swagger_spec.rb
```diff
@@ -484,6 +484,13 @@
           )
         end
       end
+
+      it 'supports getting rating info' do
+        expect(subject).to validate(:get, '/v0/disability_compensation_form/find_rating_info_pid', 401)
```
johnpaulashenfelter: Not sure you need to test this here -- seems like it should be a different spec if its important

spec/support/schemas/rating_info_response.json
```diff
@@ -0,0 +1,29 @@
...
+  "properties": {
+    "data": {
+      "properties": {
+        "attributes": {
+          "properties": {
+            "service_connected_combined_degree": {
+              "type": "string"
+            },
+            "user_percent_of_disability": {
+              "type": "string"
```
johnpaulashenfelter: should this match the type in the object?

kathleencrawford: Yes...thank you...good catch.  I updated the other files but forgot this one.

spec/request/swagger_spec.rb
kathleencrawford: I was doing this because it seems to be consistent with the rest of the spec.  I'm happy to remove it if you don't think it belongs here.  I didn't want to *not* add it and then have someone ask why its missing.

spec/request/rating_info_request_spec.rb
kathleencrawford: It doesn't match a schema because for the factory user, this value is null.

spec/lib/evss/common_service_spec.rb
kathleencrawford: Yes, the factory user does not have a rating so this value comes back as null.

config/settings.yml
```diff
@@ -144,6 +144,7 @@ evss:
   mock_ppiu: true
   mock_itf: true
   mock_disabilities_form: true
+  mock_rating_info: true
```
omgitsbillryan: Just a heads up that this will need to be overridden for staging and prod in the `devops` repo:

- https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/deployment/config/vets-api/staging-settings.local.yml.j2
- https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/deployment/config/vets-api/prod-settings.local.yml.j2

config/routes.rb
```diff
@@ -35,12 +35,13 @@
     end
 
     resource :disability_compensation_form, only: [] do
+      get 'find_rating_info_pid'
```
omgitsbillryan: @johnpaulashenfelter made an earlier comment about the RESTiness of this endpoint that I still agree with. All of your objects, including the model itself, is named `RatingInfo`. Thoughts on renaming as such? e.g. `GET /v0/disability_compensation_form/rating_info`

kathleencrawford: I agree that I don't love this name either @omgitsbillryan and @johnpaulashenfelter.  The name of the service we are calling from EVSS is findRatingInfoPID so I wasn't sure what are standards are and if I needed to keep it consistent.  I did shorten the name for the models/objects, etc because we typically don't use action verbs (get/find).  I am more than happy to rename this to something else.

PR #3447, opened 2019-10-18
update tests and data with sta_no

whuang85:
## Description of change
Use station number from PSSG for querying vha facility

https://github.com/department-of-veterans-affairs/vets-contrib/issues/3420

## Testing done
Updated unit tests

#### Applies to all PRs

- [x] Appropriate logging


app/workers/facilities/pssg_download.rb
```diff
@@ -37,8 +36,8 @@ def create_and_save_drive_time_data(drive_time_data)
     end
 
     def extract_id(attributes)
-      name = attributes&.dig('Name')
-      name.partition(':')&.first&.strip!
+      name = attributes&.dig('Sta_No')
+      name&.strip
```
edmkitty: I think you could put the `&.strip` at the end of the previous line, and I think we don't need to encapsulate this logic in a method anymore

PR #3448, opened 2019-10-18
correctly identify unique drive time bands

whuang85:
## Description of change
Correctly identify drive time bands by name instead of facility id so they don't override each other when saving multiple drive time bands of the same facility id but different drive times

closes https://github.com/department-of-veterans-affairs/vets-contrib/issues/3426

## Testing done
Unit tests updated

#### Applies to all PRs

- [x] Appropriate logging


app/workers/facilities/pssg_download.rb
```diff
@@ -22,16 +25,17 @@ def create_and_save_drive_time_data(drive_time_data)
...
         drive_time_band.min = attributes&.dig('FromBreak')
         drive_time_band.max = attributes&.dig('ToBreak')
-        drive_time_band.name = attributes&.dig('Name')
+        drive_time_band.name = name
         drive_time_band.polygon = extract_polygon(drive_time_data)
         drive_time_band.save
         facility.save
       rescue => e
-        logger.error e.message
+        raise PSSGDownloadError, e.message
```
edmkitty: I think this begin/rescue should be around the `download_data` line in the `perform` method.

spec/jobs/facilities/pssg_download_spec.rb
```diff
@@ -53,15 +53,15 @@
...
 
       allow(pssg_client_stub).to receive(:get_drivetime_bands).with(0, 1).and_return(drive_time_data_402)
       subject.perform
       expect(BaseFacility.find_facility_by_id('vha_648A4').drivetime_bands).not_to be_nil
       expect(BaseFacility.find_facility_by_id('vha_648A4').drivetime_bands.size).to be(1)
-      expect(BaseFacility.find_facility_by_id('vha_648A4').drivetime_bands[0].name).to eql('drive_time_band_^0')
+      expect(BaseFacility.find_facility_by_id('vha_648A4').drivetime_bands[0].name).to eql('648A4 : 0 - 10')
       expect(BaseFacility.find_facility_by_id('vha_648A4').drivetime_bands[0].unit).to eql('minutes')
-      expect(DrivetimeBand.find_by(vha_facility_id: '648A4').name).to eql('drive_time_band_^0')
+      expect(DrivetimeBand.find_by(vha_facility_id: '648A4').polygon.to_s).to eql(existing_drive_time.polygon.to_s)
```
edmkitty: the `eql` matcher is apparently meant for integers. I suggest using `eq` instead

edmkitty: You might be able to compare the polygon objects with the `eq` matcher instead of converting them to strings, too. Maybe. Not sure

PR #3450, opened 2019-10-18
Update Benefits Claims API Docs

AubreySuter:
## Description of change
Fleshing out the Benefits Claims API's documentation to be a bit more descriptive and more closely align with our new documentation for the Benefits Intake API. 

Tagging @ryanlink and @tristarowan for visibility - if either of you have feedback, I'd be interested in that as well!

#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging


modules/claims_api/app/swagger/claims_api/description/v0.md
```diff
@@ -1,11 +1,55 @@
...
+Internal VA consumers such as mail processing vendors can use an organizational token to establish disability compensation claims automatically.
 
-Use this API to submit a Veteran's benefits claim, and to request the status of a Veteran's benefits claim.
+Whereas the Benefits Intake API sends PDFs to be uploaded to the Centralized Mail Portal, the Benefits Claims API accepts more data, using it to automatically establish those claims so they're immediately accessible to Veteran Service Representatives in VBMS (Veterans Benefits Management System). This bypasses about 10 days of processing time.
 
-The Benefit Claim API passes data through to Electronic Veterans Self Service, EVSS.
+Currently, the API supports automatic establishment of: 
 
-EVSS uses VAAFI to authenticate requests.
+ - Application for Disability Compensation (21-526EZ) **(non-original claims only)**
```
ryanlink: Should we define "non-original claims" below, or is the term widely understood among our target audience?

modules/claims_api/app/swagger/claims_api/description/v1.md
```diff
@@ -1,18 +1,56 @@
...
+Internal VA consumers such as mail processing vendors can use an organizational token to establish disability compensation claims automatically.
 
-Use this API to submit a Veteran's benefits claim, and to request the status of a Veteran's benefits claim.
+Whereas the Benefits Intake API sends PDFs to be uploaded to the Centralized Mail Portal, the Benefits Claims API accepts more data, using it to automatically establish those claims so they're immediately accessible to Veteran Service Representatives in VBMS (Veterans Benefits Management System). This bypasses about 10 days of processing time.
 
-The Benefit Claim API passes data through to Electronic Veterans Self Service, EVSS.
+Currently, the API supports automatic establishment of: 
 
-EVSS uses VAAFI to authenticate requests.
+ - Application for Disability Compensation (21-526EZ) **(non-original claims only)**
```
ryanlink: Should we define "non-original claims" below, or is the term widely understood among our target audience?

modules/claims_api/app/swagger/claims_api/description/v0.md
AubreySuter: Good point. In my experience our consumers are familiar with these terms, but it would be good to document our definition (which we did put some effort into nailing down).

Unless you disagree with this approach, I'd like to propose we merge this as-is, and add that definition within the next few weeks, when we add a "Original Claims" section to describe the signature path.

AubreySuter: Actually... the non-original claims piece applies only to v1 (which is also where we'll be adding the signature path). Internal consumers using v0 can submit 526 claims regardless of whether they're original or not. Updating v0 to remove the disclaimer.

PR #3451, opened 2019-10-18
Remove type filter from nearby facilities

nfasulo:
## Description of change
When /nearby was originally set up it was basically a copy and paste of the request and response from the existly facilities endpoint. It had used the Bing Isochrone API to determine the drivetime to a facility for a veteran, but we are now in the process of switching to using a PSSG API to determine these drivetimes. This dataset only has VHA facilities which is fine for consumers as this endpoint is only being used to determine community care eligibility for the MISSION Act, but means that the inclusion of a request param that allows the consumer to request a particular facility type no longer makes sense.

This PR adds the value `health` to the `type` param in the NearbyController for each request. This is necessary for two reasons: so that all request are of only VHA facilities and so that we can validate the `services[]` param to against existing health services.

Resolves department-of-veterans-affairs/vets-contrib/issues/3352.

## Testing done
Updated nearby request tests to expect only VHA facilities to be returned.

#### Unique to this PR
- [x] Only VHA facilities are returned by /nearby.
#### Applies to all PRs

- [x] Appropriate logging


app/models/nearby_facility.rb
```diff
@@ -72,9 +72,7 @@ def get_facilities_in_isochrone(params, isochrone_response)
         # find all facilities that lie inside of the polygon
         conditions = "ST_Intersects(#{make_polygon}, ST_MakePoint(long, lat))"
         facilities_query_base_instance = FacilitiesQuery::Base.new(params)
-        BaseFacility::TYPES.flat_map do |facility_type|
-          facilities_query_base_instance.get_facility_data(conditions, params[:type], facility_type, params[:services])
-        end
+        Array.new facilities_query_base_instance.get_facility_data(conditions, params[:type], params[:type], params[:services])
```
edmkitty: I think you can do a `.to_a` at the end of the line here instead of using Array.new. Not sure

edmkitty: Although, why are we making this an array instead of a relation? Does it matter?

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
```diff
@@ -51,6 +52,10 @@ def set_default_format
         request.format = :json if params[:format].nil? && request.headers['HTTP_ACCEPT'].nil?
       end
 
+      def set_facility_type
+        params[:type] = 'health'
```
edmkitty: what do you think about making this name consistent with the types in BaseFacility, for example: "va_health_facility"

modules/va_facilities/spec/requests/nearby_request_spec.rb
```diff
@@ -209,7 +223,7 @@ def query_params(url)
     end
 
     it 'paginates according to parameters' do
-      setup_pdx
+      create_list(:generic_vha, 10)
```
edmkitty: Fancy! I didn't know `create_list` was a thing!

nfasulo: I saw `create_list` being used to generate an arbitrary number of VBA facilities and knew I had to steal the pattern.

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
```diff
@@ -51,6 +52,10 @@ def set_default_format
         request.format = :json if params[:format].nil? && request.headers['HTTP_ACCEPT'].nil?
       end
 
+      def set_facility_type
```
rtravitz: Do we need to run this as a before action if this endpoint is only ever going to return health facilities? Would it be more clear to pass in `'health'` to Bing/future Postgres query and not mess with params at all?

modules/va_facilities/README.v1.yml
```diff
@@ -434,15 +434,14 @@ paths:
     get:
       tags:
         - facilities
-      summary: Query facilities based on drive time from a specified location
+      summary: Query health facilities based on drive time from a specified location
```
rtravitz: Does `VHA` versus `health` make any semantic difference here?

modules/va_facilities/spec/requests/nearby_request_spec.rb
```diff
@@ -175,6 +175,20 @@ def query_params(url)
         expect(health_service['wait_times']['effective_date']).to eq('2018-03-05')
       end
     end
+
+    it 'only queries health facilities ignoring type paramater' do
```
rtravitz: Do we specify to supply a type parameter of health?

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
nfasulo: I'm not a big fan of setting the param this way, but we need to validate the `services[]` param and need to know the facility type to do that. Setting the `type` param allows us to continue to utilize `validate_type_and_services_known` for this validation.

modules/va_facilities/README.v1.yml
nfasulo: They are synonymous in my reading. I used "health" because that is what they were called in the sentence I modified in the next paragraph, but if VHA is more explicit I don't mind changing it.

modules/va_facilities/spec/requests/nearby_request_spec.rb
nfasulo: I removed the reference to the `type` parameter from the documentation. This test was just to verify that the response would be unaffected if a consumer supplied the parameter anyway.

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
rtravitz: Makes sense 👍

PR #3454, opened 2019-10-21
Add evss claims async to rack attack throttling

edmkitty:
Rapid polling of the claims status endpoint is causing multiple issues. This PR hopes to slow down the polling by adding throttling to rack attack middleware.

https://dsva.slack.com/archives/C30LCU8S3/p1571673172048900

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -24,7 +24,7 @@ class Rack::Attack
     req.ip if req.path == '/v0/vic/vic_submissions' && req.post?
   end
 
-  throttle('evss_claims_async', limit: 10, period: 1.minute) do |req|
+  throttle('evss_claims_async', limit: 10, period: 60) do |req|
     req.ip if req.path == '/v0/evss_claims_async'
   end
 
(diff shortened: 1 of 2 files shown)


config/initializers/rack_attack.rb
```diff
@@ -24,6 +24,10 @@ class Rack::Attack
     req.ip if req.path == '/v0/vic/vic_submissions' && req.post?
   end
 
+  throttle('evss_claims_async', limit: 10, period: 1.minute) do |req|
```
jholton: So, best case scenario only 10 users can hit this route per minute (which is polled 12 times per minute per user by the frontend)?
Seems low to me.

edmkitty: I thought this meant that one ip address can hit the endpoint only 10 times per minute.

jholton: Ah, you're right.  That's what the line below does.  Perhaps we should raise it to `12` to match the frontend polling.

annaswims:
`1.minute` is part of rails https://api.rubyonrails.org/classes/Numeric.html

PR #3455, opened 2019-10-21
Update evss policy

edmkitty:
Does not continue execution if the user does not exist.

### DIFF
--- a/app/policies/evss_policy.rb
+++ b/app/policies/evss_policy.rb
@@ -16,7 +16,7 @@ EVSSPolicy = Struct.new(:user, :evss) do
   end
 
   def form526_attrs?
-    user.present? && user.birls_id.present? && evss_attrs?
+    evss_attrs? && user.birls_id.present?
   end
 
   def log_success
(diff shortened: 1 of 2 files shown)


app/policies/evss_policy.rb
```diff
@@ -12,20 +12,20 @@ def access_form526?
   private
 
   def evss_attrs?
-    user.edipi.present? && user.ssn.present? && user.participant_id.present?
+    user.present? && user.edipi.present? && user.ssn.present? && user.participant_id.present?
   end
 
   def form526_attrs?
-    user.birls_id.present? && evss_attrs?
+    user.present? && user.birls_id.present? && evss_attrs?
```
jholton: So as to not duplicate the `user.present?` check:
```suggestion
    evss_attrs? && user.birls_id.present?
```

edmkitty: Oops, good call!

edmkitty:
Learned that the real problem behind this is Sidekiq jobs being in the queue for too long, which causes the user to expire in redis. Would rather see an error in this case than a Forbidden. Closing this PR without merging.

PR #3456, opened 2019-10-21
Kr/add serializer tests

edmkitty:
Add tests for serializers in va_facilities.
Added a fix for an hours issue in the csv serializer.

https://github.com/department-of-veterans-affairs/vets-contrib/issues/3331


lib/facilities/vha_facility.rb
```diff
@@ -39,6 +40,17 @@ def add_mental_health(facilities)
         end
       end
 
+      def format_phone_numbers(facilities)
```
edmkitty: This extension issue is present in other facility types and I haven't changed those files. I'm not a big fan of this fix, since it iterates through the entire facilities set again. What do you guys think of me combining the `add_mental_health`, `format_phone_numbers`, and `add_websites` methods? Or more specifically, moving their guts into methods and pulling out the map so we only iterate once?

rtravitz: It seems like phone numbers would be handled the same way regardless of facility type. Would it be better to have phone number formatting on `BaseFacility` in that case? If that's what you're suggesting, then 👍

rtravitz: And why does this iterate through the entire facilities set again? Would it be simpler to change the phone numbers before they're stored in the `base_facilities` table as each facility is being read in from GIS/ArcGIS?

modules/va_facilities/lib/va_facilities/csv_serializer.rb
```diff
@@ -26,7 +26,7 @@ def self.headers
         mailing_city mailing_state mailing_zip
         phone_main phone_fax phone_mental_health_clinic phone_pharmacy phone_after_hours
         phone_patient_advocate phone_enrollment_coordinator
-        hours_monday hours_tuesay hours_wednesday hours_thursday hours_friday
+        hours_monday hours_tuesday hours_wednesday hours_thursday hours_friday
```
rtravitz: D'oh. Nice find 👍

modules/va_facilities/lib/va_facilities/csv_serializer.rb
```diff
@@ -57,7 +57,11 @@ def self.phone_attrs(object)
 
     def self.hours_attrs(object)
       hours = object.hours
-      %w[monday tuesday wednesday thursday friday saturday sunday].map { |day| hours[day] }
+      if object.facility_type_prefix == 'vc'
+        %w[monday tuesday wednesday thursday friday saturday sunday].map { |day| hours[day] }
```
rtravitz: Would it be simpler if we modified VetCenter days as they're being read in from ArcGIS so that we no longer need this type of condition in our system? I bet there are other instances of exceptions for different day names around, too.

spec/serializers/va_facilities/geo_serializer_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    it 'makes a geometry object' do
+      geo_object = VaFacilities::GeoSerializer.geometry(vha_facility)
+
+      expect(geo_object.keys).to eq(%w[type coordinates])
+      expect(geo_object.values).to eq(['Point', [vha_facility.long, vha_facility.lat]])
+    end
+  end
+
+  context '::properties' do
+    it 'collects basic facility attributes' do
```
rtravitz: Does this farm out to the facility serializer already tested above for each of these properties?

edmkitty: No, the geo_serializer has a `properties` method which calls these attributes on the object being serialized.

modules/va_facilities/lib/va_facilities/csv_serializer.rb
edmkitty: I would like this, but I haven't looked into the haul necessary to make that change.

edmkitty: I will make a tech debt ticket. I think it shouldn't be too bad of a change

lib/facilities/vha_facility.rb
edmkitty: Moving the phone numbers formatting to the parser

spec/serializers/va_facilities/geo_serializer_spec.rb
nfasulo: It would be nice to have the same testing of the values of these properties for the GeoSerializer as the FacilitySerializer has.

rtravitz: Even though they use the same ApiSerializer module right now, that might not always be the case. It would be nice to explicitly make those assertions for Geo, too, so that we could change the implementation, if needed.

PR #3459, opened 2019-10-22
1816 facilities/ppms service refactor

LindseySaari:
## Description of change
This change addresses refactoring of the PPMS facilities service. To follow the current vets-api response pattern, the response class and name space refactor was based on existing coding practices within vets-api.

## Testing done
Test suite passing 

- [x] Specs passing 
- [x] response data comes from VCR cassettes


jholton:
It looks like `Facilities::PPMS::Response` and corresponding `spec` are new, and everything else is just renamed.
Is that correct?

LindseySaari:
> It looks like `Facilities::PPMS::Response` and corresponding `spec` are new, and everything else is just renamed.
> Is that correct?

@jholton Yes, correct! The response class and spec are both new, everything else is just namespacing to follow existing patterns

spec/lib/facilities/ppms/responses/response_spec.rb
```diff
@@ -0,0 +1,259 @@
...
+        expect(provider_services_response).not_to be(nil)
+        expect(provider_services_response.body).not_to be(nil)
+        expect(provider_services_response.status).to eq(200)
+      end
+
+      it 'has the proper provider specialties identifiers' do
+        expect(provider_specialties_response).not_to be(nil)
+        expect(provider_specialties_response.body).not_to be(nil)
+        expect(provider_specialties_response.status).to eq(200)
+      end
```
jholton: These are all just testing that the initialize.  That is, the `body` and `status` are just set as the arguments provided to `new`. 

I'd say we just need one of these that looks something like this:
```
    it 'has the proper response object attributes' do
        expect(response).not_to be(nil)
        expect(response.body).to eq(response_body)
        expect(response.get_body).to eq(response_body)
        expect(response.status).to eq(200)
      end
```
I'd also like to see specs that test the other methods: `self.from_provider_locator` and `self.map_provider_list`.

I see above you've declared `let(:provider_locator_response)`, but `provider_locator_response` isn't used anywhere.

spec/lib/facilities/ppms/responses/client_spec.rb
```diff
@@ -2,11 +2,11 @@
```
jholton: Should this file path be `spec/lib/facilities/ppms/client_spec.rb`

spec/lib/facilities/ppms/responses/response_spec.rb
```diff
@@ -0,0 +1,259 @@
+# frozen_string_literal: true
```
jholton: Should this file path be `spec/lib/facilities/ppms/response_spec.rb`

spec/lib/facilities/ppms/responses/response_spec.rb
```diff
@@ -0,0 +1,259 @@
...
+        'SpecialtyCode' => '101YA0400X',
+        'Name' => 'Counselor - Addiction (Substance Use Disorder) ',
+        'Grouping' => 'Behavioral Health & Social Service Providers',
+        'Classification' => 'Counselor',
+        'Specialization' => 'Addiction (Substance Use Disorder)', 'SpecialtyDescription' => 'Definition to come...'
+      }
+    ]
+  end
+
+  let(:response_body) do
```
johnpaulashenfelter: Is there a reason this isn't a file fixture?

spec/lib/facilities/ppms/responses/response_spec.rb
LindseySaari: Tests and file paths updated ✔️

spec/lib/facilities/ppms/responses/response_spec.rb
LindseySaari: Good idea @johnpaulashenfelter, moving to fixtures!

spec/lib/facilities/ppms/response_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+        'ProviderPrimaryCare' => false,
+        'QualityRanking' => 0,
+        'ProviderIdentifier' => '1114172319',
+        'Latitude' => 39.16996,
+        'Longitude' => -78.18052
+      }
+    ]
+  end
+
+  let(:response_body) { File.read('spec/fixtures/facilities/response_body.rb') }
```
jholton: It looks like this fixture data comes directly from a VCR cassette.
These are unit tests of a class used to parse a response (i.e. a request to the external API isn't made here), but what do we think about just using the data from the VCR cassette directly?
That way if the VCR cassette is updated, and the results change it'll let us know we probably need to revisit this response object class too.
```suggestion
  let(:response_body) { JSON.parse(YAML.load_file('spec/support/vcr_cassettes/facilities/va/ppms.yml')['http_interactions'][3]['response']['body']['string'])['value'][0] }
```
Same would apply for the `provider_locator_body` fixture.

Interested to hear opinions. @johnpaulashenfelter

LindseySaari: @jholton, yes I like that idea better! Reduces repeated code and will get updated when the vcr cassette is updated as well.

PR #3460, opened 2019-10-22
Update facilities v0 docs

jeff2d2:
## Description of change
Fixes https://github.com/department-of-veterans-affairs/vets-contrib/issues/3447.
This PR updates some of the language in the Facilities v0 API docs. It also removes the ugly `server` section that doesn't really work as intended and is replaced by a more clear presentation of the API's base URLs.

## Testing done
Tested locally

## Testing planned
None

- [x] `Design` copy updated
- [x] `server` section removed
- [x] `Base URLs` section added

#### Unique to this PR
None

#### Applies to all PRs

- [x] Appropriate logging


modules/va_facilities/README.yml
```diff
@@ -19,9 +19,15 @@ info:
...
+    ### Base URLs
+    
+    The base URLs for the VA Facilities API in the various environments are:
+    - Development: `https://dev-api.va.gov/services/va_facilities/v0`
+    - Production: `https://api.va.gov/services/va_facilities/v0`
+
     ### Authorization
 
-    API requests are authorized through a symmetric API token, provided in an HTTP header with name "apikey".
+    API requests are authorized through a symmetric API token, provided in an HTTP header with name `apikey`.
```
johnpaulashenfelter: I'd love to know how to get that API key -- this is something that's I could pull from the config file? `credstash`? I followed along as someone without any background on this and its the only thing that I was a bit unclear about.

Clearly you just added formatting, but if you can clarify that would be helpful to me at least

Eallan919:
@edmkitty to review

modules/va_facilities/README.yml
edmkitty: These docs are meant for users of the API, not developers of the API. An API key can be requested through this link: https://developer.va.gov/apply However, this only gets you access to dev. Access to the prod api is a different process. It would be nice to add the `apply` link here for clarity

modules/va_facilities/README.yml
```diff
@@ -64,22 +66,22 @@ info:
...
-    description: VA.gov API staging environment
-    variables:
-      version:
-        default: v0
-  - url: https://api.va.gov/services/va_facilities/{version}
-    description: VA.gov API production environment
-    variables:
-      version:
-        default: v0
+# servers:
```
edmkitty: Is there a reason not to simply delete this section? Will we be using it later?

jeff2d2: Solid point. I can remove it unless there's a reason to keep it around. It would only be useful again should we move to a new renderer other than swagger-ui. But I looked through the others team's docs and they've just removed.

edmkitty: Ok, good to know. Would you like me to remove it here and in the v1 readme?

jeff2d2: I took care of it. Thanks!

PR #3467, opened 2019-10-24
Va forms module

charleystran:
## Description of change
This pull request adds the VA Forms module and api for when forms are changed on the VA side

## Testing done
rspec and local testing

- [ ] REST endpoints exist for both the index and show actions. 

#### Applies to all PRs

- [ ] Appropriate logging
- [ ] https://github.com/department-of-veterans-affairs/vets-contrib/issues/3185


charleystran:
FYI this is not ready for review. It was marked ready for review on accident

db/migrate/20191023192329_create_versions.rb
```diff
@@ -0,0 +1,26 @@
+# This migration creates the `versions` table, the only schema PT requires.
+# All other migrations PT provides are optional.
+class CreateVersions < ActiveRecord::Migration[5.2]
```
bastosmichael: This migration should be in the module and not at the App level like we do with all of the other Projects and model should also be labeled as va_forms_versions or VAForms::Versions

db/migrate/20191023192329_create_versions.rb
```diff
@@ -0,0 +1,26 @@
...
+  # so that MySQL will use `longtext` instead of `text`.  Otherwise,
+  # when serializing very large objects, `text` might not be big enough.
+  TEXT_BYTES = 1_073_741_823
+
+  def change
+    create_table :versions do |t|
+      t.string   :item_type, {:null=>false}
+      t.integer  :item_id,   null: false, limit: 8
+      t.string   :event,     null: false
+      t.string   :whodunnit
```
bastosmichael: Not sure I understand this cell, maybe more descriptive?

modules/va_forms/app/controllers/va_forms/application_controller.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+module VaForms
+  class ApplicationController < ::OpenidApplicationController
```
bastosmichael: We don't need to inherit from OpenIDApplicationController unless we were doing an Oauth API end point, I would instead look at what VBA Documents inherits from and based it from there.

modules/va_forms/app/controllers/va_forms/v0/forms_controller.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module VaForms
+  module V0
+    class FormsController < ApplicationController
+      skip_before_action(:authenticate)
+
+      def index
+        forms = Form.all
```
bastosmichael: I'm guessing we are going to replace the .all with a search query that defaults to all like we talked about or is that for a different PR?

modules/va_forms/app/models/va_forms/form.rb
```diff
@@ -0,0 +1,86 @@
...
+    validates :title, presence: true
+    validates :form_name, presence: true, uniqueness: true
+    validates :url, presence: true
+    validates :sha256, presence: true
+
+    before_save :set_revision
+
+    BASE_URL = 'https://www.va.gov'
+
+    def self.refresh!(current_page = 0)
```
bastosmichael: This code feels like it would be better off living in a Sidekiq worker instead of in the model itself, keep the model relatively clean and move this process either into a worker or service if you can.

modules/va_forms/app/swagger/va_forms/form_model_swagger.rb
```diff
@@ -0,0 +1,83 @@
+# frozen_string_literal: true
+
+module VaForms
+  class FormModelSwagger
+    include Swagger::Blocks
+
+    swagger_schema :FormsIndex do
```
bastosmichael: Feels like this should just be the forms model since it won't change between the Index and Show

modules/va_forms/app/swagger/va_forms/forms_v0_controller_swagger.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+module ClaimsApi
+  class ClaimsV0ControllerSwagger
+    include Swagger::Blocks
```
bastosmichael: I don't see any responses here

modules/va_forms/config/routes.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+VaForms::Engine.routes.draw do
+  match '/v0/*path', to: 'application#cors_preflight', via: [:options]
+  match '/v1/*path', to: 'application#cors_preflight', via: [:options]
```
bastosmichael: you don't have any v1 so this should not be needed.

modules/va_forms/lib/tasks/va_forms.rake
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+namespace :va_forms do
+  task fetch_latest: :environment do
+    VaForms::Form.refresh!
```
bastosmichael: Since we have to run this once a day would make more sense to make the whole thing a sidekiq job that runs daily and just call the job here to refresh the model.

db/migrate/20191023192329_create_versions.rb
charleystran: As we chatted on friday, this is the default Papertrail migration, but I added a comment noting that.

modules/va_forms/app/controllers/va_forms/v0/forms_controller.rb
charleystran: My thought was that we would add that as another PR after we get a little feedback from users

modules/va_forms/app/swagger/va_forms/form_model_swagger.rb
charleystran: They actually do change as the show includes the versions. I Added a new block for the show that includes versions.

modules/va_forms/app/swagger/va_forms/forms_v0_controller_swagger.rb
charleystran: I added the responses

modules/va_forms/app/models/va_forms/form.rb
charleystran: Code has been moved into a sidekick worker

modules/va_forms/lib/tasks/va_forms.rake
charleystran: moved to sidekick worker and added entry to sidekick_scheduler

PR #3471, opened 2019-10-25
Validating 526 for no claims

bastosmichael:
## Description of change
Adding validation for ticket https://github.com/department-of-veterans-affairs/vets-contrib/issues/3434

## Testing done
- Updated current request tests

## Testing planned
- Testing different users in dev-api.va.gov

- Validate that a user has claims before allowing them to submit otherwise requiring that they disable pdf generation.

#### Applies to all PRs

- [ ] Appropriate logging


modules/claims_api/app/services/claims_api/unsynchronized_evss_claims_service.rb
```diff
@@ -25,6 +25,19 @@ def all
...
 
+    def count
+      raw_claims = client.all_claims.body
+      EVSS_CLAIM_KEYS.each_with_object([]) do |key, claim_accum|
+        next unless raw_claims[key]
+
+        claim_accum << raw_claims[key].map do |raw_claim|
+          raw_claim['id']
+        end
+      end.flatten.count
```
ATeal: instead of this you should just be able to do something like:
```
EVSS_CLAIM_KEYS.reduce(0) { |sum, key| sum + raw_claims[key].count }
```

modules/claims_api/app/services/claims_api/unsynchronized_evss_claims_service.rb
```diff
@@ -25,6 +25,19 @@ def all
       end.flatten
     end
 
+    def count
```
ATeal: It may be implied, but I think maybe renaming this to `claim_count` might make more sense. Feels weird to have `service.count`  -- personal preference though up to you

bastosmichael: Sounds good, I'll make the change.

modules/claims_api/app/services/claims_api/unsynchronized_evss_claims_service.rb
bastosmichael: implemented this instead 
```
EVSS_CLAIM_KEYS.reduce(0) do |sum, key| 
        next unless raw_claims[key]
        sum + raw_claims[key].count
      end
```

modules/claims_api/app/services/claims_api/unsynchronized_evss_claims_service.rb
bastosmichael: Done

PR #3472, opened 2019-10-25
More bug fixes for claim swagger

ATeal:
Swagger 2 doesn't really support bearer tokens natively like oas3


modules/claims_api/app/swagger/claims_api/claims_v1_swagger.rb
```diff
@@ -54,41 +54,10 @@ class ClaimsV1Swagger
...
-          }
-        }
-      ]
-
       security_definition :bearer_token do
-        key :type, :http
-        key :name, :bearer_token
-        key :scheme, :bearer
-        key :bearerFormat, :JWT
+        key :type, :apiKey
```
bastosmichael: You've probably explained this before but why are we listing an apiKey in a Claims V1 bearer_token instead of an actual bearer token?

PR #3474, opened 2019-10-25
move faraday client configuration spec set up into shared helper

LindseySaari:
## Description of change
To avoid module namespacing collisions and excessive setup/teardown, faraday configuration is now extracted into a shared helper for specs.

## Testing done
Spec tests pass

- [x] Spec tests pass
- [x] Specs identified where we're using a configuration, and extraction pattern applied

### DIFF
--- a/spec/lib/common/client/base_spec.rb
+++ b/spec/lib/common/client/base_spec.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require 'rails_helper'
-require './spec/support/default_configuration_helper'
 
 describe Common::Client::Base do
   module Specs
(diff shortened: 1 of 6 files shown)


spec/lib/common/client/base_spec.rb
```diff
@@ -1,16 +1,15 @@
 # frozen_string_literal: true
 
 require 'rails_helper'
+require './spec/support/default_configuration_helper'
```
omgitsbillryan: Thoughts on moving this into [`rails_helper.rb`](https://github.com/department-of-veterans-affairs/vets-api/blob/master/spec/rails_helper.rb#L9-L21)?

PR #3475, opened 2019-10-25
10795 Add Profile Show Notifications Feature Flag

elwoodva:
## Description of change

The VA.gov site administrator wants to be able to control who sees the Notifications checkbox on the profile page so that it can be turned off or on or only shown to a percentage of Veterans.

## Testing done

Local testing.

## Testing Planned

Test in Staging before introducing to production.

- [ ] When the feature flag is enabled, all qualifying Veterans will be able to see the checkbox below the Mobile Phone number on the profile page and on the edit modal for the mobile phone number on the profile page.
- [ ] When the feature flag is disabled, no Veterans will be able to see the checkbox below the Mobile Phone Number anywhere.


elwoodva:
@johnpaulashenfelter @LindseySaari Yes, I only updated one file.  The other files appeared here after it said i was behind and needed to update.  I pushed the button and those appeared.  We checked the changes out and there were updates being committed for them elsewhere/elsewhen.  So I wasn't sure what to do with them.

elwoodva:
@johnpaulashenfelter I pushed the update button again.  I don't see any differences.  Those 3 other changes are still here.  Do you want to help me figure this out in this PR or should I just wipe out my branch and start over with a different PR?  I don't understand why this has happened at all.  Nor do I see a way here to mark the change you requested here as resolved.

johnpaulashenfelter:
@elwoodva it looks like you manually reverted the change in master in your commit 93525f8239987f5bd8482e453e428835b3e25819 that... reverted that change from master ❗️ 

Might be easiest to rebase this branch and drop that commit since you're reverting something that should be in master. You can also manually squash the branch as part of that and get things cleaner if you wanted.

PR #3476, opened 2019-10-28
Vaos appointments controller

saneshark:
## Description of change
This PR is based on this issue: https://github.com/department-of-veterans-affairs/va.gov-team/issues/1582
 
1. updates the appointments service to introduce pagination
2. introduces JSON schema for cc_appointments and va_appointments
3. introduces controller logic with various arguments
4. introduces a new Flipper feature built into the policy
5. adds spec to test schema matches
6. introduces swagger documentation

## Testing done
Request specs using previously recorded VCR cassettes

## Testing planned
Manual regression on staging

#### Unique to this PR

- [x]  Introduces a new Flipper feature at `va_online_scheduling`

#### Applies to all PRs

- [x] Appropriate logging


kreek:
@kam for swagger docs looks like other modules do it like this https://github.com/department-of-veterans-affairs/vets-api/blob/d9bc8158b8d67ef65e90395affed5147b2db3dca/modules/claims_api/config/routes.rb#L50-L57

modules/vaos/app/services/vaos/appointment_service.rb
```diff
@@ -36,15 +36,78 @@ def get_cc_appointments(user, start_date, end_date)
 
     private
 
-    def get_va_appointments_url(icn, start_date, end_date)
-      "/appointments/v1/patients/#{icn}/appointments"\
-          "?startDate=#{date_format(start_date)}&endDate=#{date_format(end_date)}&useCache=false&pageSize=0"
+    def sanitize_va_appointments(json_hash)
```
kreek: Might be unnecessary future proofing but do we want to keep the service returning everything and have the serializer sanitize what goes to the front end?

saneshark: hmmm, thats not a bad idea tbh -- the parsing doesn't really read very well. I'm open to that, but this works for now. We need to get moving haha

saneshark:
I have absolutely no idea why these specs keep failing on Jenkins. Everything is fine locally.

modules/vaos/app/controllers/vaos/application_controller.rb
```diff
@@ -2,5 +2,16 @@
 
 module VAOS
   class ApplicationController < ::ApplicationController
+    before_action :authorize
```
kreek: Do we need the custom authorize method, how about `before_action { authorize :vaos, :access? }`

modules/vaos/app/serializers/vaos/va_appointments_serializer.rb
```diff
@@ -0,0 +1,31 @@
...
+      Digest::MD5.hexdigest(object.start_date)
+    end
+
+    set_type :va_appointments
+
+    attributes :start_date, :clinic_id, :clinic_friendly_name, :facility_id, :community_care
+
+    attribute :vds_appointments do |object|
+      Array.wrap(object&.vds_appointments).map do |vds|
+        vds.except(:patient_id) # remove patient identifiers
```
kreek: Would be more verbose but what about extracting a method called `remove_patient_identifiers` instead of adding comments to describe what's happening?

modules/vaos/app/services/vaos/appointment_service.rb
```diff
@@ -36,15 +24,42 @@ def get_cc_appointments(user, start_date, end_date)
...
+        pagination: {
+          current_page: pagination_params[:page] || 0,
+          per_page: pagination_params[:per_page] || 0,
+          total_pages: 0, # underlying api doesn't provide this; how do you build a pagination UI without it?
+          total_entries: 0 # underlying api doesn't provide this.
+        }
+      }
+    end
+
+    def get_appointments_url(user, type, start_date, end_date, pagination_params)
```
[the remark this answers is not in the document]
saneshark: +1 for the last one

.rubocop.yml
```diff
@@ -1,4 +1,4 @@
-require:
```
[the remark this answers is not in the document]
saneshark: I'm thinking it might be someone else's editor. Removing the additional whitespace makes sense.

modules/vaos/app/services/vaos/appointment_service.rb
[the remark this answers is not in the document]
saneshark: Might consider using return self for setting the user / type though -- to bypass code climate complaining about method arguments exceeding 4.

modules/vaos/app/controllers/vaos/application_controller.rb
saneshark: on second thought -- I rather like raising a custom exception.

modules/vaos/app/serializers/vaos/va_appointments_serializer.rb
saneshark: opted to keep this one the same for now too, it reads fine. If there was more going on would definitely extract into methods.

modules/vaos/app/controllers/vaos/v0/appointments_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+          appointment_service.get_appointments(type, start_date, end_date, pagination_params)
+      end
+
+      def each_serializer
+        "VAOS::#{params[:type].upcase}AppointmentsSerializer".constantize
+      end
+
+      def validate_params
+        raise Common::Exceptions::ParameterMissing, 'type' if params[:type].blank?
+        raise Common::Exceptions::InvalidFieldValue.new('type', params[:type]) unless %w[va cc].include?(params[:type])
```
jholton: Nitpick: Should we make use of your `type` method here instead of using `params[:type]`?

modules/vaos/spec/requests/appointments_request_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+        get '/services/vaos/v0/appointments'
+        expect(response).to have_http_status(:forbidden)
+        expect(JSON.parse(response.body)['errors'].first['detail'])
+          .to eq('You do not have access to online scheduling')
+      end
+    end
+
+    context 'without a type' do
+      it 'has a parameter missing exception' do
+        get '/services/vaos/v0/appointments'
```
jholton: I suggest being explicit with your missing query string parameter here.
That way in the future, if the `validate_params` lines get reordered or another exception case is added before the `type` check, this spec won't mysteriously start failing.
```suggestion
        get '/services/vaos/v0/appointments', params: params.except(:type)
```

saneshark:
desk_flip

modules/vaos/app/swagger/vaos/requests/appointments.rb
```diff
@@ -0,0 +1,87 @@
...
+          parameter do
+            key :name, :type
+            key :in, :query
+            key :required, true
+            key :type, :string
+            key :enum, %w[va cc]
+            key :description, <<~TYPE
+              Type of appointment:
+                * `va` - VA Appointments
+                * `cc` - ComunityCares Appointments
```
sjbrush: Typo in Community in case you care.

modules/vaos/app/swagger/vaos/schemas/va_appointments.rb
```diff
@@ -0,0 +1,51 @@
...
+              property :appointment_length, type: %i[string null]
+              property :appointment_time, type: :string, format: :datetime
+              property :clinic, type: :object do
+                property :name, type: :string
+                property :ask_for_check_in, type: :boolean
+                property :facility_code, type: :string
+              end
+              property :type, type: :string
+              property :current_status, type: :string
+              property :booking_note, %i[string null]
```
jholton: This might be your Jenkins problem:
```suggestion
              property :booking_note, type: %i[string null]
```

annaswims: can reproduce locally with `CI=true bundle exec rake spec`. The difference is eager loading.

saneshark: what a bizarre error for something as innocuous as swagger-blocks. Thank you!

annaswims:
I've changed the way feature toggle admins authenticate.  If you'd like to be an admin, create an id.me verified user in staging and prod and add your email address here: https://github.com/department-of-veterans-affairs/vets-api/blob/master/config/settings.yml#L397

modules/vaos/app/controllers/vaos/appointments_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+      Date.parse(params[:start_date])
+    rescue ArgumentError
+      raise Common::Exceptions::InvalidFieldValue.new('start_date', params[:start_date])
+    end
+
+    def end_date
+      Date.parse(params[:end_date])
+    rescue ArgumentError
+      raise Common::Exceptions::InvalidFieldValue.new('end_date', params[:end_date])
+    end
```
omgitsbillryan: These along with 
```ruby
raise Common::Exceptions::InvalidFieldValue.new('type', type) unless %w[va cc].include?(type)
```
look like model validations, in which case, you could then have more concise controller logic with something like:
```ruby
raise InvalidFieldValue(<model>.errors.messages) unless <model>.valid?
```

PR #3479, opened 2019-10-28
Rt kr/pssg job updates

edmkitty:
This PR adds updates to the PSSG Download Job that should prevent errors and speed up overall download time. 

https://github.com/department-of-veterans-affairs/vets-contrib/issues/3448

### DIFF
--- a/app/workers/facilities/pssg_download.rb
+++ b/app/workers/facilities/pssg_download.rb
@@ -40,16 +40,17 @@ module Facilities
     end
 
     def round_band(band)
-      if band%10 == 0 
+      if (band % 10).zero?
         band
       else
-        (band.to_i/10)*10
+        (band.to_i / 10) * 10
       end
     end
 
     def extract_polygon(drive_time_data)
       rings = drive_time_data&.dig('geometry', 'rings')
       return if rings.blank?
+
       geojson = "{\"type\":\"Polygon\",\"coordinates\":#{rings}}"
       RGeo::GeoJSON.decode(geojson)
     end
@@ -60,7 +61,7 @@ module Facilities
         response = @drivetime_band_client.get_drivetime_bands(offset, 30)
 
         break if response.blank?
-        
+
         response.each(&method(:create_and_save_drive_time_data))
         offset += 30
       end
--- a/app/workers/facilities/pssg_download.rb
+++ b/app/workers/facilities/pssg_download.rb
@@ -15,6 +15,7 @@ module Facilities
       begin
         download_data
       rescue => e
+        puts "Big fail #{e.message}"
         raise PSSGDownloadError, e.message
       end
     end
@@ -34,9 +35,14 @@ module Facilities
       drive_time_band.min = attributes&.dig('FromBreak')
       drive_time_band.max = attributes&.dig('ToBreak')
       drive_time_band.name = name
-      drive_time_band.polygon = extract_polygon(drive_time_data)
-      drive_time_band.save
-      facility.save
+      begin
+        drive_time_band.polygon = extract_polygon(drive_time_data)
+        drive_time_band.save
+        facility.save
+        puts "Success: #{name}"
+      rescue RGeo::Error::InvalidGeometry => e
+        puts "Error: #{name} #{e}"
+      end
     end
 
     def extract_polygon(drive_time_data)
--- a/app/workers/facilities/pssg_download.rb
+++ b/app/workers/facilities/pssg_download.rb
@@ -45,7 +45,7 @@ module Facilities
       if (band % 10).zero?
         band
       else
-        (band.to_i / 10) * 10
+        (band.to_i / 10 + 1) * 10
       end
     end
 
--- a/app/workers/facilities/pssg_download.rb
+++ b/app/workers/facilities/pssg_download.rb
@@ -4,7 +4,6 @@ require 'rgeo/geo_json'
 require 'sentry_logging'
 
 module Facilities
-
   class PSSGDownload
     include Sidekiq::Worker
     include SentryLogging
--- a/app/workers/facilities/pssg_download.rb
+++ b/app/workers/facilities/pssg_download.rb
@@ -15,7 +15,6 @@ module Facilities
       begin
         download_data
       rescue => e
-        puts "Big fail #{e.message}"
         raise PSSGDownloadError, e.message
       end
     end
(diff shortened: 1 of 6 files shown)


app/workers/facilities/pssg_download.rb
```diff
@@ -56,18 +50,20 @@ def round_band(band)
...
       loop do
-        response = @drivetime_band_client.get_drivetime_bands(offset, 1)
-        break if response.blank?
+        response = @drivetime_band_client.get_drivetime_bands(offset, 30)
 
+        break if response.blank?
+        
         response.each(&method(:create_and_save_drive_time_data))
-        offset += 1
+        offset += 30
```
LindseySaari: Just curious, how does an offset of 30 effect the drivetime band data?

app/workers/facilities/pssg_download.rb
```diff
@@ -31,28 +31,39 @@ def create_and_save_drive_time_data(drive_time_data)
...
+      if (band % 10).zero?
+        band
+      else
+        (band.to_i / 10) * 10
+      end
+    end
+
     def extract_polygon(drive_time_data)
       rings = drive_time_data&.dig('geometry', 'rings')
+      return if rings.blank?
```
whuang85: don't you want to check if this is blank in the `create_and_save_drive_time_data` so you can skip the whole building up of `drive_time_band` and even the `facility.save` and `drive_time_band.save`?

app/workers/facilities/pssg_download.rb
edmkitty: By requesting a chunk of 30 bands instead of requesting a single band at a time, we significantly speed up our overall download time. The reason we could not download a bigger chunk than 30 is because the PSSG server has a limit of 35 mb per request. We estimate that chunk sizes of 30 will not go over this limit.

jeff2d2: 2 possibly dumb questions, obviously coming from a non-engineer:
1. It may be overkill, but might it be good to have another job that runs separate of this one, to analyze the data and dynamically set the offset so that it grabs the biggest chunk possible without going over 35 mb?
1. What happens if the data changes and 30 becomes too big (i.e. >35 mb)? Would this lower it and try again or just fail?

edmkitty: Not dumb questions at all! 

1 is overkill. That would require a lot of extra engineering time and work. The simplest solution would be to choose the chunk size that is small enough to never go over the limit, but big enough that the download doesn't take forever. We could, perhaps, put in retries with an exponential backoff on the offset size or something like that, but I am fairly confident that 30 will work just fine. 

2. If there is a request that is too big and results in an error, then an error will be raised here: https://github.com/department-of-veterans-affairs/vets-api/pull/3479/files#diff-cb075e0b92af9cbd6df776e0cffe7e62R17 
and then that error will be rescued here: 
https://github.com/department-of-veterans-affairs/vets-api/pull/3479/files#diff-6ed8e216d7afc245ca81a1f3f468305eL17

And all that means is, if an error occurs, it will be filed away in Sentry and then that particular piece of code will be skipped. The loop continues and moves onto the next chunk. It will mean that 30 bands do not get downloaded, but we will get an error in sentry informing us of the event and then we can manually go in and decrease the offset size if necessary.

app/workers/facilities/pssg_download.rb
```diff
@@ -31,28 +32,37 @@ def create_and_save_drive_time_data(drive_time_data)
...
       offset = 0
       loop do
-        response = @drivetime_band_client.get_drivetime_bands(offset, 1)
+        response = @drivetime_band_client.get_drivetime_bands(offset, 30)
+
         break if response.blank?
 
         response.each(&method(:create_and_save_drive_time_data))
-        offset += 1
+        offset += 30
```
nfasulo: Can we get the offset size pulled out into a constant? It is used in a couple of places and would be bad if we only updated one.

config/initializers/rgeo.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+RGeo::ActiveRecord::SpatialFactoryStore.instance.tap do |config|
+  config.register(RGeo::Geographic.spherical_factory(srid: 4326, uses_lenient_assertions: true),
```
rtravitz: I think a comment here explaining why this initializer—especially the `uses_lenient_assertions` part—is important would help us or someone else in the future.

lib/facilities/drivetime_band_client.rb
```diff
@@ -11,6 +14,15 @@ def get_drivetime_bands(offset, limit)
       params = build_params(offset, limit)
       response = perform(:get, '/arcgis2/rest/services/Portal/MonthlyVAST_TTB/FeatureServer/0/query', params)
       JSON.parse(response.body)['features']
```
nfasulo: Can this line be deleted?

app/workers/facilities/pssg_download.rb
```diff
@@ -15,14 +13,17 @@ def perform
       begin
         download_data
       rescue => e
-        raise PSSGDownloadError, e.message
+        log_exception_to_sentry(e)
```
rtravitz: Will this job ever retry after failure in Sidekiq if we rescue all the exceptions?

edmkitty: I doubt it, unless sidekiq itself has some kind of error running the job

edmkitty: I think avoiding retries is best for now. We just need to keep a close eye on sentry.

app/workers/facilities/pssg_download.rb
```diff
@@ -31,28 +32,37 @@ def create_and_save_drive_time_data(drive_time_data)
...
       facility.save
     end
 
-    def extract_polygon(drive_time_data)
-      rings = drive_time_data&.dig('geometry', 'rings')
+    def round_band(band)
+      if (band % 10).zero?
+        band
+      else
+        (band.to_i / 10) * 10
```
nfasulo:
```suggestion
        (band.to_i / 10 + 1) * 10
```
I think we need to round up to the next 10 minute increment. Otherwise we are misrepresenting the amount of time it might take someone. For instance rounding 15 minutes down to 10 means that we are saying that anyone in the drivetime band should be able to get to the facility within 10 minutes, which isn't true.

PR #3485, opened 2019-10-29
break ci out into jenkins stages

omgitsbillryan:
## Description of change
Breaks out all of the various stages we have in our `make ci` into their own jenkins stage.

### Why?
- Provides more insight into build times
- Stages may be "parallelized" in the future
- Provides faster feedback for failed builds in PRs
  - Rather than drilling into jenkins console output to see why the `Run Tests` stage failed, you will quickly see the name of the stage which failed

image

### Potential Drawbacks
Some overhead time incurred by reloading the ruby environment for each docker execution. I think that since CI is `RAILS_ENV=test`, this overhead is on the order of seconds.

#### Applies to all PRs

- [ ] Appropriate logging


Makefile
```diff
@@ -19,14 +19,30 @@ ci:
...
 
-.PHONY: ci-spec
-ci-spec:
-	@$(BASH_TEST) "bin/rspec ${SPEC_PATH}"
-
 .PHONY: bash
 bash:
 	@$(COMPOSE_DEV) $(BASH)
 
+.PHONEY: ci-build
```
edmangimelli:
```suggestion
.PHONY: ci-build
```

edmangimelli: i see the same spelling (PHONEY) for `ci-down` too --or are they supposed to be like that?

omgitsbillryan: whoops, nope. good eye 👍

omgitsbillryan:
@nathanhruby I totally agree about the stable `ci` target. Perhaps I should simply remove it from make... In my changes to the `ci` **rake** task, I discovered that a number of developers are using this to verify their local changes. So I think that so long as the rake task is there, the make target can be deleted.

Keep an eye out for a separate PR for this.

Makefile
nathanhruby: That is most certainly my fault, thank you for catching it!

Makefile
```diff
@@ -15,18 +15,34 @@ default: ci
 ci:
 	@$(BASH_TEST) "bin/rails db:setup db:migrate ci"
 
-.PHONEY: ci-down
+.PHONY: ci-down
```
kfrz: :joy_cat:

PR #3488, opened 2019-10-29
Extract error handling and transformation logic

kfrz:
## Description of change
This PR attempts to address some of the complexities in `ApplicationController`, related to rescuing and coercing error classes. The purpose of transforming an error class is so that `vets-api` can exercise finer-grained control over the specific responses the client consumes. 

E.g., if the `ApplicationController` rescues an `ActionController::ParameterMissing` exception, we expect it to transform into a `Common::Exceptions::ParameterMissing` object, which inherits from `Common::Exceptions::BaseError`. 

`Common::Exceptions::BaseError` is used throughout `vets-api` as a parent class which provides support for custom messaging via `i18n` keys in `exceptions.en.yml`. This key-value mapping allows developers to normalize the messaging sent when disparate backend service errors are encountered, as the consistency in quality of "original" messages is low.

### Notable changes
- Use default Raven configuration for skipping exception types
- Move `SentryLogging` context logic into  module
- Move error transformation/coercion logic into a separate class 
- Add a common ErrorHandler to handle error coercion logic

## Testing done
- [ ] Spec suite ran via Jenkins CI
- [ ] Unit tests written for `ErrorHandler`

## Testing planned
- Will test on staging when PR is approved

#### Unique to this PR
- [ ] Error coercion logic is encapsulated into its own class
- [ ] Sentry logging and tagging is extracted into `SentryLogging` module

#### Applies to all PRs

- [x] Appropriate logging
- [x] ~Swagger docs have been updated, if applicable~
- [x] ~Provide which alerts would indicate a problem with this functionality (if applicable)~


app/controllers/concerns/ignore_not_found.rb
```diff
@@ -2,6 +2,6 @@
 
 module IgnoreNotFound
   def skip_sentry_exception_types
-    ApplicationController::SKIP_SENTRY_EXCEPTION_TYPES + [Common::Exceptions::RecordNotFound]
+    SentryLogging::SKIP_SENTRY_EXCEPTION_TYPES + [Common::Exceptions::RecordNotFound]
```
johnpaulashenfelter: Sentry has a built-in way to do this: https://docs.sentry.io/clients/ruby/config/ -- maybe it didn't when all this code was originally written? Seems like it would be much easier to move this to Sentry config

lib/sentry_logging.rb
```diff
@@ -1,6 +1,74 @@
...
+  # rubocop:disable Metrics/CyclomaticComplexity
+  # rubocop:disable Metrics/PerceivedComplexity
+  def log_error(error)
+    if skip_sentry_exception_types.include?(error.class)
+      Rails.logger.error error.message.to_s, backtrace: error.backtrace
+    else
+      extra = error.respond_to?(:errors) ? { errors: error.errors.map(&:to_hash) } : {}
+      if error.is_a?(Common::Exceptions::BackendServiceException)
+        # Add additional user specific context to the logs
+        if @current_user.present?
```
johnpaulashenfelter: Should this be a responsibility of the `Common::Exceptions::BackendServiceException`?

lib/sentry_logging.rb
```diff
@@ -1,6 +1,74 @@
...
+    Common::Exceptions::RoutingError,
+    Common::Exceptions::Forbidden,
+    Breakers::OutageException
+  ].freeze
+
+  # rubocop:disable Metrics/CyclomaticComplexity
+  # rubocop:disable Metrics/PerceivedComplexity
+  def log_error(error)
+    if skip_sentry_exception_types.include?(error.class)
+      Rails.logger.error error.message.to_s, backtrace: error.backtrace
```
johnpaulashenfelter: Doesn't this happen automatically in the log? This is basically "log it if we don't sentry it"

app/controllers/concerns/ignore_not_found.rb
LindseySaari: We could possibly move the exception types defined in the SKIP_SENTRY_EXCEPTION_TYPES constant to the config, but it looks like the method gets custom defined in other places as well

lib/sentry_logging.rb
jholton: The error passed to this method has been `rescue`d, so it wouldn't otherwise appear in the `Rails` log. Right?

lib/sentry_logging.rb
jholton: `Common::Exceptions::BackendServiceException` is raised in a number of places, and is a parent of class of a few of other error classes , many/most of which don't have access to the `@current_user` object.
That all raises the following question for me:  Should this code exist here, or in a library explicitly intended for use in `controller`s (or just refactored in the `ApplicationController`?  This library is used/included in controllers, AR models, Redis models, service classes, Sidekiq workers, etc, and most/all of this explicitly deals with `controller` concerns (request objects, current_user objects, etc)

kfrz: Yes, I think so - the logic here is one of the most confusing aspects of errors across the platform and I think extracting the context building to the caller as Sentry doesn't (or shouldn't) care/know about any context specifics it just logs hashes of data

kfrz: I think if we were to move this responsibility into the `::BackendServiceException` class it would be obfuscated, we may want more explicitness around the adding of user context info to Sentry

lib/sentry_logging.rb
LindseySaari: Really specific comment, but we can probably use the rails_logger method on line 15 instead of Rails.logger.error

app/controllers/concerns/ignore_not_found.rb
kfrz: This [issue](https://github.com/department-of-veterans-affairs/va.gov-team/issues/912) is in the backlog for updating how we tag services. Let's talk about it when the tools team is reformed.

kfrz: We could do the exclusion configuration like so, which would look like:

```rb
module IgnoreNotFound
  def skip_sentry_exception_types
    Raven.configure { |c| c.excluded_exceptions += 'Common::Exceptions::RecordNotFound' }
  end
end
```

This to me is a bit less readable, and intertwines the logic for " what errors don't we want to log to APM " logic into our " which APM system are we using " logic, whereas if we were to swap to a different service (i.e. DataDog), we'd have to move this logic to a different configuration. Alternatively, we could do something more abstract:

```rb
module IgnoreNotFound
  def skip_sentry_exception_types
    ErrorLogging::SKIP_EXCEPTION_TYPES += [Common::Exceptions::RecordNotFound]
  end
end
```

That seems a bit further out of scope, and addresses a problem we don't currently have.

@johnpaulashenfelter @jholton @LindseySaari thoughts?

jholton:
@kfrz Can you fill out the PR description please?

app/controllers/application_controller.rb
```diff
@@ -61,6 +61,57 @@ def set_app_info_headers
...
+
+  def render_error(exc)
+    render json: { errors: exc.errors }, status: exc.status
+  end
+
+  # @note Later rescue_from statements take precedence and so we must take care to respect order
+  rescue_from StandardError do |error|
+    exc = Common::Exceptions::InternalServerError.new(error)
+    handle_error(exc)
+    render json: { errors: exc.errors }, status: exc.status_code
```
LindseySaari: Could we extend the render_error method to include json rendering for the exc.status_code case (lines 76 and 82 of this PR)?

spec/lib/common/client/error_handler_spec.rb
```diff
@@ -16,69 +15,74 @@
...
     end
   end
 
   describe '#transformed_error' do
-  
     context 'when given an error without a specific counterpart' do
       subject { described_class.new(StandardError.new).transformed_error }
-      it 'returns the correct default error' do
+
+      xit 'returns the correct default error' do
```
LindseySaari: Ah, these tests might not be applicable any longer after the rescue block breakout in the application controller

omgitsbillryan:
Are you still looking for reviews on this PR as is? This PR is quite large and seems stale.

kfrz:
@omgitsbillryan we'll likely re-request reviews Monday or Tuesday, after updating w/ `master` and ensuring build is green. We can try to break it out, it's a bit large and Danger is raising issues

app/controllers/application_controller.rb
```diff
@@ -39,116 +39,81 @@ def routing_error
...
 
-  def user_context
-    {
-      uuid: current_user&.uuid,
-      authn_context: current_user&.authn_context,
-      loa: current_user&.loa,
-      mhv_icn: current_user&.mhv_icn
-    }
+  def handle_error(error)
+    Common::Client::ErrorHandler.handle(error)
```
kfrz: `Common::Client::ErrorHandler` would indicate the errors are originating in a client, though that's a separate subset of errors

app/controllers/application_controller.rb
```diff
@@ -39,116 +39,81 @@ def routing_error
...
-        tags[:sign_in_method] = current_user.identity.sign_in[:service_name]
-        # account_type is filtered by sentry, becasue in other contexts it refers to a bank account type
-        tags[:sign_in_acct_type] = current_user.identity.sign_in[:account_type]
-      else
-        tags[:sign_in_method] = 'not-signed-in'
-      end
-    end
+  def render_error(exc)
+    render json: { errors: exc.errors }, status: exc.status
   end
```
kfrz: Signature here isn't great - use "exception" or something more descriptive than `exc` and perhaps add some comments

app/controllers/application_controller.rb
```diff
@@ -39,116 +39,81 @@ def routing_error
...
+    handle_error(exc)
+    render_error(exc)
+  end
+
+  rescue_from Pundit::NotAuthorizedError do |error|
+    exc = Common::Exceptions::Forbidden.new(cause: error, detail: 'User does not have access to the requested resource')
+    handle_error(exc)
+    render_error(exc)
+  end
+
```
kfrz: In this case, we're actually reorganizing the control flow such that instead of rescuing *all* exceptions, including `ruby` exceptions, we're now rescuing from specific cases, and any `StandardErrors` will be converted into an `InternalServerError`.

Also on line 79, explain why `exc` instead of `error` is used

app/controllers/concerns/ignore_not_found.rb
```diff
@@ -2,6 +2,6 @@
 
 module IgnoreNotFound
   def skip_sentry_exception_types
-    ApplicationController::SKIP_SENTRY_EXCEPTION_TYPES + [Common::Exceptions::RecordNotFound]
+    Raven.configure { |c| c.excluded_exceptions += 'Common::Exceptions::RecordNotFound' }
```
kfrz: This could be a separate pr, including the error abstraction work

app/controllers/v0/health_care_applications_controller.rb
```diff
@@ -5,6 +5,8 @@ class HealthCareApplicationsController < ApplicationController
     FORM_ID = '1010ez'
 
     skip_before_action(:authenticate)
+    before_action(:append_to_skip_sentry_exception_types)
+    after_action(:remove_from_skip_sentry_exception_types)
```
kfrz: I'm not sure this is a great pattern, it would likely be overlooked and cause headaches for devs. Let's rethink our approach here.

config/initializers/raven.rb
```diff
@@ -13,4 +13,13 @@
...
   }
+
+  # appends to { Raven::Configuration::IGNORE_DEFAULT }
+  # @see https://docs.sentry.io/clients/ruby/config/
+  config.excluded_exceptions += %w[
+    Common::Exceptions::Unauthorized
+    Common::Exceptions::RoutingError
+    Common::Exceptions::Forbidden
+    Breakers::OutageException
+  ]
```
kfrz: Try to extract to separate PR

lib/common/client/error_handler.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Common
+  module Client
+    # This class is responsible for providing a public interface with which to perform common error handling tasks.
```
kfrz: Rename to `Common::Error::Handler` or something not tied to "client"

lib/common/client/errors.rb
```diff
@@ -10,22 +10,21 @@ class ClientError < Error
         attr_accessor :status
         attr_accessor :body
 
-        def initialize(message = nil, status = nil, body = nil)
+        def initialize(error = nil, message: nil, status: nil, body: nil)
```
kfrz: Add documentation to this class

Also add/update docs in `va.gov-team` repo/README.md

lib/common/client/errors.rb
```diff
@@ -10,22 +10,21 @@ class ClientError < Error
...
           super(message)
+          @cause = error
```
kfrz: This class inherits from `StandardError`, so why do we have to *add* the cause attribute? 

It should be [built-in](https://ruby-doc.org/core-2.6/Exception.html) - add specs/documentation/comments to explain why this is necessary.

lib/common/exceptions/internal/forbidden.rb
```diff
@@ -5,6 +5,7 @@ module Exceptions
     # Forbidden - We may eventually want different variations on this with distinct MinorCodes
     class Forbidden < BaseError
       def initialize(options = {})
+        @cause  = options[:cause]
```
kfrz: This should be on the base error class

spec/lib/common/client/error_handler_spec.rb
```diff
@@ -0,0 +1,88 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe Common::Client::ErrorHandler do
+  let(:gateway_timeout_error) { StandardError.new }
```
kfrz: a-z

spec/lib/common/client/error_handler_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+  let(:unknown_format_error) { ActionController::UnknownFormat }
+  let(:not_safe_host_error) { Common::Exceptions::NotASafeHostError }
+  let(:base_error) { Common::Exceptions::BaseError }
+  let(:breakers_outage_error) { Breakers::OutageException }
+  let(:client_error) { Common::Client::Errors::ClientError }
+
+  describe '#log_error' do
+    subject { described_class.new(client_error.new) }
+
+    xit 'calls instance log error method' do
```
kfrz: Don't skip me!

spec/lib/saml/settings_service_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 require 'rails_helper'
 require 'saml/settings_service'
-require 'lib/sentry_logging_spec_helper'
+require 'lib/shared_examples_for_sentry_logging'
```
kfrz: Easy to extract to separate PR

spec/request/health_care_applications_request_spec.rb
```diff
@@ -296,6 +296,12 @@
 
           it 'renders error message' do
             expect(Raven).to receive(:capture_exception).with(error, level: 'error').once
+            # This test is failing because app/models/health_care_application.rb:38 is raising a
```
kfrz: :crying_cat_face:

spec/request/user_request_spec.rb
```diff
@@ -181,6 +181,7 @@
     end
 
     it 'MVI success should only make a request to MVI one time per multiple requests!' do
+      pending 'after error handling changes this is calling MVI session class methods multiple times'
```
kfrz: :crying_cat_face:

spec/spec_helper.rb
```diff
@@ -23,6 +23,7 @@
     track_files '**/{app,lib}/**/*.rb'
 
     add_filter 'app/controllers/concerns/accountable.rb'
+    add_filter 'app/controllers/concerns/ignore_not_found.rb' # @todo remove this before merging
```
kfrz: Do this

kfrz:
Closing this PR as it's complex and has been stale for awhile. 

Will backlog tasks and re-prioritize in the future.

PR #3489, opened 2019-10-29
[1771] Higher Level Review submission endpoint

camerontesterman:
## Description of change
Adds a submission endpoint to create a higher level review. [1771](https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/1771)

## Testing done
Tests have been ran using the the VCR cassettes that were recorded as part of the decision review service work.

Submission endpoint allows a user to submit a higher level review via the endpoint.

#### Applies to all PRs

- [x] Appropriate logging


app/swagger/requests/appeals.rb
```diff
@@ -60,6 +60,81 @@ class Appeals
         end
       end
 
+      swagger_path '/services/appeals/v0/appeals/higher_level_reviews' do
```
annaswims: When I  [view the swagger docs locally](https://github.com/department-of-veterans-affairs/vets-api/tree/master/app/swagger), I'd expect to see a json schema. Screen Shot 2019-12-31 at 12 38 14 PM

config/locales/exceptions.en.yml
```diff
@@ -1685,7 +1691,7 @@ en:
         title: Duplicate Intake In Progress
         detail: Another duplicate intake is already in progress
         code: DR_duplicate_intake_in_progress
-        status: 422
+        status: 409
```
omgitsbillryan: This was an existing error response, but I don't see that any of the existing swagger or specs were changed accordingly - what gives? Were we not testing and/or swaggering it in the existing API's before this PR?

modules/appeals_api/spec/requests/appeals_request_spec.rb
```diff
@@ -51,6 +51,63 @@
...
+            'data' => {
+              'type' => 'HigherLevelReview',
+              'attributes' => {
+                'receiptDate' => '2019-07-10',
+                'informalConference' => true,
+                'sameOffice' => false,
+                'legacyOptInApproved' => true,
+                'benefitType' => 'compensation',
+                'veteran' => {
+                  'fileNumberOrSsn' => '123456789',
```
omgitsbillryan: Stylistically, not a fan of `fileNumberOrSsn` - are we allowed to modify this API?

camerontesterman: We are just passing the data that the decision review API expects. You'd have to discuss that with either @erluti or @edmangimelli

app/swagger/schemas/appeals.rb
```diff
@@ -1,258 +0,0 @@
-# frozen_string_literal: true
-
-module Swagger
-  module Schemas
-    class Appeals
-      include Swagger::Blocks
-
-      swagger_schema :Appeals do
```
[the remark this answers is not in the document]
camerontesterman: No, it's because the `data` array within the Appeals swagger schema object didn't have an items block. I went ahead and fixed it.

config/locales/exceptions.en.yml
camerontesterman: It wasn't in the swagger docs prior to this PR. Would've been introduced when the service was initially integrated.

PR #3491, opened 2019-10-29
Add server section back to the Facilities API v0 docs now that curl builder needs it

jeff2d2:
## Description of change
Add the `server` section back to the Facilities API v0 docs. This section is needed by the curl builder and will error if not present.

- [ ] `server` section added back as it was

#### Applies to all PRs

- [x] Appropriate logging

### DIFF
--- a/modules/va_facilities/README.yml
+++ b/modules/va_facilities/README.yml
@@ -66,6 +66,22 @@ info:
 tags:
   - name: facilities
     description: VA Facilities API
+servers:
+- url: https://dev-api.va.gov/services/va_facilities/{version}
+description: VA.gov API development environment
+variables:
+  version:
+    default: v0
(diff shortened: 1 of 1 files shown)


modules/va_facilities/README.yml
```diff
@@ -66,6 +66,22 @@ info:
...
+- url: https://dev-api.va.gov/services/va_facilities/{version}
+  description: VA.gov API development environment
+  variables:
+    version:
+      default: v0
+- url: https://staging-api.va.gov/services/va_facilities/{version}
+  description: VA.gov API staging environment
+  variables:
+    version:
+      default: v0
```
dvogel: Let's take staging out of this. We don't really want people using that environment anymore.

PR #3498, opened 2019-10-30
re-add authn_context to user profile

annaswims:
## Description of change
The direct deposit team has decided that DS Logon LOA3 is not as secure as ID.me LOA3 so we'll need to provide authentication context to the front end again so that they can appropriately layer verifications

#### Applies to all PRs

- [x] Appropriate logging


annaswims:
I also updated documeantion for what the `autn_context` means. https://github.com/department-of-veterans-affairs/vets.gov-team/blob/master/Products/Identity/Login/IDme/201810_IDme_Whitelabeled_Flow/%5B%2313806%5D%20Ensure%20that%20the%20whitelabled%20ID.me%20identity%20proofing%20flow%20appears%20for%20DS%20Logon%20and%20MHV%20users.md

PR #3499, opened 2019-10-31
Create v2 of eMIS payment service to get pay grade history data

rtravitz:
## Description of change

This PR finishes [#3118](https://github.com/department-of-veterans-affairs/vets-contrib/issues/3118). To support USAJobs as a consumer of the veteran data APIs, we want to provide pay grade information. The `payGradeHistory` action is only available in eMIS v2, so a new `PaymentServiceV2` is added here that mirrors the structure of other eMIS services and configurations.

Other notable changes:
- The settings file is changed to differentiate xml namespaces between v1 and v2 PaymentService clients. The v1 list is not as extensive as the v2 list because many of the defaults in `edipi_or_icn_message.rb` are already correct for v1. The extras included in the v2 list get merged and override those defaults of the same namespace.
- The payment service urls in `settings.yml` are similarly split into v1 and v2

## Testing done

- Added specs for the service and model along with fixtures recorded from eMIS based on a test user

#### Unique to this PR

 - [x] Create client that can retrieve information from the getPayGradeHistory endoint in eMIS v2

#### Applies to all PRs

- [x] Appropriate logging


lib/emis/responses.rb
```diff
@@ -15,6 +15,7 @@
 require_dependency 'emis/responses/get_separation_pay_response'
 require_dependency 'emis/responses/get_unit_information_response'
 require_dependency 'emis/responses/get_veteran_status_response'
+require_dependency 'emis/responses/get_pay_grade_history_response'
```
lihanli: requires are not needed since we are using autoloading now

spec/lib/emis/payment_service_v2_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+describe EMIS::PaymentServiceV2 do
+  describe 'get_pay_grade_history' do
+    let(:edipi) { '1007697216' }
+
+    context 'with a valid request' do
+      it 'calls the get_pay_grade_history endpoint with a proper emis message' do
+        VCR.use_cassette('emis/get_pay_grade_history/success') do
+          response = subject.get_pay_grade_history(edipi: edipi)
+          expect(response).to be_ok
```
lihanli: might be helpful to check if values in the response body are correct

spec/lib/emis/payment_service_v2_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+require 'rails_helper'
+require 'emis/payment_service'
+
+describe EMIS::PaymentServiceV2 do
+  describe 'get_pay_grade_history' do
+    let(:edipi) { '1007697216' }
+
+    context 'with a valid request' do
+      it 'calls the get_pay_grade_history endpoint with a proper emis message' do
+        VCR.use_cassette('emis/get_pay_grade_history/success') do
```
lihanli: when i write specs with VCR i usually add `{ match_requests_on: %i[method uri headers body] }`, otherwise VCR does not check if the request headers or body match what is in the cassette. we've already run into bugs with that before when we introduced a bug that made us send the wrong request body but all the vcr specs still passed since it wasn't checking the bodies. this makes writing specs more complicated though since you usually need to stub out a few things like uuids and current time to match what's in the cassette.

rtravitz:
Should have addressed all the listed comments in the most recent commit.

PR #3504, opened 2019-11-01
improve unauthenticated experience for feature toggles

annaswims:
rather than returning a 403 for users who are not able to toggle features provide a simple read-only view.  Only get requests are allowed for those who are not logged in or feature toggle admins.

Screen Shot 2019-11-06 at 11 28 23 AM
@U-DON How hard would it be to add a sign in link?


omgitsbillryan:
With this PR, unauthenticated users might be stuck wondering why they are unable to toggle, whereas before they'd get hit with a 403. They lose that feedback - even if it wasn't very pretty.

annaswims:
Good point @omgitsbillryan I added some help text to the screen.

Also, to address @jholton's point. we chatted and the reason that I [took out the test](https://github.com/department-of-veterans-affairs/vets-api/pull/3481/commits/ad66b874d9c8aefebd41c1b6e57baf72ba1e5f5f#diff-06d550e153be774d08afc234a76730c7L21).  Typically we disable CSRF in test, but flipper has it enabled.  I put in a solid amount of effort to make CSRF tags work but at some point I decided that it wasn't worth the effort to fork flipper for the sake of a test.

flipper_view

PR #3507, opened 2019-11-01
PSSG Download: stop drivetime_bands from being updated

nfasulo:
## Description of change
The PSSG download ran in dev for 90 minutes and inserted 11451 drive time bands (~10% of the total). The bands are being inserted into the table at a little over 2 per second, but updates are currently much slower than that. In the interest of getting all of the drive time bands into the table, I have modified the download to skip trying to update existing records for now. This will allow the job to insert more new records each time it runs instead of just checking for updates to the existing records the whole time.

Part of department-of-veterans-affairs/vets-contrib/issues/3534.

## Testing done
Ran locally

## Testing planned
Will run job in dev again to insert more records

### DIFF
--- a/app/workers/facilities/pssg_download.rb
+++ b/app/workers/facilities/pssg_download.rb
@@ -19,8 +19,6 @@ module Facilities
 
     private
 
-    # this method is still under development, will remove this exception later
-    # rubocop:disable Metrics/MethodLength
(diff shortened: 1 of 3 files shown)


nfasulo:
@whuang85 I just pushed a commit to flip that test expectation so that it will succeed (and then fail when we start updating again).

PR #3508, opened 2019-11-01
[no review needed]1880 notifications v1.5 permission bio on staging staging

annaswims:
PR into another feature branch so that I can run the tests on staging


omgitsbillryan:
Do you actually want reviews on this? 😆 You don't need approvals to squash and merge into the feature branch?

annaswims:
@omgitsbillryan I was actually about to ask you why it doesn't deploy to staging. I thought that setting `staging_branch` to this branch's name might work.

talbotmj:
@annaswims this looks encouraging!  It seems you were able to record 5 of the 8 cassettes needed - were you able to record these reading from vet360 with my code without any changes? I don't see any other edits in this PR except to the specs and the new cassettes.

annaswims:
I didn't make any changes, other than the committed ones.  Though I did search/replace internal-`dsva-vagov-staging-fwdproxy-1821450725.us-gov-west-1.elb.amazonaws.com:4433` with `int.vet360.va.gov` in the recorded cassettes. Since the first tests were originally recorded we've added a forward proxy to staging.

saneshark:
I think this branch is maybe committing to staging directly... not to master. See devops channel on DSVA slack.

PR #3509, opened 2019-11-01
first pass at dockerfile refactoring

nathanhruby:
**Part of: department-of-veterans-affairs/va.gov-team#2513**

## Description of change
This refactors the Dockerfile to use multistage builds to build both dev/test and production docker images.  This is the first step in moving vets-api to deploying in docker. 

## Testing done
TBD -- This is WIP right now, making a PR to get Jenkins to run some tests

## Testing planned
TBD

#### Unique to this PR
- [ ] check code climate
- [ ] check tests
- [ ] ask around to find some folks who use Dockerfile

#### Applies to all PRs

- [ ] Appropriate logging

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,5 @@
 tmp/
 data/
 log/*
+.git
+coverage/
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 
 # Ignore bundler config.
 /.bundle
+.envrc
 
 # Ignore the default SQLite database.
 /db/*.sqlite3
@@ -42,6 +43,8 @@ config/application.yml
 tmp/
 data/
 log/*
+.cache
+.local
 
 config/settings.local.yml
 config/settings/*.local.yml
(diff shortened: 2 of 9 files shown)


.dockerignore
```diff
@@ -1,3 +1,5 @@
 tmp/
 data/
 log/*
+.git
```
omgitsbillryan: Not very familiar with `.dockerignore` - does this mean we can alter this line? https://github.com/department-of-veterans-affairs/vets-api/blob/master/Makefile#L22

Some context: these pesky `.git` directories were being left behind by docker, such that, the `.git/refs/tags/` were all owned by `root`. This meant that Jenkins couldn't clean the workspace since it runs as the `jenkins` user.

omgitsbillryan:
FYI - prob want to start a "Draft" PR, so as not to auto-ping the `VA API Engineers` as reviewers.

.dockerignore
nathanhruby: maybe.   the .dockerignore change will prevent the git dir from being copied into the container, but the tests run with a checkout mounted into the container so I don't think that will have much effect.  I'll need to test a bit with jenkins builds to get them working, will try to fix that in the process

nathanhruby:
ok I can close this and make a new one I guess

PR #3511, opened 2019-11-04
first pass at dockerfile refactoring

nathanhruby:
## Description of change
**Relates To: department-of-veterans-affairs/va.gov-team#2513**

This PR refactors the Dockerfile to use a multistage build and to a non-root user for running services.  This allows the Dockerfile to produce dev, ci, and prod usable images from the same Dockerfile.  This is the first step toward containerizing vets-api in prod.

The goal here is to get this change merged and let it run with devs and in CI for a short time while I refactor the vets-api BRD process for review, staging and production and get that moved over.

## Testing done
- `make ci`
- local testing
- this PR doesn't change much in terms of vets-api, mostly just setup around running things

## Testing planned
- review instance
- ?

#### Unique to this PR
- [x] multistage build working
- [x] ci using docker layers effectively
- [x] services in consitner run as a non-root user

#### Applies to all PRs

- [x] Appropriate logging
  - uses stdout, looks the same
  - n/a
  - none

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,5 @@
-tmp/
-data/
-log/*
 .git
 coverage/
+data/
+log/*
+tmp/
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,5 @@
 tmp/
 data/
 log/*
+.git
+coverage/
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 
 # Ignore bundler config.
 /.bundle
+.envrc
 
 # Ignore the default SQLite database.
 /db/*.sqlite3
@@ -42,6 +43,8 @@ config/application.yml
 tmp/
 data/
 log/*
+.cache
+.local
 
 config/settings.local.yml
 config/settings/*.local.yml
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,7 @@ config/application.yml
 tmp/
 data/
 log/*
+.bash_history
 .cache
 .local
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,7 @@ ENV BUNDLE_ENTERPRISE__CONTRIBSYS__COM=$sidekiq_license
 ENV EXCLUDE_SIDEKIQ_ENTERPRISE=$exclude_sidekiq_ent
 ENV RAILS_ENV=$rails_env
 
-# only extra dev/build opts go here, common packages go in base 🖕
+# only extra dev/build opts go here, common packages go in base 👆
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
     git build-essential libxml2-dev libxslt-dev libpq-dev
 RUN curl -sSL -o /usr/local/bin/cc-test-reporter https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,7 +35,6 @@ RUN curl -sSL -o /usr/local/bin/cc-test-reporter https://codeclimate.com/downloa
     cc-test-reporter --version
 RUN freshclam
 COPY --chown=vets-api:vets-api docker-entrypoint.sh .
-USER vets-api
 ENTRYPOINT ["/usr/bin/dumb-init", "--", "./docker-entrypoint.sh"]
 
 ###
@@ -45,7 +44,7 @@ ENTRYPOINT ["/usr/bin/dumb-init", "--", "./docker-entrypoint.sh"]
 FROM development AS builder
 # XXX: move modules/ to seperate repos so we can only copy Gemfile* and install a slim layer
 ARG bundler_opts
-COPY --chown=vets-api:vets-api . .
+COPY . .
 RUN bundle install --binstubs="${BUNDLE_PATH}/bin" $bundler_opts
 
 ###
@@ -58,6 +57,6 @@ FROM base AS production
 
 ENV RAILS_ENV=production
 COPY --from=builder $BUNDLE_PATH $BUNDLE_PATH
-COPY --from=builder /srv/vets-api ./
+COPY --from=builder --chown=vets-api:vets-api /srv/vets-api ./
 USER vets-api
 ENTRYPOINT ["/usr/bin/dumb-init", "--", "./docker-entrypoint.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@
 #      https://github.com/department-of-veterans-affairs/va.gov-team/issues/3032
 
 ###
-# shared configs for all child images, reuse these layers yo
+# shared build/settings for all child images, reuse these layers yo
 ###
 FROM ruby:2.4.5-slim-stretch AS base
 
@@ -15,8 +15,10 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
 WORKDIR /srv/vets-api
 
 ###
-# dev step; use --target=development to stop here
-# you will need to mount your source to /srv/vets-api to do anything useful here
+# dev stage; use --target=development to stop here
+# Be sure to pass required ARGs as `--build-arg`
+# This stage useful for mounting your local checkout with compose
+# into the container to dev against.
 ###
 FROM base AS development
 
@@ -41,7 +43,11 @@ ENTRYPOINT ["/usr/bin/dumb-init", "--", "./docker-entrypoint.sh"]
 
 ###
 # build stage; use --target=builder to stop here
-# This is basically development with the app copied in and built
+# Also be sure to add build-args from development stage above
+#
+# This is development with the app copied in and built.  The build results are used in
+# prod below, but also useful if you want to have a container with the app and not
+# mount your local checkout.
 ###
 FROM development AS builder
 # XXX: move modules/ to seperate repos so we can only copy Gemfile* and install a slim layer
@@ -56,8 +62,8 @@ RUN bundle install --binstubs="${BUNDLE_PATH}/bin" $bundler_opts && \
 ###
 # prod stage; default if no target given
 # to build prod you probably want options like below to get a good build
-# --build-arg rails_env=production --build-arg bundler_opts="--no-cache --without development test"
-# prod
+# --build-arg sidekiq_license="$BUNDLE_ENTERPRISE__CONTRIBSYS__COM" --build-arg rails_env=production --build-arg bundler_opts="--no-cache --without development test"
+# This inherits from base again to avoid bringing in extra built time binary packages
 ###
 FROM base AS production
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,8 +7,9 @@
 ###
 FROM ruby:2.4.9-slim-stretch AS base
 
-RUN groupadd -r vets-api && \
-    useradd -r -m -d /srv/vets-api -g vets-api vets-api
+ARG userid=993
+RUN groupadd -g $userid -r vets-api && \
+    useradd -u $userid -r -m -d /srv/vets-api -g vets-api vets-api
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
     dumb-init clamav imagemagick pdftk curl poppler-utils
 WORKDIR /srv/vets-api
@@ -35,6 +36,7 @@ RUN curl -sSL -o /usr/local/bin/cc-test-reporter https://codeclimate.com/downloa
     cc-test-reporter --version
 RUN freshclam
 COPY --chown=vets-api:vets-api docker-entrypoint.sh .
+USER vets-api
 ENTRYPOINT ["/usr/bin/dumb-init", "--", "./docker-entrypoint.sh"]
 
 ###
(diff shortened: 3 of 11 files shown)


nathanhruby:
Fixing review instance builds on department-of-veterans-affairs/devops#5870

Makefile
```diff
@@ -8,20 +10,28 @@ SPEC_PATH    := spec/
...
+	$(COMPOSE_TEST) down
+
+.PHONY: ci-spec
+ci-spec:
+	@$(BASH_TEST) "bin/rspec ${SPEC_PATH}"
+
+.PHONY: bash
+bash:
+	@$(COMPOSE_DEV) $(BASH)
+
```
omgitsbillryan: Was moving the `bash` target intentional? Make targets are listed alphabetically (for better or worse).

Makefile
```diff
@@ -60,7 +70,7 @@ server:
 
 .PHONY: spec
 spec:
-	@$(BASH_TEST) "bin/rspec ${SPEC_PATH}"
+	@$(BASH_DEV) "bin/rspec ${SPEC_PATH}"
```
omgitsbillryan: Why the change? I'm guessing because this target isn't used by CI at all, in which case maybe we should just delete it.

Makefile
```diff
@@ -8,20 +10,28 @@ SPEC_PATH    := spec/
...
-.PHONY: clean
-clean:
-	rm -r data || true
-	$(COMPOSE_TEST) run vets-api rm -r coverage log tmp .git || true
+.PHONEY: ci-down
+ci-down:
 	$(COMPOSE_TEST) down
 
+.PHONY: ci-clean
+ci-clean:
```
omgitsbillryan: I don't see this target called anywhere, what gives?

Makefile
```diff
@@ -1,4 +1,6 @@
 $stdout.sync = true
+export VETS_API_ID  := $(shell id -u)
```
omgitsbillryan: Clever! It's somewhat unfortunate that our build user is named "vets-api", because `VETS_API_ID` doesn't scream out to me that this is actually a _user_ id. Maybe rename to `VETS_API_USER_ID`? 🤷‍♂

Makefile
```diff
@@ -8,20 +10,28 @@ SPEC_PATH    := spec/
...
-clean:
-	rm -r data || true
-	$(COMPOSE_TEST) run vets-api rm -r coverage log tmp .git || true
+.PHONEY: ci-down
+ci-down:
 	$(COMPOSE_TEST) down
 
+.PHONY: ci-clean
+ci-clean:
+	$(COMPOSE_TEST) run vets-api git clean -fdx || true
```
omgitsbillryan: TIL `git clean`, so 👍 .

Does `git clean` write to output what is being deleted? A use-case may exist where a developer is generating some file and wants to see it after a build. 

I think `git clean` is more elegant, but what we had before was more explicit.

annaswims:
this is what I see when I run ` make bash`

````
$ make bash
Creating network "vets-api_default" with the default driver
Creating vets-api_postgres_1 ... done
Creating vets-api_redis_1    ... done
/usr/local/src/vets-api/Gemfile not found
There was an error while trying to write to `/usr/local/bundle/config`. It is likely that you need to grant write permissions for that path.
make: *** [bash] Error 23
````

File: Makefile (hunk 1 above)
Comment (nathanhruby) [reply Makefile]: I moved things around to be grouped in "ci-" and "not ci-" fashion because that's how I tend to think.  I can re-ascibeticalize if you it works better for folks.

File: Makefile (hunk 2 above)
Comment (nathanhruby) [reply Makefile]: it isn't used in ci, so I thought it was a developer helper, so changed it to using the same config as the rest of dev helpers (since test and dev compose files are not interchangeable)

File: Makefile (hunk 3 above)
Comment (nathanhruby) [reply Makefile]: this was the `clean` target, but it wasn't needed anymore after I fixed the file permissions, but I left it in case it was in people's workflows.  I can remove it, or perhaps return an error?

File: Makefile (hunk 4 above)
Comment (nathanhruby) [reply Makefile]: Good idea!  Will do.

File: Makefile (hunk 5 above)
Comment (nathanhruby) [reply Makefile]: This will output what it removes, yes.  

FWIW, as an old user of autotools `git clean` is what I expect a `make clean` target should do, while the previous behavior was what I would have expected a `make dist` or `make proper` to do.

**Comment by nathanhruby** [comment]
@annaswims that's odd, I don't know of any `/usr/local/src/vets-api` path in the tooling or Dockerfile that would account for that error.  Perhaps you have an old docker image that compose is finding?  Try a `docker image ls vets-api:latest` to see how old your image is followed by an rm to remove it.  Re-running `make bash` which will trigger a rebuild of the container.  

If that works for you, we should probably add a `--force-recreate` to the compose command for a while to clear out caches from folks automatically on update.

For reference, here's what I get with a brand new vets-api:latest image made while I was typing this:
```
[nathanhruby@goodidea ../vets-api] (nrh/docker-cleanups)$ make bash
Starting vets-api_postgres_1 ... done
Starting vets-api_redis_1    ... done
The dependency tzinfo-data (>= 0) will be unused by any of the platforms Bundler is installing for. Bundler is installing for ruby but the dependency is only for x86-mingw32, x86-mswin32, x64-mingw32, java. To add those platforms to the bundle, run `bundle lock --add-platform x86-mingw32 x86-mswin32 x64-mingw32 java`.
The Gemfile's dependencies are satisfied
vets-api@d33bdfcad276:~$ pwd
/srv/vets-api
vets-api@d33bdfcad276:~$ ls
CONTRIBUTING.md  Dockerfile  Gemfile.lock  Jenkinsfile  Makefile  README.md  app        bin     config.ru  data  docker-compose.test.yml  docker-entrypoint.sh  lib  modules  rakelib  tmp
Dangerfile       Gemfile     Guardfile     LICENSE.md   Procfile  Rakefile   benchmark  config  coverage   db    docker-compose.yml       docs                  log  public   spec
vets-api@d33bdfcad276:~$  
```

File: Makefile (hunk 3 above)
Comment (omgitsbillryan) [reply Makefile]: I've vote to remove. My understanding is that `COMPOSE_DEV` & `BASH_DEV` are for local use. Given that this target uses `COMPOSE_TEST`, it shouldn't be used locally by anyone.

File: Makefile (hunk 1 above)
Comment (omgitsbillryan) [reply Makefile]: I agree with you, but I'm mostly indifferent TBH. I'm good either way.

File: Makefile (hunk 5 above)
Comment (omgitsbillryan) [reply Makefile]: So long as it's printing out what's being deleted, then 👍 . It's a better catch-all for cleaning the space.

File: Jenkinsfile [hunk 6]
```diff
@@ -110,7 +110,7 @@ pipeline {
   }
   post {
     always {
-      sh 'make clean'
+      sh 'make ci-down'
```
omgitsbillryan: I think we also want `make ci-clean` here as well, no?

nathanhruby:
do you have a `.bundle/config` or `~/.bundle/config` or env var that might be spilling over?  So you want to do a screenshare to try to debug for a bit?

.dockerignore
```diff
@@ -1,3 +1,5 @@
 tmp/
 data/
 log/*
+.git
```
wyattwalter: 👍

wyattwalter: also, oops

Jenkinsfile
nathanhruby: nah, the following `deleteDir()` whacks all the data, the clean was there to run from inside the container to remove the root owned files which `deleteDir()` could not because it is run as the jenins user.  

We don't have root owned files anymore because of the user changes to run as vets-api user.

Makefile
nathanhruby: so keep the clean task or whack it per above comment?

nathanhruby:
ooooh, I bet this is the bundle volume having the old bundle.  `make down && docker volume rm vets-api_bundle && make bash` should get things to a better state.

Dockerfile
```diff
@@ -1,36 +1,68 @@
...
+    cc-test-reporter --version
+RUN freshclam
+COPY --chown=vets-api:vets-api docker-entrypoint.sh .
+USER vets-api
+ENTRYPOINT ["/usr/bin/dumb-init", "--", "./docker-entrypoint.sh"]
+
+###
+# build stage; use --target=builder to stop here
+# This is basically development with the app copied in and built
+###
```
wyattwalter: What's the intended use of the builder target?

Dockerfile
```diff
@@ -1,36 +1,68 @@
...
-RUN chmod +x /docker-entrypoint.sh
-ENTRYPOINT ["/docker-entrypoint.sh"]
-
-ENV BUNDLE_PATH=/bundle \
-    BUNDLE_BIN=/bundle/bin \
-    GEM_HOME=/bundle
-ENV PATH="${BUNDLE_BIN}:${PATH}"
+ENV RAILS_ENV=$rails_env
+
+# only extra dev/build opts go here, common packages go in base 🖕
```
wyattwalter: Is this the finger you intended? 😂  I can't tell from the rendering on GitHub

Dockerfile
nathanhruby: It's where the source is copied in and bundler is run.  The results of that are copied into a fresh new container without build deps for prod use.

Also having this in a dedicated stage means we can skip it for dev/ci which makes the layers there much more reusable over time.

Do you feel this comment could be clearer?

Dockerfile
nathanhruby: it was not, thanks!

.dockerignore
```diff
@@ -1,3 +1,5 @@
 tmp/
```
johnpaulashenfelter: Minor, but can you alphabetize? Makes it easier as file grows

Dockerfile
```diff
@@ -1,36 +1,68 @@
...
+
+###
+# shared configs for all child images, reuse these layers yo
+###
+FROM ruby:2.4.5-slim-stretch AS base
+
+ARG userid=993
+RUN groupadd -g $userid -r vets-api && \
+    useradd -u $userid -r -m -d /srv/vets-api -g vets-api vets-api
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
```
johnpaulashenfelter: Won't this give us potentially different versions of `clamav` etc every time? Can we move that higher up in a build (perfect world would be in the `FROM`) so we get repeatable versions? 

I realize this might be a further down the road issue. It doesn't make anything appreciably worse to leave it as-is since that's our current state.

.dockerignore
nathanhruby: sure, done.

Dockerfile
nathanhruby: You're correct in that this is no different that what is done now (either here or in the current vets-api BRD image builds) so I don't think this is worth dealing with here and is outside the scope of my current task.

The source distro is still Debian which is conservative with updates and compatibility so this shouldn't be a big issue.

.gitignore
```diff
@@ -42,6 +43,8 @@ config/application.yml
 tmp/
 data/
 log/*
+.cache
+.local
```
wyattwalter: I'm also getting a `.bash_history` in the main directory after doing a `make bash`, could be ignored here.

Dockerfile
wyattwalter: Oh, my question was weird maybe. Mostly thinking about the "how to stop here" comment and when I'd want to stop at this point. It looks like the docker-compose file for both local dev and test is using target of development, so wasn't sure what might be the advantage here over just having a `development` and `production` target.

Dockerfile
```diff
@@ -1,36 +1,68 @@
...
+###
+FROM development AS builder
+# XXX: move modules/ to seperate repos so we can only copy Gemfile* and install a slim layer
+ARG bundler_opts
+COPY --chown=vets-api:vets-api . .
+USER vets-api
+# --no-cache doesn't do the right thing, so trim it during build
+# https://github.com/bundler/bundler/issues/6680
+RUN bundle install --binstubs="${BUNDLE_PATH}/bin" $bundler_opts && \
+    find ${BUNDLE_PATH}/cache -type f -name \*.gem -delete
```
wyattwalter: This is failing for me locally when trying to target it because it doesn't get the sidekiq license arg passed.

nathanhruby: Weird, how are you building it?  You need to pass in the var as a build arg, which is something the compose file does for us. If running a plain docker build command, the invocation is:
```
docker build --build-arg sidekiq_license="$BUNDLE_ENTERPRISE__CONTRIBSYS__COM" --target builder -t local/vetsapi:new .
```

I just retested that and it works.

.gitignore
nathanhruby: good catch, done!

Makefile
nathanhruby: I removed this target

Makefile
nathanhruby: this is removed per other comment

nathanhruby:
ok, I changed the bundle volumes names so the next build will get a fresh empty bundle which I think should solve the problems @omgitsbillryan and @annaswims are having.  Can you retest?  I'd love to get this merged.

Dockerfile
wyattwalter: 🤦‍♀ Yeah, that works for me. I looked at my shell history and I typo'ed the arg.

Jenkinsfile
omgitsbillryan: Which changes are you referring to specifically? 

I too had once believed we were running everything as the `vets-api`, but I eventually gave up and accepted that, `root` was _somehow_  coming to own various files.

nathanhruby: The old Dockerfile did a user creation in the container (eg: the `useradd`), but never set the USER directive in the Dockerfile to make docker use it, so it kept running things as root.  I added USER directives in the stages where they are needed to make it run as vets-api user.

That is also the reason for the `userid` ARG that gets passed in via the `VETS_API_USER_ID` env var set in the Makefile.  That will override the default userid of the vets-api user in the container and make it match the person building and running the container .  That way that the files written to the mounted volume have the same UID inside and outside the container and can be cleaned up easily.

nathanhruby:
Thanks everyone!

PR #3512, opened 2019-11-04
New Drivetime Band Query

rtravitz:
## Description of change

This PR finishes [#3378](https://github.com/department-of-veterans-affairs/vets-contrib/issues/3378) and [#3104](https://github.com/department-of-veterans-affairs/vets-contrib/issues/3104). It changes the underlying implementation of the `/nearby` endpoint in the Facilities API. 

Previously, the `/nearby` endpoint would accept an address or coordinate pair from the user, query Bing Maps to retrieve an isochrone for that user within a specified drivetime, and then checked that isochrone against VHA facility lat/longs stored in Postgres. Retrieval from Bing was slow, relatively imprecise, and not idempotent, as it relied on varying traffic conditions.

With this PR, the user still sends in their location, either in an address that gets geocoded with Bing Maps or a coordinate pair. But now we run a query in PostGIS to see which facility isochrones stored there intersect with the user location.

#### Notable changes
- `DrivetimeBand` now has a method to find based on location, max distance, and related facility ids
- `VHAFacility` can return facilities that offer a certain service
- `/nearby` can no longer return GeoJSON
- Geocoding functionality was extracted from the now-deleted `NearbyFacility` into its own service
- A new `NearbySerializer` was created satisfying JSON API and conforming to the updated endpoint spec
- Removed a number of deprecated isochrone VCR cassettes and added DrivetimeBand factories that align with the existing Portland-based VHAFacility factories.
- In the separate PR [#3518](https://github.com/department-of-veterans-affairs/vets-api/pull/3518), an index was added to the polygon column on the `drivetime_bands` table to speed up the intersect query

## Testing done

- Added specs for `DrivetimeBand`
- Updated specs for `VHAFacility`
- Overhauled specs in `nearby_request_spec.rb`
- Manually verified behavior against the full set of 11k+ drivetime bands

#### Unique to this PR

- [x] Return a list of facilities that include the requested location within the requested drive time band (or closer)
- [x] Allow filtering by services
- [x] The nearby controller can return data in the format specified in the 
- [x] The nearby controller can no longer return data in GeoJSON format

#### Applies to all PRs
- [x] Appropriate logging


app/models/drivetime_band.rb
```diff
@@ -2,4 +2,19 @@
 
 class DrivetimeBand < ApplicationRecord
   belongs_to :vha_facility, class_name: 'Facilities::VHAFacility'
+
+  class << self
+    def find_within_max_distance(lat, lng, drive_time, ids)
+      query = 'ST_Intersects(polygon, ST_MakePoint(:lng,:lat)) AND max <= :max'
+      params = { lng: lng, lat: lat, max: drive_time }
+
+      unless ids.nil?
```
whuang85: would it be better to check blank?

edmangimelli: the double negative throws me off a bit.
maybe `if ids.present?` ?

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
```diff
@@ -58,6 +58,10 @@ def set_facility_type
 
       private
 
+      def per_page
+        20
+      end
```
edmangimelli: why not a constant? i might be out of the loop on this repo's conventions or not aware of some best practice

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
```diff
@@ -70,21 +76,34 @@ def validate_params
...
           no_missing_fields = (req_field_names - obs_fields).empty?
           break loc_type if no_missing_fields
         end
-        QUERY_INFO[location_type]
+
+        if location_type.eql? :address
+          GeocodingService.new.query(params[:street_address], params[:city], params[:state], params[:zip])
+        else
+          params.slice(:lat, :lng)
+        end
```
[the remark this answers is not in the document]
rtravitz: Explicit in what way? This reads okay to me (as another outside party— @whuang85 put that block together). If there's an address, it geocodes and returns. Otherwise it returns the passed in lat/long. Rubocop favors the implicit returns, though as someone who came to Ruby from other languages, sometimes that throws me, too.

spec/models/nearby_facility_spec.rb
```diff
@@ -1,103 +0,0 @@
...
-
-  describe '#make_linestring' do
-    it 'converts a polygon array in a string' do
-      polygon = [[45.451913, -122.440689], [45.451913, -122.786758], [45.64, -122.440689], [45.64, -122.786758]]
-      linestring = '-122.440689 45.451913,-122.786758 45.451913,-122.440689 45.64,-122.786758 45.64'
-      expect(NearbyFacility.make_linestring(polygon)).to eq(linestring)
-    end
-  end
-
-  describe 'parsing location response' do
```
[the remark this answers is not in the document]
rtravitz: SimpleCov agrees with you 👍. Adding these kinds of cases back in.

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
```diff
@@ -70,21 +76,34 @@ def validate_params
...
-                        per_page: resource.per_page,
-                        total_pages: resource.total_pages,
-                        total_entries: resource.total_entries },
-          distances: [] }
+        {
+          pagination: {
+            current_page: resource&.try(:current_page),
+            per_page: resource&.try(:per_page),
+            total_pages: resource&.try(:total_pages),
+            total_entries: resource&.try(:total_entries)
```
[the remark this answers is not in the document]
whuang85: this is just guarding when return `DrivetimeBand.none`.. I think the only one that has issues with `DrivetimeBand.none` is `per_page` but just guarded everything.

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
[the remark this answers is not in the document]
nfasulo: Yeah, a final line in the method that returns a variable assigned in the if/else was what I was referring to. Obviously we don't literally need it, just trying to get a temperature check on whether folks thought it would make it easier to read.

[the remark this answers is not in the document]
whuang85: not sure what ruby conventions are.. or even coding best practices for VA.. Originally, I had explicit returns in each conditional and rubocop removed the `return`s

spec/models/nearby_facility_spec.rb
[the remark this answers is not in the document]
rtravitz: @nfasulo Let me know if the specs I added in a second ago resolve this comment!

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
```diff
@@ -17,31 +16,34 @@ class NearbyController < ApplicationController
       skip_before_action(:authenticate)
       before_action :set_default_format
       before_action :set_facility_type
+      before_action :set_default_drivetime
       before_action :validate_params, only: [:index]
 
+      PER_PAGE = 20
+
       REQUIRED_PARAMS = {
```
edmkitty: Out of scope/tech debt thought: Should we use Rails Strong Parameters here instead of a constant?

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
```diff
@@ -17,31 +16,34 @@ class NearbyController < ApplicationController
...
 
       def index
-        query_method = get_query_method(params)
-        params_hash = params.permit!.to_h.symbolize_keys
-        resource = query_method.call(params_hash).paginate(page: params[:page],
-                                                           per_page: params[:per_page] || NearbyFacility.per_page)
+        lat_lng = get_lat_lng(params)
+        eligible_facilities = Facilities::VHAFacility.with_services(params[:services]) if params[:services]
+
+        resource = if lat_lng.present?
```
edmkitty: I think I would prefer a more specific variable name here, perhaps `drivetime_bands` or something instead of `resource`

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
```diff
@@ -17,31 +16,34 @@ class NearbyController < ApplicationController
...
-        lat_lng: NearbyFacility.method(:query_by_lat_lng)
-      }.freeze
 
       def index
-        query_method = get_query_method(params)
-        params_hash = params.permit!.to_h.symbolize_keys
-        resource = query_method.call(params_hash).paginate(page: params[:page],
-                                                           per_page: params[:per_page] || NearbyFacility.per_page)
+        lat_lng = get_lat_lng(params)
+        eligible_facilities = Facilities::VHAFacility.with_services(params[:services]) if params[:services]
```
edmkitty: It wasn't immediately clear to me that `eligible_facilities` is actually a list of ids, and not a facilities relation. I think the select/pluck statement in `with_services` should be moved up into the controller, OR maybe just have `with_services` method renamed to `ids_with_services` to make it extra clear. I also think it should be `eligible_facility_ids`.

nfasulo: I like the idea of moving the `pluck` into the controller. That would allow us to theoretically reuse the query itself and return any fields we want.

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
```diff
@@ -70,21 +76,34 @@ def validate_params
         validate_type_and_services_known unless params[:type].nil?
       end
 
-      def get_query_method(params)
+      def get_lat_lng(params)
         obs_fields = params.keys.map(&:to_sym)
         location_type = REQUIRED_PARAMS.find do |loc_type, req_field_names|
```
edmkitty: When does this block make the decision to return a location type? It seems like it only even returns the last location type, and only if all the params are present. Otherwise, it's nil. Am I misunderstanding?

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
```diff
@@ -70,21 +76,34 @@ def validate_params
...
           no_missing_fields = (req_field_names - obs_fields).empty?
           break loc_type if no_missing_fields
         end
-        QUERY_INFO[location_type]
+
+        lat_lng = params.slice(:lat, :lng)
+        if location_type.eql? :address
```
edmkitty: We can change this block to look like this:
```suggestion
        if location_type.eql? :address
            GeocodingService.new.query(params[:street_address], params[:city], params[:state], params[:zip])
        else
            params.slice(:lat, :lng)
        end
```

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
```diff
@@ -70,21 +76,34 @@ def validate_params
...
-      def get_query_method(params)
+      def get_lat_lng(params)
         obs_fields = params.keys.map(&:to_sym)
         location_type = REQUIRED_PARAMS.find do |loc_type, req_field_names|
           no_missing_fields = (req_field_names - obs_fields).empty?
           break loc_type if no_missing_fields
         end
-        QUERY_INFO[location_type]
+
+        lat_lng = params.slice(:lat, :lng)
```
edmkitty: We can change the block to look like this:
```suggestion
        if location_type.eql? :address
            GeocodingService.new.query(params[:street_address], params[:city], params[:state], params[:zip])
        else
            params.slice(:lat, :lng)
        end
```
And then remove the `lat_lng` being returned at the end of the method.

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
whuang85: it iterates over the`REQUIRED_PARAMS` and returns the first one that has no missing fields.

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
whuang85: lol, I changed it to the current implementation because it was discussed that it was clearer to have a final return at the end of the line.

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
edmkitty: In that case,  I think the logic here is overly complicated. The `break` is unnecessary since `find` just needs a condition to return the thing. Here's what we could do:
```suggestion
        location_type = REQUIRED_PARAMS.keys.find do |loc_type|
           (REQUIRED_PARAMS[loc_type] - obs_fields).empty?
        end
```

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
edmkitty: That is unnecessary in ruby and not ruby-ish. But if that's not what the team wants, then I'm ok keeping it the way it is

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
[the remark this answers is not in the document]
edmkitty: I'm a little confused... `&.` is syntactic sugar for `try` isn't it? Why are we using both?

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
whuang85: That will return an array of key, value. So for example, if the the params are `{lat: 45, lng: 45}` then that function will return `[:lat_lng, [:lat, :lng]]`

this was also discussed here:
https://github.com/department-of-veterans-affairs/vets-api/pull/3273/files/a8c59804eca6e24421bd527e208382c180665c6f

edmkitty: Actually, in my code, we are iterating over an array of the keys, not the hash. So here we are iterating over `[:address, :lat_lng]`. Make sense?

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
edmkitty: Made a ticket for it.

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
whuang85: ah cool, didn't see the keys part

modules/va_facilities/spec/requests/nearby_request_spec.rb
```diff
@@ -1,420 +1,99 @@
...
-        health_services = json['data'][0]['attributes']['services']['health']
-        expect(health_services.length).to eq(0)
-      end
-    end
-  end
+      get base_query_path,
+          params: { lat: 45.4967668, lng: -122.6832211, drive_time: '10' },
+          headers: { 'HTTP_ACCEPT' => 'application/json' }
 
-  context 'with invalid request parameters' do
```
edmkitty: Why were these unhappy-path tests for param validations removed? Don't we need them?

modules/va_facilities/app/services/geocoding_service.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+require 'common/exceptions'
+
+class GeocodingService
```
edmkitty: This is SO much nicer than it was! This is awesome!

modules/va_facilities/app/services/geocoding_service.rb
```diff
@@ -0,0 +1,46 @@
...
+  def query(street_address, city, state, zip)
+    return nil unless [street_address, city, state, zip].all?(&:present?)
```
edmkitty: Neat pattern here! I didn't know about it!

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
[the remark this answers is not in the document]
whuang85: if you call `try` on a `nil` object, wouldn't it fail?  In this case, I don't think resource will ever be `nil` so yep, it is probably not needed.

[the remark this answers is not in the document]
edmkitty: We discussed offline a bit and found that `&.` will help us in the case of nils, but not if the method doesn't exist. `.try` will return nil in both of these cases: the caller is nil, and the method being called doesn't exist on the object. I think `&.` makes sense in this scenario, and if there is a `NoMethodError` on an existing object then something has probably gone horribly wrong

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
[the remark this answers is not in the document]
rtravitz: Calling this six/half-dozen and leaving as is. Strong objections welcomed and I'll reopen.

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
rtravitz: I'm going with `bands`

modules/va_facilities/spec/requests/nearby_request_spec.rb
rtravitz: My mistake. Took them out when doing the overhaul to reduce noise. Adding them back.

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
rtravitz: Moved pluck into the controller and made the method on `VHAFacility` return the whole facility.

modules/va_facilities/app/controllers/va_facilities/v1/nearby_controller.rb
rtravitz: I'm going to leave this however it ended up. Six/half-dozen.

PR #3515, opened 2019-11-06
create unique config for vaos user service

kreek:
## Description of change
Creates a unique service configuration for VAOS UserService so it has the correct headers

## Testing done
unit tests

## Testing planned
staging integration tests

#### Unique to this PR
- [ ] Create a user service config file that extends the vaos configuration and overrides the connection

#### Applies to all PRs

- [ ] Appropriate logging


modules/vaos/spec/request/appointments_request_spec.rb
```diff
@@ -4,12 +4,10 @@
 RSpec.describe 'vaos appointments', type: :request do
   include SchemaMatchers
 
-  let(:rsa_private) { OpenSSL::PKey::RSA.generate 4096 }
-
   before do
     Flipper.enable('va_online_scheduling')
     sign_in_as(current_user)
-    allow_any_instance_of(VAOS::JWT).to receive(:rsa_private).and_return(rsa_private)
+    allow_any_instance_of(VAOS::UserService).to receive(:session).and_return('stubbed_token')
```
saneshark: Nitpick: see this in a couple of the specs and there will likely be more in the future, possibly introduce rails helper or even shared examples where prudent. Can do as cleanup / refactor later.

modules/vaos/app/services/vaos/user_service.rb
```diff
@@ -16,7 +16,7 @@ def session(user)
     def create_session(user)
       url = '/users/v2/session?processRules=true'
       token = VAOS::JWT.new(user).token
-      response = perform(:post, url, token, 'Content-Type' => 'text/plain', 'Referer' => 'https://api.va.gov')
+      response = perform(:post, url, token)
       if response&.body
         SessionStore.new(user_uuid: user.uuid, token: response.body).save
         response.body
```
johnpaulashenfelter: This returns either a `response.body` from the `if` or a `response`. Seems like you might want something more predictable in the event of an `else` since this is the output that goes into session creation.

I'd also rather see the storage of the session token happen closer to the cache check in the `session` method.

saneshark: I think from testing what we've seen is that anything not returning a response body will raise an exception which would break execution and not cache, but this is a good catch! There might be some edge case we have unaccounted for which could present difficult to debug scenarios with current handling (in the event we return a 2xx without response body).

PR #3516, opened 2019-11-06
Set PR jenkins job name to the branch name

omgitsbillryan:
## Description of change
Our jenkins PR job names take the form : `PR-<GITHUB_PR_ID>`, e.g.
image

This PR renames those jobs to instead take the form: `<BRANCH_NAME>`, so it will look like our "branches" jobs:
image

The new job name is more human-readable.

## Testing done
None.

#### Applies to all PRs

- [ ] Appropriate logging

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -20,8 +20,11 @@ pipeline {
   }
 
   stages {
-    stage('Checkout Code') {
+    stage('Checkout Code & Set Job Name') {
       steps {
+        script {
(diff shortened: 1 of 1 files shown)


omgitsbillryan:
I dont think there's a good way to do this in Jenkins. One can change the build number to include the branch name... but that's not great IMO and creates a non-standard jenkins environment. Going to close since it's not viable + not easy and probably not worth the effort.

PR #3441, opened 2019-10-17
vet360 address validation endpoint

lihanli:
## Description of change
https://github.com/department-of-veterans-affairs/va.gov-team/issues/2532

## Testing done
automated tests

## Testing planned
test in staging

#### Applies to all PRs

- [x] Appropriate logging
- [x] Swagger docs have been updated, if applicable
- [x] Provide link to originating GitHub issue, or connected to it via ZenHub
- [x] Does not contain any sensitive information (i.e. PII/credentials/internal URLs/etc., in logging, hardcoded, or in specs)
- [x] Provide which alerts would indicate a problem with this functionality (if applicable)

PR #3517, opened 2019-11-06
Danger-bot: More meaningful maximum PR size message

omgitsbillryan:
## Description of change
Create a more meaningful MAX_PR_SIZE message from `danger-bot`.

**Before**
image

**After PR**
image

UPDATE:
**After PR Comments**
image

#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging


app/controllers/application_controller.rb
```diff
@@ -7,6 +7,7 @@
 require 'sentry_logging'
 require 'aes_256_cbc_encryptor'
 
+# a change
```
omgitsbillryan: ignore

app/serializers/appointment_serializer.rb
```diff
@@ -8,4 +8,264 @@ class AppointmentSerializer < ActiveModel::Serializer
   def id
     nil
   end
+  # long lines
```
LindseySaari: Does this chunk need to be removed (Lines 11-270)?  Or maybe you did already? I'm guessing it was for testing purposes. Otherwise, looks good! I like the message details 👍

dhall:
If we're going for human readable here I'd suggest something like 
'This pull request changes 321 lines of code. Our approach is that smaller is better, you may consider breaking your changes up into multiple PRs that are less than 250 Loc.'

annaswims:
I vote we exclude the `spec/` directory from the line count entirely and just adjust the expected line count, if necessary.  

I'd also lean towards excluding `app/swagger`

We've got a fair few PRs that are adding, documenting and testing endpoints.  For example I don't think I'd suggest breaking this one up into several PR's
https://github.com/department-of-veterans-affairs/vets-api/pull/3441

app/serializers/appointment_serializer.rb
omgitsbillryan: Yep just for testing 👍

omgitsbillryan:
@annaswims I disagree with `/spec` but agree with `/swagger`. In any case, t's a bit outside the scope for this PR so I opened a new one : https://github.com/department-of-veterans-affairs/vets-api/pull/3522

jholton:
I agree with Derek on taking another pass at the verbage.

> You changed: 273 > 250 LoC

Is a bit opaque in what is being conveyed, especially for engineers new to the repo.  His version (after fixing the run-on sentence 😃 ) does a better job of communicating that 250 lines changed is our threshold for this warning.

jholton:
I agree on excluding Swagger. 

I think I agree with @annaswims on excluding `spec/`. We don't want devs skimping on specs to come in under the threshold.

PR #3518, opened 2019-11-06
Add migration to index polygon on drivetime_bands table

rtravitz:
## Description of change

This adds an index to the `polygon` column on the `drivetime_bands` table to significantly speed up PostGIS intersect queries. This is to support [#3104](https://github.com/department-of-veterans-affairs/vets-contrib/issues/3104).

## Testing done

- Ran and verified migration locally

## Testing planned

- Will run and verify the migration in lower environments first

#### Applies to all PRs

- [x] Appropriate logging


rtravitz:
> This migration is fine, but there should be a corresponding `schema.rb` change as well.

Missed the add. Thanks @johnpaulashenfelter!

johnpaulashenfelter:
@rtravitz no problem -- its also why the Jenkins build was failing.

PR #3519, opened 2019-11-06
Add bgs external key settings

charleystran:
## Description of change
Add the settings for the BGS external keys to be sent to BGS Gem

## Testing done
Local testing

- [ ] VetsApi can load and set the external uuid and external key for the bgs gem

#### Applies to all PRs

- [ ] Appropriate logging

### DIFF
--- a/config/initializers/bgs.rb
+++ b/config/initializers/bgs.rb
@@ -8,6 +8,8 @@ LighthouseBGS.configure do |config|
   config.client_station_id = Settings.bgs.client_station_id
   config.client_username = Settings.bgs.client_username
   config.env = Rails.env.to_s
+  config.external_uuid = Settings.bgs.external_uuid
+  config.external_key = Settings.bgs.external_key
   config.mock_response_location = Settings.bgs.mock_response_location
   config.mock_responses = Settings.bgs.mock_responses
 end
--- a/token_creator.sh
+++ b/token_creator.sh
@@ -1,0 +1,75 @@
+#!/usr/bin/env bash
+
+#
+# JWT Encoder Bash Script
+#
+
+secret='L1ghth0us3B1p'
+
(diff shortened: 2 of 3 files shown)


token_creator.sh
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env bash
```
kfrz: This seems rather unpolished approach, is there a more `ruby` way to accomplish your goal? At least I'd like to see it in `bin/scripts` with some more error handling if it is to be bash

token_creator.sh
```diff
@@ -0,0 +1,75 @@
...
+
+#
+# JWT Encoder Bash Script
+#
+
+secret='L1ghth0us3B1p'
```
kfrz: Guessing this is temporary?

config/initializers/bgs.rb
```diff
@@ -8,6 +8,8 @@
   config.client_station_id = Settings.bgs.client_station_id
   config.client_username = Settings.bgs.client_username
   config.env = Rails.env.to_s
+  config.external_uuid = Settings.bgs.external_uuid
```
kfrz: Might need to add these to [Jenkins](http://jenkins.vfs.va.gov/blue/organizations/jenkins/testing%2Fvets-api/detail/add_bgs_external_key_info/1/pipeline) somehow

saneshark:
This might not be related to the intent of this PR, but I recently ran into an issue with this initializer when I was running specs offline. The check for an ipaddress seems a little intrusive and perhaps when iplookup fails you can default to 127.0.0.1?

Something like this, which I've since reverted.
https://github.com/department-of-veterans-affairs/vets-api/pull/3552/commits/40f343b56899fb28e2e8a9a8e7c22e6cea01768d

PR #3520, opened 2019-11-07
VAOS get facilities endpoint

kreek:
## Description of change
Adds the `GET /v0/vaos/facilities` endpoint and service method.

## Testing done
unit tests

## Testing planned
staging integration tests

#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging


modules/vaos/config/routes.rb
```diff
@@ -3,5 +3,6 @@
 VAOS::Engine.routes.draw do
   resources :appointments, only: %i[index], defaults: { format: :json }
   resources :systems, only: :index, defaults: { format: :json }
+  resources :facilities, only: :index, defaults: { format: :json }
   get 'api', to: 'apidocs#index', defaults: { format: :json }
 end
```
jholton: Nitpick:  instead of listing the same `defaults` on every line, I think this will work:
```
VAOS::Engine.routes.draw do
  defaults format: :json do
    resources :appointments, only: %i[index]
    resources :systems, only: :index
    resources :facilities, only: :index
    get 'api', to: 'apidocs#index'
  end
end

```

modules/vaos/spec/request/facilities_request_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+        end
+      end
+    end
+
+    context 'when the facility code param is missing' do
+      it 'returns a 200 with the correct schema' do
+        VCR.use_cassette('vaos/systems/get_facilities', match_requests_on: %i[host path method]) do
+          get '/v0/vaos/facilities'
+
+          expect(response).to have_http_status(:bad_request)
```
jholton: > it 'returns a 200 with the correct schema' do

> expect(response).to have_http_status(:bad_request)

😕

modules/vaos/spec/services/systems_service_spec.rb
```diff
@@ -27,4 +27,25 @@
       end
     end
   end
+
+  describe '#get_facilities' do
+    context 'with 141 facilities' do
+      it 'returns an array of size 10' do
+        VCR.use_cassette('vaos/systems/get_facilities', match_requests_on: %i[host path method]) do
+          response = subject.get_facilities(user, 688)
+          expect(response.size).to eq(141)
```
jholton: > it 'returns an array of size 10'

> expect(response.size).to eq(141)

Am I misunderstanding something?

modules/vaos/spec/request/facilities_request_spec.rb
saneshark: good catch

PR #3522, opened 2019-11-07
Exclude app/swagger from danger-bot

omgitsbillryan:
## Description of change
Excludes swagger doc from our danger-bot "Big PR" warning.

## Testing done
None.

## Testing planned
None.

#### Applies to all PRs

- [ ] Appropriate logging


omgitsbillryan:
Thoughts on excluding `/specs` as well? cc @annaswims 

I'd lean towards no, but I can see how that code is usually more "boilerplate" than regular app code. Maybe a middle ground could be to just bump the `250` max up to `500`?

annaswims:
I lean towards excluding `/specs`. I could see reducing the MAX_PR_SIZE though since about half of a good PR is specs.  I don't want to do anything to discourage people from writing specs.

jholton:
I'm still leaning toward excluding `specs` for the same reason as Anna:
> I don't want to do anything to discourage people from writing specs.

Theoretically we have Rubocop to keep specs tidy, but I think a lot of those cops are disabled.

As long as we keep the substantive changes small, I usually don't have an issue groking a PR.

PR #3523, opened 2019-11-07
Lower the VAOS JWT session TTL to 15 min

kreek:
## Description of change
Discovered in VAOS security review that their jwt session tokens expire after 15min. [Future work](https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/vets.gov-team/19933) will set the correct expires from the jwt and push out the session when we make requests. 

For now if a user does not make a request in 15 min then our stored jwt token should expire.

## Testing done
N/A

## Testing planned
staging integration tests

#### Unique to this PR

- [ ] Lowers the VAOS JWT TTL to just under 15min (14:55, 855s)

#### Applies to all PRs

- [ ] Appropriate logging


kreek:
@saneshark going to add that to the future work ticket, want to unblock FE for now

config/redis.yml
```diff
@@ -88,7 +88,7 @@ development: &defaults
     each_ttl: 60 # 1 minute
   va_mobile_session:
     namespace: va-mobile-session
-    each_ttl: 3600
+    each_ttl: 855
```
kfrz: Can we add an example for this, something like:

```rb
# modules/vaos/spec/services/jwt_spec.rb

# ...
  describe '#token' do
    context 'when a user makes no requests for 15m' do
      it 'expires the token' do
        Timecop.travel(now + 15.minutes)
        expect(subject).to be_expired
        Timecop.return
      end
    end
  end
# ... 
```

PR #3525, opened 2019-11-08
VAOS facility clinics

kreek:
## Description of change
Adds `/v0/vaos/facilities/:id/clinics` and updated systems service

Also updates brakeman gem:
```
Name: brakeman
Version: 4.6.1
Advisory: CVE-2019-18409
Criticality: Medium
URL: https://brakemanscanner.org/blog/2019/10/14/brakeman-4-dot-7-dot-1-released
Title: brakeman world writable files allow local privilege escalation
Solution: upgrade to >= 4.7.1
```

## Testing done
unit tests

## Testing planned
staging integration tests

#### Unique to this PR
- [ ] adds clinic controller
- [ ] updates systems service to return facility clinics

#### Applies to all PRs

- [ ] Appropriate logging


modules/vaos/app/controllers/vaos/clinics_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+require_dependency 'vaos/application_controller'
+
+module VAOS
+  class ClinicsController < ApplicationController
+    def index
+      response = systems_service.get_facility_clinics(
+        current_user,
+        clinics_params[:facility_id],
+        clinics_params[:type_of_care_id],
+        clinics_params[:system_id]
```
omgitsbillryan: Thoughts on using a `Clinic` model?

saneshark: will definitely be doing this for POSTs and PUTs I think where we get a lot of data. Might do something like this as a future refactor for some of the GETs but I think it's going to decide if these types of things can go in as mixins vs poro once we've got more of them built.

PR #3526, opened 2019-11-08
Move nearby endpoint to v0 after schema changes

rtravitz:
## Description of change

This PR finishes [#3597](https://github.com/department-of-veterans-affairs/vets-contrib/issues/3597) and moves `/nearby` from `v1` to `v0`. 

Previously it made sense to have `/nearby` be in `v1` because it was returning a facility object, and we wanted it to return the one with the `lng` attribute and `wait_times` inside of `services`. But now we return a `DrivetimeBand`, which is an entirely new construct. More than that, we also use this "v1" /nearby to link out to routes like /facilities and /facilities/{id}, which currently only exist in v0. We can still make this change without affecting public production consumers.

#### Notable changes
- Remove `v1` documentation and routes
- Add `/nearby` information to `v0` Open API Spec

## Testing done

- Updated the `nearby_request_spec` to hit the v0 route instead

## Testing planned

- Will verify in lower envs

#### Unique to this PR

- [x] Move code from v1 to v0 in the namespaces, change spec and tests
- [x] Add v0 path to kong route
- [x] Update Readme.yml

#### Applies to all PRs
- [x] Appropriate logging


nfasulo:
I think we need the changes from the [Update README.v1.yml PR](https://github.com/department-of-veterans-affairs/vets-api/pull/3371) to be reflected in the v0 documentation.

modules/va_facilities/README.yml
```diff
@@ -424,6 +424,197 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                required:
+                  - data
+                properties:
+                  data:
+                    type: array
+                    items:
+                      $ref: "#/components/schemas/NearbyFacility"
```
nfasulo: The `NearbyFacility` schema object this references needs to be included as well.

rtravitz: editor.swagger.io didn't show the error until refresh 🤦‍♂

PR #3527, opened 2019-11-08
[No review needed] improve Flipper specs to test routes for unauthorized users

jholton:
No review needed.

The target of this PR is another feature branch.  These changes will be reviewed in that branch.


lib/flipper/action_patch.rb
```diff
@@ -12,9 +12,9 @@ def view(name)
 
       raise "Template does not exist: #{path}" unless path.exist?
 
-      contents = path.read
-      compiled = Flipper::UI::Eruby.new(contents)
-      compiled.result proc {}.binding
+      # rubocop:disable Lint/Eval
+      eval(Erubi::Engine.new(path.read, escape: true).src)
+      # rubocop:enable Lint/Eval
```
jholton: [following example in Flipper gem](https://github.com/jnunemaker/flipper/pull/407/files)

spec/request/flipper_request_spec.rb
```diff
@@ -3,25 +3,52 @@
 require 'rails_helper'
 
 RSpec.describe 'Flipper UI', type: :request do
+  before(:all) do
+    Rails.application.routes.draw do
+      mount Flipper::UI.app(
+        Flipper.instance,
+        rack_protection: { except: :authenticity_token }
```
jholton: The addition of [this line](https://github.com/jnunemaker/flipper/blob/fa78a0030c7f139aecc3f9c8468baf9fd1498eb9/lib/flipper/ui.rb#L44) in the latest version makes setting this option possible.

Gemfile
```diff
@@ -45,10 +45,10 @@ gem 'faraday_middleware'
 gem 'fast_jsonapi'
 gem 'fastimage'
 gem 'figaro'
-gem 'flipper'
-gem 'flipper-active_record'
-gem 'flipper-active_support_cache_store'
-gem 'flipper-ui'
+gem 'flipper', '0.17.1'
```
annaswims: This, to me, declares intent to keep it at version `0.17.1` Can we change this to >=

spec/request/flipper_request_spec.rb
annaswims: That's a good find!  

I'm a bit torn over whether we should override the route here or just change it in routes.rb.  Mostly, because if someone were to remove the `constraints` from routes tests would still pass.

jholton: Refactored and added new spec that won't pass if constraint is not present.

spec/request/flipper_request_spec.rb
```diff
@@ -3,25 +3,68 @@
...
     end
 
-    it 'Displays list of features' do
-      get '/flipper/features', params: nil
-      expect(response.body).to include('flipper/features')
-      assert_response :success
+    context 'with LOA1 access' do
+      let(:user) { build(:user) }
+
+      it 'does not allow feature toggling' do
```
jholton: This test does not pass if constraint is not present due to the constraint's LOA3 requirement

PR #3531, opened 2019-11-08
Increase VAOS timeout

kreek:
## Description of change
Increases the VAOS (VA Mobile) read timeout to 30 seconds.

## Testing done
unit tests

## Testing planned
staging integration tests

#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging


saneshark:
hopefully you don't need to call &.to_i on the Settings.yml stuff. It can interpret integers in yaml right?

edmangimelli:
>hopefully you don't need to call &.to_i on the Settings.yml stuff. It can interpret integers in yaml right? Also don't want &.to_i, because nil would become 0 and fail the || 30

it seems like I always end up using `Integer()`, but then you need to put it in a `begin/rescue` which is maybe a bit clunky

edmangimelli:
should a default value (30) live somewhere else? just curious if we had some convention

kreek:
@annaswims @edmangimelli created a devops PR that sets the timeout to 30s for staging https://github.com/department-of-veterans-affairs/devops/pull/5920

modules/vaos/spec/services/configuration_spec.rb
```diff
@@ -32,4 +32,10 @@
       end
     end
   end
+
+  describe '#read_timeout' do
+    it 'has a default timeout of 30 seconds' do
```
annaswims:
```suggestion
    it 'has a default timeout of 15 seconds' do
```

edmangimelli:
https://github.com/department-of-veterans-affairs/vets-api/pull/3531#issuecomment-553577976

noice

PR #3532, opened 2019-11-08
make requests and appointments a bit more consistent and fix params

saneshark:
## Description of change
Fixes requests to use params properly.
Makes appointments service use params more consistently rather than building url.

## Testing done
Local and on staging.

## Testing planned
integration with FE

#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging


modules/vaos/app/services/vaos/appointment_requests_service.rb
```diff
@@ -42,19 +42,16 @@ def deserialize(json_hash)
...
-      url
+    def get_appointment_requests_url
+      "/var/VeteranAppointmentRequestService/v4/rest/appointment-service/patient/ICN/#{user.icn}/appointments"
     end
 
-    def get_date_url(start_date, end_date)
-      { startDate: date_format(start_date), endDate: date_format(end_date) }
+    def date_params(start_date, end_date)
+      { startDate: date_format(start_date), endDate: date_format(end_date) }.compact
     end
```
edmangimelli: This kinda feels like the right time for making a class instead (but I might be missing some context).

saneshark:
There will be a class / model for the POST validations, but introducing a model just for GET params seemed inconsistent with most of the other APIs.

PR #3533, opened 2019-11-11
do not gate feature toggle admin for development

annaswims:
## Description of change
There is no reason to gate feature toggle administration behind a LOA3 login for dvelopment, so don't. 

## Testing done
Screen Shot 2019-11-11 at 5 29 18 PM


lib/flipper/views/features.erb
```diff
@@ -36,8 +36,8 @@
       </tbody>
     </table>
 
-    <% if Thread.current[:flipper_user_email_for_log].blank? %>
-      <a href ="<%= SAML::URLService::VIRTUAL_HOST_MAPPINGS.dig(request.base_url, :base_redirect) %>">Sign in</a> with an identity-verified admin account to exit feagure toggles.
+    <% if Thread.current[:flipper_user_email_for_log].blank? &&  !Rails.env.development? %>
```
johnpaulashenfelter: this and line20 are pretty similar but not the same -- might it make sense to extract all of it into a method (`fipper_enabled?` or whatever makes sense in context) so the logic is all in one place?

annaswims: I had the same thought, but thought it probably wasn't worth a new class/file just for that.  This view [overrides](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/flipper/action_patch.rb) the view of the flipper-ui gem.  Any thoughts were you'd put it?

johnpaulashenfelter: Argh. Helper method? Don't love that either.

PR #3534, opened 2019-11-11
Add clamd and BRD support

nathanhruby:
## Description of change
This adds clamd support into the container and allows for it to be run with foreman.  We do this because by default clamdscan requests clamd to scan a file on the filesystem, and since the container with vets-api and host where clamd run are different file systems, things don't work well. 

Alternatives are
- Change vets-api to use STREAM requests to clamd and bypass fs.
  - This may have performance, PII, and network impacts that would be hard to identify, so I opted to not do that here for simplicity.  I would like revisit this as part of a project to move clamd to a standalone service.
- Make the filesystem for the container be identical to the host system
  - This will get fussy very quickly.  I started down this road, and realized the pathing changes were needed for other prod uses (certs, configs, etc..) but keeping all the various tmpdir uses pointed at the right place seemed like the wrong choice.

## Testing done
- ci
- manual running

## Testing planned
- once we build the container in the AMI and get it deployed to dev, I need to do manually retest before flipping the deploy to use docker.  There will be seperate tickets for this

- clamd runs in container
- freshclam runs in container
- adtional pathing simialr to prod deploys in container to make mounting files into correct locations possible

#### Unique to this PR
- ⚠️ This PR moves the source code again, as a result, will require a rebuild of container and bundler volume.  I've added a `docker-clean` make target to make this easy.

#### Applies to all PRs

- [x] Appropriate logging
  - more foreman logging
  - n/a
  - department-of-veterans-affairs/va.gov-team#2513
  - department-of-veterans-affairs/devops#5901
  - one updated, developers will need to rebuild, will announce this in various slack channels
  - clamd stuff is current unused, no alerts for it currently


nathanhruby:
Ping.  Hi Need a review.

PR #3536, opened 2019-11-12
Messages service and controller

saneshark:
## Description of change
Introduces MessagesService and Controllers as a nested route under appointment_requests
Includes Schema
Includes VCR Cassette
Skipping Swagger for now.

## Testing done
Local and on staging in the console.

## Testing planned
Integration testing on FE.

#### Unique to this PR
See above.

#### Applies to all PRs

- [x] Appropriate logging


johnpaulashenfelter:
@saneshark this is failing because of codecoverage

```
[2019-11-12T05:50:49.558Z] Coverage report generated for RSpec to /srv/vets-api/coverage. 28441 / 28674 LOC (99.19%) covered.

[2019-11-12T05:50:49.558Z] File (/srv/vets-api/modules/vaos/app/services/vaos/messages_service.rb) is only (88.89%) covered. This is below the expected minimum coverage per file of (90.00%).

```

modules/vaos/app/services/vaos/messages_service.rb
```diff
@@ -0,0 +1,53 @@
...
+    STATSD_KEY_PREFIX = 'api.vaos'
+
+    def initialize(user, request_id)
+      @user = user
+      @request_id = request_id
+    end
+
+    def get_messages
+      with_monitoring do
+        response = perform(:get, messages_url, headers)
```
LindseySaari: Wondering if the response portion of this code could be refactored out into a response class for uniformity purposes?

saneshark:
> @saneshark this is failing because of codecoverage
> 
> ```
> [2019-11-12T05:50:49.558Z] Coverage report generated for RSpec to /srv/vets-api/coverage. 28441 / 28674 LOC (99.19%) covered.
> 
> [2019-11-12T05:50:49.558Z] File (/srv/vets-api/modules/vaos/app/services/vaos/messages_service.rb) is only (88.89%) covered. This is below the expected minimum coverage per file of (90.00%).
> ```

@johnpaulashenfelter how can I view coverage? It's no longer in simplecov and codeclimate is showing it passed.

modules/vaos/app/services/vaos/messages_service.rb
saneshark: There might be an opportunity for doing a custom base class that makes this more DRY, but so that "with_monitoring" for example happens automatically, but I think I favor composability over abstraction here.

Can you point me to examples of where response was abstracted into a class in other services/clients?

PR #3537, opened 2019-11-12
Add VAOS feature flags

jbalboni:
## Description of change
This adds the VAOS feature flags that we're going to need over the next month in order to roll out the application. It also adds users that can access the UI to change the flags.

## Testing done
None

## Testing planned
Verify the flags in staging

#### Applies to all PRs

- [ ] Appropriate logging

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -405,7 +405,6 @@ flipper:
     - jeff@adhocteam.us
     - kam@adhocteam.us
     - alastair@adhocteam.us
-    - mark.greenburg@adhocteam.us
     - narin@adhocteam.us
 
 bgs:
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -406,6 +406,7 @@ flipper:
     - kam@adhocteam.us
     - alastair@adhocteam.us
     - mark.greenburg@adhocteam.us
+    - narin@adhocteam.us
 
 bgs:
   application: ~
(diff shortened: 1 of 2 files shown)


config/settings.yml
```diff
@@ -401,6 +401,12 @@ flipper:
     - travis.hilton@oddball.io
     - rian.fowler@adhoc.team
     - rianfowler@outlook.com
+    - jbalboni@gmail.com
```
rianfowler: @annaswims We don't have a process for updating this list but this seems fine for now. you mentioned off-boarding this morning and cleaning up this list could be part of that

saneshark: I'd like to add a few others to this list for redundancy. 

@kreek @markgreenburg @lalexanderson-dsva do you all know what your ID.me wallet email addresses are for Staging and Production environments? Can you also double check and make sure that these accounts are properly secured and do not use "throw-away" password? If you do not want to be added as an admin, just reply back to this thread with opt-out, otherwise provide:

Production: kam@adhocteam.us
Staging: kam@adhocteam.us

saneshark: on second thought -- would be nice to have this in staging asap, can add additional redundancy in a future PR.

annaswims: Do the id.me users exist for all of these users in both staging and prod?  Email validation can be bypassed in staging, which plausibly could cause an issue.

PR #3540, opened 2019-11-13
VAOS service error middleware

kreek:
## Description of change
Moves VAOS service error handling to middleware.

## Testing done
unit tests

## Testing planned
staging integrations tests

#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging

### DIFF
--- a/modules/vaos/app/services/vaos/middleware/response/errors.rb
+++ b/modules/vaos/app/services/vaos/middleware/response/errors.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+module VAOS
+  module Middleware
+    module Response
+      class Errors < Faraday::Response::Middleware
+        def on_complete(env)
+          return if env.success?
+
+          caller = "VAOS.#{caller_locations(14, 1)[0].label}"
+          Raven.extra_context(message: env.body, url: env.url)
+          case env.status
+          when 403
+            raise Common::Exceptions::BackendServiceException.new(
+              'VAOS_403',
+              source: caller
+            )
+          when 500
+            raise Common::Exceptions::BackendServiceException.new(
+              'VAOS_502',
+              source: caller
+            )
+          else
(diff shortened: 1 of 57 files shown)


modules/vaos/app/services/vaos/middleware/response/errors.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module VAOS
+  module Middleware
+    module Response
+      class Errors < Faraday::Response::Middleware
+        def on_complete(env)
+          return if env.success?
+
+          caller = "VAOS.#{caller_locations(14, 1)[0].label}"
```
kreek: this gets the calling method e.g. VAOS.get_systems don't know if there's a better way to get the class and method that called it?

saneshark: This is kind of nice, but is it really necessary? I would think the backtrace should tell us what we need.

modules/vaos/app/services/vaos/middleware/response/errors.rb
```diff
@@ -0,0 +1,35 @@
...
+          case env.status
+          when 403
+            raise Common::Exceptions::BackendServiceException.new(
+              'VAOS_403',
+              source: caller
+            )
+          when 500
+            raise Common::Exceptions::BackendServiceException.new(
+              'VAOS_502',
+              source: caller
```
saneshark: these are going to get pretty unruly fast... do we want a range of 500s? Also does breakers still have an opportunity to do it's thing or does this middleware circumvent that?

kreek: it's after breakers in the chain

saneshark:
This should probably be merged for production testing to begin. Gives much more robust error handling and makes Sentry debugging easier.

PR #3541, opened 2019-11-13
[2454] Add Transaction ID

camerontesterman:
[Link to Originating Issue](https://github.com/department-of-veterans-affairs/va.gov-team/issues/2454)

## Description of change
Adds a transaction_id to every request sent by a service that inherits from the EVSS::Service base class. Said transaction_id will also be reported into sentry if a request fails for some reason.

This will allow for easier tracking of failed of requests, as we will be able to use the transaction ID logged into sentry since that same transaction ID will be sent to caseflow, allowing for 1-to-1 matching of the requests.

## Testing done
This branch has been deployed to staging, was able to successfully trigger an error which displayed the appropriate transaction_id within sentry.

Transaction IDs are sent with every request sent by an EVSS::Service inherited class service, and every transaction_id is logged to sentry upon failure.

#### Applies to all PRs

- [x] Appropriate logging


camerontesterman:
> LGTM.
> 
> Only note I would provide is that instead of removing tests I would try to find a work around. Even though the transaction id is a random uuid we want to make sure we dont lose test functionality. For example, we want to make sure the rest of the auth headers are still being built properly even if we cant verify that the uuid is the EXACT one that the AuthHeaders object built.

I agree with this, this change does add a complicating factor that will require rewriting some of the tests. I will make sure this effort gets captured in a ticket and prioritized accordingly.

PR #3543, opened 2019-11-13
Thread safety rubocop plugin

omgitsbillryan:
## Description of change
Adds a thread safety rubocop plugin to our linter.

## Testing done
Run the lint locally using the command to only look for thread safety issues.
```
rubocop -r rubocop-thread_safety --only ThreadSafety,Style/GlobalVars,Style/ClassVars,Style/MutableConstant
```
#### Applies to all PRs

- [ ] Appropriate logging


johnpaulashenfelter:
@omgitsbillryan just curious if we've had any problems from thread safety in the past that this would pick up

annaswims:
I like it!  How about committing the results of `rubocop --auto-gen-config` so that we can start checking future PR's right away and address existing problems one-by-one?

Gemfile
```diff
@@ -4,6 +4,8 @@ source 'https://rubygems.org'
 
 ruby '2.4.5'
 
+gem 'rubocop-thread_safety'
```
jholton: Should this be moved to the `group :development, :test` block?

edmangimelli:
I'm sad by how little this readme tells me
https://github.com/covermymeds/rubocop-thread_safety

omgitsbillryan:
The jenkins build passed 🎉 

But it looks like code climate is unable to load the plugin 🤔 

```
/usr/local/lib/ruby/site_ruby/2.4.0/rubygems/core_ext/kernel_require.rb:117:in `require': cannot load such file -- rubocop-thread_safety (LoadError)
```

annaswims:
ugh!  So, codeclimate doesn't run the version of rubocop from the gemfile.  It does it's own nonsense and  rubocop-thread_safety isn't in their gemfile https://github.com/codeclimate/codeclimate-rubocop/blob/channel-rubocop-0.71/Gemfile 

I could see ditching codeclimate for this.

annaswims:
different rubocop gem, similar issue https://github.com/codeclimate/codeclimate-rubocop/issues/131

omgitsbillryan:
It might be possible to execute the thread safety cop in a separate Jenkins stage purely on the command line rather than include it as a plugin in our `.rubocop.yml` config. This would give us the best of both worlds - except that thread safety exceptions won't show up in our codeclimate report.

johnpaulashenfelter:
We already run rubocop on CI, so why are we bothering running it again on CodeClimate, especially when it adds the "keep versions in sync" issue. 

Isn't rubocop, for our purposes on the platform, like specs -- pass/fail?

PR #3548, opened 2019-11-13
accidentally broke dates here

saneshark:
## Description of change
This is a really important fix that needs to go in before testing can begin tomorrow. The date format for params sent was incorrect. This fixes that.

## Testing done
manual regression identified the problem

## Testing planned
more manual regression

#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging


saneshark:
@johnpaulashenfelter I thought as a general rule you're not supposed to write specs for private methods. An artifact of how VCR cassettes ignore params is at fault.

kfrz:
@saneshark I do think this could have benefited from a unit spec, but really the issue is the `date_params()` helper which could arguably be tested against. 

Alternatively a spec that exercises the date behavior could do some input validation which would be verifying behavior:

quick example off-top

```rb
    context 'with junk input for dates' do
      it 'does the right thing' do
        VCR.use_cassette('vaos/appointment_requests/get_requests', match_requests_on: %i[host path method]) do
          expect { subject.get_requests('junk', 'junk') }.to raise NoMethodError
        end
      end
    end

```

saneshark:
I think we could tell VCR to match on URI which would include the changes to params we send over in the request. 

```
match_requests_on: %i[host path method])

to

match_requests_on: %i[host uri method])

saneshark:
You'll notice I added an additional spec and updated all vcr cassettes elsewhere in the project to also throw an error based on query parameter changes. They will better match the request generated by the service.

PR #3552, opened 2019-11-14
Cancel appointment

saneshark:
## Description of change
Added Cancel Appointments Form object.
Added Cancel Appointments service endpoint
Added Cancel Appointment Routes and Controller endpoint.

## Testing done
VCR cassettes and staging manual testing in console.

## Testing planned
Integration testing. 

#### Unique to this PR

#### Applies to all PRs

- [x] Appropriate logging


config/initializers/bgs.rb
```diff
@@ -4,7 +4,7 @@
 
 LighthouseBGS.configure do |config|
   config.application = Settings.bgs.application
-  config.client_ip = Socket.ip_address_list.detect(&:ipv4_private?).ip_address
+  config.client_ip = Socket.ip_address_list.detect(&:ipv4_private?)&.ip_address
```
kreek: Does this apply to this PR? Maybe a rubocop thing?

modules/vaos/app/services/vaos/appointment_service.rb
```diff
@@ -21,17 +21,29 @@ def self.for_user(user)
...
       end
     end
 
+    def put_cancel_appointment(request_object_body)
+      params = VAOS::CancelForm.new(request_object_body).params
+      params.merge!(patient_identifier: { unique_id: user.icn, assigning_authority: 'ICN' })
+
+      with_monitoring do
+        perform(:put, put_appointment_url, params, headers(user))
+        ''
```
kreek: Could we return `true` for success if there's no object to return?

modules/vaos/spec/request/appointments_request_spec.rb
```diff
@@ -10,123 +10,206 @@
...
+                'tly to cancel your appointment. <a class="external-link" href="https://www.va.gov/find-locations/">V' \
+                'A Facility Locator</a>')
+          end
+        end
+      end
 
-        expect(response).to have_http_status(:success)
-        expect(response.body).to be_a(String)
-        expect(response).to match_response_schema('vaos/cc_appointments')
+      context 'when appointment can be canncelled' do
```
kreek:
```suggestion
      context 'when appointment can be cancelled' do
```

modules/vaos/app/services/vaos/appointment_service.rb
saneshark: we could, but I figured we usually return JSON or raise an exception and this was at least closer to returning to JSON than boolean.

config/initializers/bgs.rb
saneshark: no, it was just something I caught in testing with vets-api while not having an internet connection. I could remove it from this PR and do a separate one if it's a problem.

johnpaulashenfelter: I'm not sure letting it be `nil` is helpful. Pretty much an edge case imho when you don't have an internet connection -- no different than changing it to a one-off URL.

Maybe a default would be a better choice here (eg `|| 127.0.0.1`)

modules/vaos/app/models/vaos/appointment_time.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module VAOS
+  # This is a custom type class for ensuring Time is always coerced as UTC
+  class AppointmentTime < Virtus::Attribute
```
johnpaulashenfelter: We're moving away from Virtus since its deprecated. Maybe go ahead and use `dry-types` instead to avoid creating more debt?

config/initializers/bgs.rb
saneshark: k, maybe i'll pull it out right now and can be addressed in a different PR.

saneshark: pulled in most recent commit. @johnpaulashenfelter

modules/vaos/app/models/vaos/appointment_time.rb
saneshark: Yes, I've noticed that, but we've already used virtus all over the place and I don't have the time to grok dry-types right now. Will have to revisit as code debt unfortunately.

saneshark: This feature along with its in progress, sister PR, for cancelling an appointment request ideally should have been wrapped up by last week. I'm a little behind and will be out of pocket for the next couple days. I'm happy to open a ticket to make sure it's something we address along side swagger stuff that needs to be done too post launch.

saneshark: https://github.com/department-of-veterans-affairs/va.gov-team/issues/3546

saneshark: Swagger: https://github.com/department-of-veterans-affairs/va.gov-team/issues/3545

modules/vaos/app/services/vaos/configuration.rb
```diff
@@ -17,13 +17,18 @@ def service_name
     def connection
       Faraday.new(base_path, headers: base_request_headers, request: request_options) do |conn|
         conn.use :breakers
-
+        conn.request :camelcase
         conn.request :json
 
+        if ENV['VAOS_DEBUG'] && !Rails.env.production?
```
johnpaulashenfelter: We don't normally have debugging like this in modules. Is this a temporary thing that should have been removed or is it in for dealing with issues during UAT?

Is there a devops PR that goes with this ENV var?

modules/vaos/config/routes.rb
```diff
@@ -2,7 +2,9 @@
 
 VAOS::Engine.routes.draw do
   defaults format: :json do
-    resources :appointments, only: :index
+    resources :appointments, only: :index do
+      put 'cancel', on: :collection
```
johnpaulashenfelter: 🤔  I'm back and forth on a separate cancel as opposed to the more semantic destory/DELETE. 

Maybe a refactor towards this being an `AppointmentCancellation` controller if its substantially different from an appointment (which based on the index action it is pretty different)

modules/vaos/spec/request/appointments_request_spec.rb
```diff
@@ -10,123 +10,206 @@
...
+      context 'when request body validation fails' do
+        it 'returns validation failed' do
+          put '/v0/vaos/appointments/cancel'
+
+          expect(response).to have_http_status(:unprocessable_entity)
+          expect(JSON.parse(response.body)['errors'].size).to eq(2)
+        end
+      end
+
+      context 'when appointment cannot be cancelled' do
```
johnpaulashenfelter: what can't this be cancelled? because the params are wrong?

modules/vaos/spec/services/appointment_service_spec.rb
```diff
@@ -5,12 +5,55 @@
 describe VAOS::AppointmentService do
   subject { described_class.for_user(user) }
 
-  let(:user) { build(:user, :mhv) }
+  let(:user) { build(:user, :vaos) }
```
johnpaulashenfelter: The reason for the change from `mhv` to `vaos` isn't really clear -- there's no new factory and it seems like a lot of other places in the spec its a `mhv` user. Is there a consistent pattern for the kind of user in these specs?

modules/vaos/spec/services/appointment_service_spec.rb
```diff
@@ -5,12 +5,55 @@
...
+      it 'returns the bad request with detail in errors' do
+        VCR.use_cassette('vaos/appointments/put_cancel_appointment_400', match_requests_on: %i[method uri]) do
+          expect { subject.put_cancel_appointment(request_body) }.to raise_error(
+            Common::Exceptions::BackendServiceException
+          )
+        end
+      end
+    end
+
+    context 'when appointment can be cancelled' do
```
johnpaulashenfelter: Still not clear to me what makes the difference here between can and can't cancel.

modules/vaos/app/services/vaos/appointment_service.rb
saneshark: the idea is that we want this controller action to return no response body -- the question here is what should the service return...

if we were passing around response objects with status and body, then that would be ideal. The general pattern has been to pass the response body. In this case there is none. So I'm thinking Nil or '' makes the most sense.

modules/vaos/app/models/vaos/appointment_time.rb
```diff
@@ -0,0 +1,14 @@
...
+    def coerce(value)
```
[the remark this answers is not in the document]
saneshark: The underlying API that we're building against have about 7 different date time formats. This and other custom attributes are in an effort to wrangle those so that our front-end only needs to send 1-2 variations at most.

modules/vaos/spec/services/appointment_service_spec.rb
```diff
@@ -5,12 +5,55 @@
...
+          cancel_code: 'PC',
+          remarks: '',
+          clinic_name: 'CHY OPT VAR1'
+        }
+      end
+
+      it 'cancels the appointment' do
+        VCR.use_cassette('vaos/appointments/put_cancel_appointment', match_requests_on: %i[method uri]) do
+          response = subject.put_cancel_appointment(request_body)
+          expect(response).to be_an_instance_of(String).and be_empty
```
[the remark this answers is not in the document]
saneshark: This particular response IS successful. An exception would have been raised if the service was not successful. This is a fairly common that virtually every service in vets-api uses. The question before was whether the service should return true vs ''. In truth, it should probably return a response object and leave it to the controller to decide. The pattern has been to return the response body. In this case the most appropriate response body for a controller action that will return no body is empty string.

modules/vaos/app/services/vaos/configuration.rb
saneshark: it's typically commented out... but with the amount of debugging necessary already for testing this, this is useful. You can see this used in Rx and SM configurations, but there it's commented out. ENV flag makes it much easier to run this on a staging machine.

modules/vaos/config/routes.rb
saneshark: I agree, but those actions typically take an ID associated with the object being deleted. They are member actions. This is a collection action, and there is no actual ID. A synthetic id might be some combination of appointmentTime and clinicID, but the underlying API doesn't treat it at that way, they expect a full JSON payload to be processed.

modules/vaos/spec/request/appointments_request_spec.rb
saneshark: because it is the response the underlying API returns as recorded in that particular cassette.

modules/vaos/spec/services/appointment_service_spec.rb
saneshark: the rest will eventually use vaos... trying to keep this change incremental.

modules/vaos/spec/services/appointment_service_spec.rb
saneshark: We are writing a request spec that tests that the exception that is raised from the service, properly renders the error that is under test here. In particular, that detail matches what the service returned in the response body, that our middlewares effectively handled things.

modules/vaos/spec/services/appointment_service_spec.rb
```diff
@@ -5,12 +5,55 @@
...
+          appointment_time: '11/15/19 20:00:00',
+          clinic_id: '408',
+          cancel_reason: 'whatever',
+          cancel_code: '5',
+          remarks: nil,
+          clinic_name: nil
+        }
+      end
+
+      it 'returns the bad request with detail in errors' do
```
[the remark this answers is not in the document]
saneshark: I think it's clear if you look at what's different on the next line where the cassette is referenced. 🤷‍♂

config/locales/exceptions.en.yml
```diff
@@ -474,6 +474,12 @@ en:
         code: 'SEARCH_429'
         detail: 'Exceeded Search.gov rate limit'
         status: 429
+      VAOS_400:
+        <<: *external_defaults
+        title: Bad Request
+        code: 'VAOS_400'
+        detail: # deliberarly left nil to reference response.body which is unparsable string (not json)
```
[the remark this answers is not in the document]
saneshark: good catch

modules/vaos/app/models/vaos/appointment_time.rb
saneshark: I'm seeing

https://github.com/department-of-veterans-affairs/va.gov-team/issues/466
https://github.com/department-of-veterans-affairs/va.gov-team/issues/816

Is there why you feel there needs to be a greater sense of urgency for our team to address this? I certainly see value in doing making the switch for longer term maintainability, but I am having a hard time understanding the degree of urgency you're attributing here.

PR #3553, opened 2019-11-14
Working to reduce db migration noise

jbritt1:
**Refs: https://github.com/department-of-veterans-affairs/va.gov-team/issues/2514 && https://github.com/department-of-veterans-affairs/devops/pull/5923**

Working to reduce the duplication of vets-api-worker and vets-api-server warnings for database migrations (we think just having server post will suffice). Setting the parameter for migration_status on the apps specifically for deploys/releases will help accomplish this.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -112,11 +112,13 @@ pipeline {
       steps {
         build job: 'deploys/vets-api-server-vagov-dev', parameters: [
           booleanParam(name: 'notify_slack', value: true),
+          booleanParam(name: 'migration_status', value: true),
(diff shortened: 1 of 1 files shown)


jbritt1:
> can we just automatically run migrations after a deploy? I don't see a benefit to an extra manual step.

Anna I believe I thought the same thing, but it seemed like something we had approached that perhaps didn't quite work out in the future? Or perhaps a project that has been back-burner'd for now? Here is a link to the thread about it: https://dsva.slack.com/archives/C7S6EA0ES/p1571852022231600.

PR #3554, opened 2019-11-14
Adding Uploading Capability to 526 pdf

bastosmichael:
## Description of change
Adding document functionality to 526 for ticket https://github.com/department-of-veterans-affairs/vets-contrib/issues/3435

## Testing done
- Rspec testing

## Testing planned
- Will test in dev-api.va.gov

- Create the ability to upload a specific 526 form through a PUT end point.

#### Applies to all PRs

- [ ] Appropriate logging


modules/claims_api/app/models/claims_api/auto_established_claim.rb
```diff
@@ -64,5 +65,23 @@ def set_md5
     def status_from_phase(*)
       status
     end
+
+    def file_name
```
ATeal: I feel like these same methods are duplicated all over the place in claims api module...could you refactor?

Gemfile.lock
```diff
@@ -501,7 +501,8 @@ GEM
     rack (2.0.7)
     rack-attack (5.0.1)
       rack
-    rack-cors (0.4.1)
+    rack-cors (1.0.6)
```
LindseySaari: Wondering if update for rack and nokogiri should be in a separate PR?

bastosmichael: There is a separate PR for the gemfile updates but I also included them in here because they were causing a CI/CD failure for this PR https://github.com/department-of-veterans-affairs/vets-api/pull/3560/files

modules/claims_api/app/models/claims_api/concerns/file_data_validation.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module ClaimsApi
+  module FileDataValidation
```
ATeal: this doesn't actually validate anything so maybe a different name?

modules/claims_api/app/models/claims_api/auto_established_claim.rb
bastosmichael: Refactored.

modules/claims_api/app/models/claims_api/concerns/file_data_validation.rb
bastosmichael: Just calling it FileData then...

PR #3558, opened 2019-11-15
VAOS get preferences endpoint

kreek:
## Description of change
Adds the `/v0/vaos/preferences` endpoint and backing service

## Testing done
unit tests

## Testing planned
staging integration tests

#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging

### DIFF
--- a/modules/vaos/app/controllers/vaos/preferences_controller.rb
+++ b/modules/vaos/app/controllers/vaos/preferences_controller.rb
@@ -5,13 +5,13 @@ require_dependency 'vaos/application_controller'
 module VAOS
   class PreferencesController < ApplicationController
     def index
-      response = preference_service.get_preferences(current_user)
+      response = preferences_service.get_preferences(current_user)
       render json: VAOS::PreferenceSerializer.new(response)
     end
 
     private
 
-    def preference_service
+    def preferences_service
       VAOS::PreferencesService.new
     end
   end
--- a/modules/vaos/app/controllers/vaos/preferences_controller.rb
+++ b/modules/vaos/app/controllers/vaos/preferences_controller.rb
@@ -12,7 +12,7 @@ module VAOS
     private
 
     def preference_service
-      VAOS::PreferenceService.new
+      VAOS::PreferencesService.new
     end
   end
 end
--- a/modules/vaos/app/controllers/vaos/preferences_controller.rb
+++ b/modules/vaos/app/controllers/vaos/preferences_controller.rb
@@ -6,7 +6,7 @@ module VAOS
   class PreferencesController < ApplicationController
     def index
       response = preferences_service.get_preferences(current_user)
-      render json: VAOS::PreferenceSerializer.new(response)
+      render json: VAOS::PreferencesSerializer.new(response)
     end
 
     private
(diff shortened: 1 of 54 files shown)


modules/vaos/spec/services/preference_service_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  before { allow_any_instance_of(VAOS::UserService).to receive(:session).and_return('stubbed_token') }
+
+  describe '#get_preferences' do
+    context 'with a 200 response' do
+      it 'includes' do
+        VCR.use_cassette('vaos/preferences/get_preferences', match_requests_on: %i[method uri]) do
+          response = subject.get_preferences(user)
+          expect(response.notification_frequency).to eq('Never')
+          expect(response.email_allowed).to be_truthy
+          expect(response.email_address).to eq('Vilasini.reddy@va.gov')
```
saneshark: Please remove real email address from source control. Can use a fake one instead. Spam bots will find it if its checked in. Can probably remove and rebase.

modules/vaos/app/services/vaos/preference_service.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+require_relative '../vaos/concerns/headers'
+
+module VAOS
+  class PreferenceService < Common::Client::Base
```
saneshark: nitpick by why is this singular?

modules/vaos/app/controllers/vaos/preferences_controller.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+require_dependency 'vaos/application_controller'
+
+module VAOS
+  class PreferencesController < ApplicationController
```
saneshark: here it's plural... a user can only have 1 right? Maybe it needs to be a singular resource?

Show instead of Index?

kreek: If we change it to be show, get single preference, what do we use for the id? I think we'd have to make a custom route.

saneshark: oh the preference also has an id? Shouldn't the ID just be the patient identifier / icn??

saneshark: i think i'm ok with index, since "preferences" reads better anyway. Lets just make it consistent maybe.

modules/vaos/app/controllers/vaos/preferences_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+    def index
+      response = preference_service.get_preferences(current_user)
+      render json: VAOS::PreferenceSerializer.new(response)
+    end
+
+    private
+
+    def preference_service
```
saneshark: still singular here

modules/vaos/app/controllers/vaos/preferences_controller.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+require_dependency 'vaos/application_controller'
+
+module VAOS
+  class PreferencesController < ApplicationController
+    def index
+      response = preference_service.get_preferences(current_user)
+      render json: VAOS::PreferenceSerializer.new(response)
```
saneshark: still singular here

PR #3563, opened 2019-11-19
Vbms eservice upload

charleystran:
## Description of change
This pull request is to add logic to the ClaimsApi Power of attorney update endpoint to upload the submitted file to the VBMS eFolder service

## Testing done
rspec and local testing against lower level vbms envs

- [ ] The document uploaded to the claimsAPI is uploaded to VBMS and can be verified in the VBMS system.

#### Applies to all PRs

- [ ] Appropriate logging


modules/claims_api/app/swagger/claims_api/form_2122_model_swagger.rb
```diff
@@ -19,19 +19,19 @@ class Form2122ModelSwagger
         key :description, 'Required by JSON API standard'
         key :required, %i[poa_code poa_first_name poa_last_name]
 
-        property :poa_code do
+        property :poaCode do
```
bastosmichael: Why change it to camel case instead of underscore?

charleystran: as per our offline convo. This change was because the docs didnt match the schema. The schema uses camelCase.

modules/claims_api/app/workers/claims_api/vbms_uploader.rb
```diff
@@ -0,0 +1,69 @@
...
+
+    def perform(power_of_attorney_id)
+      power_of_attorney = ClaimsApi::PowerOfAttorney.find(power_of_attorney_id)
+      uploader = ClaimsApi::PowerOfAttorneyUploader.new(power_of_attorney_id)
+      uploader.retrieve_from_store!(power_of_attorney.file_data[:filename])
+      filepath = "#{uploader.file.file}/#{power_of_attorney.file_data['filename']}"
+
+      upload_token_response = fetch_upload_token(
+        filepath: filepath,
+        file_number: power_of_attorney.auth_headers['va_eauth_pnid']
```
ATeal: I know if the benefits intake api it can be file number or ssn, is it always ssn in this case or should we support a filenumber input?

modules/claims_api/db/migrate/20191118152927_add_vbms_id_fields_to_claims_api_power_of_attorney.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddVbmsIdFieldsToClaimsApiPowerOfAttorney < ActiveRecord::Migration[5.2]
+  def change
+    add_column :claims_api_power_of_attorneys, :vbms_new_document_version_ref_id, :string
```
ATeal: do you think we should serialize any of these fields back out to the conumer or are we storing them meerly for tracking purposes?

Gemfile.lock
```diff
@@ -90,15 +90,15 @@ PATH
       sidekiq
 
 PATH
-  remote: modules/veteran_verification
+  remote: modules/veteran
```
ATeal: is it just me or do these change around on like every pr??

Gemfile
```diff
@@ -36,7 +36,7 @@ gem 'breakers'
 gem 'carrierwave-aws'
 gem 'clam_scan'
 gem 'config'
-gem 'connect_vbms', git: 'https://github.com/department-of-veterans-affairs/connect_vbms.git', branch: 'master'
+gem 'connect_vbms', git: 'https://github.com/department-of-veterans-affairs/connect_vbms.git', branch: 'master', require: 'vbms'
```
ATeal: oh that naming convention

modules/claims_api/app/workers/claims_api/vbms_uploader.rb
charleystran: Its a good question. The vbms gem asks for a file number, but I am not sure if under the hood they allow either.

modules/claims_api/db/migrate/20191118152927_add_vbms_id_fields_to_claims_api_power_of_attorney.rb
charleystran: My thought was for debugging because you would need access to the VBMS app to do anything with them. Mainly it was so we could verify with VBMS when a file we said was uploaded was actually uploaded. However, I don't think it would hurt to send them in the status endpoint.

modules/claims_api/app/workers/claims_api/vbms_uploader.rb
ATeal: Hrm who can we talk to to figure this out, seems like it would be relevant

modules/claims_api/db/migrate/20191118152927_add_vbms_id_fields_to_claims_api_power_of_attorney.rb
ATeal: yeah that makes sense, dunno how much value it would provide to the consumer

charleystran: cool cool. We can always add them down the line if something useful comes up

config/settings.yml
```diff
@@ -442,3 +442,13 @@ va_mobile:
   mock: false
   key_path: /fake/client/key/path
   timeout: 15
+
+vbms:
+  url: "https://internal-dsva-vagov-dev-fwdproxy-1893365470.us-gov-west-1.elb.amazonaws.com:4449"
+  saml: "vetsapi.client.vbms.aide.oit.va.gov.saml-token.xml"
+  keypass: "{{ lookup('credstash', 'vbms.dev.key_password') }}"
+  ca_cert: VBMS-Client-Signing-CA.crt
```
bastosmichael: if these certs are fake then I would include the work fake somewhere in the title so that anyone looking at this won't think we somehow accidentally exposed a cert without context.

charleystran: These are the names of real certs, but not the certs themselves. They are dropped on the box via the ansible deploy script. What I believe @ATeal was referring to was the fake certs in the spec fixtures folder.

ATeal: ^ Yes I was

PR #3565, opened 2019-11-19
1816 facilities refactor 2

LindseySaari:
## Description of change
This change addresses namespacing and cleanup of duplicate code within the facilities directory (lib/facilities). In order to better understand separation of facility services, namespacing improvements were introduced. 

## Testing done
Test suite passing

#### Unique to this PR
- [x] Passing test suite

#### Applies to all PRs
- [ ] Appropriate logging


spec/lib/facilities/drivetime_bands/response_spec.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require 'facilities/ppms/response'
```
edmkitty: Why is the ppms response being required here?

lib/facilities/drivetime_bands/response.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require 'common/models/base'
+
+module Facilities
+  module DrivetimeBands
+    class Response < Common::Base
```
edmkitty: What is the purpose of this class? I don't see where it's used

spec/lib/facilities/drivetime_bands/response_spec.rb
LindseySaari: Good catch, definitely not needed 👍

lib/facilities/drivetime_bands/response.rb
LindseySaari: Ah, looks like a change of mine got squashed 😞I ran into a big merge that I had to go through manually. Thanks for this catch 🙇‍♀

lib/facilities/drivetime_bands/response.rb
```diff
@@ -0,0 +1,23 @@
...
+      attribute :body, String
+
+      def initialize(body)
+        self.body = body
```
edmkitty: You could move the json parse step to this initialize function and store the json as another instance variable. I think that would simplify some of the code in the client

lib/facilities/drivetime_bands/client.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    # Core class responsible for api interface operations
+    class Client < Common::Client::Base
+      configuration Facilities::DrivetimeBands::Configuration
+
+      def get_drivetime_bands(offset, limit)
+        params = build_params(offset, limit)
+        response = perform(:get, '/arcgis2/rest/services/Portal/MonthlyVAST_TTB/FeatureServer/0/query', params)
+        drivetime_band_response = Facilities::DrivetimeBands::Response.new(response.body)
+        response_body = drivetime_band_response.parsed_json
```
edmkitty: This line is unnecessary now

LindseySaari: response_body = drivetime_band_response.parsed_json is referring to an instance variable which now equals the parsed json response body

edmkitty: Instead of reassigniing `drivetime_band_response.parsed_json` to another instance variable, you could use `drivetime_band_response.parsed_json` in the error checking step. I think this would make the code more cohesive, since `drivetime_band_response` is used again later in the method

LindseySaari: Ah, I see what you're saying now 😄

omgitsbillryan:
I think you could break this up into multiple PR's.

The namespace changes look fairly straight forward, but I find I'm getting some of that worked mixed up with the other refactor you've done to DRY up some other code.

LindseySaari:
Breaking this out into separate PRs

PR #3566, opened 2019-11-19
VAOS system facilities endpoint

kreek:
## Description of change
Adds the `/v0/vaos/systems/:id/facilities` endpoint which returns facilities available for direct scheduling.

## Testing done
unit tests

## Testing planned
staging integration tests

#### Unique to this PR
- [ ] adds direct scheduling facilities controller
- [ ] adds get_system_facilities method to the systems service

#### Applies to all PRs

- [ ] Appropriate logging


modules/vaos/app/services/vaos/systems_service.rb
```diff
@@ -18,6 +18,22 @@ def get_systems(user)
       end
     end
 
+    def get_system_facilities(user, system_id, parent_code, type_of_care_id)
+      with_monitoring do
+        url = '/var/VeteranAppointmentRequestService/v4/rest/direct-scheduling/institutions'
```
johnpaulashenfelter: Hhere you're inlining the URL in the method. In @saneshark's earlier PR it was in a method.

Might there be a way to configure this either consistently in the module *or* register all the URLs all in one place?

modules/vaos/config/routes.rb
```diff
@@ -8,7 +8,9 @@
     resources :appointment_requests, only: :index do
       resources :messages, only: :index
     end
-    resources :systems, only: :index
+    resources :systems, only: :index do
+      resources :facilities, only: :index, controller: 'direct_scheduling_facilities'
```
johnpaulashenfelter: Its minor, but seems that since you've got a specific "kind" of facility here in this controller (direct_scheduling) it feels like there are other kinds (indirect_scheduling? who knows). Mapping Facilities to DirectSchedulingFacilities seems like a problem unless there are no other kinds of facilities.

Basically really wondering why you just didn't do `direct_scheduling_facilities` as the resource name or rename the controller.

modules/vaos/spec/request/systems_request_spec.rb
```diff
@@ -25,63 +25,80 @@
...
 
-    context 'with a valid GET systems response' do
-      it 'returns a 200 with the correct schema' do
-        VCR.use_cassette('vaos/systems/get_systems', match_requests_on: %i[method uri]) do
-          get '/v0/vaos/systems'
-
-          expect(response).to have_http_status(:ok)
-          expect(response.body).to be_a(String)
-          expect(response).to match_response_schema('vaos/systems')
+    describe 'GET /v0/vaos/systems' do
```
johnpaulashenfelter: 👍 definitely clearer in this diff

modules/vaos/spec/request/systems_request_spec.rb
```diff
@@ -25,63 +25,80 @@
...
-          expect(response).to have_http_status(:forbidden)
-          expect(JSON.parse(response.body)['errors'].first['code']).to eq('VAOS_403')
-          expect(response).to match_response_schema('errors')
+      context 'with a 403 response' do
+        it 'returns a VAOS 403 error response' do
+          VCR.use_cassette('vaos/systems/get_systems_403', match_requests_on: %i[method uri]) do
+            get '/v0/vaos/systems'
+
+            expect(response).to have_http_status(:forbidden)
+            expect(JSON.parse(response.body)['errors'].first['code']).to eq('VAOS_403')
```
johnpaulashenfelter: Would a helper method here DRY things up? Something like

```
def response_error(response)
	JSON.parse(response.body)['errors'].first['code'])
end
```

Would make it a lot simpler to read

johnpaulashenfelter: Looks like someone already did something similar with the response schema matcher below.

modules/vaos/spec/services/systems_service_spec.rb
```diff
@@ -110,4 +110,25 @@
       end
     end
   end
+
+  describe '#get_system_facilities' do
+    context 'with a 200 response' do
+      it 'returns an array of size 6' do
```
johnpaulashenfelter: What does 6 mean here? A translation of the magic number might help someone understand in the future.

modules/vaos/app/controllers/vaos/direct_scheduling_facilities_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+require_dependency 'vaos/application_controller'
+
+module VAOS
+  class DirectSchedulingFacilitiesController < ApplicationController
+    def index
+      response = systems_service.get_system_facilities(
+        current_user,
+        facilities_params[:system_id],
+        facilities_params[:parent_code],
+        facilities_params[:type_of_care_id]
```
johnpaulashenfelter: This is mapping to `clinical-service`, right?

modules/vaos/app/controllers/vaos/direct_scheduling_facilities_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    def facilities_params
+      params.require(:parent_code)
+      params.require(:system_id)
+      params.require(:type_of_care_id)
+      params.permit(
+        :parent_code,
+        :system_id,
+        :type_of_care_id
+      )
```
johnpaulashenfelter: Looking at the cassette, a succesful request could be `?clinical-service=323&facility-code=688&parent-code=688`

And looking at this all are required, so could this be rewritten as 

`params.require(:parent_code, :system_id, :type_of_care_id).permit!` ?

modules/vaos/config/routes.rb
johnpaulashenfelter: I checked in case I missed something in the commit history that would help explain it, but those are pretty terse commit messages.

modules/vaos/app/controllers/vaos/direct_scheduling_facilities_controller.rb
kreek: Yes, we have a [WIP swagger](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/products/health-care/appointments/engineering/api-contract.yml) where we mapped type_of_care_id to clinical-service for this and other endpoints.

modules/vaos/app/services/vaos/systems_service.rb
kreek: IMO it doesn’t need to be extracted here, it would be a method that returned a constant (maybe it should be a constant 😄), the reason it’s in a variable to begin with is so the perform line isn’t too long.

modules/vaos/config/routes.rb
kreek: Not having a custom controller name conflicts with the existing facilities controller, this routing has facilities belonging to systems which is the case. I do like the idea of renaming the route direct_scheduling_facilities 👍- If our devs (@saneshark @jbalboni @narin) agree with the change I’ll implement that.

modules/vaos/spec/services/systems_service_spec.rb
saneshark: The array returned in the VCR cassette is 6. We're just verifying the response returned from downstream here. There isn't really anything magic about it. Could have just as easily verified that `is_an_instance_of(Array)` but this is a bit more detailed.

modules/vaos/config/routes.rb
jbalboni: Yeah, that make sense. Especially because I think we're now going to have a more generic clinics endpoint in the future.

modules/vaos/app/services/vaos/systems_service.rb
johnpaulashenfelter: I'm a fan of the constant 👍 but implementation is all over the map both here and in other parts of the API

kreek: I don't agree that here it's "all over the map". The implementation here is that when a line would be too long and/or requires string interpolation the url is extracted into a variable to make it more readable (and pass linting). In other services we use a method where there is logic required to build the url or when the url uses implicit concatenation (multi line with a `\`).

saneshark: Yes, this is very much a pattern you will see throughout the code and a perfect example of a code review comment that does nothing more than hamper / handicap development. Please, if its not a PII issue or egregious security violation of some sort can we stop with the whole "There's a new Sherrif in town." This is getting really tired @johnpaulashenfelter

PR #3568, opened 2019-11-19
WIP - Post va appointment request

saneshark:
## Description of change
Introduces Appointment Request cancelation endpoint to the service.
Introduces cancellation endpoint and routes to controller
Adds specs, factories, and models to assist in coercion.
Skips swagger for now.

UPDATE 11/20:
In testing it was recently discovered that the JWT we are currently using does not adequately capture a few attributes that are required for this service integration to work. I will continue to do discovery around this within this branch, but I will be separating out the JWT specific changes into another separate PR. It appears as though I cannot retroactively change this to a draft, but please do not bother with reviewing this. Furthermore, the commit history reflects an already merged PR that I needed to piggy back off of for this feature. This is only open as a PR vs a branch to help inform my team of the models being added. When this work is complete I will likely close this PR and cherry pick the necessary commits into 2 separate PRs: 1 for JWT changes, and 1 for AppointmentRequest cancellations.

## Testing done
Local

## Testing planned
Staging

#### Unique to this PR
See above.

#### Applies to all PRs

- [ ] Appropriate logging


modules/vaos/spec/factories/va_appointment_requests.rb
```diff
@@ -0,0 +1,39 @@
...
+        facility_state: 'VA',
+        facility_code: '652',
+        facility_name: 'Richmond VAMC',
+        facility_parent_site_code: '688',
+        facility_type: 'VAMC'
+      }
+    end
+
+    email { 'test@va.gov' }
+    phone_number  { '+1 (555) 555-5555 x55.55' }
```
johnpaulashenfelter: dots are allowed in the phone extension?

modules/vaos/spec/models/va_appointment_request_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    end
+
+    it 'coerces params to correct types' do
+      expect(subject.params).to eq(
+        appointment_time: '11/13/2019 20:19:12',
+        clinic_id: '455',
+        cancel_reason: '5',
+        cancel_code: 'PC',
+        remarks: nil, # doesn't coerce nil to ""
+        clinic_name: ''
```
johnpaulashenfelter: Trying to find the place clinic_id and clinic_name come from -- didn't see in factory or any of the new code. Is it matching those values from a VAOS:Facility factory?

modules/vaos/spec/services/appointment_requests_service_spec.rb
```diff
@@ -5,13 +5,39 @@
...
+    context 'when appointment_request can be cancelled' do
+      let(:request_body)  { build(:va_appointment_request) }
+
+      it 'cancels the appointment_request' do
+        VCR.use_cassette('vaos/appointment_requests/put_cancel_request', match_requests_on: %i[method uri]) do
+          response = subject.put_cancel_appointment_request(request_body)
+          expect(response).to be_an_instance_of(String).and be_empty
+        end
+      end
+    end
```
johnpaulashenfelter: The request here is the same in both paths -- a new `build(:va_appointment_request)` but one results in an error and one doesn't. Seems like it would be clearer to show what changed to lead to a failure recording in particular. Or just pass some kind of mock request_body if the values aren't important.

The first spec says its a bad request but that's confusing when the second request, which seems exactly the same, is a valid request.

modules/vaos/spec/services/appointment_requests_service_spec.rb
```diff
@@ -5,13 +5,39 @@
...
+  end
+
   describe '#get_requests' do
+    let(:start_date) { Date.parse('2019-08-20') }
+    let(:end_date) { Date.parse('2020-08-22') }
```
johnpaulashenfelter: These two values don't have a clear reason to have been added from the immediate context/code -- this method seems like it doesn't have anything to do with the rest of the changes. Was this from another commit or did something else change that required this change?

modules/vaos/spec/factories/va_appointment_requests.rb
saneshark: Yes. It's not a typo.

modules/vaos/spec/models/va_appointment_request_spec.rb
saneshark: Which is largely why this PR is still a WIP.

modules/vaos/spec/services/appointment_requests_service_spec.rb
saneshark: I write these specs, and then I get on a staging box and I record the VCR cassettes. Then I commit the cassettes directly from staging machine. Perhaps you didn't notice that the requests identified are also not included as part of this PR. It's because they haven't been recorded yet. Which is why this is WIP.

modules/vaos/spec/services/appointment_requests_service_spec.rb
saneshark: I moved stuff in the specs around. You'd have to look at the full diff more carefully to understand the rationale for why. It's largely because your previous code review comments suggested that the descriptions for examples were not clear to you.

modules/vaos/app/services/vaos/appointment_requests_service.rb
```diff
@@ -31,6 +31,19 @@ def get_requests(start_date = nil, end_date = nil)
       end
     end
 
+    def put_cancel_request(request_object_body)
+      params = VAOS::Patient.new(request_object_body).params
+      params.merge!(patient_identifier: { unique_id: user.icn, assigning_authority: 'ICN' })
+
+      binding.pry
```
saneshark: these binding.pry are here as breakpoints to help facilitate recording cassettes. They will be removed.

modules/vaos/app/services/vaos/jwt.rb
```diff
@@ -18,18 +18,33 @@ def payload
         sub: @user.icn,
         idType: 'ICN',
         iss: 'gov.va.vaos',
-        firstName: @user.first_name,
-        lastName: @user.last_name,
+        firstName: @user.mvi&.profile&.given_names&.first, # from MVI not SAML assertion
```
kreek: Could this be a method named for the comment? e.g. `mvi_first_name` (same for last name)

modules/vaos/app/models/vaos/va_appointment_request.rb
```diff
@@ -0,0 +1,54 @@
...
+
+require 'common/models/form'
+
+module VAOS
+  class VAAppointmentRequest < Common::Form
+    include ActiveModel::Validations
+
+    attribute :email, String
+    attribute :phone_number, String
+    attribute :option_date1, String
```
kreek: We're constrained by the service so this is rhetorical, but if we weren't, could this be an options array?

modules/vaos/spec/request/appointment_requests_request_spec.rb
```diff
@@ -13,78 +13,146 @@
...
-        expect(response).to have_http_status(:bad_request)
-        expect(response.body).to be_a(String)
-        expect(JSON.parse(response.body)['errors'].first['detail'])
-          .to eq('"invalid" is not a valid value for "end_date"')
+      context 'without an end_date' do
+        it 'has a parameter missing exception' do
+          get '/v0/vaos/appointment_requests', params: params.except(:end_date)
+          expect(response).to have_http_status(:bad_request)
+          expect(response.body).to be_a(String)
+          expect(JSON.parse(response.body)['errors'].first['detail'])
```
kreek: Could the invalid cases be wrapped in another description/context and have `JSON.parse(response.body)['errors'].first['detail']` be a let?

modules/vaos/app/models/vaos/va_appointment_request.rb
saneshark: Yes, i'd love to that when the underlying service can support it as such. This is the craziest form object I've ever worked with and I've seen some really bizarre things in the past. It actually requires a flattened version of much of the same data, but balks if that flattened patient data isn't also provided in an object. My plan is to keep the form validations and handle that kloodgy mess in the service so that the FE doesn't need to be exposed to it.

saneshark:
Closing because rebase and cherry picking is proving too difficult. Part of this PR (messaging and other pieces were already merged last week, and there's an open PR for the JWT changes). I am going to have to create a branch off of the JWT changes since those have not been merged to master yet. I'll just copy and paste from the diff here. 

The problem is that one thing depends on another, but we also want to keep PR's small for more efficient code review. Not sure if there is a better way, but I'm often forced to branch off of a branch that has not yet been merged pending review. @here

PR #3571, opened 2019-11-21
VAOS facility limits endpoint

kreek:
## Description of change
Adds the `/v0/vaos/facilities/:facility_id/limits` endpoint and a `#get_facility_limits` to the systems service. 

## Testing done
unit tests

## Testing planned
staging integration tests

#### Unique to this PR
- [ ] adds limits controller
- [ ] adds get_facility_limits method to systems service

#### Applies to all PRs

- [ ] Appropriate logging


modules/vaos/app/services/vaos/systems_service.rb
```diff
@@ -39,6 +39,18 @@ def get_facility_clinics(user, facility_id, type_of_care_id, system_id)
       end
     end
 
+    def get_facility_limits(user, facility_id, type_of_care_id)
+      with_monitoring do
+        url = "/var/VeteranAppointmentRequestService/v4/rest/direct-scheduling/patient/ICN/#{user.icn}/request-limit"
+        url_params = {
+          'institution-code' => facility_id,
+          'clinical-service' => type_of_care_id
+        }
```
saneshark: I think this is fine since there are only two and you're doing the `param` validation in the controller. If there were any more than 2-3, or if we knew we needed to validate on more than just `presence` I might suggest a form object.

modules/vaos/app/controllers/vaos/limits_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+
+    def limit_params
+      params.require(:facility_id)
+      params.require(:type_of_care_id)
+      params.permit(
+        :facility_id,
+        :type_of_care_id
+      )
+    end
```
saneshark: Mostly a nitpick, but wouldn't read a little bit better if you create methods for each of these, passing the method as arguments to your service?

```ruby
def facility_id
  params.require(:facility_id).permit(:facility_id)
end

def type_of_care_id
  params.require(:type_of_care_id).permit(:type_of_care_id)
end
```

modules/vaos/app/controllers/vaos/limits_controller.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+require_dependency 'vaos/application_controller'
```
saneshark: Have we found that this is actually required? I would have thought it would be loaded automatically as part of the engine.

modules/vaos/app/controllers/vaos/limits_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+module VAOS
+  class LimitsController < ApplicationController
+    def index
+      response = systems_service.get_facility_clinics(
+        current_user,
+        clinics_params[:facility_id],
+        clinics_params[:type_of_care_id]
+      )
+
+      render json: VAOS::ClinicSerializer.new(response)
```
LindseySaari: Is this the correct serializer?

modules/vaos/app/controllers/vaos/limits_controller.rb
```diff
@@ -5,13 +5,13 @@
...
+      response = systems_service.get_facility_limits(
         current_user,
-        clinics_params[:facility_id],
-        clinics_params[:type_of_care_id]
+        limit_params[:facility_id],
+        limit_params[:type_of_care_id]
       )
 
-      render json: VAOS::ClinicSerializer.new(response)
+      render json: VAOS::LimitSerializer.new(response)
```
LindseySaari: Looks like you fixed it here! 😄 👍

modules/vaos/app/controllers/vaos/limits_controller.rb
jholton: `require` returns the value(s) of the key(s) provided, so in this example, you'd be calling `permit` on a string.
And then `permit` still returns a `params` object.
So in this case you can just call `require` since we're not worried about mass assignment:
```ruby
def facility_id
  params.require(:facility_id)
end
```

PR #3573, opened 2019-11-21
VA Forms Search

ATeal:
## Description of change
Add a new query param for searching functionality

resolves https://github.com/department-of-veterans-affairs/vets-contrib/issues/3648

## Testing done
- rspec

nique to your PR_
#### Applies to all PRs

- [x] Appropriate logging

### DIFF
--- a/modules/va_forms/app/controllers/va_forms/v0/forms_controller.rb
+++ b/modules/va_forms/app/controllers/va_forms/v0/forms_controller.rb
@@ -21,7 +21,7 @@ module VaForms
 
       def get_forms
         if params[:query].present?
-          Form.where('name like ? OR title like ?', "%#{params[:query]}%", "%#{params[:query]}%")
+          Form.where('form_name like ? OR title like ?', "%#{params[:query]}%", "%#{params[:query]}%")
         else
           Form.all
         end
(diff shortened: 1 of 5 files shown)


modules/va_forms/app/controllers/va_forms/v0/forms_controller.rb
```diff
@@ -17,6 +16,16 @@ def show
         render json: forms,
                serializer: VaForms::FormDetailSerializer
       end
+
+      private
+
+      def get_forms
+        if params[:query].present?
+          Form.where('name like ? OR title like ?', "%#{params[:query]}%", "%#{params[:query]}%")
```
bastosmichael: Can you use % instead of like for the search operator, might need to include pg_trim but will give you way more fuzzy searches in a way then just a like term https://www.rdegges.com/2013/easy-fuzzy-text-searching-with-postgresql/

charleystran: Also this should be form_name over name

ATeal: just using ilike, the other seems to be a little heavy handed for this initial release. will keep it in our back pockets for future requests

PR #3575, opened 2019-11-21
update facilities metadata client and configuration namespacing

LindseySaari:
## Description of change
This change addresses namespacing cleanup within the facilities metadata directory (lib/facilities/metadata). In order to better understand separation of facility services, namespacing improvements were introduced.

It was decided to break the original PR up into smaller PRs. Link to Original PR: #3565

## Testing done
Test suite passing

#### Unique to this PR
- [x] Passing test suite

#### Applies to all PRs

- [ ] Appropriate logging


lib/facilities/metadata/configuration.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+require 'common/client/configuration/rest'
+require 'common/client/middleware/response/raise_error'
+require 'common/client/middleware/response/facility_parser'
```
jholton: I realize you just copy/pasted these lines, but can these be removed?
I'm wondering if they are relics from a time when autoloading wasn't configured properly for `lib`.
`facility_parser` isn't even used in this file.
I commented them out (and the `require` in the `client.rb`), and all specs pass.  But I'm worried I'm missing something.

lib/facilities/metadata/client.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require 'common/client/base'
+
+module Facilities
+  module Metadata
+    class Client < Common::Client::Base
+      configuration Facilities::Metadata::Configuration
+
+      def get_metadata(facility_type)
```
kfrz: This should be using `with_monitoring do ...`

lib/facilities/metadata/configuration.rb
```diff
@@ -0,0 +1,35 @@
...
+      def service_name
+        'FL'
+      end
+
+      def connection
+        Faraday.new(base_path, headers: base_request_headers, request: request_options) do |conn|
+          conn.use :breakers
+          conn.request :json
+
+          # Uncomment this if you want curl command equivalent or response output to log
```
kfrz: Remove this comment block

lib/facilities/metadata/configuration.rb
LindseySaari: Ok, awesome. I'll take a look too!

LindseySaari:
@kfrz It looks like i'm blocked on merging this

PR #3578, opened 2019-11-21
Set GI timeout to 1s

annaswims:
## Description of change
As a folloup to https://github.com/department-of-veterans-affairs/vets.gov-team/blob/master/Postmortems/2019-06-10-GI-Latency.md  and Today's outage, reduce GI timeout to 1s.

[Typical latency is below 200 ms.](http://grafana.vfs.va.gov/d/000000050/backend-service-report?orgId=1&fullscreen&panelId=121&from=1574333041713&to=1574371541713)  The default timeout is 15 s system-wide but due to the high number of autocomplete requests we've see cascading effects to vets-api when requests are slow. 
## Testing done
specs

#### Applies to all PRs

- [ ] Appropriate logging


annaswims:
yeah, it can be monitored here http://grafana.vfs.va.gov/d/000000050/backend-service-report?orgId=1&from=now-3h&to=now . but pagerduty will alert us in #oncall if it goes bad.  

This is configurable per env because we use the [config](https://github.com/railsconfig/config) gem. so. here's the prod config https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/deployment/config/vets-api/prod-settings.local.yml.j2

you can also set an environment variable to override, like we do [here](https://github.com/department-of-veterans-affairs/vets-api/blob/517cd59caa6bcdf1e4dc0839cfc5a38a4843511e/docker-compose.yml#L32)

PR #3579, opened 2019-11-21
Updates VAOS JWT to support endpoints that depend on it

saneshark:
## Description of change
It was recently discovered as part of https://github.com/department-of-veterans-affairs/vets-api/pull/3568 that the JWT we were previously using does not get a returned token from user_service that includes certain necessary parameters. Those `patient` parameters are only returned if JWT includes the parameters as part of the root payload. When it is verified and signed by the user_service, the attributes are added to a new patient object as part of the returned response.

## Testing done
Local and Integration testing on a dedicated staging sandbox.

## Testing planned
More smoke testing on actual staging.

#### Unique to this PR
Improves the JWT and adds additional test coverage.

#### Applies to all PRs

- [x] Appropriate logging

### DIFF
--- a/modules/vaos/spec/services/jwt_spec.rb
+++ b/modules/vaos/spec/services/jwt_spec.rb
@@ -5,14 +5,58 @@ require 'rails_helper'
 describe VAOS::JWT do
   subject { VAOS::JWT.new(user) }
 
-  let(:user) { build(:user, :mhv) }
+  let(:user) { build(:user, :vaos) }
+  let(:rsa_private) { OpenSSL::PKey::RSA.generate(4096) }
+  let(:jwt_regex) { %r{^[A-Za-z0-9\-_=]+\.[A-Za-z0-9\-_=]+\.?[A-Za-z0-9\-_.+/=]*$} }
 
   describe '#token' do
-    it 'encodes a payload' do
-      rsa_private = OpenSSL::PKey::RSA.generate 4096
-      allow(File).to receive(:read).and_return(rsa_private)
-      decoded = JWT.decode(subject.token, rsa_private.public_key, true, algorithm: 'RS512').first
-      expect(decoded['firstName']).to eq(user.first_name)
+    before { allow(File).to receive(:read).and_return(rsa_private) }
+
+    it 'returns a JWT string' do
+      expect(subject.token).to be_a(String).and match(jwt_regex)
+    end
+
+    context 'decoded payload' do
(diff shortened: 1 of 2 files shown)


modules/vaos/spec/services/jwt_spec.rb
```diff
@@ -5,14 +5,58 @@
 describe VAOS::JWT do
   subject { VAOS::JWT.new(user) }
 
-  let(:user) { build(:user, :mhv) }
+  let(:user) { build(:user, :vaos) }
+  let(:rsa_private) { OpenSSL::PKey::RSA.generate(4096) }
+  let(:jwt_regex) { %r{^[A-Za-z0-9\-_=]+\.[A-Za-z0-9\-_=]+\.?[A-Za-z0-9\-_.+/=]*$} }
```
omgitsbillryan: 😱 

Could you provide a human-readable comment for what this is representing?

modules/vaos/spec/services/jwt_spec.rb
```diff
@@ -5,14 +5,58 @@
...
   describe '#token' do
-    it 'encodes a payload' do
-      rsa_private = OpenSSL::PKey::RSA.generate 4096
-      allow(File).to receive(:read).and_return(rsa_private)
-      decoded = JWT.decode(subject.token, rsa_private.public_key, true, algorithm: 'RS512').first
-      expect(decoded['firstName']).to eq(user.first_name)
+    before { allow(File).to receive(:read).and_return(rsa_private) }
+
+    it 'returns a JWT string' do
+      expect(subject.token).to be_a(String).and match(jwt_regex)
```
omgitsbillryan: What exactly are you validating here? For example - what would be an example of a non-JWT string?

modules/vaos/spec/services/jwt_spec.rb
saneshark: I think "jwt_regex" is pretty self explanatory. JWTs have 3 parts: a header, a payload, and a signature. Each part is base64url encoded. You want me to put that in the header? I think it's common knowledge for anyone working with JWT, but happy to do it if you think it is helpful.

modules/vaos/spec/services/jwt_spec.rb
saneshark: A string that doesn't match the jwt regex: "this string would not validate." It has spaces, it does not have 3 base64 url encoded parts.

saneshark:
Closing this PR as the other one has now been approved and it includes these changes.

PR #3580, opened 2019-11-22
Fixing Duplicate UUID issue

bastosmichael:
## Description of change
Fix for ticket https://github.com/department-of-veterans-affairs/vets-contrib/issues/3562

- Catch and ignore rare duplicate state ID issues returning from GDIT.

#### Applies to all PRs

- [ ] Appropriate logging


modules/vba_documents/app/workers/vba_documents/upload_processor.rb
```diff
@@ -117,7 +118,7 @@ def to_faraday_upload(file_io, filename)
     end
 
     def process_response(response)
-      if response.success?
+      if response.success? || response.body.match?(NON_FAILING_ERROR_REGEX)
```
ATeal: Were you able to reproduce this situation? Maybe by doing to `PUT`s back to back? 

Probably a good candidate for a test either way

omgitsbillryan: Would definitely want to see a spec to exercise this code.

bastosmichael: Added a spec

PR #3582, opened 2019-11-22
Post and put va appointment request

saneshark:
## Description of change
This PR closes: 
https://github.com/department-of-veterans-affairs/va.gov-team/issues/1788 (PUTS /requests/id)
and
https://github.com/department-of-veterans-affairs/va.gov-team/issues/2969 (POSTS /requests

NOTE: This PR depends on https://github.com/department-of-veterans-affairs/vets-api/pull/3579 getting merged first at which point I will rebase with that branch and then rebase with master. I might have to not squash commits on 3579 though to that cleanly. Alternately, I can merge this in and close the other PR.

Adds POST and PUT capability for appointment requests.
Uses a form object to normalize data coming in, and adds on attributes that the FE does not have available but the service depends on. There is plenty of opportunity to refactor some of the other controllers based on the approach done here. This seems to be the most flexible way to tackle the challenge of normalizing data for these set of services (none of which are consistent). A quick look at the vcr cassette payloads will demonstrate that dates are wildly inconsistent. And null valued keys are often dropped, which makes normalizing a consistent set of keys for the FE to consume that much more challenging.

Form validations are currently absent pending further discovery and reverse engineering. It is not clear which attributes are required and which are optional. This discovery will be more readily facilitated by FE integration testing.

## Testing done
Staging Sandbox to record VCR cassettes and then local testing.

## Testing planned
Staging / Vets-Website integration.

FE to validate and provide feedback.

#### Unique to this PR
See above

#### Applies to all PRs

- [x] Appropriate logging


modules/vaos/spec/request/appointment_requests_request_spec.rb
```diff
@@ -10,80 +10,127 @@
...
-        .to eq('You do not have access to online scheduling')
-    end
-  end
-
-  context 'loa3 user' do
-    let(:current_user) { build(:user, :mhv) }
-    let(:start_date) { Date.parse('2019-08-20') }
-    let(:end_date) { Date.parse('2020-08-22') }
-    let(:params) { { start_date: start_date, end_date: end_date } }
+  describe 'GET /v0/vaos/appointment_requests' do
```
saneshark: NOTE: most of these specs were just tabbed over, not actually modified. Wrapped them in a new describe block. GET requests are unchanged as part of this PR.

modules/vaos/app/controllers/vaos/appointment_requests_controller.rb
```diff
@@ -4,21 +4,42 @@
...
+    MASSASSIGN_PARAMS_POST = MASSASSIGN_PARAMS_PUT - %i[created_date last_access_date last_updated_date]
 
     def index
-      render json: AppointmentRequestsSerializer.new(requests[:data], meta: requests[:meta])
+      response = appointment_requests_service.get_requests(start_date, end_date)
+      render json: AppointmentRequestsSerializer.new(response[:data], meta: response[:meta])
+    end
+
+    def create
+      response = appointment_requests_service.post_request params.permit(*MASSASSIGN_PARAMS_POST)
```
kreek: could we have parens here? 
```ruby
appointment_requests_service.post_request(params.permit(*MASSASSIGN_PARAMS_POST))
```

modules/vaos/app/controllers/vaos/appointment_requests_controller.rb
```diff
@@ -4,21 +4,42 @@
 
 module VAOS
   class AppointmentRequestsController < ApplicationController
-    before_action :validate_params
+    before_action :validate_params, only: :index
+
+    MASSASSIGN_PARAMS_PUT = %i[
```
kfrz: Nitpick, but I'd prefer to see something like `MASSASSIGN_PARAMS_PUT` named something more like `MASS_ASSIGN_PUT_PARAMS`, for specific clarity

modules/vaos/app/controllers/vaos/appointment_requests_controller.rb
```diff
@@ -4,21 +4,42 @@
...
+      email phone_number option_date1 option_time1 option_date2 option_time2 option_date3 option_time3
+      status appointment_type visit_type text_messaging_allowed phone_number purpose_of_visit
+      other_purpose_of_visit purpose_of_visit provider_id second_request second_request_submitted
+      provider_name requested_phone_call type_of_care_id has_veteran_new_message has_provider_new_message
+      provider_seen_appointment_request requested_phone_call type_of_care_id created_date last_access_date
+      last_updated_date facility patient best_timeto_call appointment_request_detail_code
+    ].freeze
+
+    MASSASSIGN_PARAMS_POST = MASSASSIGN_PARAMS_PUT - %i[created_date last_access_date last_updated_date]
```
kfrz: Ditto on the naming comment

modules/vaos/app/models/vaos/appointment_request_form.rb
```diff
@@ -0,0 +1,155 @@
...
+    attribute :best_timeto_call, Array[String] # VAR camel cases this incorrectly so we will snakecase it incorrectly
+    attribute :appointment_request_detail_code, Array[String] # This setter is overriden
+
+    def initialize(user, json_hash)
+      @user = user
+      @id = json_hash[:id]
+      @appointment_request_id = @id # ensure these are the same (and the FE doesn't have to pass in redundant data)
+      @unique_id = @id # ensure these are the same (and the FE doesn't have to pass in redundant data)
+      @date = Time.current.strftime('%Y-%m-%dT%H:%M:%S.%L%z') if @id.blank? # only set for create in proper format
+      @patient_id = edipi # TODO: discover if the patient_id ever comes from elsewhere or if it is always EDIPI
```
kfrz: This `TODO` should either be addressed & removed or extracted to an issue & removed from the code

modules/vaos/app/models/vaos/appointment_request_form.rb
```diff
@@ -0,0 +1,155 @@
...
+    def patient_id=(_value); end
+
+    # These values are provided directly from FE without coercion; we may need to validate some of these
+    def facility=(values_hash)
+      @facility = values_hash.empty? ? {} : values_hash.merge(object_type: 'Facility').compact
+    end
+
+    # These values ought to be derived from MVI vs user input, except for inpatient and text_messaging_allowed
+    def patient=(values_hash)
+      @patient = {
```
kfrz: Feels like this should be a model/struct? A `Patient` model? How can we test behavior on this?

modules/vaos/app/services/vaos/appointment_requests_service.rb
```diff
@@ -31,6 +31,28 @@ def get_requests(start_date = nil, end_date = nil)
       end
     end
 
+    def post_request(request_object_body)
+      with_monitoring do
+        params = VAOS::AppointmentRequestForm.new(user, request_object_body).params
+        response = perform(:post, url, params, headers(user))
+
+        {
+          data: OpenStruct.new(response.body)
```
kfrz: Is this a conscious choice to not use a `VAOS::PostResponseObject` or something similar to other clients?

modules/vaos/spec/request/appointment_requests_request_spec.rb
```diff
@@ -10,80 +10,127 @@
...
+    context 'loa1 user with flipper enabled' do
```
kfrz: I prefer to see `context` blocks using `when/with` as a pattern, because it makes the specs read more logically in `rspec` output

I.e.:

```suggestion
    context 'when user is loa1 with flipper enabled' do
```

Also not necessary for this PR but as an aside, the above example is interlocking two contexts which could probably be separate. :man_shrugging:

modules/vaos/spec/request/appointment_requests_request_spec.rb
```diff
@@ -10,80 +10,127 @@
...
     end
 
-    context 'without a start_date' do
-      it 'has a parameter missing exception' do
-        get '/v0/vaos/appointment_requests', params: params.except(:start_date)
-        expect(response).to have_http_status(:bad_request)
-        expect(response.body).to be_a(String)
-        expect(JSON.parse(response.body)['errors'].first['detail'])
-          .to eq('The required parameter "start_date", is missing')
+    context 'loa3 user' do
```
kfrz:
```suggestion
    context 'when user is loa3' do
```

modules/vaos/spec/services/jwt_spec.rb
```diff
@@ -5,14 +5,58 @@
 describe VAOS::JWT do
   subject { VAOS::JWT.new(user) }
 
-  let(:user) { build(:user, :mhv) }
+  let(:user) { build(:user, :vaos) }
+  let(:rsa_private) { OpenSSL::PKey::RSA.generate(4096) }
```
kfrz: Any way to avoid calling out to generate a new key? Probably not a big deal, just wondering if there's a significant performance cost here.

modules/vaos/app/services/vaos/jwt.rb
```diff
@@ -2,6 +2,13 @@
 
 module VAOS
   class JWT
```
kfrz: Please add top-level class documentation for [YARDocs](https://www.rubydoc.info/github/department-of-veterans-affairs/vets-api

modules/vaos/app/services/vaos/jwt.rb
```diff
@@ -2,6 +2,13 @@
...
+    VERSION = 2.1
+    ISS = 'gov.va.vaos'
+    ID_TYPE = 'ICN'
+    AUTHORITY = 'gov.va.iam.ssoe.v1'
```
kfrz: Would like to see some docs on what these constants represent? Maybe that's covered elsewhere or could go into a top-level class docs

modules/vaos/app/models/vaos/appointment_request_form.rb
```diff
@@ -0,0 +1,155 @@
+# frozen_string_literal: true
+
+require 'active_model'
+require 'common/models/form'
+
+module VAOS
+  class AppointmentRequestForm < Common::Form
+    attribute :email, String
+    attribute :phone_number, String
+    attribute :option_date1, String # TODO: ideally these would be better abstracted as an array for a FE consumer
```
kfrz: Please either address this TODO before merge or extract to an issue and remove comment lest we have further tech debt :D

modules/vaos/app/models/vaos/appointment_request_form.rb
```diff
@@ -0,0 +1,155 @@
+# frozen_string_literal: true
+
+require 'active_model'
+require 'common/models/form'
+
+module VAOS
+  class AppointmentRequestForm < Common::Form
```
kfrz: Please add top level class [rubydocs](https://www.rubydoc.info/github/department-of-veterans-affairs/vets-api and public method docs

kfrz:
Also for housekeeping could you connect this PR w/ the appropriate issue (if it exists)

modules/vaos/app/controllers/vaos/appointment_requests_controller.rb
omgitsbillryan: At first glance I read "assassin params put"

saneshark: LOL I'll make this change this before merging.

modules/vaos/app/models/vaos/appointment_request_form.rb
saneshark: That's actually very sound advice. I'll go ahead and remove it and add a ticket to follow up with VAOS SMEs on this.

modules/vaos/app/models/vaos/appointment_request_form.rb
saneshark: It was originally. There was also a Facility model. And I'll probably go back to that once it comes time to add validations as it will make it easier. But for keeping this PR light / small I went with this approach.

modules/vaos/app/services/vaos/appointment_requests_service.rb
saneshark: Yes, it's kind of a bizarre that the Netflix serialize we're using only accepts an object and not JSON. We don't really have a need for a model. This is just a hack to get the much faster Netflix serializers accept our json.

modules/vaos/spec/services/jwt_spec.rb
saneshark: Oh I don't think any of these are actually calling out. They should all be using a VCR cassette. And if you look in UserService once the JWT is returned it is cached for 15 minutes.

modules/vaos/app/services/vaos/jwt.rb
saneshark: I can link to this but I think the documentation lives in coderepo.va.mobile. With their permission I might be able to expose that documentation.

modules/vaos/app/models/vaos/appointment_request_form.rb
saneshark: @jbalboni what are your thoughts on these attributes? We could have you provide it to us in a  better way and then break it out like VAMF needs as part of re-serialization. For now I kept it like this because any errors VAMF returns would otherwise not be particularly sensible. That being said, if we're doing our own validations before making the request it shouldn't be a problem.

modules/vaos/app/models/vaos/appointment_request_form.rb
saneshark: Will do, in follow-up PR. Will open a ticket before merging to document that debt.

modules/vaos/app/models/vaos/appointment_request_form.rb
jbalboni: I would leave it as is. I think we should eventually completely restructure this data, rather than doing one off changes here and there.

modules/vaos/app/services/vaos/jwt.rb
saneshark: https://github.com/department-of-veterans-affairs/va.gov-team/issues/3689

modules/vaos/app/models/vaos/appointment_request_form.rb
saneshark: https://github.com/department-of-veterans-affairs/va.gov-team/issues/3685

modules/vaos/app/models/vaos/appointment_request_form.rb
saneshark: department-of-veterans-affairs/va.gov-team#3689

PR #3583, opened 2019-11-22
Update VCR gem

johnpaulashenfelter:
## Description of change
We are pretty far behind (5 v 3.0.3) but the upgrade seemed pretty seemless. All the specs still work so cassettes are running properly.

#### Applies to all PRs

- [x] ~Appropriate logging~
- [x] ~Swagger docs have been updated, if applicable~
- [x] ~Provide link to originating GitHub issue, or connected to it via ZenHub~
- [x] ~Provide which alerts would indicate a problem with this functionality (if applicable)~


kfrz:
Two major versions in one go? No version pinning in `Gemfile`? No additional context or tests added? Seems a bit dubious to me - how can we be certain there aren't any significant breakages?

What features do we gain? Just curious :D

omgitsbillryan:
[Link to changelog](https://github.com/vcr/vcr/blob/master/CHANGELOG.md)

Copy pasted here for the lazy:
```
5.0.0 (May 27, 2019)
Full Changelog

[breaking] disabling webmock excon adapter due to issues
[breaking] change the compressed extension from gz to zz
[breaking] don't eject cassettes on UnhandledHTTPRequestError
[breaking] when logging the before_playback placeholder replacement, VCR would state "replacing with " which does not correctly represent what is actually happening, it's now semantically correct
[breaking] removing more ruby 1.8 weirdness, re: ordered hashes quirk
[patch] VCR.use_cassettes raises no implicit conversion of nil into Hash https://github.com/vcr/vcr/pull/730#issuecomment-454049764
faster JSON comparison
recompress_response option+hook
unignore_hosts option: remove hosts added to @ignored_hosts Set with ignore_hosts
support ruby 2.6 in test matrix
support ruby 2.5 in test matrix
support Typhoeus 1.1 in test matrix
updating repository meta files

4.0.0
Full Changelog

[Feature] Output non-matched headers on error
[Fix] Allow non-latin charaters in cassette filenames
[Breaking] Remove deprecated VCR::RSpec::Macros
[Breaking] Remove support for ruby 1.9.3
[Breaking] Remove support for Fakeweb
```

omgitsbillryan:
Might I recommend that you record (or re-record) a new cassette just as a test for added assurance that all is well. Once that's done it LGTM.

kfrz:
Thanks for the changelog link - looks pretty minimal in terms of risk! +1 to the "record a cassette" smoke test, then I'm happy to :heavy_check_mark:

kfrz:
Updated this PR with a re-recorded `search` spec, works like a charm. 

Yes, the body response is now binary-encoded, but that's actually how it should be. The readable JSON string that did exist was manually entered, by me & Harry when this service was created. It was mostly for reference/FE use but we shouldn't need to be looking at cassette contents directly IMO.

@johnpaulashenfelter will leave up to you if you want to merge or not w/ the binary-encoded string.

PR #3585, opened 2019-11-22
Update awrence gem

johnpaulashenfelter:
Update the awrence gem, which is only used in two specs to handle camelcase conversions.

#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging


Gemfile.lock
```diff
@@ -692,7 +692,7 @@ GEM
     uri_template (0.7.0)
     utf8-cleaner (0.2.5)
       activesupport
-    vcr (3.0.3)
+    vcr (5.0.0)
```
kfrz: This seems errant, though if we merge the VCR PR #3583 first then we'll be fine.

PR #3588, opened 2019-11-25
Adding data type checks to some specs

thilton-oddball:
## Description of change
Adding in data type checks to specs on some Virtus models


thilton-oddball:
@LindseySaari Yes, the plan is still to move away from Virtus. Just adding some type checks to beef up specs prior to converting over to dry-types.

PR #3589, opened 2019-11-25
Fixes 3667 VAOS facilities array of ids

kreek:
## Description of change
Fixes [3667](https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/3667). Allows `/v0/vaos/facilities` to take multiple `facility_codes` url params e.g. `/v0/vaos/facilities?facility_codes[]=983&facility_codes[]=984`

Note that there's a bug in VCR where it does not record the url correctly when you flatten the url params (don't use `[]=`, which the VAOS service doesn't). However I did test this out intercepting the request and it build the url as VAOS expects outside of tests 
```
/var/VeteranAppointmentRequestService/v4/rest/direct-scheduling/parent-sites?facility-code=983&facility-code=984
```

## Testing done
unit tests

## Testing planned
staging integration tests

#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging


modules/vaos/app/services/vaos/systems_service.rb
```diff
@@ -18,10 +18,11 @@ def get_systems(user)
       end
     end
 
-    def get_facilities(user, facility_code)
+    def get_facilities(user, facility_codes)
       with_monitoring do
         url = '/var/VeteranAppointmentRequestService/v4/rest/direct-scheduling/parent-sites'
-        response = perform(:get, url, { 'facility-code' => facility_code }, headers(user))
+        options = { params_encoder: Faraday::FlatParamsEncoder }
```
saneshark: I take it this is needed because of how the array of facility codes deviates from the Rails []= convention? I'm glad no monkey patching was necessary!

spec/support/vcr_cassettes/vaos/systems/get_facilities_multiple.yml
```diff
@@ -0,0 +1,1334 @@
+---
+http_interactions:
+  - request:
+      method: get
+      uri: https://veteran.apps.va.gov/var/VeteranAppointmentRequestService/v4/rest/direct-scheduling/parent-sites?facility-code=984
```
saneshark: In your request here it doesn't appear as though you are sending an array, what am I missing here?

kreek: > Note that there's a bug in VCR where it does not record the url correctly when you flatten the url params (don't use []=, which the VAOS service doesn't). However I did test this out intercepting the request and it build the url as VAOS expects outside of tests

kreek: No matter how many url param array items you pass it only captures the last one in VCR, but I tested it by making a Sinatra service locally and outside of specs it does pass full set of url params correctly.

PR #3590, opened 2019-11-25
Upgrade eMIS Military Information Service to v2

nfasulo:
## Description of change
This PR actually finishes #3118. To support USAJobs as a consumer of the veteran data APIs, we want to provide a normalized rank across services, which is included in the MilitaryInformationServiceV2. The new MilitaryInformationServiceV2 is added here that mirrors the structure of other eMIS services and configurations, but only `get_military_service_episodes` has been added as an endpoint at this time.

Other notable changes:

- The settings file is changed to differentiate xml namespaces between v1 and v2 MilitaryInformationService clients. The v1 list is not as extensive as the v2 list because many of the defaults in edipi_or_icn_message.rb are already correct for v1. The extras included in the v2 list get merged and override those defaults of the same namespace.
- The payment service urls in settings.yml are similarly split into v1 and v2.

## Testing done
- Added specs for the service and model along with fixtures recorded from eMIS based on a test user

#### Unique to this PR
- [x] Create client that can retrieve information from the getServiceEpisode endoint in eMIS v2

#### Applies to all PRs

- [x] Appropriate logging


spec/support/vcr_cassettes/emis/get_military_service_episodes/valid_no_end_date_v2.yml
```diff
@@ -0,0 +1,72 @@
+---
+http_interactions:
+- request:
+    method: post
+    uri: https://localhost:4438/VIERSService/eMIS/v2/MilitaryInformationService
```
rtravitz: The base url in this cassette is going to have to change to match the base in `settings.yml`: https://vaausvrsapp81.aac.va.gov

spec/support/vcr_cassettes/emis/get_military_service_episodes/valid_v2.yml
```diff
@@ -0,0 +1,69 @@
+---
+http_interactions:
+- request:
+    method: post
+    uri: https://localhost:4438/VIERSService/eMIS/v2/MilitaryInformationService
```
rtravitz: Same note about the base url changing

lib/emis/military_information_configuration_v2.rb
```diff
@@ -0,0 +1,25 @@
...
+    def base_path
+      URI.join(Settings.emis.host, Settings.emis.military_information_url.v2).to_s
+    end
+
+    # :nocov:
+
+    # Military Information Service breakers name
+    # @return [String] Military Information Service breakers name
+    def service_name
+      'EmisMilitaryInformation'
```
rtravitz: Do you think this this service name should be versioned? I'm still unclear on how exactly these service names are used in Breakers.

spec/lib/emis/military_information_service_v2_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+describe EMIS::MilitaryInformationServiceV2 do
+  describe 'get_military_service_episodes' do
+    context 'with a valid request' do
+      let(:edipi) { '1607472595' }
+
+      it 'calls the get_military_service_episodes endpoint with a proper emis message' do
+        VCR.use_cassette('emis/get_military_service_episodes/valid_v2') do
+          response = subject.get_military_service_episodes(edipi: edipi)
+          expect(response).to be_ok
+        end
```
rtravitz: It might be good to add some assertions about the expected fields.

lib/emis/service.rb
```diff
@@ -19,29 +19,37 @@ class Service < Common::Client::Base
...
-    #  are the same or an Array containing the method name and endpoint path
+    #  are the same or an Array containing the method name, endpoint path,
+    #  and optional version of response and soapaction
+    #
+    # rubocop:disable Metrics/MethodLength
     def self.create_endpoints(endpoints)
       endpoints.each do |endpoint|
         operation = nil
         request_name = nil
         if endpoint.is_a?(Array)
```
rtravitz: We can probably extract this if/else to a separate method to satisfy rubocop instead of disabling the metric. It has clear inputs and outputs.

spec/lib/emis/military_information_service_v2_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      end
+
+      it 'includes new fields from v2' do
+        VCR.use_cassette('emis/get_military_service_episodes/valid_v2') do
+          response = subject.get_military_service_episodes(edipi: edipi)
+
+          first_item = response.items.first
+          expect(first_item.narrative_reason_for_separation_txt).to eq('UNKNOWN')
+          expect(first_item.pay_plan_code).to eq('MW')
+          expect(first_item.pay_grade_code).to eq('04')
```
rtravitz: 🎉

PR #3593, opened 2019-11-26
[Search] Apply page limit based on search.gov offset limit

ncksllvn:
## Description of change

This PR fixes a minor bug on va.gov/search. Due to a limitation with the `offset` property on search.gov searches, the max page a user can visit is page 99. This is accounted for. However, if the user manually changes the page query param to 100, the current page goes haywire. 

### Example issue

https://www.va.gov/search/?page=100&query=test

image

Underlying JSON -
image

### How this fixes it
I really don't follow this pagination logic well, but from what I understand, search.gov returns a value `next_offset` that we use to derive the `current_page`. However, `next_offset` is `nil` when it reaches the search.gov offset limitation, `999`. This messes up that logic for deriving the `current_page`. This PR fixes this issue by accounting for the `total_viewable_pages` property as an additional max value for `current_page`.

There are still pagination issues here. For example, since 99 is the last viewable page, then results 981-990 are the last viewable search results, but the API's `totalEntries` property says 999. I tinkered with fixing this but it seems a little complicated. I think page 100 (result 991-1000) should be accessible though.

Ticket - https://github.com/department-of-veterans-affairs/va.gov-team/issues/2268

## Testing done

On localhost using a search.gov access key w/the FE running -
- Navigate to http://localhost:3001/search/?page=100&query=test
- Observe that the search results range doesn't go haywire

## Testing planned
This is pretty minor, so not much more besides testing in Staging before the deployment.

#### Unique to this PR

- [x] I needed a search.gov access key to be able to test this locally

#### Applies to all PRs

- [x] Appropriate logging


lib/search/pagination.rb
```diff
@@ -54,7 +54,7 @@ def current_page
 
     def pagination_object
       {
-        'current_page' => [current_page, total_pages].min,
+        'current_page' => [current_page, total_viewable_pages].min,
```
ncksllvn: Below, on Line 77, there is this -

```
def maximum_viewable_entries
      (ENTRIES_PER_PAGE * total_viewable_pages) + (ENTRIES_PER_PAGE - 1)
    end
```

I cannot figure out why that `+ (ENTRIES_PER_PAGE - 1)` is there but it seems to be the root of an issue where page 100 which should contain results 991-1000 but instead it isn't accessible.

ncksllvn:
@kfrz looking for your guidance on this.

lib/search/pagination.rb
LindseySaari: @kfrz any insight on this?

johnpaulashenfelter: @ncksllvn any reason not to just limit the *input* to 99? Avoids taking a misunderstood calculation and changing it to another misunderstood one :)

ncksllvn: The logic was written so that the page sizes are flexible and can be changed, so many values are derived. I had some trouble understanding it, and tracing it to how the front-end responds. I think it could all be simpler though.

ncksllvn:
I would like to merge this today. Can someone review it?

lib/search/pagination.rb
kfrz: @ncksllvn you're right - that's the reason we added the weird `-1` logic, the last page wouldn't be accessible otherwise. Albeit an edge case but that was the solution we shipped....

I'm in agreement with you folks, about simplifying/rewriting.

@johnpaulashenfelter I like the idea of limiting the input. Do we really need to have a flexible page size? Why not preload 99 results and handle any display pagination on the f/e (show xx results...) 

Though this change should be a good stop-gap but we *really* need to refactor this service.

PR #3594, opened 2019-11-27
VAOS facility available appointments endpoint

kreek:
## Description of change
Adds `/v0/vaos/facilities/:facility_id/available_appointments`

## Testing done
unit tests

## Testing planned
staging integration tests

#### Unique to this PR
- [ ] Adds `/v0/vaos/facilities/:facility_id/available_appointments` endpoint: service, controller, cassette, schema, serializer
- [ ] Updates common resource to include `Dry.Types(default: :nominal)` rather than `Dry::Types.module` which is deprecated 

#### Applies to all PRs

- [ ] Appropriate logging


modules/vaos/app/models/vaos/facility_availability.rb
```diff
@@ -0,0 +1,23 @@
...
+    attribute :clinic_display_start_time, Types::String
+    attribute :display_increments, Types::String
+    attribute :stop_code, Types::String
+    attribute :ask_for_check_in, Types::Bool
+    attribute :max_overbooks_per_day, Types::Integer
+    attribute :has_user_access_to_clinic, Types::Bool
+    attribute :primary_stop_code, Types::String
+    attribute :secondary_stop_code, Types::String
+    attribute :list_size, Types::Integer
+    attribute :object_type, Types::String
```
jbalboni: I don't think we ever need these object type params on the front end.

modules/vaos/app/controllers/vaos/available_appointments_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    def appt_params
+      params.require(:facility_id)
+      params.require(:start_date)
+      params.require(:end_date)
+      params.require(:clinic_ids)
+      params.permit(
+        :facility_id,
+        :start_date,
+        :end_date,
+        clinic_ids: []
```
jbalboni: Would we be passing this in on the FE in as `clinic_ids[]`? Also, not sure if it matters, but the VAOS service takes this is a comma separated list of ids (because they needed to add yet another way of passing an array in a query param, I guess).

kreek: From our FE it will the Rails way `../?clinic_ids[]=1&clinic_ids[]=2&clinic_ids[]=3`, I'll double check how we're sending it down to VAOS.

modules/vaos/app/models/vaos/facility_availability.rb
```diff
@@ -0,0 +1,22 @@
...
+    attribute :stop_code, Types::String
+    attribute :ask_for_check_in, Types::Bool
+    attribute :max_overbooks_per_day, Types::Integer
+    attribute :has_user_access_to_clinic, Types::Bool
+    attribute :primary_stop_code, Types::String
+    attribute :secondary_stop_code, Types::String
+    attribute :list_size, Types::Integer
+    attribute :empty, Types::Bool
+    attribute :appointment_time_slot, Types::Array.of(VAOS::AppointmentTimeSlot)
+  end
```
saneshark: NITPICK: Can we revisit using dry-tpes later and just keep things consistent for now? This is totally going to trip me up I think if we run into issues and you're out. I've not yet had a chance to grok dry-types but I'm pretty sure all but mastered Virtus.

lib/common/models/resource.rb
```diff
@@ -4,7 +4,7 @@
 require 'dry-struct'
 
 module Types
-  include Dry::Types.module
+  include Dry.Types(default: :nominal)
```
saneshark: Can we stick to Virtus and revisit this later? This will most certainly trip me up if I run into any issues.

modules/vaos/app/controllers/vaos/available_appointments_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    def systems_service
+      VAOS::SystemsService.new(current_user)
+    end
+
+    def appt_params
+      params.require(:facility_id)
+      params.require(:start_date)
+      params.require(:end_date)
+      params.require(:clinic_ids)
```
saneshark: you can pass all these in as part of an array.

modules/vaos/app/controllers/vaos/available_appointments_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+      )
+
+      render json: VAOS::AvailabilitySerializer.new(response)
+    end
+
+    private
+
+    def systems_service
+      VAOS::SystemsService.new(current_user)
+    end
```
saneshark: hmmm, this works, but we've been using `.for_user` elsewhere.

modules/vaos/app/models/vaos/appointment_time_slot.rb
```diff
@@ -0,0 +1,24 @@
...
+    transform_types do |type|
+      case type.name
+      when :start_date_time || :end_date_time
+        type.constructor do |value|
+          DateTime.strptime(value, '%m/%d/%Y %H:%M:%S')
+        end
+      else
+        type
+      end
+    end
```
saneshark: Please take a look at the AppointmentTime < Virtus::Attributes. Almost every VAMF case is probably captured there. I think this ends up being code duplication by introducing dry-types prematurely.

modules/vaos/app/controllers/vaos/available_appointments_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+      DateTime.parse(appt_params[:start_date]).in_time_zone
+    rescue ArgumentError
+      raise Common::Exceptions::InvalidFieldValue.new('start_date', params[:start_date])
+    end
+
+    def end_date
+      DateTime.parse(appt_params[:end_date]).in_time_zone
+    rescue ArgumentError
+      raise Common::Exceptions::InvalidFieldValue.new('end_date', params[:end_date])
+    end
```
saneshark: these methods might be worth moving into a base controller or mixin -- they're definitely used elsewhere.

modules/vaos/app/controllers/vaos/available_appointments_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    def facility_id
+      appt_params[:facility_id]
+    end
+
+    def start_date
+      DateTime.parse(appt_params[:start_date]).in_time_zone
+    rescue ArgumentError
+      raise Common::Exceptions::InvalidFieldValue.new('start_date', params[:start_date])
+    end
```
saneshark: I think start_date and end_date are used in a couple places already and either worth moving into a PORO object to validate / coerce properly, a mixin / concern, or inherited from some base VAOS controller.

modules/vaos/app/controllers/vaos/available_appointments_controller.rb
kreek: What `.for_user` do over standard ruby initialization?

saneshark: Mostly a nitpick comment I guess, but most of the other services initialize with `self.for_user`

PR #3597, opened 2019-11-30
2604 - Enabling Betamocks for Rating Info Service

kathleencrawford:
## Description of change

A new endpoint was recently implemented that retrieves a Veteran's rating info from EVSS.  Initially, betamocks was configured correctly, but due to some changes, the betamocks configuration for testing this service needs to be updated.  A developer or tester can update the mock_rating_info to true if they want to test the new rating_info service and don't have access to the EVSS dev environment locally.   

Original GitHub issue can be found here: https://github.com/department-of-veterans-affairs/va.gov-team/issues/2604

## Testing done

Local Testing

## Testing planned

Testing with FE engineers.  

#### Unique to this PR

- [x] Submitted a PR in vets-api-mockdata to add the new betamocks files here: https://github.com/department-of-veterans-affairs/vets-api-mockdata/pull/107

#### Applies to all PRs

- [x] Appropriate logging


lib/evss/common_service.rb
```diff
@@ -7,6 +7,11 @@ class CommonService < BaseService
     API_VERSION = Settings.evss.versions.common
     BASE_URL = "#{Settings.evss.url}/wss-common-services-web-#{API_VERSION}/rest/"
 
+    def initialize(*args)
+      super
+      @use_mock = Settings.evss.mock_rating_info || false
```
LindseySaari: Looks straight forward. 👍As a suggestion to keep the code more uniform, wondering if the conn method in EVSS base_service.rb should be moved to a configuration class instead

charleystran: Why are we modify the common service? I think these settings belong in inherited class like what is done in VSO search and reference data.

bastosmichael: I agree with Charlie, there is a lot of critical code that depends on this common service object so something like this should be done much higher up the chain if at all possible.

kathleencrawford: Thanks @charleystran and @bastosmichael.  My original PR had its own custom service and configuration classes (similar to VSO search), but I was directed to use common service since the base URL and configuration already lived there, and I was just adding another endpoint.  However, the mock configuration is not enabled for common services, so I'm not able to use betamocks locally to test the new endpoint.  If someone could give me direction on how to just enable betamocks without duplicating, I'm open to suggestions.  @johnpaulashenfelter tagging you as well for visibility so we can discuss further.

charleystran: The issue I see is that all requests to common service will be mocked or not mocked based on the rating info flag. We do want the dev servers to still use betamocks even though we have some local access now. It may make sense to change your flag to be to mock_common_service vs mock_rating_info

johnpaulashenfelter: @bastosmichael @charleystran I was going to mention the same thing. @omgitsbillryan had asked her to do this: https://github.com/department-of-veterans-affairs/vets-api/pull/3446

charleystran: A closer look informed my last comments. Common is not shared like other common services. Naming is fun. However I still recommend that the mock flag be to the whole service and not the endpoint.

charleystran: Also one interesting note is that vso_search also is a common service so we do have a pattern to break these into individual services.

kathleencrawford: Thank you all for your comments.  I went ahead and updated the settings flag to "mock_common_service" as requested by @charleystran

omgitsbillryan: > I still recommend that the mock flag be to the whole service and not the endpoint

Agree with this ☝️. I'd define "service" as a collection of "endpoints" that share configuration. So we have to pick our poison - either we mock a single endpoint in a service or we define a new service along with new duplicated config.

I think we've run up into some tech debt here. This `CommonService` doesn't use our `Configuration` model. Also, there doesn't seem to be anything at all "common" about this service.

omgitsbillryan:
I had previously recommended that you re-use this `CommonService` so as not to duplicate configuration. At that time, I didn't realize that the config had [_already_ been duplicated](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/evss/vso_search/configuration.rb#L10-L16).

There's a fair bit of tech debt in our services hierarchy. For example, this `CommonService` (along with a few others) does not implement the `Configuration` abstraction that most of our other services use. My understanding of a "service" is a collection of "endpoints" that share configuration.

omgitsbillryan:
Be sure to update the devops settings.local.yml files for dev, staging, & prod
https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/deployment/config/vets-api/dev-settings.local.yml.j2

Also, since we will now be mocking the `CommonService` in `dev`, be sure that not having `vets-api-mockdata` for `def create_user_account` doesn't break login. I'm 90% sure that when betamocks middleware doesn't match the request (via the `services_config.yml`) that it just passes the request along, but just be sure to double-check.

kfrz:
@kathleencrawford could you connect this PR with the associated issue? I think it's [#4257](https://github.com/department-of-veterans-affairs/va.gov-team/issues/4257) but the issue is in closed pipeline, so it won't allow ZH to connect it.

kathleencrawford:
@kfrz I've re-opened the ticket and connected it again...didn't realize that Zenhub removes the connection when the ticket gets closed.

kathleencrawford:
I haven't merged this in yet, because the comments caused some confusion.  My intent was to have the mock enabled flag set to false by default, I didn't realize that in dev we set them to true by default.  Is there someone I can discuss this with further, @johnpaulashenfelter or someone else?  I don't want to inadvertently affect services that I didn't change/touch.

kathleencrawford:
Updated the devops settings.local.yml files for dev, sandbox, staging and prod.

https://github.com/department-of-veterans-affairs/devops/pull/6166

kathleencrawford:
@charleystran I updated the settings flag to true as discussed.  Thank you for your help!  Do I need to wait until the devops PR has been merged in before I merge this one?  Thanks!

PR #3600, opened 2019-12-02
Fix/3722 missing cc appointment request

kreek:
## Description of change
Fixes Community Care requests not having a cc_appointment_request object https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/3722

## Testing done
unit tests

## Testing planned
staging integrations tests

#### Unique to this PR
- [ ] Updates service and serializers to include the cc_appointment_request object

#### Applies to all PRs

- [ ] Appropriate logging


modules/vaos/spec/services/appointment_requests_service_spec.rb
```diff
@@ -60,11 +59,37 @@
...
+    context 'without data params but with cc appointment data' do
+      it 'includes the cc appointment data' do
         VCR.use_cassette('vaos/appointment_requests/get_requests', match_requests_on: %i[method uri]) do
           response = subject.get_requests
-          expect(response[:data].size).to eq(40)
+          request_with_cc = response[:data][40]
+          expect(request_with_cc.cc_appointment_request).to eq(
+            data_identifier: {},
+            patient_identifier: {},
+            surrogate_identifier: {},
```
saneshark: we might need to drop some of these parameters if they include PII. I know patient_identifier does this a lot. You should be able to find some examples where I've needed to normalize the data for the FE.

modules/vaos/spec/services/appointment_requests_service_spec.rb
```diff
@@ -60,11 +59,37 @@
...
+            office_hours: ['Evenings'],
+            preferred_providers: [
+              {
+                first_name: 'Vilasini',
+                last_name: 'Reddy',
+                practice_name: 'Test clinic 2',
+                address: {},
+                preferred_order: 0,
+                object_type: 'Provider',
+                link: []
```
saneshark: drop link

modules/vaos/spec/services/appointment_requests_service_spec.rb
```diff
@@ -60,11 +59,37 @@
...
+              }
+            ],
+            object_type: 'CCAppointmentRequest',
+            link: []
```
saneshark: drop link

modules/vaos/spec/services/appointment_requests_service_spec.rb
saneshark: nevermind -- this PR will address

modules/vaos/spec/services/appointment_requests_service_spec.rb
saneshark: i *think* this PR will address this if i'm not mistaken.

modules/vaos/spec/request/appointment_requests_request_spec.rb
```diff
@@ -84,7 +84,7 @@
 
           expect(response).to have_http_status(:success)
           expect(response.body).to be_a(String)
-          expect(response).to match_response_schema('vaos/appointment_requests')
+          expect(response).to match_response_schema('vaos/appointment_requests', strict: false)
```
saneshark: hmm, is it necessary because the object is sometimes nil and sometimes a hash? I wonder if this will evaluate to with strict: false if you change nil valued ones to an empty hash maybe?

modules/vaos/spec/request/appointment_requests_request_spec.rb
```diff
@@ -132,7 +132,7 @@
         put "/v0/vaos/appointment_requests/#{id}", params: params
         expect(response).to have_http_status(:success)
         expect(response.body).to be_a(String)
-        expect(json_body_for(response)).to match_schema('vaos/appointment_request')
+        expect(json_body_for(response)).to match_schema('vaos/appointment_request', strict: false)
```
saneshark: same comment here

modules/vaos/app/serializers/vaos/appointment_requests_serializer.rb
```diff
@@ -44,6 +44,7 @@ class AppointmentRequestsSerializer
                :booked_appt_date_time,
                :type_of_care_id,
                :friendly_location_name,
+               :cc_appointment_request,
```
saneshark: is there anything clever we can do here to still have the json evaluate strictly? It might not be possible to do, but I'm asking because VAMF has a tendency to just drop null-valued keys. Wondering if doing a reverse_merge or something to ensure it's always at least an empty hash would allow you to still evaluate with `strict: true`

modules/vaos/app/serializers/vaos/appointment_requests_serializer.rb
```diff
@@ -4,14 +4,23 @@ module VAOS
...
+    attribute :patient do |object|
+      object.patient.slice(:inpatient, :text_messaging_allowed)
+    end
+
+    attribute :appointment_request_detail_code do |object|
+      Array.wrap(object.appointment_request_detail_code).map do |element|
+        element.slice(:appointment_request_detail_code_id, :created_date)
+               .merge(detail_code: element[:detail_code].slice(:code, :provider_message, :veteran_message))
+      end
     end
```
omgitsbillryan: What in the world is happening here?? 😎

omgitsbillryan: Is there some way to make this more readable?

kreek: yep, this was merged in from another branch should be cleaner now

modules/vaos/app/serializers/vaos/appointment_requests_serializer.rb
kreek: got it to evaluate strictly using `oneOf` in the schema. I think this is better for the FE otherwise they'll be checking for empty hashes rather than null 😄

kreek:
@kam I merged your branch in but made an update to appointment serializer appointment_request_detail_code attribute

modules/vaos/app/serializers/vaos/appointment_requests_serializer.rb
saneshark: sounds good! as long as we're not just dropping the root level key, I'm ok with that. Root level objects should ideally always be null or object, not just dropped. 😂

modules/vaos/app/serializers/vaos/appointment_requests_serializer.rb
saneshark: basically appointment_request_detail_code is an array of complex object that includes a bunch of stuff including patient identifiers that we'd like to NOT pass up to the FE. 

Perhaps it could have its own serializer in the future, but we'll revisit that when we refactor.

edmangimelli: i like it 👍 
although the double slice caught me off-guard.
seems pretty rubyish

saneshark: it's technically more performant than doing a merge! or slice! when iterating a large array. I'm not sure that the changes made make it any more "readable." The best way to address readability would be to give its own serializer and we'll get to that eventually when we refactor. This would have the added benefit of ensuring that the data coming in doesn't include some key we've never seen before that includes some kind of PII we don't want to expose to the FE. We're doing a ton of regression testing and haven't seen anything like that yet (and apparently it's ok in the legacy app we're replacing even if we were to find it), but something we'd prefer to avoid where possible.

PR #3602, opened 2019-12-03
speed up VAOS specs 3x by loading rsa key instead of generating new one

saneshark:
## Description of change
Speeds up the specs for the VAOS test suite by 3x, from 18 seconds to 6 seconds simply by loading a fake RSA private key from a fixture vs generating one for testing.

## Testing done
Locally running `bundle exec rspec modules/vaos --profile` to identify slow running specs.

## Testing planned
none

#### Unique to this PR

#### Applies to all PRs

- [x] Appropriate logging


saneshark:
I can't make sense of the failing spec? 🤷‍♂

modules/vaos/spec/services/jwt_spec.rb
```diff
@@ -1,12 +1,13 @@
...
 
 require 'rails_helper'
+require_relative '../support/fixture_helper'
 
 describe VAOS::JWT do
   subject { VAOS::JWT.new(user) }
 
   let(:user) { build(:user, :vaos) }
-  let(:rsa_private) { OpenSSL::PKey::RSA.generate(4096) }
+  let(:rsa_private) { OpenSSL::PKey::RSA.new(read_fixture_file('open_ssl_rsa_private.pem')) }
```
johnpaulashenfelter: Definitely better.

It's worth documenting somewhere how the file is generated (maybe just a comment line in the file itself?) since now its not necessarily clear that it was with the 4096 param.

saneshark: I'm not sure it needs to be anything specific per se, certainly not for specs. I say this because according to jwt.io you can have a lot of different types of algos, it's part of the header. Would think one could infer the algo from the header. Usually the first few bytes includes this information.

saneshark: I'm happy to add a comment if you feel it's necessary, but for example:

`openssl rsa -in test.key -text | head -n 1`                                                                                                                                       
Private-Key: (4096 bit)

PR #3603, opened 2019-12-03
VAOS: first pass at removing pii and unneeded attributes from serialized data

saneshark:
## Description of change
This is the first pass removing unneeded data from serialized data sent to the FE. In addition we're removing all link objects since these often include ICN and EDIPI information.

## Testing done
Local

## Testing planned
Staging

Need @jbalboni to review the schema changes and let me know if he is comfortable with the changes being made.

#### Unique to this PR

#### Applies to all PRs

- [X] Appropriate logging
- [X] Swagger docs have been updated, if applicable

### DIFF
--- a/modules/vaos/app/serializers/vaos/appointment_requests_serializer.rb
+++ b/modules/vaos/app/serializers/vaos/appointment_requests_serializer.rb
@@ -4,14 +4,23 @@ module VAOS
   class AppointmentRequestsSerializer
     include FastJsonapi::ObjectSerializer
 
-    set_id do |object|
-      object.data_identifier[:unique_id]
-    end
+    set_id(&:appointment_request_id)
 
     set_type :appointment_requests
 
     attribute :facility do |object|
-      object.facility.reverse_merge(type: nil, address: nil)
+      object.facility.reverse_merge(type: nil, address: nil).except(:link, :object_type)
+    end
+
+    attribute :patient do |object|
+      object.patient.slice(:inpatient, :text_messaging_allowed)
+    end
+
+    attribute :appointment_request_detail_code do |object|
+      Array.wrap(object.appointment_request_detail_code).map do |element|
+        element.slice(:appointment_request_detail_code_id, :created_date)
+               .merge(detail_code: element[:detail_code].slice(:code, :provider_message, :veteran_message))
+      end
     end
 
     attributes :last_updated_at,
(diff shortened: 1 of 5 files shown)


saneshark:
I think we can probably go ahead and remove object_type from all the serializers, since we're setting type in the serializer ourselves.

modules/vaos/app/serializers/vaos/clinic_serializer.rb
```diff
@@ -17,7 +17,6 @@ class ClinicSerializer
                :display_to_patient_flag,
                :institution_name,
                :institution_code,
-               :object_type,
-               :link
+               :object_type
```
jholton: @saneshark did you mean to leave this `:object_type`?
> I think we can probably go ahead and remove object_type from all the serializers, since we're setting type in the serializer ourselves.

modules/vaos/app/serializers/vaos/appointment_requests_serializer.rb
```diff
@@ -34,25 +43,15 @@ class AppointmentRequestsSerializer
...
-               :appointment_request_detail_code,
                :has_veteran_new_message,
                :has_provider_new_message,
                :provider_seen_appointment_request,
                :requested_phone_call,
                :booked_appt_date_time,
                :type_of_care_id,
                :friendly_location_name,
-               :patient_id,
-               :appointment_request_id,
```
jbalboni: We use this variable on the front end, but maybe we shouldn't?

modules/vaos/app/serializers/vaos/clinic_serializer.rb
saneshark: i can remove it as long as we remove it and set our own "type" attribute as part of JSON API spec consistently.

modules/vaos/app/serializers/vaos/appointment_requests_serializer.rb
saneshark: you're using appointment_request_id or patient_id? 

patient_id happened to be EDIPI for Judy Morrison. I'm not sure if it is the same for other users or not. 

appointment_request_id I could probably leave in if it makes it easier for you to make requests that way for messages and other things. It's just a duplicate of both id and unique_id. We add id to be consistent with other endpoints -- to me it seemed redundent, but you bring up a good point that it would be necessary for posting messages and stuff.

jbalboni: Yeah, appointment_request_id. We could get it elsewhere, but let's leave it in for now, so we don't break the front end. If we want to remove it later, we can update the front end first and then remove it from the back end.

saneshark:
Going to close this PR for now, but keep the branch locally. Will sync with you @jbalboni and we can kind of step through and identify places for quick improvements when more of the endpoints are fully fleshed out.

PR #3604, opened 2019-12-03
fixes best_timeto_call getting blanked out

saneshark:
## Description of change
https://github.com/department-of-veterans-affairs/va.gov-team/issues/3884

You'll notice that there are no tests to verify this. I would love to have a better way of writing specs for this type of thing, but we record a successful VCR cassette as part of the service interaction since thats a unit level, and we replay the same cassette for request specs. Even if I introduce specs for the form object, that is going to accept a normal hash. I don't know of a good way to test and verify ActiveSupport::Parameters. Maybe for my FormObject I should initialize an ActiveSupport::Parameters object but then i'd have to make sure to also pass it in similar to the controller. Maybe the FormObject should get initialized in the controller instead of the service, but this seems wonky too. 🤷‍♂ 

I could re-record VCR cassettes as part of request spec, but then any future changes to `ActiveSupport::Parameters#permit` would have the same issue of a previous cassette returning success.

## Testing planned
Local

Verify working on staging

#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging


johnpaulashenfelter:
Honestly I lean towards form object in the controller since you can let it handle all the validation for you. But this keeps with the existing code patterns so 🤷‍♂

saneshark:
> Honestly I lean towards form object in the controller since you can let it handle all the validation for you. But this keeps with the existing code patterns so 🤷‍♂

Yup, I totally agree with you. My comment about:

> Maybe the FormObject should get initialized in the controller instead of the service, but this seems wonky too

is that I shouldn't have to change that simply to be able to test something related to ActiveSupport::Parameters... it strikes me as a bad abstraction of how rails manages mass assignment. But I totally agree with you that doing the form validation in the controller makes the most sense and will be happy to do a refactor for this once the dust settles.

PR #3606, opened 2019-12-03
No POA checking in 2122 end points

bastosmichael:
## Description of change
Looks like we didn't implement the poa check logic properly for 2122 and need to bypass the poa check in order to allow VSO's to submit POA's for Veteran's they don't yet have POA access for etc.

Resolves https://github.com/department-of-veterans-affairs/vets-contrib/issues/3801

- Allow VSO's to submit POA forms on behalf of Veterans they don't yet have POA access for.
- Ensure we are tracking whom is doing the submission and saving to source

#### Applies to all PRs

- [ ] Appropriate logging


bastosmichael:
@charleystran because while the poa_request? method is utilized in other parts of the code no specifically associated with the poa validation, the verify_power_of_attorney method is really the only one that needs to be called in order to verify poa ownership in some way, I could rename that concern to poa ownership concern and move other poa related stuff into a separate concern but determining whether something is a poa request vs owership of a poa need to be kept as separate concerns.

bastosmichael:
@charleystran Added contextual comments to each sidekiq conditional in order to explain when and why certain jobs are happening and why etc.

PR #3607, opened 2019-12-04
introduce mvi rake task to fetch mvi attributes given csv

saneshark:
## Description of change
Not sure if anyone has any value in this; but it was very handy for me so I'm submitting as PR. In particular because DS Logon sign on staging tends to be unreliable, the idea is that having these users in ID.me would give us an alternate means of signing in.

We recently needed to have some users validated and properly staged in ID.me's lower environment such that signing in as these users would properly fetch their MVI records. What ID.me needs to be able to stage these users are the set of SAML attributes: `first_name, middle_name, last_name, gender, birth_date, ssn, address`

Given a CSV file with a column ICN and any number of additional columns, we want to create a new CSV appending to the old these additional columns.

For example if my input CSV named 'icns_for_staging.csv' is:

```csv
icn, notes
1013124304V115761, (patient999@id.me)
1013125218V696863, (patient998@id.me)
1013125219V724755, (patient997@id.me)
1012845331V153043, (judy.morrison)
1012853802V084487, (ruben.moreno)
```

then my output file will be 'icns_for_staging.csv.out:
```csv
icn, notes,first_name,middle_name,last_name,gender,birth_date,ssn,address
1013124304V115761, (patient999@id.me),Patient,"",Zztest,F,19620101,999999998,"{""street"":""123 Easy Street"",""city"":""New Orleans"",""state"":""LA"",""postal_code"":""58321"",""country"":""USA""}"
1013125218V696863, (patient998@id.me),Patient,"",Zztest,M,19720101,999999997,"{""street"":""122 Main Street"",""city"":""Boston"",""state"":""MA"",""postal_code"":""02101"",""country"":""USA""}"
1013125219V724755, (patient997@id.me),Patient,"",Zztest,M,19820101,999999996,"{""street"":""123 Main Street"",""city"":""Los Angeles"",""state"":""CA"",""postal_code"":""90001"",""country"":""USA""}"
1012845331V153043, (judy.morrison),Judy,Snow,Morrison,F,19530401,796061976,null
1012853802V084487, (ruben.moreno),Ruben,Gustav,Moreno,M,19660516,796264187,"{""street"":""9813 ST CHARLES PL"",""city"":""DALLAS"",""state"":""TX"",""postal_code"":""75253"",""country"":""USA""}"
```

## Testing done
Verified on Staging

## Testing planned
none

#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging


johnpaulashenfelter:
@saneshark seems like this should live in a `dev` or `staging` rake task to make it clear that its for "devs" or its useful on "staging" -- I'm kind of surprised we don't have a `dev.rake` for random things that are useful like this.

saneshark:
> Cool, this seems very useful since ID.me needs the additional columns to fetch MVI records. It looks like there may be a few rubocop errors, but otherwise I think it's a nice tool. Could the 'notes' header be renamed to 'username'?

@LindseySaari I'll clean up the rubocop errors. The notes header can be called whatever you like. I just used that as my example. You can upload any csv, but as long as there is an icn column, thats all that is referenced in the rake task. The additional attributes are appended columns to the original. So if it were called email or username it wouldn't make any difference. :)

> seems like this should live in a dev or staging rake task to make it clear that its for "devs" or its useful on "staging" -- I'm kind of surprised we don't have a dev.rake for random things that are useful like this.

@johnpaulashenfelter I think you'll find that most of the rake tasks are mostly meant for staging or dev environments. To me it seems more logical that it be categorized based on the service it is invoking seeing as rake tasks have an environment argument. And if anything, if we want to limit a particular rake task from being used on production we would maybe have some kind of Settings.yml check. I could some value to this rake task on production too should we ever need to stage production users. MVI has a few fake users setup in production I believe.

PR #3610, opened 2019-12-04
Add Breakers to VAOS

annaswims:
## Description of change
VAOS was missing from the breakers initializer so we it wasn't reporting stats.

## Testing done
spec

## Testing planned
confirm presence of VAOS in [grafana](http://grafana.vfs.va.gov/d/000000025/external-service-status?orgId=1&fullscreen&edit&panelId=11)


saneshark:
@kreek I think this also explains maybe why we were not getting STATSD. The STATSD plugin is part of breakers. 

@johnpaulashenfelter I think doing some variety of load testing and making sure that there are proper grafana dashboards in place is (or should be) part of a checklist. This would have certainly been caught before a broader roll out, but I do share your concern.

annaswims:
I updated the example at https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/engineering/backend/tutorials/endpoint-punch-list-presentation/supporting_code/service-object-implementation.md . 

and we have on " Breakers initializer" in the [Vets-API Endpoint Punch List]( https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/engineering/backend/vets-api-endpoint-punch-list.md)

PR #3612, opened 2019-12-04
Kr/fix expired token error

edmkitty:
See https://github.com/department-of-veterans-affairs/vets-contrib/issues/2047


app/controllers/application_controller.rb
```diff
@@ -87,7 +87,7 @@ def skip_sentry_exception_types
       case exception
       when Pundit::NotAuthorizedError
         Common::Exceptions::Forbidden.new(detail: 'User does not have access to the requested resource')
-      when JWT::ExpiredSignature
+      when OpenidApplicationController::ExpiredTokenError
```
LindseySaari: Just a note: @kfrz We may run into a merge here in our ApplicationController Error Handling refactor

app/controllers/openid_application_controller.rb
```diff
@@ -11,6 +11,11 @@
 class OpenidApplicationController < ApplicationController
   before_action :authenticate
   TOKEN_REGEX = /Bearer /.freeze
+  class ValidationError < StandardError
```
LindseySaari: Wondering if there is a better place to define custom error types?

app/controllers/openid_application_controller.rb
```diff
@@ -51,6 +56,10 @@ def establish_session
     @current_user = OpenidUser.build_from_identity(identity: user_identity, ttl: ttl)
     @session = build_session(token, token_identifiers.uuid, ttl)
     @session.save && user_identity.save && @current_user.save
+  rescue JWT::ExpiredSignature
+    raise ExpiredTokenError
+  rescue ExpiredTokenError
+    raise ExpiredTokenError
   rescue => e
     Rails.logger.warn(e)
```
edmkitty: Can anybody explain to me why we are swallowing errors here? Validation errors get raised and then rescued and then logged, but I have no clue why.

app/controllers/openid_application_controller.rb
edmkitty: I think I would prefer making a series of Error classes under Common::Exceptions that deal specifically with OAuth, since this isn't the only place tokens are used. This would make it easier to identify/handle in the Application Controller as well. However, I am unsure if the health team would find this useful or use it at all. I would love to hear your thoughts on this idea.

app/controllers/application_controller.rb
edmkitty: If there is anything I can do to make that merge conflict go smoother, please let me know! I would happy to change my code to conform to yours.

spec/controllers/openid_application_controller_spec.rb
```diff
@@ -52,6 +52,19 @@ def index
...
       end
+
+      it 'rejects access when expired' do
+        with_okta_configured do
+          exp_token = token
+          exp_token[0]['iat'] = Time.current.utc.to_i - 7200
+          exp_token[0]['exp'] = Time.current.utc.to_i - 3600
+          allow(JWT).to receive(:decode).and_return(exp_token)
+          get :index
+          expect(response.status).to eq(403)
```
edmkitty: This should be a 401, my bad.

app/controllers/openid_application_controller.rb
rtravitz: @edmkitty If you're talking about the team that manages [this health api](https://developer.va.gov/explore/health/docs/fhir?version=current), then they live outside of the context of vets-api. Kong forwards to a set of Java applications with their own client-facing error handling.

rtravitz: At the moment it looks like only `veteran_verification` and `claims` are using the `OpenidApplicationController` in vets-api.

app/controllers/openid_application_controller.rb
patrickvinograd: When validate_token raises an error, it will jump to the rescue block and skip over establishing the `@session` object and saving it to Redis. Then when it returns to `authenticate_token`, `@session` is nil, so that method returns false, and causes the `authenticate` method to render_unauthorized, which will generate the right HTTP status code.

If you don't want to swallow the error, that's fine, but some caller has to do something meaningful with it, (which could be for example, rendering the unauthorized error with more detail about why, such as "token is expired".

patrickvinograd: If you just removed the `rescue`, I think it would probably result in a 500 error somewhere in the parent `ApplicationController`'s `rescue_from 'Exception' do |exception|` block.

app/controllers/openid_application_controller.rb
```diff
@@ -69,10 +69,23 @@ def token_payload
...
+  def valid_issuer?
+    token_payload['iss'] == Settings.oidc.issuer
+  end
+
+  def valid_audience?
+    token_payload['aud'] == Settings.oidc.audience
+  end
+
+  def valid_expiration?
+    token_payload['exp'] >= Time.current.utc.to_i
```
LindseySaari: Nice refactor here 😸

app/controllers/openid_application_controller.rb
```diff
@@ -109,6 +123,14 @@ def expected_key(token)
     decoded_token = JWT.decode(token, nil, false, algorithm: 'RS256')
     kid = decoded_token[1]['kid']
     OIDC::KeyService.get_key(kid)
+  rescue JWT::ExpiredSignature
+    raise error_klass('Validation error: token has expired')
+  rescue JWT::DecodeError
+    raise error_klass('Validation error: token could not be validated')
```
omgitsbillryan:
```suggestion
    raise error_klass('Validation error: token could not be decoded')
```
Bit of a cyclical error message as written :thinking: , proposed a new change

edmkitty:
> Does the swagger doc need to be updated with this 401?
> 
> There's a lot of validation logic WRT the `token_payload` bloating this controller. Maybe we could create a new model and put that validation logic there? Otherwise, the code looks quite good, and LGTM.

Good idea. I've made a new class to encapsulate the token logic, take a look. I'm fairly sure the swagger doesn't need to be updated, since the 500 response was a bug, but I'll double check.

app/models/token.rb
```diff
@@ -0,0 +1,73 @@
...
+    @payload ||= if @token_string
+                   pubkey = expected_key
+                   return if pubkey.blank?
+
+                   JWT.decode(@token_string, pubkey, true, algorithm: 'RS256')[0]
+                 end
+  rescue JWT::ExpiredSignature
+    raise error_klass('Validation error: token has expired')
+  rescue JWT::DecodeError
+    raise error_klass('Validation error: token could not be decoded')
```
kalilsn: There's a lot of different kinds of `JWT::DecodeError` that could be raised here (such as https://github.com/jwt/ruby-jwt/blob/v2.1.0/lib/jwt/verify.rb#L32-L82). It looks to me like the messages from these errors are generally suitable for raising to the user, so could we do that?

If that makes you queasy, could we at least log the more specific error details here?

app/models/token.rb
```diff
@@ -0,0 +1,73 @@
...
+  end
+
+  def expected_key
+    decoded_token = JWT.decode(@token_string, nil, false, algorithm: 'RS256')
+    kid = decoded_token[1]['kid']
+    OIDC::KeyService.get_key(kid)
+  rescue JWT::ExpiredSignature
+    raise error_klass('Validation error: token has expired')
+  rescue JWT::DecodeError
+    raise error_klass('Validation error: token could not be validated')
```
kalilsn: Since we have this custom error handling logic for `JWT.decode` could we wrap that in a method so we can deduplicate it (and therefore keep it in sync)?

kalilsn: Nevermind, this call to `JWT.decode` doesn't do verification, so I actually think there's no need for these rescues at all.

kalilsn: Instead, I think we should just raise a more specific error like "Unable to get `kid` from JWT" or something like that

app/models/token.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  def payload
+    @payload ||= if @token_string
+                   pubkey = expected_key
+                   return if pubkey.blank?
+
+                   JWT.decode(@token_string, pubkey, true, algorithm: 'RS256')[0]
+                 end
+  rescue JWT::ExpiredSignature
+    raise error_klass('Validation error: token has expired')
```
kalilsn: This is actually the reverse of what I was trying to suggest. I think we should be raising the original JWT error message here like you're doing down below:

```ruby
  rescue JWT::DecodeError => e
    raise error_klass("Validation error: #{e.message}")
```

app/models/token.rb
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  def expected_key
```
kalilsn: Unless I'm misunderstanding what this does, I believe this method extracts the `kid` from our JWT, then looks up + returns the associated public key from Okta. What about just `public_key`?

app/models/token.rb
```diff
@@ -0,0 +1,69 @@
...
+    payload['iss'] == Settings.oidc.issuer
+  end
+
+  def valid_audience?
+    payload['aud'] == Settings.oidc.audience
+  end
+
+  def valid_expiration?
+    payload['exp'] >= Time.current.utc.to_i
+  end
```
kalilsn: We probably don't need to do this at all since the jwt library is already throwing an error if the token is expired

app/models/token.rb
omgitsbillryan: If `e.message` has a dynamic element to it (e.g. "blah blah token ending in xyz123 cannot be decided") then sentry won't group it properly. 

Also, there are some scenarios I end up reading exceptions in a shell or somewhere where the line breaks are `\n` instead of actual line breaks, making the stack trace difficult to read, to the point where being able to simple grep through our code for an exception that is obviously ours becomes a real time saver.

app/models/token.rb
edmkitty: I would prefer to keep this in to fail early

app/models/token.rb
edmkitty: Will it group improperly even if the error class is the same for each dynamic message? I don't think the stack trace will be an issue, since we aren't sending the stack trace as a string here.

kalilsn: I'm not sure that matters. Doesn't the change to `app/controllers/application_controller.rb` mean that these errors won't go to sentry at all?

edmkitty: Ha, you're right Kalil, I didn't even realize that.

PR #3613, opened 2019-12-05
VAOS facility visits endpoint

kreek:
## Description of change
Adds the `/v0/vaos/facilities/:facility_id/visits/:schedule_type` endpoint. Ticket # [2993](https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/2993)

## Testing done
unit tests

## Testing planned
staging integration tests

#### Unique to this PR

- [x] adds the `/v0/vaos/facilities/:facility_id/visits/:schedule_type` endpoint

#### Applies to all PRs

- [x] Appropriate logging


modules/vaos/config/routes.rb
```diff
@@ -16,6 +16,7 @@
       resources :cancel_reasons, only: :index
       resources :available_appointments, only: :index
       resources :limits, only: :index
+      get 'visits/:schedule_type', to: 'visits#index'
```
saneshark: There's probably a good reason for this but I'm not seeing it. Why not controller_name#index?

kreek: we need `../:schedule_type` appended, is there another way to do that?

saneshark: Ahh, right I see. If schedule_type is something limited, might want to limit the route to only those constraints, easy to do if constraints are a constant in your controller.

```
# from config/routes.rb
get '/sessions/:type/new',
      to: 'v0/sessions#new',
      constraints: ->(request) { V0::SessionsController::REDIRECT_URLS.include?(request.path_parameters[:type]) }
```

but also not critical if you're doing that in the controller already.

saneshark: I personally think for readability it helps to know how routes are limited

saneshark: alternately just do 

`resources :visits, only: :index`

and just make `:schedule_type` a required parameter

saneshark: If we're going to do anything custom like this for vanity urls, I would suggest introducing routing specs `specs/routing/vaos/visits_spec.rb` or something too.

modules/vaos/app/services/vaos/systems_service.rb
```diff
@@ -52,9 +52,9 @@ def get_facility_clinics(facility_id, type_of_care_id, system_id)
       with_monitoring do
         url = "/var/VeteranAppointmentRequestService/v4/rest/clinical-services/patient/ICN/#{@user.icn}/clinics"
         url_params = {
-          'three-digit-code' => facility_id,
+          'three-digit-code' => system_id,
           'clinical-service' => type_of_care_id,
-          'institution-code' => system_id
+          'institution-code' => facility_id
```
LindseySaari: Just want to double check that the system_id and facility_id should be swappeed

modules/vaos/app/controllers/vaos/visits_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+    SCHEDULE_TYPES = %w[direct request].freeze
+
+    def index
+      response = systems_service.get_facility_visits(system_id, facility_id, type_of_care_id, schedule_type)
+      render json: VAOS::FacilityVisitSerializer.new(response)
+    end
+
+    private
+
+    def systems_service
```
annaswims: I don't have any strong feelings on this, but i looks strange to me.

Why not just something like 
````ruby
def visit_params
  params.require([:system_id, :facility_id, :type_of_care_id, :schedule_type, :type_of_care_id])
end
````

saneshark: I agree, could do it that way -- what's problematic is when you then try to chain permit on the array, but just saying what is required is not a problem.

I think I'll fix this when we go in and refactor. There's a lot of cleanup ahead to make the work we've done more readable and consistent.

PR #3616, opened 2019-12-05
Create Forms route consumable for www.va.gov

ncksllvn:
## Description of change

This PR creates a new API method, `v0/forms?query=information`, that basically signs the request with an `apiKey` in the headers in order to leverage the Lighthouse [vaForms](https://developer.va.gov/explore/vaForms/docs/vaForms?version=current) method.

Ticket - https://github.com/department-of-veterans-affairs/va.gov-team/issues/3563

### Later steps
- [ ] Create a Lighthouse `apiKey` for vets-api and add it into Credstash. Any guidance on this?
- [ ] Add Swagger docs. Does this have to be done now?

## Testing done

- Added spec and cassettes

### How to run this
- Pull this branch
- In your `settings.local.yml`, add a `lighthouse.api_key` property with an API key authorized for the Lighthouse Forms API
- Start the API
- Navigate to `http://localhost:3000/v0/forms?query=information`

## Testing planned

- Testing for awhile in Staging. This isn't used by the front-end website yet.

#### Unique to this PR

- [x] Doesn't do anything w/o local configuration, because we need to create a Lighthouse `apiKey` for Vets-API.

#### Applies to all PRs

- [x] Appropriate logging


config/settings.yml
```diff
@@ -443,6 +443,13 @@ va_mobile:
   key_path: /fake/client/key/path
   timeout: 15
 
+lighthouse:
+  api_key: fake_key
```
ncksllvn: The idea here is that we can share this API key across any services that leverage Lighthouse APIs.

lib/forms/client.rb
```diff
@@ -0,0 +1,63 @@
...
+      end
+    end
+
+    def save_error_details(error_message)
+      Raven.extra_context(
+        message: error_message,
+        url: config.base_path
+      )
+      Raven.tags_context(search: 'general_forms_query_error')
+    end
```
ncksllvn: I borrowed most of this from other services/clients. I could use some guidance on if this monitoring is set up right. There doesn't seem to be any specific errors to look out for in the Lighthouse VA Forms API right now.

ncksllvn:
Omg, my specs pass. The only thing is Code Climate for identical blocks of code, but I think that's okay.

spec/lib/forms/client_spec.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe Forms::Client do
+  subject { described_class.new(query) }
+
+  let(:query) { nil }
+
+  describe '#get_all' do
```
SMLuthi: Are there any specific errors that the GET call can return? Wondering if those should be covered in the tests to make sure we handle them properly. Most likely you cant get a cassette for it...

spec/support/vcr_cassettes/forms/200_all_forms.yml
```diff
@@ -0,0 +1,345 @@
...
+      string: ''
+    headers:
+      Accept:
+      - application/json
+      Content-Type:
+      - application/json
+      User-Agent:
+      - Vets.gov Agent
+      Apikey:
+      - 5sBmaAWUXwjUtG1qO7yZI6bAtYpg9KPH
```
SMLuthi: Im assuming this is your test apikey? You should replace this with a fake one

spec/support/vcr_cassettes/forms/200_form_query.yml
```diff
@@ -0,0 +1,122 @@
...
+      string: ''
+    headers:
+      Accept:
+      - application/json
+      Content-Type:
+      - application/json
+      User-Agent:
+      - Vets.gov Agent
+      Apikey:
+      - 5sBmaAWUXwjUtG1qO7yZI6bAtYpg9KPH
```
SMLuthi: Im assuming this is your test apikey? You should replace this with a fake one

spec/support/vcr_cassettes/forms/200_all_forms.yml
ncksllvn: That is an excellent idea. I'm gonna replace it, but also have it regenerated.

spec/lib/forms/client_spec.rb
ncksllvn: Yeah, nothing listed right now in the dev docs. There's a Lighthouse route for an API healthcheck that I'll likely need to add later though.

app/controllers/v0/forms_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+             status: response.status
+    end
+
+    def query_params
+      params.permit(:query)
+    end
+
+    def query
+      sanitize query_params['query']
+    end
```
johnpaulashenfelter: "query" and "params" gets used a lot here -- seems like it might be clearer with something like

`Forms::Client.new(search_term)` -- assuming its a search term. query is a pretty generic word

app/controllers/v0/forms_controller.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module V0
+  class FormsController < ApplicationController
+    include ActionView::Helpers::SanitizeHelper
+
+    skip_before_action :authenticate
+
+    def index
+      response = Forms::Client.new(query).get_all
```
johnpaulashenfelter: this would typically be `query_params` or just `params`. You're bogged down a bit by your two helper methods

app/controllers/v0/forms_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+      render json: response.body,
+             status: response.status
+    end
+
+    def query_params
```
johnpaulashenfelter: these are `private` methods, right?

app/controllers/v0/forms_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+      params.permit(:query)
+    end
+
+    def query
+      sanitize query_params['query']
```
johnpaulashenfelter: I *think* you might just want this to be a one-liner if you can -- i think the multiple methods with the naming here are more confusing rather than lesss

kfrz:
@ncksllvn this looks great! I **would** like to see swagger docs added if this is getting committed to `master` and used, but I won't block approval. As for getting a key into credstash, check with operations they should have a process defined.

app/controllers/v0/forms_controller.rb
johnpaulashenfelter:
```
def search_params
	query = params.permit(:query)
	sanitize query
end
```
or a one-liner.

config/routes.rb
```diff
@@ -234,6 +234,9 @@
 
     resources :search, only: :index
 
+    get 'forms', to: 'forms#index'
+    get 'forms/healthcheck', to: 'forms#healthcheck'
```
johnpaulashenfelter: Does this exist?

lib/forms/client.rb
```diff
@@ -0,0 +1,65 @@
...
+        raw_response = perform(:get, 'forms', query_params)
+        Forms::Responses::Response.new(raw_response.status, raw_response.body, 'forms')
+      end
+    rescue => e
+      handle_error(e)
+    end
+
+    private
+
+    def query_params
```
johnpaulashenfelter: Heh! Here's its query_params 👍

lib/forms/responses/response.rb
```diff
@@ -0,0 +1,34 @@
...
+      attribute :status, Integer
+
+      def initialize(status, body, schema_name)
+        self.body = body if json_format_is_valid?(body, schema_name)
+        self.status = status
+      end
+
+      private
+
+      def json_format_is_valid?(body, schema_name)
```
johnpaulashenfelter: 👍 validating from the json schema

lib/forms/responses/response.rb
```diff
@@ -0,0 +1,34 @@
...
+        schema_path = Rails.root.join('lib', 'forms', 'schemas', "#{schema_name}.json").to_s
```
johnpaulashenfelter: Would this ever be anything but `forms.json` for you?

spec/controllers/v0/forms_controller_spec.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe V0::FormsController, type: :controller do
+  context 'with query param' do
+    it 'returns forms' do
+      VCR.use_cassette('forms/200_form_query') do
+        get :index, params: { query: 'health' }
+        expect(JSON.parse(response.body)['data'][0]['type']).to eq('va_form')
```
johnpaulashenfelter: You _could_ refactor the common bit of these two out

```
....
	expect type_of_response(response) to eq('va_form')
....
def type_of_response(response)
	JSON.parse(response.body)['data'][0]['type'])	
end
```

I think that then could be refactored out even further but that's a different issue 😄

spec/lib/forms/client_spec.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe Forms::Client do
+  subject { described_class.new(query) }
+
+  let(:query) { nil }
```
johnpaulashenfelter: You don't need to define it (:query) here since its in each context, which is a better idea in any case

lib/forms/responses/response.rb
ncksllvn: Not right now, but potentially. That health-check route may come into play later but I think this PR is large enough already.

ncksllvn:
@kfrz I added a ticket for the Swagger part - https://github.com/department-of-veterans-affairs/va.gov-team/issues/4281. I figured this PR was big enough already.

PR #3619, opened 2019-12-06
VAOS systems pact endpoint

kreek:
## Description of change
Adds the `/v0/vaos/systems/:system_id/pact` endpoint. Ticket # [3991](https://github.com/department-of-veterans-affairs/va.gov-team/issues/3991)

## Testing done
unit tests

## Testing planned
staging integration tests

#### Unique to this PR
- [ ] Adds the `/v0/vaos/systems/:system_id/pact` endpoint
#### Applies to all PRs

- [ ] Appropriate logging


saneshark:
@annaswims @johnpaulashenfelter I think we might want to relax the code-climate "code similarity" thresholds a bit.

Screen Shot 2019-12-06 at 2 25 07 PM

are NOT "similar" enough in my opinion. 😂

modules/vaos/app/services/vaos/systems_service.rb
```diff
@@ -92,6 +92,15 @@ def get_facility_available_appointments(facility_id, start_date, end_date, clini
       end
     end
 
+    def get_system_pact(system_id)
+      with_monitoring do
+        url = "/var/VeteranAppointmentRequestService/v4/rest/direct-scheduling/site/#{system_id}" \
+                "/patient/ICN/#{@user.icn}/pact-team"
+        response = perform(:get, url, nil, headers(@user))
```
saneshark: Totally a nitpick, but I think elsewhere user is attr_reader, and therefore we call the method `user` instead of instance variable.

modules/vaos/app/controllers/vaos/pact_controller.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module VAOS
+  class PactController < ApplicationController
+    def index
+      binding.pry
```
LindseySaari: Looks like a leftover debugging line 🐛

saneshark: I think this has been resolved, but I'm not sure why it's still appearing in the diff.

saneshark:
If you fetch the above branch, there is no binding.pry on that line. I am not sure why it appears in the diff as though it is there. Is this something code climate is messing up?

saneshark:
nevermind, seems to be working now -- maybe github is having caching issues.

PR #3623, opened 2019-12-10
limit retries in pager_duty/poll_maintenance_windows

annaswims:
## Description of change
Limit retries to 3-ish minutes since the job runs every three minutes.  

Sidekiq will retry failures with an exponential backoff using [the formula](https://github.com/mperham/sidekiq/wiki/Error-Handling#automatic-job-retry).  3 retries is roughly 2.5 miniutes. 

* rename the files to be consistent with the class name

### DIFF
--- a/app/workers/pager_duty/poll_maintenance_windows.rb
+++ b/app/workers/pager_duty/poll_maintenance_windows.rb
@@ -6,7 +6,7 @@ module PagerDuty
   class PollMaintenanceWindows
     include Sidekiq::Worker
     include SentryLogging
-    sidekiq_options retry: 3
+    sidekiq_options retry: 1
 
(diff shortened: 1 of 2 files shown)


app/workers/pager_duty/poll_maintenance_windows.rb
```diff
@@ -6,6 +6,7 @@ module PagerDuty
   class PollMaintenanceWindows
     include Sidekiq::Worker
     include SentryLogging
+    sidekiq_options retry: 3
```
johnpaulashenfelter: How about just one retry @annaswims (or none). If this runs every 3m, then we'll have 4 failures every 3m when reachability is not ephemeral. I get 1 retry for ephemeral stuff. 

Is the 3m scheduled frequency good enough?

PR #3631, opened 2019-12-11
Add post appt req message

saneshark:
## Description of change
Fixes: https://github.com/department-of-veterans-affairs/va.gov-team/issues/2977

Introduced POST message, as a new endpoint. This change includes a form object for validation.

Since the response returns a single message instead of an array of messages, we made some minor changes to the schema. We also added 3 new VCR cassettes. Some of the other changes in service and request specs might appear to show modification to existing GET requests, those are a false positive in the diff as a result of embedding that code in a new `describe` block.

## Testing done
MessageForm specs, MessageControllerSpec, MessageServiceSpec, and comprehensive manual testing against the service on staging.

## Testing planned
Integration testing with the FE on actual staging.

Verify on staging by FE

#### Applies to all PRs

- [X] Appropriate logging


modules/vaos/app/models/vaos/message_form.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    def message_date_time
+      ''
+    end
+
+    def message_sent
+      true
+    end
+
+    # rubocop:disable Naming/PredicateName
```
johnpaulashenfelter: You've got control of this variable name, so I'd recommend following the pattern and using `last_message`

config/locales/exceptions.en.yml
```diff
@@ -486,6 +486,12 @@ en:
         code: 'VAOS_400'
         detail: # deliberarly left nil to reference response.body which is unparsable string (not json)
         status: 400
+      VAOS_204_hack:
```
johnpaulashenfelter: "hack"? No real context for why that's named like that. There's no other VAOS_204 so maybe some more explanation

modules/vaos/app/models/vaos/message_form.rb
```diff
@@ -0,0 +1,59 @@
...
+    attribute :is_last_message, Boolean
+    attribute :_appointment_request_id, String
+
+    def initialize(user, request_id, json_hash = {})
+      super(json_hash)
+      @user = user
+      @appointment_request_id = request_id
+    end
+
+    # HACK: addresses the fact that downstream requires this as AppointmentRequestId (upper camelcase)
```
johnpaulashenfelter: both `appointment_require_id` and `_appointment_request_id` become `AppointmentRequestId` when using `.camelize` -- this comment really doesn't make it clear where or why you need the extra variable

modules/vaos/app/services/vaos/messages_service.rb
```diff
@@ -31,10 +31,31 @@ def get_messages(request_id)
...
+        else
+          handle_error(response)
+        end
+      end
+    end
+
     private
 
+    def handle_error(response)
+      key = response.status == 204 ? 'VAOS_204_hack' : nil
```
johnpaulashenfelter: Still trying to figure this out -- you're using VAOS_204_hack to indicate a 400 error? seems like it would be easier to understand to mask it right at the source (here, where the error is handled) and make it clear what code is being returned and why.

modules/vaos/app/models/vaos/message_form.rb
saneshark: we do not have control of this name, need this to be sent to VAMF. In fact, if you look more closely we're overriding it and setting it manually such that it cannot be assigned by FE. This is a required field by VAMF to validate this form object.

config/locales/exceptions.en.yml
saneshark: I think we can probably just call it VAOS_204. 

What's happening here isn't even really a hack, I can fix this. Basically the host is returning a 204 which would otherwise be interpreted as a success. What they probably should be returning is a 400 or 404 because the parent association of the message is invalid or not found.

modules/vaos/app/models/vaos/message_form.rb
saneshark: NOT TRUE. Camelize as it's written in the middleware does lowerCamelCase. All other attributes in the request are in fact required to be lowerCamelCase, but this particular attribute for whatever reason must be UpperCamelCase.

modules/vaos/app/services/vaos/messages_service.rb
saneshark: I know precisely what is being returned, but the http status being returned (204) is not helpful seeing that the request actually did nothing at all. We need to interpret this SUCCESS as a FAILURE. Whether that means a 400 Bad Request or 404 Not Found. Technically it's the parent object (the appointment request) that was not found, and this was identified when sending appointment_request_id as lowerCamelCase instead of the upperCamelCase they require.

it's unlikely we will ever run into this issue, but I felt it was important to capture it in the event that there are other ways that this issue could manifest itself.

saneshark:
Everywhere that you see references to "hack" there is something that VAMF has an opportunity to fix. It's unlikely that those things are going to get fixed any time soon though. But at least we're identifying the opportunities to improve those things by calling attention to them. 

I personally feel like the code is the best place for this to live, not issues that end up in an icebox and fall by the waist-side as has happened in the past.

saneshark:
This is ready for second review as I believe all the changes have now been addressed or commented on as a fix is not necessary.

config/locales/exceptions.en.yml
kfrz: Yeah the pattern here can be followed - `VAOS_204` is fine imo. There are other examples of this

modules/vaos/app/controllers/vaos/messages_controller.rb
```diff
@@ -6,8 +6,22 @@ def index
...
     end
 
+    def create
+      response = messages_service.post_message(appointment_request_id, post_params)
+      render json: MessagesSerializer.new(response[:data], meta: response[:meta])
+    end
+
     private
 
+    def appointment_request_id
```
annaswims: I don't care too much about this, but  it feels a bit strange to declare a method to reference `params[:appointment_request_id]` if you're just going to reference it directly on line 26.

modules/vaos/app/models/vaos/message_form.rb
```diff
@@ -0,0 +1,59 @@
...
+      true
+    end
+
+    # rubocop:disable Naming/PredicateName
+    def is_last_message
+      true
+    end
+    # rubocop:enable Naming/PredicateName
+
+    # validates :appointment_request_id, presence: true
```
annaswims: commented code.

modules/vaos/app/controllers/vaos/messages_controller.rb
```diff
@@ -6,8 +6,22 @@ def index
...
+      params[:appointment_request_id]
+    end
+
+    def post_params
+      params.require(:message_text)
```
kfrz: You could do:

`params.require(:message_text).permit(:message_text)`

modules/vaos/app/models/vaos/message_form.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+require 'active_model'
+require 'common/models/form'
+
+module VAOS
+  class MessageForm < Common::Form
+    attribute :message_text, String # only this attribute gets set by user, everything else is overridden
```
kfrz: Who/what overrides this?

modules/vaos/app/models/vaos/message_form.rb
kfrz: Can we add a test to definitively show this?

modules/vaos/app/models/vaos/message_form.rb
kfrz: I don't love this but tbh it's not VSP code - a solution would have to change the name when the response is sent anyways.

modules/vaos/app/models/vaos/message_form.rb
annaswims: Maybe just take of the word hack? To me, "hack" implies that it could/should be done better and that it should be looked at later.

modules/vaos/app/services/vaos/messages_service.rb
```diff
@@ -31,10 +31,31 @@ def get_messages(request_id)
...
     end
 
+    def post_message(request_id, request_object_body)
+      with_monitoring do
+        params = VAOS::MessageForm.new(user, request_id, request_object_body).params
+        response = perform(:post, messages_url(request_id), params, headers)
+
+        if response.status == 200
+          {
+            data: OpenStruct.new(response.body),
```
kfrz: Any need for a full response object here?

modules/vaos/app/controllers/vaos/messages_controller.rb
saneshark: Will use the defined method. Good catch!

modules/vaos/app/models/vaos/message_form.rb
saneshark: removed

saneshark:
I agree. I will change HACK to "DOWNSTREAM_FIX"

modules/vaos/spec/request/messages_request_spec.rb
```diff
@@ -4,45 +4,145 @@
...
   before do
     Flipper.enable('va_online_scheduling')
     sign_in_as(current_user)
     allow_any_instance_of(VAOS::UserService).to receive(:session).and_return('stubbed_token')
   end
 
-  context 'loa1 user with flipper enabled' do
-    let(:current_user) { build(:user, :loa1) }
+  describe 'GET messages' do
+    let(:request_id) { '8a4886886e4c8e22016e5bee49c30007' }
```
kfrz: seems like a mystery number

modules/vaos/spec/request/messages_request_spec.rb
```diff
@@ -4,45 +4,145 @@
...
-        .to eq('You do not have access to online scheduling')
+      it 'does not have access' do
+        get "/v0/vaos/appointment_requests/#{request_id}/messages"
+        expect(response).to have_http_status(:forbidden)
+        expect(JSON.parse(response.body)['errors'].first['detail'])
+          .to eq('You do not have access to online scheduling')
+      end
+    end
+
+    context 'loa3 user' do
```
kfrz: nitpick but contexts read better imo when they start with "when" or "with"

modules/vaos/app/models/vaos/message_form.rb
saneshark: OK, that totally makes sense to me. I've changed it to:

```ruby
# DOWNSTREAM_BUG: addresses the fact that downstream requires this as AppointmentRequestId (upper camelcase)
```

modules/vaos/app/services/vaos/messages_service.rb
saneshark: no need for a full response object, but I think in future refactor I'll change to that. I think the validations and stuff should happen at the controller level. It's a poor abstraction but one based on precedent in other services all over the platform where the service is responsible for normalizing the data both in and out. It's tricky, but a separation of concerns seems appropriate.

saneshark: > Can we add a test to definitively show this?

I wish I had recorded the cassette for that. Would have to go back and re-record it or mock up a fake one. It's fairly obvious as a bug when looking at the VCR cassettes though. And when VAMF makes the change we will likely have to fix as part of a v2, so i'm ok with no spec for now if you are.

modules/vaos/spec/request/messages_request_spec.rb
saneshark: i'll make a note to change this in all our specs as part of a refactor PR I have planned.

modules/vaos/spec/request/messages_request_spec.rb
saneshark: it's the only request_id that would match what the VCR cassette has? The idea being that we record those and then have to play them back against the VCR cassette -- if they don't match the request would throw an error. Not sure how to address this. 🤔

modules/vaos/app/models/vaos/message_form.rb
saneshark: So the form object receives input from our FE, this field is validated. But there are other attributes that need to get passed to VAMF for the request to validate on their side. Those other attributes are overridden such that the FE cannot mass-assign them but do get included in the form object as part of the outbound request payload.

modules/vaos/app/controllers/vaos/messages_controller.rb
saneshark: i know, but sometimes you have an array of .requires keys and thats where things get tripped up. I haven't seen a clearly defined convention for this across the code. I personally think it's a poor design choice by Rails maintainers to do this and they ought to favor this happening in form objects, but convention over configuration even at the expense of separation of concerns is the rails way!

modules/vaos/spec/request/messages_request_spec.rb
jholton: You can inject variable values into the VCR cassette at runtime.  [Here is a recent PR](https://github.com/department-of-veterans-affairs/vets-api/pull/3592/files#diff-cda0492bae902639014fa5d79a611bd2) with a very similar use case to yours where the author had to do this to get his specs/cassettes to work.

PR #3632, opened 2019-12-12
2032 implement add person mvi

SMLuthi:
## Description of change

Integration for the  `add person` SOAP action with the MVI API. This will allow veterans who are missing their corp id(participant_id)/birls id (or if they are just missing a corp id(participant_id)) to be added to the corp database and have these ids generated for them. This will then allow them to start a compensation claim (since these ids are required). Also, since an `add_person` call requires a search token from a previous `find_profile` call in the request payload, changes have been made to store this variable in the mvi profile.

The capability for this has been added to the mvi.rb model. Its important to note that we have to do, what MVI calls, an orchestrated search before attempting to run proxy add. All this means is that we have to include the `representedOrganization` block within the find_profile call. This, according to them, will identify us to BGS and prepare them for a possible incoming add_person call. Once the proxy add is completed, the mvi profile can be replenished and it will include the newly created keys.

## Testing done

Testing being done by hand via a dev environment, unit tests.

## Testing planned

end-to-end tests within staging.

- [x] Add new SOAP endpoint call to MVI service class
- [x] Add Response parsing and error handling
- [x] Add Request XML payload builder
- [x] Store `search_token` from `find_profile` call
- [x] add mvi add person method to mvi model

#### Applies to all PRs
- [x] Appropriate logging


kfrz:
@SMLuthi can you link this to a ticket for reference?

lib/mvi/messages/add_person_message.rb
```diff
@@ -0,0 +1,51 @@
...
+        )
+
+        template.render!(build_content(@user))
+      end
+
+      private
+
+      def build_content(user)
+        current_time = Time.now.utc
+        ip_address = Socket.ip_address_list.find { |ai| ai.ipv4? && !ai.ipv4_loopback? }.ip_address
```
kfrz: Why is the accumulator named `ai`, not `ip`?

lib/mvi/messages/add_person_message.rb
```diff
@@ -0,0 +1,51 @@
...
+          File.read(File.join('config', 'mvi_schema', SCHEMA_FILE_NAME))
+        )
+
+        template.render!(build_content(@user))
+      end
+
+      private
+
+      def build_content(user)
+        current_time = Time.now.utc
```
kfrz: `Time.current` should return the time in UTC or w/e Rails is set to, rather than the server time.

lib/mvi/responses/add_parser.rb
```diff
@@ -0,0 +1,69 @@
...
+module MVI
+  module Responses
+    # Parses a MVI response and returns a MviProfile
+    class AddParser
+      include SentryLogging
+
+      BODY_XPATH = 'env:Envelope/env:Body/idm:PRPA_IN201306UV02'
+      CODE_XPATH = 'acknowledgement/typeCode/@code'
+
+      ACKNOWLEDGEMENT_DETAIL_CODE_XPATH = 'acknowledgement/acknowledgementDetail/code'
```
kfrz: Nitpick, please alphabetize these and remove excess whitespace

lib/mvi/service.rb
```diff
@@ -33,6 +33,34 @@ class Service < Common::Client::Base
...
+      log_console_and_sentry("MVI add_person connection failed: #{e.message}", :warn)
+      mvi_add_exception_response_for('MVI_504', e)
+    rescue Common::Client::Errors::ClientError, Common::Exceptions::GatewayTimeout => e
+      log_console_and_sentry("MVI add_person error: #{e.message}", :warn)
+      mvi_add_exception_response_for('MVI_504', e)
+    rescue MVI::Errors::Base => e
+      mvi_error_handler(user_identity, e)
+      mvi_add_exception_response_for('MVI_502', e)
+    end
+    # rubocop:enable Metrics/MethodLength
```
kfrz: Looks like a good place to extract this error handling chain to a private method (or middleware if you're so inclined)

lib/mvi/messages/add_person_message.rb
```diff
@@ -0,0 +1,51 @@
...
+        )
+
+        template.render!(build_content(@user))
+      end
+
+      private
+
+      def build_content(user)
+        current_time = Time.current
+        ip_address = Socket.ip_address_list.find { |ip| ip.ipv4? && !ip.ipv4_loopback? }.ip_address
```
patrickvinograd: Curious about the intent here. Obviously our vets-api instances get recycled all the time so in terms of auditability this doesn't help that much. Though not surprising that MVI expects more static IPs

SMLuthi: Yea, its literally a field thats required for their endpoint. TBH, we could probably fill this in with whatever but I'm not trying to give them a reason to reject us if its just filled with a garbage ip.

config/mvi_schema/mvi_add_person_template.xml
```diff
@@ -0,0 +1,93 @@
...
+                  </asOtherIDs>
+                  <asOtherIDs classCode="PAT">
+                    <id extension="PROXY_ADD^PI^200VBA^USVBA" root="2.16.840.1.113883.4.349"/>
+                    <scopingOrganization determinerCode="INSTANCE" classCode="ORG">
+                      <id extension="VBA" root="2.16.840.1.113883.4.349"/>
+                      <name>MVI.ORCHESTRATION</name>
+                    </scopingOrganization>
+                  </asOtherIDs>
+                </patientPerson>
+                <providerOrganization determinerCode="INSTANCE" classCode="ORG">
```
patrickvinograd: This and the `custodian` element look like placeholder/sample values that we would want to populate correctly; are we getting any guidance about what values we are supposed to supply?

SMLuthi: That IS the guidance we were given. In fact, I checked with ebenefits just to see if they do anything differently in this regard and they fill it exactly the same way.

Dockerfile
```diff
@@ -52,7 +52,7 @@ RUN curl -sSL -o /usr/local/bin/cc-test-reporter https://codeclimate.com/downloa
     cc-test-reporter --version
 COPY --chown=vets-api:vets-api config/freshclam.conf docker-entrypoint.sh ./
 USER vets-api
-# XXX: this is tacky 
+# XXX: this is tacky
```
kfrz: :face_with_head_bandage:

lib/mvi/messages/add_person_message.rb
SarahJaine: Silvio, I think your instinct to err on the side of caution and give them the real IP is correct. They may know the CIDR blocks for our VPCs; it's possible that they check that the stated ip is within the CIDR.

PR #279, opened 2016-10-17
Added doc on sorting multiple fields

mphprogrammer:

PR #265, opened 2016-10-14
Mhv session caching (RX and SM)

saneshark:

PR #263, opened 2016-10-14
Drafts redux

mphprogrammer:
* Removed sorting of folders
* Updated documentation
* added post_create_reply_draft
* added create_reply_draft and update_reply_draft

PR #20, opened 2016-08-17
upgrade rspec

lihanli:
upgrade rspec to fix this deprecation
```
[DEPRECATION] `last_comment` is deprecated.  Please use `last_description` instead.
```

PR #3635, opened 2019-12-12
upgrade pg

annaswims:
## Description of change
upgrade pg gem.

I read the changelog and I don't believe any additional changes are necessary.

There was a time that pg > 1.0 was incompatible with rails but that has been [fixed](https://github.com/rails/rails/pull/31671/files) as of 5.1.15
## Testing done
So many tests depend on the`pg` gem that if specs pass I'm confident in the upgrade.


annaswims:
Changelog:

== v1.1.4 [2019-01-08] Michael Granger <ged@FaerieMUD.org>

- Fix PG::BinaryDecoder::Timestamp on 32 bit systems. # 284
- Add new error-codes of PostgreSQL-11.
- Add ruby-2.6 support for Windows fat binary gems and remove ruby-2.0 and 2.1.


== v1.1.3 [2018-09-06] Michael Granger <ged@FaerieMUD.org>

- Revert opimization that was sometimes causing EBADF in rb_wait_for_single_fd().


== v1.1.2 [2018-08-28] Michael Granger <ged@FaerieMUD.org>

- Don't generate aliases for JOHAB encoding.
  This avoids linking to deprecated/private function rb_enc(db)_alias().


== v1.1.1 [2018-08-27] Michael Granger <ged@FaerieMUD.org>

- Reduce deprecation warnings to only one message per deprecation.


== v1.1.0 [2018-08-24] Michael Granger <ged@FaerieMUD.org>

Deprecated (disable warnings per PG_SKIP_DEPRECATION_WARNING=1):
- Forwarding conn.exec to conn.exec_params is deprecated.
- Forwarding conn.exec_params to conn.exec is deprecated.
- Forwarding conn.async_exec to conn.async_exec_params.
- Forwarding conn.send_query to conn.send_query_params is deprecated.
- Forwarding conn.async_exec_params to conn.async_exec is deprecated.

PG::Connection enhancements:
- Provide PG::Connection#sync_* and PG::Connection#async_* query methods for explicit calling syncronous or asynchronous libpq API.
- Make PG::Connection#exec and siblings switchable between sync and async API per PG::Connection.async_api= and change the default to async flavors.
- Add async flavors of exec_params, prepare, exec_prepared, describe_prepared and describe_portal.
  They are identical to their syncronous counterpart, but make use of PostgreSQL's async API.
- Replace `rb_thread_fd_select()` by faster `rb_wait_for_single_fd()` in `conn.block` and `conn.async_exec` .
- Add PG::Connection#discard_results .
- Raise an ArgumentError for strings containing zero bytes by #escape, #escape_literal, #escape_identifier, #quote_ident and PG::TextEncoder::Identifier. These methods previously truncated strings.

Result retrieval enhancements:
- Add PG::Result#tuple_values to retrieve all field values of a row as array.
- Add PG::Tuple, PG::Result#tuple and PG::Result#stream_each_tuple .
  PG::Tuple offers a way to lazy cast result values.
- Estimate PG::Result size allocated by libpq and notify the garbage collector about it when running on Ruby-2.4 or newer.
- Make the estimated PG::Result size available to ObjectSpace.memsize_of(result) .

Type cast enhancements:
- Replace Ruby code by a faster C implementation of the SimpleDecoder's timestamp decode functions. github #20
- Interpret years with up to 7 digists and BC dates by timestamp decoder.
- Add text timestamp decoders for UTC vs. local timezone variations.
- Add text timestamp encoders for UTC timezone.
- Add decoders for binary timestamps: PG::BinaryDecoder::Timestamp and variations.
- Add PG::Coder#flags accessor to allow modifications of de- respectively encoder behaviour.
- Add a flag to raise TypeError for invalid input values to PG::TextDecoder::Array .
- Add a text decoder for inet/cidr written in C.
- Add a numeric decoder written in C.
- Ensure input text is zero terminated for text format in PG::Coder#decode .

Source code enhancements:
- Fix headers and permission bits of various repository files.

Bugfixes:
- Properly decode array with prepended dimensions. #272
  For now dimension decorations are ignored, but a correct Array is returned.
- Array-Decoder: Avoid leaking memory when an Exception is raised while parsing. Fixes #279


== v1.0.0 [2018-01-10] Michael Granger <ged@FaerieMUD.org>

Deprecated:
- Deprecate Ruby older than 2.2.
- Deprecate Connection#socket in favor of #socket_io.

Removed:
- Remove compatibility code for Ruby < 2.0 and PostgreSQL < 9.2.
- Remove partial compatibility with Rubinius.
- Remove top-level constants PGconn, PGresult, and PGError.

Enhancements:
- Update error codes to PostgreSQL-10
- Update Windows binary gems to Ruby-2.5, PostgreSQL 10.1 and
  OpenSSL 1.1.0g.

Bugfixes:
- Fix URI detection for connection strings. #265 (thanks to jjoos)
- MINGW: Workaround segfault due to GCC linker error in conjunction with MSVC.
  This happens when linking to PostgreSQL-10.0-x64 from EnterpriseDB.

Documentation fixes:
- Add PostgreSQL version since when the given function is supported. #263
- Better documentation to `encoder` and `decoder` arguments of COPY related methods.


== v0.21.0 [2017-06-12] Michael Granger <ged@FaerieMUD.org>

Enhancements:
- Move add_dll_directory to the Runtime namespace for newest versions
  of RubyInstaller.
- Deprecate PGconn, PGresult, and PGError top-level constants; a warning
  will be output the first time one of them is used. They will be
  removed in the upcoming 1.0 release.

Documentation fixes:
- Update the docs for PG::Result#cmd_tuples

New Samples:
- Add an example of the nicer #copy_data way of doing `COPY`.


== v0.20.0 [2017-03-10] Michael Granger <ged@FaerieMUD.org>

Enhancements:
- Update error codes to PostgreSQL-9.6
- Update Windows binary gems to Ruby-2.4, PostgreSQL 9.6.1 and
  OpenSSL 1.0.2j.
- Add support for RubyInstaller2 to Windows binary gems.

Bugfixes:
- Use secure JSON methods for JSON (de)serialisation. #248
- Fix Result#inspect on a cleared result.
- Fix test case that failed on Ruby-2.4. #255

Documentation fixes:
- Talk about Integer instead of Fixnum.
- Fix method signature of Coder#encode.

jholton:
> So many tests depend on the`pg` gem that if specs pass I'm confident in the upgrade.

It's probably fine, but it looks like our CI is using Postgres 9.6 (via docker-compose) while production is on Postgres 9.4.20.
What are you running locally?

annaswims:
ooh, good catch @jholton I didn't realize pg was 9.4 in prod.  I'll hold off until it's tested with 9.4

johnpaulashenfelter:
@annaswims  there’s one direct PG call in the code in a controller `app/controllers/v0/disability_compensation_forms_controller.rb` — while that's a problematic bit of code (seems like it should be a validation) my only concern would be to make sure that's still in the error codes. Seems hard to believe NotNullViolation would change :)

Extremely minor risk -- its rescuing and reraising so worst case you'd lose the log message

annaswims:
Verified [NotNullViolation](https://github.com/ged/ruby-pg/blob/685c7db6388171161e5c172acf667da8d5da33cf/ext/errorcodes.def#L375) is still present.  

will test pg on stage

PR #3639, opened 2019-12-13
VAOS - update preferences

saneshark:
## Description of change
Updates the existing GET requests for Preferences as well to be more consistent, including making the route a singular resource.
Introduces request and service specs for updating preferences.

Fixes: https://github.com/department-of-veterans-affairs/va.gov-team/issues/3865

## Testing done
VCR cassettes recorded from actual staging data and being played back. 

## Testing planned
FE to verify on staging

FE verification

#### Applies to all PRs

- [X] Appropriate logging


modules/vaos/app/controllers/vaos/preferences_controller.rb
```diff
@@ -2,15 +2,29 @@
...
 
     private
 
     def preferences_service
-      VAOS::PreferencesService.new
+      VAOS::PreferencesService.new(current_user)
+    end
+
+    def put_params
+      params.require(:notification_frequency)
```
sjbrush: Is notification frequency the only required param?

saneshark: Yes. Good question. I did a little bit of reverse engineering / testing of the endpoint on staging in the console and it looks that way. If the FE identifies other validations and required fields they can always let us know and we'll add those.

modules/vaos/spec/request/preferences_request_spec.rb
```diff
@@ -36,5 +36,28 @@
...
+      it 'returns a 200 with correct schema' do
+        VCR.use_cassette('vaos/preferences/put_preferences', match_requests_on: %i[method uri]) do
+          put '/v0/vaos/preferences', params: request_body
+
+          expect(response).to have_http_status(:ok)
+          expect(response.body).to be_a(String)
+          expect(response).to match_response_schema('vaos/preferences')
+        end
+      end
+    end
```
kfrz: Looks like a happy path test, but can we test the sad-paths?

modules/vaos/spec/services/preferences_service_spec.rb
```diff
@@ -20,4 +22,36 @@
...
+        notification_frequency: 'Each new message',
+        email_allowed: true,
+        email_address: 'abraham.lincoln@va.gov',
+        text_msg_allowed: false,
+        text_msg_ph_number: ''
+      }
+    end
+
+    context 'with valid params' do
+      it 'updates preferences', :skip_mvi do
```
kfrz: Why are we skipping MVI? It's not obvious to someone who's not involved w/ VAOS details, could you add a succinct comment?

modules/vaos/app/controllers/vaos/preferences_controller.rb
kfrz: You can chain like this: `params.require(:params).permit(:other, :params)`

modules/vaos/app/controllers/vaos/preferences_controller.rb
```diff
@@ -2,15 +2,29 @@
...
     end
 
     private
 
     def preferences_service
-      VAOS::PreferencesService.new
+      VAOS::PreferencesService.new(current_user)
+    end
+
+    def put_params
```
kfrz: It's more rails-y to call this either `preferences_params`. `put_params` to me sounds like I'm trying to put some params somewhere.

modules/vaos/app/models/vaos/preference_form.rb
```diff
@@ -0,0 +1,46 @@
...
+      @user = user
+    end
+
+    def params
+      raise Common::Exceptions::ValidationErrors, self unless valid?
+
+      attributes.compact
+    end
+
+    # The following values are all overridden, needed by VAMF, but should not be passed in by FE.
```
kfrz: Where are they overridden?

modules/vaos/app/services/vaos/preferences_service.rb
```diff
@@ -11,12 +11,36 @@ class PreferencesService < Common::Client::Base
...
+    def initialize(user)
+      @user = user
+    end
+
+    def get_preferences
       with_monitoring do
-        url = "/var/VeteranAppointmentRequestService/v4/rest/patient/ICN/#{user.icn}/preference"
         response = perform(:get, url, nil, headers(user))
-        OpenStruct.new(response.body.merge(id: SecureRandom.uuid))
+        OpenStruct.new(response.body.merge(id: preference_id))
```
kfrz: Any chance this would return an error worth `rescue`-ing? Ditto for the `put_preferences` action?

modules/vaos/app/controllers/vaos/preferences_controller.rb
```diff
@@ -2,15 +2,29 @@
 
 module VAOS
   class PreferencesController < VAOS::BaseController
-    def index
-      response = preferences_service.get_preferences(current_user)
+    def show
+      response = preferences_service.get_preferences
+      render json: VAOS::PreferencesSerializer.new(response)
+    end
```
kfrz: This reads to me as if it is returning a collection of `Preference` resources, in which case a `show` seems inappropriate, why not an `index` still?

modules/vaos/config/routes.rb
```diff
@@ -20,7 +20,7 @@
       resources :limits, only: :index
       get 'visits/:schedule_type', to: 'visits#index'
     end
-    resources :preferences, only: :index
+    resource :preferences, only: %i[show update]
```
kfrz: This is confusing - maybe we need to consider a different name, if we're actually returning *one* resource that's a `PreferenceCollection` or something, otherwise the plurality doesn't make sense. Maybe I'm just not understanding the functionality on VAMF side correctly.

modules/vaos/spec/services/preferences_service_spec.rb
saneshark: We do this for all of our more recent specs, specifically the ones that use the vaos user factory. Without this, when the user is reloaded, the instance variable that we've set in the factory somehow gets reset and mvi and ends up being re-generated from a different factory. 

The whole stub_mvi convenience method is really rather confusing and ought to be refactored or removed in my opinion. What we're doing in vaos_users.rb factory is I think more efficient.

saneshark: I would have to add this comment in A LOT of places.

modules/vaos/app/controllers/vaos/preferences_controller.rb
saneshark: I kind of prefer to have to have these as distinct lines, than chaining. Particularly because it's a little inconsistent. For example if I pass an array of keys to require, I can no longer chain permit.

saneshark: happy to change the name, but what if your method ends up being different for Update vs Create? In other words, if I need to distinguish then preferences_params would become what: `preferences_params_for_update` and `preferences_params_for_create`?

modules/vaos/app/models/vaos/preference_form.rb
saneshark: Right here in the class beginning below that line. `attribute :some_attribute` creates a Getter and Setter method. We're overriding the getter method below this comment.

modules/vaos/app/services/vaos/preferences_service.rb
saneshark: no, it would raise an exception in the JSON middleware before it raises an exception here.

modules/vaos/app/controllers/vaos/preferences_controller.rb
saneshark: I would characterize Preferences as a singular resource having multiple attributes, not a collection. It's a 1-1 resource belonging to User though.

modules/vaos/config/routes.rb
saneshark: I don't think it's confusing at all. I mean rails allows you to have inflections as you deem prudent for a reason. Sometimes things just read better one way or the other. In this case your resource is a single resource, that in spoken language is usually pluralized. It's not a collection though -- it's a set of attributes that characterize preferences. A collection would imply that every single attribute should have it's own ID. Thats not the case at all.

modules/vaos/app/controllers/vaos/preferences_controller.rb
saneshark: Not a big fan of chaining these since the behavior is considerably different when you pass an array to require. If it were more consistent in what require returns, I might consider it.

modules/vaos/spec/request/preferences_request_spec.rb
saneshark: We have certain degree of confidence that our middleware are handling the really unhappy paths like 500 already, we certainly wouldn't write specs for those. And I couldn't get any of the attributes here to return errors. I'd have to sit and reverse engineer each one until I find something that returns an error. Would prefer to wait for FE to notify me that something's causing issues and addressing at that point than to spend extra time on reverse engineering unhappy paths.

modules/vaos/app/controllers/vaos/preferences_controller.rb
saneshark: If each attribute returned was part of an array, each attribute having its own ID, then I would agree with you on "collection." But that's not the case. It's a set of attributes (which in spoken language is typically pluralized as "preferences"). Rails makes exceptions for different inflections. But whats important to note is that a User can at most only have 1 set of this resource, hence why it is singular.

modules/vaos/spec/request/preferences_request_spec.rb
kfrz: I understand this approach - just wanted to make sure any *documented* paths or known errors are captured.

modules/vaos/config/routes.rb
saneshark: `resource` vs `resources` distinguishes whether it is a singular resource or collection. How you choose to inflect is more a matter of language you wish to use.

modules/vaos/spec/services/preferences_service_spec.rb
kfrz: Ok - thanks for the explanation. After looking through the code it's more clear, definitely ripe for refactor. 

Also `:skip_mvi` is indicative enough.

saneshark: I haven't had a chance to go in and re-record VCR cassettes for the specs that are still using :mhv user. Most of the ones from the last month have been using :vaos user. Go ahead and try it out int he console for yourself -- you'll notice that for unit level specs things work just fine. Then for integration level specs, for whatever odd reason, stub_mvi starts doing some really wonky and unexpected things.

PR #3640, opened 2019-12-13
VAOS - GET CC Eligibility

saneshark:
## Description of change
Fixes https://github.com/department-of-veterans-affairs/va.gov-team/issues/2973

Introduces mechanism to fetch community care eligibility based on service type.

## Testing done
Local

## Testing planned
Staging

#### Unique to this PR
1. Introduces CC Eligibility GET endpoint
2. Identified bug in limit and pact controller not requiring vaos application controller, which causes application controller to default to root level application controller which then doesn't do proper validations.
3. Added schemas
4. Added routing specs for custom route

#### Applies to all PRs

- [X] Appropriate logging


saneshark:
Please ignore code climate. It's not at all similar blocks of code.

config/initializers/inflections.rb
```diff
@@ -13,5 +13,6 @@
   inflect.acronym 'VAOS'
   inflect.acronym 'VBA'
   inflect.acronym 'VIC'
+  inflect.acronym 'CC'
```
johnpaulashenfelter: Thank you for inflecting!

modules/vaos/app/controllers/vaos/limits_controller.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_dependency 'vaos/application_controller'
```
johnpaulashenfelter: Would prefer if these were removed by doing something like VAOSBaseController or something similar.

saneshark: To be fair, I think most (if not all) of the modules are doing something similar and we drew heavily from those examples, but is something I've addressed in a PR you've since approved so TY. :)

saneshark:
This is ready for a second review now that the BaseController changes have been merged.

PR #3645, opened 2019-12-16
include source_app label in StatsD

omgitsbillryan:
## Description of change
Adds a `source_app` label to the StatsD metrics that vets-api serves to prometheus. This value comes from a header that `vets-website` passes on each request to vets-api ([link to that PR](https://github.com/department-of-veterans-affairs/vets-website/pull/11222))

## Testing done
rspec

## Testing planned
- [ ] Discover if nginx (revproxy) automatically passes on the header from the browser onto vets-api

#### Applies to all PRs

- [ ] Appropriate logging


lib/statsd_middleware.rb
```diff
@@ -3,6 +3,52 @@
...
+    'proxy-rewrite',
+    'public-outreach-materials',
+    'search',
+    'static-pages',
+    'terms-and-conditions',
+    'vaos',
+    'verify',
+    'veteran-id-card',
+    'veteran-representative',
+    'vic-v2'
```
omgitsbillryan: @cvalarida I got this list by looking in `vets-website` for `manifest.json` and finding the line:
```json
"entryName": "foo"
```

Do you know if this list is complete?

lib/statsd_middleware.rb
```diff
@@ -31,16 +77,24 @@ def call(env)
     if path_parameters
       controller = path_parameters[:controller]
       action = path_parameters[:action]
+      source_app = env['HTTP_SOURCE_APP_NAME']
```
omgitsbillryan:
```ruby
    # We should never use a dynamic path to apply the tag for the instrumentation,
    # since this will permit a rogue actor to increase the number of time series
    # exported from the process and causes instability in the metrics system.
```

@wyattwalter do you see a problem with this given the ominous comment ☝️ . We do perform a whitelist later on.

lib/statsd_middleware.rb
```diff
@@ -3,6 +3,52 @@
...
+    proxy-rewrite
+    public-outreach-materials
+    search
+    static-pages
+    terms-and-conditions
+    vaos
+    verify
+    veteran-id-card
+    veteran-representative
+    vic-v2
```
omgitsbillryan: @cvalarida I got this list by looking in vets-website for all manifest.json files and finding the line:
```
"entryName": "foo"
```
Do you know if this list is complete?

lib/statsd_middleware.rb
```diff
@@ -3,6 +3,52 @@
 class StatsdMiddleware
   STATUS_KEY   = 'api.rack.request'
   DURATION_KEY = 'api.rack.request.duration'
+  SOURCE_APP_NAMES = %w[
```
annaswims: how do we intend for FE engineers to know to keep this list up to date?  Can you add a comment, so people in the future know that this should be kept up to date manually and FE engineers

omgitsbillryan: Good question. Some thoughts and stuff:

- added logic to warn to sentry when an unrecognized `Source-App-Name` header comes in
- add instructional comment on how to regenerate 
- if we're out of date on the whitelist it isnt _that big of a deal_, it will result in some small holes in our metrics

lib/statsd_middleware.rb
cvalarida: That's as complete as I think we can get right now. You can verify this by running `yarn apps` or (`npm run apps`). That'll do the same thing.

lib/statsd_middleware.rb
omgitsbillryan: I suppose my real question is: does this warrant some changes in our prometheus storage config?

I came across [a prior postmortem](https://github.com/department-of-veterans-affairs/vets.gov-team/blob/master/Postmortems/2017-02-21-prometheus-metrics-count.md), where we created a new label that had an infinite number of possible values. In practice, I'm not sure where that breaking point is of how many possible values warrant config changes.

Is trying to calculate the storage impact upfront a practical solution? Or is it best to merge and then monitor?

lib/statsd_middleware.rb
```diff
@@ -31,16 +81,35 @@ def call(env)
...
+    return source_app if SOURCE_APP_NAMES.include?(source_app)
+
+    Raven.capture_message('Unrecognized Source App Request Header',
+                          level: 'warning', extra: { source_app_name: source_app })
+    nil
+  end
+
+  def instrument_statsd(status, duration, controller, action, source_app)
+    duration_tags = ["controller:#{controller}", "action:#{action}"]
+    duration_tags.push("source_app:#{source_app}") unless source_app.nil?
```
wyattwalter: We need to make sure that every metric sent to statsd with the same name has the same labels too. I'm not sure if `null` or an empty string is preferable to the library, but they should all have the same "shape" (i.e. labels).

lib/statsd_middleware.rb
```diff
@@ -31,16 +81,35 @@ def call(env)
...
+
+  private
+
+  def get_source_app(env)
+    source_app = env['HTTP_SOURCE_APP_NAME']
+
+    return nil if source_app.nil?
+    return source_app if SOURCE_APP_NAMES.include?(source_app)
+
+    Raven.capture_message('Unrecognized Source App Request Header',
```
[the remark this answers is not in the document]
omgitsbillryan: Yes it sends to sentry, and yes we want to know when this happens.

You're poking at the fact that a malicious user could load up our sidekiq queues? Since we use sidekiq enterprise, we could enforce some rate limiting / throttling on this call to Sentry.

Alternatively, we could use the Rails.logger and setup a metric log filter + alert.

I think I'll just remove it altogether for now and create a new ticket since knowing when this happens can happen a bit later.

lib/statsd_middleware.rb
omgitsbillryan: K. J/w - what happens if we don't?

lib/statsd_middleware.rb
omgitsbillryan: Created a ticket for this: https://github.com/department-of-veterans-affairs/va.gov-team/issues/4421

PR #2708, opened 2019-01-04
MVI integration for HCA dupes

lihanli:
## Description of change

https://github.com/department-of-veterans-affairs/vets.gov-team/issues/16032
add a method that looks up a person's ICN based on attributes from the HCA form, which we will later use to look up that person through the enrollment system api

## Testing done
automated tests, tested MVI in console on dev

#### Applies to all PRs

- [x] Appropriate logging
- [x] Swagger docs have been updated, if applicable
- [x] Provide link to originating GitHub issue, or connected to it via ZenHub
- [x] Does not contain any sensitive information (i.e. PII/credentials/internal URLs/etc., in logging, hardcoded, or in specs)
- [x] Provide which alerts would indicate a problem with this functionality (if applicable)

PR #3649, opened 2019-12-17
Pension automation work — Veteran MVI verification for automated claim creation

bshyong:
#### This PR is set up to merge into the feature branch `pa-claims-dev`

Within the 530 Burial Benefits form flow, check whether the Veteran on the form exists in MVI. This work supports upcoming pension automation work for directly creating a claim via API from 530 form data rather than routing through Central Mail.

Resolves https://github.com/department-of-veterans-affairs/pension-and-fiduciary-automation/issues/188

## Testing done
Local testing

## Testing planned
Staging/review instance

- [x] Appropriate logging


app/controllers/v0/burial_claims_controller.rb
```diff
@@ -2,6 +2,32 @@
...
+
+      else
+        # TODO: Move call to CentralMail here
+      end
+
+      claim.process_attachments! # upload claim and attachments to Central Mail
+
+      StatsD.increment("#{stats_key}.success") # TODO: potentially log creation of claims with Claims API; veteran/claimant lookup failure
+      Rails.logger.info "ClaimID=#{claim.confirmation_number} Form=#{claim.class::FORM}"
+      validate_session
```
annaswims: Just curious: What leads you to verify that a user is logged in towards the end of a controller action?

annaswims:
Are we allowed (through our ATO) to take data from unauthenticated users and query MVI?

bshyong:
I am checking on the ATO question

jholton:
FWIW, [the Health Care Application queries MVI based on user attributes from non-logged-in users](https://github.com/department-of-veterans-affairs/vets-api/blob/6f721a6881601edadae4130d990cb8552192dfc4/app/models/health_care_application.rb#L119).

You don't have to be logged in to fill out a Health Care Application.
They were getting a lot of duplicate submissions.
So this was implemented to take user attributes from the health care application form, query MVI for their ICN, then use that to check with HCA if the user already has a submission being processed.

jholton:
More context:  
[HCA PR](https://github.com/department-of-veterans-affairs/vets-api/pull/2708)
In that PR there is mention that "@gabrielzucker says we can amend the agreement with IAM".
I don't see much else detail about it though.

bshyong:
ATO question confirmed by Faith that we are covered. Under current ATO we are allowed to receive PII
https://dsva.slack.com/archives/CBU0KDSB1/p1576684045090600?thread_ts=1576615584.073000&cid=CBU0KDSB1

lib/bip_claims/service.rb
```diff
@@ -0,0 +1,82 @@
...
+        first_name: full_name['first'],
+        last_name: full_name['last'],
+        birth_date: bday
+      )
+    end
+
+    def lookup_veteran_from_mvi(claim)
+      MVI::AttrService.new.find_profile(veteran_attributes(claim))&.profile
+    rescue MVI::Errors::Base
+      nil
```
jjulian: Should we do anything interesting when this happens? Update a stat counter? Log a message?

lib/bip_claims/service.rb
```diff
@@ -0,0 +1,82 @@
...
+    end
+
+    def create_claim(form_data)
+      veteran_record = lookup_veteran_from_mvi(form_data)
+      claimant_record = false # TODO: look up from MVI
+      return false unless veteran_record && claimant_record
```
jjulian:
```suggestion
      return true if veteran_record && claimant_record
```
Just a thought to clarify double-negative "return false if not both" to "return true if both"

bshyong: makes sense; though in this context I'm trying to have the method return early with a false value, so that the submit_claim API call never occurs

lib/bip_claims/service.rb
bshyong: Yes; I am not sure what sort of stats we want to capture here; it would be worth it to discuss what metrics or logging we want here. Could make sense to logging into Sentry

lib/bip_claims/service.rb
jholton: Yeah, if I'm reading it right, he wants to
```ruby
submit_claim(veteran: veteran_record, claimant: claimant_record) if veteran_record && claimant_record
```
So for clarity the last few lines of the method could be
```ruby
return submit_claim(veteran: veteran_record, claimant: claimant_record) if veteran_record && claimant_record
false
```
or 
```ruby
if veteran_record && claimant_record
  submit_claim(veteran: veteran_record, claimant: claimant_record) 
else
  false
end
```
What's most clear?

lib/bip_claims/service.rb
jjulian: Lets tick statsd counters for "mvi lookup miss" and "mvi lookup hit"

lib/bip_claims/service.rb
jjulian: ah ok, ignore my suggestion please

lib/bip_claims/service.rb
bshyong: Added

bshyong:
Closing and merging into feature branch

PR #3650, opened 2019-12-17
exclude swagger from codeclimate similar code

annaswims:
## Description of change
Codeclimate complains a lot about similar code, but swagger is more like documentation than code, so exclude it from duplication checks.

I think that this change will only take effect once it's on master.

https://codeclimate.com/github/department-of-veterans-affairs/vets-api/issues?category%5B%5D=duplication&page=2&status%5B%5D=&status%5B%5D=open

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -28,7 +28,8 @@ checks:
     config:
       threshold: # language-specific defaults. an override will affect all languages.
     exclude_patterns:
-    - '*/engine.rb'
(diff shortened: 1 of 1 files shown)


annaswims:
Didn't work... 😢 https://codeclimate.com/github/department-of-veterans-affairs/vets-api/issues?category%5B%5D=duplication&status%5B%5D=&status%5B%5D=open&status%5B%5D=confirmed

PR #3664, opened 2019-12-18
MVI Find Profile by Attributes Spike

rtravitz:
## Description of change

This PR is a proposal for how to add a public method to query MVI by attributes without passing in a User object created from authenticating with an identity provider. This type of querying is needed by as part of [#3655](https://github.com/department-of-veterans-affairs/vets-contrib/issues/3655), which is [part of a larger epic](https://github.com/department-of-veterans-affairs/vets-contrib/issues/3633) to add an API to Lighthouse that indicates status based on passed in identity traits. 

### More details
The new Veteran Confirmation API will be secured by an API key and have a `/status` endpoint which accepts attributes like name, birth date, and social security number in a POST body. That information will then be used to query MVI to retrieve an ICN, which can then be used to query eMIS. A response of "confirmed" or "not confirmed" is then returned to the end user. 

The `MVI::Service` currently assumes that a user object will be provided. This use-case deviates from that assumption and requires the additional public method.

### Notable changes
- Adds a new public method called `find_profile_by_attributes` to `MVI::Service`
- Modifies error reporting logic to not assume that a user identity will be provided

## Testing done

- Replicates tests, including for the various MVI error cases from existing find profile implementation

#### Unique to this PR

- [x]  A way to query MVI by user traits is created

#### Applies to all PRs
- [x] Appropriate logging

### DIFF
--- a/lib/mvi/service.rb
+++ b/lib/mvi/service.rb
@@ -67,6 +67,35 @@ module MVI
         mvi_profile_exception_response_for('MVI_502', e)
       end
     end
+
+    def find_profile_from_attributes(attributes)
+      with_monitoring do
+        raw_response = perform(
+          :post, '',
+          create_profile_message_from_attributes(attributes),
+          soapaction: OPERATIONS[:find_profile]
+        )
+
+        MVI::Responses::FindProfileResponse.with_parsed_response(raw_response)
+      end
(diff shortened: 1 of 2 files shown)


rtravitz:
Closing in favor of leveraging the existing `MVI::AttrService` as seen in #3676

PR #3668, opened 2019-12-18
Log tokens with expired signatures

jasalisbury:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/3770


app/models/token.rb
```diff
@@ -20,6 +20,7 @@ def payload
                    JWT.decode(@token_string, pubkey, true, algorithm: 'RS256')[0]
                  end
   rescue JWT::ExpiredSignature => e
+    Rails.logger.info('JWT: EXPIRED SIGNATURE', token: @token_string)
```
johnpaulashenfelter: @jasalisbury I don't see any harm in the logging, but in the original issue, getting rid of the error altogether is part of the suggested solution.

Seems like this adds a log that will be a 2nd source of noise (in sentry through original error  as we do currently and now in log) without really changing the underlying issue. Was the goal to get a better count of expirations vs other errors through the log? Maybe changing the error message here to "Validation error: expired token (e.message)" would be clearer -- though maybe that's exactly what `e.message` says.

dvogel: @johnpaulashenfelter Since expired tokens are generally an expected condition, we would like to eventually get rid of these errors altogether. However the issue we're debugging involves a bunch of unexpectedly expired tokens. The logging approach here is to at least get a sample of the expired tokens to aide debugging. The original issue mentions converting this to a prometheus metric because that's what we _hope_ to do but cannot do yet until we identify why we're seeing so many expired tokens.

jasalisbury: > Seems like this adds a log that will be a 2nd source of noise (in sentry through original error as we do currently and now in log)

https://github.com/department-of-veterans-affairs/vets-api/pull/3612 excluded these errors from hitting Sentry. Can verify by looking at the error [in Sentry](http://sentry.vfs.va.gov/vets-gov/platform-api-production/issues/80772/).

johnpaulashenfelter: Awesome. Thanks @dvogel @jasalisbury

edmkitty: Yes, `e.message` is just forwarded the original error message, so in this case it would be "JWT Expired Signature." I recommend reusing the original error message instead of the hard-coded string. The raised error below is not reported to sentry since it is excluded. So if that `raise` is actually just logging, then I think this is duplicated logic.

PR #2972, opened 2019-04-17
Ruby 2.3.8

kreek:
## Description of change
Updates the .ruby-version rbenv file and docs to reference Ruby 2.3.8 to match the docker and deployed environments.

## Testing done
local specs

## Testing planned
N/A

#### Unique to this PR

- [ ] Native setup uses Ruby 2.3.8

#### Applies to all PRs

- [ ] Appropriate logging
- [ ] Swagger docs have been updated, if applicable
- [ ] Provide link to originating GitHub issue, or connected to it via ZenHub
- [ ] Does not contain any sensitive information (i.e. PII/credentials/internal URLs/etc., in logging, hardcoded, or in specs)
- [ ] Provide which alerts would indicate a problem with this functionality (if applicable)

PR #3675, opened 2019-12-19
VAOS #2972 get cc supported sites

sjbrush:
## Description of change
Added service call to get vaos sites that support community care.  The service call on the back end has a misleading name, but is the right one.
https://github.com/department-of-veterans-affairs/va.gov-team/issues/2972

#### Unique to this PR

#### Applies to all PRs

- [x] Appropriate logging


modules/vaos/app/serializers/vaos/cc_supported_sites_serializer.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require 'fast_jsonapi'
+
+module VAOS
+  class CCSupportedSitesSerializer
+    include FastJsonapi::ObjectSerializer
+
+    set_id :id
```
saneshark: So this is going to be interesting. Even though we get an array, it's not really a collection, right? A collection would imply that each member has an id. In fact, I'm not sure any members have an "id" so maybe, since your parameter that you're passing back essentially returning a constant, which could happens to be an array, we could consider setting the value of ID to that.

Lets record the VCR cassettes and we can revisit this. Maybe there is something in the payload returned that is better. But right now, this doesn't make any sense, since you're not likely to get an "id" in the payload returned to you.

modules/vaos/app/serializers/vaos/cc_supported_sites_serializer.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    set_type :objectType
```
saneshark: This is camel-case, should be snake_case

modules/vaos/app/services/vaos/cc_supported_sites_service.rb
```diff
@@ -0,0 +1,35 @@
...
+    def initialize(user)
+      @user = user
+    end
+
+
+    def get_supported_sites(site_codes)
+      with_monitoring do
+        response = perform(:get, url(site_codes), nil, headers(user))
+        OpenStruct.new(response.body)
+      end
```
saneshark: nice work -- this is exactly what you want.

modules/vaos/config/routes.rb
```diff
@@ -9,6 +9,7 @@
       resources :messages, only: %i[index create]
     end
     get 'community_care/eligibility/:service_type', to: 'cc_eligibility#show'
+    get 'community_care/supported_sites', to: 'cc_supported_sites#index'
```
saneshark: I'm still not sure you want this to be an index action -- we're not returning a collection of results. If we had a unique id for each member, it would be an index. Instead we're getting a constant back that just happens to be an array. But we can revisit this once cassettes are recorded.

modules/vaos/spec/services/cc_supported_sites_service_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      let(:site_codes) { '' }
+
+      it 'handles bad parameters appropriately' do
+        VCR.use_cassette('vaos/cc_supported_sites/get_supported_sites_error', match_requests_on: %i[method uri]) do
+          expect { subject.get_supported_sites(site_codes) }.to raise_error(
+            Common::Exceptions::BackendServiceException
+          )
+        end
+      end
+    end
```
saneshark: good job, we'll do something similar for request specs once cassettes have been recorded.

modules/vaos/config/routes.rb
saneshark: If site-code could be thought of as an "id" then we could build the route such that site_code is part of the url, not a query string. I think this would make more sense, but then you'd have a "show" (member action with site_code as id) not an "index" (which returns a collection).

modules/vaos/app/serializers/vaos/cc_supported_sites_serializer.rb
sjbrush: i expect to get an id (ie, '983') for each of the objects in the array.  Well, it's called 'id' and it's the site_id so i thought that would be appropriate to use here.

modules/vaos/config/routes.rb
sjbrush: site_code is a comma separated list of site ids, so not sure if that changes your mind one way or another.

modules/vaos/app/serializers/vaos/cc_supported_sites_serializer.rb
saneshark: I see, in that case you'll probably want to do something like:

set_id :site_id

modules/vaos/config/routes.rb
saneshark: hmmm, this doesn't seem particularly RESTful to me to be honest. I guess if site_code is thought of as an array of parameters, then perhaps index makes the most sense.

saneshark: cool, looks like we kept it as index -- resolved

saneshark:
LGTM! 🚀

modules/vaos/app/services/vaos/cc_supported_sites_service.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    private
+
+    def deserialize(json_hash)
+      json_hash[:sites_supporting_var].map do |request|
+        OpenStruct.new(request)
+      end
+    rescue => e
+      log_message_to_sentry(e.message, :warn, invalid_json: json_hash)
+      []
```
kreek: Only concern here is that we miss out on automated error monitoring/metrics by catching the error. Is it likely that we'd get a 200 response with malformed JSON? IMO if it's ok that CC data is blank the FE could handle the error gracefully while still allowing the backend to raise an error (like a 422)

saneshark: I think this was a cut and paste, and yeah -- i think I agree with you on this! Lets tackle this across all the services that are doing this as part of a refactor. And maybe we want to prioritize that to make sure we catch anything with it sooner rather than later.

kfrz:
@sjbrush please update the PR description to include the AC for this work, thanks! 

LGTM overall, will defer to previous VSP reviewers.

PR #3676, opened 2019-12-19
Create Veteran Confirmation /status endpoint

rtravitz:
## Description of change

This PR addresses [#3655](https://github.com/department-of-veterans-affairs/vets-contrib/issues/3655).

The PR adds the implementation for an endpoint "Veteran Confirmation API", which was created for Lighthouse in #3638 . This `/status` endpoint is secured by an API key and accepts attributes like name, birth date, and social security number in a POST body. That information is then used to query MVI to retrieve an ICN, which can then be used to query eMIS. A response of "confirmed" or "not confirmed" is then returned to the end user. 

### Notable changes
- Creates validations in the controller to check POST body 
- Adds a StatusService to handle the details of interacting with the MVI and EMIS services
- Adds an AttrService, which inherits from the `MVI::AttrService`. This inheritance is necessary because the `MVI::AttrConfiguration` used in the base AttrService uses a different url in some environments than the standard MVI url. This is because of requirements for the service that originally used the `MVI::AttrService` to hit the same MVI environment as HCA.

### Other notes:
- This endpoint is intended for use by identity providers that provide LOA3 by proxy.
- The MVI part of the work is based on [previous art by Health Care Application](https://github.com/department-of-veterans-affairs/vets-api/pull/3649#issuecomment-566747969), which queries MVI based on the attributes of non-logged-in users.

## Testing done

- Adds unit tests for all outcomes from the StatusService
- Adds integration tests for the interaction from the controller all the way to point where the MVI and EMIS services make calls to their backing systems
- Manually confirmed that Betamocks interactions are all returning appropriately for test users with the correct data in both MVI and eMIS

## Testing planned

- Confirm that Betamocks returns correct data in nonprod environments

#### Unique to this PR

- [x] A way to query MVI by user traits is created
- [x] Veteran status is confirmed or not confirmed when provided relevant user data

#### Applies to all PRs
- [x] Appropriate logging


modules/veteran_confirmation/app/controllers/veteran_confirmation/v0/veteran_status_controller.rb
```diff
@@ -5,8 +5,70 @@
...
+
+        render json: { veteran_status: status }
+      end
+
+      private
+
+      def validate_body
+        raise error_klass('Body must not be empty') if string_body.blank?
+
+        body = JSON.parse(string_body)
```
edmkitty: We should probably rescue JSON parse errors and tell the user something meaningful. Just in case

rtravitz: I made changes to how we're parsing the JSON body. It is automatically parsed into params, so we don't need to handle it manually.

modules/veteran_confirmation/VETERAN_CONFIRMATION.yml
```diff
@@ -51,24 +64,40 @@ paths:
                 properties:
                   data:
                     $ref: '#/components/schemas/VeteranStatusConfirmation'
-        '502':
-          description: eMIS failed to respond or responded in a way we cannot handle. 
+        '503':
```
nfasulo: The Facilities OAS has good examples of additional error statuses that can be added: https://github.com/department-of-veterans-affairs/vets-api/blob/cfd79ba48a17545f5b340217a7fa0e038a8b394f/modules/va_facilities/README.yml#L131

modules/veteran_confirmation/app/controllers/veteran_confirmation/v0/veteran_status_controller.rb
```diff
@@ -5,8 +5,59 @@
...
+        to_check.each do |field|
+          if params[field].blank?
+            raise Common::Exceptions::ParameterMissing.new(field,
+                                                           detail: "Must supply #{field} to query Veteran status")
+          end
+        end
+      end
+
+      def validate_ssn_format(ssn)
+        raise Common::Exceptions::InvalidFieldValue.new('ssn', 'the provided') unless valid_ssn?(ssn)
```
nfasulo: Not logging the value :1st_place_medal:

modules/veteran_confirmation/app/controllers/veteran_confirmation/v0/veteran_status_controller.rb
```diff
@@ -5,8 +5,59 @@
...
+        render json: { veteran_status: status }
+      end
+
+      private
+
+      def validate_body
+        parameters_exist?(%w[first_name last_name ssn birth_date])
+
+        validate_ssn_format(params['ssn'])
+        vali_date(params['birth_date'])
```
jholton: Do these methods need arguments since the arguments are `params` values which are available in the method definitions without passing them in?
In fact, both of these methods' definitions reference both the argument and the `params` value directly. (e.g. `vali_date`'s parameter is `date`, which is `params['birth_date']` and also references `params['birth_date']` directly)

modules/veteran_confirmation/app/controllers/veteran_confirmation/v0/veteran_status_controller.rb
```diff
@@ -5,8 +5,59 @@
...
+          birth_date: params['birth_date']
+        )
+
+        render json: { veteran_status: status }
+      end
+
+      private
+
+      def validate_body
+        parameters_exist?(%w[first_name last_name ssn birth_date])
```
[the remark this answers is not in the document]
rtravitz: Thanks for this one. Added!

modules/veteran_confirmation/app/controllers/veteran_confirmation/v0/veteran_status_controller.rb
rtravitz: They don't. I have an easier time tracing methods with explicit arguments, but this may not be Ruby-ish or fit the style we should be aim for. I took it out of the top-level methods as you suggested.

modules/veteran_confirmation/app/controllers/veteran_confirmation/v0/veteran_status_controller.rb
```diff
@@ -5,8 +5,50 @@
...
+
+      def all_digits?(ssn)
+        /^\d{9}$/.match?(ssn)
+      end
+
+      def all_digits_with_hyphens?(ssn)
+        /^\d{3}-\d{2}-\d{4}$/.match?(ssn)
+      end
+
+      def vali_date
```
kfrz: :smile_cat:

PR #3677, opened 2019-12-20
switch to using request store instead of thread.current

lihanli:
## Description of change
https://github.com/department-of-veterans-affairs/va.gov-team/issues/2264

## Testing done
automated tests

## Testing planned
test in staging

#### Applies to all PRs

- [x] Appropriate logging


lib/sidekiq/set_request_attributes.rb
```diff
@@ -2,7 +2,7 @@
 
 class Sidekiq::SetRequestAttributes
   def call(_worker, job, _queue, _redis_pool)
-    additional_request_attributes = Thread.current['additional_request_attributes'] || {}
+    additional_request_attributes = RequestStore.store['additional_request_attributes'] || {}
```
kfrz: Looks like this might need the [`request_store-sidekiq`](https://github.com/madebylotus/request_store-sidekiq) dependency as well? Just based on the README

kfrz: would apply to the other sidekiq changes too probably

lihanli: right now `Thread.current` is never directly used by code that will be run by a sidekiq worker, so it doesn't look we need request_store-sidekiq. `Sidekiq::SetRequestAttributes` is still running on the same thread that the web server is on, that's why it has access to `additional_request_attributes` which was set by the controller

kfrz: Awesome. Thanks for clearing that up

PR #3684, opened 2019-12-20
Make parent_code param optional

jbalboni:
## Description of change
The parent_code parameter is optional on the underlying service, this updates the vets-api service to reflect that, so that we can request a wider range of facilities from the FE and fix a bug around parent-child facility relationships.

## Testing done
Local and unit testing

- [x] Service doesn't return a 400 error when parent_code isn't passed

#### Applies to all PRs

- [x] Appropriate logging


jbalboni:
@saneshark I don't really understand what's going on with the spec, or if I actually need to make a copy of that cassette.

PR #3687, opened 2019-12-23
Fixing bad test sequence

bastosmichael:
## Description of change
Fixing test sequence fix, bad loading of the test was masking our lack of requiring X-Consumer-Username in the code, running the test individually like charlie did causes it to fail, running it as a group causes it to pass.

- Fixed test sequence

#### Applies to all PRs

- [ ] Appropriate logging

### DIFF
--- a/modules/claims_api/app/controllers/claims_api/application_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/application_controller.rb
@@ -38,7 +38,7 @@ module ClaimsApi
 
     def target_veteran(with_gender: false)
       if header_request?
-        headers_to_validate = %w[X-VA-SSN X-VA-First-Name X-VA-Last-Name X-VA-Birth-Date]
+        headers_to_validate = %w[X-VA-SSN X-VA-First-Name X-VA-Last-Name X-VA-Birth-Date X-Consumer-Username]
         headers_to_validate << 'X-VA-LOA' if v0?
         validate_headers(headers_to_validate)
         if v0?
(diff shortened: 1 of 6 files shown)


modules/claims_api/app/controllers/claims_api/application_controller.rb
```diff
@@ -38,7 +38,7 @@ def header_request?
 
     def target_veteran(with_gender: false)
       if header_request?
-        headers_to_validate = %w[X-VA-SSN X-VA-First-Name X-VA-Last-Name X-VA-Birth-Date]
+        headers_to_validate = %w[X-VA-SSN X-VA-First-Name X-VA-Last-Name X-VA-Birth-Date X-Consumer-Username]
```
ATeal: we have x-va-consumer-username and x-consumer-username in a couple of places can you grep and unify with the correct one

bastosmichael: X-consumer-username is the correct one because it comes from Kong and is only suppose to be used for V0 where we need the consumer name as the source for the user stored in the models

ATeal: Yeah thought so as well can you change the others

bastosmichael: I'm not seeing an x-va-consumer-usernames anywhere, can you point to where in the code you're seeing it?

bastosmichael: Fixed

PR #3694, opened 2019-12-23
Fix CC appointment request

saneshark:
## Description of change
Fixes: https://github.com/department-of-veterans-affairs/va.gov-team/issues/3975

Adds the ability to POST CC requests to the same endpoint. Uses a different form to validate based on presence of type field. 

## Testing done
Tested against staging instance and works, doesn't break existing functionality for VA requests.

## Testing planned
Integration testing on staging against the FE

#### Unique to this PR

#### Applies to all PRs

- [X] Appropriate logging


modules/vaos/app/controllers/vaos/appointment_requests_controller.rb
```diff
@@ -38,13 +45,13 @@ def id
...
 
-    def put_params
+    def params_for_update
       params.require(BASE_REQUIRED_PARAMS + [:created_date])
       params[:facility].require(%i[name facility_code parent_site_code])
       params.permit(*(BASE_PARAMS_WHITELIST + [:created_date]))
     end
 
-    def post_params
+    def params_for_create
```
sjbrush: not gonna lie, i like these names better.

modules/vaos/app/controllers/vaos/appointment_requests_controller.rb
```diff
@@ -5,16 +5,23 @@ class AppointmentRequestsController < VAOS::BaseController
...
 
     BASE_PARAMS_WHITELIST = [
-      :option_date1, :option_time1, :option_date2, :option_time2, :option_date3, :option_time3, :status,
+      :type, :option_date1, :option_time1, :option_date2, :option_time2, :option_date3, :option_time3, :status,
       :appointment_type, :visit_type, :phone_number, :email, :purpose_of_visit, :other_purpose_of_visit,
       :provider_id, :provider_name, :second_request, :second_request_submitted, :requested_phone_call, :type_of_care_id,
-      best_timeto_call: [], facility: %i[name facility_code state city parent_site_code object_type],
-      appointment_request_detail_code: [], patient: %i[inpatient text_messaging_allowed]
+      :additional_information, :address, :city, :state, :zip_code, :distance_willing_to_travel, :new_message,
+      :office_hours, :preferred_city, :preferred_state, :preferred_zip_code, :provider_option, :preferred_language,
```
johnpaulashenfelter: Office hours here is a string -- the attributes in the form is an array of strings. If that's correct, no problem. Just an FYI.

modules/vaos/app/services/vaos/appointment_requests_service.rb
```diff
@@ -74,11 +82,11 @@ def filter_cc_appointment_data(request)
       request
     end
 
-    def url(id = nil)
+    def url(id = nil, type = 'appointments')
       if id
         url + "/system/var/id/#{id}"
       else
-        "/var/VeteranAppointmentRequestService/v4/rest/appointment-service/patient/ICN/#{user.icn}/appointments"
+        "/var/VeteranAppointmentRequestService/v4/rest/appointment-service/patient/ICN/#{user.icn}/#{type}"
```
johnpaulashenfelter: The if/else here seems to do two pretty different things on each logic path. Its clear one path generates an appointment endpoint URL -- the other not so clear. Maybe `system/var/id/:id` is another URL for an appointment but its pretty obtuse without specific knowledge.

This code plus the `if` in the form object makes it seem like there are really two different flows that are pushed together into this one process. Wondering if there's a higher level refactor that makes everything simpler.

modules/vaos/app/controllers/vaos/appointment_requests_controller.rb
saneshark: Yes, there is a lot of redundancy, but garbage in, garbage out. 🤷‍♂

modules/vaos/app/services/vaos/appointment_requests_service.rb
saneshark: it's kind of a recursive if statement. I think I got a little code golf crazy here. I thought it was elegant, but not necessarily readable. I'll certainly be refactoring this. I'm actually looking to move most of the services logic into a base services class with url helper methods, etc.

PR #3704, opened 2019-12-26
Kr/change vet verification swagger docs

edmkitty:
Add metadata swagger docs to veteran verification.

https://github.com/department-of-veterans-affairs/vets-contrib/issues/3831


modules/veteran_verification/METADATA.yml
```diff
@@ -0,0 +1,96 @@
...
+    receiving authorization using an Open ID Connect flow: 
+
+    #### Veteran Status - Request confirmation from the VA of an individual's Veteran status 
+
+    The Veteran Verification API passes requests through to eMIS, the Enterprise
+    Military Information Service, and formats the response into consumable data.
+
+     #### Disability Rating - Access the disability rating of a Veteran
+
+    The Disability Rating API passes requests through to EVSS, the Electronic Veterans Self Service, and formats the response into consumable data.
```
rtravitz: Disability rating is just an endpoint and not an API

modules/veteran_verification/METADATA.yml
```diff
@@ -0,0 +1,96 @@
...
+  description: |
+
+    ## Background
+
+    The use-case of this API is to allow third-parties to do the following after
+    receiving authorization using an Open ID Connect flow: 
+
+    #### Veteran Status - Request confirmation from the VA of an individual's Veteran status 
+
+    The Veteran Verification API passes requests through to eMIS, the Enterprise
```
rtravitz: Veteran Verification _is_ an API, but in this case I think we mean the "Veteran Confirmation" endpoint.

modules/veteran_verification/METADATA.yml
```diff
@@ -0,0 +1,96 @@
...
+    Military Information Service, and formats the response into consumable data.
+
+     #### Disability Rating - Access the disability rating of a Veteran
+
+    The Disability Rating API passes requests through to EVSS, the Electronic Veterans Self Service, and formats the response into consumable data.
+
+    #### Service History - Access the service history of a Veteran
+
+    The Service History API passes requests through to eMIS, the Enterprise
```
rtravitz: Service History endpoint instead of API

modules/veteran_verification/METADATA.yml
```diff
@@ -0,0 +1,96 @@
...
+  
+    ## Design
+
+    ### Authorization
+
+    API requests are authorized using a Bearer token issued through an OpenID
+    Connect service to allow third-party applications. The token should be
+    submitted as an `Authorization` header in the form `Bearer <token>`.
+
+    ### Veteran Status Request
```
rtravitz: Adding a section for each request schema doesn't seem very useful or sustainable. Could we make this generic and only say it once? Most of this information is also repeated in the schemas displayed below.

modules/veteran_verification/METADATA.yml
edmkitty: Maybe "Veteran Status endpoint"? I feel like calling it Confirmation when we are adding another confirmation endpoint will just be confusing

modules/veteran_verification/METADATA.yml
```diff
@@ -0,0 +1,81 @@
...
+    
+    ### 2. Test the  API
+    
+    In the endpoint documentation below, we've provided a curl command builder for trying out the API before implementation with your app.
+
+    ### 3. Build your app
+
+    The base URI for the Veteran Verification API in the Development environment is:
+
+    dev-api.va.gov/services/veteran_verification/v0
```
rtravitz: Should we include the protocol `https://` here?

modules/veteran_verification/METADATA.yml
```diff
@@ -0,0 +1,81 @@
...
+
+    Check out some of our sample apps located on our [GitHub page](https://github.com/department-of-veterans-affairs/vets-api-clients/tree/master/samples/). Please visit our VA Lighthouse [Support portal](/support) should you need further assistance.
```
rtravitz: We can probably just link this as `[sample apps](linkgoeshere)` instead of specifying Github.

modules/veteran_verification/METADATA.yml
```diff
@@ -0,0 +1,81 @@
...
+    See our [OAuth Guide](/explore/verification/docs/authorization) for more details
+
+
+    ## Reference
+    
+    ### Development vs. Production Data
+    APIs accessed via the Development environment are using the same underlying logic as VA’s production APIs; only the underlying data store is different.
+  
+    ### Rate Limiting
+    We implemented basic rate limiting of 60 requests per minute. If you exceed this quota, your request will return a 429 status code. You may petition for increased rate limits by emailing   and requests will be decided on a case by case basis.
```
rtravitz: There are extra spaces between `emailing` and `and requests`. That actually needs a comma, too. 
```suggestion
    We implemented basic rate limiting of 60 requests per minute. If you exceed this quota, your request will return a 429 status code. You may petition for increased rate limits by emailing, and requests will be decided on a case by case basis.
```

modules/veteran_verification/spec/requests/api_docs_request_spec.rb
```diff
@@ -26,4 +26,12 @@
       JSON.parse(response.body)
     end
   end
+
+  describe '#get /docs/v0/veteran_verification' do
+    it 'returns Open API Spec v3 JSON' do
+      get '/services/veteran_verification/docs/v0/veteran_verification'
+      expect(response).to have_http_status(:ok)
+      JSON.parse(response.body)
```
nfasulo: I would add testing to verify that the parts of the schema pulled into VETERAN_VERIFICATION.yml from the existing yml files are present.

modules/veteran_verification/VETERAN_CONFIRMATION.yml
```diff
@@ -1,17 +1,17 @@
 openapi: '3.0.0'
 info:
   version: 0.0.1
-  title: Veteran Confirmation
+  title: Veteran Verification
   description: |
-    Veteran Verification - Veteran Confirmation
+    Veteran Verification - Veteran Status
```
edmkitty: Well, this probably shouldn't be happening.

PR #3705, opened 2019-12-26
Use bgs env from settings

charleystran:
## Description of change
Setting up the BGS connection to use the BGS env over the Rails env in the initializer. Its not 100% necessary for our connection to BGS since we connect through the proxy, but we should have it setup correctly in case anyone does want a direct connection. The devops changes for this have been done

## Testing done
tested on dev

- [ ] BGS env is properly set 

#### Applies to all PRs

- [ ] Appropriate logging


omgitsbillryan:
What's the difference between `config.env` & `config.forward_proxy_url`? Forgive my ignorance on the subject, I know very little about BGS.

charleystran:
We will be hitting the forward proxy URL but the the actual BGS url will be passed as a HOST header in the soap call. That being said, we need to pass the right bgs env for each environment so the proper url can be generated for the host headers

omgitsbillryan:
I'm still confused, so `config.env` makes its way into the `Host` request header for when this client calls out to BGS? So does that mean each of our fwd proxy connections in dev, staging, & prod all have access to each "webtest", "linktest", & "production" environments?

Also, just want to make sure it's ok for `LighthouseBGS` to have `config.env = nil` - which is what's going to happen for local envs. You'll want to set this to some "fakeenv" in `settings.yml` if not.

charleystran:
Local envs will not be able to talk directly to BGS in the same regard that local envs can't talk to EVSS or MVI without tunneling or a properly configured GFE. In those cases the proper setting needs to be added to the config/settings.local.yml based on which env the developer wishes to hit. 

The gem builds a URL based on the environment for the host header when a forward proxy is used.. We were improperly passing the Rails env to build that url. We would be getting development.vba.va.gov when we wanted webtest.vba.va.gov The purpose of this change is to set the correct values in line with what the proper BGS environment is for the deployment environment on vets api .

PR #3718, opened 2020-01-02
add global downtime notification flag

rianfowler:
## Overview 

Adds global downtime notification feature flag.


annaswims:
How do you expect this to work?   The values of the feature flag are cached in memory for 1 minute but stored in the database.  Will this work for your scenario?

rianfowler:
@annaswims sorry for the late reply. My plan is to turn this feature flag on for me in staging so that i can force the FE into a state where it thinks the two status apis are showing all of the services are down. 

we have a global downtime this weekend and I want to be able to test everything without taking vets-api down for an extended period on staging.

PR #3723, opened 2020-01-02
Get more info about token validation failures

kalilsn:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/3982

The /internal/auth/v0/validation endpoint has started sending a lot of events to sentry: http://sentry.vfs.va.gov/vets-gov/platform-api-production/issues/88135/. The error message here isn't very clear, but it seems to indicate that the public key specified by the access tokens being validated isn't being found. This PR adds some additional monitoring and more specific error messages to help us test our theories as we debug this problem.

### DIFF
--- a/app/models/token.rb
+++ b/app/models/token.rb
@@ -15,7 +15,6 @@ class Token
   def payload
     @payload ||= if @token_string
                    pubkey = public_key
-                   return if pubkey.blank?
 
                    JWT.decode(@token_string, pubkey, true, algorithm: 'RS256')[0]
                  end
--- a/app/models/token.rb
+++ b/app/models/token.rb
@@ -30,7 +30,10 @@ class Token
     decoded_token = JWT.decode(@token_string, nil, false, algorithm: 'RS256')
     kid = decoded_token[1]['kid']
     key = OIDC::KeyService.get_key(kid)
-    raise error_klass("Public key not found for kid specified in token: '#{kid}'") if key.blank?
+    if key.blank?
+      Rails.logger.info("Public key not found", kid: kid, exp: decoded_token[0]['exp'])
+      raise error_klass("Public key not found for kid specified in token: '#{kid}'")
+    end
 
     key
   rescue JWT::DecodeError => e
--- a/app/models/token.rb
+++ b/app/models/token.rb
@@ -31,7 +31,7 @@ class Token
     kid = decoded_token[1]['kid']
     key = OIDC::KeyService.get_key(kid)
     if key.blank?
-      Rails.logger.info("Public key not found", kid: kid, exp: decoded_token[0]['exp'])
+      Rails.logger.info('Public key not found', kid: kid, exp: decoded_token[0]['exp'])
       raise error_klass("Public key not found for kid specified in token: '#{kid}'")
     end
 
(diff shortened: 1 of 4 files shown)


app/models/token.rb
```diff
@@ -21,21 +21,25 @@ def payload
                  end
   rescue JWT::ExpiredSignature => e
     Rails.logger.info(e.message, token: @token_string)
-    raise error_klass("Validation error: #{e.message}")
+    raise error_klass(e.message)
+  rescue JWT::DecodeError => e
+    raise error_klass(e.message)
```
edmkitty: Is there a potential for token information to be exposed when the response is given here?

app/models/token.rb
```diff
@@ -21,21 +21,25 @@ def payload
...
   end
 
   def public_key
     decoded_token = JWT.decode(@token_string, nil, false, algorithm: 'RS256')
     kid = decoded_token[1]['kid']
-    OIDC::KeyService.get_key(kid)
+    key = OIDC::KeyService.get_key(kid)
+    raise error_klass("Public key not found for kid specified in token: '#{kid}'") if key.blank?
```
edmkitty: should we raise this error to the user or log it? I'm not sure how sensitive a kid is

app/models/token.rb
```diff
@@ -21,21 +21,28 @@ def payload
...
 
   def public_key
     decoded_token = JWT.decode(@token_string, nil, false, algorithm: 'RS256')
     kid = decoded_token[1]['kid']
-    OIDC::KeyService.get_key(kid)
+    key = OIDC::KeyService.get_key(kid)
+    if key.blank?
+      Rails.logger.info('Public key not found', kid: kid, exp: decoded_token[0]['exp'])
+      raise error_klass("Public key not found for kid specified in token: '#{kid}'")
+    end
```
dvogel: Since this function will never return a blank key, I think the `return` on line 18 can go away.

lib/okta/service.rb
```diff
@@ -51,5 +45,26 @@ def delete_grant(user_id, grant_id)
...
+        get_url_with_token(Settings.oidc.auth_server_metadata_url)
+      end
+    end
+
+    def keys
+      url = metadata.body['jwks_uri']
+      with_monitoring do
+        get_url_with_token(url)
+      end
+    end
```
dvogel: There's lots of stuff related to "keys" when it comes to okta. Can this function be named `oidc_jwks_keys` or something like that?

app/models/token.rb
kalilsn: I updated this to do both. The kid isn't sensitive at all + it's actually coming from the token they send us. So I don't think there's any risk to including it in the error (but there's also not that much to gain, just might steer a future provider in the right direction while debugging).

app/models/token.rb
kalilsn: No, none of these errors include any sensitive info. And it would be fine to include anything in the token in the response since the user has to send their token with the request.

The main concern would be something sensitive showing up in the logs, but none of the error messages generated by the jwt library seem problematic.

app/models/token.rb
kalilsn: Good point!

PR #3728, opened 2020-01-03
Veteran MVI lookup for 530 burials form 

bshyong:
## Description of change
Within the 530 Burial Benefits form flow, check whether the Veteran on the form exists in MVI. This work supports upcoming pension automation work for directly creating a claim via API from 530 form data rather than routing through Central Mail.

This work also includes the preliminary BIP Claims API service configuration

Resolves https://github.com/department-of-veterans-affairs/pension-and-fiduciary-automation/issues/188

**This PR will enable StatsD counter for MVI hit/miss on Veteran info within the 530 burial claims forms. By default this data will be retained for 15 days, but can be extended with a corresponding PR that modifies https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/deployment/config/prometheus/rules/job.rules

## Testing done
Local

## Testing planned
Staging/review instance

#### Unique to this PR
- [x] Appropriate logging


lib/bip_claims/service.rb
```diff
@@ -0,0 +1,37 @@
...
+module BipClaims
+  class Service < Common::Client::Base
+    STATSD_KEY_PREFIX = 'api.bip_claims'
+    include Common::Client::Monitoring
+
+    configuration BipClaims::Configuration
+
+    def veteran_attributes(claim)
+      case claim.form_id
+      when '21P-530'
```
jjulian: Is there a case where the form_id will _not_ be 530? What should we do then?

bshyong: This will only be the case if we are trying to use this service for other types of forms beyond 530 burials. For other types of forms we can add logic to accommodate the different key names/data shapes in those forms

lib/bip_claims/service.rb
```diff
@@ -0,0 +1,39 @@
...
+        ssn, full_name, bday = claim.parsed_form.values_at(
+          'veteranSocialSecurityNumber',
+          'veteranFullName',
+          'veteranDateOfBirth'
+        )
+      end
+
+      BipClaims::Veteran.new(
```
annaswims: Would you want a BipClaims::Veteran with a nil ssn or bday (if claim.form_id != '21P-530' )?

lib/bip_claims/service.rb
annaswims: This is more of a style thing, but Why have a case statement when `if` would do?

lib/bip_claims/service.rb
bshyong: I was anticipating having to support some other type of forms in the future but the timing on those are undetermined.

Would you recommend raising an error for "unsupported form id" or just returning a nil veteran? For this service, we expect only `21P-530` forms for the indefinite future

bshyong: @annaswims

I was anticipating having to support some other type of forms in the future but the timing on those are undetermined.

Would you recommend raising an error for "unsupported form id" or just returning a nil veteran? For this service, we expect only `21P-530` forms for the indefinite future

PR #3742, opened 2020-01-07
Throttle okta public keys requests for kids that have failed previously 

kalilsn:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/3982

## Description of change
When a token is sent to api.va.gov/internal/auth/v0/validation we look up the okta public key associated with the `kid` header in the token. These keys are cached in memory, but when the key isn't found we have to make two requests to okta before returning a failure. Since we're receiving many requests with invalid `kid` values, this PR stores previous bad `kid` values and ensures that we only make an upstream request once per minute for a given `kid`. In cases where the same bad `kid` is sent repeatedly (which is what's happening now), this should enable much faster failures for most requests.

This PR also adds a counter metric for kid lookup failures labeled by the `kid` so we can more easily address this kind of issue in the future.

## Testing done
Local testing + unit tests

#### Applies to all PRs

- [x] Appropriate logging


app/models/token.rb
```diff
@@ -30,6 +30,7 @@ def public_key
     kid = decoded_token[1]['kid']
     key = OIDC::KeyService.get_key(kid)
     if key.blank?
+      StatsD.increment('okta_kid_lookup_failure', 1, ["kid:#{kid}"])
```
kalilsn: Open to suggestions for other names for this metric. Also are there any other labels that would be useful here? The docs (https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/engineering/backend/sending-metrics-using-statsd.md#tagging) indicate that we won't be able to add additional dimensions to this metric in the future if we need to.

lib/oidc/key_service.rb
```diff
@@ -7,23 +7,45 @@ class KeyService
     @mutex = Mutex.new
     # Map from kid to OpenSSL::RSA::PKey for all current keys
     @current_keys = {}
-
-    class << self
-      attr_reader :current_keys
-    end
+    @cache_miss_kids = {}
+    KID_CACHE_PERIOD = 60
+    KID_CACHE_MAX_SIZE = 1000
```
kalilsn: The cache period and max size are both pretty arbitrary. The max size could probably be larger without much risk, but given that this is per instance, per deploy, I think 1000 should be fine for now.

dvogel: Specifying the unit (`KID_CACHE_PERIOD_SECS`) for the period could be a nice readability improvement in the comparison logic below.

I would suggest settings the max cache size lower to start. A malicious user could easily generate enough requests to fill the cache whether it's 100 entries, 1,000 entries, or 10,000 entries. A smaller cache size minimizes the time to update it, leaving more compute time and memory for other activities. Due to the ineffectiveness of memory compaction in ruby, the added memory fragmentation caused by the cache is more of a concern than the absolute memory usage at any given instant.

spec/lib/oidc/key_service_spec.rb
```diff
@@ -23,27 +23,64 @@
 
   describe '::get_key' do
     after do
-      described_class.instance_variable_set(:@current_key, {})
+      described_class.reset!
     end
 
     it 'returns the current key if it already exists' do
-      described_class.current_keys['key'] = 'key'
+      described_class.instance_variable_set(:@current_keys, 'key' => 'key')
```
dvogel: Thanks for cleaning up the public API of this class. I much prefer the test suite overriding the protection of the instance variable to the public API of the class exposing it's state _just_ to enable a testing scenario.

app/models/token.rb
dvogel: I think this is fine. We aren't relying on this yet in any operational capacity. Once our debugging is complete and we've used the metric a bit w can rename it, either to add new labels or because we find a better name. New acronym: GEFN (good enough for now) :smile:

lib/oidc/key_service.rb
```diff
@@ -7,23 +7,43 @@ class KeyService
     @mutex = Mutex.new
     # Map from kid to OpenSSL::RSA::PKey for all current keys
     @current_keys = {}
-
-    class << self
-      attr_reader :current_keys
-    end
+    @cache_miss_kids = {}
+    KID_CACHE_PERIOD_SECS = 60
+    KID_CACHE_MAX_SIZE = 10
```
mjlumetta: what's the current rate of cache misses, approximately?

kalilsn: Once this is deployed it should be about 1/day/instance, because there's only a single bad kid value that's being sent. Assuming all the clients are well behaved and that okta rotates the keys once/month there really should only ever be a maximum of one entry in here.

kalilsn: Actually we've received two different invalid `kid` values in the past 24 hours. Some non-json lines in the logs were breaking my script and I hadn't noticed.

kalilsn: `awslogs get -GS -s 24h -f 'Public key' dsva-vagov-prod/srv/vets-api/src/log/vets-api-server.log | jq -R '. as $raw | try (fromjson | .payload.kid) catch $raw' | sort | uniq -c`

PR #3721, opened 2020-01-02
validate nil attachment POST param

omgitsbillryan:
## Description of change
Adds basic param validation to all controllers using the `FormAttachmentCreate` concern. 

The motivation for this PR was a [`NoMethodError` in Sentry](http://sentry.vfs.va.gov/vets-gov/platform-api-production/issues/90775/?referrer=slack) where nested `params` parameters were being read w/o the safe-navigation operator or a validation that the param existed.

Connects: https://github.com/department-of-veterans-affairs/vets-api/pull/3721

## Testing done
Specs

## Testing planned
None.

#### Unique to this PR

- [ ] _Replace this line with individual tasks unique to your PR_

#### Applies to all PRs

- [ ] Appropriate logging
- [ ] Swagger docs have been updated, if applicable
- [ ] Provide link to originating GitHub issue, or connected to it via ZenHub
- [ ] Does not contain any sensitive information (i.e. PII/credentials/internal URLs/etc., in logging, hardcoded, or in specs)
- [ ] Provide which alerts would indicate a problem with this functionality (if applicable)

PR #3746, opened 2020-01-08
Bugfix: NoSuchKey error

annaswims:
https://github.com/department-of-veterans-affairs/va.gov-team/issues/4633

## Description of change
There was a bug introduced in https://github.com/department-of-veterans-affairs/vets-api/pull/3721 that attempted to save all of the parameters rather than just the file 

````
[1] pry(#<V0::HcaAttachmentsController>)> file_data
=> <ActionController::Parameters {"file_data"=>#<ActionDispatch::Http::UploadedFile:0x00007fdd6ded08f8 @tempfile=#<Tempfile:/var/folders/gs/qp8mkx392qjfgj22mv__h_br0000gn/T/RackMultipart20200107-29459-1gtucql.pdf>, @original_filename="extras.pdf", @content_type="", @headers="Content-Disposition: form-data; name=\"hca_attachment[file_data]\"; filename=\"extras.pdf\"\r\nContent-Type: \r\nContent-Length: 12161\r\n">} permitted: true>
````

## Testing done
specs

## Testing planned
upload in staging

### DIFF
--- a/app/controllers/concerns/form_attachment_create.rb
+++ b/app/controllers/concerns/form_attachment_create.rb
@@ -11,10 +11,9 @@ module FormAttachmentCreate
     form_attachment_model = self.class::FORM_ATTACHMENT_MODEL
     form_attachment = form_attachment_model.new
     namespace = form_attachment_model.to_s.underscore.split('/').last
+    filtered_params = params.require(namespace).permit(:file_data)
 
-    file_data = params.require(namespace).permit(:file_data)
-
-    form_attachment.set_file_data!(file_data)
+    form_attachment.set_file_data!(filtered_params[:file_data])
     form_attachment.save!
     render(json: form_attachment)
   end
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: app/controllers/concerns/form_attachment_create.rb [hunk 1]
```diff
@@ -11,10 +11,9 @@ def create
     form_attachment_model = self.class::FORM_ATTACHMENT_MODEL
     form_attachment = form_attachment_model.new
     namespace = form_attachment_model.to_s.underscore.split('/').last
+    filtered_params = params.require(namespace).permit(:file_data)
 
-    file_data = params.require(namespace).permit(:file_data)
-
-    form_attachment.set_file_data!(file_data)
+    form_attachment.set_file_data!(filtered_params[:file_data])
```
annaswims: @omgitsbillryan Should have been `filtered_params[namespace][:file_data]`, right?

omgitsbillryan: ah crap. ya that's a pretty important detail that we overlooked

annaswims: actually it works...
Screen Shot 2020-01-08 at 7 55 07 PM

PR #3748, opened 2020-01-08
Add source data to power of attorney and sent to BGS

charleystran:
## Description of change
This PR saves extra source data for the 2122 endpoints and sends them to BGS as external keys

## Testing done
Rspec

## Testing planned
Once deployed on dev, verifying call with BGS team

- [ ] External keys are sent to BGS in the service initialization

#### Applies to all PRs

- [ ] Appropriate logging


modules/claims_api/app/models/claims_api/power_of_attorney.rb
```diff
@@ -8,6 +8,7 @@ class PowerOfAttorney < ApplicationRecord
     include FileData
     attr_encrypted(:form_data, key: Settings.db_encryption_key, marshal: true, marshaler: ClaimsApi::JsonMarshal)
     attr_encrypted(:auth_headers, key: Settings.db_encryption_key, marshal: true, marshaler: ClaimsApi::JsonMarshal)
+    attr_encrypted(:source_data, key: Settings.db_encryption_key, marshal: true, marshaler: ClaimsApi::JsonMarshal)
```
kfrz: Anyway you'd think this could be DRY-d up? `confg = {}; attr_encrypted(:attr_name, config)` etc

kfrz:
Though typically we do like migrations & app code to live in [separate PRs](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/engineering/code_review_guidelines.md) and I'm not sure it applies to Lighthouse policy-wise it's still probably a good idea.

modules/claims_api/app/models/claims_api/power_of_attorney.rb
charleystran: Hey @kfrz i agree with you in most cases. However this dB change removed a column in favor of it being part of the encrypted hash. So a migration only commit would break implementation code. Mind you we also don’t have any production consumers using this endpoint yet so it’s safe to drop that column.

PR #3749, opened 2020-01-08
add system certs if given

nathanhruby:
## Description of change
This will copy ca certs to system locations and update the system bundle.  This injects them into the container during build, there is a devops PR to put them into place for the build

## Testing done
- Did a test build by hand for dev and prod builds

## Testing planned
- Watch builds from accompanying patch for devops repo, once built, login and check container to make sure correct certs are in place and working

- Resulting containers have correct certs after build and can talk to hosts using them

#### Unique to this PR
- This can not be done in entrypoint script because we drop root privs in the container build so do not have perms to alter the files.

#### Applies to all PRs

- [x] Appropriate logging
  - department-of-veterans-affairs/va.gov-team#2513
  - as we discovered Monday, Sentry will notify about failure to talk to various external things without added root ca certs.


nathanhruby:
confirmed that certs are in place and traffic works as expected after deploy

PR #3756, opened 2020-01-09
Add capability of POSTing to /appointments (direct scheduling)

saneshark:
## Description of change
Fixes https://github.com/department-of-veterans-affairs/va.gov-team/issues/2968

Introduces a new controller action on existing appointments endpoint, adding the ability to POST / create a new appointment. This type of appointment is known as a "direct scheduled appointment" which is different an appointment request. 

NOTE: I would have liked to also update the Gemfile. Namely moving faker and faker/medical into the development and test groups. If you need to invoke user factories, which in turn depend on mvi_profile factory, it is required. Having to constantly `require 'faker/medical` gets tedious. But frankly the gem is only used in one place from what I can tell so maybe `faker/medical` could be replaced altogether. It would be very easy to stub out the one place it is used in factories for gender with `%w[M F].sample`

## Testing done
Tested by reverse engineering the api capability manually on staging instance and recording vcr cassettes.

## Testing planned
FE verification / integration.

The FE confirms that the the new endpoint is working as expected / UAT. 

#### Applies to all PRs

- [x] Appropriate logging


modules/vaos/spec/factories/appointment_forms.rb
```diff
@@ -0,0 +1,64 @@
...
+      appt_type { 'P' }
+      purpose { '9' }
+      lvl { '1' }
+      ekg { '' }
+      lab { '' }
+      x_ray { '' }
+      preferred_email { 'test@va.gov' }
+      time_zone { 'America/Denver' }
+      desired_date { DateTime.new(2020, 0o1, 31, 0o0, 0o0, 0).iso8601(3) }
+      date_time { DateTime.new(2020, 0o1, 31, 21, 0o0, 0).iso8601(3) }
```
saneshark: I guess rubocop thinks this is more readable somehow?

PR #3762, opened 2020-01-09
expanding upon the usefulness of the SSO success KPI metric

ericbuckley:
## Description of change
- we should update the login success statsd metrics to only be emitted once (the success login metrics are sometimes getting logged twice. there are a number of users who can be up-leveled from LOA1 to LOA3 upon login, those users are redirected back to the IDP for a second round and thus get "logged in" twice)
- update the KPI to be slightly more generic, for example api.auth.no_shared_cookie (the current KPI, api.auth.mhv_cookie.no_user, is checking users that don't have a MHV shared cookie. when the new SSOe integration rolls out the user will have an SSOe shared cookie, thus we should update the name of the metric to be inclusive of users who don't have ANY shared cookie)
- add tags to the statsd KPI to include
  - the IDP the user authenticated with
  - what their LOA is

This PR addresses the first 3 steps outlined in https://github.com/department-of-veterans-affairs/va-sso/issues/50

#### Applies to all PRs

- [x] Appropriate logging


app/controllers/v0/sessions_controller.rb
```diff
@@ -99,11 +99,14 @@ def user_login(saml_response)
       if user_session_form.valid?
         @current_user, @session_object = user_session_form.persist
         set_cookies
-        # track users who need to re-login on MHV
-        StatsD.increment(STATSD_MHV_COOKIE_NO_ACCOUNT_KEY) unless @current_user.mhv_correlation_id
         after_login_actions
         redirect_to url_service.login_redirect_url
-        stats(:success, saml_response)
+        if self.location.start_with?(url_service.base_redirect_url)
```
patrickvinograd: Interesting is self.location the value of the `Location` header that will get sent in the redirect_to response above?

ericbuckley: yeah https://github.com/rails/rails/blob/be3d9daaa39aa69603af16b602fe62ae47548469/actionpack/lib/action_controller/metal/redirecting.rb#L63

app/controllers/v0/sessions_controller.rb
```diff
@@ -131,6 +134,10 @@ def stats(status, saml_response = nil, failure_tag = nil)
         StatsD.increment(STATSD_LOGIN_NEW_USER_KEY) if request_type == 'signup'
         StatsD.increment(STATSD_SSO_CALLBACK_KEY,
                          tags: ['status:success', "context:#{saml_response.authn_context}"])
+        # track users who have a shared sso cookie
+        StatsD.increment(STATSD_SSO_SHARED_COOKIE,
```
patrickvinograd: We seem to. be initializing all the StatsD counters in `/config/initializers/statsd.rb`. These also appear to spit out a bunch of warnings on app startup but I think the alternative is that we get some weird behavior in prometheus with uninitialized counters.

ericbuckley: @patrickvinograd did we need something else other than this?  https://github.com/department-of-veterans-affairs/vets-api/pull/3762/files#diff-da38dd1d4d165deac7ffeeb8f413bbdeR16

patrickvinograd:
@annaswims particularly interested in your review on this. The "SSO shared cookie" stuff is relatively new that Ben introduced in the context of the SSO work that Eric is now refactoring. But the big change is the conditional on invoking the `stats` method, meant to only measure stats on the terminal SAML response, and not double-count the upleveling requests. So more measuring "user authentication events" instead of "saml responses".  If you think this causes any loss of fidelity/useful info in any existing dashboards that's helpful info.

annaswims:
I don't have read access to the va-sso, repo so I don't know what KPI's you're going for, but the success rate numbers are particularly squishy. [The panel](http://grafana.vfs.va.gov/d/000000033/site-authentication?orgId=1&fullscreen&panelId=21) describing why got moved a bit fro the graph that it's describing, but 

> The Vague approximation of auth success rate graph measures login attempts vs. login success, however there isn't a 1-1 relationship between the Authentication context (login type) for attempts vs. success.
> 
> For example a user may log in with DS Logon and then use id.me for 2-factor authentication, which would count as 1 DS logon attempt, 1 DS Logon success and 1 id.me success. Therefore, the success rate of id.me is likely to be greater than 100%.
> 
> Also, there are plenty of reasons auth may not be successful, like user abandonment but a significant dip in success rate is a sign of trouble.
> 
> Typical succes rates are: DS Logon & MHV~ 0.8 id.me ~ 1.2`
> 
> 

There's not a 1-1 mapping between contexts for attempts and successful logins.  For example an id.me signup is `signup` -> `http://idmanagement.gov/ns/assurance/loa/1/vets` or `http://idmanagement.gov/ns/assurance/loa/3/vets` . So, the result of a successful signup looks the same as a login. 

> **Attempt context:**
> dslogon
> idme
> mfa
> mhv
> signup
> slo
> verify
> 
> **Success context:**
> dslogon
> dslogon_loa3
> dslogon_multifactor
> http://idmanagement.gov/ns/assurance/loa/1/vets
> http://idmanagement.gov/ns/assurance/loa/3/vets
> multifactor
> myhealthevet
> myhealthevet_loa3
> myhealthevet_multifactor

If you do decide to adjust the metrics, take [the alerts](https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/deployment/config/prometheus/rules/external_service.rules.j2#L108) into consideration.

The 2 things most affecting these metrics are: "is the identity provider reliable?" and the quick answer to that is ["not at all"](https://dsva.pagerduty.com/services/P9DJJAV?status=acknowledged%2Ctriggered%2Cresolved) for DS Logon. Multi-day outages for DS Logon are common (I'm guessing 3-4 this year) and multi-hour outages are becoming a weekly occurrence.  

We have some insight to that with https://dslogon.statuspage.io (this uses a pupeteer script to log Boris every minute or so).

Also, all of these metrics include human error.  If a user forgets their password or for some reason decides not to continue with the login process that is included in this metric.

spec/controllers/v1/sessions_controller_spec.rb
```diff
@@ -86,6 +86,7 @@ def expect_logger_msg(level, msg)
                 .to trigger_statsd_increment(described_class::STATSD_SSO_NEW_KEY, tags: ["context:#{type}"], **once)
 
               expect(response).to have_http_status(:found)
+              # FIXME: should this not be a PINT url?
```
ericbuckley: @patrickvinograd / @hinzed1127 I believe these response.location values should be pointing to PINT urls.  I'm guessing that its just using the same mocked build response values that v0 is using.  However nothing jumped out at me in the `spec/support/saml/response_builder.rb` module.  Any ideas to where this is being set?

ericbuckley: think I figured it out, working on the changes now...https://github.com/department-of-veterans-affairs/vets-api/blob/master/spec/factories/rubysaml_settings.rb

ericbuckley:
@hinzed1127 / @patrickvinograd if one of you have a chance could you paste in some details as to why Jenkins can't build this branch?  Without the SSH proxy I don't have access to http://jenkins.vfs.va.gov/job/testing/job/vets-api/job/PR-3762/5/display/redirect

ericbuckley:
@patrickvinograd / @hinzed1127 all the tests and linting checks have passed now, so this is ready for a full review.  Let me know if I should ping anyone else for a review.

ericbuckley:
> Could you hold off on merging this for just a bit? [department-of-veterans-affairs/va.gov-team#5260](https://github.com/department-of-veterans-affairs/va.gov-team/issues/5260) . I was doing some digging last week and saw that MHV login success rate dropped 15% one Day in December and have tried to reach out to the MHV team to see if they can see an issue. I'd really like to see if we can validate that fix before we change how we measure success
> 
> Also, we have [downtime banners](https://github.com/department-of-veterans-affairs/devops/blob/master/docs/External%20Service%20Integrations/DS_logon_Banner.png) that automatically show up on the login modal based upon these metrics. So the alerts would need to be tuned as soon as this is released.

@annaswims sure thing, and I will make sure to have a matching PR in the devops repo ready to roll out before merging this

config/initializers/loa.rb
```diff
@@ -8,5 +8,5 @@ module LOA
 
   IDME_LOA1 = 'http://idmanagement.gov/ns/assurance/loa/1/vets'
   IDME_LOA2 = 'http://idmanagement.gov/ns/assurance/loa/2/vets'
-  IDME_LOA3 = 'http://idmanagement.gov/ns/assurance/loa/3/vets'
```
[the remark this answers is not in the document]
ericbuckley: The change to `IDME_LOA3` works for both v0 and v1.  The IAM team requested that we make this change for SSOe authentication, however I'm not sure why they didn't make the same ask for `IDME_LOA1`.  I'll work on getting a better explanation from their team and post that on this PR.

annaswims:
The MHV login success rate has gone back up.  They made a fix with their proxy rules overnight.  So, that's not longer an issue.  Thanks for waiting 🎉

config/initializers/loa.rb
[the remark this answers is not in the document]
ericbuckley: I've ran tests locally using the `v0` controller and the new IDME_LOA3 works as expected.  To add some context, the change was made because SSOe already recognised the AuthNContext urls without the `/vets` suffix, so ID.ME made a change to support both.  Going forward we could probably use either of the 2 urls with the `v0` controller, but I'm guessing they might deprecate `http://idmanagement.gov/ns/assurance/loa/3/vets` sometime in the future.

ericbuckley:
> @annaswims particularly interested in your review on this. The "SSO shared cookie" stuff is relatively new that Ben introduced in the context of the SSO work that Eric is now refactoring. But the big change is the conditional on invoking the `stats` method, meant to only measure stats on the terminal SAML response, and not double-count the upleveling requests. So more measuring "user authentication events" instead of "saml responses". If you think this causes any loss of fidelity/useful info in any existing dashboards that's helpful info.

@patrickvinograd / @annaswims  I refactored the stats methods, breaking them out into 2 sets.  One to track every callback iteration (pretty much the stats we are currently tracking) and a new method to track stats that are only marked when the SAML roundtrips are complete.

The thought here is we can leave the [alerts as is](https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/deployment/config/prometheus/rules/external_service.rules.j2#L108-L124), because they will continue to track callback invocations.  It would have been complicated to modify previously because the rate of successfull or failed logins wouldn't always be half of our STATSD_API_AUTH_NEW requests (because some users would only have LOA1 access and not need an uplevel).

Let me know what you think https://github.com/department-of-veterans-affairs/vets-api/pull/3762/commits/61c37de6c1b4efa54c02ac66fb5db008eeac7d52

patrickvinograd:
I definitely like the change to separating out callback metrics from logical login events, 👍 

Will wait for final PR approval until this gets rebased with the changes that went in to master (sorry for the extra back and forth)

ericbuckley:
@patrickvinograd merge is complete, def a much simplier PR now.  The codeclimate issues are related to some duplicate functions in v0/v1 sessions controller.  Is there a way to comment those out from being analyzed by codeclimate or can we just ignore them?

patrickvinograd:
CodeClimate is not a required check so other than the aesthetics of the partially red circle, we're fine on that front. Will take a look.

app/controllers/v0/sessions_controller.rb
```diff
@@ -125,12 +129,39 @@ def user_logout(saml_response)
...
+        # track users who have a shared sso cookie
+        if cookies.key?(Settings.sso.cookie_name)
+          StatsD.increment(STATSD_LOGIN_SHARED_COOKIE,
+                           tags: ["loa:#{@current_user.loa[:current]}",
+                                  "idp:#{@current_user.identity.sign_in[:service_name]}"])
+        end
+        StatsD.increment(STATSD_LOGIN_STATUS,
+                         tags: ['status:success',
+                                "idp:#{@current_user.identity.sign_in[:service_name]}",
+                                "context:#{saml_response.authn_context}"])
```
annaswims: do we gain any additional information from having both `idp` and `context`? 

id.me context can be rather long (`http://idmanagement.gov/ns/assurance/loa/1/vets`) and has a colon in it, which I think may cause problems for statsd.

in current dashboards, context has values like  dslogon, idme, mfa, mhv, signup, slo, verify. not `authn_context`

annaswims:
I don't see any changes to the /v0 specs

app/controllers/v0/sessions_controller.rb
```diff
@@ -99,17 +100,20 @@ def user_login(saml_response)
       if user_session_form.valid?
         @current_user, @session_object = user_session_form.persist
         set_cookies
-        # track users who need to re-login on MHV
-        StatsD.increment(STATSD_MHV_COOKIE_NO_ACCOUNT_KEY) unless @current_user.mhv_correlation_id
         after_login_actions
         redirect_to url_service.login_redirect_url
-        stats(:success, saml_response)
+        if location.start_with?(url_service.base_redirect_url)
```
annaswims: is there a corresponding PR for the alert rules?  https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/deployment/config/prometheus/rules/external_service.rules.j2#L100 . If this goes out without a reduction in the alert threshold, someone will get paged and users will see a downtime notification that looks like this https://github.com/department-of-veterans-affairs/devops/blob/master/docs/External%20Service%20Integrations/DS_logon_Banner.png

app/controllers/v0/sessions_controller.rb
ericbuckley: the goal was to know the IDP and the level the user is authenticating at (eg level 1 vs level 3).  If we only use the path and drop the `http:` prefix, would that work better with statsd?

app/controllers/v0/sessions_controller.rb
ericbuckley: the existing `stats` method was renamed to `callback_stats`.  These are the values that are added on every callback, and should be what the alerts are triggered off of.  The new `login_stats` method, only gets called once per user login, and is tracking new metrics that are not being used in any alert rules. https://github.com/department-of-veterans-affairs/vets-api/pull/3762#issuecomment-583475348

ericbuckley:
> I don't see any changes to the /v0 specs

missed that, will update now

app/controllers/v0/sessions_controller.rb
annaswims: I was mistaken, they're fine as is 
like this graph...
http://grafana.vfs.va.gov/d/000000033/site-authentication?orgId=1&fullscreen&panelId=16

PR #3763, opened 2020-01-10
Set default workers / threads

jbritt1:
## Description of change

I am hesitant to try this again due to what happened earlier in the week, but after talking with @wyattwalter and @johnpaulashenfelter, I think this will set a default in the Puma config to what is currently being used (meaning no changes should be seen by what is committed here) and allow for those configurations to be overwritten by environment variables. This would allow for us to do a dev-env deploy to test, allow for different settings dependent on environment, etc. 

Other than setting `- "-e PUMA_THREADS=16"` and `- "-e PUMA_WORKERS=1"` in the `docker_options` within the deployment configs, I am unsure what other changes would need to be made to align with the docker stuff @nathanhruby is working on. Any insight on that would be awesome.

In addition, I know there are some more settings like `preload_app!` and instructions for `on_worker_boot`, but a slower / more methodical approach with incremental changes is likely the safest route.

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -2,6 +2,6 @@
 
 plugin :statsd
 
-workers Integer(ENV.fetch("PUMA_WORKERS", 0))
-threads_count = Integer(ENV.fetch("PUMA_THREADS", 16))
+workers Integer(ENV.fetch('PUMA_WORKERS', 0))
+threads_count = Integer(ENV.fetch('PUMA_THREADS', 16))
 threads(threads_count, threads_count)
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -1,3 +1,7 @@
 # frozen_string_literal: true
 
 plugin :statsd
+
+workers Integer(ENV.fetch("PUMA_WORKERS", 0))
+threads_count = Integer(ENV.fetch("PUMA_THREADS", 16))
+threads(threads_count, threads_count)


johnpaulashenfelter:
@jbritt1 This is about as basic as it gets 👍 

The defaults should match what we have currently have and I verified both a `rails server` and `foreman start` bootup with defaults

```
$ bundle exec  foreman s web

10:42:05 web.1       | started with pid 79313
10:42:12 web.1       | => Booting Puma
10:42:12 web.1       | => Rails 5.2.4.1 application starting in development
10:42:14 web.1       | Puma starting in single mode...
.... lots of statsd
10:42:14 web.1       | * Version 4.3.1 (ruby 2.4.5-p335), codename: Mysterious Traveller
10:42:14 web.1       | * Min threads: 16, max threads: 16
10:42:14 web.1       | * Environment: development
10:42:14 web.1       | * Listening on tcp://0.0.0.0:3000
10:42:14 web.1       | Use Ctrl-C to stop
```

and

```
 $ PUMA_WORKERS=2 PUMA_THREADS=8 bundle exec foreman start web
10:43:56 web.1       | started with pid 79605
10:44:03 web.1       | => Booting Puma
10:44:03 web.1       | => Rails 5.2.4.1 application starting in development
....lots of statsd
10:44:05 web.1       | [79605] Puma starting in cluster mode...
10:44:05 web.1       | [79605] * Version 4.3.1 (ruby 2.4.5-p335), codename: Mysterious Traveller
10:44:05 web.1       | [79605] * Min threads: 8, max threads: 8
10:44:05 web.1       | [79605] * Environment: development
10:44:05 web.1       | [79605] * Process workers: 2
10:44:05 web.1       | [79605] * Phased restart available
10:44:05 web.1       | [79605] * Listening on tcp://0.0.0.0:3000
10:44:05 web.1       | [79605] Use Ctrl-C to stop
10:44:05 web.1       | [79605] - Worker 0 (pid: 79634) booted, phase: 0
10:44:05 web.1       | [79605] - Worker 1 (pid: 79635) booted, phase: 0
```

I don't think it is a problem to add the `on_worker_boot` since its only going to happen when a worker boots (PUMA_WORKERS >= 1) but i'm fine with that in the next release.

@jbritt1 can you or someone on devops deploy this branch (rebased off the docker deploy going out later today) to a dev/staging box so we can kick the tires a little?

jbritt1:
Thanks guys! @kfrz I think it was decided best to pin it until next week due to the number of other things in progress currently.

PR #3768, opened 2020-01-10
X-Consumer-Username should not be required field

bastosmichael:
## Description of change

Resolves https://github.com/department-of-veterans-affairs/vets-contrib/issues/4040

- Don't require X-Consumer-Username for submissions needing required headers.

#### Applies to all PRs

- [ ] Appropriate logging

### DIFF
--- a/modules/claims_api/app/controllers/claims_api/application_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/application_controller.rb
@@ -57,6 +57,7 @@ module ClaimsApi
       elsif !request.headers['X-Consumer-Username']
         log_message_to_sentry('Kong no longer sending X-Consumer-Username', :error,
                               body: request.body)
+        validate_headers(['X-Consumer-Username'])
       end
     end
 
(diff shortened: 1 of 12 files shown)


modules/claims_api/app/controllers/claims_api/application_controller.rb
```diff
@@ -51,6 +51,15 @@ def v0?
       request.env['PATH_INFO'].downcase.include?('v0')
     end
 
+    def check_source_user
+      if !header('X-VA-User') && request.headers['X-Consumer-Username']
```
ATeal: why not just use the `v0` based conditional, we already use it for loa, feel like that might simplify and provide an opportunity to refactor with the other v0 headers.

Also could you add a test?

bastosmichael: Yes I'll include a test, I want to keep loa seperate from source user but both do fall under the V0 if statement above so that it stays within the claims v0 scope.

bastosmichael: So I couldn't figure out a way to do a test for the sentry entry but I did add the X-Consumer-Username validation back on just set it only to work under V0 in this case. Updated the test accordingly.

PR #3771, opened 2020-01-13
Use sed instead of ruby sub for memory performance

ATeal:
## Description of change
In production ruby's `sub` was failing due to memory with ~100mb base 64 payloads. Using sed is a bit more memory efficient

resolves https://github.com/department-of-veterans-affairs/vets-contrib/issues/4016

## Testing done
- prod server + worker server console parsing

## Testing planned
- rerunning payloads once deployed

#### Applies to all PRs

- [x] Appropriate logging

### DIFF
--- a/modules/vba_documents/lib/vba_documents/multipart_parser.rb
+++ b/modules/vba_documents/lib/vba_documents/multipart_parser.rb
@@ -40,9 +40,8 @@ module VBADocuments
     end
 
     def self.create_file_from_base64(infile)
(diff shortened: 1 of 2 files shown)


bastosmichael:
I think sed is also missing in the CI/CD containers but that should be a quick fix.

johnpaulashenfelter:
@bastosmichael wonder if they're not running docker. I'll check w/ @nathanhruby

PR #3772, opened 2020-01-13
4129 Facility Locator Urgent Care Service Type Splits (Data)

darkbushido:
Extracting radius calculations from build params
reformatting the hash to make it easier to read
renaming build_params to provider_locator_params to be more descriptive
creating pharmacy_locator_params
using a new path for pharmacy_locator

Following the ruby style guide
  - No Explicit return
  - case vs if-else

## Description of change

- Duplicating the Community Care Provider search to the CCP controller
- Modifying Community Care Provider search to support
  - Pharmacy
  - Walk-In Care
  - Urgent Care

## Testing planned
This needs to be tested in an environment with access to PPMS

- Within a given bounding box search for the following Community Care Places of Service
  - Pharmacies (PoS 01)
  - Walk-In Care (PoS 17)
  - Urgent Care (PoS 20)

#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging


app/controllers/v0/facilities/ccp_controller.rb
```diff
@@ -1,10 +1,26 @@
...
+  def index
+    case params[:type]
+    when 'cc_provider'
+      provider_locator
+    when 'cc_pharmacy'
+      pos_locator('01')
+    when 'cc_walkin'
+      pos_locator('17')
+    when 'cc_urgent_care'
+      pos_locator('20')
```
johnpaulashenfelter: It seems like these should have corresponding changes in the controller spec

app/controllers/v0/facilities/ccp_controller.rb
```diff
@@ -14,13 +30,60 @@ def show
...
+
+    pages = { current_page: page, per_page: BaseFacility.per_page,
+              total_pages: place_of_service.length / BaseFacility.per_page + 1, total_entries: total }
+
+    render json: providers,
+           each_serializer: ProviderSerializer,
+           meta: { pagination: pages }
+  end
+
+  def ppms
```
johnpaulashenfelter: What do you think about calling it `client` or `api`? Since its private I shouldn't care about what it is, and with it you get controller methods that look something like:

```
    result = api.specialties
    render json: result
```

or

```
    result = client.specialties
    render json: result
```

Just a thought

app/controllers/v0/facilities/ccp_controller.rb
```diff
@@ -1,10 +1,26 @@
 # frozen_string_literal: true
 
 class V0::Facilities::CcpController < FacilitiesController
+  before_action :validate_params,
```
johnpaulashenfelter: I don't think you gain a lot by using these `only: SINGLE_CONTROLLER_ACTION` before actions. Its one more place to have to look in code to see what's going on. Leaving them in the controller method seems like it would be easier to understand

lib/facilities/ppms/client.rb
```diff
@@ -47,21 +55,46 @@ def specialties
...
+          driveTime: 10_000,
+          posCodes: pos_code,
+          network: 0,
+          maxResults: 20 * page + 1
+        }
+      end
+
+      def provider_locator_params(params)
+        page = Integer(params[:page] || 1)
+        # more estimation fun about 69 miles between latitude lines, <= 69 miles between long lines
```
johnpaulashenfelter: Comment doesn't seem connected to this code anymore

config/routes.rb
```diff
@@ -145,7 +145,7 @@
 
     scope :facilities, module: 'facilities' do
       resources :va, only: %i[index show], defaults: { format: :json }
-      resources :ccp, only: %i[show], defaults: { format: :json }
+      resources :ccp, only: %i[index show], defaults: { format: :json }
```
johnpaulashenfelter: I know it was already there, but its weird you'd need the format specified here as json since its the default for /v0

app/controllers/v0/facilities/ccp_controller.rb
```diff
@@ -1,10 +1,26 @@
...
+                :validate_no_services_without_type,
+                only: [:index]
+
   before_action :validate_id, only: [:show]
 
+  def index
+    case params[:type]
+    when 'cc_provider'
+      provider_locator
```
johnpaulashenfelter: I wonder if it would be clearer here to return the object and the pages and put the render method here in the controller action where you'd typically find it:

```
providers, pages = provider_locator
render json: providers,
	each_serializer: ProviderSerializer,
     meta: { pagination: pages }
```

or even

```
providers, = provider_locator
render json: providers,
	each_serializer: ProviderSerializer,
     meta: { pagination: paginate_providers(providers }
```

app/controllers/v0/facilities/ccp_controller.rb
darkbushido: I don't have any issues with either.

app/controllers/v0/facilities/ccp_controller.rb
darkbushido: this was a cargo cult from the VA controller.

lib/facilities/ppms/client.rb
darkbushido: I abstracted the radius calculation and didn't move the comment.

app/controllers/v0/facilities/ccp_controller.rb
darkbushido: This is an odd one. It's similar to the original implementation. This really shouldn't have pagination since we can't get page 2 from PPMS.

spec/controllers/v0/facilities/ccp_controller_spec.rb
```diff
@@ -3,35 +3,256 @@
...
+    data = fake_data(attributes)
+    if set
+      { 'data' => [data] }
+    else
+      { 'data' => data }
+    end
+  end
+
+  describe '#index' do
+    def strong_params(wimpy_params)
```
johnpaulashenfelter: This is pretty unusual -- i'm confused why you're needing to do this in your part of the API instead of just simply sending params

spec/controllers/v0/facilities/ccp_controller_spec.rb
```diff
@@ -3,35 +3,256 @@
...
+
+  def fake_provider_serializer(provider_hash, details_hash = {}, set = true)
+    attributes = details_hash.merge(provider_hash)
+    data = fake_data(attributes)
+    if set
+      { 'data' => [data] }
+    else
+      { 'data' => data }
+    end
+  end
```
johnpaulashenfelter: Shouldn't the response you're creating from the controller action already handle all of this?

config/routes.rb
johnpaulashenfelter: Looks like we do it everywhere.

lib/facilities/ppms/client.rb
```diff
@@ -15,8 +15,16 @@ class Client < Common::Client::Base
...
-        response = perform(:get, 'v1.0/ProviderLocator?', qparams)
+        qparams = provider_locator_params(params)
+        response = perform(:get, 'v1.0/ProviderLocator', qparams)
+        return [] if response.body.nil?
+
+        Facilities::PPMS::Response.from_provider_locator(response, params)
+      end
+
+      def pos_locator(params, pos_code)
+        qparams = pos_locator_params(params, pos_code)
```
johnpaulashenfelter: qparams == query params?

spec/factories/providers.rb
```diff
@@ -2,15 +2,32 @@
...
+    trait :from_provider_locator do
+      AddressStreet { Faker::Address.street_address }
+      AddressCity { Faker::Address.city }
+      AddressStateProvince { Faker::Address.state_abbr }
+      AddressPostalCode { Faker::Address.zip }
+      CareSitePhoneNumber { Faker::PhoneNumber.phone_number }
+      IsAcceptingNewPatients { Faker::Boolean.boolean }
+      Latitude { Faker::Address.latitude }
+      Longitude { Faker::Address.longitude }
+    end
```
johnpaulashenfelter: `trait` was a good call

app/controllers/v0/facilities/ccp_controller.rb
```diff
@@ -3,24 +3,74 @@
...
+  def page
+    Integer(params[:page] || 1)
+  end
+
+  def pages(ppms_results)
+    total = ppms_results.length
+    {
+      current_page: page,
+      per_page: BaseFacility.per_page,
+      total_pages: ppms_results.length / BaseFacility.per_page + 1,
```
johnpaulashenfelter: Is this always an integer? maybe it needs round/floor/ceiling?

spec/lib/facilities/ppms/client_spec.rb
```diff
@@ -1,89 +1,183 @@
...
 
-    describe 'route_fuctions' do
-      context 'with an http timeout' do
-        before do
-          allow_any_instance_of(Faraday::Connection).to receive(:get).and_raise(Faraday::TimeoutError)
-        end
+  context 'with an http timeout' do
+    before do
+      allow_any_instance_of(Faraday::Connection).to receive(:get).and_raise(Faraday::TimeoutError)
+    end
```
johnpaulashenfelter: Indentation diff makes it harder to see on GitHub, but is this a before for a single example? If it is I'd consider just making it part of the example

spec/lib/facilities/ppms/client_spec.rb
```diff
@@ -1,89 +1,183 @@
...
-          expect(r[0]['CareSiteAddressStreet']).not_to be(nil)
-        end
+  describe '#provider_services' do
+    it 'returns Services' do
+      VCR.use_cassette('facilities/va/ppms', match_requests_on: %i[path query]) do
+        r = Facilities::PPMS::Client.new.provider_services(1_407_842_941)
+        expect(r.each_with_object(Hash.new { |h, k| h[k] = Hash.new(0) }) do |service, count|
+          %w[Name AffiliationName RelationshipName CareSiteName CareSiteAddressZipCode].each do |key|
+            count[key][service[key]] += 1
+          end
```
johnpaulashenfelter: This might be pretty hard to figure out when someone reads in the future -- though i get the alternative is pretty huge. Would it be worth matching to a fixture so the code is a little simpler?

```
expect(r).to match(JSON.parse(some_giant_fixture_on_disk))
```

Might be just as easy to consider a simpler cassette if that path makes sense.

spec/request/swagger_spec.rb
```diff
@@ -1278,6 +1278,18 @@
...
       context 'successful calls' do
+        let(:provider) { FactoryBot.build(:provider, :from_provider_info) }
+        let(:provider_services_response) do
+          {
+            'CareSiteAddressStreet' => Faker::Address.street_address,
+            'CareSiteAddressCity' => Faker::Address.city,
+            'CareSiteAddressZipCode' => Faker::Address.zip,
+            'CareSiteAddressState' => Faker::Address.state_abbr,
+            'Latitude' => Faker::Address.latitude,
+            'Longitude' => Faker::Address.longitude
```
johnpaulashenfelter: I'd recommend hardcoding. Not sure fake info here brings much value

spec/request/v0/facilities/ccp_request_spec.rb
```diff
@@ -0,0 +1,175 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require File.expand_path(
+  Rails.root.join(
+    'spec',
+    'support',
+    'shared_contexts',
+    'facilities_ppms.rb'
+  )
```
johnpaulashenfelter: Surprises me you'd have to load it this way -- guessing its because we're not loading spec/support/* in the rspec helper, right?

lib/facilities/ppms/client.rb
darkbushido: it's the request params that we are sending to ppms. I left it the way it was done originally.

app/controllers/v0/facilities/ccp_controller.rb
darkbushido: [per_page](https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/models/base_facility.rb#L104-L106) it is for now. I have an issue to make it a parameter.

spec/lib/facilities/ppms/client_spec.rb
darkbushido: it is, I'll move it into the `it` block

spec/lib/facilities/ppms/client_spec.rb
darkbushido: I'm not happy with this one at all. currently, we jump through two objects to [get_body](https://github.com/department-of-veterans-affairs/vets-api/blob/4129-facility-locator-urgent-care-service-type-splits/lib/facilities/ppms/response.rb#L31-L33)

This is something I'm going to change in the future but I'm fine with this test case going away since we don't do anything with the response. it's mostly here so I could get the 100% code coverage.

spec/request/swagger_spec.rb
darkbushido: ok, I'll make the change.

spec/request/v0/facilities/ccp_request_spec.rb
darkbushido: [This](https://github.com/department-of-veterans-affairs/vets-api/blob/4129-facility-locator-urgent-care-service-type-splits/spec/rails_helper.rb#L13-L23) and [This](https://github.com/department-of-veterans-affairs/vets-api/blob/4129-facility-locator-urgent-care-service-type-splits/spec/spec_helper.rb#L4-L15) could be replaced with 
```
Dir[Rails.root.join("spec/support/**/*.rb")].each {|f| require f}
```

darkbushido:
> I think this reads pretty well. A bunch of suggestions, but only real serious thing for you to look at is whether rounding the `total_pages` is required -- I think it might be.

I'll cast it to an Integer for now.

spec/controllers/v0/facilities/ccp_controller_spec.rb
darkbushido: I'm using strong parameters, `ActionController::Parameters` no longer extends `Hash`. so the `.with()` mock won't get a `Hash`

spec/controllers/v0/facilities/ccp_controller_spec.rb
darkbushido: I split these into two methods

PR #3773, opened 2020-01-13
3602 VAOS full institution data

kreek:
## Description of change
Closes [#3602](https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/3602)

Adds `/v0/voas/systems/:system_id/institutions` which returns institution data for one or more clinic ids passed as url params.

## Testing done
unit tests

## Testing planned
staging integration tests

#### Unique to this PR
- [ ] Adds `/v0/voas/systems/:system_id/institutions`

#### Applies to all PRs

- [ ] Appropriate logging


modules/vaos/app/controllers/vaos/institutions_controller.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module VAOS
+  class InstitutionsController < VAOS::BaseController
```
jbalboni: I'm not sure how I feel about "institution" as a description. This service is returning data about a clinic, which is at best an office within a facility, and sometimes just an abstract bucket for scheduling purposes. In other places "institutionCode" refers to a code or id for a facility, not a clinic, and similarly, in the data we get back in the array for this, the fields prefixed with institution refer to the facility the clinic is associated with.

modules/vaos/app/services/vaos/systems_service.rb
```diff
@@ -109,11 +109,24 @@ def get_facility_visits(system_id, facility_id, type_of_care_id, schedule_type)
           'institution-code' => facility_id,
           'clinical-service' => type_of_care_id
         }
-        response = perform(:get, url, url_params, headers(@user))
+        response = perform(:get, url, url_params)
```
saneshark: is headers not needed for this request? howcome?

kreek: good catch, updated

kreek:
@jbalboni @saneshark updated the path to be `/systems/:system_id/clinics`

modules/vaos/app/serializers/vaos/institution_serializer.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+require 'fast_jsonapi'
+
+module VAOS
+  class InstitutionSerializer
+    include FastJsonapi::ObjectSerializer
+
+    set_id :institution_ien
```
jbalboni: I don't think this will always be unique, if that matters here. `location_ien` should be.

PR #3774, opened 2020-01-13
Ensure that the test environment is not using a real rails log

johnpaulashenfelter:
## Description of change

There are a lot of logs written when running the specs, which are rarely useful. This
takes CPU and diskIO (since we're logging to file in the test environment). Occasionally in local
testing I've seen 

```
[42292:SemanticLogger::Appenders] SemanticLogger::Appenders -- Async: Appender thread has fallen behind by 61431867.117476 seconds with 1 messages queued up. Consider reducing the log level or changing the appenders
```

so I thought I'd try removing logging in the environment.

## Testing done

After 5 test runs, I am no longer receiving that error (almost always received it previously)

Test runs are also now faster

* ~7.25m before
* ~6.25m After

That's ~15% savings in time. I'm verifying that on CI`.

Hat tip to https://jtway.co/speed-up-your-rails-test-suite-by-6-in-1-line-13fedb869ec4

## Testing planned

Testing time savings on CI

~One test fails, mostly because its testing the wrong thing. Talking to the
author and will likely rewrite that spec to be more accurate.~

#### Unique to this PR
- [ ] Add an ENV var to toggle back on in rare case its necessary

#### Applies to all PRs

- [x] Appropriate logging

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -43,6 +43,8 @@ Rails.application.configure do
   # Print deprecation notices to the stderr.
   config.active_support.deprecation = :stderr
 
+  config.logger = Logger.new(nil)
+  config.log_level = :fatal
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
 end
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -43,8 +43,11 @@ Rails.application.configure do
   # Print deprecation notices to the stderr.
   config.active_support.deprecation = :stderr
 
-  config.logger = Logger.new(nil)
-  config.log_level = :fatal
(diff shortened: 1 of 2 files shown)


johnpaulashenfelter:
Turns out the spec is really testing whether `log.info` is completed on the last step of the `perform` method -- not whether it was performed properly. I think we can just write a more accurate spec.

Thoughts @dhall? (wrote the original spec a looooong time ago)

dhall:
> Thoughts @dhall? (wrote the original spec a looooong time ago)

If I'm tracking, it sounds like the spec is relying on the return value of the log method.. I'm fine with adding a return value to the perform method or updating the spec to test for success another way.

PR #3731, opened 2020-01-03
Lighthouse - Appeals API - add v1 docs

erluti:
## Description of change
https://app.zenhub.com/workspaces/developervagov-5ac272fff41bfa50c0e46c48/issues/department-of-veterans-affairs/vets-contrib/3969
New Appeals API work was originally in Caseflow.  The appeal team will continue to build out the API within Caseflow, but now we are beginning to build the access to it in the `appeals_api` module within vets-api.  The first task was to move the api documentation and set up framework for new v1 access to the Caseflow API being built.

## Testing done
Added route testing for new v1 docs controller.

PR #3775, opened 2020-01-13
get contestable issues through lighthouse

edmangimelli:
## Description of change
Resolves https://app.zenhub.com/workspaces/developervagov-5ac272fff41bfa50c0e46c48/issues/department-of-veterans-affairs/vets-contrib/3971

Picking up where https://github.com/department-of-veterans-affairs/vets-api/pull/3731 leaves off, this PR continues building the lighthouse layer of the Appeals API (instead of Kong sending requests directly to Caseflow, Kong will now (soon) forward requests to vets-api controllers, which will then call Caseflow).
This PR, specifically, adds appeals service method `get_contestable_issues` and a contestable issues controller.

## Acceptance Criteria
- [x] Tests
- [x] Appropriate logging -- using `with_monitoring` in service class
- [ ] ~Swagger docs have been updated, if applicable~ n/a
- [x] Does not contain any sensitive information (i.e. PII/credentials/internal URLs/etc., in logging, hardcoded, or in specs) --using dummy Caseflow UAT ssn (pulled from existing VCR cassette)


modules/appeals_api/app/controllers/appeals_api/application_controller.rb
```diff
@@ -3,5 +3,10 @@
 module AppealsApi
   class ApplicationController < ::ApplicationController
     skip_before_action :set_tags_and_extra_context, raise: false
+
+    def render_response(resp)
+      response.headers = resp.headers
```
erluti: we should probably remove the lighthouse api key from the headers we pass along

edmangimelli: whoops! yes we should (although I believe I'll do it in the contestable_issues_controller or service.rb)

edmangimelli: or, it sounds like this is stripped out before it hits vets-api

lib/appeals/service.rb
```diff
@@ -26,11 +28,28 @@ class Service < Common::Client::Base
     #
     def get_appeals(user, additional_headers = {})
       with_monitoring do
-        response = perform(:get, '/api/v2/appeals', {}, request_headers(user, additional_headers))
+        response = perform(
+          :get,
+          '/api/v2/appeals',
```
erluti: Can we make a constant for this too while we're changing it around?

PR #3782, opened 2020-01-14
wip - filter endpoint prometheus label in breakers plugin

omgitsbillryan:
#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging

### DIFF
--- a/lib/breakers/statsd_plugin.rb
+++ b/lib/breakers/statsd_plugin.rb
@@ -5,22 +5,7 @@ module Breakers
     def get_tags(request)
       tags = []
       if request
-        if request.url&.path
-          # replace identifiers with 'xxx'
-          # this nasty-looking regex attempts to cover:
-          # * (possibly negative) digit identifiers
-          # * uuid's with or without dashes
-          # * institution id's of form 111A2222 or 11A22222
-          digit = /\-?\d+/
-          contact_id = /\d{10}V\d{6}(%5ENI%5E200M%5EUSVHA)*/
-          uuids = /[a-fA-F0-9]{8}(\-?[a-fA-F0-9]{4}){3}\-?[a-fA-F0-9]{12}/
-          institution_ids = /[\dA-Z]{8}/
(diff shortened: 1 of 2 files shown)


lib/breakers/statsd_plugin.rb
```diff
@@ -5,22 +5,7 @@ class StatsdPlugin
...
-          digit = /\-?\d+/
-          contact_id = /\d{10}V\d{6}(%5ENI%5E200M%5EUSVHA)*/
-          uuids = /[a-fA-F0-9]{8}(\-?[a-fA-F0-9]{4}){3}\-?[a-fA-F0-9]{12}/
-          institution_ids = /[\dA-Z]{8}/
-          provider_ids = /Providers\(\d{10}\)/
-          r = %r{(\/)(#{digit}|#{contact_id}|#{uuids}|#{institution_ids}|#{provider_ids})(\/|$)}
-          endpoint = request.url.path.gsub(r, '\1xxx\5')
-          tags.append("endpoint:#{endpoint}")
-        end
-
```
omgitsbillryan: The 2 new regex's i added broke the lint for maximum lines, so I moved this into a private method below.

spec/lib/breakers/statsd_plugin_spec.rb
```diff
@@ -53,6 +53,22 @@
...
+
+        request.url = URI(test_host + '/api/v1/users/00u2i1p1u8m7l7FYb297/grants')
+        expect(subject.get_tags(request)).to include('endpoint:/api/v1/users/xxx/grants')
+      end
+    end
+
+    context 'request without an id' do
+      it 'doesnt replace anything' do
+        request.url = URI(test_host + '/foo/bar')
+        expect(subject.get_tags(request)).to include('endpoint:/foo/bar')
```
omgitsbillryan: Didn't need to do this, but it seems like a negative test case is appropriate to make sure that we dont accidentally filter an endpoint that we shouldn't.

PR #3784, opened 2020-01-14
filter dynamic ids from prometheus label

omgitsbillryan:
## Description of change
Connects: https://github.com/department-of-veterans-affairs/va.gov-team/issues/4854

- adds 2 regex's to filter for `/api/v1/users/<DYNAMIC_ID>` for the `endpoint` label of the `api_external_http_request_duration_seconds` metric
- slight refactor of `statsd_plugin.rb` because the new lines went over rubocops max lines for method
- adds 1 unit test for ensuring that our regex's don't wrongly filter a normal endpoint

## Testing done
Specs only.

## Testing planned
None.

#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging


lib/breakers/statsd_plugin.rb
```diff
@@ -5,22 +5,7 @@ class StatsdPlugin
...
-          digit = /\-?\d+/
-          contact_id = /\d{10}V\d{6}(%5ENI%5E200M%5EUSVHA)*/
-          uuids = /[a-fA-F0-9]{8}(\-?[a-fA-F0-9]{4}){3}\-?[a-fA-F0-9]{12}/
-          institution_ids = /[\dA-Z]{8}/
-          provider_ids = /Providers\(\d{10}\)/
-          r = %r{(\/)(#{digit}|#{contact_id}|#{uuids}|#{institution_ids}|#{provider_ids})(\/|$)}
-          endpoint = request.url.path.gsub(r, '\1xxx\5')
-          tags.append("endpoint:#{endpoint}")
-        end
-
```
omgitsbillryan: I had to refactor this into it's own method because my 2 additional regex's broke the maximum method length rubocop.

lib/breakers/statsd_plugin.rb
```diff
@@ -46,5 +31,30 @@ def send_metric(status, service, request_env, response_env)
...
+      institution_ids = /[\dA-Z]{8}/
+      provider_ids = /Providers\(\d{10}\)/
+      v1_user_ids_1 = /[0-9a-f]{22}$/ # e.g. /api/v1/users/bb9c0f499977be68611151
+      v1_user_ids_2 = /00u2[0-9a-zA-Z]{16}/ # e.g. /api/v1/users/00u2gskfa6kXUvU0N792
+
+      r = %r{
+        (\/)
+        (#{digit}|#{contact_id}|#{uuids}|#{institution_ids}|#{provider_ids}|#{v1_user_ids_1}|#{v1_user_ids_2})
+        (\/|$)
+      }x
```
omgitsbillryan: Had to refactor this regex to be multi-line because adding `|#{v1_user_ids_1}|#{v1_user_ids_2}` broke the rubocop line length.

spec/lib/breakers/statsd_plugin_spec.rb
```diff
@@ -53,6 +53,22 @@
...
+
+        request.url = URI(test_host + '/api/v1/users/00u2i1p2u2m3l7FYb712/grants')
+        expect(subject.get_tags(request)).to include('endpoint:/api/v1/users/xxx/grants')
+      end
+    end
+
+    context 'request without an id' do
+      it 'doesnt replace anything' do
+        request.url = URI(test_host + '/foo/bar')
+        expect(subject.get_tags(request)).to include('endpoint:/foo/bar')
```
omgitsbillryan: Added a "negative case" spec. Not specific to this PR, but I wanted to be sure that I wasnt blowing away "normal" url paths that _didnt_ contain dynamic ID's.

lib/breakers/statsd_plugin.rb
omgitsbillryan: I could use a 2nd pair of eyes... are these two equivelant?
```ruby
r = %r{(\/)(/foo/|/bar/})(\/|$)}
```
I changed it to:
```ruby
r = %r{
  (\/)
  (/foo/|/bar/)
  (\/|$)
}x
```

lib/breakers/statsd_plugin.rb
```diff
@@ -46,5 +31,30 @@ def send_metric(status, service, request_env, response_env)
...
+      # this nasty-looking regex attempts to cover:
+      # * (possibly negative) digit identifiers
+      # * uuid's with or without dashes
+      # * institution id's of form 111A2222 or 11A22222
+      digit = /\-?\d+/
+      contact_id = /\d{10}V\d{6}(%5ENI%5E200M%5EUSVHA)*/
+      uuids = /[a-fA-F0-9]{8}(\-?[a-fA-F0-9]{4}){3}\-?[a-fA-F0-9]{12}/
+      institution_ids = /[\dA-Z]{8}/
+      provider_ids = /Providers\(\d{10}\)/
+      v1_user_ids_1 = /[0-9a-f]{22}$/ # e.g. /api/v1/users/bb9c0f499977be68611151
```
annaswims: are we confident that these are the two types of id's or should we just go for something like "everything after `/api/v1/users/`"?

lib/breakers/statsd_plugin.rb
johnpaulashenfelter: why not just leave it on one line and add to the rubocop ignore?

johnpaulashenfelter:
@omgitsbillryan what does success look like inside Prometheus? How do we know it worked and/or failed?

lib/breakers/statsd_plugin.rb
omgitsbillryan: Eh, reasonably confident. I [ran this query](http://prometheus-prod.vetsgov-internal:9090/prometheus/graph?g0.range_input=1h&g0.expr=count%20by%20(endpoint)%20(api_external_http_request_duration_seconds%7Bendpoint%3D~%22%2Fapi%2Fv1%2Fusers%2F.%2B%22%7D)&g0.tab=1), then copy+pasted the output to my text editor, cleaned up the output and "eyeballed" it. At the worst, I've missed some and we'll see them in grafana.

I tried following how other regex's were written in this file. OTOH, I do like how blunt / least-surprise just filtering everything after `/api/v1/users/` is - the only problem is that [some urls took the form `/api/v1/users/abcd1234..../grants`](https://github.com/department-of-veterans-affairs/vets-api/pull/3784/files#diff-18132ff30b1bdbf5e33dc76464095e08R64).

I'm 🤷‍♂ and could go either way.

lib/breakers/statsd_plugin.rb
omgitsbillryan: I suppose you could make the case for ignore since fixing could introduce a bug, but we do have a pretty good spec that makes me confident that I didn't. This line is pretty unreadable, so I think a fix is warranted.

omgitsbillryan:
@johnpaulashenfelter one of two ways:

1. This list will get much shorter: http://prometheus-prod.vetsgov-internal:9090/prometheus/graph?g0.range_input=1h&g0.expr=count%20by%20(endpoint)%20(api_external_http_request_duration_seconds%7Bendpoint%3D~%22%2Fapi%2Fv1%2Fusers%2F.%2B%22%7D)&g0.tab=1
2. In the [Breakers Grafana dashboard](http://grafana.vfs.va.gov/d/000000022/breakers?orgId=1&var-data_source=Prometheus%20(Production)&var-service=All&var-endpoint=All), the `endpoint` dropdown will get shorter.

PR #3787, opened 2020-01-15
Add optional fields to query Veteran Confirmation

rtravitz:
## Description of change

This PR addresses [#4047](https://github.com/department-of-veterans-affairs/vets-contrib/issues/4047). 

### Notable changes
- Add two optional fields—`middle_name` and `gender`—to query the Veteran Confirmation `/status` endpoint. Having these optional fields present increases the likelihood of a successful MVI query for a given person of interest. 
- Update documentation to include the optional fields

## Testing done

- Added both fields to the happy path payload for the request spec
- Add tests to validate the incoming gender field to comply with the enum that MVI expects

#### Unique to this PR

- [x] API is updated to allow `Gender` and `Middle Name` as optional parameters
- [x] Add and test validations for new optional params
- [x] API docs are updated on the developer portal to reflect this change

#### Applies to all PRs
- [x] Appropriate logging

### DIFF
--- a/modules/veteran_confirmation/VETERAN_CONFIRMATION.yml
+++ b/modules/veteran_confirmation/VETERAN_CONFIRMATION.yml
@@ -80,7 +80,6 @@ paths:
       security:
         - apikey: []
       requestBody:
-        description: ''
         required: true
         content:
           application/json:
--- a/modules/veteran_confirmation/VETERAN_CONFIRMATION.yml
+++ b/modules/veteran_confirmation/VETERAN_CONFIRMATION.yml
@@ -29,13 +29,13 @@ info:
     ## Authorization
     This API requires an API key in combination with identifiable information for the person being confirmed listed below. API requests are authorized through a symmetric API token provided in an HTTP header with name `apikey`. Including more information has a better chance of making a match and returning a Confirmed status. 
     #### Required information:
-    * First Name  
-    * Last Name  
-    * Date of Birth  
+    * First Name 
+    * Last Name 
+    * Date of Birth 
     * Social Security Number 
 
     #### Optional information:
-    * Middle Name  
+    * Middle Name
     * Gender
     
     ## Reference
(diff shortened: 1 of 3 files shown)


jeff2d2:
Looked through the yml and it is exactly how I expected. From a product perspective, this is 💯.

modules/veteran_confirmation/VETERAN_CONFIRMATION.yml
```diff
@@ -29,13 +29,13 @@ info:
...
     #### Required information:
-    * First Name
 
-    * Last Name
 
-    * Date of Birth
 
+    * First Name 
+    * Last Name 
+    * Date of Birth
```
nfasulo: Any idea what git is up to here?

modules/veteran_confirmation/app/controllers/veteran_confirmation/v0/veteran_status_controller.rb
```diff
@@ -43,6 +46,13 @@ def valid_ssn?(ssn)
         ssn.is_a?(String) && (all_digits?(ssn) || all_digits_with_hyphens?(ssn))
       end
 
+      def validate_gender
+        gender_options = %w(M F)
+        no_matching_option = params['gender'] && !gender_options.include?(params['gender'])
```
edmkitty: Shall we upcase `params['gender']` in the cases of 'm' and 'f'?

modules/veteran_confirmation/VETERAN_CONFIRMATION.yml
jeff2d2: I see an extra symbol at the end of these lines on master, probably leftovers from copy/pasting from Figma. Thanks for taking care of those @rtravitz!

rtravitz: Removed some hidden characters

modules/veteran_confirmation/app/controllers/veteran_confirmation/v0/veteran_status_controller.rb
rtravitz: Good call. I'll add in the upcasing.

PR #3803, opened 2020-01-20
Contact info endpoint change

lihanli:
## Description of change
https://github.com/department-of-veterans-affairs/va.gov-team/issues/2483
update contact info endpoint to new path. current path is already being forwarded to the new path and will be deprecated in the future

## Testing done
tested in staging

## Testing planned
test in production

#### Applies to all PRs

- [x] Appropriate logging


spec/models/async_transaction/vet360/base_spec.rb
```diff
@@ -52,7 +52,7 @@
     end
 
     it 'persists the messages from vet360' do
-      VCR.use_cassette('vet360/contact_information/email_transaction_status_w_message') do
```
jholton: I don't think this cassette is used anywhere now.  Can we delete it?

lihanli: sounds good, i deleted it

PR #3805, opened 2020-01-21
Add NCA websites

nfasulo:
## Description of change
ArcGIS has a field for website url for NCA, which currently contains no data. We received a list of websites from NCA so that they can be included in the Facility Locator. These have been added to our website CSV for loading into our facility cache.

The data received from NCA does not include a leading `0` to make all of the station numbers three digits (e.g. `042` is `42`). Instead of modifying this in the CSV (and then subsequently having to remember to do it again when we receive updates), I opted to strip the leading `0` from a facility's station number before trying to match it to the CSV. I made this change in `NCAFacility` instead of the `WebsiteURLService` in an attempt to be explicit about what records need this modification and avoid future side effects.

For https://github.com/department-of-veterans-affairs/vets-contrib/issues/4064

## Testing done
Added test to NCA facilities tests

## Testing planned
Will verify that NCA websites are available in dev after this is merged

#### Unique to this PR
- [x] NCA Facilities now have website urls

#### Applies to all PRs

- [x] Appropriate logging


nfasulo:
@rtravitz I failed to move the instantiation of new `NCAFacility` objects to after the website is pulled from `WebsiteURLService`. This meant that while the code was passing the test (because the website was part of the object in memory it was not being updated in the database. I see 172 NCA facilities with `website` filled in now.

PR #3807, opened 2020-01-21
fail out for hca attachments that were uploaded in error

omgitsbillryan:
## Description of change
Trying to help out veterans who encountered a bug introduced when we allowed null files to be uploaded [link to PR](https://github.com/department-of-veterans-affairs/vets-api/pull/3721)

#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging

### DIFF
--- a/app/workers/hca/submission_job.rb
+++ b/app/workers/hca/submission_job.rb
@@ -3,7 +3,6 @@
 module HCA
   class SubmissionJob
     include Sidekiq::Worker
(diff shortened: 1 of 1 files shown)


omgitsbillryan:
It turned out to be pretty difficult to reproduce this specific exception since we use carrierwave and not the aws s3 sdk directly. I'm going to close this pr unmerged because 

1. Each user who was unlucky enough to upload their attachments while the original bug was live in prod has now received their failure email notification.
2. I don't want to deploy any untested code into prod.

PR #3809, opened 2020-01-22
Enable Sidekiq's `super_fetch` feature

jholton:
## Description of change
I've [researched/documented](https://github.com/department-of-veterans-affairs/va.gov-team/issues/4976) occurrences of Sidekiq processes getting killed due to "out of memory" occurrences.  When this happens, these jobs are lost 

Enabling Sidekiq's `super_fetch` feature will help ensure that if a thread/process crashes the job it was processing isn't lost.

[`super_fetch` documentation](https://github.com/mperham/sidekiq/wiki/Reliability#using-super_fetch)

#### Unique to this PR

- [x] enable `super_fetch!` in Sidekiq config for `production` environments

#### Applies to all PRs

- [ ] Appropriate logging


jholton:
@department-of-veterans-affairs/vsp-operations Any concerns with enabling this?

johnpaulashenfelter:
@jholton all for trying this, but worth knowing about the latency: https://gitter.im/mperham/sidekiq?at=5bbca3fdbbdc0b25050ec832

Do we have an obvious dashboard (or can someone from ops help with) something where we can track the latency of our sidekiq requests?

config/initializers/sidekiq.rb
```diff
@@ -4,6 +4,8 @@
 
 Sidekiq.configure_server do |config|
   config.redis = REDIS_CONFIG['redis']
+  config.super_fetch! if Rails.env.production?
```
omgitsbillryan:
```suggestion
  config.super_fetch! unless Settings&.sidekiq&.disable_super_fetch
```

I think'd it'd be best to enable this for local environments by default, and then have a mechanism to disable it if a developer wishes to do so.

johnpaulashenfelter: I really don't like the &.&.&. pattern -- its an error if we expect a setting and its not there

omgitsbillryan: I just remembered that [we already have the `EXCLUDE_SIDEKIQ_ENTERPRISE` env var](https://github.com/department-of-veterans-affairs/vets-api/blob/master/Gemfile#L173), so maybe we could re-write to:

```ruby
config.super_fetch! unless ENV['EXCLUDE_SIDEKIQ_ENTERPRISE'] == 'true'
```

jholton:
> does this mean that it might be possible for a successful job run could get interrupted before it's pulled off the job queue, to where, it might get re-run later on?

A little more info:
`super_fetch` uses Redis' `RPOPLPUSH` command:

> the consumer fetches the message and at the same time pushes it into a processing list. It will use the LREM command in order to remove the message from the processing list once the message has been processed.

> An additional client may monitor the processing list for items that remain there for too much time, and will push those timed out items into the queue again if needed.

So there's technically another "processing" queue used here.

As to your concern of a job being run twice, I think Sidekiq best practices would say "Your jobs should be idempotent"
Certainly not all of ours are, so your concern is valid.  I wonder what safeguards would be possible.  (Wrapping code in transactions, setting flags in a DB to mark certain blocks of code as done, etc)

omgitsbillryan:
"Your jobs should be idempotent" is sound advice. This still LGTM.

PR #3811, opened 2020-01-23
GIBCT Search results not showing #4280

zurbergram:
## Description of change
https://github.com/department-of-veterans-affairs/va.gov-team/issues/4280

The search results are not displaying when searching for a school on staging. Instead, a user is only able to see the "Loading search results" animation. This issue is occurring in all environments.

Navigate to https://va.gov/gi-bill-comparison-tool
Type "College of Charleston" into the search bar
Click Search schools

Setting timeout to the default that is defined at https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/common/client/configuration/base.rb#L38 if there is not a settings value

#### Unique to this PR
Requires https://github.com/department-of-veterans-affairs/devops/pull/6282 to be merged first so as to not negatively impact production environment

#### Applies to all PRs

- [ ] Appropriate logging


annaswims:
For the stability of Vets-api we cannot accept a 15s timeout for a request that should take 300ms or less. The problem isn’t that vets-api doesn’t wait long enough for a response, the problem is that GIDS performance.

zurbergram:
> For the stability of Vets-api we cannot accept a 15s timeout for a request that should take 300ms or less. The problem isn’t that vets-api doesn’t wait long enough for a response, the problem is that GIDS performance.

The default timeout is 15 seconds, additionally most other configurations have a timeout of at least 15 seconds, some are 30 seconds.

https://github.com/department-of-veterans-affairs/devops/pull/6282 will set the timeout for production to 2 seconds and be merged first. This value is based on http://grafana.vfs.va.gov/d/000000050/backend-service-report?orgId=1&from=1579669200000&to=1579755599000 

The GIDS issue that warranted the original changes was fixed in https://github.com/department-of-veterans-affairs/gibct-data-service/pull/507.  

Currently there are issues with timeouts occurring in dev, staging, and production due to the timeout either being set at 1 second or defaulting to 1 second.  This 1 second default timeout is related to the issue being discussed around https://dsva.slack.com/archives/C30LCU8S3/p1579726768127200

annaswims:
I'd suggest you routinely run the GIDS load tests as you make substantial changes.  Over the past month, we've seen a substantial uptick in [latency to the /gids/v0/institutions endpoint](http://grafana.vfs.va.gov/d/000000050/backend-service-report?orgId=1&from=now-30d&to=now&fullscreen&panelId=121).
Screen Shot 2020-01-23 at 3 40 15 PM 

Do you look at the caching PR independently to see if it helped/harmed?

PR #3814, opened 2020-01-23
Use puma executable to start web application instead of bin/rails

johnpaulashenfelter:
## Description of change

We have been using `bin/rails s` to start the rails server through foreman for quite some time, which is atypical. Typically the puma executable is used to start
the application (https://github.com/puma/puma#frameworks) so that its easier to configure. 

This wasn't a problem until we tried to use clustered mode to better use the capacity of our EC2 instances. Even running with a single worker in clustered mode,
no logfiles would be generated. Our production site experienced HUGE (2x+) memory usage and clear memory leakage that is 100% correlated with running in clustered
mode. Switching back to single mode in puma (PUMA_WORKERS=0) clearly put memory usage back in the normal range and restarted logfile collection. The "no logs"
problem was reproduced on staging and in local development, though no tests were run to check the memory usage.

@nathanruby did some initial work that indicates the issue is related to puma forking the process and handling stdout improperly.

Without digging deeply into the detailed "why" of the issue, the more standard configuration of using `puma` as the web runner in the Procfile definitely fixes the
logging issue. I expect that the memory issue will be fixed as well which will be verified in staging and production

## Testing done

- [x] Verify logs work in development
- [x] Verify memory usage is back to normal in development

## Testing planned

- [ ] Verify logs work in staging
- [ ] Verify memory usage is back to normal in staging
- [ ] Verify logs work in production
- [ ] Verify memory usage is back to normal in production

#### Unique to this PR

- [ ] Memory usage remains ~constant with PUMA_WORKERS=1 in staging and production

#### Applies to all PRs

- [ ] Appropriate logging

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,4 +1,4 @@
-web: bundle exec bin/rails s -b 0.0.0.0 -p 3000
+web: bundle exec puma -p 3000 -C ./config/puma.rb
 job: bundle exec sidekiq -q default -q critical -q tasker
 freshclam: /usr/bin/freshclam -d --config-file=config/freshclam.conf
 clamd: /usr/sbin/clamd -c config/clamd.conf


omgitsbillryan:
My first thought looking at this, is that we don't make use of this `procfile` in our actual dev/staging/prod deployments. That said, @nathanhruby might know more than I on this subject since he just converted all our vets-api deploys to use docker.

omgitsbillryan:
I think we might be using an upstart config here: https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/build/roles/vets-api/tasks/main.yml#L173

johnpaulashenfelter:
@omgitsbillryan you're even more OOL -- we're doing *BOTH* as i understand it 😄

PR #3820, opened 2020-01-27
Supplemented source app mapping with matching of root URLs

U-DON:
## Description of change
One of the existing issues with the source app mapping is that many of the apps are reporting `not_provided`, in part due to certain API calls from vets-website not supplying the header. To address those cases, we can look at the referer (the URL the request came from) and match that to the root URL of the corresponding app.

## Testing done
Unit tests. Simple local test where I checked that referer was used when source app name header wasn't sent.

## Testing planned
Will monitor metrics for the mapping in Grafana after merging.

#### Applies to all PRs

- [ ] Appropriate logging


lib/statsd_middleware.rb
```diff
@@ -95,6 +107,18 @@ def call(env)
...
+    # match the referrer to the root URL of one of the whitelisted apps.
+    if source_app.nil?
+      referer = env['HTTP_REFERER']
+
+      if referer
+        *, source_app = SOURCE_APP_URL_MAP.find do |root_url, _|
+          URI.parse(referer).path.start_with? root_url
+        end
+      end
+    end
```
omgitsbillryan: I'm a little concerned at the performance of this code since it runs on every single request. One easy thing is to move 

```ruby
path = URI.parse(referer).path
```

outside the enumeration.

omgitsbillryan: Maybe some testing is in order to see how much time iterating over our entire `SOURCE_APP_URL_MAP` adds.

U-DON: It may run a lot from apps where the source_app isn't being sent as a header, but this will be avoided when it is provided.

U-DON:
There are apps using a different method of making requests to the APIs than the common helper function that had the change to include the header, but it seems there may be fewer of them than I thought. My original thinking was that this would catch everything we missed, since it might take a bit of lift to update all the apps to use that common helper.

Since the number of apps that need to be updated seems manageable, I can add headers to those cases and revisit this if there are still too many unaccounted `not_provided` reports.

PR #3824, opened 2020-01-29
Require fields on veteran rep search

charleystran:
## Description of change
This PR adds the proper error handling for when the rep search endpoint is called without first name and last name
## Testing done
Local testing

- [ ]  when no first or last name params are in the api call, a 422 is returned with a message

#### Applies to all PRs

- [ ] Appropriate logging

resolve https://github.com/department-of-veterans-affairs/vets-contrib/issues/4110


edmangimelli:
LGTM 🎉

modules/veteran/app/controllers/veteran/v0/representatives_controller.rb
```diff
@@ -17,6 +18,18 @@ def find_rep
...
         end
       end
+
+      private
+
+      def check_required_fields
+        if params[:first_name].blank? || params[:last_name].blank?
+          render json: {
+            errors: [
+              details: 'First name and Last name are required to complete this request', status: 422
```
edmangimelli: do you both default to `details` for errors? (over title)

bastosmichael: It's technically detail (no s) and it's part of the JSON API spec error response https://jsonapi.org/format/#errors

edmangimelli: we fell into the habit of using `title` for general messages, and `details` when the message included unique info (to the request) (that's how we interpreted jsonapi)

charleystran: @bastosmichael this has been updated

edmangimelli: still looks like `details` to me

edmangimelli: do we never use `title` or `code`?

charleystran: if you go into files_changed it should show as detail:

edmangimelli: weird. that's where i thought i was looking 🤷‍♀

charleystran:
@bastosmichael I updated the error hash to include the extra fields

modules/veteran/app/controllers/veteran/v0/representatives_controller.rb
```diff
@@ -17,6 +18,24 @@ def find_rep
...
+      def check_required_fields
+        errors = []
+        errors << error_hash('first_name') if params[:first_name].blank?
+        errors << error_hash('last_name') if params[:last_name].blank?
+        render json: { errors: errors }, status: 422 if errors.any?
+      end
+
+      def error_hash(parameter)
+        {
+          detail: "#{parameter.humanize} are required to complete this request",
```
edmangimelli: should `are` be `is`?

edmangimelli:
does this resolve https://github.com/department-of-veterans-affairs/vets-contrib/issues/4110
?

bastosmichael:
yes @edmangimelli this does resolve that ticket, I usually put the Resolves in the description myself but @charleystran sometimes forgets to do so.

modules/veteran/app/controllers/veteran/v0/representatives_controller.rb
bastosmichael: Good catch yes it should

charleystran: This has been updated

PR #3826, opened 2020-01-29
Veteran Confirmation API Documentation Updates

rtravitz:
## Description of change

This PR addresses [#4114](https://github.com/department-of-veterans-affairs/vets-contrib/issues/4114). It updates the Open API Spec for the Lighthouse Veteran Confirmation API to include information about Test Users. That information is needed for an API consumer to get a successful `"confirmed"` response from the API in lower environments.

## Testing done

- Validated the yaml for the Open API Spec

#### Unique to this PR

- [x] Documentation for the Veteran Confirmation API indicates that Test User attributes should be used in lower environments and where those users can be found

#### Applies to all PRs
- [x] Appropriate logging

### DIFF
--- a/modules/veteran_confirmation/VETERAN_CONFIRMATION.yml
+++ b/modules/veteran_confirmation/VETERAN_CONFIRMATION.yml
@@ -15,11 +15,14 @@ info:
 
     ### 2. Test the API
     In the endpoint documentation below, we've provided a curl command builder for trying out the API before implementation with your app.
+    Use [Test User](https://github.com/department-of-veterans-affairs/vets-api-clients/blob/master/test_accounts.md) attributes to populate the curl.
 
     ### 3. Build your app
     The base URI for the Veteran Confirmation API in the Development environment is:
     
     https://dev-api.va.gov/services/veteran_confirmation/v0
+
+    In this environment, use attributes from the list of [Test Users](https://github.com/department-of-veterans-affairs/vets-api-clients/blob/master/test_accounts.md). Only Test Users can return a `"confirmed"` response.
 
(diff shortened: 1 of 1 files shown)


modules/veteran_confirmation/VETERAN_CONFIRMATION.yml
```diff
@@ -15,12 +15,15 @@ info:
 
     ### 2. Test the API
     In the endpoint documentation below, we've provided a curl command builder for trying out the API before implementation with your app.
+    Use [Test User](https://github.com/department-of-veterans-affairs/vets-api-clients/blob/master/test_accounts.md) attributes to populate the curl.
```
kalilsn:
```suggestion
    Use [Test User](https://github.com/department-of-veterans-affairs/vets-api-clients/blob/master/test_accounts.md) attributes to populate the request body.
```

I think we have been overloading `curl` in a lot of places (it's both the tool and the command line invocation itself etc.). Also, "Request Body" is the header that actually shows where these attributes actually go.

modules/veteran_confirmation/VETERAN_CONFIRMATION.yml
```diff
@@ -15,12 +15,15 @@ info:
...
 
     ### 3. Build your app
     The base URI for the Veteran Confirmation API in the Development environment is:
     
     https://dev-api.va.gov/services/veteran_confirmation/v0
 
+    In this environment, use attributes from the list of [Test Users](https://github.com/department-of-veterans-affairs/vets-api-clients/blob/master/test_accounts.md). Only Test Users can return a `"confirmed"` response.
```
kalilsn: Do you think it's worth mentioning that some of them may not return confirmed? I don't really know what the proportion is, so maybe that's not necessary at all!

rtravitz: I think a larger recognition of that is probably necessary but am going to ignore it for this update.

PR #3830, opened 2020-01-30
review instance docker-compose file for vets-api

wyattwalter:
## Description of change

Adds a new docker-compose file for review instances. A similar change was made in vets-website this week as well.

I chose to make it a separate docker-compose file and not deal with consolidating them because the intent is to move to a container orchestrator soon and this will keep everything we need for that transition in one place.

ref: https://github.com/department-of-veterans-affairs/va.gov-team/issues/5238

## Testing done

Tested with a deploy from this branch: https://github.com/department-of-veterans-affairs/devops/pull/6303

#### Applies to all PRs

- [x] Appropriate logging


docker-compose.review.yml
```diff
@@ -0,0 +1,53 @@
...
+      POSTGRES_USER: "${POSTGRES_USER:-postgres}"
+    volumes:
+      - db-data:/var/lib/postgresql/data:cached
+    ports:
+      - "54320:5432"
+    restart: unless-stopped
+  vets-api:
+    build:
+      context: .
+      target: development
```
omgitsbillryan: 👍 👍 vets-api has a number of `Rails.env` checks in it, but I dont think any that we cant work through. Happy to see this.

docker-compose.review.yml
```diff
@@ -0,0 +1,53 @@
...
+      args:
+        exclude_sidekiq_ent: "${EXCLUDE_SIDEKIQ_ENTERPRISE:-false}"
+        sidekiq_license: "${BUNDLE_ENTERPRISE__CONTRIBSYS__COM}"
+        userid: "${VETS_API_USER_ID}"
+    command: >
+      bash -c "bundle exec rake db:migrate || bundle exec rake db:setup db:migrate
```
omgitsbillryan: Is the idea here to only run a setup on first deploy and migrations on PR commits thereafter?

PR #3833, opened 2020-01-31
add review slug back to saml relay state

omgitsbillryan:
## Description of change
Connects: https://github.com/department-of-veterans-affairs/va.gov-team/issues/5090

Step 1 for re-enabling login on review instances. This PR:

1. Adds `review_instance_slug ` back into the RelayState of the AuthNRequest.
2. Allow the after login success relay to be overrided with a settings config so that review instances can be configured to do so.

## Testing done
Local testing.

#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging


lib/saml/url_service.rb
```diff
@@ -39,7 +39,7 @@ def initialize(saml_settings, session: nil, user: nil, params: {})
 
     # REDIRECT_URLS
     def base_redirect_url
-      VIRTUAL_HOST_MAPPINGS[current_host][:base_redirect]
+      Settings.saml.relay || VIRTUAL_HOST_MAPPINGS[current_host][:base_redirect]
```
jholton: 1) [Here](https://github.com/department-of-veterans-affairs/devops/blob/775cc72c84e13f5b93b8fcb3fc862bc514a3c30c/ansible/roles/review-instance-configure/vars/settings.local.yml#L103) is the `Settings.saml.relay` value (right?).
This being the `base_redirect_url`, shouldn't it be the base URL and not contain the `/auth/login/callback` path since that is provided by the `LOGIN_REDIRECT_PARTIAL` constant? 

2) Can this data be configured to be part of the `VIRTUAL_HOST_MAPPINGS` constant instead of using this `||` operator?  I know the chances are probably very low, but that would prevent this from breaking in the case where `Settings.saml.relay` is erroneously present in `prod`, `staging`, or `dev`

omgitsbillryan: 1. Yes. And yes! Definitely an oversight by me that I'll correct.
2. This is a good point about `Settings.saml.relay` being erroneously present - unfortunately I don't think it's possible since the base url will be unique for each review instance (e.g. `http://617bed45ccb1fc2a87872b567c721009-api.review.vetsgov-internal`), making it impossible to hardcode into a ruby file.

As a tangent, I don't think _any_ of the dev/stag/prod mappings should be hardcoded into a ruby file. I'd rather see each of these as config in their respective settings-env.yml files, which would eliminate the need for this `def base_redirect_url` method entirely.

jholton: Good point.  That whole constant is environment specific config.

omgitsbillryan:
@jholton once I added some specs, I realized my previous code wouldn't work since the callback url path must also change from:

- `/auth/saml/callback`

to:

- `/review_instance/saml/callback`

...by the end of this PR I will definitely be writing / updating some auth flow doc.

spec/lib/saml/url_service_spec.rb
```diff
@@ -5,20 +5,20 @@
...
-    end
+  around do |example|
+    User.create(user)
+    Timecop.freeze('2018-04-09T17:52:03Z')
+    RequestStore.store['request_id'] = '123'
+    example.run
+    Timecop.return
+  end
 
+  SAML::URLService::VIRTUAL_HOST_MAPPINGS.each do |vhost_url, values|
```
omgitsbillryan: Git is showing this diff weirdly... all I did was move the `let` and `around` declarations outside of the `SAML::URLService::VIRTUAL_HOST_MAPPINGS.each`, so that I could re-use them at the bottom of the file.

PR #3836, opened 2020-01-31
Use postgresql v11 in CI

johnpaulashenfelter:
#### Applies to all PRs


docker-compose.test.yml
```diff
@@ -3,7 +3,7 @@ services:
   redis:
     image: redis:3.2-alpine
   postgres:
-    image: mdillon/postgis:9.6-alpine
+    image: mdillon/postgis:11
```
jholton: Do we want `mdillon/postgis:11` or `mdillon/postgis:11-alpine`?

johnpaulashenfelter: Honestly i'm not THAT sure it matters. Production is AWS RDS so always going to be slightly different. Alpine is certainly smaller, so 👍

johnpaulashenfelter:
Verified specs pass on local Docker instance using `docker ci` and this configuration.

PR #3838, opened 2020-01-31
Adding new endpoint for BGS find dependents

kathleencrawford:
## Description of change
Adding a new API endpoint for the BGS findDependentsByPtcpntId method, which returns a list of dependents associated with a veteran.

Original ticket here - https://github.com/department-of-veterans-affairs/va.gov-team/issues/3762

## Testing done
Local and dev

## Testing planned
Staging

#### Applies to all PRs

- [x] Appropriate logging


app/services/bgs/dependent_service.rb
```diff
@@ -0,0 +1,14 @@
...
+
+module BGS
+  class DependentService
+    def get_dependents(current_user)
+      service = LighthouseBGS::Services.new(
+        external_uid: current_user.icn,
+        external_key: current_user.email
+      )
+
+      service.claimants.find_dependents_by_participant_id(current_user.participant_id, current_user.ssn)
```
johnpaulashenfelter: What happens if this fails?

kathleencrawford: There is a "rescue" in the controller that calls this method that will catch the error from the lighthouse_bgs service and log it to sentry.

app/controllers/v0/dependents_applications_controller.rb
```diff
@@ -21,7 +21,12 @@ def create
     end
 
     def show
-      render(json: DependentsApplication.find(params[:id]))
+      dependent_service = BGS::DependentService.new
+      response = dependent_service.get_dependents(current_user)
+      render json: response, each_serializer: DependentsSerializer
+    rescue => e
```
charleystran: Not sure if this should really rescue all exceptions, or just the BGS Share exception that is raised.

johnpaulashenfelter: I tend to agree

config/settings/test.yml
```diff
@@ -69,8 +69,13 @@ vba_documents:
 
 bgs:
   mock_response_location: /cache/bgs
-  mock_responses: true
+  mock_responses: false
```
johnpaulashenfelter: Shouldn't this be true in test? or is this to record a cassette?

charleystran: Mock false allows VCR to work. Mock true would be like Betamocks being on.

app/controllers/v0/dependents_applications_controller.rb
charleystran: @johnpaulashenfelter  since this is raised to sentry still, I am ok with the catch all, because it returns a properly formed response vs a 500 and the client having to handle that. What do you think? I am not against being specific either and having better errors.

PR #3853, opened 2020-02-03
[5337] Added MDOT Prefill Form

camerontesterman:
## Original Issue
[Issue 5337](https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/5337)

## Description of change
Provides prefill form data for the Medical Devices Ordering Tool form, which the veteran will use to verify that they are ordering for the correct person.

I also fixed the cognitive complexity complaint around the `FormProfile.prefill_enabled_forms` method.

## Testing done
Tested locally with mock user.

## Testing planned
None.

#### Unique to this PR
None.

#### Applies to all PRs

- [x] Appropriate logging


app/models/form_profile.rb
```diff
@@ -121,22 +125,11 @@ class FormProfile
...
-    forms += PENSION_BURIAL_FORMS if Settings.pension_burial.prefill
-    forms += EDU_FORMS if Settings.edu.prefill
-    forms += VIC_FORMS if Settings.vic.prefill
-    forms << '21-686C'
-    forms << '40-10007'
-    forms += EVSS_FORMS if Settings.evss.prefill
-    forms += HLR_FORMS if Settings.decision_review.prefill
-
+    forms = %w[21-686C 40-10007]
+    ALL_FORMS.each { |type, form_list| forms += form_list if Settings[type].prefill }
```
SMLuthi: The reason I actually didnt do this very thing recently when I added in the HLR forms (and opted for the CyclomaticComplexity enable/disable) because I thought it was an odd way of accessing settings. It makes it less explicit in how prefill options are being added. But its a personal preference... I just like things to be explicit instead of clever sometimes :) but thats not the ruby way!

Gemfile
```diff
@@ -103,7 +103,7 @@ gem 'swagger-blocks'
 gem 'typhoeus'
 gem 'upsert'
 gem 'utf8-cleaner'
-gem 'vets_json_schema', git: 'https://github.com/department-of-veterans-affairs/vets-json-schema', branch: 'master'
+gem 'vets_json_schema', git: 'https://github.com/department-of-veterans-affairs/vets-json-schema', tag: '3.140.0'
```
jholton: Can we reverse this?  Typically users don't create tags/releases.
Typically the way to update is the version is to just run `bundle update vets_json_schema`.
That'll get you the same result here (i.e. version 3.140.0 will be specified in Gemfile.lock)

PR #3854, opened 2020-02-03
Add HTTP into example code in docs

ncksllvn:
This little guy got me good today. I tried to run Swagger locally through Option 2, but it was missing the `http` part of the `web_origin` so I kept getting a CORS error. This PR fixes that.


annaswims:
So, @ncksllvn's change is an improvement, but it makes me wonder if maybe we should just put that setting in settings.yml rather than istructing everyone to DIY

jholton:
> So, @ncksllvn's change is an improvement, but it makes me wonder if maybe we should just put that setting in settings.yml rather than istructing everyone to DIY

Welp, [looks like we did](https://github.com/department-of-veterans-affairs/vets-api/blob/9780c207d89a9e2ae6497d84063860c913c41489/config/settings.yml#L5).

PR #3856, opened 2020-02-03
change ca cert loading in docker image for non-prod environments

wyattwalter:
## Description of change

Adds a config/ca-trust directory to the git repository and updates the Dockerfile to use that as a source for CA certs.

I want to make this change so that review instances can also have the CA certs of the image updated during the standard build process without going through any hoops. The primary reason that I put it here is that 1) I need it in the repository and with a non-empty dir (`COPY` can't be conditional and fails if there's nothing to copy) and 2) review instances aren't "prod" by their nature. Other considerations: I want to be able to drop these certs into a vets-api checkout and have it be clean from a git perspective and leave them in place so the image can be rebuild in place (again, to support review instance use case).

## Testing done

Tested extensively in the context of a review instance.

## Testing planned

Need to update the Ansible a bit for prod AMI builds to adjust the directory location since that changed and then test an AMI build.

#### Applies to all PRs

- [x] Appropriate logging - NA
- [x] Swagger docs have been updated, if applicable - NA
- [x] Provide link to originating GitHub issue, or connected to it via ZenHub - https://github.com/department-of-veterans-affairs/va.gov-team/issues/5239
- [x] Provide which alerts would indicate a problem with this functionality (if applicable) - this would result in vets-api being unable to reach backend services in dev and staging if it were to not work correctly


Dockerfile
```diff
@@ -79,6 +86,5 @@ ENV RAILS_ENV=production
 COPY --from=builder $BUNDLE_APP_CONFIG $BUNDLE_APP_CONFIG
 COPY --from=builder --chown=vets-api:vets-api /srv/vets-api/src ./
 COPY --from=builder --chown=vets-api:vets-api /srv/vets-api/clamav/database ../clamav/database
-RUN if [ -d certs-tmp ] ; then cd certs-tmp ; for i in * ; do cp $i /usr/local/share/ca-certificates/${i/pem/crt} ; done ; fi && update-ca-certificates
```
nathanhruby: certs-tmp is used in https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/build/roles/vets-api/tasks/main.yml#L90 to load in the host system ca's, so this removal will break that.

why not move this line to base stage as-is and have review instances follow the same process as linked above (make a temp dir, copy into it, mount into container during build, whack dir when done" ?

wyattwalter: Yeah, I know where this gets picked up and have a branch to update it, should we decide to do this.

I want review instances to be setup to run a build again outside of Ansible. So, ideally, I'd like to put them in a directory that's gitignore'd that can hang out indefinitely on those instances. I could just gitignore this particular path, just seemed nice to get it out of the root of the app if it was going to hang out like that.

wyattwalter: And actually, this won't work in base because we haven't copied in files from the app into the image yet. That only happens in the builder target and beyond.

nathanhruby: ahh yeah, totally valid.  This looks fine, tough I would ask that you leave this RUN statement in for now and remove it separately once the build side in devops is there.  That should reduce rollout complexity by not requiring coordination across two repos at the cost of some duplication here.

wyattwalter: Done!

docker-compose.review.yml
```diff
@@ -29,6 +29,7 @@ services:
       - .:/srv/vets-api/src:cached
       - dev_bundle:/usr/local/bundle
       - ../.secret:/srv/vets-api/secret:cached
+      - ../.pki:/srv/vets-api/pki:cached
```
omgitsbillryan: So this is temporary? Am I understanding that correctly?

Dockerfile
```diff
@@ -19,6 +19,13 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
 RUN mkdir -p /srv/vets-api/{clamav/database,pki/tls,secure,src} && \
     chown -R vets-api:vets-api /srv/vets-api && \
     ln -s /srv/vets-api/pki /etc/pki
+# XXX: get rid of this when we have a better model for it
+COPY config/ca-trust/* /usr/local/share/ca-certificates/
+# normalize file extension between pem/crt
```
omgitsbillryan:
```suggestion
# rename .pem files to .crt because update-ca-certificates ignores files that are not .crt
```

.gitignore
```diff
@@ -56,6 +56,8 @@ rspec.xml
 
 # Ignore certificates directory
 config/certs
+config/ca-trust
+!config/ca-trust/README.md
```
omgitsbillryan: Can i get a tl;dr of what ! does in gitignore? The documentation I found is a bit confusing.

We'd only want to ignore the `.pem` & `.crt` files, yes? In order to maintain a clean `git` state when developing on a review instance.

wyattwalter: I'm telling git to ignore that directory but exclude the README.md from that ignore ( gotta love a double-negative 😄 ). I'm ignoring everything there except that one file.

Yeah, I guess that would be another way to approach it. Either way is fine by me.

docker-compose.review.yml
wyattwalter: this particular change is permanent. It's to let vets-api access certs we provide for authentication to backend services. The other cert stuff is for CA cert store (and accessed through /etc/ssl)

PR #3857, opened 2020-02-03
Increase read timeout to 60 seconds for problem endpoints

saneshark:
## Description of change
During UAT it was identified that certain endpoints related to direct scheduling do not return a response in under 45 seconds. This change is to increase the timeout to ~60~ 55 seconds for these three endpoints. This will ensure that breakers does not cause a disruption across all of the other services as well. 

VAOS shared services team is actively looking at improving the latency upstream. This is just a temporary fix for UAT. We will not want to go live with these changes.

## Testing done
None

## Testing planned
Verify in staging that it's working.

#### Applies to all PRs

- [x] Appropriate logging

### DIFF
--- a/modules/vaos/app/services/vaos/appointment_service.rb
+++ b/modules/vaos/app/services/vaos/appointment_service.rb
@@ -6,7 +6,7 @@ module VAOS
       params = date_params(start_date, end_date).merge(page_params(pagination_params)).merge(other_params).compact
 
       with_monitoring do
-        response = perform(:get, get_appointments_base_url(type), params, headers, timeout: 60)
+        response = perform(:get, get_appointments_base_url(type), params, headers, timeout: 55)
         {
           data: deserialized_appointments(response.body, type),
           meta: pagination(pagination_params)
--- a/modules/vaos/app/services/vaos/appointment_service.rb
+++ b/modules/vaos/app/services/vaos/appointment_service.rb
@@ -6,7 +6,7 @@ module VAOS
       params = date_params(start_date, end_date).merge(page_params(pagination_params)).merge(other_params).compact
 
       with_monitoring do
-        response = perform(:get, get_appointments_base_url(type), params, headers)
+        response = perform(:get, get_appointments_base_url(type), params, headers, timeout: 60)
         {
           data: deserialized_appointments(response.body, type),
           meta: pagination(pagination_params)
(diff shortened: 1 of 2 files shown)


annaswims:
What's the maximum number or requests/min these services would make during UAT?

saneshark:
Very low. At most there are 3-5 sessions per day, and these three endpoints comprise less than a third of the total endpoints. All UAT is behind a feature toggle with email based opt-in, so we're in full control here. Furthermore, if a problem does arise that appears to compromise the integrity of the platform we can shit it down at any time since there is no implicit promise or contract that the service should be available while we are in testing.

markgreenburg:
@annaswims PACT calls are probably going to be one per UAT session. Clinics calls could be as high as 5 per UAT. Though like @saneshark mentioned, the most sessions we have scheduled per day right now is 4.

omgitsbillryan:
Can the timeout values be configified in a yaml file somehow? (excuse my ignorance - I dont know if modules share our settings.yml file 😬 ). It would make it easier for someone who isn't as familiar to change it in the event that it did "compromise the integrity of the platform" 🙂 .

In that vein, is there any reason that this call couldn't be done asynchronously with sidekiq?

saneshark:
Yes, asynchronous just isn't a good alternative. Would require considerable rewrite and only really provides lift to FE -> vets-api, not vets-api -> to upstream (which would still require an open connection. The upstream api-gateway does not support asynchronous requests.

I could certainly configure this as a yaml file, but this is a short term solution, incidentally it is needed it for tomorrow UAT.

saneshark:
Sometime tomorrow we'll have a bit more context on whether or not the upstream service in CDW can be updated / improved. That would solve the problem in a more ideal way. If for some reason that doesn't work, we'll have to look at alternatives such as building an aggregate table of our own using a batched sidekiq job that fetches all possible records from CDW nightly and builds a more efficient mechanism to query against.

If for some reason there is too much user specific context required in the query and we cannot build a local table, we're kind of stuck with the timeouts and will look at alternatives such as breaking out the VAOS module into its own separate AWS instance so that auto-scaling can be done independent of the platform or various other options. But I don't think asynchronous calls solves the problem really.

saneshark:
Again, I want to re-iterate, this is behind a feature toggle. There is no way it can compromise the integrity of the platform unless we were to somehow open the flood gates which have no intention of doing.

PR #3858, opened 2020-02-04
Veteran Verification API JWS Signing Service

rtravitz:
## Description of change

This PR addresses [#3119](https://github.com/department-of-veterans-affairs/vets-contrib/issues/3119). 

It adds a service to return responses as a JWT that can be verified with a public key by the consumer. The JWT response is in addition to the existing JSON response; it only comes back when the `Accept` header is `application/jwt`. This functionality is being added for USAJOBS, which receives its response through a third-party and needs to validate that the information is unchanged. 

This PR contains only the signing service, which wraps responses from the `/disability_rating` and `/service_history` endpoints. To complete the functionality, more PRs are needed to:
1. Inject a secret RSA key for each deployed environment (the one included with the current PR is only for testing and local development purposes). **A configuration PR with the keys should be merged prior to this one.**
2. Create a new endpoint that provides public keys that can verify the signed tokens generated by this `Notary` service. The public keys will be matched based on the `kid` provided with each generated token.
3. Update the Open API Spec for the Veteran Verification API to showcase this new functionality on `/disability_rating` and `/service_history` and document the new public key endpoint.

#### Notable changes
- Adds a new mime type of `application/jwt`, which maps to the symbol `:jwt`. When an incoming request has an `Accept` header of `application/jwt`, the `ActionController::MimeResponds` module will know how to respond to the request inside the `respond_to` block.
- Creates a `VeteranVerification::Notary` class to handle reading in an RSA key, signing arbitrary payloads, and generating a `kid` (key id) for future lookup of the corresponding public key.

#### Other notes
- The PR removes `json` as the default format in Veteran Verification's routes file. "Default" in this case turns out to mean "always." The "default" is not overridden when another `Accept` header is provided. Removing this route constraint also meant having to explicitly add  `application/json` as the `Accept` header to request specs, which request `text/html` by default. The `set_default_format` method was added to the controllers to prevent a regression in behavior from the view of our consumers who could previously still get a response without an `Accept` header because of the route constraint. Without that method, consumers would receive a 406 if they don't provide a value for `Accept` that we support.

## Testing done

- Added integration and unit tests for signing a payload from Veteran Verification endpoints

## Testing planned

- Manual validation of functionality in deployed lower environments

#### Unique to this PR

- [x] Add JWS wrapper to Disability Rating response
- [x] Add JWS wrapper to Service History response

#### Applies to all PRs
- [x] Appropriate logging


modules/veteran_verification/spec/requests/service_history_request_spec.rb
```diff
@@ -7,12 +7,16 @@
 
   let(:scopes) { %w[profile email openid service_history.read] }
 
+  def headers(auth)
+    auth.merge('Accept' => 'application/json')
+  end
```
kalilsn: I think it would be good to have a test in here that either removes the default `Accept: text/html` or that uses `Accept: */*` to verify the `set_default_format` method is working as intended.

modules/veteran_verification/app/controllers/veteran_verification/v0/disability_rating_controller.rb
```diff
@@ -1,21 +1,38 @@
...
 
 require_dependency 'veteran_verification/application_controller'
+require_dependency 'notary'
 
 module VeteranVerification
   module V0
     class DisabilityRatingController < ApplicationController
+      include ActionController::MimeResponds
+
+      NOTARY = VeteranVerification::Notary.new(Settings.vet_verification.key_path)
```
kalilsn: Is this going to happen once per request? If so, do you think it's worth moving this instantiation into the jwt response on L26 so that we don't add this unnecessary IO to json responses?

modules/veteran_verification/app/controllers/veteran_verification/v0/service_history_controller.rb
```diff
@@ -1,17 +1,37 @@
...
+
+        respond_to do |format|
+          format.json { render json: serialized.to_json }
+          format.jwt { render body: NOTARY.sign(serialized.serializable_hash) }
+        end
+      end
+
+      private
+
+      def set_default_format
```
kalilsn: Could you add your explanation for using this method (rather than the route default) as a comment?

kalilsn: Also, I think this method could be extracted to the `ApplicationController` if you want!

modules/veteran_verification/app/controllers/veteran_verification/v0/disability_rating_controller.rb
rtravitz: This only gets loaded on the first request the first time that the controller is loaded. Had the same concern about the syscall and moved the Notary instantiation up here to prevent it 🚀

modules/veteran_verification/app/controllers/veteran_verification/v0/service_history_controller.rb
rtravitz: Moved this to the ApplicationController as suggested

modules/veteran_verification/spec/requests/service_history_request_spec.rb
rtravitz: Changed the default format for the request specs to `Accept: */*`, which still passes everything.

modules/veteran_verification/app/controllers/veteran_verification/v0/disability_rating_controller.rb
kalilsn: Excellent. I thought maybe rails instantiated the controller class on each request – seems I have some reading to do!

modules/veteran_verification/lib/notary.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+require 'jwt'
+
+module VeteranVerification
+  class NotaryException < StandardError
+  end
+
+  class Notary
```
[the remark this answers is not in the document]
rtravitz: Added a comment explaining the service

modules/veteran_verification/app/controllers/veteran_verification/v0/disability_rating_controller.rb
rtravitz: You should be able to verify with something as simple as a `puts` if you want to see it with your own eyes! Add the line, make a couple of requests, and look for the output.

PR #3866, opened 2020-02-06
Displaying EVSS Errors Part 2

bastosmichael:
## Description of change
We had to revert a change with PR https://github.com/department-of-veterans-affairs/vets-api/pull/3852 and resolves https://github.com/department-of-veterans-affairs/vets-contrib/issues/3946 this is our attempt at fixing the problems discovered and testing for them.

- Show error for errored statuses in Claims submissions

#### Applies to all PRs

- [ ] Appropriate logging


modules/claims_api/app/controllers/claims_api/v0/claims_controller.rb
```diff
@@ -22,25 +22,7 @@ def index
...
-          end
-        end
-      end
-
-      private
-
-      def fetch_and_render_evss_claim(id)
-        claim = claims_service.update_from_remote(id)
-        render json: claim, serializer: ClaimsApi::ClaimDetailSerializer
+        super
```
edmangimelli: this is super bizarro. i wonder why you need to do this

bastosmichael: https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/app/controllers/claims_api/v0/forms/disability_compensation_controller.rb#L52-L54 I ran into the same problem when I tried consolidating our validate end point, I'm not sure if it's just how vets-api is configured or maybe even how our module is configured but it'll simply fail and not recognize the method if I remove it completely from the controller

PR #3868, opened 2020-02-06
Feature toggle by string

annaswims:
## Description of change
If we modify the Front end to pass a string like the Google analytics id, we could better roll out featues for a percentage of users. It could be configurable on a per-toggle basis whether we should look at the logged in user or the GA id.

We already [get the GA id's](https://github.com/department-of-veterans-affairs/vets-website/blob/7e33f4acf79780b02a704abc241debb402b559be/src/platform/user/authentication/utilities.js#L51)  on the front end.

Something like: http://localhost:3000/v0/feature_toggles?features=ssoe&cookie_id=GA-abc123

## Testing done
Text if configured to use logged in user
Screen Shot 2020-02-19 at 5 16 40 PM
Text if configured to use GA id from front end 
Screen Shot 2020-02-19 at 5 15 57 PM

#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging


lib/flipper/views/feature.erb
```diff
@@ -1,4 +1,9 @@
 <% if RequestStore.store[:flipper_user_email_for_log].present? || Rails.env.development? %>
+
+  <% actor_type = FLIPPER_FEATURE_CONFIG['features'].dig(feature_name, 'actor_type')%>
```
annaswims: Ideally, these would be in some sort of helper, but I'm not sure that it's worth the effort to open that can of worms since the controller for this view is in the gem.

config/features.yml
```diff
@@ -2,36 +2,53 @@
 # Add a new feture toggle here to ensure that it is initialized in all environments.
 # Features are defaulted to enabled in development and test environments and disabled in all others.
 # The description should contain any relevant information for an admin who may toggle the feature.
+#
+# The actor_type should be either `user` for featires you want to be "sticky" for a logged in user (default)
+#  or `cookie_id` of you wish to use the Google Analytics id as the unique identifier.
 
 features:
   facility_locator_show_community_cares:
+    actor_type: user
```
annaswims: technically,  `actor_type: user` is optional, but I thought it was more clear this way

config/initializers/flipper.rb
```diff
@@ -32,10 +32,6 @@
 
   # Labeling what Flipper calls "actors" as "users" in the UI
   config.actors.title = 'Users'
-  config.percentage_of_actors.title = 'Percentage of Logged in Users'
```
annaswims: moved these strings into the view with some additional logic.

app/controllers/v0/feature_toggles_controller.rb
```diff
@@ -11,12 +11,21 @@ def index
...
       end
       render json: { data: { type: 'feature_toggles', features: features } }
     end
+
+    def actor(feature_name)
+      if FLIPPER_FEATURE_CONFIG['features'].dig(feature_name, 'actor_type') == FLIPPER_ACTOR_STRING
+        Flipper::Actor.new(params[:cookie_id])
+      else
+        @current_user
+      end
```
omgitsbillryan: Can you educate me on `Flipper::Actor`? Is there any reason we shouldn't also wrap `@current_user` as well, i.e. `Flipper::Actor.new(@current_user)`? Excuse the ignorance if this is a dumb question - my knowledge of `flipper` is pretty minimal.

config/features.yml
omgitsbillryan: What happens when `actor_type` is omitted for a given feature config? Is there a validation somewhere / is there a default value?

app/controllers/v0/feature_toggles_controller.rb
annaswims: user has a method called [fliper_id](https://github.com/department-of-veterans-affairs/vets-api/blob/dac70152cbd9c40ab0d70233b7e443c487e8ee07/app/models/user.rb#L307) that returns a unique id for a logged in user. 

[Flipper::Actor](https://github.com/jnunemaker/flipper/blob/master/lib/flipper/actor.rb) is a class that has the bare mininum number of methods to be n actor, just a `fliper_id` methods and an initializer.

config/features.yml
annaswims: it will [default to the logged in user](https://github.com/department-of-veterans-affairs/vets-api/pull/3868/files#diff-d14889a0f335ac94fca7645d09edb8d6R24).

PR #3869, opened 2020-02-06
va.gov-team#2628: Tune sidekiq to only kick off 16 threads concurrently

benniemosher:
## Description of change

Currently our sidekiq threads are ramping up to 25. We only allow 16 connections to a DB, so this could solve a problem. This change sets sidekiq to only allow up to 16. The default in v4 of sidekiq is 20 and in v5 was changed to 10. The current version is v6, so we are 2 major versions behind. Maybe it's time for an upgrade?

There is currently no metrics for sidekiq being tracked, and may need to be addressed in another issue. (SEE: https://github.com/mperham/sidekiq/wiki/Pro-Metrics)

SEE: https://app.zenhub.com/workspaces/vsp-5cedc9cce6e3335dc5a49fc4/issues/department-of-veterans-affairs/va.gov-team/2628

## Testing done

```
make up
docker exec -it 2cbfa98e7861 /bin/bash
ps aux | grep sidekiq

vets-api@2cbfa98e7861:~/src$ ps aux | grep sidekiq
vets-api    12  4.0 15.1 2153168 307676 pts/0  Sl+  20:19   0:23 sidekiq 4.2.10 src [0 of 16 busy]
vets-api   114  0.0  0.0  11112   968 pts/1    S+   20:29   0:00 grep sidekiq
```

#### Unique to this PR

#### Applies to all PRs

- [x] Appropriate logging

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -1,0 +1,2 @@
+---
+:concurrency: 16


wyattwalter:
@benniemosher (and anyone who sees this later) we _do_ have metrics from sidekiq already. They're being renamed by the Prometheus statsd_exporter under "api_". I'm not sure why that was done, but they are being collected and enabled.

nathanhruby:
Bennie and I were chatting and I neither of us saw thread count metrics (either in prom os the sidekiq docs) which would be the obvious way to check success here.

benniemosher:
Grafana board from @wyattwalter found here: http://prometheus-prod.vetsgov-internal:9090/prometheus/graph?g0.range_input=1h&g0.expr=api_jobs_working&g0.tab=1

PR #3871, opened 2020-02-06
Vaos logging middleware

sjbrush:
## Description of change
Introducing middleware logging for service calls.

## Testing done
Specs for the middleware. No integration spec, but verified in console it is working as expected.

To reverify locally:

`RAILS_ENABLE_TEST_LOG=true bundle exec rspec modules/vaos/services/middleware/vaos_logging_spec.rb`

while simultaneously running:

 `tail -f log/test.log | grep --line-buffered 'VAOS service call'`

## Testing planned
Need to write some cloud watch scripts to verify in other environments. In particular, want to be able to query AWS Cloudwatch based on request_id found in Sentry.

Pretty much done. Just logging.

#### Applies to all PRs

- [x] Appropriate logging

Technically it does include some sensitive information such as ICN and other information, but these are needed for logging purposes. It's in cloudwatch though, which is more secure than Sentry.


modules/vaos/spec/services/middleware/vaos_logging_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+      conn.use :vaos_logging, service_name
+
+      conn.adapter:test do |stub|
+        stub.get('success') { [200, { 'Content-Type' => 'text/plain' }, response_data] }
+        stub.post(user_service_uri) { [200, { 'Content-Type' => 'text/plain' }, response_data] }
+      end
+    end
+  end
+
+  let(:service_name) { 'jalepeno' }
```
kreek: 🌶

modules/vaos/app/services/vaos/middleware/vaos_logging.rb
```diff
@@ -0,0 +1,65 @@
...
+        @app.call(env).on_complete do |response_env|
+          jti = jti(env, response_env)
+          duration = Time.current - start_time
+          status = response_env.status
+
+          if status.between?(200..299) 
+            logger.info('vaos service call succeeded:',
+              duration: duration,
+              status: status,
+              jti: jti,
```
kreek: jti includes ssn will this get filtered out?

saneshark: The JTI includes it or JWT includes it? I think it's just a unique identifier generated by VAMF. If we filter it, we can't reconcile logs -- i thought this was our unique identifier we go to VAMF with. It includes SSN really? 😬

modules/vaos/spec/services/middleware/vaos_logging_spec.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require_relative '../../support/fixture_helper'
+
+describe VAOS::Middleware::VaosLogging do
```
kreek: Could we add a spec that confirms no PII is logged?

saneshark: how? And what are you characterizing as PII? Because we definitely need to log all of the following:

ICN, JTI, the URL (which likely also include ICN or EDIPI), the users UUID, etc.

kreek: Yep those are fine, but I'm pretty sure the one that comes back from VAMF includes SSN

saneshark: oh, well we're only logging the JTI from the decoded header, so if it includes anything else it shouldn't appear in logs.

saneshark:
LGTM, but I also contributed to this, so someone else can feel free to review.

modules/vaos/spec/services/middleware/vaos_logging_spec.rb
saneshark: We are not logging the entire decoded JWT. Only the JTI key.

saneshark:
Do you think maybe we should add the http method? could probably change url to be "GET ..." vs "PUT ..." if we don't want to add too many more tags, but I don't see a problem with either.

modules/vaos/app/services/vaos/jwt.rb
```diff
@@ -17,6 +17,10 @@ def token
       ::JWT.encode(payload, rsa_private, 'RS512')
     end
 
+    def self.decode(jwt, key = nil, verify = true, options = {})
+      ::JWT.decode(jwt, key, verify, options)
+    end
```
omgitsbillryan: Does this need to exist? Can't `vaos_logging#decode_jwt` call `JWT.decode(jwt, key, verify, options)` directly?

modules/vaos/app/services/vaos/middleware/vaos_logging.rb
```diff
@@ -0,0 +1,61 @@
...
+      def initialize(app)
+        super(app)
+      end
+
+      def call(env)
+        start_time = Time.current
+
+        @app.call(env).on_complete do |response_env|
+          log_tags = {
+            jti: jti(env, response_env),
```
omgitsbillryan: What's jti? 

Might be worth spelling out directly if you think someone like me who doesn't know this stands for could someday be reading these logs.

modules/vaos/app/services/vaos/middleware/vaos_logging.rb
```diff
@@ -0,0 +1,61 @@
...
+        VAOS::JWT.decode(token, rsa_private.public_key, false, algorithm: 'RS512').first
+      end
+
+      def user_session_request?(env)
+        env.url.to_s.include?('users/v2/session?processRules=true') ? true : false
+      end
+
+      def rsa_private
+        OpenSSL::PKey::RSA.new(File.read(Settings.va_mobile.key_path))
+      end
```
omgitsbillryan: Reading a file on disk + initializing an OpenSSL RSA private key seems like an expensive thing to do on every request. Any merit to memoizing this as a class variable instead?

modules/vaos/app/services/vaos/jwt.rb
saneshark: we're not verifying the signature. You'll notice the actual call to the decode method used in specs and the code has the verify flag as false. This is because the JWT is signed using staging certs that are unavailable locally and in test environment. It's mostly there for show, and shouldn't be causing performance problems -- but it's handy for verifying on a staging instance.

modules/vaos/app/services/vaos/middleware/vaos_logging.rb
saneshark: i think it's a common enough nomenclature for anyone familiar with JWT. It's basically a unique identifier for the JWT session. I could put a comment around it if helpful.

modules/vaos/app/services/vaos/middleware/vaos_logging.rb
saneshark: so I don't think that code actually gets invoked seeing as we're not verifying the signature. The JWT is not encrypted, it's just signed and to get the JWT we do not currently verify signature.

modules/vaos/app/services/vaos/middleware/vaos_logging.rb
omgitsbillryan: Whoops. My bad. No need.

modules/vaos/app/services/vaos/middleware/vaos_logging.rb
omgitsbillryan: Starting on line 15: `jti()` --> `decode_jwt()` --> `rsa_private.public_key` would execute this code, no?

modules/vaos/app/services/vaos/jwt.rb
omgitsbillryan: I'm not seeing how any of this is relevant to the initial question 😬 .

modules/vaos/app/services/vaos/middleware/vaos_logging.rb
```diff
@@ -0,0 +1,61 @@
...
+
+      private
+
+      def log(type, message, tags)
+        Rails.logger.send(type, message, tags)
+      end
+
+      def decode_jwt(token)
+        VAOS::JWT.decode(token, rsa_private.public_key, false, algorithm: 'RS512').first
```
omgitsbillryan:
```suggestion
       ::JWT.decode(token, rsa_private.public_key, false, algorithm: 'RS512').first
```

modules/vaos/app/services/vaos/jwt.rb
omgitsbillryan: To be clear, I'm asking why we cant delete the entire `self.decode(jwt, key = nil, verify = true, options = {})` class method from this `VAOS::JWT` class. It seems like it's only usage is in `VAOS::Middleware::VaosLogging` (which could be [updated to use `::JWT` directly](https://github.com/department-of-veterans-affairs/vets-api/pull/3871/files#r382709855))

Don't mean to belabor such an unimportant question, just trying to make sure we're on the same page.

modules/vaos/app/services/vaos/middleware/vaos_logging.rb
saneshark: yes, you're right. I've changed the code. And I opened up a ticket to investigate memoizing that as part of user-service as well. Very good catch.

modules/vaos/app/services/vaos/middleware/vaos_logging.rb
saneshark: I went ahead and added a bunch of YARD documentation to the logging middleware. Should help document JTI and other attributes a bit better.

PR #3873, opened 2020-02-07
wrong file path check in dockerfile

wyattwalter:
## Description of change

Fixes logic error in Dockerfile. This check should be looking for the presence of .pem files to rename, not .crt. It works, but only because we happen to have both that get added by Ansible.

Part of https://github.com/department-of-veterans-affairs/va.gov-team/issues/5239

## Testing done

Tested locally with `make ci-build` in scenarios with .crt and with .pem files.

#### Applies to all PRs

- [x] Appropriate logging

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,7 @@ RUN mkdir -p /srv/vets-api/{clamav/database,pki/tls,secure,src} && \
 # XXX: get rid of the CA trust manipulation when we have a better model for it
 COPY config/ca-trust/* /usr/local/share/ca-certificates/
 # rename .pem files to .crt because update-ca-certificates ignores files that are not .crt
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -22,7 +22,7 @@ RUN mkdir -p /srv/vets-api/{clamav/database,pki/tls,secure,src} && \
 # XXX: get rid of the CA trust manipulation when we have a better model for it
 COPY config/ca-trust/* /usr/local/share/ca-certificates/
 # rename .pem files to .crt because update-ca-certificates ignores files that are not .crt
```
erluti: does this comment make sense with the change?

wyattwalter: yep! I just had the logic flipped in the `if`

PR #3878, opened 2020-02-07
Allow SSOe session controller to uplevel successfully

patrickvinograd:
## Description of change
Per contract with SSOe, we need to use a new `ial` parameter to determine if a session can be upleveled, and we need to pass the standard `loa/3` authn_context instead of `loa/3/vets`. 

Most of this PR (particularly around the specs) provides for the v0 sessions controller to continue as is while the v1 sessions controller uses the new context. It also cleans up some obsolete code that was used to suppress up-leveling earlier during development. 

## Testing done
Exercised ID.me login locally through both v0 and v1 session controllers.

#### Unique to this PR
v1 sessions controller is still hidden in production. 

#### Applies to all PRs
- [ ] Appropriate logging


app/controllers/v1/sessions_controller.rb
```diff
@@ -194,7 +192,8 @@ def request_type
     end
 
     def url_service
-      SAML::URLService.new(saml_settings, session: @session_object, user: current_user, params: params)
+      SAML::URLService.new(saml_settings, session: @session_object, user: current_user,
+                                          params: params, loa3_context: LOA::IDME_LOA3)
```
ericbuckley: this looks fine, but the name of the param, loa3_conext, might be a little misleading as this only changes the idme behavior and not mhv/dslogon right?

patrickvinograd:
The LOC warning is due largely to the URL service specs which were cloned to be relevant for the v1 sessions controller.

ericbuckley:
@patrickvinograd as part of the SSOe upleveling cleanup effort, I believe the `skip_uplevel` parameter can be removed from the `login_redirect_url` function.  https://github.com/department-of-veterans-affairs/vets-api/pull/3878/files#diff-b612ab0a1a74aaaac08287659f2b70a7R48

patrickvinograd:
Removed the obsolete skip_uplevel parameter from url_service

config/initializers/loa.rb
```diff
@@ -8,5 +8,6 @@ module LOA
 
   IDME_LOA1 = 'http://idmanagement.gov/ns/assurance/loa/1/vets'
   IDME_LOA2 = 'http://idmanagement.gov/ns/assurance/loa/2/vets'
-  IDME_LOA3 = 'http://idmanagement.gov/ns/assurance/loa/3/vets'
+  IDME_LOA3 = 'http://idmanagement.gov/ns/assurance/loa/3'
+  IDME_LOA3_VETS = 'http://idmanagement.gov/ns/assurance/loa/3/vets'
```
omgitsbillryan: If you're going to change `IDME_LOA3` to `IDME_LOA3_VETS`, then you should probably stay consistent also do it for 1 & 2 as well. Since that will require some annoying updates to code, you might want to save yourself a headache and just call this new one `LOA3`.

spec/factories/users.rb
```diff
@@ -105,7 +105,7 @@
     end
 
     trait :loa3 do
-      authn_context { 'http://idmanagement.gov/ns/assurance/loa/3/vets' }
+      authn_context { 'http://idmanagement.gov/ns/assurance/loa/3' }
```
omgitsbillryan:
```suggestion
      authn_context { LOA::IDME_LOA3 }
```

This trait has historically been for the `/vets` variety of an loa3 user. I'm surprised that this didn't break any existing specs... I guess we weren't using it anywhere?

spec/controllers/v1/sessions_controller_spec.rb
```diff
@@ -24,8 +24,8 @@
   let(:callback_url)        { "http://#{request_host}/v1/sessions/callback" }
   let(:logout_redirect_url) { 'http://127.0.0.1:3001/logout/' }
 
-  let(:settings_no_context) { build(:settings_no_context, assertion_consumer_service_url: callback_url) }
-  let(:rubysaml_settings)   { build(:rubysaml_settings, assertion_consumer_service_url: callback_url) }
+  let(:settings_no_context) { build(:settings_no_context_v1, assertion_consumer_service_url: callback_url) }
```
omgitsbillryan: I just realized that `:settings_no_context` isn't used anywhere in `sessions_controller_spec.rb` 😬

spec/factories/rubysaml_settings.rb
```diff
@@ -31,4 +31,34 @@
...
+    private_key                     { Settings.saml.key }
+    sp_entity_id                    { Settings.saml.issuer }
+    assertion_consumer_service_url  { Settings.saml.callback_url }
+    idp_cert                        { File.read("#{::Rails.root}/spec/fixtures/files/idme_cert.crt") }
+    idp_cert_fingerprint            { '74:DC:33:BE:D6:92:69:3F:81:65:F6:CF:ED:55:82:E0:A5:65:B3:32' }
+    idp_entity_id                   { 'https://pint.eauth.va.gov/isam/sps/saml20idp/saml20' }
+    idp_slo_target_url              { 'https://pint.eauth.va.gov/pkmslogout' }
+    idp_sso_target_url              { 'https://pint.eauth.va.gov/isam/sps/saml20idp/saml20/login' }
+    name_identifier_format          { 'urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified' }
+  end
```
omgitsbillryan: I think these 2 factories are identical, except for `rubysaml_settings_v1 ` having the `:rollover_cert` trait.  I think maybe you don't need the `no_context` variety (since neither has an `authn_context`).

config/initializers/loa.rb
omgitsbillryan: Actually, I just realized that renaming to LOA3 would result in `LOA::LOA3` 😢 . My main ask is for the `/vets` flavored ones to all be consistent with one another.

patrickvinograd: Yeah...actually not too bad to change the the uses of `LOA1` to `LOA1_VETS` so I'll do that

PR #3879, opened 2020-02-07
5674 vaos empty appointments

kreek:
## Description of change
Fixes a bug where VAOS empty appointments removes the JSON key rather than returning an empty array. The API will now catch that and return an empty array e.g.

 ```json
{
  "data": [],
  "meta": {
    "pagination": {
      "current_page": 0,
      "per_page": 0,
      "total_entries": 0,
      "total_pages": 0
    }
  }
}
```

## Testing done
unit tests

## Testing planned
staging integration tests

Does not throw an error

#### Unique to this PR
- [ ] returns an empty array when there are no VAOS appointments for the user

#### Applies to all PRs

- [ ] Appropriate logging

### CODE REVIEW

File: modules/vaos/app/services/vaos/appointment_service.rb [hunk 1]
```diff
@@ -42,12 +42,14 @@ def put_cancel_appointment(request_object_body)
...
-                                                  .map { |appointments| OpenStruct.new(appointments) }
-      end
+      appointment_list = if type == 'va'
+                           json_hash.dig(:data, :appointment_list)
+                         else
+                           json_hash[:booked_appointment_collections].first[:booked_cc_appointments]
+                         end
+      return [] unless appointment_list
+
+      appointment_list.map { |appointments| OpenStruct.new(appointments) }
```
sjbrush: this looks like it occurs after method returns.  is this just a ruby thing or is this not used?

saneshark: you could also do Array.wrap

since Array.wrap(nil) returns []

kreek: @sjbrush ya `unless` is a rubyism to check that a statement returns false (e.g. instead of `if !something`), in this case it's a guard clause to return early if appointment_list is falsey

PR #3882, opened 2020-02-10
add --parallel to rubocop cmd

wyattwalter:
## Description of change

Adds `--parallel` to the rubocop command.

Noticed this today while poking around in a Jenkins node. We aren't running rubocop in `--parallel` mode, so it's tied to a single CPU core. We can't use this option for non-CI as we have `--auto-correct` enabled, and the two are mutually exlusive.

I didn't pick up this epic, but would squarely fall under https://github.com/department-of-veterans-affairs/va.gov-team/issues/4464

## Testing done

Tested locally and in CI.

#### Applies to all PRs

- [x] Appropriate logging


johnpaulashenfelter:
Happy to see this go in

PR #3889, opened 2020-02-11
Lighthouse Appeals API - switch to ssn field

edmangimelli:
resolves https://github.com/department-of-veterans-affairs/vets-contrib/issues/4150

Change the `veteranId` and `fileNumberOrSsn` fields to `ssn` [in accordance with changes in Caseflow](https://github.com/department-of-veterans-affairs/caseflow/pull/13166).

Updates documentation accordingly.

All tests in `modules/appeals_api` have been updated and tested.

### Acceptance Criteria
- [x] tests pass.
- [x] all uses of `veteranId` and `fileNumberOrSsn` have been changed to `ssn`


edmangimelli:
> Instead of the header being just SSN, can you make the header X-VA-SSN so that it's more uniform with the other requests across the rest of the platform?
> 
> https://github.com/department-of-veterans-affairs/vets-api/blob/890be296a5637c6f89696785c4f03ddfab100a7c/modules/claims_api/app/controllers/claims_api/application_controller.rb#L37

i like it! i'd rather be consistent across lighthouse
questions and concerns:
1) Is `X-VA` or `X-` the standard across Lighthouse?
2) Browsing through the dev portal, I don't see a ton of endpoints using headers --are the use of these prefixes pretty much exclusive to Benefits Claims?
3) It's my understanding that naming custom headers with the prefix `X-` is no longer recommended --I wonder if in the future we (Lighthouse) wanna get away from this or no? I think this is really minor, and I don't really care one way or another, but was just curious if this had already been discussed.
4) Lastly, I'm not sure if this should be changed only in vets-api and then translated for Caseflow:
```
X-VA-SSN,  X-VA-Receipt-Date  ==>  ssn,  receiptDat
```
or if I should change this in Caseflow (also not sure how the Caseflow team would feel about it --maybe they don't care).
If it's up to me, I'd rather _not_ transform the keys.

cc @charleystran

charleystran:
I would say that since apiKey is not funnel back to us, we are ok for that. I am a little confused on this but overall we should get X-VA headers from users but pass whatever is standard to people  like caseflow.

modules/appeals_api/app/swagger/v1/decision_reviews.yaml
```diff
@@ -256,9 +255,6 @@ paths:
                       a veteran's contestable issues to select from, and include
                       the _ContestableIssue_ objects that you want to contest in this
                       array.
-                      When including a _ContestableIssue_ obtained from the
-                      `/contestable_issues` endpoint, use the `notes` field to provide
-                      additional information about contesting this issue.
```
edmangimelli: this change is just to be in-sync with Caseflow's documentation (which accidentally became more up-to-date)

modules/appeals_api/app/swagger/v1/decision_reviews.yaml
```diff
@@ -269,9 +265,6 @@ paths:
                       attributes:
                         type: object
                         properties:
-                          notes:
-                            type: string
-                            nullable: true
```
edmangimelli: again:
>this change is just to be in-sync with Caseflow's documentation (which accidentally became more up-to-date)

modules/appeals_api/app/swagger/v1/decision_reviews.yaml
```diff
@@ -308,35 +301,6 @@ paths:
...
-                        participantId: "33333333"
-                        payeeCode: "11"
-                  included:
-                    - type: ContestableIssue
-                      attributes:
-                        decisionIssueId: 206
-                    - type: ContestableIssue
-                      attributes:
-                        ratingIssueId: 80
       responses:
```
edmangimelli: again:
>this change is just to be in-sync with Caseflow's documentation (which accidentally became more up-to-date)

PR #3899, opened 2020-02-13
Create CARMA Service (Salesforce Client)

kevinmirc:
Fixes: https://github.com/department-of-veterans-affairs/va.gov-team/issues/5326
Fixes: https://github.com/department-of-veterans-affairs/va.gov-team/issues/5453
Fixes: https://github.com/department-of-veterans-affairs/va.gov-team/issues/5454

## Description of change
Create a service that can authenticate against CARMA and send it a CaregiversAssistanceClaim (10-10CG Submission).

## Technical Summary
- Added `Form1010cg` namespace in the main application.
- Added a `Form1010cg::Submission` model.
Local record of the `CaregiversAssistanceClaim` being submitted.
- Added `Form1010cg::Service`: Manages interaction between `CaregiversAssistanceClaims`, `Form1010cg::Submission`s, and the `CARMA` library.
Handles the business logic.
- Added a `CARMA` lib. This has an HTTP client and domain models for the mail app's consumption.
- `POST v0/caregivers_assistance_claim` now returns a `Form1010cg::Submission` vs a `SavedClaim`.

**Requirements**:
- [x] It should be able to authenticate against the salesforce service (extend Salesforce::Service)
- [x] It should have the ability to transform a CaregiversAssistanceClaim's form to a valid format that CARMA accepts (data contract is still in the works).
- [x] It should be able to POST the submission data to the Salesforce service.
- [ ] It should be able to handle errors from an unsuccessful submission (both 4xx and 5xx errors)
Will be implemented when we do a real request to CARMA.
- [x] It should be able to handle a successful response (return the CARMA Case Id)
- [x] Place private signing keys in config for each env
https://github.com/department-of-veterans-affairs/devops/pull/6648
- [ ] Update vets-api settings for each env (url, username, signing_key_path, consumer_key)
Not needed since we're using a stub response for now
- [x] Pass related certs with public keys to CARMA

#### Applies to all PRs

- [ ] Appropriate logging


app/controllers/v0/caregivers_assistance_claims_controller.rb
```diff
@@ -6,14 +6,14 @@ class CaregiversAssistanceClaimsController < ApplicationController
     skip_before_action(:authenticate)
 
     def create
-      claim = service.submit_claim!(nil, claim_params)
-      render json: claim, serializer: SavedClaimSerializer
+      submission = service.submit_claim!(nil, claim_params)
+      render json: submission, serializer: ::Form1010cg::SubmissionSerializer
```
jholton, line 10: Does this form use `in_progress_forms` to save form data for logged in users?
You have code in place to delete those forms after submission, but you never invoke it since you hard code the `user_context` param as `nil` in the call (line 9).

There's already some helpers for this situation, so here's a suggestion:
```suggestion
      submission = service.submit_claim!(claim_params)
      load_user
      clear_saved_form(SavedClaim::CaregiversAssistanceClaim::FORM)
      render json: submission, serializer: ::Form1010cg::SubmissionSerializer
```
And you can strip everything related to `user_context` and destroying `in_progress_forms` from your service class.

lib/carma/client/client.rb
```diff
@@ -0,0 +1,47 @@
...
+            '/services/apexrest/carma/v1/1010-cg-submissions',
+            payload,
+            'Content-Type': 'application/json'
+          ).body
+        end
+
+        response_body
+      end
+
+      # The CARAM Staging and Prod enviornments will not exist until ~May 2020
```
jholton, line 31: typo:
```suggestion
      # The CARMA Staging and Prod enviornments will not exist until ~May 2020
```

lib/carma/models/submission.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+module CARMA
+  module Models
+    class Submission < Base
+      attr_reader :metadata
+      attr_accessor :data
+      attr_accessor :carma_case_id, :submitted_at
```
jholton, line 8: Nitpicky, but any reason not to include these all on one line?
```suggestion
      attr_accessor :carma_case_id, :data, :submitted_at
```

lib/carma/models/submission.rb
```diff
@@ -0,0 +1,60 @@
...
+          metadata: metadata.merge(claim_id: claim.id)
+        )
+      end
+
+      def initialize(args = {})
+        @carma_case_id = args[:carma_case_id]
+        @submitted_at = args[:submitted_at]
+        @data = args[:data]
+
+        send('metadata=', args[:metadata] || {})
```
jholton, line 31: Why not just
```suggestion
        @metadata = Metadata.new(args[:metadata] || {})
```
and remove the `metadata=` method?  Is this the `metadata=` method used anywhere else?

app/models/form1010cg/submission.rb
```diff
@@ -0,0 +1,26 @@
...
+    # More information about CARMA can be found in lib/carma/README.md
+
+    attr_accessor :carma_case_id # The id of the CARMA Case created from this form submission
+    attr_accessor :submitted_at # The timestamp of when the submission was accepted in CARMA
+
+    def id
+      nil
+    end
+
+    def persisted?
```
jholton, line 18: How is this used?

spec/lib/carma/models/submission_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+
+    context 'when :data is invalid' do
+      xit 'raises exception' do
+      end
+    end
+
+    context 'when :metadata is invalid' do
+      xit 'raises exception' do
+      end
+    end
```
jholton, line 133: Need to be implemented?

lib/carma/models/submission.rb
kevinmirc: I'm using `:send` so the logic to parse the setter value to the `Metadata` model, lives in one place. Although `:metadata=` is not being called anywhere, semantically I think it's appropriate for it to be a public method.

If you feel that I'm adding a method prematurely, I could remove `:metadata=` for now, and add it back in when it actually needs to be called by `/app`.

app/models/form1010cg/submission.rb
```diff
@@ -0,0 +1,26 @@
...
+    # Used to store data of the relating record created in the form's ultimate destination (CARMA).
+    #
+    # More information about CARMA can be found in lib/carma/README.md
+
+    attr_accessor :carma_case_id # The id of the CARMA Case created from this form submission
+    attr_accessor :submitted_at # The timestamp of when the submission was accepted in CARMA
+
+    def id
+      nil
+    end
```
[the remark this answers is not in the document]
kevinmirc: Good catch. Thanks!

app/models/form1010cg/submission.rb
kevinmirc: Removed. `:id` and `:persisted?`were placeholders because I thought I'd be inheriting from ActiveRecord by now, but that will come in the future.

spec/lib/carma/models/submission_spec.rb
kevinmirc: Removed. I decided not to add `ActiveRecord::Validations` with a `:valid?` pattern and just build form a single `:from_claim` method that properly assigns values. I may use `ActiveRecord::Validations` in the future if consumers of this library (aka `app`) requires a more atomic interface.

app/controllers/v0/caregivers_assistance_claims_controller.rb
kevinmirc: Sorry for the confusion here. There is no longer an authenticated experience for this form, it is only available through an unauthenticated session, and no save in progress will be available for it. So I've removed this functionality entirely.

app/serializers/form1010cg/submission_serializer.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Form1010cg
+  class SubmissionSerializer < ActiveModel::Serializer
+    attribute(:confirmation_number) { object.carma_case_id }
+    attribute(:submitted_at) { object.submitted_at }
```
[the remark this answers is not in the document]
kevinmirc: Turns out `attribute :submitted_at` will not work since the model is not an ActiveRecord instance. So I will need to use the more verbose version for now.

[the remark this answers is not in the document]
jholton: huh, interesting.  I guess they're serious about the `active_model` part of `active_model_serializers` 😄

lib/carma/models/submission.rb
jholton: I don't feel strongly.  But it is a public method, so you don't need to use `send`, right?
```suggestion
        self.metatdata = args[:metadata] || {}
```

app/controllers/v0/caregivers_assistance_claims_controller.rb
jholton: That's... interesting.  So, if I visit `va.gov` and log in, I _cannot_ start this form?  Will I be forced to log out first?

lib/carma/models/submission.rb
kevinmirc: Yes, I can! Good call. I've updated it.

app/controllers/v0/caregivers_assistance_claims_controller.rb
kevinmirc: I believe if you are already logged into va.gov, and you visit this page, you can still fill it out as normal. On the backend, we will treat it as if you're unauthenticated. The end-user will not have pre-fill or save-in-progress features for it.

From what I understand save-in-progress is disabled on the front-end (removing the button / not using the specific component).

kevinmirc: I'm checking in with the team to confirm that this is the user flow, it changes quite a bit in the lifetime of this Draft PR.

kevinmirc:
Thanks for the review @jholton!

PR #3901, opened 2020-02-13
3546 VAOS replace virtus with dry types/schema

kreek:
## Description of change
Prototype PR for swapping out the virtus gem, which has been discontinued, with it's successor dry-types (specifically [dry-struct](https://dry-rb.org/gems/dry-struct/1.0/) and [dry-schema](https://dry-rb.org/gems/dry-schema/1.4/)). The main difference is that you have to explicitly mark where you want type coercion to occur e.g.

```ruby
attribute :appointment_length, Integer
``` 
vs
```ruby
attribute :appointment_length, Types::Coercible::Integer
``` 

Forms have been replaced with Params which is a class based on dry-schema which provides a DSL for changing type and validation rules.

## Testing done
unit tests, local integration tests

## Testing planned
staging integration tests

Replace a model with dry-structs, replace a form with dry-schema

#### Applies to all PRs

- [ ] Appropriate logging


modules/vaos/lib/vaos/exceptions/validation_errors.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+require 'common/exceptions/base_error'
+require 'common/exceptions/serializable_error'
+
+module VAOS
+  module Exceptions
+    class ValidationErrors < Common::Exceptions::BaseError
```
kreek: The base error class selects the i18n key based on the class name `ValidationErrors` => `common.exceptions.validation_errors`

saneshark:
Is it fair to say that we're forced to more or less punt on this given how we are using type coercions as both param / form validation and deserialization of incoming responses in some cases?

kreek:
@saneshark we could still use dry-types for VAMF API responses but yes for now forms would keep using virtus unless we switch to using schemas as in this PR.

kreek:
closing this for now

PR #3902, opened 2020-02-14
Logic for display of Community Care name

darkbushido:
## Description of change
Set the provider name based on the provider type

We need to display the correct name based on the provider type
it was always using ProviderName but that's only applicable to Individuals

Adding in the ability to trim all the fields we get from ppms
this will make it so there aren't any leading or trailing whitespaces
and it will remove extra spaces inside the value
`a  a -  b` should look like `a a - b`

department-of-veterans-affairs/va.gov-team#4762
## Testing done
I used the previously recorded VCR tapes to verify that we are getting the data and making the right decision.

## Testing planned
Verify it on a review instance

 - [x] When the ProviderType is GroupPracticeOrAgency display the value under CareSite
 - [x] When the ProviderType is Individual display the value under ProviderName
 - [x] When the trim param is added to a ccp request the values returned should have consistant spacing.

#### Applies to all PRs

- [ ] Appropriate logging


app/controllers/v0/facilities/ccp_controller.rb
```diff
@@ -43,9 +43,13 @@ def search_params
     params.permit(:type, :address, services: [], bbox: [])
   end
 
+  def trim_params
+    params.permit(:trim)
```
erluti: it looks like everywhere this is used, `[:trim]` is accessed.  Maybe this method should do that?

darkbushido: moved this to a feature flag

lib/facilities/ppms/client.rb
```diff
@@ -70,6 +82,21 @@ def specialties
 
       private
 
+      def trim_response_attributes(response)
```
brangi: Would it make sense to add an unit test for `trim_response_attributes` ?

darkbushido: since its a private method I don't want add direct tests for it. I am checking how this effects the other public methods. checkout `spec/lib/facilities/ppm/client_spec.rb`

PR #3904, opened 2020-02-14
Pr stats

thilton-oddball:
## Description of change
Adding a rake task that pulls PR data from Github. Will only be run in Jenkins and will eventually be removed when no longer needed.


rakelib/github_stats.rake
```diff
@@ -0,0 +1,50 @@
...
+].freeze
+
+namespace :github_stats do
+  desc 'Hit the Github API and grab data on open PRs'
+  task get_open_prs: :environment do
+    # hit API endpoints for vets-api and vets-website to get open PRs and add data to array
+    responses = []
+    REPOS.each do |repo|
+      url = "https://api.github.com/repos/department-of-veterans-affairs/#{repo}/pulls?state=open&per_page=100"
+      resp = HTTParty.get(url)
```
kfrz: Probably can sub in `Faraday` for this so we keep down on deps

Gemfile
```diff
@@ -112,6 +112,7 @@ gem 'zero_downtime_migrations'
 group :development do
   gem 'benchmark-ips'
   gem 'guard-rubocop'
+  gem 'httparty'
```
kfrz: Dislike, we already have `Faraday` and ruby has `Net::HTTP` , even though it's only in dev I don't see the need to add another external dependency

rakelib/github_stats.rake
thilton-oddball: Removed httparty and changed to:
resp = Net::HTTP.get(uri)
responses.concat(JSON.parse(resp))

Gemfile
thilton-oddball: Removed it.

.rubocop_todo.yml
```diff
@@ -35,55 +35,3 @@ RSpec/ContextMethod:
 # Configuration parameters: AllowConsecutiveOneLiners.
 RSpec/EmptyLineAfterExample:
   Enabled: false
-
```
kfrz: Did you mean to remove these? I.e. is this why the build is failing, or is this related to the rake task somehow?

thilton-oddball: Yes, when you run rubocop --auto-gen-config it adds any offenses that it finds to that todo file. Once you correct the offenses and rerun auto-gen-config it will no longer show up in there.

thilton-oddball: The reason it is failing is on some rubocop offenses that have nothing to do with any of the files I changed. I think they came in when I merged master, but I'm not quite sure how they could have gotten in master without having them specifically ignored in the rubocop.yml

thilton-oddball: Added the cops to the rubocop.yml and all pass now.

config/sidekiq_scheduler.yml
```diff
@@ -149,3 +149,9 @@ VaForms::FetchLatest:
   cron: "0 0 * * * America/New_York"
   class: VaForms::FormReloader
   description: "Fetches latest VA Forms"
+
+ExportGithubStats:
+  cron: "0 */2 * * * America/New_York"
+  class: GithubStatsJob
+  description: "Send Github open PR stats to prometheus"
+  enabled: <%= Settings.hostname == 'dev-api.vets.gov' %>
```
omgitsbillryan: I thought we were going with a Jenkins job to run the rake task instead a sidekiq scheduler job - what gives? (I'm not necessarily opposed to the scheduler strategy, I just didn't expect to see it)

thilton-oddball: Good catch. I totally forgot to remove that.

PR #3908, opened 2020-02-15
optionally lookup Accounts on sec_id, and update when needed

ericbuckley:
## Description of change
When looking up an Account for a User, we will now search on either the `idme_uuid` or the `sec_id`.  With the upcoming SSOe integration changes, there may be some authentication responses coming back from SSOe without an `idme_uuid`, thus we need another idenitifier to look them up by.

Additionally this request adds the `Account.update_if_needed!` method that will update the Account record if any of the User's id values have changed.  In practice we don't expect any values to change, however since `sec_id` was not previously stored, we need this functionality to backfill existing Account records with their `sec_id` value.

deps https://github.com/department-of-veterans-affairs/vets-api/pull/3930
closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/5289

#### Applies to all PRs

- [x] Appropriate logging


ericbuckley:
@patrickvinograd I still need to write test cases, but I was hoping to get your feedback on the following `TODOS` first.

- [PII data?](https://github.com/department-of-veterans-affairs/vets-api/pull/3908/files#diff-1299db718a4681d9e44c3708ebbab542R59)
- [missing redis field?](https://github.com/department-of-veterans-affairs/vets-api/pull/3908/files#diff-1299db718a4681d9e44c3708ebbab542R80)
- [unique index?](https://github.com/department-of-veterans-affairs/vets-api/pull/3908/files#diff-b4d704f6c4e7946101c5bc9027b38418R5)

db/migrate/20200214224815_index_accounts_on_sec_id.rb
```diff
@@ -0,0 +1,8 @@
+class IndexAccountsOnSecId < ActiveRecord::Migration[5.2]
+  disable_ddl_transaction!
+
+  def change
+    # TODO: should this be a unique index instead?
```
patrickvinograd: I don't think so - for the same reason that the icn/edipi indexes are not unique, we could have somebody with two credentials mapping to the same MVI identity and I think those would result in duplicate SECIDs.

app/models/account.rb
```diff
@@ -36,18 +39,54 @@ class Account < ApplicationRecord
...
+    attrs = account_attrs_from_user(user)
+
+    accts = where(idme_uuid: attrs[:idme_uuid])
+            .where.not(idme_uuid: nil)
+            .or(
+              where(sec_id: attrs[:sec_id])
+              .where.not(sec_id: nil)
+            )
+    if accts.length > 1
+      # TODO: are any ids in an Account record considered PII? if so we need
```
patrickvinograd: Generally all of these fields are low-sensitivity PII in the sense that they only really have value within VA. I would like to log at least the account UUIDs for each, and preferably a hash of all the other values so we can see where things are colliding.

app/models/account.rb
```diff
@@ -15,7 +16,9 @@ class Account < ApplicationRecord
   has_many :notifications, dependent: :destroy
 
   validates :uuid, presence: true, uniqueness: true
-  validates :idme_uuid, presence: true, uniqueness: true
+  validates :idme_uuid, uniqueness: true
```
patrickvinograd: I'm wondering if we need to relax this validation (and the corresponding database unique constraint if it exists).

app/models/account.rb
```diff
@@ -36,18 +39,54 @@ class Account < ApplicationRecord
...
+  end
+
+  def self.update_if_needed!(account, user)
+    # account has yet to be saved, no need to update
+    return account unless account.id
+
+    # return account as is if all user attributes match up to be the same
+    attrs = account_attrs_from_user(user)
+    return account if attrs.all? { |k, v| account.send(k) == v }
+
```
patrickvinograd: I would like to consider logging the old and new values of account/attrs here. Basically if we have a logic error or some misunderstanding of the constraints around identifiers in some external system, we potentially want to be able to reconstruct the updates that we are making to these records. 
If so we should call out in PR review that we are logging VA-internal identifiers to cloudwatch, which is allowed per ATO but should still be a conscious decision.

app/models/account.rb
ericbuckley: what use cases are you thinking that would cause a collision on idme_uuid?

app/models/account.rb
ericbuckley: will do, in general what critiera do you use to decided if the log should go to sentry or cloudwatch?

ericbuckley:
FYI, the tests are failing because I moved the [DB migrations into a seperate PR](https://github.com/department-of-veterans-affairs/vets-api/pull/3930) late last week.  Once that merges, and this updates with master, they should pass again.

app/models/account.rb
patrickvinograd: I'm not sure, just that we've seen almost every kind of data anomaly we can think of. I know we've seen the same user with two ID.me wallets, but that would be different ID.me UUIDs and hence not a collision. I guess the converse would be a person who winds up with two different SECIDs that map to the same ID.me UUID. I can't quite imagine the scenario that would lead to that. I guess for now leaving the constraint in place should be fine, we'd probably see a unique constraint violation get raised and visible in Sentry if it occurred.

patrickvinograd:
Looking good, thanks for the thorough test cases.

ericbuckley:
@patrickvinograd if it looks good, could you approve so I can ask the VSP team to review?

app/models/account.rb
```diff
@@ -36,18 +40,56 @@ class Account < ApplicationRecord
...
-      account.edipi = user&.edipi
-      account.icn   = user&.icn
+    attrs = account_attrs_from_user(user)
+
+    accts = where(idme_uuid: attrs[:idme_uuid])
+            .where.not(idme_uuid: nil)
+            .or(
+              where(sec_id: attrs[:sec_id])
+              .where.not(sec_id: nil)
+            )
```
omgitsbillryan: A named scope could be easier on the eyes here.

app/models/account.rb
```diff
@@ -36,18 +40,56 @@ class Account < ApplicationRecord
...
+      data = accts.map(&:attributes)
+      log_message_to_sentry('multiple Account records with matching ids', 'warning', data)
     end
+
+    accts.length.positive? ? accts[0] : create(**attrs)
+  end
+
+  def self.update_if_needed!(account, user)
+    # account has yet to be saved, no need to update
+    return account unless account.id
```
omgitsbillryan:
```suggestion
    return account unless account.persisted?
```
Does `persisted?` work here instead? Seems easier to read and could warrant removing the above comment.

app/models/account.rb
```diff
@@ -36,18 +40,56 @@ class Account < ApplicationRecord
...
+            .where.not(idme_uuid: nil)
+            .or(
+              where(sec_id: attrs[:sec_id])
+              .where.not(sec_id: nil)
+            )
+
+    if accts.length > 1
+      data = accts.map(&:attributes)
+      log_message_to_sentry('multiple Account records with matching ids', 'warning', data)
     end
```
omgitsbillryan: Could we instead create a custom uniqueness constraint on this model requiring that (`idme_uuid `+`sec_id`) must be  unique?

Also, is it ok to just continue marching on after finding more than 1 account? Some other options would be raising, or trying to consolidate the accounts into 1.

patrickvinograd: I think we're okay with constraints where we are. For a duplicate idme_uuid to occur, MVI would have to have mapped an identity on to different individuals at different times, associating the idme_uuid with two different secids. We should definitely log that to sentry and analyze if it does happen but I don't think it's likely, and I don't think the custom uniqueness constraint express much of use. 

If we do find more than one account, you're right that we probably need a heuristic to pick one other than the unstable ordering coming back from the database, but I'd also like to observe this and see if sentry picks up any of these events.

PR #3909, opened 2020-02-15
Fixing Form Urls and checking for valid_pdf

bastosmichael:
## Description of change
Resolves https://github.com/department-of-veterans-affairs/vets-contrib/issues/4172

## Testing done
- Fix existing urls and data associated with urls so that all data points to proper pdfs
- Remove issued on date

#### Applies to all PRs

- [ ] Appropriate logging


modules/va_forms/app/workers/va_forms/form_reloader.rb
```diff
@@ -44,49 +44,52 @@ def parse_table_row(row)
       if row.css('a').try(:first) && (url = row.css('a').first['href'])
         return if url.starts_with?('#') || url == 'help.asp'
 
-        begin
-          parse_form_row(row, url)
-        rescue
```
charleystran: @bastosmichael  removing the rescue here means that if an error is thrown on any form the reloaded dies. The purpose of the rescue was to log the bad form but still be able to move on to the next form. There are errors in the current urls for example some forms can only be accessed from the VA network so trying to open the URL throws an error.

modules/va_forms/app/workers/va_forms/form_reloader.rb
```diff
@@ -44,49 +44,52 @@ def parse_table_row(row)
...
+      end
+      form
+    rescue
+      form.valid_pdf = false
+      form
+    end
+
     def get_full_url(url)
-      "#{BASE_URL}/vaforms/#{url.gsub('./', '')}" if url.starts_with?('./va') || url.starts_with?('./medical')
+      "#{BASE_URL}/vaforms/#{url.gsub('./', '')}" if url.include?('./va') || url.include?('./medical')
```
charleystran: I am not sure what this change is solving. which URL's have ./va or ./medical that dont start with that string?

modules/va_forms/app/workers/va_forms/form_reloader.rb
bastosmichael: I was able to get it to successfully run all the way through without a single error which is why I removed it but what you're saying makes sense so I'll bring it back.

charleystran: yea, I mean if we could trust the dataset I would be less inclined to rescue, but as we know its a bit of a mess

modules/va_forms/app/workers/va_forms/form_reloader.rb
bastosmichael: I'm trying to account for any instance where the data set might include any sort of weird characters in front of the string before being parsed, I fixed most if not all of the failures associated with crawling the pages so let me reset this back and see if your original is good enough again.

modules/va_forms/app/workers/va_forms/form_reloader.rb
bastosmichael: I've tested this again and again and the data set is good now and every url is a proper pdf with a few of the minor tweaks to how we segment dates etc but good to be cautious here for sure.

modules/va_forms/app/workers/va_forms/form_reloader.rb
```diff
@@ -53,36 +53,43 @@ def parse_table_row(row)
...
       current_sha256 = form.sha256
-      form.title = title
-      issued_string = line.css('td:nth-child(3)').text
+      form.title = line.css('font').text
       revision_string = line.css('td:nth-child(4)').text
-      form.first_issued_on = parse_date(issued_string) if issued_string.present?
       form.last_revision_on = parse_date(line.css('td:nth-child(4)').text) if revision_string.present?
       form.pages = line.css('td:nth-child(5)').text
-      form_url = url.starts_with?('http') ? url : get_full_url(url)
+      form_url = url.starts_with?('http') ? url.gsub('http:', 'https:') : get_full_url(url)
```
charleystran: @bastosmichael  Are we sure all urls are https? Or they will be https?

edmangimelli: how about `sub` over `gsub`? should url be downcased first?

bastosmichael: Yes I was able to confirm that all 532 forms do support https and in fact the reason some of the forms were failingn previously is because not converting from http to https was actually not getting redirected properly from http to https. Don't know if that changes in the future but I doubt any new forms will be added without https support and redirection at least.

bastosmichael: I don't want to mess with the casing because not all VA systems are clean enough to understand downcased vs upcased etc. Leaving it as it is makes no assumptions as to the nature of the applications hosting these pdfs and gsub ensures that the linear structure of the swap is maintained moving forward.

PR #3915, opened 2020-02-19
[Forms] Remove deprecated property from v0/forms

ncksllvn:
## Description of change

This Lighthouse PR, https://github.com/department-of-veterans-affairs/vets-api/pull/3909, removed a property, `first_issued_on`, that was determined to not be valuable at all, so it's always `null`. We need to update our JSON schema for the va.gov route to reflect this. I could have changed it to allow `null` values, but I just removed it instead to be cleaner.

Error - http://sentry.vfs.va.gov/vets-gov/platform-api-staging/issues/101356/?query=is%3Aunresolved

https://github.com/department-of-veterans-affairs/va.gov-team/issues/6050

## Testing done

Locally ran this API to make sure it wasn't returning a 500 anymore.

## Testing planned

This API route isn't used in production right now, besides in a Drupal draft. We'll test in all environments.

#### Unique to this PR

- [x] _Replace this line with individual tasks unique to your PR_

#### Applies to all PRs

- [x] Appropriate logging


app/swagger/schemas/forms.rb
```diff
@@ -38,13 +38,6 @@ class Forms
                 key :example, 'Information for Pre-Complaint Processing'
               end
 
-              property :first_issued_on do
-                key :description, 'The date the form first became available'
-                key :type, :string
-                key :example, '2012-01-01'
-                key :format, 'date'
-              end
-
```
ncksllvn: Should I also update the cassette to reflect this property being removed? It's always `null` from Lighthouse right now.

johnpaulashenfelter: If its part of the response, it should probably stay in the cassette

ncksllvn: Sounds good!

ncksllvn:
Hey @johnpaulashenfelter I think is good to go?

lib/forms/schemas/forms.json
```diff
@@ -28,11 +28,8 @@
               "title": {
                 "type": "string"
               },
-              "first_issued_on": {
-                "type": "string"
-              },
               "last_revision_on": {
-                "type": "string"
+                "type": ["string", "null"]
```
ncksllvn: This field is still valid, but now null-able.

app/swagger/schemas/forms.rb
```diff
@@ -38,14 +38,7 @@ class Forms
...
-              property :last_revised_on do
+              property :last_revision_on do
```
ncksllvn: This was an inconsistency I caught between the Swagger docs and the actual response JSON.

ncksllvn:
This can be verified as working via the review instance - http://9eead66401523831f676457729ca0eb6.review.vetsgov-internal/find-va-forms-beta/?q=health

PR #3919, opened 2020-02-20
Fixing an issue where the provider uses a Name instead of ProviderName

darkbushido:
Expanding the responses hashes in the ccp_request_spec.rb
The mocks were hiding the errors

department-of-veterans-affairs/va.gov-team#6010

## Description of change
PPMS `v1.0/ProviderLocator` Endpoint returns a hash with the following keys
`CareSite` and `ProviderName`

PPMS `v1.0/Providers(1407842941)` Endpoint returns a hash with the following key
`Name`

`v1.0/ProviderLocator` uses `CareSite` and `ProviderName`
`v1.0/Providers(1407842941)` uses `Name`

With the provider name change pr #3902, it didn't look like anything was using `Name`
This is not the case and this PR addresses the issue.

 - [x] Facility or Provider name should appear above Facility type text on the Detail page
 - [x] The Detail page's title should match the provider name

#### Applies to all PRs

- [ ] Appropriate logging


app/serializers/provider_serializer.rb
```diff
@@ -13,17 +13,22 @@ def unique_id
...
   end
 
+  def fipper_enabled_name
+    possible_name = case object.ProviderType
+                    when /GroupPracticeOrAgency/i
+                      object.CareSite
+                    else
+                      object.ProviderName
+                    end
+    [possible_name, object.Name].find(&:present?)
```
omgitsbillryan: I find myself getting tripped up on this line - is this equivalent to this?

```ruby
possible_name || object.Name
```

omgitsbillryan:
Mind being more verbose in the OP description. I don't understand this:

>PPMS Response arrays use ProviderName
Single Responses use Name
For some reason Provider was getting Name

app/serializers/provider_serializer.rb
darkbushido: it is, `possible_name || object.Name` was causing an issue with codeclimate

omgitsbillryan: Gotcha, weird, but they're functionally equivelant, yes? Suppose I'm just asking for my own curiosity.

omgitsbillryan: Also weird that codeclimate seems to be OK with line 18: `object.ProviderName || object.Name`, but not here? 🤔

PR #3920, opened 2020-02-20
Fixing ccp pagination and changing the returned results to 10

darkbushido:
## Description of change
Attempting to fix an issue with pagination. 

Adding support for the per_page param
changing the default per page to 10

department-of-veterans-affairs/va.gov-team#5897

This also addresses 
department-of-veterans-affairs/va.gov-team#5823
## Testing planned
 - [ ] Check the review instance API to make sure the page and per_page params are working.

 - [ ] Search will return 10 results per page by default.
 - [ ] Data matches between the back end and what is displayed to the user on front.
 - [ ] Search results displayed are relevant to facility type searched
 - [ ] User experience includes pagination when the search returns more than 20 results
 - [ ] Pagination is displayed at bottom of the results list
 - [ ] Number of search results is shown
 - [ ] User can move to the next page of search results

#### Applies to all PRs

- [ ] Appropriate logging


app/controllers/v0/facilities/ccp_controller.rb
```diff
@@ -55,16 +55,20 @@ def api
...
   end
 
   def pages(ppms_results)
     total = ppms_results.length
     {
       current_page: page,
-      per_page: BaseFacility.per_page,
-      total_pages: Integer(ppms_results.length / BaseFacility.per_page + 1),
+      per_page: per_page,
+      total_pages: page + 1,
```
jholton: I'm confused here.  `total_pages` is always `current_page + 1`?
So, if we're on page 1, `total_pages` is 2 even if there are 7 pages worth of results...?

spec/request/v0/facilities/ccp_request_spec.rb
```diff
@@ -250,8 +302,8 @@
           'meta' => {
             'pagination' => {
               'current_page' => 1,
-              'per_page' => 20,
-              'total_pages' => 1,
+              'per_page' => 10,
+              'total_pages' => 2,
               'total_entries' => 10
```
jholton: This is illustrative of my questions from above:
* we are on page 1
* there are 10 results per page
* there are 10 results total

Shouldn't `total_pages` be 1?

app/controllers/v0/facilities/ccp_controller.rb
darkbushido: The PPMS response does not include any pagination metadata. This is a hack to allow the front end to request the next page of results.

spec/request/v0/facilities/ccp_request_spec.rb
darkbushido: - We are on Page 1
- We have no idea how many pages there are
- if we leave it as it is we will never request the next set of results.

jholton: My first thought was "Why do we have no idea how many pages there are?"
And then I went down the rabbit hole.
[Here](https://github.com/department-of-veterans-affairs/vets-api/blob/a5cf56e9ace70ab64c38acba41c8f87e0b16845c/lib/facilities/ppms/client.rb#L169) is why we don't know the total number of results:
`maxResults: 20 * page + 1`

We're requesting a different number of results based on what page the end user is on.

The logic for this ([PR](https://github.com/department-of-veterans-affairs/vets-api/pull/2595)) was
```
Tweaks the number of results to work around performance issues with the provider locator ppms endpoint identified in the staging environment. Results now limited to the amount necessary to display the current page (+1 so that a next page will show up in the front end if appropriate)
```

So, given all that, the logic for `total_pages` should be something like:
```ruby
total_pages = if ppms_results.length == per_page * page + 1 
  current_page + 1
else
  current_page
end
```
This should prevent the frontend from displaying a "next page" link when there is no next page.

We also need to ensure that `maxResults` is using the `per_page` value as determined in the controller instead of being hard coded as `20`.

jholton: And even with all that logic, the service responses can be weird.  I was playing around in a console:
```ruby
ppms = Facilities::PPMS::Client.new
params ={:address=>"366 Main Street, Waterville, Maine 04901, United States", :bbox=>["-70.39", "43.82", "-68.89", "45.32"], :type=>"cc_provider", :services=>["204D00000X"], :page=>"1", :per_page=>"20"}
providers = ppms.provider_locator(params)
providers.count
# => 21.
# OK, that means there is page 2 data.
params[:page] = "2"
providers = ppms.provider_locator(params)
providers.count
# => 32
# OK, that means we've retrieved all the results.  We would have received a count of 41 if there were page 3 results.  But maybe I'll try page 3 anyway
params[:page] = "3"
providers = ppms.provider_locator(params)
providers.count
# => 37
# WUT!?!?
```

app/controllers/v0/facilities/ccp_controller.rb
```diff
@@ -55,16 +52,20 @@ def api
...
-  def pages(ppms_results)
-    total = ppms_results.length
+  def pages(_ppms_results)
+    total = (page + 1) * per_page
     {
       current_page: page,
-      per_page: BaseFacility.per_page,
-      total_pages: Integer(ppms_results.length / BaseFacility.per_page + 1),
+      per_page: per_page,
+      total_pages: page + 1,
```
jholton: With our conversation this in mind, does it make sense for this value be `page` when `ppms_results.empty?`
Is it possible in that scenario that they go to the next page and get results?

PR #3921, opened 2020-02-20
WIP [5807] MDOT GET Supplies Endpoint

camerontesterman:
## Original issue
[Link to Issue 5807](https://github.com/department-of-veterans-affairs/va.gov-team/issues/5807)

## Description of change
Creates a new endpoint `GET /v0/mdot/supplies` which will allow veterans via the frontend app to pull in an array of medical devices and supplies available to them through our partners at the DLC.

## Testing done
All testing has been completed locally.

## Testing planned
When the DLC API get_supplies endpoint is available, we will be testing to ensure this endpoint successfully pulls in the data from DLC endpoint.

Through tests we can successfully demonstrate that an array of medical devices and supplies are returned to the veteran when `GET /v0/mdot/supplies` is called.

#### Applies to all PRs

- [x] Appropriate logging


app/controllers/v0/mdot/supplies_controller.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module V0
+  module MDOT
+    class SuppliesController < ApplicationController
+      before_action { authorize :mdot, :access? }
+
+      def index
+        supplies = client.get_supplies
```
omgitsbillryan: Do you know the timeout for this? I only ask because for synchronous calls, we need to be weary of long timeouts that can block requests and ultimately cause us to run out of puma threads.

camerontesterman: I am unsure at this point, the DLC haven't even created their side of the integration yet, so it may need to be tweaked in a later PR.

omgitsbillryan: OK, so long as you're aware. I don't think we have an official policy or anything in place, but I'll shoot from the hip and say that anything beyond the 10s mark is probably cause for moving this API call to an asynchronous one using a sidekiq worker.

camerontesterman: Yeah I think that is a fair point, I'll definitely keep that in mind once we start integrating and will refactor if necessary.

PR #3923, opened 2020-02-20
increase hca submission timeout

lihanli:
## Description of change
https://github.com/department-of-veterans-affairs/va.gov-team/issues/5596
i talked with josh faulkner on the hca team about this bug. he was able to change some server settings to fix the bug but then i ran into the issue of submissions with large attachments timing out. after changing the timeout setting i was able to successfully submit an application with a 9MB attachment on the staging server

## Testing done
tested on staging

#### Applies to all PRs

- [x] Appropriate logging


omgitsbillryan:
Is this service cal done synchronously? To where it could hold a thread in `vets-api` for 30 seconds?

@annaswims did we ever come to a consensus on allowable synchronous timeouts?

lihanli:
if the user is logged in or if they put in their email it submits asynchronously but otherwise it submits synchronously

omgitsbillryan:
Gotcha - would it be possible to have separate timeouts for each? 

It's NBD for a sidekiq worker job to hold for 30 seconds, but holding a `vets-api-server` request for that long can hog the available threads and lead to downtime.

lihanli:
well we need a longer timeout because a shorter one is not working with submissions with large attachments. if 30s is too long then we can implement a system like in `ClaimsBaseController` where it submits using a background job with no retries and then frontend polls to see if it succeeded or failed, that would take a lot more work though

PR #3924, opened 2020-02-20
Update duration calculation to use floats and rename metric

kfrz:
## Description of change
Updates the metric name and the way we're calculating duration :D 

## Testing done
Tested on the dev server
#### Applies to all PRs

- [ ] ~Appropriate logging~
- [ ] ~Swagger docs have been updated, if applicable~
- [ ] ~Provide link to originating GitHub issue, or connected to it via ZenHub~
- [ ] ~Provide which alerts would indicate a problem with this functionality (if applicable)~

### DIFF
--- a/rakelib/github_stats.rake
+++ b/rakelib/github_stats.rake
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-STATSD_ERROR_KEY = 'worker.github_stats_scraper.error'
+STATSD_METRIC = 'tasks.github_stats_scraper'
 
 REPOS = %w[
   vets-api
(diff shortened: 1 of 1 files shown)


rakelib/github_stats.rake
```diff
@@ -37,12 +37,12 @@ namespace :github_stats do
...
         end
       end
-      h['duration'] = (DateTime.now - DateTime.parse(h['updated_at']))
+      h['duration'] = (DateTime.now.to_f - DateTime.parse(h['updated_at']).to_f)
       open_prs << h
     end
     # send each duration to StatsD
     open_prs.each do |pr|
-      StatsD.measure('github:pull_request_duration', pr['duration'], tags: { repo: pr['repo_name'] })
+      StatsD.measure(STATSD_METRIC, pr['duration'], tags: { repo: pr['repo_name'] })
```
omgitsbillryan: Since this is measuring a duration, can we use the word "duration", "seconds", "time" or something?

rakelib/github_stats.rake
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-STATSD_ERROR_KEY = 'worker.github_stats_scraper.error'
+STATSD_METRIC = 'tasks.github_stats_scraper'
```
omgitsbillryan: J/w - is there anything else under `tasks.`?

kfrz: Nope, I opted to change it to `tasks` as it's not quite a worker and we don't have another similar bucket. 

Didn't spend a *ton* of time thinking about it but happy to change if there's a better idea

rakelib/github_stats.rake
kfrz: duration - good idea

PR #3925, opened 2020-02-20
[Forms] Treat queries as a list of terms

ncksllvn:
## Description of change

This PR modifies the API route `/services/va_forms/v0/forms`, which accepts a query parameter. Currently, query parameter operates by a strict match - if you search `health care`, then that full compound word must be present in the form's name. This PR modifies that behavior by splitting compound words by whitespace, so `health care` becomes `['health', 'care']`. Then, if a form name contains the word `health` or `care`, it is returned as a result. This way, a search for `health care` will pull up form 10-10EZ, the health care form. This is also necessary to achieve closer parity with the search function of the legacy forms search at `/vaforms/`.

https://github.com/department-of-veterans-affairs/va.gov-team/issues/5693
https://github.com/department-of-veterans-affairs/va.gov-team/issues/6048

## Testing done

I connected my local front-end website to it and did some queries. 

This query confirms that a query `health care` will render forms that contain either `health` or `care` in its name -

image

## Testing planned

Will be tested via Postman and using the FE app, /find-va-forms-beta, in all environments.

#### Unique to this PR

- [x] Work being done by VSA Public Websites but coordinated with Lighthouse team

#### Applies to all PRs

- [x] Appropriate logging


ncksllvn:
I don't know what's going on with the review instance here. It was working previously, but it doesn't look like it's using the changes in this PR anymore.

bastosmichael:
@ncksllvn is this actually giving you closer results to what you were looking for without any additional data?

ncksllvn:
@bastosmichael Yes sir! This seems to closely mirror what `/vaforms/` is doing, minus the DB access.

PR #3862, opened 2020-02-05
VAOS appointments openapi docs

kreek:
## Description of change
OpenAPI docs for the VAOS appointments endpoints. Should be merged after the systems PR https://github.com/department-of-veterans-affairs/vets-api/pull/3829

## Testing done
unit tests

## Testing planned
confirm docs are rendered correctly

## Acceptance Criteria (Definition of Done)
docs are publicly accessible

#### Applies to all PRs

- [ ] Appropriate logging
- [ ] Swagger docs have been updated, if applicable
- [ ] Provide link to originating GitHub issue, or connected to it via ZenHub
- [ ] Does not contain any sensitive information (i.e. PII/credentials/internal URLs/etc., in logging, hardcoded, or in specs)
- [ ] Provide which alerts would indicate a problem with this functionality (if applicable)

PR #3926, opened 2020-02-21
revert unintentional changes to Gemfile.lock

annaswims:
revert Gemfile.lock changes from https://github.com/department-of-veterans-affairs/vets-api/pull/3862


kreek:
@annaswims Thanks for fixing this 🤦‍♂wondering if we can add a spec that checks if the enterprise gem is bundled? e.g. 
```ruby
it 'has sidekiq enterprise in the bundle' do
  expect { Sidekiq::Limiter }.not_to raise_error(NameError)
end
```
It would require the gem to move from the production group in the Gemfile. Another option could be an initializer but that wouldn't get caught in specs.

PR #3928, opened 2020-02-21
block certain users from direct deposit

lihanli:
## Description of change
https://github.com/department-of-veterans-affairs/va.gov-team/issues/6093

## Testing done
automated tests

## Testing planned
test in staging

#### Applies to all PRs

- [x] Appropriate logging


lib/evss/ppiu/payment_information.rb
```diff
@@ -26,6 +26,16 @@ class PaymentInformation
...
+
+      delegate :authorized?, to: :control_information
+
+      def payment_account
+        authorized? ? @payment_account : EVSS::PPIU::PaymentAccount.new
+      end
+
+      def payment_address
+        authorized? ? @payment_address : EVSS::PPIU::PaymentAddress.new
+      end
```
omgitsbillryan: Can you help me understand why this is here? 

This part looks to be doing more than only restrict access to the `ppiu#update` endpoint. Are you also trying to filter out data from a response from some other endpoint as well?

omgitsbillryan: Also probably worth having a `payment_information_spec.rb` now that there's some logic in it.

lihanli: yes we also want to prevent unauthorized users from seeing the information in payment account and payment address

app/controllers/v0/ppiu_controller.rb
```diff
@@ -5,6 +5,7 @@ class PPIUController < ApplicationController
     before_action { authorize :evss, :access? }
     before_action { authorize :ppiu, :access? }
     before_action :validate_pay_info, only: :update
+    before_action(only: :update) { authorize(:ppiu, :access_update?) }
```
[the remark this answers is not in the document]
lihanli: `authorize(:ppiu, :access_update?)` needs to be passed in a block and the block has to go after the method arguments

lib/evss/ppiu/payment_information.rb
lihanli: thanks for the suggestion, i added `payment_information_spec.rb`

lihanli:
> Does swagger doc need to be updated since we added a new authz to the update endpoint?

the swagger already has an authorization parameter and error state, there doesn't seem to be anything specific for the evss policy or ppiu policy so not sure if anything is needed for the new update policy

lihanli:
> LGTM, but maybe you want to consider breaking this data into it's own endpoint? That way, a client would get a very obvious unauthorized http response as opposed to just getting `nil` data for direct deposit info and wondering why.

ok i can discuss it with the frontend engineers and look into doing it later

PR #3932, opened 2020-02-21
okta id does not have a specific format

annaswims:
## Description of change
okta has no set user id format, so we can just escape what comes after `api/v1/users/` for statsd

https://devforum.okta.com/t/is-user-id-unique-and-always-20-character-length/2487
## Testing done
no spec changes necessary


lib/breakers/statsd_plugin.rb
```diff
@@ -45,22 +45,19 @@ def filtered_endpoint_tag(path)
...
         |#{institution_ids}
-        |#{provider_ids}
-        |#{v1_user_ids_1}
-        |#{v1_user_ids_2})
+        |#{provider_ids})
         (\/|$)
       }x
 
-      path.gsub(r, '\1xxx\5')
+      path.gsub(r, '\1\3xxx\6')
```
omgitsbillryan: Can you explain the `\1\3xxx\6` part to me? I never quite understood this part in the code.

annaswims: yeah, it's in need of a serious comment/refactor

\1 refers to the first capture group and \6 refers to the 6th capture group.
a capture group is just the stuff matching the regex in parentheses.  So, it's just a complicated search and replace.

falling back to trusty https://rubular.com/
Screen Shot 2020-02-24 at 5 18 51 PM

annaswims:
@omgitsbillryan Is this any more easy to read?

PR #3933, opened 2020-02-21
disabling the MHV SSO shared cookie when logging in via SSOe

ericbuckley:
## Description of change
Authenticating via SSOe is going to set a cookie in the users browser to give them access to any IAM authenticated VA application, so there is no need to set the MHV SSO cookie.

closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/6038

## Testing done
- [x] manually tested authentication using the v0 session controller to confirm a MHV SSO cookie was created
- [x] manually tested authentication using the v1 session controller to cofirm no MHV SSO cookie was created

#### Applies to all PRs

- [ ] Appropriate logging


ericbuckley:
@patrickvinograd is this the type of change you we're thinking for disabling the MHV SSO cookie?

app/controllers/concerns/authentication_and_sso_concerns.rb
```diff
@@ -137,4 +138,13 @@ def sso_logging_info
       request_host: request.host
     }
   end
+
+  def can_set_sso_cookie?
+    return Settings.sso.cookie_enabled &&
+            @session_object.present? &&
+            # if the user logged in via SSOe, there is no benefit from
+            # creating a MHV SSO shared cookie
+            @current_user&.issuer !~ /https:\/\/.*eauth.va.gov\/[\S]+/
```
patrickvinograd: We use a slightly different regex in `lib/saml/user.rb#user_attributes_class` so might be good to reconcile those.

It feels a little off to have `issuer` bubble up as an attribute of `User` from an encapsulation perspective, but I think this is largely code that will get refactored/cleaned up as we standardize on SSO, so we can live with it for the time being.

ericbuckley: I'm open to other ideas if you can think of a better way to send info captured in the saml response to here

patrickvinograd: The only thing that comes to mind is a method like `User.has_va_sso` that can in turn delegate to identity and at least avoids this exact eauth issuer string bubbling up to this class.

patrickvinograd: I don't know if that's meaningfully better 🤔

ericbuckley:
> The code looks good, but I'm wondering if we can find a way to disambiguate the cookie we're referencing in `Settings.sso.cookie_enabled` (mhv shared cookie) from the kind of `sso` we're talking about in `has_va_sso` (ssoe)

good point, maybe just a name change?  replace the `has_va_sso` function with `authenticated_by_ssoe`?

lib/saml/user.rb
```diff
@@ -58,7 +58,7 @@ def to_hash
     private
 
     def serializable_attributes
-      %i[authn_context]
+      %i[authn_context has_va_sso]
```
omgitsbillryan: Does it make sense to instead include `issuer` as a serializable attribute as opposed to `has_va_sso`, that is, serailize the "data" and not the "logic on that data"? 

I could be misunderstanding how these `serializable_attributes` are used.

ericbuckley: there was a commit that tried that, https://github.com/department-of-veterans-affairs/vets-api/pull/3933/commits/27936b74cc8245d99688b91ceb5f789bb36358cb, however the thought was we we're just seperating the "eauth.va.gov" domain matching into 2 different places and if that regex ever needed to be changed it was more likely to cause problems having the logic in multiple modules.

app/controllers/concerns/authentication_and_sso_concerns.rb
```diff
@@ -84,9 +85,9 @@ def set_api_cookie!
 
   # Sets a cookie used by MHV for SSO
   def set_sso_cookie!
-    Rails.logger.info('SSO: ApplicationController#set_sso_cookie!', sso_logging_info)
+    return unless can_set_sso_cookie?
```
omgitsbillryan: Is "can" the right word? Might be more readable as simply `set_sso_cookie?`.

Or maybe a slightly over-engineered idea that just popped into my head:

```ruby
def set_sso_cookie!
  return unless can_set_sso_cookie? && should_set_sso_cookie?
  ...
end

def can_set_sso_cookie?
  Settings.sso.cookie_enabled && @session_object.present?
end

def should_set_sso_cookie?
  !@current_user&.has_va_sso
end
```

lib/saml/user.rb
omgitsbillryan: Ah, glad I wasn't being crazy. That makes sense. 👍

app/controllers/concerns/authentication_and_sso_concerns.rb
```diff
@@ -2,6 +2,7 @@
 
 # This module only gets mixed in to one place, but is that cleanest way to organize everything in one place related
 # to this responsibility alone.
+# rubocop:disable Metrics/ModuleLength
```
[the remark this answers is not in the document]
omgitsbillryan: Pls create a tech-debt ticket if you don't intend on fixing.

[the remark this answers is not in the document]
ericbuckley: I'll fix the rubocop issues, I should be able to consolidate some things to bring it under 100 LOC

ericbuckley:
@omgitsbillryan I [removed the rubocop wanrings](https://github.com/department-of-veterans-affairs/vets-api/pull/3933/commits/7c611b50f74fc2bd76ac3d0c3053d277f1bfd0f6) and [renamed to the has_va_sso function](https://github.com/department-of-veterans-affairs/vets-api/pull/3933/commits/b3dca7cbdd939e0ec2d89bdfde1b6706ce3d2c1b) to give better clarity as to whats being checked

PR #3934, opened 2020-02-22
Maps nil values for SSOe SAML assertions

patrickvinograd:
## Description of change
Corrects mapping of SSOe SAML paylods which use NOT_FOUND for empty values. 
Refactors SAML response specs to allow for using factories for SAML attributes to allow finer-grained tweaking of individual attribute values. I was otherwise finding it very difficult to write specs for things like testing the various mappings of the `gender` attribute but being able to override factory attributes per-spec makes it trivial. 

ResponseBuilder class is greatly simplified in favor of factories. Test cases are now responsible for instantiating the relevant attributes factory object, but otherwise all existing test case logic is preserved. 

Some of the SSOe specs are skipped pending being able to record the actual attribute sets that we get back for MHV and DSLogon credentials.

Fixes #6074

#### Applies to all PRs

- [ ] Appropriate logging


lib/saml/user_attributes/ssoe.rb
```diff
@@ -78,27 +79,27 @@ def loa_current
...
 
       # This is the ID.me highest level of assurance attained
       def loa_highest
-        attributes['va_eauth_ial']&.to_i
+        safe_attr('va_eauth_ial')&.to_i
       end
 
       def multifactor
-        attributes['multifactor']
+        safe_attr('va_eauth_multifactor') == 'true'
```
ericbuckley: should the comparison be case insenstive?

spec/factories/saml_attributes.rb
```diff
@@ -0,0 +1,466 @@
...
+  end
+
+  factory :mhv_premium, class: OneLogin::RubySaml::Attributes do
+    transient do
+      authn_context { 'myhealthevet' }
+    end
+    mhv_icn { ['1012853550V207686'] }
+    mhv_profile {
+      [
+        '{"accountType":"Premium","availableServices":{"21":"VA Medications","4":"Secure Messaging","3":"VA Allergies","2":"Rx Refill","12":"Blue Button (all VA data)","1":"Blue Button self entered data.","11":"Blue Button (DoD) Military Service Information"}}'
```
omgitsbillryan: Can we pretty this up with some line breaks for readability?

patrickvinograd: Done

PR #3935, opened 2020-02-24
Update Facilities Drivetime Bands

nfasulo:
## Description of change
This PR is for adding logic the the PSSG download job to update previously inserted drivetime bands. The reason that this logic was not included originally was the time required for each drivetime band to get updated was much longer than the time for inserting. The cause of this was the validation of the update, specifically the polygon itself which is massive (overall this dataset is 2 GB in Postgres).

To speed the processing of updates along I am using `update_columns` which writes directly to SQL without the extra rails validation and callbacks (the only of these add-ons I recreated manually was updating the `update_at` column). This allows the entire dataset to be processed in less than 90 minutes which is in line with the time it took to insert the bands in the first place.

Additionally, I ended up adding a default value to `vssc_extract_date`. This should be overwritten by the actual value from PSSG on insert, but adds a bit of resiliency in case that is ever missing. This migration does not need to be in place before the rest of the code is merged, so I did not create a separate PR.

Jira ticket: https://vasdvp.atlassian.net/browse/IN-38?atlOrigin=eyJpIjoiZjc0ZmYwMWQ2NTdlNDFmM2FjM2QwN2NmN2U2MzM3YmEiLCJwIjoiaiJ9

## Testing done
Ran PSSG download job on my machine and verified results.

## Testing planned
Will verify the values and counts of `vssc_extract_date` across environments after PSSG download job has been run and check the logs of the job to verify that nothing unexpected occurred.

#### Unique to this PR
- [x] drivetime bands are updated in step with PSSG API
- [x] new bands continue to be inserted
- [x] the speed of the job is not impacted by this additional processing

#### Applies to all PRs

- [x] Appropriate logging


spec/jobs/facilities/pssg_download_spec.rb
```diff
@@ -89,9 +99,10 @@
     end
 
     it 'leaves facility with original drive time band' do
+      create :vha_402QA
```
nfasulo: I found that because the facility wasn't being created drivetime bands were not actually being inserted for it. I changed this to a facility that already has a factory.

app/workers/facilities/pssg_download.rb
```diff
@@ -56,19 +56,31 @@ def extract_polygon(rings)
...
+        # Rails.logger.info "PSSG Band successfully inserted: #{@band_name}" # temporary logging
+      elsif @vssc_extract_date > drive_time_band.try(:vssc_extract_date)
+        # rubocop:disable Rails/SkipsModelValidations
+        drive_time_band.update_columns(
+          min: round_band(@attributes&.dig('FromBreak')),
+          max: round_band(@attributes&.dig('ToBreak')),
+          polygon: extract_polygon(@rings),
+          vssc_extract_date: @vssc_extract_date,
+          updated_at: Time.zone.now
+        )
```
edmangimelli: just out of curiosity, why skip validations?

app/workers/facilities/pssg_download.rb
```diff
@@ -56,19 +56,31 @@ def extract_polygon(rings)
...
         drive_time_band = facility.drivetime_bands.new(vha_facility_id: @id, name: @band_name)
         drive_time_band.unit = 'minutes'
         drive_time_band.min = round_band(@attributes&.dig('FromBreak'))
         drive_time_band.max = round_band(@attributes&.dig('ToBreak'))
-        drive_time_band.name = @band_name
         drive_time_band.polygon = extract_polygon(@rings)
+        drive_time_band.vssc_extract_date = @vssc_extract_date
 
-        Rails.logger.info "PSSG Band successfully saved: #{@band_name}" # temporary logging
         drive_time_band.save
```
edmangimelli: could this be?
```
        facility.drivetime_bands.create(
          vha_facility_id: @id,
          name: @band_name,
          unit: 'minutes',
          min: round_band(@attributes&.dig('FromBreak')),
          max: round_band(@attributes&.dig('ToBreak')),
          polygon: extract_polygon(@rings),
          vssc_extract_date: @vssc_extract_date
        )
```

app/workers/facilities/pssg_download.rb
nfasulo: We are skipping validations and callbacks because they take much longer than the update itself. For instance when running locally 30 records can be updated in around 3 seconds on average, but it takes more than 200 seconds when passing through all of the extra steps that Active Record includes.

The field `polygon` is a giant array of latitude and longitude points (a lot of bands have thousands of points). In Postgres these are stored as PostGIS geographic polygons that allow us to quickly compare them to individual lat/long points. We use the gem `rgeo-geojson` to represent these PostGIS objects in Active Model. There is a lot of overhead involved in converting and comparing between JSON and these objects. Even comparing an unchanged polygon to itself to see if the column needs to be updated is very slow. By instead just checking if our source has updated their polygon and replacing our cached polygon if so, we are able to greatly speed up this process. Additionally `extract_polygon` does will throw an error if the band is malformed, so we have already vetted the value.

I suppose we could update the rest of the record with validations and callbacks turned on, but that would me sending two separate update statements to Postgres, which is less than ideal.

app/workers/facilities/pssg_download.rb
nfasulo: `create` could definitely replace setting the properties individually for a new record and then saving (lines 62-69). I'm not clear on what `persisted?` is doing though. Won't it always return `false`?

app/workers/facilities/pssg_download.rb
```diff
@@ -56,19 +56,31 @@ def extract_polygon(rings)
...
+        # rubocop:disable Rails/SkipsModelValidations
+        drive_time_band.update_columns(
+          min: round_band(@attributes&.dig('FromBreak')),
+          max: round_band(@attributes&.dig('ToBreak')),
+          polygon: extract_polygon(@rings),
+          vssc_extract_date: @vssc_extract_date,
+          updated_at: Time.zone.now
+        )
+        # rubocop:enable Rails/SkipsModelValidations
+        # Rails.logger.info "PSSG Band successfully updated: #{@band_name}" # temporary logging
```
rtravitz: Any particular reason we're hanging on to these comments?

db/migrate/20200221144825_add_vssc_extract_date_drivetime_bands_default.rb
```diff
@@ -0,0 +1,5 @@
+class AddVsscExtractDateDrivetimeBandsDefault < ActiveRecord::Migration[5.2]
```
rtravitz: Do these migrations need to be extracted to a separate PR?

spec/jobs/facilities/pssg_download_spec.rb
rtravitz: Silent failure there could be difficult to debug in the future. Should we add a log?

db/migrate/20200221144825_add_vssc_extract_date_drivetime_bands_default.rb
nfasulo: Because this code is not affecting the database schema at all (not to mention the fact that this code isn't triggered by requests but rather a sidekiq job) it can be applied after the code has been merged. It is mostly just a paranoid attempt at handling a possible reversion in the PSSG API.

app/workers/facilities/pssg_download.rb
nfasulo: I left a few log lines commented out because although they are overkill for logging everyday, I thought they might be useful for someone coming to reason about this code or watch it run in the future. If it wasn't for the implicit `else` on line 82 I would have just deleted them them already. You are a good test case for that as you haven't looked at this code in months, so if you don't see any value in holding on to them, I can let them go.

spec/jobs/facilities/pssg_download_spec.rb
nfasulo: There actually is logging for this path through the logic. I just felt like not actually changing any drivetime bands wasn't fulling the promise of this test.
https://github.com/department-of-veterans-affairs/vets-api/blob/5d9af96ab8f37d4e01c9264a13e0e695200100bc/app/workers/facilities/pssg_download.rb#L36

nfasulo:
CI build failing due to security vulnerability in Puma. Once this PR is merged into master I will merge master into this branch again: https://github.com/department-of-veterans-affairs/vets-api/pull/3960

PR #3937, opened 2020-02-24
Setting the ProviderType for providers found by the POS locator

darkbushido:
## Description of change
We aren't setting the ProviderType on providers we get from the Place of Service locator
These providers are always GroupPracticeOrAgency
Provider Type is used to determine the name that gets displayed.

department-of-veterans-affairs/va.gov-team#6114
## Testing planned
Verify the front end displays the care site name instead of the provider name in a review instance.

#### Applies to all PRs

- [ ] Appropriate logging


omgitsbillryan:
The `vets-api` endpoint in question here is `/v0/facilities/ccp` - yes? 

If I understand properly, you're not modifying the structure of the JSON response of that endpoint. Instead, you're changing the how the response of the upstream service call to `PPMS` gets marshaled/de-serialized into our existing `data.attributes.name` json attribute?

darkbushido:
> The `vets-api` endpoint in question here is `/v0/facilities/ccp` - yes?
> 
> If I understand properly, you're not modifying the structure of the JSON response of that endpoint. Instead, you're changing the how the response of the upstream service call to `PPMS` gets marshaled/de-serialized into our existing `data.attributes.name` json attribute?

Correct

lib/facilities/ppms/client.rb
```diff
@@ -44,6 +44,7 @@ def pos_locator(params)
           providers = Facilities::PPMS::Response.from_provider_locator(response, request_params)
           providers.each do |provider|
             provider.posCodes = request_params[:posCodes]
+            provider.ProviderType = 'GroupPracticeOrAgency'
```
omgitsbillryan: I think a "why-style" comment is warranted here. It took me a good bit of time to realize that "GroupPracticeOrAgency" is a special string used in the provider serializer.

PR #3939, opened 2020-02-24
adding PR number and user to github_stats.rake

thilton-oddball:
## Description of change
Added PR number and user to the data that gets passed to Prometheus in github_stats.rake


kfrz:
@omgitsbillryan I like the refactor, and your "psuedocode" looks a lot like valid ruby :D

PR #3940, opened 2020-02-24
Add default CircleCi configuration

johnpaulashenfelter:
## Description of change

Evaluating CircleCI build process for vets-api.

#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging


.circleci/config.yml
```diff
@@ -0,0 +1,15 @@
+version: 2.1
+orbs:
+  ruby: circleci/ruby@0.1.2 
+
+jobs:
+  build:
+    docker:
+      - image: circleci/ruby:2.6.3-stretch-node
```
johnpaulashenfelter: I know this is slightly different than our current one. Digging into the docs to see what our options are.

annaswims:
all PR's now show " Your tests failed on CircleCI" should we unto the status check until this is merged?

omgitsbillryan:
Is there a ticket this is connected to? Also, any additional context you could provide as to why we're doing this would help greatly.

johnpaulashenfelter:
@omgitsbillryan issue is upcoming. readout from devops was heading this way, frontend is trying as well. seems like a small lift but this is acting weird.

johnpaulashenfelter:
Closing because there's no easy way to do in *this* repository without breaking deploys (review instance deploys at least). Jenkins expects *all* checks to be passing and since this CircleCi check will be broken until it works, AND since every branch gets the same set of checks including CircleCI we end up in a broken situation.

The right way to do this is figure out the issues on a fork

PR #3944, opened 2020-02-25
Introduce jwt session extending middleware - Phase 1

saneshark:
## Description of change
The goal here is to introduce a middleware that will allow for the JWT token to automatically be updated based on a refreshed token we receive on each successful response. This is an incremental step in that direction by consolidating to a single configuration. I've tested this thoroughly on review instances.

## Testing done
Lots of testing on review instance and detached staging instance.

## Testing planned
verify on staging and production that everything is working as expected.

This is just phase 1. Incremental step needed to introduce the additional middleware, but the changes would get too gnarly for review in a single PR.

#### Applies to all PRs

- [x] Appropriate logging


lib/common/client/middleware/request/camelcase.rb
```diff
@@ -10,7 +10,7 @@ def initialize(app)
           end
 
           def call(env)
-            env[:body] = camelcase(env[:body])
+            env[:body] = camelcase(env[:body]) unless env.request_headers['Content-Type'] == 'text/plain'
```
kreek: All connections that use camelcase also use json, and this middleware assumes it's working with an Array or a Hash (json_has is the arg), so maybe it should be run only if it's 'application/json'? (unless we have VA services returning JSON with mislabelled content-types, which is entirely possible) 😄

saneshark: I thought about that -- but then I worried that there might be some weird hosts out there that don't respond well to "Content-Type" headers and we'd be excluding those.

PR #3946, opened 2020-02-25
Move VAOS cert file read to configuration and memoize it

kreek:
## Description of change
Moves the VAOS cert that is used for signing JWT requests to the base service configuration file.

## Testing done
unit tests

## Testing planned
staging integration tests

does not break things and loads the cert file only once per server instance

#### Applies to all PRs

- [ ] Appropriate logging


modules/vaos/app/services/vaos/configuration.rb
```diff
@@ -14,6 +14,10 @@ def service_name
       'VAOS'
     end
 
+    def cert_key
+      @key ||= OpenSSL::PKey::RSA.new(File.read(Settings.va_mobile.key_path))
+    end
```
omgitsbillryan: `VAOS::Configuration` is a singleton, and therefore any instance variables in the singleton are cached for the life of the app... am I following that right? Just want to make sure this doesn't get loaded over and over again.

Secondly, not very important, but I think `key` is a more succinct name for this method.

kreek: yes it's a singleton, fairly long inheritance chain but the base class is `Common::Client::Configuration::Base`

saneshark: I can confirm, it is a singleton and therefore should not be loaded over and over again.

modules/vaos/app/services/vaos/configuration.rb
```diff
@@ -14,6 +14,10 @@ def service_name
       'VAOS'
     end
 
+    def key
```
saneshark: NITPICK: possibly consider calling this "cert" or "private_key" as "key" is kind of generic and could be confused with "hash key"

modules/vaos/app/services/vaos/jwt.rb
```diff
@@ -14,7 +14,7 @@ def initialize(user)
     end
 
     def token
-      ::JWT.encode(payload, rsa_private, 'RS512')
+      ::JWT.encode(payload, Configuration.instance.key, 'RS512')
```
saneshark: NITPICK: is Configuration as clear as "VAOS::Configuration?" I guess so.

saneshark:
Also, just FYI, I have a PR in the works where I'm trying to remove "user_configuration" so it's nice that you did this as part of the main "configuration" and not "user_configuration"

modules/vaos/app/services/vaos/configuration.rb
kreek: Used to be called `cert_key` but changed it as I liked @omgitsbillryan suggestion to make it more succinct 😃Variable naming ~fight~ discussion... go! 🥊

kreek: `rsa_key` makes sense to me

PR #3949, opened 2020-02-26
GitHub stats modifications

thilton-oddball:
## Description of change
Refactored github_stats.rake to get a much more accurate calculation of the duration between when a PR is opened and when it is first responded to.


rakelib/github_stats.rake
```diff
@@ -31,15 +42,29 @@ namespace :github_stats do
...
+        review_comments_url = pr['review_comments_url']
+        reviews_url = "#{pr_url}/reviews"
+        # determine first_responded_at date
+        first_responded_at = if comments.positive?
+                               get_first_responded_at(comments_url)
+                             elsif review_comments.positive?
+                               get_first_responded_at(review_comments_url)
+                             else
+                               get_first_responded_at(reviews_url)
+                             end
```
omgitsbillryan: What about a scenario where a review occurs, then sometime later a regular comment occurs? This if-else logic block would determine that the comment came first even though the review came before it. Am I reading that right?

omgitsbillryan: Is it correct to assume that if a comment exists, that it must have come first?

thilton-oddball: Yeah, I think you're right. In the first hit that pulls everything, we have access to comment counts and review_comment counts for each PR but, for whatever reason, not review counts. In trying to limit the number of times we have to hit the API I figured I'd check the comment and review_comment counts first, since if one of them is positive then we hit the specific endpoint for it. If both are zero then we can hit the review endpoint and determine if there are either zero reviews (nothing has been done at all with the PR yet) or one or more reviews, in which case we grab the first and get its date. We'll have to increase the number of hits to the API but I think I know what I need to do to account for the scenario you describe.

thilton-oddball: After discussing with my group, we determined that we should only calculate duration based on actual reviews and not on comments or review_comments, which fortunately makes this a lot simpler.

rakelib/github_stats.rake
```diff
@@ -33,13 +25,25 @@ namespace :github_stats do
...
-        repo_name = repo
-        updated_at = pr['updated_at']
+        pr_created_at = DateTime.parse(pr['created_at']).to_f
+        pr_url = pr['url']
+        reviews_url = "#{pr_url}/reviews"
+        # determine first_reviewed date
+        uri = URI(reviews_url)
+        resp = Net::HTTP.get(uri)
+        first = JSON.parse(resp).first
+        first_reviewed = if first.nil?
```
omgitsbillryan:
```suggestion
        first_reviewed_at = if first.nil?
```

I like the `_at` suffix to indicate time. NBD though.

rakelib/github_stats.rake
omgitsbillryan: Interesting. I can provide my own anecdote that I sometimes will first respond with plain old comments if I don't understand something. I find that getting all my comments in, and then submitting my review might take as long as 45 minutes! Whereas if I make simple comments one by one, the author may respond in real time as I am still reviewing.

PR #3951, opened 2020-02-26
Add exclusion to dependabot for vets_json_schema

johnpaulashenfelter:
## Description of change

Some of our gems should never be auto-updated by dependabot. This PR adds vets_json_schema as the first of those gems

## Testing planned

None necessary other than YAML validation


annaswims:
I still approve 👍

PR #3954, opened 2020-02-27
Cache global downtimes

U-DON:
## Description of change
When the API goes down for maintenance, the website needs to display the proper messaging, but it needs to access the downtime schedule without invoking the `maintenance_windows` endpoint. Currently, it's relying on a hard-coded time frame in the source, which isn't ideal.

The solution here is to get that data from PagerDuty so we can dynamically show the global downtime notification. We didn't want to have the website excessively invoke the PD API with direct requests, so instead, we're going with a job that periodically makes that request and caches the response to an S3 bucket that the website can read.

Before this can be merged, we need to have the service added to PD and the bucket set up for storing the JSON file.

Changes:
- Updated configuration to recognize the new `global` service from PagerDuty ("Internal: API").
- Created a job to get only the global downtimes from PD and store them in a JSON file in S3.

## Testing done

Added unit tests.

## Testing planned

Will verify that the file becomes publicly accessible and contains the correct data.

#### Unique to this PR

- [ ] Global service has been added to PagerDuty.
- [ ] Buckets have been set up for storing the downtime JSON file.

#### Applies to all PRs

- [ ] Appropriate logging


config/sidekiq_scheduler.yml
```diff
@@ -104,6 +104,11 @@ FacilityLocationVHADriveTime:
   class: Facilities::PSSGDownload
   description: "Download and store drive time bands"
 
+GlobalDowntimeCache:
+  every: "60m"
```
U-DON: Initial thought was to have it run every hour, but open to changing this if a different frequency makes more sense.

omgitsbillryan:
My first thought - why not cache in redis instead of creating a new bucket? Are you trying to keep the number of calls to vets-api down?

Such a call to vets-api to retrieve a value from redis would be quite performant.

lib/pagerduty/maintenance_client.rb
```diff
@@ -6,30 +6,32 @@ module PagerDuty
...
       while resp['more']
         offset = resp['offset'] + resp['limit']
-        resp = get_raw(offset = offset)
+        resp = get_raw(options.merge('offset' => offset))
         windows += resp['maintenance_windows']
       end
       windows
     end
 
-    def get_raw(offset = 0)
```
omgitsbillryan: Were there any instances of this being called where `offset` was being overridden to something other than 0 that will now need to be updated to `{offset: 0}`?

U-DON:
@omgitsbillryan: We're caching this to S3 so the website can access the downtime schedule when vets-api is down, so it can put up the right messaging. Something like, "maintenance is expected to end at `#{end_time}`".

lib/pagerduty/maintenance_client.rb
U-DON: We're only overriding the offset in the loop where we're reading through everything by incrementing the offset. It's seems to be isolated since `get_raw` is private and `get_all_raw` (also private) was the only place where it was called.

PR #3955, opened 2020-02-27
Improve logging for 526 timeouts

edmangimelli:
Relates to [DASH-61](https://vasdvp.atlassian.net/secure/RapidBoard.jspa?rapidView=140&projectKey=DASH&modal=detail&selectedIssue=DASH-61&assignee=5e386d921b1d910e5dfd793b
)
>Tim Barto is having intermittent upstream errors happening in production when he hits v0/forms/526. ...

Also cleaned up some tech debt / made rubocop happier about `#request`.


lib/common/client/base.rb
```diff
@@ -119,10 +117,26 @@ def request(method, path, params = {}, headers = {}, options = {}) # rubocop:dis
...
+              body: resp.body,
+              headers: resp.headers
+            }
+          }
+        end
 
-        response_hash = e.response&.to_hash
-        client_error = error_class.new(e.message, response_hash&.dig(:status), response_hash&.dig(:body))
-        raise client_error
+        { response: resp.try(:to_hash) || resp.try(:to_h) }
```
edmangimelli: If this looks paranoid, it's because I wanted it to be as general-purpose as possible. Given any error (Faraday or not), it will try to extract a response hash. That leaves room for us to author our own errors that conform somewhat to Faradays (which is what we might be doing already?)
Also, I noticed we were already using `to_hash`, but I noticed in _Gemfile.lock_ we're using Faraday *0.17.0*.
`Response` s in that version of Faraday don't have a `to_hash` method (it wasn't introduced until `1.0.0`).
Therefore, to be maximally, safe, I try to use the `status`, `body`, and `headers` first (which are in all versions).
https://www.rubydoc.info/gems/faraday/0.17.0/Faraday/Response
https://www.rubydoc.info/gems/faraday/Faraday/Response

bastosmichael: Should we be doing a Faraday Response object instead of just a hash if we are being paranoid?

edmangimelli: I'm not sure if it would be serialized when handed to sentry, if that makes sense.
maybe tacking it on in addition to the hash would be good

lib/common/client/base.rb
```diff
@@ -119,10 +117,26 @@ def request(method, path, params = {}, headers = {}, options = {}) # rubocop:dis
...
+      def response_hash_from_error(error)
+        return { response: nil } unless error.respond_to?(:response)
+
+        resp = error.response
+
+        if %i[status body headers].all? { |method| resp.respond_to? method }
+          return {
+            response: {
+              status: resp.status,
+              body: resp.body,
```
jholton: What kind of data is in a `Timeout::Error` or  `Faraday::TimeoutError` response?
As Michael pointed out most external service calls go through this code. 
We don't want PII ending up in Sentry.

jholton:
Can you provide more details as to the motivation for this?
I can't open the linked ticket.

charleystran:
Hey @jholton I might be able to provide some context and I am glad you all are involved in this PR. We have a random failing call to EVSS for our 526 auto establishment. The consumer can't consistently reproduce it.  And since all we know is that the claim didn't establish, we dont know enough to attempt to reproduce it. The purpose of more logging is to raise up the response from EVSS so we might be able to find the failed calls and work from there. There very well could be a better way to do that so if you have thoughts we are open to ideas.

jholton:
Happy to make suggestions.
Do the requests in question always result in a timeout?  I'm curious what info you can get from logging a a timeout error "response".  Presumably a response wasn't received from the external service, hence the timeout.  Right?

charleystran:
@jholton we are not sure to be honest. However timeouts are timeouts. If that is problem it would just be good to know it and explain that the customer. This PR also catches Common::Exceptions::BackendServiceException which is more of our concern. I 100% agree we dont want PII in the logs but we just want to captures these failed responses in some manor.

lib/common/client/base.rb
edmangimelli: >We don't want PII ending up in Sentry.

Good point! And not sure what those errors hold --I'd like to find out!

edmangimelli:
@jholton @charleystran 
>We have a random failing call to EVSS for our 526 auto establishment.

I _think_ it's EVSS timeouts. I've seen EVSS timeouts in the logs as I've been hunting, but I'm not 100% certain.

>Do the requests in question always result in a timeout?

Not sure. I think logging these timeout response will help us figure that out.

>This PR also catches Common::Exceptions::BackendServiceException which is more of our concern.

This PR does _not_ do that. That exception was already being caught, and I'm not adding anything to that rescue block (it just got rubocopped).

lib/common/client/base.rb
```diff
@@ -106,10 +106,9 @@ def request(method, path, params = {}, headers = {}, options = {}) # rubocop:dis
         end.env
       rescue Common::Exceptions::BackendServiceException => e
         # convert BackendServiceException into a more meaningful exception title for Sentry
-        raise config.service_exception.new(
-          e.key, e.response_values, e.original_status, e.original_body
-        )
-      rescue Timeout::Error, Faraday::TimeoutError
+        raise config.service_exception.new(e.key, e.response_values, e.original_status, e.original_body)
+      rescue Timeout::Error, Faraday::TimeoutError => e
+        PersonalInformationLog.create(data: response_hash_from_error(e), error_class: e.class)
```
jholton: Can we move all this into area of code that is making the service calls you care about?  It's the "526 auto establishment" calls, right?  Where does that happen?  `modules/claims_api/app/workers/claims_api/claim_establisher`? Is that it?

Just rescue all errors there, log them, and re-raise.

If you leave it here, any `Timeout::Error` or `Faraday::TimeoutError` from _any_ service will get logged.
Also, your `error_class` value should basically be treated as a key/value to allow you to easily identify your records of interest from the others.  Call it "EdsErrors" if you want. 😄 
As is right now, all of your entries will have an `error_class` of `Timeout::Error` or `Faraday::TimeoutError`, which will make it very hard to distinguish the ones you care about from the others.

Additionally, looking at Sentry I don't see a lot of `Common::Exceptions::GatewayTimeout` errors (which gets raised 2 lines below.  I do see a whole bunch of `Faraday::TimeoutError` errors.  [Example](http://sentry.vfs.va.gov/vets-gov/platform-api-production/issues/101527/?query=%22Faraday%3A%3ATimeoutError%22).  Looks like the `rescue` isn't catching them currently. Not sure why.

edmangimelli: @jholton 
> Can we move all this into the area of code that is making the service calls you care about?  It's the "526 auto establishment" calls, right?  Where does that happen?  `modules/claims_api/app/workers/claims_api/claim_establisher`? Is that it?

I was advised to plop my code into `lib/common/client/base.rb`, but, yeah, putting it somewhere closer to the 526 code makes a lot more sense (although I ended up in a slightly different spot).

>... Call it "EdsErrors" if you want. 😄 

That _does_ have a nice ring to it.


> Additionally, looking at Sentry I don't see a lot of `Common::Exceptions::GatewayTimeout` errors (which gets raised 2 lines below).  I do see a whole bunch of `Faraday::TimeoutError` errors.  [Example](http://sentry.vfs.va.gov/vets-gov/platform-api-production/issues/101527/?query=%22Faraday%3A%3ATimeoutError%22).  Looks like the `rescue` isn't catching them currently. Not sure why.

That is pretty curious. I wonder if moving this line up:
```
rescue Timeout::Error, Faraday::TimeoutError
```
or adding `Net::ReadTimeout` would help. 🤷‍♀️

Also,
**thank you so much for all of your input!** It's been invaluable.

lib/evss/disability_compensation_form/service.rb
```diff
@@ -50,7 +78,18 @@ def submit_form526(form_content)
         with_monitoring_and_error_handling do
           headers = { 'Content-Type' => 'application/json' }
           options = { timeout: Settings.evss.disability_compensation_form.submit_timeout || 355 }
-          raw_response = perform(:post, 'submit', form_content, headers, options)
+          begin
+            raw_response = perform(:post, 'submit', form_content, headers, options)
+          rescue Net::ReadTimeout, Timeout::Error, Faraday::TimeoutError => e
```
jholton: Hmmm, theoretically I don't think this `rescue` should ever fire for `Timeout::Error` and `Faraday::TimeoutError`.
The `rescue` in `lib/common/client/base` should catch them and raise `Common::Exceptions::GatewayTimeout`.
So the timeout errors never make it to this level. There's [a spec for this scenario](https://github.com/department-of-veterans-affairs/vets-api/blob/d28cc999502a96a58f60378e415deb64a4290448/spec/lib/evss/disability_compensation_form/service_spec.rb#L30).

However, as I pointed out, there are a bunch of timeout errors in Sentry.  Will this catch them?  Or will they not get rescued for the same reason their getting past `common/client/base`?  I don't know.
Maybe play around with some code like [this SO answer](https://stackoverflow.com/questions/56649234/how-i-can-generate-faraday-timeout-error-manually-in-ruby-on-rails).

Also, you're trying to log `raw_response` in this rescue. I believe it's value will be `nil`.  It never gets set since an error occurred inside the method that is supposed to return it's value.  Thinking about it broadly, if you get a timeout error, there is no response to analyze.  None was received because the request timed out.

Any rescuing/logging you do in this file, you have to account for any manipulation (`rescue`/`raise` stuff) that happens in `common/client/base`.  Likewise anything that happens in `common/client/base` is dependent on what happens in the Faraday middlewares. That's the closest you can get to the actual request/response.  That's probably why the `:logging` middleware was written.

edmangimelli: >Hmmm, theoretically I don't think this rescue should ever fire for Timeout::Error and Faraday::TimeoutError.
The rescue in lib/common/client/base should catch them and raise Common::Exceptions::GatewayTimeout.

ugh. I forgot that all the network calls were going through there.

>Also, you're trying to log raw_response in this rescue. I believe it's value will be nil. It never gets set since an error occurred inside the method that is supposed to return it's value. Thinking about it broadly, if you get a timeout error, there is no response to analyze. None was received because the request timed out.

this was just for completeness sake, but also, i don't know if something beyond EVSS could timeout, and then EVSS would really return a response (that describes a further upstream timeout). don't know. just wanted to cover my bases

modules/claims_api/app/controllers/claims_api/v0/forms/disability_compensation_controller.rb
```diff
@@ -35,6 +35,12 @@ def submit_form_526
         rescue EVSS::ErrorMiddleware::EVSSError => e
           track_526_validation_errors(e.details)
           render json: { errors: format_errors(e.details) }, status: :unprocessable_entity
+        rescue ::Common::Exceptions::GatewayTimeout, ::Timeout::Error, ::Faraday::TimeoutError => e
+          req = { auth: auth_headers, form: form_attributes, source: source_name, auto_claim: auto_claim.as_json }
+          PersonalInformationLog.create(
+            error_class: "submit_form_526 #{e.class.name}", data: { request: req, error: e.try(:as_json) || e }
+          )
+          render json: { errors: [{ status: 504, detail: 'Timeout', source: nil }] }, status: :gateway_timeout
```
jholton: Someone on your team should chime in here (@charleystran ?), but why explicitly `render` here instead of re-raising the error and letting the existing code path handle it?

edmangimelli: ah. that change happened probably when i was troubleshooting why my tests were passing locally but failing in the pipeline (ended up being a weird scope resolution thing).
originally it was just re-raising

edmangimelli: reverted

PR #3956, opened 2020-02-27
Filter out unneeded identifiers in VAOS systems and messages services

jbalboni:
## Description of change
We have ICNs and other identifiers coming back in the systems service. We also have an ICN coming back in the messages service in the sender_id field. This PR removes those identifiers, so that the front end never sees them.

## Testing done
Unit testing

#### Unique to this PR
- [ ] systems service only returns dfn and CRNR ids
- [ ] messages services does not return sender_id

#### Applies to all PRs

- [ ] Appropriate logging


modules/vaos/app/services/vaos/messages_service.rb
```diff
@@ -19,6 +19,7 @@ def post_message(request_id, request_object_body)
         response = perform(:post, messages_url(request_id), params, headers)
 
         if response.status == 200
+          response.body.delete(:sender_id)
```
jbalboni: Is this ok syntax? I'm not sure exactly how you note if blocks in Ruby.

modules/vaos/app/services/vaos/systems_service.rb
```diff
@@ -7,7 +7,10 @@ class SystemsService < VAOS::BaseService
     def get_systems
       with_monitoring do
         response = perform(:get, '/mvi/v1/patients/session/identifiers.json', nil, headers)
-        response.body.map { |system| OpenStruct.new(system) }
+        response
+          .body
+          .select { |system| system[:assigning_authority].include?('dfn-') || system[:assigning_code].include?('CRNR') }
```
jbalboni: We would need to allow Cerner ids through in the future, so I made an attempt at that now. Could remove it if we don't want to go there yet.

modules/vaos/app/services/vaos/messages_service.rb
kreek: Yep that's ok. Rails (via the active_support gem) adds an `except` method to Ruby Hashes, so you could also do `response.body.except(:sender_id)`.  Realize the if/else logic was there before, but that could also use a guard clause to handle the non 200 case upfront:

```ruby
handle_error(response) unless response.status == 200

{
  data: OpenStruct.new(response.body.except(:sender_id)),
  meta: {}
}
```

or `if response != 200` if you're allergic to `unless` 😄

modules/vaos/app/services/vaos/messages_service.rb
```diff
@@ -18,14 +18,12 @@ def post_message(request_id, request_object_body)
...
-          }
-        else
-          handle_error(response)
-        end
+        handle_error(response) unless response.status == 200
+
+        {
+          data: OpenStruct.new(response.body.except(:sender_id)),
+          meta: {}
+        }
```
omgitsbillryan: I think this change introduces some coupling with the `handle_error` method. It assumes that `handle_error` will exit the code execution (which it currently does do by raising), but that assumption might not always be true.

modules/vaos/spec/services/messages_service_spec.rb
```diff
@@ -29,7 +29,7 @@
       end
     end
 
-    it 'handles 500 errors approprietaly' do
+    it 'handles 500 errors appropriately' do
```
omgitsbillryan: 😉

PR #3957, opened 2020-02-27
check for existence of `Sidekiq::Pro` when configuring `super_fetch!`

jholton:
## Description of change
Change logic in enabling `super_fetch!` to check for existence of `Sidekiq::Pro` class instead of evaluating `ENV['EXCLUDE_SIDEKIQ_ENTERPRISE']`.

The old way is causing issues for users who are running the app via Docker who don't have the env variable set at runtime.

## Testing done
Tested various ways to start the app (docker vs native) with various configurations (credentials, env variables)

## Testing planned
N/A

#### Unique to this PR

- [ ] When enabling `super_fetch!` to check for existence of `Sidekiq::Pro` class instead of evaluating `ENV['EXCLUDE_SIDEKIQ_ENTERPRISE']`

#### Applies to all PRs

- [ ] ~~Appropriate logging~~


config/initializers/sidekiq.rb
```diff
@@ -4,7 +4,7 @@
 
 Sidekiq.configure_server do |config|
   config.redis = REDIS_CONFIG['redis']
-  config.super_fetch! unless ENV['EXCLUDE_SIDEKIQ_ENTERPRISE'] == 'true'
+  config.super_fetch! if defined?(Sidekiq::Pro)
```
omgitsbillryan:
```suggestion
  # super_fetch! is only available in sidekiq-pro and will cause 
  #   "undefined method `super_fetch!'"
  # for those using regular sidekiq
  config.super_fetch! if defined?(Sidekiq::Pro)
```

omgitsbillryan: Seems worthy of a "why-style" comment.

PR #3958, opened 2020-02-27
increase hca service timeout and submit all hcas with background job

lihanli:
## Description of change
https://github.com/department-of-veterans-affairs/va.gov-team/issues/5596
increases timeout to fix hca submission bug, also will submit hca applications via a background job with no retries if the user does not put their email in the form

## Testing done
automated tests

## Testing planned
test in staging

#### Applies to all PRs

- [x] Appropriate logging


app/controllers/v0/health_care_applications_controller.rb
```diff
@@ -10,7 +10,7 @@ def create
       validate_session
 
       health_care_application = HealthCareApplication.new(params.permit(:form))
-      health_care_application.async_compatible = params[:async_compatible]
+      health_care_application.async_compatible = params[:async_all]
```
jholton: Is there going to be any lag in the frontend making this param change?  i.e. should you check for both for a while?

lihanli: the `async_compatible` param was used to migrate the frontend from all synchronous submissions to async if the user has an email. right now it's always sending true for `async_compatible` and never expects synchronous submissions for applications with an email

jholton: Got it.
But what about the new `params[:async_all]` param.  Is it sending that too?

lihanli: backend will be deployed first and frontend initially wont send `async_all`. then frontend changes will be made and they will add `async_all` when they are ready to handle async with all submissions

jholton: OK, so for a while only forms that contain an email will be processed async, then they will set the flag so forms without email can also be processed async. 👍

jholton: Since we are increasing the timeout and we will still be processing submissions synchronously (for a time), will this increase risk of tying up Rails processes if/when the HCA service is slow?
Looking at our dashboards, it doesn't appear that timeouts/long processing requests are an issue for the vast majority of HCA requests.  I think the risk is low.  
Do you have any thoughts?

lihanli: currently we're not tying up processes with the 15s timeout, i think most people are submitting the application without attachments, so the 30s timeout probably won't slow things down much. maybe i can wait until the frontend changes are done and then merge this PR first then have frontend merge their changes the next day to decrease the risk.

jholton: Yeah I think we’ll be fine. Just wanted your perspective.

jholton: Yeah I think we’ll be fine. Just wanted your perspective.

PR #3959, opened 2020-02-28
Modifying github_stats.rake to exclude weekends

thilton-oddball:
## Description of change
Modifying github_stats.rake to exclude weekends when calculating the time span for duration.


rakelib/github_stats.rake
```diff
@@ -10,6 +10,9 @@ REPOS = %w[
 namespace :github_stats do
   desc 'Hit the Github API and grab data on open PRs'
   task get_open_prs: :environment do
+    # setup calculator for determining duration excluding weekends
+    schedule = { %i[mon tue wed thu fri] => [[0 * 3600, 24 * 3600]] }
```
omgitsbillryan: Is it worth including holidays?

thilton-oddball: It does have the ability to include holidays and I did consider it but it is something that would have to be updated on a yearly basis. I figured the chance of that being overlooked in a year was probably fairly high vs. the small amount of variance that a few holidays being included in calculations might cause.

rakelib/github_stats.rake
```diff
@@ -25,21 +28,22 @@ namespace :github_stats do
...
-                           DateTime.now.to_f
-                         else
-                           DateTime.parse(first['submitted_at']).to_f
-                         end
-        # calculate the duration
-        duration = (first_reviewed - pr_created_at).round
+        first_reviewed_at = if first.nil?
+                              # PR has not yet been reviewed so use current timestamp
+                              # and set time zone offset to zero so we don't get negatives
+                              DateTime.now.new_offset('+0')
```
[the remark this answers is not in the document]
thilton-oddball: I did look at at the business_hours gem but it didn't seem to do quite what I needed it to do. As for the offset, the timestamps that are returned by the Github API have zero offset but when doing DateTime.now it will include the timezone offset so there is the chance of ending up with negative numbers. Setting the offset to zero gives you the proper timestamp format for calculating the correct duration.

PR #3982, opened 2020-03-02
Add CSRF protection for non-GET routes

jholton:
## Description of change
This is a solution for the issues found [here](https://github.com/department-of-veterans-affairs/va.gov-team-sensitive/issues/44), [here](https://github.com/department-of-veterans-affairs/va.gov-team-sensitive/issues/42) and [here](https://github.com/department-of-veterans-affairs/va.gov-team-sensitive/issues/61)

`vets-api` operates as an API for our frontend, `vets-website`
Since our frontend content is not rendered by the server, Rails' `protect_from_forgery` will not work.
Still, our frontend provides form submissions for our end users.  Many of these forms do not require authentication
CSRF proctection is still needed.  This PR adds a module that provides that protection following the same basic pattern that is used by Rails' `protect_from_forgery`
Whereas `protect_from_forgery` uses `session` and hidden form fields, we are using `cookies` and request headers.

For now, CSRF violations will be logged to Sentry.
When we are confident in the functionality we will change this to raise `ActionController::InvalidAuthenticityToken` exceptions.

"Lighthouse" endpoints can't use/don't need this protection. We have created a `ExternalApiApplicationController` (name suggestions welcome) to act as a parent controller for all endpoints that are not used by `vets-website`, including "Lighthouse" endpoints.
This controller skips the CSRF functionality.

## Testing done
Specs

## Testing planned
Review instances, staging, production.  For now, no exceptions will be intentionally raised while testing.

#### Unique to this PR

- [x] create `CSRFProtection` module for adding CSRF protection
- [x] exempt "Lighthouse" endpoints from CSRF protection

#### Applies to all PRs

- [x] Appropriate logging


app/controllers/concerns/authentication_and_sso_concerns.rb
```diff
@@ -4,16 +4,28 @@
 # to this responsibility alone.
 module AuthenticationAndSSOConcerns
   extend ActiveSupport::Concern
-  include ActionController::HttpAuthentication::Token::ControllerMethods
```
omgitsbillryan: Don't we still need `ActionController::HttpAuthentication::Token::ControllerMethods` to do token authentication?

jholton: This should probably be in a separate PR. 
This should have been removed in PR #2692. None of the methods it provides are in use anymore.

jholton: Those methods documented [here](https://api.rubyonrails.org/v5.2.3/classes/ActionController/HttpAuthentication/Basic/ControllerMethods.html)

spec/request/csrf_request_spec.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+# This file is for exercising routes that should require CSRF protection.
+# It is very much a WIP
+
+# TODO:
+# Lighthouse API endpoints
+# Check routes.rb for other `POST` routes
```
jholton: Please add to the `TODO` list as you think of cases to consider.

spec/request/csrf_request_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+
+require 'rails_helper'
+
+RSpec.describe 'CSRF scenarios', type: :request do
+  before do 
+    get(v0_maintenance_windows_path)
```
jholton: innocuous route chosen for setting the CSRF token in the cookie

omgitsbillryan:
```suggestion
    # innocuous route chosen for setting the CSRF token in the cookie
    get(v0_maintenance_windows_path)
```
😄

app/controllers/concerns/csrf_protection.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+# This module only gets mixed in to one place, but is that cleanest way to organize everything in one place related
+# to this responsibility alone.
```
omgitsbillryan: This comment might not age well.

I'd like to see a "why-style"  comment explaining how we're api-only and cannot use the normal rails `protect_from_forgery`.

app/controllers/concerns/csrf_protection.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def validate_csrf_token!
+    if request.headers['X-CSRF-Token'].nil? || request.headers['X-CSRF-Token'] != cookies['X-CSRF-Token']
+      # raise ActionController::InvalidAuthenticityToken
+
+      # for now we are just logging when there's no CSRF protection
+      # when this is going to be enforced return a meaningful error (and turn up logging level)
+      log_message_to_sentry('Request susceptible to CSRF', :info, controller: self.class, action: action_name)
+    end
```
omgitsbillryan: How does the validation work when in the event where multiple cookie values were generated? Does rails keep some kind of memory of cookie values? (This is just for my own education)

spec/request/csrf_request_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+require 'rails_helper'
+
+RSpec.describe 'CSRF scenarios', type: :request do
+  around do |example|
+    original_val = ActionController::Base.allow_forgery_protection
+    ActionController::Base.allow_forgery_protection = true
+    example.run
+    ActionController::Base.allow_forgery_protection = original_val
+  end
```
omgitsbillryan: Is this exists because `ActionController::Base.allow_forgery_protection = false` in test mode? 

1. Probably worthy of a comment.
2. Any merit to doing a before(:all) and after(:all) to set these instead? I'm not sure if setting this value incurs any performance hit.

jholton: Yes.
1. 👍 
2. Agreed, seems like a good use case for `before/all :all`

app/controllers/concerns/csrf_protection.rb
jholton: We are just comparing the token value found in the cookie with the token value found in the request header.  Rails remembers nothing.  

We are using `cookies['X-CSRF-Token']` and `request.headers['X-CSRF-Token']`

Rails' `protect_from_forgery` uses `session[:_csrf_token]` and the value in the meta tag/hidden form field.

We can't use `session` because it's encrypted and our frontend would not be able to read it.  `cookies` are unencrypted.

app/controllers/concerns/csrf_protection.rb
jholton: Updated.

jholton:
cc @bastosmichael @charleystran this PR touches Lighthouse code.

patrickvinograd:
Glad to see this getting implemented.

jholton:
closing in favor of a different approach [here](https://github.com/department-of-veterans-affairs/vets-api/pull/4056)

PR #3984, opened 2020-03-03
Claimant child lookup

bshyong:
## Description of change
When claimant is a child, attempts lookup of corresponding Person record in BGS for eventual use in BIP Claims establishment.

Also adds statsD tracking to record when claimant record can be found.

Resolves https://github.com/department-of-veterans-affairs/va.gov-team/issues/5579

## Testing done
Local

## Testing planned
Local

- [x] Appropriate logging


lib/structured_data/utilities.rb
```diff
@@ -0,0 +1,26 @@
...
+      results = find_dependents(
+        veteran.participant_id
+      )
+
+      children = results[:dependent].select do |c|
+        [
+          c[:ptcpnt_rlnshp_type_nm] == 'Child',
+          c[:first_nm].downcase == c_name['first'].downcase,
+          c[:last_nm].downcase == c_name['last'].downcase
+        ].reduce(&:&)
```
bshyong: Basic lookup logic for now, only uses name and relationship. Middle name is not available on the 530 form for claimant. Selects if all conditions are true

lib/structured_data/utilities.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module StructuredData
+  class Utilities
+    def self.find_dependents(file_number_or_ssn)
+      service = LighthouseBGS::Services.new(external_uid: 0, external_key: 'VA.gov')
```
jjulian: Do we need specific "external" parameters here?

lib/structured_data/utilities.rb
```diff
@@ -0,0 +1,26 @@
...
+      service.people.find_dependents(file_number_or_ssn).fetch(:dependent, nil)
+    end
+
+    def self.find_dependent_claimant(veteran, c_name, _c_address)
+      results = find_dependents(
```
jjulian: It may be easier to read this method if `find_dependents` is extracted here

lib/structured_data/utilities.rb
charleystran: These keys should be specific to the person/veteran who is making the call. The external_uid should be the users icn number and the external_key should be the users email address.

app/workers/structured_data/process_data_job.rb
```diff
@@ -17,18 +16,31 @@ def perform(saved_claim_id)
...
+        claimant = lookup_claimant(relationship_type, veteran, claimant_name, claimant_address) if veteran
       ensure
         @claim.process_attachments! # upload claim and attachments to Central Mail
 
         # veteran lookup for hit/miss metrics in support of Automation work
         StatsD.increment("#{stats_key}.success", tags: [
                            "relationship:#{relationship_type}",
-                           "veteranInMVI:#{veteran&.participant_id}"
+                           "veteranInMVI:#{veteran&.participant_id.present?}",
+                           "claimantFound:#{claimant.present?}"
```
omgitsbillryan: If the code in the `begin` block raises somehow, then `relationship_type`, `veteran`, & `claimant`, could all be `nil`. I think we want the code in the `ensure` block to not be dependent on anything going on inside the `begin` block.

app/workers/structured_data/process_data_job.rb
```diff
@@ -17,18 +16,31 @@ def perform(saved_claim_id)
...
+        claimant_address = @claim.parsed_form['claimantAddress']
+
         veteran = BipClaims::Service.new.lookup_veteran_from_mvi(@claim)
+
+        claimant = lookup_claimant(relationship_type, veteran, claimant_name, claimant_address) if veteran
       ensure
         @claim.process_attachments! # upload claim and attachments to Central Mail
 
         # veteran lookup for hit/miss metrics in support of Automation work
         StatsD.increment("#{stats_key}.success", tags: [
```
omgitsbillryan: Outside the scope of this PR - but are you aware of https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/common/client/concerns/monitoring.rb - it may be able to replace this ensure block.

app/workers/structured_data/process_data_job.rb
```diff
@@ -8,27 +8,40 @@ class ProcessDataJob
...
+
+    def lookup_claimant(veteran, claim)
+      relationship_type = claim.parsed_form['relationship']&.fetch('type', nil)
+      claimant_name = claim.parsed_form['claimantFullName']
+      claimant_address = claim.parsed_form['claimantAddress']
+
+      case relationship_type
+      when 'child'
+        StructuredData::Utilities.find_dependent_claimant(veteran, claimant_name, claimant_address)
+      end
```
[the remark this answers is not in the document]
omgitsbillryan: Woops, code changed in the middle of my review. It would return `nil`.

Case statement with 1 case is still weird to me though.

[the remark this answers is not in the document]
bshyong: We need to add cases for other types of claimants later on but good point,  we can leave this in a simple `if` and use cases when they are needed

app/workers/structured_data/process_data_job.rb
bshyong: I'm using `present?` here so that our stats trigger with a false/not found if those values are `nil`—does that work?

lib/structured_data/utilities.rb
bshyong: @jjulian @charleystran this form can be filled out by anonymous users so we may not have that info. what should we fill out in that case?   The BGS calls will fail if nil values are provided for these uids

bshyong: One thing we could do is pass the claimant email and phone number if we really want to keep track of that these calls relate to. I have a message out to the BGS team to ask them how this information is being used on their end

charleystran: Its a BGS requirement that we pass enough information in these values to identify the end user making the call. Its all a part of their logging and audit trail. They want to be able to track down any user who made the call. Something outside our control. Maybe email and phone number would be enough? We would need to check with BGS on that.

bshyong: @charleystran @johnpaulashenfelter I followed up with Cory Easley from BGS on this. I can confirm that it is used for logging/audit trail, no other use. For this burials form, an anonymous user will be able to fill out the form. I suggest we use the Claimant email and phone number for this purpose as they are the ones who are submitting the form to VA.gov

charleystran: Did Corey say he was ok with using those values? I know when we setup to use icn and email address, there were a couple people like John Dell that had to give it the all clear. If they are ok with sending those values in those headers, then I am ok with it.

bshyong: Will ask John Dell in BGS email thread to confirm and update the PR accordingly

jjulian: Is the UID numeric-only? I'm considering collisions between a 10-digit claimant phone number and an ICN number.
If UID is a string, we can make phone numbers unique with a prepended string, like `+` or `tel:`

kfrz:
Gentle nudge to @charleystran and @bshyong  - any update on the aforementioned parameter issue?

app/workers/structured_data/process_data_job.rb
kfrz: :heavy_plus_sign: :point_up:

lib/structured_data/utilities.rb
kfrz: Following up @bshyong did you get a reply from the BGS email thread/John Dell?

lib/structured_data/utilities.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module StructuredData
+  class Utilities
+    def self.find_dependents(file_number_or_ssn)
```
kfrz, line 5: Please document this class and class-level method using YARDoc.

kfrz (changes requested):
Overall this looks fine. Wondering if there's any updates, this PR has been opened for quite some time.

bshyong:
@kfrz I am no longer working at Ad Hoc and I’m not sure where this structured data work landed. But happy to provide context whoever is taking it over if the project is still active. @jjulian or @alistair may have more context

jjulian:
Thanks @bshyong for commenting! This work has been indefinitely put on hold. What is the protocol? Close PR but leave branch? Are there any labels we can use to communicate this status?

patrickvinograd:
This work is long since mothballed/obsolete, and Ben is no longer working at VA, closing this out.

PR #3988, opened 2020-03-04
adding a forceAuthn value to the SAML Settings

ericbuckley:
## Description of change

Adding a `forceAuthn` value to the SAML payload sent to eauth for the v1 sessions controller.  By default this value will be set to "false", however if the request has explicitly set the `?force=true` query parameter, then it will be set to true.

When a user has authenticated with eauth (IAM), before coming to va.gov, there will be cases when that user doesn't have the necessary ID.ME values in their profile.  In order to get those, we need to force eauth to reauthenticate with ID.ME to have those values set.

https://github.com/department-of-veterans-affairs/va.gov-team/issues/5957

#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging

### DIFF
--- a/app/controllers/v1/sessions_controller.rb
+++ b/app/controllers/v1/sessions_controller.rb
@@ -81,6 +81,9 @@ module V1
     private
 
     def saml_settings(options = {})
+      # add a forceAuthn value to the saml settings based on the initial options or
+      # the "force" value in the query params
+      options[:force_authn] = options[:force_authn] || params[:force]&.downcase == 'true'
       SAML::SSOeSettingsService.saml_settings(options)
     end
 


ericbuckley:
@patrickvinograd i tried a few different ways to write test cases this afternoon to test this, however because the [saml settings are getting mocked](https://github.com/department-of-veterans-affairs/vets-api/blob/master/spec/controllers/v1/sessions_controller_spec.rb#L28), there doesn't seem to be a good way to verify that the "force_authn" value is being set in the SAMLRequest parameter.  I did verify that it works when manually testing this though.  Any ideas on how to implement a test case?

ericbuckley:
> @patrickvinograd i tried a few different ways to write test cases this afternoon to test this, however because the [saml settings are getting mocked](https://github.com/department-of-veterans-affairs/vets-api/blob/master/spec/controllers/v1/sessions_controller_spec.rb#L28), there doesn't seem to be a good way to verify that the "force_authn" value is being set in the SAMLRequest parameter. I did verify that it works when manually testing this though. Any ideas on how to implement a test case?

I think I found a workaround using the rspec `.with` function, https://github.com/department-of-veterans-affairs/vets-api/pull/3988/commits/f97179e8daed911f9fdcd72e8e6eefe1ff52525d

app/controllers/v1/sessions_controller.rb
```diff
@@ -24,7 +24,8 @@ def new
       type = params[:type]
       raise Common::Exceptions::RoutingError, params[:path] unless REDIRECT_URLS.include?(type)
 
-      StatsD.increment(STATSD_SSO_NEW_KEY, tags: ["context:#{type}"])
+      StatsD.increment(STATSD_SSO_NEW_KEY,
+                       tags: ["context:#{type}", "forceauthn:#{force_authn?}"])
```
johnpaulashenfelter: @ericbuckley is there a dashboard somewhere that matches this new tag?

johnpaulashenfelter: Or maybe an alert?

johnpaulashenfelter: Chatted about it and likely to be done

PR #3989, opened 2020-03-04
Allows in progress forms to be found by account id

patrickvinograd:
Mitigates dependency on ID.me UUIDs by letting in progress forms
be resolved by either idme_uuid or account_id. Preferentially adds
new forms using account_id. Updates existing IPF lookups to be
tolerant of either lookup style.

#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging


app/models/in_progress_form.rb
```diff
@@ -19,8 +20,25 @@ def cast(value)
   before_save :serialize_form_data
   before_save :set_expires_at
 
+  scope :for_user, lambda { |user|
```
kreek: 👍nice way to scope it to both ids

app/models/in_progress_form.rb
```diff
@@ -19,8 +20,25 @@ def cast(value)
...
   def self.form_for_user(form_id, user)
-    InProgressForm.find_by(form_id: form_id, user_uuid: user.uuid)
+    InProgressForm.for_user(user).find_by(form_id: form_id)
+  end
+
+  # Resolve a form given an "indifferent" user_id, i.e.
+  # one that may be either an idme_uuid or a prefixed account_id
+  def self.form_for_user_id(form_id, user_id)
+    acct = Account.find_by(idme_uuid: user_id)
+    acct = Account.find_by(uuid: user_id.delete_prefix(ACCT_ID_PREFIX)) if user_id.start_with?(ACCT_ID_PREFIX)
```
kreek:
```suggestion
    args = user_id.start_with?(ACCT_ID_PREFIX) ? { uuid: user_id.delete_prefix(ACCT_ID_PREFIX) } : { idme_uuid: user_id }
    acct = Account.find_by(args)
```

kreek:
One more question is the not null constraint on user_uuid the reason for favoring using the `acct:` prefix over adding a new related `account_id` column. Maybe we could drop not null on `user_uuid` and once there are no legacy forms remove the column? It would mean allowing `account_id` to be null temporarily at the db level until that happened but we could enforce it for new forms in rails.

app/models/in_progress_form.rb
ericbuckley: @patrickvinograd would there ever be a case where `user.uuid` is NULL?

app/models/in_progress_form.rb
patrickvinograd: Yup that's better, Rubocop had a multi-line ternary complaint that i was trying to get rid of and my mind went in a different direction

app/models/in_progress_form.rb
patrickvinograd: @ericbuckley Yes, I think we're headed for a case where we get an inbound SSOe session that will not have an idme_uuid. Do you think the query needs updating for that case?

app/controllers/v0/in_progress_forms_controller.rb
```diff
@@ -5,12 +5,13 @@ class InProgressFormsController < ApplicationController
...
     def index
-      render json: InProgressForm.where(user_uuid: @current_user.uuid)
+      render json: InProgressForm.for_user(@current_user)
+      # render json: InProgressForm.where(user_uuid: @current_user.uuid)
     end
 
     def show
       form_id = params[:id]
-      form    = InProgressForm.form_for_user(form_id, @current_user)
+      form    = InProgressForm.for_user(@current_user).where(form_id: form_id).first
```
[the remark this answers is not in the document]
patrickvinograd: No I just went through iterating on the form_for_user method and missed this case the second time around. Will update to match the others.

patrickvinograd:
I went back and forth on whether to add a new column or not. If we add a column, then are we going to record both the idme_uuid (when it exists) and the account_id? And then what do we do if there's somehow a mismatch between the two? Which seems unlikely but possible. 

Maybe it's six of one half a dozen of the other, but this way there is one field that can be used to match an in_progress_form to a person and that seemed semantically cleaner to me even at the cost of some lookup complexity.

app/models/in_progress_form.rb
ericbuckley: I think so, if we get a user with a NULL uuid, this query will return all users that also have a NULL uuid plus the one use with the matching account_id.

patrickvinograd: But there's not a separate uuid column on the InProgressForm model - we're overloading the user_uuid column to always either be an idme_uuid or an account_id (prefixed with 'acct:'). The column still has a non-null constraint. 
In other words I think the difference is that you needed that logic in the Account model because you added a new secid column and then had to look up across two columns either of which might be null, but that's not the case here. 
But I should add a test case for nil user.uuid.

patrickvinograd: And maybe a guard that account_id and user.uuid are not both nil.

ericbuckley: that makes sense, if `user_uuid` is never NULL then ignore my last comment

patrickvinograd: I can't easily add such a spec for nil uuid because it's validated to be non-nil, but I'll add some more careful handling of nil account_ids

app/controllers/v0/in_progress_forms_controller.rb
```diff
@@ -20,13 +20,16 @@ def show
     end
 
     def update
-      form = InProgressForm.where(form_id: params[:id], user_uuid: @current_user.uuid).first_or_initialize
+      form = InProgressForm.where(form_id: params[:id], user_uuid: @current_user.uuid).first
+      alt_id = "#{InProgressForm::ACCT_ID_PREFIX}#{@current_user.account_id}"
```
annaswims: How about making a method for `alt_id` in `User`?  Looks like there are quite a few places where these two strings are cobined.

app/workers/evss/disability_compensation_form/submit_form526_cleanup.rb
```diff
@@ -16,7 +16,8 @@ class SubmitForm526Cleanup < Job
       def perform(submission_id)
         super(submission_id)
         with_tracking('Form526 Cleanup', submission.saved_claim_id, submission.id) do
-          InProgressForm.find_by(form_id: FORM_ID, user_uuid: submission.user_uuid)&.destroy
+          InProgressForm.form_for_user_id(FORM_ID, submission.user_uuid)&.destroy
+          # InProgressForm.find_by(form_id: FORM_ID, user_uuid: submission.user_uuid)&.destroy
```
annaswims: extra comment

lib/evss/disability_compensation_form/data_translation.rb
```diff
@@ -285,8 +285,9 @@ def greater_itf_date?
 
       def application_create_date
         # Application create date is the date the user began their application
-        @acd ||= InProgressForm.where(form_id: VA526ez::FORM_ID, user_uuid: @user.uuid)
-                               .first.created_at
+        # TODO is this IPF guaranteed to exist at this point?
+        # or should we do safe navigation?
```
annaswims: Todo?

app/models/user.rb
```diff
@@ -46,6 +46,10 @@ def account
     Account.cache_or_create_by!(self)
   end
 
+  def account_id
+    account.uuid
+  end
+
```
annaswims: User#in_progress_forms needs to be updated, right?

patrickvinograd:
closing - we can defer some of the account ID switchover work based on alternate approach in  va.gov-team#7110.

PR #3990, opened 2020-03-04
add logging around /mhv-api/patient/v1/session endpoint

lihanli:
## Description of change
https://github.com/department-of-veterans-affairs/va.gov-team/issues/6391

## Testing done
automated tests

#### Applies to all PRs

- [x] Appropriate logging


lib/sentry_logging.rb
```diff
@@ -63,6 +63,13 @@ def non_nil_hash?(h)
     h.is_a?(Hash) && !h.empty?
   end
 
+  def set_raven_tag(k, v)
+    Raven.tags_context(k => v)
+    return_val = yield
+    Raven.context.tags.delete(k)
+    return_val
+  end
```
johnpaulashenfelter: This method is only used in your client, so I'm wondering if it makes more sense to just put in your client instead of pulling it into common code.

It expects a block, and the setting and then deleting the tag key is something that on quick inspection won't make sense without a lot of context for the _why_ -- myself included :)

So no qualms about letting you do what you need to do, but I think it makes more sense closer to the code that needs this method instead of the sentry library if at all possible.

lib/common/client/concerns/mhv_session_based_client.rb
```diff
@@ -45,7 +46,9 @@ def authenticate
       # @return [Sm::ClientSession] if a SM (Secure Messaging) client session
       #
       def get_session
-        env = perform(:get, 'session', nil, auth_headers)
+        env = set_raven_tag(:error, 'mhv_session') do
```
johnpaulashenfelter: Can you find of a method name that captures what's really going on here? It's not setting the tag in the way a casual reading would indicate (esp since the corresponding method literally deletes the tag.

lihanli:
@johnpaulashenfelter ok thanks for the suggestion, i moved the method out of `SentryLogging` to a local method

lib/sentry_logging.rb
lihanli: i set the tag and then delete it because i only want the errors inside the block to be tagged. any errors that happen afterwards shouldn't be tagged.

PR #3993, opened 2020-03-04
3688 vaos debug middleware log line

kreek:
## Description of change
Fixes a bug where the vaos logging middleware was correctly writing success lines but not failures.

## Testing done
unit tests, local integration tests

## Testing planned
staging integration tests

Failure log lines written by the middleware show up in CloudWatch.

#### Applies to all PRs

- [ ] Appropriate logging


modules/vaos/app/services/vaos/configuration.rb
```diff
@@ -34,6 +34,7 @@ def connection
         conn.response :snakecase
         conn.response :json, content_type: /\bjson$/
         conn.response :vaos_errors
+        conn.use :vaos_logging
```
saneshark: interesting...

saneshark: i really thought that the .use middleware takes precedence over response middlewares... this is kind of bizarre to me... how does breakers not run into this problem when it is always at the top?

kreek: From the Faraday docs: 

> the first middleware on the list wraps all others, while the last middleware is the innermost one

`use` is for middleware that has both a request and response component. That you configure middleware in a list makes it confusing because it's more like wrapping methods. Just because a middleware only has a response doesn't mean it's still not wrapped by a `use` one.

```
request  ------------------------------->
breakers( vaos_errors( vaos_logging() ) )
<------------------------------- response
```

Breakers works because it's catching an re-raising the error. That's an option for this fix too although since this logging is specifically for the VAMF requests it makes sense to have it as close to the http call as possible.

saneshark: That's insightful and a very good TIL. Thanks for clarifying.

kreek:
@saneshark I agree about having both middleware unit test and the integration one in this PR, I'll update.

kreek:
@johnpaulashenfelter thanks for the review, @saneshark 's suggested change of having both a unit and integration test is complete so this is good to go

PR #3995, opened 2020-03-05
Settings.yml config for disabling local saml authn signing

omgitsbillryan:
## Description of change
- Adds new settings.yml boolean config: `Settings.saml. authn_requests_signed`
- Use ☝️ config to determine whether or not we sign SAML AuthNRequests in `settings_service.rb`
- Adds 2 specs:
  - Assert that we default to `true` for signing AuthNRequests in the event that `Settings.saml. authn_requests_signed` is not defined
  - Assert that we use the `Settings.saml. authn_requests_signed` setting when it _is_ defined

### Why?
See ticket: https://github.com/department-of-veterans-affairs/va.gov-team-sensitive/issues/65

## Testing done
- Lots of localhost authentication testing
- specs

## Testing planned
Test in dev & staging

#### Applies to all PRs

- [ ] Appropriate logging


config/settings.yml
```diff
@@ -12,6 +12,7 @@ saml:
   issuer: saml-rp.vetsgov.localhost
   callback_url: http://localhost:3000/auth/saml/callback
   metadata_url: https://api.idmelabs.com/saml/metadata/provider
+  authn_requests_signed: false
```
omgitsbillryan: Even though I've created the devops PR to override this to `true`, ( https://github.com/department-of-veterans-affairs/devops/pull/6488 ) I'm tempted to have it `true` here as well and force all local devs to override it to `false` in their respective `settings.local.yml` files, thoughts?

omgitsbillryan: It'd be more annoying for local devs, but also seems more appropriate since this setting would only ever be disabled for localhost authentication.

johnpaulashenfelter: Let's err on best for developers!

PR #4003, opened 2020-03-08
add new profile_schema_forms feature flag

erikphansen:
#### Unique to this PR

#### Applies to all PRs

- [ ] Appropriate logging

### DIFF
--- a/config/features.yml
+++ b/config/features.yml
@@ -24,10 +24,10 @@ features:
     actor_type: user
     description: >
       On https://www.va.gov/find-locations/ toggles the deduplication of of services from a community care provider
-  profile_show_direct_deposit:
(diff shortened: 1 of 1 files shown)


config/features.yml
```diff
@@ -24,10 +24,10 @@ features:
     actor_type: user
     description: >
       On https://www.va.gov/find-locations/ toggles the deduplication of of services from a community care provider
-  profile_show_direct_deposit:
+  profile_schema_forms:
```
erikphansen: We no longer use the `profile_show_direct_deposit` flag at all. And in fact because it was deleted from the https://api.va.gov/flipper/features dashboard it doesn't even show up there anymore even though it was still in this `features.yml` file.

PR #4005, opened 2020-03-09
invalid Settings.hostname on staging instances causing Bad Requests

saneshark:
## Description of change
VAMF appears to be doing some kind of DNS resolution on the HTTP Referrer we pass. I would like to make the change in Settings.hostname changing 'staging-api.vets.gov' to be the correct 'staging-api.va.gov' but I'm afraid this change might break something with how ID.me is configured so for now I'm just going to do a gsub.

## Testing done
verified on staging instance that making this change does not result in a Bad Request.

## Testing planned
None

None

#### Applies to all PRs

- [x] Appropriate logging

### DIFF
--- a/modules/vaos/app/services/vaos/base_service.rb
+++ b/modules/vaos/app/services/vaos/base_service.rb
@@ -27,7 +27,7 @@ module VAOS
     # Set the referrer (Referer header) to distinguish review instance, staging, etc from logs
     def referrer
       if Settings.hostname.ends_with?('.gov')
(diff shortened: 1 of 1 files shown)


saneshark:
Yup, I'm going to look at making the changes to Settings.hostname longer term, but I'm worried it might break sign in since the ID.me metadata also depends on that hostname and this might cause problems to them. It should probably be fine, but this is a quick hack to work around the issue for now and unblock further testing before our roll-out Wednesday.

Thanks for the quick turn-around!

PR #4008, opened 2020-03-10
VAOS change breakers error_threshold to 90 instead of 50

saneshark:
## Description of change
Because VAOS makes calls out to a large number of different endpoints and they are all characterized as the same VAOS service name and all URI begin with the same hostname, there is greater risk that 1 or 2 of these endpoints might cause an outage across the entirety of the app.

As we continue to learn more we might look at coming up with custom matchers based on client class names that we use instead of URI. But for now, we just want to avoid this problem and we can do that by setting the error_threshold so high that it's unlikely to ever trigger a breakers outage without affecting all of the plugins such as STATSD and others which we depend on.

## Testing done
None

## Testing planned
None

#### Unique to this PR
I believe this is the the first time someone has changed the error_threshold for a client configuration breakers service, there should be a better way to do it by setting a method on Configuration::Base I would think.

#### Applies to all PRs

- [x] Appropriate logging

### DIFF
--- a/lib/common/client/configuration/base.rb
+++ b/lib/common/client/configuration/base.rb
@@ -120,9 +120,14 @@ module Common
           )
         end
 
+
+        # The percentage of errors over which an outage will be reported as part of breakers gem
+        #
+        # @return [Integer] corresponding to percentage
         def breakers_error_threshold
           50
         end
+
 
         private
 
--- a/lib/common/client/configuration/base.rb
+++ b/lib/common/client/configuration/base.rb
@@ -108,11 +108,20 @@ module Common
             end
           end
 
-          @service = Breakers::Service.new(
+          @service = create_new_breakers_service(matcher, exception_handler)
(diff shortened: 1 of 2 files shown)


lib/common/client/configuration/base.rb
```diff
@@ -108,13 +108,22 @@ def breakers_service
...
+        def create_new_breakers_service(matcher, exception_handler)
+          Breakers::Service.new(
             name: service_name,
             request_matcher: matcher,
+            error_threshold: breakers_error_threshold,
             exception_handler: exception_handler
           )
         end
 
+        def breakers_error_threshold
```
annaswims: Can we add a comment making clear that this is `The percentage of errors over which an outage will be reported.`, just to spare someone the need to read breakers documentation?

saneshark: done

PR #4010, opened 2020-03-10
Get In Progress Form endpoint calls MVI proxy add

SarahJaine:
## Description of change

**Closes [va.gov-team 4855](https://github.com/department-of-veterans-affairs/va.gov-team/issues/4855)**
In the simplest terms, calling `GET` on the in progress form endpoint will now start an "add person" call to MVI.  This allows first time applicants to use the online form.  

There is some debate about where this logic should reside-- putting it in the controller for in progress forms isn't elegant, but our other option (adding a new controller just for this use case) feels heavy. See more details on this debate in the comments below.

**_Dearest Reviewers,  I'm new to ruby/rails so please feel free to correct any stylistic conventions or add in ruby/rails-specific advice you think I could benefit from._**

## Testing done

- [x] Added unit tests

#### Unique to this PR

- [x] new proxy_add method
- [x] call new proxy_add method via before_action only on 'show' call
- [x] redis_store add :clear_cache method, use instead of :destroy

#### Applies to all PRs

- [x] Appropriate logging - already in place
- [x] Swagger docs have been updated


app/swagger/requests/in_progress_forms.rb
```diff
@@ -55,7 +55,10 @@ class InProgressForms
         operation :get do
           extend Swagger::Responses::AuthenticationError
 
-          key :description, 'Get form data'
+          key :summary,     'Get form data'
+          key :description, 'When a GET request is made for a user completing disability claim form 21-526EZ
+            for the first time, the API service will make a call to add the user to the Master Veteran Index (MVI)
+            and a new blank form will be served to the client.'
```
SMLuthi: This is close but not quite correct. The proxy add is asking MVI to make a request to birls and/or corp database and asking them create identities there. This will populate corp/birls ids in the MVI which are both required to make a successful 526 submission through EVSS. The veteran *should* already exist in MVI. Once this action has been done, the in progress controller can move forward and serve up the form and attempt to prefill the data as it normally does.

lib/common/models/concerns/cache_aside.rb
```diff
@@ -26,20 +26,29 @@ def self.redis_config_key(key)
       attribute :response
     end
 
+    # get method
+    def cached?(key:)
```
SMLuthi: Good idea! Makes testing this much easier.

spec/request/in_progress_forms_request_spec.rb
```diff
@@ -241,6 +241,52 @@
...
+        context 'and is completing form 526' do
+          let(:user) { build(:user_with_no_ids) }
+          let!(:in_progress_form) { FactoryBot.create(:in_progress_form, user_uuid: user.uuid, form_id: '21-526EZ') }
+
+          it 'call is made to add user to MVI' do
+            VCR.use_cassette('mvi/add_person/add_person_success') do
+              VCR.use_cassette('mvi/find_candidate/orch_search_with_attributes') do
+                expect(in_progress_form.form_id).to eq('21-526EZ') # check test set up
+                # expect success to be achieved by calling mvi_add_person
+                expect_any_instance_of(Mvi).to receive(:mvi_add_person).once
```
SMLuthi: Nothing to add about the work here except I want people to know how annoying this test was to get working :)

app/controllers/v0/in_progress_forms_controller.rb
```diff
@@ -32,5 +33,16 @@ def destroy
...
       render json: form
     end
+
+    def proxy_add
+      if params[:id] == '21-526EZ'
+        if @current_user.participant_id.nil?
+          # Add person to MVI if missing participant_id (no matter if birls_id present or absent)
+          @current_user.mvi.mvi_add_person
+        elsif @current_user.birls_id.nil?
+          raise Common::Exceptions::ValidationErrors, 'No birls_id while participant_id present'
```
SMLuthi: It may be worth our time adding in some logging here. This is an extreme edge case (as it *technically* should never be possible to have a corp id but not a birls id) and we may want to log some user information so we can investigate these cases with upstream services.

patrickvinograd:
I get why this is being done but I have to say this is kind of an un-intuitive point to hook in the AddPerson call, since the InProgressForm controller is general to all kinds of forms. I realize it's done selectively for that 526ez form but can you add some explanation here in the PR or in the commit message about why this was the best place to hook in this functionality as opposed to doing it somewhere that is specific to the 526 application?

SMLuthi:
Hey @patrickvinograd good question. Its actually something I bounced around for awhile before settling on this. Theres a couple factors that go into it. First, the in progress call is the very first call the form flow makes when starting the 526 flow. Directly following that call is the ITF hook (which retrieves and/or creates an ITF). On top of that, the prefill work is dependent on the add person call being made (if the ids dont exist for said user) since it relies on EVSS and EVSS requires both ids in the auth header. I therefore made the call to hook it into in progress controller. The other idea was to create an entirely separate controller to handle the work but MVI has made it very clear that this call should only be made in regards to 526. I didnt want to expose an entire endpoint that could just be accessible to anyone... but hooking it into and forcing it on the form `21-526ez` identifier seemed to be more in line with their vision.

app/controllers/v0/in_progress_forms_controller.rb
SarahJaine: I agree that we should log this error. My reading of this controller is that it inherits from ApplicationController and the application controller will log the error to sentry since `Common::Exceptions::ValidationErrors` isn't in the `SKIP_SENTRY_EXCEPTION_TYPES` array. Am I reading this? ref:https://github.com/department-of-veterans-affairs/vets-api/blob/4e72f8bf232efbd02ec9d2945b261ca35ecb69cf/app/controllers/application_controller.rb#L109

app/swagger/requests/in_progress_forms.rb
SarahJaine: Thanks for catching this. I wrote a new description. Could you do another pass on it for me please?

app/controllers/v0/in_progress_forms_controller.rb
SMLuthi: I think you are correct and looking deeper into that it looks like `log_exception_to_sentry` will also log it to the rails console, so i think we are good there.

app/swagger/requests/in_progress_forms.rb
SMLuthi: Looks solid!

app/controllers/v0/in_progress_forms_controller.rb
```diff
@@ -32,5 +33,16 @@ def destroy
       form.destroy
       render json: form
     end
+
+    def proxy_add
+      if params[:id] == '21-526EZ'
```
annaswims: Could we use the constant `VA526ez::FORM_ID` rather than the string `21-526EZ`?

app/controllers/v0/in_progress_forms_controller.rb
```diff
@@ -32,5 +33,16 @@ def destroy
...
+        if @current_user.participant_id.nil?
```
annaswims: Is this meant to be the check for if the user is in MVI?  Do we know for sure that the user isn't in MVI with a blank `participant_id`? 

We're only querying MVI for LOA3 users so do we know at this point that the user is LOA3?   We don't want to add non-identity verified users to MVI, do we?

spec/request/in_progress_forms_request_spec.rb
annaswims: @SMLuthi were there any hard lessons learned worth sharing?

app/controllers/v0/in_progress_forms_controller.rb
SMLuthi: Good question. The form 526 flow is gated to log in on the front end AND the service list for their user object. The authorization verifies that the necessary information for the user exists which, I believe, includes them being `user.loa3?`.

But, even if they managed to attempt an add proxy call to MVI, if they dont exist in MVI, it would fail. This call isnt creating an MVI entry but rather, creating entries in CORP and/or BIRLS depending on whether the veteran exists in MVI already. The add proxy call _requires_ an orchestrated search to be made to MVI *before* an add proxy call is made. This accomplishes two things... 1) It creates a timestamped entry in mvi which is important because the proxy add call is time limited to ~90 seconds to the search call that preceded it and 2) the orchestrated search call retrieves a `search_token` var that needs to be included in the proxy call that is made.

SMLuthi: Long story short, it doesnt check for mvi existence, it just checks if the veteran has an entry in the CORP database

spec/request/in_progress_forms_request_spec.rb
SMLuthi: Nothing that im not embarrassed about. It mostly came down to making `expect` calls AFTER making the `GET` request... essentially it was checking whether `:mvi_add_person` was being called after we had already made the request. So it was failing haha

app/controllers/v0/in_progress_forms_controller.rb
SMLuthi: Yes, thats a good idea.

app/controllers/v0/in_progress_forms_controller.rb
SMLuthi: Also, the mvi profile isnt accessible at all UNLESS the user is loa3 certified to begin with.

SMLuthi: Oh, I should add, the auth gating for the service checks our form526 policy. This policy verifies the existence of an `edipi`. The user object, when accessing `edipi` checks whether the user is `loa3?`

app/controllers/v0/in_progress_forms_controller.rb
SarahJaine: Absolutely-- I didn't realize it was set as a constant somewhere. Thanks!

app/controllers/v0/in_progress_forms_controller.rb
SarahJaine: @annaswims RE: "Do we know for sure that the user isn't in MVI with a blank participant_id?"  

Great question.  When I was writing that part I looked for evidence that the participant and birls ids were `nil` when "empty/blank/none" and I found the sanitize methods for the mvi profile construction.  https://github.com/department-of-veterans-affairs/vets-api/blob/d238959fcd3e040698e35faf41410b1992b9357e/lib/mvi/responses/profile_parser.rb#L164
My reading is that 1. we expect empty/blank/none values to be represented as `nil` in the data source and 2. if the data source value contains only non-digit characters we transform that to `nil` (the return value of an unsuccessful `match`.) https://ruby-doc.org/core-2.5.1/Regexp.html

PR #4013, opened 2020-03-10
multiple Account matches should select value with matching idme uuid

ericbuckley:
## Description of change
- simplifying Account lookup query with 2 named scopes
- when more than 1 account is found (a user could have multiple credential sets for logging in), attempt to select the one with a matching idme uuid value

closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/6702

## Testing done
- wrote a new test case to verify the correct Account is matched

#### Applies to all PRs

- [ ] Appropriate logging


patrickvinograd:
LGTM

ericbuckley:
@patrickvinograd you worked with one of the engineers at ID.me to identify credential sets that mapped to the same sec id.  Do you know if any of these had mismatched LOA values?

patrickvinograd:
We had a face-face discussion about the issue Anna raised above. Noting here for posterity:
- A user can have both LOA1 and LOA3 accounts using separate credentials. 
- But, a user will only have an ICN and SECID if they are LOA3 - because both SSOe and VA.gov will only resolve a user in MVI if they are LOA3. If a user logs in via SSOe at LOA1, they arrive at VA.gov as a "csp-only" user with a much-reduced trait set, including no SECID.

So, there should be no way that we wind. up resolving  a user's LOA3 account ID using their LOA1 credential:
- An LOA1 login would only look up account IDs by idme_uuid, because no secid is present on the user object. So it would only find the account object with the same idme_uuid, and would never be able to add a secid to that account object.
- An LOA3 login would look up by idme_uuid or secid, but would not resolve a user's other LOA1 account object because the latter's idme_uuid would be different, and it would not have a secid stored.

annaswims:
one thing to note: we do get ICN from the MHV saml response as well.

app/models/account.rb
```diff
@@ -56,16 +59,16 @@ def self.cache_or_create_by!(user)
...
 
     if accts.length > 1
       data = accts.map(&:attributes)
       log_message_to_sentry('multiple Account records with matching ids', 'warning', data)
+      # sort the accounts so the ones with matching idme_uuid values come first
+      # in the array, this will provide the users with a more consistent
+      # experience in case they have multiple credentials to login with
+      # https://github.com/department-of-veterans-affairs/va.gov-team/issues/6702
+      accts = accts.sort_by { |a| a.idme_uuid == attrs[:idme_uuid] ? 0 : 1 }
     end
```
johnpaulashenfelter: @ericbuckley since you've already got the explanation here (which is great -- its not obvious whats going on w/o some context), maybe you could turn this into a (private?) method that describes the function -- 
```
accts = sort_with_idme_uuid_priority(accts)

# https://github.com/department-of-veterans-affairs/va.gov-team/issues/6702
def  sort_with_idme_uuid_priority(accts)
   if accts.length > 1
      data = accts.map(&:attributes)
      log_message_to_sentry('multiple Account records with matching ids', 'warning', data)
      accts = accts.sort_by { |a| a.idme_uuid == attrs[:idme_uuid] ? 0 : 1 }
end
``` 

or something similar. Would fix the complexity issue in codeclimate and make it easier to read

app/models/account.rb
```diff
@@ -108,7 +101,22 @@ def self.get_key(user)
...
+  # Sort the given list of Accounts so the ones with matching ID.me UUID values
+  # come first in the array, this will provide users with a more consistent
+  # experience in the case they have multiple credentials to login with
+  # https://github.com/department-of-veterans-affairs/va.gov-team/issues/6702
+  #
+  # @return [Array]
+  def self.sort_with_idme_uuid_priority(accts)
+    if accts.length > 1
+      data = accts.map(&:attributes)
+      log_message_to_sentry('multiple Account records with matching ids', 'warning', data)
```
annaswims: `data` has PII that we don't want logged to sentry.

ericbuckley: thanks @annaswims I just made this change to only log the Account ID.  https://github.com/department-of-veterans-affairs/vets-api/pull/4013/commits/43d912c04f27a1643bd66d1242d5a756bef9374b

PR #4014, opened 2020-03-11
Add deployments method to MilitaryInformationServiceV2

rtravitz:
## Description of change

This PR supports [IN-21](https://vasdvp.atlassian.net/browse/IN-21?atlOrigin=eyJpIjoiYmFiYjJlOTdlYzdiNGVlN2JhMTVkNTI5NTJkZTM0OWEiLCJwIjoiaiJ9). It adds the ability to retrieve deployments from v2 of the eMIS militaryInformationService. Those deployments will be used in a later PR to add additional fields to the Lighthouse `/service_history` endpoint. It is followed by #4015 and #4016.

Notably this PR:
- Adds the `get_deployment` method to `MilitaryInformationServiceV2`
- Creates `DeploymentV2` and `DeploymentLocationV2` models to hold the parsed response from the backing service
- Adds recordings of v2 responses taken from a staging eMIS server and restructures some previous `military_service_episodes` recordings
- Adds the `military_service_episodes_response_v2` and `deployment_response_v2` to the requires in the main response class. Not having those there causes a bug in the more complete implementation where the `Oj` library can't load the class if it hasn't been referenced yet, such as in a server restart. 

## Testing done

- Added an additional unit test based on the new recordings for deployment

#### Unique to this PR

- [ ] Add `pay_grade_code` to `/service_history`
- [ ] Add type of separation - Retired, discharged, etc. (reason provided if discharged, but nothing indicating they were retired) to `/service_history`.

#### Applies to all PRs
- [x] Appropriate logging


lib/emis/responses.rb
```diff
@@ -15,6 +15,8 @@
 require_dependency 'emis/responses/get_separation_pay_response'
 require_dependency 'emis/responses/get_unit_information_response'
 require_dependency 'emis/responses/get_veteran_status_response'
+require_dependency 'emis/responses/get_deployment_response_v2'
+require_dependency 'emis/responses/get_military_service_episodes_response_v2'
```
nfasulo: Good job catching out that these needed to be added!

lib/emis/military_information_service_v2.rb
```diff
@@ -10,6 +10,7 @@ class MilitaryInformationServiceV2 < Service
 
     create_endpoints(
       [
+        [:get_deployment, 'deploymentRequest', 'V2'],
```
nfasulo: Has adding a second endpoint to v2 brought up any concerns with the way we are currently handling the versioning of eMIS dependencies?

rtravitz: Can you say more on this question? Are you concerned that we're duplicating too much, or something else?

nfasulo: No, I don't have any particular concern, I'm just wondering if you ran into any pain points in adding to v2 or saw anything that might be troubling as more endpoints get added.

rtravitz: It's pretty straightforward. You add an extra response and a model or two for each new method. That only feels a bit bad in this case because of the almost exact duplication between the V1 and V2 responses and models. @dbrown-va and I felt that the duplication is a fair trade-off to make to keep the different major versions of the API logically separate.

PR #4015, opened 2020-03-11
Add caching model for v2 of eMIS MilitaryInformationService

rtravitz:
## Description of change

This PR supports [IN-21](https://vasdvp.atlassian.net/browse/IN-21?atlOrigin=eyJpIjoiYmFiYjJlOTdlYzdiNGVlN2JhMTVkNTI5NTJkZTM0OWEiLCJwIjoiaiJ9). It caches responses from v2 of the eMIS militaryInformationService in order to avoid overloading the upstream. This is the second of three eventual PRs to add additional fields to the Lighthouse `/service_history` endpoint. It is preceded by #4014 and followed by #4016.

This PR just picks and chooses the parts of the original `EMIS::MilitaryInformation` model to replicate and updates the tests where appropriate with new VCR cassettes. There is a lot of duplication because the upstream service changed only slightly between major versions. In part because it seems like v1 has also been updated in several cases from the fields that existed two years ago when the initial recordings were taken. My team feels like duplication is an acceptable tradeoff here to keep the implementations for the different major versions of the eMIS `militaryInformationService` separate.

## Testing done

- Add updated tests for the part of `MilitaryInformation` that carry forward

#### Unique to this PR

- [ ] Add `pay_grade_code` to `/service_history`
- [ ] Add type of separation - Retired, discharged, etc. (reason provided if discharged, but nothing indicating they were retired) to `/service_history`. `discharge_status` is already included in the response; adding `narrative_reason_for_separation_txt` appears to provide the missing context. (could be 'separation_reason')

#### Applies to all PRs
- [x] Appropriate logging


rtravitz:
> Out of curiousity, how did you discover `military_information_v2_spec.rb` should be added RSpec/SubjectStub exclusion list? Did you get a linting error that lead you there?

It copies testing over from the first version, for the most part. And that version is missing the coverage mark, too, for understandable reasons. Figured there was an exception somewhere and searched for it.

PR #4021, opened 2020-03-12
fix va profile facility list

ericbuckley:
## Description of change
fixing an issue with the facility ids showing up in the user profile to not show the MVI identifiers

closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/6807

## Testing done
- enhanced an existing test case to include an MVI identifier in the response to show that the new VA profile facility list does not include that value

#### Applies to all PRs

- [ ] Appropriate logging


ericbuckley:
@johnpaulashenfelter I'm working on creating a new spec to explicitly test the new function I've introduced, but was wondering about the current implementation in master.  Based on how the [facility_specific value](https://github.com/department-of-veterans-affairs/vets-api/blame/master/config/settings.yml#L231) is being used in the [current va_patient? function](https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/models/user.rb#L191) is there a problem with the value?  Meaning should the value be `['741MM']` instead of `[['741MM']]`

ericbuckley:
> @ericbuckley i'm a little surprised that a change like this didn't require new specs -- it seems like you're changing some behavior in the filtering but there's not spec showing the new behavior. Maybe i misread the issue but it seems like there would be a spec for "here's the one that should be included" and "here's the one that shouldn't be included" because of the new filtering. Maybe one that shows a difference for the user with a vha_facility_id with 200MHS?

@johnpaulashenfelter I just added a new test case for the `va_treatment_facility_ids` function.  https://github.com/department-of-veterans-affairs/vets-api/pull/4021/commits/a5336ec6c599de216e9640ba5ac2542b44e06ee0

PR #4022, opened 2020-03-12
Remove duplicate code in swagger documentation for contact information API

lihanli:
## Description of change
https://github.com/department-of-veterans-affairs/va.gov-team/issues/6002

## Testing done
automated tests
looked at swagger json

#### Applies to all PRs

- [x] Appropriate logging


Gemfile
```diff
@@ -59,6 +59,7 @@ gem 'httpclient'
 gem 'ice_nine'
 gem 'iconv'
 gem 'iso_country_codes'
+gem 'js_regex'
```
johnpaulashenfelter: Its unclear why this gem needs to be added to remove duplicate code.

app/swagger/schemas/vet360/address.rb
```diff
@@ -6,300 +6,103 @@ module Vet360
...
-                   example: '38843',
-                   maxLength: 5,
-                   pattern: ::Vet360::Models::Address::VALID_NUMERIC_REGEX.inspect
+        def self.domestic_address_fields(block)
+          block.property :state_code,
+                         type: :string,
+                         example: 'MS',
+                         minLength: 2,
+                         maxLength: 2,
+                         pattern: SwaggerHelpers.convert_regex(::Vet360::Models::Address::VALID_ALPHA_REGEX)
```
johnpaulashenfelter: This seems like it must change the schema -- how is the new pattern different from the old one? Same question for all three instances.

lihanli: @johnpaulashenfelter 
the old pattern was incorrect, when running specs against the old pattern it would throw this error `'//v0/profile/address_validation#/addresses/0/address/zip_code' value "11249" did not match the regex '/[0-9]+/'`

according to swagger docs the pattern needs to be in javascript format and not include the starting and ending `/` characters https://swagger.io/docs/specification/data-models/data-types/

the reason the specs don't fail in master is because the `Vet360AddressSuggestion` schema doesn't include the patterns, once you add a pattern in with the old format then the specs fail

johnpaulashenfelter:
Currently not planning to merge

PR #4023, opened 2020-03-12
using the account_uuid as the redis cache index for user data

ericbuckley:
user.uuid will no longer be guaranteed when authenticating via SSOe

closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/6061

#### Applies to all PRs

- [ ] Appropriate logging

### DIFF
--- a/app/models/mvi.rb
+++ b/app/models/mvi.rb
@@ -130,7 +130,7 @@ class Mvi < Common::RedisStore
   private
 
   def response_from_redis_or_service
-    do_cached_with(key: user.uuid) do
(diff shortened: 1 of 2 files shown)


ericbuckley:
@patrickvinograd using the account id as a caching index is creating a circular dependency.  One one side we need the `account.uuid` to look for the MVI record in cache, on the other end we need the MVI record (with its edipi/icn/sec_id) to create a new Account.

Are there any other unique identifiers we could use?  What about email address + authn context?

ericbuckley:
closing in favor of #7110

PR #4025, opened 2020-03-13
validate nil form attachment param

LindseySaari:
## Description of change
Adds inputs validation to the FormAttachmentCreate concern pertaining to a [NoMethodError in Sentry](http://sentry.vfs.va.gov/vets-gov/platform-api-production/issues/90775/?referrer=slack) related to the form attachments. The validation here addresses nil input params rather than allowing NoMethodError to be raised. The namespace param in particular was being read without validating if the param was in existence.

This [original PR](https://github.com/department-of-veterans-affairs/vets-api/pull/3721) caused a breaking change by passing in the entire params hash instead of just the file data attribute for all controllers that utilized the FormAttachmentCreate concern.
 
## Testing done
- [x] Passing specs
- [x] Test upload to S3 in staging

#### Unique to this PR
- [x] Tested file upload locally
- [x]  Updated specs and ran test suite
- [x]  Updated swagger docs and corresponding tests

#### Applies to all PRs

- [x] Appropriate logging


spec/controllers/v0/hca_attachments_controller_spec.rb
```diff
@@ -11,5 +11,12 @@
 
       expect(JSON.parse(response.body)['data']['attributes']['guid']).to eq HcaAttachment.last.guid
```
omgitsbillryan: I think creating an additional expectation that the attachment file stored in the DB matches the one we uploaded? `pdf_fill/extras.pdf` in this case.

What came around to bite us last time, was the lack of happy path testing. That empty hash was getting stored to the database but we were none the wiser. We should verify that the file uploaded is the one that's stored.

PR #4027, opened 2020-03-13
alternative user UUID

ericbuckley:
## Description of change
Adding `sec_id` to the SAML user attributes for all SSOe calls, then using that value as an "alternate" uuid in the case where the user does not have a ID.me uuid value.

closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/7110

#### Reasoning

The [EPIC around migrating ID.me uuids to Account uuids](https://github.com/department-of-veterans-affairs/va.gov-team/issues/5290) goes really deep as lots of different apps store references to users via their ID.me uuid.  While an Account uuid change can address almost everything, there are a couple of situations where it becomes difficult (MVI lookup would depend on an Account record and the Account record would depend on an MVI lookup).

This PR acknowledges that at some point in time significant steps need to be made to create an internally created identifier that can be used to reference Users (a foreign key).  However, the thought is that this should be done in isolation outside the changes needed for SSOe and with deeper thought as to what the best long term solution should be (eg should we persist user records?).

The changes here are relatively simple and keep up the guarantee that all Users will have a UUID value when initialized.  Some downstream services of this value, have interpreted this value as an ID.me uuid, and they need to be updated, but most just use the value for a User foreign key, and thus just need a unique non-null value.

## Testing
- local testing to make sure login still works
- added a few functional test cases

#### Unique to this PR

#### Applies to all PRs


ericbuckley:
@patrickvinograd I know you'd like to see some changes to this PR, would you be able to list them at a high level so I can take another crack at it?

lib/saml/user.rb
```diff
@@ -49,14 +50,22 @@ def changing_multifactor?
       authn_context.include?('multifactor')
     end
 
+    def uuid
+      raise Common::Exceptions::InvalidResource, user_attributes unless user_attributes.idme_uuid or user_attributes.sec_id
+
+      return user_attributes.idme_uuid if user_attributes.idme_uuid
+
+      Digest::UUID.uuid_v3('sec-id', user_attributes.sec_id).replace('-', '')
```
patrickvinograd: i like this UUID v3 approach 👍

lib/saml/user_attributes/ssoe.rb
```diff
@@ -50,6 +54,19 @@ def email
       end
 
       ### Identifiers
+      def idme_uuid
+        return safe_attr('va_eauth_uid') if sign_in[:service_name] == 'idme'
```
patrickvinograd: this whole `sign_in` object is pretty clunky; i'm wondering if there's anything in the assertion itself that we can use to distinguish. If you look at the factories under `spec/factors/saml_attributes.rb`, there's examples  of all  the login variations, there are some CSP attributes

ericbuckley: I saw the `va_eauth_csid` attribute, but I've been unable to test any MHV / DS Logon accounts so I wasn't sure that it would be available every time.

patrickvinograd: Those factories should be current/accurate w.r.t. what traits are available, and have instances of each credential variation. `csid` looks to be equal to `idme` for any case where ID.me is brokering the authentication. I'll have data on what it looks like for inbound existing sessions pretty soon.

app/models/account.rb
```diff
@@ -56,13 +55,7 @@ def self.cache_or_create_by!(user)
...
     attrs = account_attrs_from_user(user)
 
-    accts = where(idme_uuid: attrs[:idme_uuid])
-            .where.not(idme_uuid: nil)
-            .or(
-              where(sec_id: attrs[:sec_id])
-              .where.not(sec_id: nil)
-            )
-
+    accts = where(uuid: user.uuid)
```
patrickvinograd: Wait I don't think this is right - you're trying to match up the (idme_uuid or secid_uuid) with the (VA.gov-generated) account_id

patrickvinograd: And also since we've been gradually adding secids to the accounts table for the last  few weeks, need to make sure we're querying by the raw secid, not the secid_uuid.

spec/lib/saml/ssoe_user_spec.rb
```diff
@@ -359,8 +409,42 @@
...
+        )
+      end
+    end
+
+    context 'DSLogon premium user with idme uuid in gcIds' do
+      let(:authn_context) { 'dslogon' }
+      let(:account_type) { '3' }
+      let(:highest_attained_loa) { '3' }
+      let(:multifactor) { true }
+      let(:saml_attributes) { build(:ssoe_idme_dslogon_level2, va_eauth_uid: ['0000028007'], va_eauth_csid: ['NOT_FOUND']) }
```
ericbuckley: @patrickvinograd I added this test case to trigger the new gcId parsing logic in ssoe.rb, however looking at all the new saml_attribute factories you've created it doesn't appear to be possible to get a response back from SSOe with a `va_eauth_csid != 'idme' value.  Is this because we haven't captured any SAML payloads for inbound authentication yet?

patrickvinograd: That's correct. I'll push a branch with those payloads tonight at least for visibility; if you want to cherry pick it into your PR that would work too.

lib/saml/user_attributes/base.rb
```diff
@@ -16,9 +16,17 @@ def initialize(saml_attributes, authn_context)
...
       # ID.me unique identifier
       def uuid
+        idme_uuid
+      end
+
+      def idme_uuid
         attributes['uuid']
       end
 
+      def sec_id
```
patrickvinograd: Any user_attributes derived from Base (which really means "the variants of SAML payloads we get direct from ID.me) will never have a sec_id attribute (because ID.me will never have it, nor will either of the federated credential types). So I think this just ought to return nil.

app/models/account.rb
```diff
@@ -85,7 +81,7 @@ def self.update_if_needed!(account, user)
   # @return [Hash]
   #
   def self.account_attrs_from_user(user)
-    { idme_uuid: user.uuid, edipi: user.edipi, icn: user.icn, sec_id: user.sec_id }
+    Hash[%i[idme_uuid sec_id edipi icn].map { |k| [k, user.send(k)] }]
```
saneshark: I ❤️ this from terseness standpoint, but might be more readable to do:

{
idme_uuid: user.idme_uuid,
sec_id: user.sec_id,
edipi: user.edipi,
icn: user.icn
}

lib/saml/user_attributes/ssoe.rb
```diff
@@ -1,19 +1,25 @@
 # frozen_string_literal: true
 
+require 'digest'
+
 module SAML
   module UserAttributes
-    class SSOe < Base
+    class SSOe
```
saneshark: Could this class be a candidate for SimpleDelegator / or decorator pattern? It looks like you're running up against a CodeClimate refactoring warning. Just a thought.

patrickvinograd: The reason I suggested this inheritance go away is that logically there is no structural relationship between the "Base" SAML attribute class (which represents the commonality inherent between the ID.me-issued assertions) and this SSOe class which represents an SSOe-issued assertion. 

There's a wee bit of shared logic primarily around the sign_in/account_type logic that maybe could be a mix-in or something but I think splitting apart this inheritance is the right call for clarity.

app/models/account.rb
```diff
@@ -43,11 +43,9 @@ class Account < ApplicationRecord
   # @return [Account] A persisted instance of Account
   #
   def self.cache_or_create_by!(user)
-    return unless user.uuid || user.sec_id
+    return unless get_key(user)
```
omgitsbillryan: I think the `return` statement has become unreachable code as a result of this change.

Assuming that `user.uuid` & `user.sec_id` are both `nil`:
- `user.uuid || user.sec_id` evaluates falsey
- `get_key(user)` returns "sec:" and evaluates truthy

app/models/user.rb
```diff
@@ -134,7 +135,7 @@ def edipi
   end
 
   def sec_id
-    va_profile&.sec_id
+    identity.sec_id || va_profile&.sec_id
```
omgitsbillryan: Are `identity.sec_id` and `va_profile&.sec_id` always guaranteed to be the same? Looks like we're prioritizing `identity.sec_id` in the event that both are non-nil. Making sure that's intentional / expected.

lib/saml/user_attributes/ssoe.rb
```diff
@@ -50,6 +56,27 @@ def email
       end
 
       ### Identifiers
+      def uuid
+        raise Common::Exceptions::InvalidResource, @attributes unless idme_uuid || sec_id
+
+        return idme_uuid if idme_uuid
+
+        Digest::UUID.uuid_v3('sec-id', sec_id).tr('-', '')
```
omgitsbillryan: I got a bit tripped up reading this - is `Digest::UUID.uuid_v3` some kind of convenience way of generating a "uuid-like" hash? And then we're removing the dashes?

Maybe a comment would help. Or maybe wrapping this behavior in a helper method named in a way for added readability?

app/models/account.rb
ericbuckley: good catch, the original implementation of `get_key` didn't prefix with "sec:"

app/models/user.rb
ericbuckley: To my knowledge yes, as they both come from the MVI database.  `identity.sec_id` will only be available if the user is authenticated via SSOe, and in that case should be preferred as it will be immediately available and a MVI call will not be needed.  @patrickvinograd can you confirm that my reasoning is correct?

app/models/account.rb
```diff
@@ -85,7 +81,12 @@ def self.update_if_needed!(account, user)
   # @return [Hash]
   #
   def self.account_attrs_from_user(user)
-    { idme_uuid: user.uuid, edipi: user.edipi, icn: user.icn, sec_id: user.sec_id }
+    {
+      idme_uuid: user.idme_uuid,
```
annaswims: mostly just for my benefit while reviewing - None of these are attributes stored on user.  

      idme_uuid: user.idme_uuid, - > delegated to identity
      sec_id: user.sec_id, -> identity.sec_id || Mvi.for_user(self).profile&.sec_id  (mvi is cached user.user.uuid)
      edipi: user.edipi, - identity.edipi ||  mvi&.edipi
      icn: user.icn -> delegated to mvi

lib/saml/user_attributes/ssoe.rb
ericbuckley: The `sec_id` is not a UUID, and while unique this has a potential to cause issues in downstream processes that are expecting a user UUID to be 32 bytes.  For example, if there is a log filtering process that was striping out any 32 byte id, an 10 byte sec id would be missed.  Using a one way UUID hash, will convert the sec id to a 32 byte unique identifier so that any downstream processes will treat it exactly the same as a typical 32 byte ID.me identifier.

I'm relatively new to the team, so maybe this isn't needed, but I thought this would be the safest route.

@omgitsbillryan if that logic is sound, I will add the above comment or something similar to the code.

NOTE: there is the possibility of a collision between an ID.me UUID and a hashed sec id, however that probability should be so small that I believe it can be ignored

app/models/account.rb
annaswims: and then, we should have all of those things available because the objects are found/created/instantiaed in this order:

-   Find User. if exists using uuid from saml (existing_user)
-   create UserIdentity from SAML attributes
-   Create User. uuid: @user_identity.attributes[:uuid]
-   Create Session. uuid: @user.uuid
-   In a job: find or create Account using User.uuid

app/models/user.rb
annaswims: identity.sec_id is from the v1 saml response, and va_profile&.sec_id is from MVI.  Both of those are written by the IAM team, so I'd imagine that they're the same.  If someone is using the v0 saml response we wouldn't have the `identity.sec_id` response so we could fall back to MVI

lib/saml/user_attributes/ssoe.rb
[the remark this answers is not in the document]
ericbuckley: @annaswims I don't know.  My guess is V4, but even if they did use V3 there is still a chance for collision since the namespace would be different (although the chance is tiny).

In my [comment to Bill](https://github.com/department-of-veterans-affairs/vets-api/pull/4027#discussion_r400364760) earlier was trying to imply that I think the chance of a UUID collision is minimal compared to the possibility of a downstream process assuming that all user UUIDs are going to be 32 bytes.  However, if that's a bad assumption, we should remove the hashing process and just return the raw `sec_id`

lib/saml/user_attributes/ssoe.rb
```diff
@@ -50,6 +56,27 @@ def email
...
+      end
+
+      def idme_uuid
+        return safe_attr('va_eauth_uid') if safe_attr('va_eauth_csid') == 'idme'
+
+        gcids = safe_attr('va_eauth_gcIds')&.split('|')
```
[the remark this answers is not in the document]
ericbuckley: I'll add a comment, but I believe the `va_eauth_gcIds` are equivalent to the full MVI ids that we parse from an MVI response.  @patrickvinograd could you confirm?

app/models/account.rb
ericbuckley: @annaswims based on that logic, it probably make sense to parse the ICN number out of the `va_eauth_gcIds` attribute on an SSOe response, and update the `user.icn` function to use that if possible.  That way we wouldn't need to query MVI to create an Account when a user logs in via SSOe.

app/models/user.rb
patrickvinograd: Confirming above expectation is correct.

lib/saml/user_attributes/ssoe.rb
patrickvinograd: Correct, for example the InProgressForm model has some validation around the uuid being a 32 byte uuid. 

I think the UUIDv3 vs. the UUIDv4 (ID.me) algorithms are designed to minimize the possibility of collisions to astronomically low numbers.

lib/saml/user_attributes/ssoe.rb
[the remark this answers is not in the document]
patrickvinograd: Yes, you can see examples in the `spec/factories/saml_attributes.rb` factory - it is a pipe-delimited concatenation of the MVI correlation IDs (minus the weird "base/extension" cruft).

app/models/account.rb
annaswims: ooh, that'd be nice!

patrickvinograd: Agreed, I think this could be an optimization for a future ticket, and even starts laying the groundwork for moving away from the separate MVI lookups altogether.

ericbuckley: created an implementation issue for this https://github.com/department-of-veterans-affairs/va.gov-team/issues/7424

lib/saml/user_attributes/ssoe.rb
```diff
@@ -50,6 +56,35 @@ def email
       end
 
       ### Identifiers
+      def uuid
+        raise Common::Exceptions::InvalidResource, @attributes unless idme_uuid || sec_id
+
+        return idme_uuid if idme_uuid
+
+        # The sec_id is not a UUID, and while unique this has a potential to cause issues
+        # in downstream processes that are expecting a user UUID to be 32 bytes. For
```
patrickvinograd:
```suggestion
        # in downstream processes that are expecting a user UUID to be 32 bytes. For
```

PR #4031, opened 2020-03-17
Log details for clinics who do not support VAOS

kreek:
## Description of change
Adds a log line that records clinic information for sites that don't support VAOS (use the `VAOS_400` key)

## Testing
Unit tests

Adds a warn log line with the clinic details hash

#### Applies to all PRs

### DIFF
--- a/modules/vaos/app/services/vaos/appointment_service.rb
+++ b/modules/vaos/app/services/vaos/appointment_service.rb
@@ -44,7 +44,11 @@ module VAOS
       rescue Common::Exceptions::BackendServiceException => e
         # TODO: Reevaluate the need to log clinic data three months after launch (6/15/20)
         if e.key == 'VAOS_400'
-          Rails.logger.warn('Clinic does not support VAOS appointment cancel', clinic_id: params[:clinic_id], facility_id: params[:facility_id])
+          Rails.logger.warn(
+            'Clinic does not support VAOS appointment cancel',
+            clinic_id: params[:clinic_id],
+            facility_id: params[:facility_id]
+          )
         end
         raise e
       end
--- a/modules/vaos/app/services/vaos/appointment_service.rb
+++ b/modules/vaos/app/services/vaos/appointment_service.rb
@@ -26,9 +26,7 @@ module VAOS
         }
       rescue Common::Exceptions::BackendServiceException => e
         # TODO: Reevaluate the need to log clinic data three months after launch (6/15/20)
-        if e.key == 'VAOS_400'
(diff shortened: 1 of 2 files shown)


kreek:
@johnpaulashenfelter I'll add a TODO. The log line will stay in for a couple months after launch.

modules/vaos/app/services/vaos/appointment_service.rb
```diff
@@ -24,6 +24,12 @@ def post_appointment(request_object_body)
           data: OpenStruct.new(response.body),
           meta: {}
         }
+      rescue Common::Exceptions::BackendServiceException => e
+        # TODO: Reevaluate the need to log clinic data three months after launch (6/15/20)
+        if e.key == 'VAOS_400'
+          Rails.logger.warn('Clinic does not support VAOS appointment create', clinic: params['clinic'])
```
jbalboni: Do we need to log the facility id, too? I don't think clinic ids are unique across sites.

saneshark: might be able to put some of those in as semantic logging elements / aka tags. Easier to query for later in cloud watch.

modules/vaos/app/services/vaos/appointment_service.rb
```diff
@@ -24,6 +24,12 @@ def post_appointment(request_object_body)
           data: OpenStruct.new(response.body),
           meta: {}
         }
+      rescue Common::Exceptions::BackendServiceException => e
+        # TODO: Reevaluate the need to log clinic data three months after launch (6/15/20)
+        if e.key == 'VAOS_400'
```
saneshark: NITPICK: consider putting this logic in a method -- surprised it didn't trigger code climate, but it certainly will next time someone adds one or more lines to this.

modules/vaos/app/services/vaos/appointment_service.rb
kreek: an example of the clinic hash is:
```ruby
{"site_code"=>"983", "clinic_id"=>"308", "clinic_name"=>"CHY PC KILPATRICK", "clinic_friendly_location_name"=>"Green Team Clinic1", "institution_name"=>"CHYSHR-Cheyenne VA Medical Center", "institution_code"=>"983"}
```
I think `site_code` === `facility_id` right? If I pull that out then both create and cancel log lines would be the same.

jbalboni: Yeah, I think site_code is the right facility id in this case.

PR #4032, opened 2020-03-17
evss reference data service refactor

omgitsbillryan:
## Description of change
Connects: https://github.com/department-of-veterans-affairs/va.gov-team/issues/6934

### TL;DR
Reworks how we conditionally use the evss reference data service.

### More Details
Before this PR, the `states` & `countries` routes were conditional on `Settings.evss.reference_data_service.enabled`. After this PR the routes are static and instead the usage of the reference data service client or not is conditional upon said setting.

### Why?
1. rspec speeds

Before this PR, the `spec/request/address_request_spec.rb` was reloading routes in order to test all routes (since the conditional logic lived in `routes.rb`). Such reloading is not performant. 

2. Less code / fewer routes

### Other Info
`Settings.evss.reference_data_service.enabled = false` in all our `dev`, `staging`, `review-instance`, & `prod` environments. I'm pretty sure this never got used.

## Testing
Specs only

#### Unique to this PR

#### Applies to all PRs


spec/routing/address_routing.rb
```diff
@@ -1,62 +0,0 @@
-# frozen_string_literal: true
```
omgitsbillryan: This spec was testing our conditional routing logic that no longer exists - which is why we don't need it anymore.

omgitsbillryan:
This service integration was created before we started using Flipper. 

It could be converted to use flipper instead. However, I'd consider that (along with maybe just deleting the rds service class and it's usages altogether 😱 ) a separate piece of work. The purpose of this PR is just to speed up our tests.

johnpaulashenfelter:
@omgitsbillryan did you have some before/after timing for the specs?

omgitsbillryan:
The last 2 runs of the "Run tests" stage in this PR took `8m40s` & `9m13s`.

I clicked on 3 of the latest PR builds and their run times were:
- `10m50s`
- `9m45s`
- `9m30s`

I think since we use `--parallel` in CI, there's some luck/randomness involved & the savings won't be the same as the savings when local specs run (which is about 1 minute).

omgitsbillryan:
🤦‍♂ I got lint mixed up with spec... specs are _not_ run in parallel, but I can't explain why we're not seeing a more significant. 

For context, I was getting this when running specs locally using `--profile` option:

```
Top 10 slowest example groups:
  address
    8.32 seconds average (108.12 seconds / 13 examples) ./spec/request/address_request_spec.rb:5
  ...
```

Nearly 2 minutes! For running specs on my local machine, here's the difference:
- before: `Finished in 15 minutes 41 seconds`
- after `Finished in 14 minutes 18 seconds`

app/controllers/v0/addresses_controller.rb
```diff
@@ -41,26 +41,14 @@ def states
...
-
     private
 
     def service
-      EVSS::PCIUAddress::Service.new(@current_user)
-    end
-
-    def rds_service
-      EVSS::ReferenceData::Service.new(@current_user)
+      if Settings.evss&.reference_data_service&.enabled
```
erluti, line 47: Did you consider moving this check outside the method so when the class is loaded the mode is determined instead of rechecking settings?  (I'm under the impression Settings can't be changed without restarting Rails) Might have to do some  stubbing to get tests to work though

kfrz, line 47: `Settings.evss.reference_data_service.enabled` 

Will `Settings.evss` ever be nil? Not sure we need the first guard clause

omgitsbillryan: That's clever, I hadn't considered that. 

I think I'll save myself some extra commits to fix up tests and just leave it as is for now. But I'll keep this in mind.

omgitsbillryan: I think you're referring to the safe navigator, right(`&.`)?

Ya, probably not, I'll remove it. If `Settings.evss` was ever `nil` I think this class would be the least of our problems 😄 .

PR #4033, opened 2020-03-17
Move MHV constants to only class that uses them

johnpaulashenfelter:
Found this as part of a code review. These two constants are in the settings file but only used in one file, for one method.
It seems like it makes sense to keep the information close together so its easier to understand what is going on as close to the
use of the values as possible

#### Unique to this PR

#### Applies to all PRs

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -185,10 +185,14 @@ class User < Common::RedisStore
   # Facilities in the defined range are treating facilities, indicating
   # that the user is a VA patient.
   def va_patient?
+    mhv_facility_range = [[358,718],[720,740],[742,758]].freeze
+    mhv_facility_specific = [['741MM']].freeze # 741 is excluded, but 741MM is included
+
+    # include ranges first, then individual exceptions to the ranges last.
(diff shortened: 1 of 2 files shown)


johnpaulashenfelter:
@patrickvinograd thanks for the info -- where do we inject that information into the staging environment? I looked through code, so it must be set during BRD in Jenkins?

patrickvinograd:
`devops/ansible/deployment/config/vets-api/staging-settings.local.yml.j2` - you can see all the settings overrides for staging in that environment (and likewise for dev/prod in correspondingly-named files).

As a matter of course we don't build separate variants of vets-api for  dev/staging/prod. nor do we typically like having any kind of `if env==staging` logic in Rails code - Settings is the accepted  way to inject environment-specific values.  I think a code comment mentioning that why this VHA facilities range needs to be paramaterized per-environment would be totally justified.

johnpaulashenfelter:
We definitely need a better way to deal with the various places we handle config. Closing until post-infrastructure-recommit devops project

PR #4036, opened 2020-03-18
Build Optimization: Lint only the files changed in the PR

omgitsbillryan:
## Description of change
Connects: https://github.com/department-of-veterans-affairs/va.gov-team/issues/3407

### TL;DR
Reduces the time spent linting in CI from ~1m30s down to ~15-20 seconds. This is accomplished by only linting the files changed in the PR instead of all the files in the repo.

image1
becomes:
image2

### More Details
- Adds a new Jenkins stage called `Lint Changed Files`
  - Call [`getGithubChangedFiles()`](https://github.com/department-of-veterans-affairs/va.gov-devops-jenkins-lib/commit/2e5467c3bb0df698064fb01e2f3df1e3982e828a#diff-b109130bb8389bb3abcc8bdb2b6dea9d) from the `Jenkinsfile` which calls the Github api to get a list of changed files in the PR
- Update our `make lint` build target to accept a list of files
- Update our `rake lint` task to accept a list of files to lint

## Testing
Tested in CI.


Jenkinsfile
```diff
@@ -1,3 +1,6 @@
+// TODO REMOVE THIS!!
+@Library('va.gov-devops-jenkins-lib@github-api') _
```
omgitsbillryan: Will be removed once https://github.com/department-of-veterans-affairs/va.gov-devops-jenkins-lib/pull/14 is merged.

Jenkinsfile
```diff
@@ -43,7 +43,25 @@ pipeline {
...
+        script {
+          files_to_lint = ''
+          try {
+            files_to_lint = getGithubChangedFiles('vets-api', env.CHANGE_ID.toInteger(),
+                              change_types: ['modified', 'added']).join(' ')
+          } catch(IOException e) {
+            echo "WARNING: Unable to fetch changed PR files from Github!"
+            echo "${e}"           
+          }
+        }
```
johnpaulashenfelter: @omgitsbillryan might it make sense to fall back here to linting everything instead of failing? So you'd get only changed files if things go smooth and current behavior if it doesn't with the getGithubChangedFiles method

Makefile
```diff
@@ -14,7 +14,7 @@ BASH_DEV     := $(COMPOSE_DEV) $(BASH) -c
 BASH_TEST    := $(COMPOSE_TEST) $(BASH) --login -c
 SPEC_PATH    := spec/
 DB		     := "bin/rails db:setup db:migrate"
-LINT    	 := "bin/rails lint"
+LINT    	 := "bin/rails lint['$(files)']"
```
johnpaulashenfelter: How does this effect running lint locally?

Jenkinsfile
omgitsbillryan: Good catch. I didn't realize that I undid that functionality when breaking these up into 2 separate jenkins stages. I'll put it back in place 👍

omgitsbillryan: Whoops, I didn't realize that it already does fall back to linting all fails. On line 50: `files_to_lint = ''`, then if `getGithubChangedFiles()` raises, we catch and keep marching on to line 59.

Makefile
omgitsbillryan: Tested `make lint` locally and it still works 👍 .

Did some further testing:
- `make files= lint`... lints all files👍 
- `make files='' lint` ... lints all files 👍 
- `make files='app/controllers/application_controller.rb' lint`... lints the one file 👍

Jenkinsfile
omgitsbillryan: Is there some way the code could be reworked to make this the behavior more obvious?

kfrz (approved):

PR #4037, opened 2020-03-18
Create a JSON Schema for Appeals API / HLR Create

edmangimelli:
resolves [DASH-10](https://vasdvp.atlassian.net/browse/DASH-10)

Create a JSON Schema for Appeals API - HigherLevelReview POST endpoint.

*Acceptance Criteria*
- [x] complete (and updated) JSON Schema for HLR create
- [x] use Draft 7 features
- [x] test it

this has been pretty handy for one-off testing
https://www.jsonschemavalidator.net/


modules/appeals_api/config/schemas/hlr_post.json
```diff
@@ -0,0 +1,270 @@
...
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "properties": {
+    "data":     { "$ref": "#/definitions/data" },
+    "included": { "$ref": "#/definitions/included" }
+  },
+  "additionalProperties": false,
+  "required": ["data", "included"],
+  "definitions": {
+
```
bastosmichael: What's up with your linter? Two paragraph spacers seems unnecessary.

edmangimelli: Unnecessary, but made it much easier for me to read.

bastosmichael: I understand, just want to make sure we agree that this is how we want to have it organized and we stay consistent across all of our projects using this that's all.

modules/appeals_api/config/schemas/hlr_post.json
```diff
@@ -0,0 +1,270 @@
...
+    "informalConferenceTimes": {
+      "type": "array",
+      "items": { "$ref": "#/definitions/informalConferenceTime" },
+      "minItems": 1,
+      "maxItems": 2,
+      "uniqueItems": true
+    },
+
+
+    "informalConferenceTime": {
```
bastosmichael: I would probably do the time slots themselves and the Timezone differently from a data perspective. Even if the end consumer of the PDF expects it all to be in EST time, we can calculate that period but it doesn't make any sense that someone on the west coast would have to figure out what time they need to submit on the east coast even if all of the decision making itself happens on the east coast. You're basically asking the Veteran in California if they want a 5am appointment slot and there could be confusion that he somehow thinks it's 8am his time and doesn't pay attention to the actual time zone, keeping them as seperate zones also allows us to have the actual times themselves change per zone in a conditional so if they pick PST then we give them the 5am to 7am time slot if they want it and we can translate the correct time needed on the form if it's Eastern at the pdf generation portion. I know this is probably a battle that you guys had in discussions before but feels like we could and should do better here at least from a data and schema perspective up front.

edmangimelli: So, this has been discussed over months ad nauseum, and I don't think it needs to be re-opened again. But, for a variety of reasons, I don't think we should diverge from VA form 20-0996 (possibly legal reasons).
A front-end can display time however they please --whatever formatting and timezone. But, the HLR POST endpoint gives you 4 options for time slots that are inflexible and are in eastern time, just like the PDF.
This might be something we want to escalate in the future, but I don't think we want to for MVP.

charleystran: For the record, I agree with @bastosmichael but based on @edmangimelli 's response I am not sure its open for debate. I am of the opinion we make our API as wasy to use and understand as possible to the consumers, and we heavy lift converting the time into what the PDF expects. Thats just my thoughts though on how I would want to use it from a consumer point of view.

bastosmichael: What we do in the schema is what the front end does ultimately I think it should be up for discussion, was the push back from caseflow or from the va.gov front end team?

kellylein: I don't want our API to be in the business of converting the time-zones based on the User, that should be on the front-end if they so choose.  Our API needs to be agnostic of things like that.  I know that the actual thing that the User is choosing is confusing for a multitude of reasons, but as far as the API is concerned, they are selecting up to 2 of the 4 available options.  How those options are displayed is up to the UI.  We will need to take special care to make sure our documentation is as clear as possible about what those 4 options mean, but I want to try to separate the confusion of time zones and date formats from the fact that we just need to allow the ability to select up to 2 of the 4 options.

edmangimelli: >and we heavy lift converting the time into what the PDF expects
the problem is, the veteran isn't given many options. Taking in arbitrary time ranges isn't an option (hence the string enum). We _could_ receive the 4 set ranges in UTC, but I think 1) expressing them in ET stays true to the requirement 2) expressing them in ET _also_ explains why they are what they are (if these ranges seem arbitrary now, they would seem extra arbitrary in UTC) and 3) I'd like this enum to be a reminder (and to cause devs to pause and take notice) that these time ranges are a strange requirement of the form, and you really can't put it whatever you want.

kellylein: We did some quick hallway testing last year around the best way to format the options to make them as clear as possible, and there wasn't a strong answer from the developers that responded.  Due to that, I think we need to pick the simplest implementation we can and ensure our documentation is as clear as possible, and then move forward.

bastosmichael: Without trying to push this debate on further can we consider a compromise where instead of including this data as a single string with multiple values that would be hard to validate within the context of a string that instead we have the type be something more like

```
{
    "type": "object",
    "properties": {
        "start": {
            "type": "string",
            "format": "date-time"
        },
        "end": {
            "type": "string",
            "format": "date-time",
            "minimum": {"$data": "1/start"}
        },
        "zone": {
          "type": "string",
          "enum": [
            "EST",
            "PST",
            etc, etc...
          ]
        }
    },
    "required": ["start", "end", "zone"]
}
```

It'll make this data easier set set defaults based on objects instead of strings and better to parse down the road no matter how it ends up being represented on the form.

edmangimelli: > Without trying to push this debate on further can we consider a compromise...

what we've ended up with is a compromise many times over

As an engineer, I like you solution a lot, but our hands are tied on this one.
Because the options available are so few and inflexible, I think your solution would over-complicate things. As a simple 4 value enum, validation is completely doable with JSON Schema. Your proposal is probably theoretically validatable with JSON Schema, but it would start looking pretty hairy. Also, I would be opposed to a data structure that gave the illusion that more options are available (because they aren't).

But I'm with you, I hope in the future, with HLR 2.0, we can escalate this and get something better.

My best guess is that the form is written in this way because all of the call centers / organizations that work with / respond to this form are eastern time. ¯\_(ツ)_/¯

bastosmichael: I think you're thinking that the above object type is meant to give them free reign to enter in whatever information they want, you can do a non string enumerable on a selector if you'd like and still store the data as an object instead of a string in the final payload and have it fail if the object doesn't match the only options you've been given. This both ensures the validation still returns exactly the options you're limited to without storing it in a string literal that's all. Makes it much easier down the road for more complex things that we may need or want to do. Let me know if you'd like me write up an example of what I'm talking about for you to reference but it's essentially the object type above with an enumerable list of objects as the final selector points.

kellylein: I agree with @edmangimelli that the above format gives the idea that the time ranges could be created however they want, which is definitely NOT true.  The form allows for 4 options and the Veteran can pick up to two.  I want us to not focus so much on what those options are and instead on the ability to pick 0-2 of 4 options.

bastosmichael: I'm merely talking about it from a data store perspective which is why I would consider it a compromise, there will still only be 0-2 out of 4 options but stored as an object instead of a string that's all. I can write it up if you need me to. Also if you can only pick two out of the 4 we should codify that in the schema as well by using something like 

```
{
  "type": "array",
  "items": {
    "type": "string",
    "enum": [
        "800-1000 ET",
        "1000-1230 ET",
        "1230-1400 ET",
        "1400-1630 ET"
      ]
  },
  "uniqueItems": true,
  "minItems": 2
}
```
This will ensure that it will fail it if they pick 3 or more etc.

edmangimelli: As of yesterday, here's the schema:
```
    "informalConferenceTimes": {
      "type": "array",
      "items": { "$ref": "#/definitions/informalConferenceTime" },
      "minItems": 1,
      "maxItems": 2,
      "uniqueItems": true
    },


    "informalConferenceTime": {
      "type": "string",
      "enum": [
        "800-1000 ET",
        "1000-1230 ET",
        "1230-1400 ET",
        "1400-1630 ET"
      ]
    },
```
the checks you're refering to have already been put into place

edmangimelli: (although 0 items is forbidden. If you are including that field, you can't include an empty array)

bastosmichael: Got it I missed the difference between the string vs the array.

edmangimelli: Basically.
We're creating the digital version of this:
https://www.vba.va.gov/pubs/forms/VBA-20-0996-ARE.pdf

edmangimelli: > I can write it up if you need me to.

No need. I understand that this can be done.

I think you're assuming that we’ll be parsing this string at some point-- that is not the case.
We could (and considered at one point) using the enum `[0,1,2,3]`.

I think you're assuming that either

A) we can safely diverge from form 20-0996

or

B) that getting those in charge to change this form is easy (and is a fight we're even interested in fighting at this moment).

My understanding of the situation:

A) No
B) No (No)


From the data storage perspective:

My hope is that the column that will store this will be an enum type (so, internally, just the integers 0,1,2,3)
https://api.rubyonrails.org/classes/ActiveRecord/Enum.html

Storing these as structured data of start and stop times, even with validations in place, is putting into place an abstraction that we _think_ we're going to need in the future, which i think would be a bad move. I don't think we should be coding stuff that we _might_ want in the future.

bastosmichael: We'll be storing the entire json payload generated from the schema as a whole so I don't think we'll need the 0,1,2,3 in integer format like you're saying but I won't fight you anymore on this as I don't think we'll agree but I'll have to trust your judgement here. I would prefer data be stored in it's proper state and format but if you're saying we are just taking the '800-1000 ET' and setting it as that in the pdf itself so string to string and you just want to simplify it so that it gets added the way it's selected then I'll forgo pushing back on this.

edmangimelli: @bastosmichael , yeah, we could get really buck wild if we wanted to:
```
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "properties": { "informalConferenceTimes": { "$ref": "#/definitions/informalConferenceTimes" } },
  "required": [ "informalConferenceTimes" ],
  "definitions": {

    "informalConferenceTimes": {
      "type": "array",
      "items": { "$ref": "#/definitions/timeRange" },
      "minItems": 1,
      "maxItems": 2,
      "uniqueItems": true
    },

    "timeRange": {
      "oneOf": [
        {
          "type": "object",
          "properties": {
            "start": {
              "type": "object",
              "properties": { "hour": { "const": 8 }, "minute": { "const": 0 } },
              "required": [ "hour", "minute" ]
            },
            "stop": {
              "type": "object",
              "properties": { "hour": { "const": 10 }, "minute": { "const": 0 } },
              "required": [ "hour", "minute" ]
            },
            "zone": { "const": "UTC-5" }
          },
          "required": [ "start", "stop", "zone" ]
        },
        {
          "type": "object",
          "properties": {
            "start": {
              "type": "object",
              "properties": { "hour": { "const": 10 }, "minute": { "const": 0 } },
              "required": [ "hour", "minute" ]
            },
            "stop": {
              "type": "object",
              "properties": { "hour": { "const": 12 }, "minute": { "const": 30 } },
              "required": [ "hour", "minute" ]
            },
            "zone": { "const": "UTC-5" }
          },
          "required": [ "start", "stop", "zone" ]
        },
        {
          "type": "object",
          "properties": {
            "start": {
              "type": "object",
              "properties": { "hour": { "const": 12 }, "minute": { "const": 30 } },
              "required": [ "hour", "minute" ]
            },
            "stop": {
              "type": "object",
              "properties": { "hour": { "const": 14 }, "minute": { "const": 0 } },
              "required": [ "hour", "minute" ]
            },
            "zone": { "const": "UTC-5" }
          },
          "required": [ "start", "stop", "zone" ]
        },
        {
          "type": "object",
          "properties": {
            "start": {
              "type": "object",
              "properties": { "hour": { "const": 14 }, "minute": { "const": 0 } },
              "required": [ "hour", "minute" ]
            },
            "stop": {
              "type": "object",
              "properties": { "hour": { "const": 16 }, "minute": { "const": 30 } },
              "required": [ "hour", "minute" ]
            },
            "zone": { "const": "UTC-5" }
          },
          "required": [ "start", "stop", "zone" ]
        },
        {
          "type": "object",
          "properties": {
            "start": {
              "type": "object",
              "properties": { "hour": { "const": 13 }, "minute": { "const": 0 } },
              "required": [ "hour", "minute" ]
            },
            "stop": {
              "type": "object",
              "properties": { "hour": { "const": 15 }, "minute": { "const": 0 } },
              "required": [ "hour", "minute" ]
            },
            "zone": { "const": "UTC+0" }
          },
          "required": [ "start", "stop", "zone" ]
        },
        {
          "type": "object",
          "properties": {
            "start": {
              "type": "object",
              "properties": { "hour": { "const": 15 }, "minute": { "const": 0 } },
              "required": [ "hour", "minute" ]
            },
            "stop": {
              "type": "object",
              "properties": { "hour": { "const": 17 }, "minute": { "const": 30 } },
              "required": [ "hour", "minute" ]
            },
            "zone": { "const": "UTC+0" }
          },
          "required": [ "start", "stop", "zone" ]
        },
        {
          "type": "object",
          "properties": {
            "start": {
              "type": "object",
              "properties": { "hour": { "const": 17 }, "minute": { "const": 30 } },
              "required": [ "hour", "minute" ]
            },
            "stop": {
              "type": "object",
              "properties": { "hour": { "const": 19 }, "minute": { "const": 0 } },
              "required": [ "hour", "minute" ]
            },
            "zone": { "const": "UTC+0" }
          },
          "required": [ "start", "stop", "zone" ]
        },
        {
          "type": "object",
          "properties": {
            "start": {
              "type": "object",
              "properties": { "hour": { "const": 19 }, "minute": { "const": 0 } },
              "required": [ "hour", "minute" ]
            },
            "stop": {
              "type": "object",
              "properties": { "hour": { "const": 21 }, "minute": { "const": 30 } },
              "required": [ "hour", "minute" ]
            },
            "zone": { "const": "UTC+0" }
          },
          "required": [ "start", "stop", "zone" ]
        }
      ]
    }
  }
}
```

example input:
```
{
  "informalConferenceTimes": [
    {
      "start": { "hour": 17, "minute": 30 },
      "stop": { "hour": 19, "minute": 0 },
      "zone": "UTC+0"
    },
    {
      "start": { "hour": 19, "minute": 0 },
      "stop": { "hour": 21, "minute": 30 },
      "zone": "UTC+0"
    },
  ]
}
```

edmangimelli: ^^^ and this isn't even right because JSON Schema's uniqueness constraint isn't aware that
```
    {
      "start": { "hour": 14, "minute": 0 },
      "stop": { "hour": 16, "minute": 30 },
      "zone": "UTC-5"
    },
```
```
    {
      "start": { "hour": 19, "minute": 0 },
      "stop": { "hour": 21, "minute": 30 },
      "zone": "UTC+0"
    },
```
are the same thing

edmangimelli: validation / readability / maintainability suffer quickly, IMO

edmangimelli: > Can you include a GET end point and swagger that's the same as the POST but exposes the schema externally?
> 
> https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/config/routes.rb#L14 Example of get end point route
> 
> https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/app/controllers/claims_api/base_form_controller.rb#L13 Example of schema being exposed in controller

ah. I assumed that fell under's @charleystran 's _schema implementation_ DASH-11 ticket.

edmangimelli:
> Can you include a GET end point and swagger that's the same as the POST but exposes the schema externally?
> 
> https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/config/routes.rb#L14 Example of get end point route
> 
> https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/app/controllers/claims_api/base_form_controller.rb#L13 Example of schema being exposed in controller

ah. I assumed that fell under's @charleystran 's _schema implementation_ DASH-11 ticket.

bastosmichael:
His is explicit about implementing the POST so I think it's safe to say that you can do the GET unless you think it's really too out of scope for this?

PR #4039, opened 2020-03-18
Model- Old 526 Submission Flow Cleanup

SarahJaine:
## Description of change

**Part 1 of 2 for [va.gov-team 6822](https://github.com/department-of-veterans-affairs/va.gov-team/issues/6822)**
Remove unused code related to abandoned asynchronous approach to form 526 submissions.
Second PR is https://github.com/department-of-veterans-affairs/vets-api/pull/4043, contains `db/` related changes

#### Unique to this PR

- [x] Remove file defining `VA526ezSubmitTransaction` Class and related test specs
- [x] Remove `migrate_dry_run` and `migrate_data` rake tasks from `form526.rake` file

#### Applies to all PRs
- [n/a] Swagger docs have been updated, if applicable
- [n/a] Provide which alerts would indicate a problem with this functionality (if applicable)


db/migrate/20200317162403_remove_disability_compensation_job_statuses.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveDisabilityCompensationJobStatuses < ActiveRecord::Migration[5.2]
+  def change
+    drop_table :disability_compensation_job_statuses
```
SarahJaine: I don't really like using `change` here because the automagic `down` would be to create the table.  This doesn't actually reverse the effects of dropping a table because creating the table doesn't reinstate dropped data, dropping a table is in reality irreversible. So, I originally _wanted_ to write`up/down` functions instead and have my `down` `raise ActiveRecord::IrreversibleMigration` but that kind of stinks too because there isn't any data in this table or the `disability_compensation_submissions` table in prod-- it seems silly to fail rollbacks for this. In summation, all these solutions are imperfect, so I picked the least controversial way forward.

PR #4042, opened 2020-03-18
Allow users to redirect to external application after SSOe authentication

ericbuckley:
## Description of change
Cerner users that use the new standalone login page, need to authenticate and return to the cerner application.  This PR adds a new redis store, `LoginRedirectApplication` to temporarily store the users request to redirect out.  It is keyed on the SAML request id, and thus can be looked up on return using the `in_response_to` attribute value on the SAML response.

The exact URLs to redirect a user to will be configured in the settings module with a list of predetermined keys.  So for example if a request was made to authenticate via `GET /v1/sessions/idme/new?application=cerner`, upon a successful authentication with SSOe, the user would be redirected back to "https://cernercare.com/".

closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/6917

## Testing
- a new unit test for the `RedisStore.pop` method
- a new functional test on the v1 sessions controller new method to show the application value is stored
- a new functional test on the v1 sessions controller saml callback to show that if a LoginRedirectApplication value exists, the controller will redirect to that value

#### Applies to all PRs


config/redis.yml
```diff
@@ -92,6 +92,9 @@ development: &defaults
   va_mobile_session:
     namespace: va-mobile-session
     each_ttl: 855
+  login_redirect_application:
+    namespace: login-redirect-application
+    each_ttl: 300 # 5 minutes
```
ericbuckley: @patrickvinograd thinking through this TTL, would we want to give the user more than 5 minutes to login?  Is it possible that a signup process, with verification procedures will take longer than this?

annaswims: Good call.  5 minutes is way to short for a first-time sign up. After watching a user research session with a blind veteran who used a screen reader who needed to both verify his email address and set up 2fa I'd say minimum of 30 minutes, but that 1h would be perfectly reasonable.

ericbuckley: thanks Anna, just increased this to 1 hour

lib/saml/url_service.rb
```diff
@@ -182,5 +189,15 @@ def add_query(url, params)
...
       end
     end
+
+    def create_redirect_application(uuid)
+      LoginRedirectApplication.create(
+        uuid: uuid, redirect_application: @redirect_application
+      )
+    end
+
+    def find_redirect_application(uuid)
```
annaswims: nitpick: a method called `find` doesn't' strike me a something that pops a record from redis.

app/models/login_redirect_application.rb
```diff
@@ -0,0 +1,21 @@
...
+
+require 'common/models/redis_store'
+
+# Users logging in via SSOe may need to be redirected back to an external
+# application after authentication rather than the VA.gov home page.  When a
+# authentication request comes in, with the necessary parameter, we can use
+# this Redis namespace to temporarily store a value so that when a matching
+# SAML response comes back from SSOe we know where to redirect the newly
+# authenticated user.
+class LoginRedirectApplication < Common::RedisStore
```
annaswims: I really like this and think we could use this to do a better job of tracking login metrics (timing, abandon rate), so I'm wondering if we could give this a more generic name so that we could use it for other login-related attributes in the future.

patrickvinograd:
Interesting, in my head this redirect was going to happen on the frontend - i.e. normal SAML callback process would happen and land back at the sign-in page app, which would have enough context to know to bounce the user back to Cerner. 

I don't know if I had a strong rationale for that, and it is an unnecessary set of page loads/redirects. 

@hinzed1127 does this approach (where the return path never arrives at the sign-in page) cause any issues that you can see?

PR #4044, opened 2020-03-19
Remove config.cache_classes true (default to false) in test env

kreek:
# Description of change
I believe since the Rails 5 upgrade there's been an issue for those that develop against docker version of vets-api. A code change in-between rspec runs would cause a `LoadError` to occur. This made developing with docker much slower than native as you'd have to prepend `bin/spring stop && ` before each `bin/rspec ...` command to pick up any changes.

Removing `config.cache_classes = true` and using the default value of `false` fixes the issue 

## Testing
Local unit tests/code changes/unit tests with docker

Docker rspec runs no longer throw LoadError when code changes

#### Applies to all PRs


kreek:
😐apparently fixed docker but breaks ci

PR #4049, opened 2020-03-19
Updated make command in README for Docker configuration.

U-DON:
## Description of change
After changes in #4029, running `make up` in the Docker configuration started causing issues with logins. This is because the default `make up` behavior starts a test environment, which doesn't create the `vets_api_development` database. It should be clear to developers using the Docker configuration for local development that they should run the following:

```
env=dev make up
```

## Testing
Ran `make up` with and without the environment variable and found that the `vets_api_development` DB only gets created when the environment is specified as `dev`.

#### Applies to all PRs

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ VA.gov Team Engineers should follow instructions [here](https://github.com/depar
 
 Once you have the `EXCLUDE_SIDEKIQ_ENTERPRISE` set you can run the application with:
 ```
-make up
+env=dev make up
 ```
 
 You should then be able to navigate to [http://localhost:3000/v0/status](http://localhost:3000/v0/status) in your


johnpaulashenfelter:
@LindseySaari i think you might want to take a look at this

U-DON:
Let me know if this is an accurate instruction, and also if I'm missing other places where the environment variable makes a significant difference in the `make` command. I see that it's different for a few others, but I'm not sure it makes enough of a difference to point out. Or perhaps there's a better place in the docs to explain how to use the variable.

LindseySaari:
This PR can be closed, we are working on a change to default the env arg to dev [here](https://github.com/department-of-veterans-affairs/vets-api/pull/4050) @johnpaulashenfelter

jholton:
A fix making this change (`env=dev`) unnecessary was merged in #4050

PR #4052, opened 2020-03-23
log vet360 async transaction failures

lihanli:
## Description of change
https://github.com/department-of-veterans-affairs/va.gov-team/issues/6847

## Testing

tested in staging
#### Applies to all PRs


lihanli:
> PII should go in PersonalInformationLog not Sentry

@annaswims we don't need to record the PII but want the errors to show up in sentry, i think that all the fields that are sensitive are already filtered

http://sentry.vfs.va.gov/vets-gov/platform-api-staging/issues/109900/?query=is%3Aunresolved%20error%3Avet360

PR #4053, opened 2020-03-23
Hlr validate endpoint

charleystran:
## Description of change
This PR does 4 things
1. exposes a validation endpoint for the appeals_api HLR controller
2. calls the validation against the JSON schema in the create and validate endpoint
3. refactors the json validation logic out of the claims api to a shared library
4. Exposes the JSON schema for hlr via a get request

## Testing
rspec

- [ ] HLR create and validate endpoints hit the json schema for validation

#### Applies to all PRs


lib/json_schema/form_schemas.rb
```diff
@@ -0,0 +1,35 @@
...
+    def schemas
+      return_val = {}
+
+      Dir.glob(File.join(base_dir, '/*')).each do |schema|
+        schema_name = File.basename(schema, '.json').upcase
+        return_val[schema_name] = MultiJson.load(File.read(schema))
+      end
+
+      return_val
+    end
```
edmangimelli: should we cache this ?
```
def schemas
  schemas ||= get_schemas
end

def get_schemas
  return_val = {}

  Dir.glob....
```
(or all in one method)

charleystran: Yea we could do that.

modules/appeals_api/app/controllers/appeals_api/v1/decision_review/higher_level_reviews_controller.rb
```diff
@@ -1,22 +1,50 @@
...
   end
 
   private
 
-  def create_params
+  def validate_json_schema
+    AppealsApi::FormSchemas.new.validate!(self.class::FORM_NUMBER, @json_body)
+  rescue JsonSchema::JsonApiMissingAttribute => e
+    render json: e.to_json_api, status: e.code
+  end
```
edmangimelli: i thought this would be in a base controller that HigherLevelReviewController would inherit from ? isn't that how the claims stuff works?

charleystran: Claims works that way because we have multiple forms/controllers/api_versions needing to share. Until appeals has a second form to share the logic, it doesnt need to inherit from anything (yet).

edmangimelli: 👍

lib/json_schema/form_schemas.rb
charleystran: done

lib/json_schema/form_schemas.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+# require_relative './json_schema/json_api_missing_attribute'
+
+module JsonSchema
+  class FormSchemas
+    def base_dir
```
bastosmichael: Is this too much future proofing?

charleystran: Its really just an abstraction out of what was in claims_api in modules/claims_api/lib/claims_api/form_schemas.rb which has been moved out so claims and appeals can share the code

modules/appeals_api/spec/config/schemas/200996.json.spec.rb
```diff
@@ -3,13 +3,6 @@
 require 'rails_helper'
 require 'pp'
 
-# it('has no errors') do
-#   pp errors.map { |error| error['data_pointer'] } unless errors.empty?
-#   puts
-#   pp errors unless errors.empty?
-#   expect(errors).to be_empty
-# end
-
```
edmangimelli: thanks

modules/appeals_api/app/controllers/appeals_api/v1/decision_review/higher_level_reviews_controller.rb
```diff
@@ -1,22 +1,50 @@
...
+require 'json_schema/json_api_missing_attribute'
+require 'appeals_api/form_schemas'
+require_dependency 'appeals_api/concerns/json_format_validation'
+
 class AppealsApi::V1::DecisionReview::HigherLevelReviewsController < AppealsApi::ApplicationController
+  include AppealsApi::JsonFormatValidation
+
   skip_before_action(:authenticate)
+  before_action :validate_json_format, if: -> { request.post? }
+  before_action :validate_json_schema, only: %i[create validate]
```
edmangimelli: these names threw me off for a sec. I was thinking `validate_json_schema` --he's making sure our json schema is valid with each create or validate call?
i don't know what would be better. i'm probably too long winded:
```
before_action :ensure_request_body_is_json
before_action :validate_request_json
```
(_not_ a blocking comment)

lib/json_schema/form_schemas.rb
edmangimelli: i love that appeals and clams are sharing this. very cool

modules/appeals_api/app/controllers/appeals_api/v1/decision_review/higher_level_reviews_controller.rb
```diff
@@ -1,22 +1,50 @@
...
-    render_response(Appeals::Service.new.create_higher_level_review(create_params))
+    render json: { data: { success: true } }
+  end
+
+  def validate
+    render json: validation_success
+  end
+
+  def schema
+    render json: { data: [AppealsApi::FormSchemas.new.schemas[self.class::FORM_NUMBER]] }
```
edmangimelli: maybe in the future we can think about convert the form schema stuff into a concern so schema files aren't being read all the time.

edmangimelli: or caching the schemas at the class level of  `AppealsApi::JsonFormatValidation`

PR #4054, opened 2020-03-23
Update to Concatenate Pay Plan and Pay Grade

dbrown-va:
## Description of change
This PR supports https://vasdvp.atlassian.net/browse/IN-137.

In order to determine the grade for a military service member, a combination of Pay Plan Code and Pay Grade Code is needed.
Pay Grade is a value between 01 and 10.
Pay Plan is represented as a two-character code, and can be one of five values:

 - CC (Commissioned Corps)
 - MC (Cadet)
 - ME (Enlisted)
 - MO (Officer)
 - MW (Warrant Officer)

Pay Plan needs to be concatenated with Pay Grade to determine the full grade, with the leading character (M or C) stripped from Pay Plan. For example, if a service member had Pay Plan=MW and Pay Grade=01, the concatenated value to be returned in the API would be W01.

Pay Plan exists in the eMIS getMilitaryServiceEpisodes response, so there's no additional service connectivity needed. It needs to be mapped/translated as mentioned above, and the final value can remain in the existing API field being used for Pay Grade.

## Testing

Added an additional unit test based on the pay grade calculation.

- `Pay Plan` and `Pay Grade` are concatenated in the `Pay Grade` field as described above.
- The API's documentation is updated to describe how this data is treated.

#### Unique to this PR

- [x] Adds new `build_pay_grade` method that performs desired `pay_plan_code` & `pay_grade_code` concatenation. 

#### Applies to all PRs

### DIFF
--- a/modules/veteran_verification/VETERAN_VERIFICATION.yml
+++ b/modules/veteran_verification/VETERAN_VERIFICATION.yml
@@ -267,6 +267,7 @@ components:
                  - ME (Enlisted)
                  - MO (Officer)
                  - MW (Warrant Officer)
+                 
                 Pay Grade Code is a value between 01 and 10.
                 Pay Plan Code is concatenated with Pay Grade Code to determine the full Pay Grade value, with the leading character (M or C) stripped from Pay Plan Code.
               example: "W01"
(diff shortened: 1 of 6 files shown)


dbrown-va:
Definitely.  I've added a partially invalid cassette that is missing the pay plan code and a corresponding test to verify the expected `unknown` response for `pay_grade`.

PR #4055, opened 2020-03-24
fix: using "myvahealth" as the key for a login redirect

ericbuckley:
Instead of "cerner", use "myvahealth" as the application redirect value when a user lands on the standalone login page.

Also adding a check/spec for handling unknown application values.

closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/6917

## Testing
- updated existing specs with new application key name
- added a new spec for ignoring unknown application redirect values

#### Applies to all PRs

### DIFF
--- a/spec/controllers/v1/sessions_controller_spec.rb
+++ b/spec/controllers/v1/sessions_controller_spec.rb
@@ -200,10 +200,10 @@ RSpec.describe V1::SessionsController, type: :controller do
 
       it 'redirect user to external site' do
         LoginRedirectApplication.create(
-          uuid: login_uuid, redirect_application: Settings.sso.ssoe_redirects['myvahealth']
+          uuid: login_uuid, redirect_application: Settings.ssoe.redirects['myvahealth']
         )
         allow(SAML::User).to receive(:new).and_return(saml_user)
-        expect(post(:saml_callback)).to redirect_to(Settings.sso.ssoe_redirects['myvahealth'])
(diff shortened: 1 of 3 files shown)


ericbuckley:
missed something, guessing I should make the application name check case-insensitive.  Meaning `/v1/session/idme/new?application=MyVAHealth` should also work

spec/controllers/v1/sessions_controller_spec.rb
```diff
@@ -187,10 +200,10 @@ def expect_logger_msg(level, msg)
 
       it 'redirect user to external site' do
         LoginRedirectApplication.create(
-          uuid: login_uuid, redirect_application: 'cerner'
+          uuid: login_uuid, redirect_application: Settings.sso.ssoe_redirects['myvahealth']
         )
         allow(SAML::User).to receive(:new).and_return(saml_user)
-        expect(post(:saml_callback)).to redirect_to(Settings.sso.ssoe_redirects['cerner'])
+        expect(post(:saml_callback)).to redirect_to(Settings.sso.ssoe_redirects['myvahealth'])
```
patrickvinograd: Looks good, I would consider another test case  here capturing the ultimate intended redirect behavior for an unknown redirect_application.

ericbuckley: this test case captures it from perspective of when authentication starts, https://github.com/department-of-veterans-affairs/vets-api/pull/4055/files#diff-b6ff071088096f7d49f49f8b428ab449R133-R144

PR #4056, opened 2020-03-24
Jh/csrf protection via headers

jholton:
#### Unique to this PR

#### Applies to all PRs


app/controllers/application_controller.rb
```diff
@@ -8,6 +8,10 @@
 require 'aes_256_cbc_encryptor'
 
 class ApplicationController < ActionController::API
+  include ActionController::RequestForgeryProtection
+  protect_from_forgery with: LoggingForgeryStrategy, if: -> { ActionController::Base.allow_forgery_protection }
```
jholton: `LoggingForgeryStrategy` is a temporary "strategy" for handling violations.  That being, logging them to Sentry for evaluation.
Will change this to `with: :exception` when ready to turn it on, which will raise `ActionController::InvalidAuthenticityToken` exception.

erluti: This looks really clean.  Does this mean you need to create a LoggingForgeryStrategy?

app/controllers/concerns/logging_forgery_strategy.rb
```diff
@@ -0,0 +1,24 @@
...
+      'Request susceptible to CSRF',
+      :info,
+      controller: @controller.controller_name,
+      action: @controller.action_name
+    )
+  end
+
+  private
+
+  def request
```
erluti: do you need this method?

jholton: Good catch.  Left over from copy pasta

jholton:
Corresponding frontend PR is [here](https://github.com/department-of-veterans-affairs/vets-website/pull/12005).

app/controllers/application_controller.rb
```diff
@@ -53,6 +57,10 @@ def clear_saved_form(form_id)
 
   attr_reader :current_user
 
+  def set_csrf_header
+    response.set_header('X-CSRF-Token', session[:_csrf_token] || form_authenticity_token)
```
erluti: Does rails normally re-use a session-stored token?  I was thinking Rails generates that over again for each form in normal protect mode, and was wondering if that might be a security concern?

jholton: Yeah, you're right.  I had it generating a new one every request, but concerned it'd be too easy to end up with tokens that don't validate when they should (what if someone has two tabs open with different tokens in each?)
But, it looks like even though they are different they will decode to all be valid.  See [here](https://stackoverflow.com/questions/47723379/why-does-the-csrf-token-in-rails-not-prevent-multiple-tabs-from-working-properly/49783739#49783739).

I'm going to change it to generate a new one on each request and verify that it works in this implementation.

jholton: Reverted.  Thanks.

app/controllers/external_api_application_controller.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+# controllers that aren't a part of the vets-website backend should inherit from this controller
+class ExternalApiApplicationController < ApplicationController
+  skip_before_action :verify_authenticity_token
+  skip_after_action :set_csrf_header
+end
```
omgitsbillryan: I'm not fond of this pattern, that is, adding inheritance between a controller and application_controller. It quickly gets unreadable once there's more than a couple of layers of inheritance up until application_controller.

Have you considered using a mixin instead? Something like, `include CsrfProtection`, and then have:

```ruby
included do
  skip_before_action :verify_authenticity_token
end
```

omgitsbillryan: Also, I'm not sure how realistic it is to expect that other developers will know that this controller exists (i.e. `controllers that aren't a part of the vets-website backend should inherit from this controller`). I'd imagine they would just manually write

```ruby
  skip_before_action :verify_authenticity_token
```
at the top of their controllers.

jholton: Refactored and removed `ExternalApiApplicationController` controller. 
Whether it's using a new parent controller or having to manually `skip` the before filters, there's too much implicit universal (for Lighthouse) required code here that should be built in to all Lighthouse endpoints.

spec/request/csrf_request_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+    ActionController::Base.allow_forgery_protection = true
+  end
+
+  after(:all) do
+    Settings.sentry.dsn = nil
+    ActionController::Base.allow_forgery_protection = @original_val
+  end
+
+  before do
+    # innocuous route chosen for setting the CSRF token in the cookie
```
erluti, line 20:
```suggestion
    # innocuous route chosen for setting receiving the CSRF token in a header
```

erluti: or something to keep the comment in line with the implementation

jholton: fixed.  thanks.

app/controllers/v0/education_benefits_claims_controller.rb
```diff
@@ -14,7 +14,7 @@ def create
 
       StatsD.increment("#{stats_key}.success")
       Rails.logger.info "ClaimID=#{claim.id} RPO=#{claim.education_benefits_claim.region} Form=#{form_type}"
-      validate_session
+      load_user
```
omgitsbillryan, line 17: I didnt realize that `validate_session` was being called outside of authentication_and_sso_concerns.rb. What's the motivation for a controller action to both:

1. `skip_before_action(:authenticate)`

and,

2. call `validate_session` from within a specific action

? I suppose this question is just for my own understanding and not to do with this PR.

spec/request/csrf_request_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+      Rails.application.reload_routes!
+    end
+
+    %i[post put patch delete].each do |verb|
+      context "for #{verb.upcase} requests" do
+        context 'without a CSRF token present' do
+          it 'raises an exception' do
+            expect(Raven).to receive(:capture_message).with('Request susceptible to CSRF', level: 'info')
+            send(verb, '/csrf_test')
+            # expect(response.status).to eq 500
```
kfrz, line 45: What's the purpose of this commented expectation?

spec/request/csrf_request_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+  end
+
+  # SAML callback
+  describe 'POST SAML callback' do
+    context 'without a CSRF token' do
+      context 'v0' do
+        it 'does not raise an error' do
+          expect(Raven).not_to receive(:capture_message).with('Request susceptible to CSRF', level: 'info')
+          post(auth_saml_callback_path)
+          # expect(response.body).not_to match(/ActionController::InvalidAuthenticityToken/)
```
kfrz, line 84: What's the purpose of this commented expectation?

spec/request/csrf_request_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+        end
+      end
+
+      context 'v1' do
+        it 'does not raise an error' do
+          expect(Raven).not_to receive(:capture_message).with('Request susceptible to CSRF', level: 'info')
+          post(v1_sessions_callback_path)
+          # expect(response.body).not_to match(/ActionController::InvalidAuthenticityToken/)
```
kfrz, line 92: What's the purpose of this commented expectation?

spec/request/csrf_request_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+        'X-VA-Gender': 'M' }
+    end
+    let(:path) { '/services/claims/v0/forms/526' }
+    let(:data) { File.read(Rails.root.join('modules', 'claims_api', 'spec', 'fixtures', 'form_526_json_api.json')) }
+
+    it 'skips CSRF validation' do
+      post path, params: data, headers: headers
+      expect(response.status).to eq(200)
+      expect(Raven).not_to receive(:capture_message).with('Request susceptible to CSRF', level: 'info')
+      # expect(response.body).not_to match(/ActionController::InvalidAuthenticityToken/)
```
kfrz, line 117: What's the purpose of this commented expectation?

spec/spec_helper.rb
```diff
@@ -133,4 +133,24 @@
     example.run
     Timecop.return
   end
+
+  # enable `allow_forgery_protection` in Lighthouse specs to ensure that the endpoints
+  # in those modules have explicitly skipped the CSRF protection functionality
+  lighthouse_dirs = %r{
+    modules/
+    (appeals_api|claims_api|openid_auth|va_facilities|va_forms|vba_documents|
+      veteran|veteran_confirmation|veteran_verification)/
```
omgitsbillryan, line 142: Can this somehow be defined with a wildcard in the event that a module is added/modified/removed? Would be a serious pain to understand what the heck is going on here for some poor developer in the future.

spec/spec_helper.rb
```diff
@@ -133,4 +133,24 @@
...
+  end
+
+  config.before(:all, :enable_csrf_protection) do
+    @original_allow_forgery_protection = ActionController::Base.allow_forgery_protection
+    ActionController::Base.allow_forgery_protection = true
+  end
+
+  config.after(:all, :enable_csrf_protection) do
+    ActionController::Base.allow_forgery_protection = @original_allow_forgery_protection
+  end
```
omgitsbillryan, line 155: I think maybe you want `config.before(:suite, ...`, since it will run only once, whereas `:all` will run before each example group.

spec/request/csrf_request_spec.rb
jholton: When we roll this out it will be only logging to Sentry for a bit (a few days?) so we can see if any routes aren't being covered properly by CSRF protection.  Then we'll switch it to actually raise the exception.  At that point I'm going to change these specs.  i.e. remove the `Raven` expectations and start checking that the exception is being raised.

app/controllers/v0/education_benefits_claims_controller.rb
jholton: On most (all?) of them, they are forms that can be completed by logged-in or logged-out users.  If they were logged-in when completing the form, there will be an `in_progress_forms` record to delete after submission.  Loading the user allows the record deletion to success.  See [`#clear_saved_form`](https://github.com/department-of-veterans-affairs/vets-api/blob/d834efae2725b6a7eeaabb033cdd51d85d8a52a0/app/controllers/application_controller.rb#L42)

The `gi_bill_feedbacks_controller` also includes user data in the form, if present.

spec/spec_helper.rb
jholton: I agree it's not ideal. We are relying on any future Lighthouse modules being added manually too.
I'm open to suggestions.  There's nothing in there naming that identifies them as Lighthouse manuals.  And I can't just include everything in `modules`. `vaos`, for example, lives there too.

spec/spec_helper.rb
jholton: I think you're right.  I was messing with `:all` vs `:suite` earlier, and I thought I changed it.  Obviously I didn't!

jholton: OK, we were wrong. You can't use `:suite` here:
> WARNING: `:suite` hooks do not support metadata since they apply to the suite as a whole rather than any individual example or example group that has metadata. The metadata you have provided ([:enable_csrf_protection]) will be ignored.

PR #4057, opened 2020-03-25
install Sidekiq pro/ent if license is present

jholton:
## Description of change
Changes Gemfile to install Sidekiq pro/ent if license is present
Removes need for `EXCLUDE_SIDEKIQ_ENTERPRISE` var

Developers often run into the scenario where they don't have the license set or the `EXCLUDE_SIDEKIQ_ENTERPRISE` variable set.  Then installation blows up when Bundler tries to install Sidekiq pro/ent.

With this change Sidekiq Pro/Ent gets installed if the license is present and it doesn't if the license is not present.  That's it.  No extra config.  No blow-ups.

TODO: Remove references to `EXCLUDE_SIDEKIQ_ENTERPRISE` from documentation.

## Testing
N/A

#### Unique to this PR

- [x] Check for presence of Sidekiq Ent license before installing Sidekiq Pro/Ent gems

#### Applies to all PRs

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -173,7 +173,7 @@ end
 group :production do
   # sidekiq enterprise requires a license key to download but is only required in production.
   # for local dev environments, regular sidekiq works fine
-  if Bundler::Settings.new['enterprise.contribsys.com'] || ENV['BUNDLE_ENTERPRISE__CONTRIBSYS__COM']
+  unless Bundler::Settings.new['enterprise.contribsys.com']&.empty? && ENV['BUNDLE_ENTERPRISE__CONTRIBSYS__COM']&.empty?
     source 'https://enterprise.contribsys.com/' do
       gem 'sidekiq-ent'
       gem 'sidekiq-pro'
--- a/Gemfile
+++ b/Gemfile
@@ -173,7 +173,7 @@ end
 group :production do
   # sidekiq enterprise requires a license key to download but is only required in production.
   # for local dev environments, regular sidekiq works fine
-  unless ENV['EXCLUDE_SIDEKIQ_ENTERPRISE'] == 'true'
+  if Bundler::Settings.new['enterprise.contribsys.com'] || ENV['BUNDLE_ENTERPRISE__CONTRIBSYS__COM']
     source 'https://enterprise.contribsys.com/' do
       gem 'sidekiq-ent'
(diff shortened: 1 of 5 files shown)


johnpaulashenfelter:
@jholton can you doublecheck how the contribsys value is put into CI? And i think @LindseySaari might be able to help clean up the docker files.

jholton:
Yeah, there are some references in Docker related files.  Cleaning those up now...

Gemfile
```diff
@@ -173,7 +173,7 @@ end
 group :production do
   # sidekiq enterprise requires a license key to download but is only required in production.
   # for local dev environments, regular sidekiq works fine
-  unless ENV['EXCLUDE_SIDEKIQ_ENTERPRISE'] == 'true'
+  unless Bundler::Settings.new['enterprise.contribsys.com']&.empty? && ENV['BUNDLE_ENTERPRISE__CONTRIBSYS__COM']&.empty?
```
jholton: Rails isn't loaded at this point, so I can't use `present?` or `blank?`.

johnpaulashenfelter: for the ENV, i'd recommend `ENV.fetch('BUNDLE_ENTERPRISE__CONTRIBSYS__COM', false)` -- that way you're ensuring a default

jholton: If someone is running this natively and `ENV['BUNDLE_ENTERPRISE__CONTRIBSYS__COM']` is not set, it will be `nil`. 
However, if using Docker and it's not set, Docker defaults the value to an empty string `''`.
I do not know of a way to have Docker default a variable value to "not set".  Do you?

So, I could do `ENV.fetch('BUNDLE_ENTERPRISE__CONTRIBSYS__COM', '').empty?`

jholton:
> @jholton can you doublecheck how the contribsys value is put into CI? 

It gets set for CI and deployment builds [here](https://github.com/department-of-veterans-affairs/devops/blob/24d138e7f2e1911066262295faf93c117eb7f8aa/ansible/build/roles/vets-api/tasks/main.yml#L115).
[Here](https://github.com/department-of-veterans-affairs/devops/blob/24d138e7f2e1911066262295faf93c117eb7f8aa/ansible/roles/review-instance-configure/tasks/main.yml#L100) for review instances. 
We're covered.

PR #4059, opened 2020-03-26
Cleanup v1 logout api

patrickvinograd:
## Description of change
Removes the SAML-based logout callback from v1 sessions controller, and cleans up extraneous routes and test cases. Also unifies the triggering URL for frontend to be `v1/sessions/slo/new` to match `v0/sessions/slo/new`.

SSOe does not implement SAML single logout, so we never expect a SAML response for a logout. Instead, they have a custom 'PKMS logout' endpoint that will terminate the IAM session and then redirect. back to an endpoint of your choosing via a simple redirect. The `ssoe_slo_callback` method is expected to receive that redirect and do the final additional redirect back to va.gov frontend.

Fixes va.gov-team#7137, but will need corresponding config change in devops repo to update pkms logout endpoint per-environment.

## Testing
Awaiting testing in conjunction with IAM team; require them to set a redirect after their PKMS logout endpoint completes.

#### Unique to this PR

#### Applies to all PRs

### DIFF
--- a/app/controllers/v1/sessions_controller.rb
+++ b/app/controllers/v1/sessions_controller.rb
@@ -22,12 +22,11 @@ module V1
     # For more details see SAML::SettingsService and SAML::URLService
     def new
       type = params[:type]
-      if type == 'slo'
-        type = 'ssoe_slo'
-       raise Common::Exceptions::RoutingError, params[:path] unless REDIRECT_URLS.include?(type)
+      raise Common::Exceptions::RoutingError, params[:path] unless REDIRECT_URLS.include?(type)
 
       StatsD.increment(STATSD_SSO_NEW_KEY,
                        tags: ["context:#{type}", "forceauthn:#{force_authn?}"])
+      type = 'ssoe_slo' if type == 'slo'
       url = url_service.send("#{type}_url")
 
       if type == 'ssoe_slo'
--- a/app/controllers/v1/sessions_controller.rb
+++ b/app/controllers/v1/sessions_controller.rb
@@ -65,6 +65,7 @@ module V1
 
     private
 
+    # rubocop:disable Metrics/CyclomaticComplexity
     def redirect_url(type)
       case type
       when 'signup'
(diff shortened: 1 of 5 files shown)


app/controllers/v1/sessions_controller.rb
```diff
@@ -26,9 +26,10 @@ def new
 
       StatsD.increment(STATSD_SSO_NEW_KEY,
                        tags: ["context:#{type}", "forceauthn:#{force_authn?}"])
+      type = 'ssoe_slo' if type == 'slo'
       url = url_service.send("#{type}_url")
 
-      if %w[slo ssoe_slo].include?(type)
+      if type == 'ssoe_slo'
         Rails.logger.info("LOGOUT of type #{type}", sso_logging_info)
```
ericbuckley: its a little confusing that the StatsD above would log a type of 'slo' and then we'd have a Rails logger message of type 'sloe' here.  Could we move the type reassignment on line 29 above the StatsD call?

patrickvinograd: I wound up reworking this anyway because it was complaining about the unsafe send based on user input.

app/controllers/v1/sessions_controller.rb
```diff
@@ -81,6 +65,29 @@ def metadata
 
     private
 
+    # rubocop:disable Metrics/CyclomaticComplexity
```
patrickvinograd: Yes I know 😬 
But, this method is fundamentally not hard to understand, and is there to avoid the "unsafe send based on user input" that was in place before and that the security scan is now complaining about.

patrickvinograd:
@hinzed1127 The sequence of steps is documented in the ticket. pkmslogout should redirect to the v1/sessions/ssoe_logout endpoint which then redirects to www.va.gov/logout/

hinzed1127:
Cool, that's what I thought, just double checking 👍

app/controllers/v1/sessions_controller.rb
annaswims: line 25 made the `send` safe, but I still prefer this.

PR #4060, opened 2020-03-27
fix Gemfile guard clause for installing Sidekiq Pro/Ent

jholton:
## Description of change
Got bit by using the safe operator on a `nil` value.
```ruby
''&.empty?     # true
 nil&.empty?   # nil, which is falsey
```
I need to check explicitly for `nil` and an empty string.
Why?
Natively, when you have no Sidekiq license config set:
```ruby
Bundler::Settings.new['enterprise.contribsys.com']
# => nil
```
However, if when you start up via Docker, it defaults the value of `BUNDLE_ENTERPRISE__CONTRIBSYS__COM` env variable to ''  and thus:
```ruby
Bundler::Settings.new['enterprise.contribsys.com']
# => ''
```

#### Unique to this PR

- [x] Fix guard clause logic

#### Applies to all PRs

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -173,7 +173,7 @@ end
 group :production do
   # sidekiq enterprise requires a license key to download but is only required in production.
   # for local dev environments, regular sidekiq works fine
-  unless Bundler::Settings.new['enterprise.contribsys.com']&.empty? &&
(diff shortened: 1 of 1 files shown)


jholton:
re: #4057

PR #4062, opened 2020-03-30
Fix mhv account type attribute for v1 login

patrickvinograd:
## Description of change
* Ensures that MHV account status and identifiers are mapped correctly into an identity object from the v1 session controller. SSOe user attribute class was missing the `mhv_account_type` attribute and was incorrectly populating the `sign_in` object. 
* An SSOe assertion has two potential sources for an MHV UUID; this change updates the class to read from either of them and throw a fatal login error in case of a mismatch. 

Fixes va.gov-team#7337

This PR is for merge to master but builds on top of https://github.com/department-of-veterans-affairs/vets-api/pull/4027. I will rebase this branch once that is merged but for now this diff is probably more illustrative:
https://github.com/department-of-veterans-affairs/vets-api/compare/eb-alternative-user-uuid...pv-fix-mhv-account-type-attribute

## Testing
Tested locally and ensured that a user with an MHV account is correctly granted access to health tools and eligible for account upgrade.

#### Applies to all PRs

### DIFF
--- a/lib/saml/user_attributes/ssoe.rb
+++ b/lib/saml/user_attributes/ssoe.rb
@@ -7,8 +7,8 @@ module SAML
     class SSOe
       include SentryLogging
       SERIALIZABLE_ATTRIBUTES = %i[email first_name middle_name last_name zip gender ssn birth_date
-                                   uuid idme_uuid sec_id mhv_icn mhv_correlation_id dslogon_edipi
-                                   loa sign_in multifactor].freeze
+                                   uuid idme_uuid sec_id mhv_icn mhv_correlation_id mhv_account_type
+                                   dslogon_edipi loa sign_in multifactor].freeze
       IDME_GCID_REGEX = /^(?<idme>\w+)\^PN\^200VIDM\^USDVA\^A$/.freeze
 
       attr_reader :attributes, :authn_context, :warnings
@@ -94,6 +94,10 @@ module SAML
         safe_attr('va_eauth_mhvien')
       end
 
+      def mhv_account_type
+        safe_attr('va_eauth_mhvassurance')
+      end
+
       def dslogon_edipi
         safe_attr('va_eauth_dodedipnid')
       end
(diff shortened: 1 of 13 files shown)


lib/saml/user_attributes/ssoe.rb
```diff
@@ -98,25 +130,41 @@ def multifactor
...
+                                  .merge(account_type: account_type)
+      end
+
+      def to_hash
+        Hash[SERIALIZABLE_ATTRIBUTES.map { |k| [k, send(k)] }]
+      end
+
+      # Raise any fatal exceptions due to validation issues
+      def validate!
+        raise SAML::UserAttributeError, 'MHV Identifier mismatch' if mhv_id_mismatch?
```
ericbuckley: @patrickvinograd should we be doing this check with other ids as well.  For example, if the user is LOA3 and they logged in using ID.me credentials, their ID.me uuid should be in both the `va_eauth_uuid` and `va_eauth_gcIds` attributes.

patrickvinograd: I think we can tighten this up over time but I don't think an ID.me mismatch is possible - any LOA3 ID.me account should be getting added to MVI. 

With MHV, we can have the case of an MHV advanced user already having a correlation in MVI, but potentially an MHV basic+ID.me LOA3 identity mapping to the same MVI record and thus having mhv_uuid (from MHV's part of the SAML payload) not match mhv_ien (from MVI). 

Could this potentially happen for DSLogon with the EDIPI parameter? Maybe :/

patrickvinograd: (But, I think it's less of an issue for DSLogon because at this point we should only be getting DSLogon Level 2 authentications. This is really an artifact of the continuance of MHV Basic/Advanced accounts.)

lib/saml/user_attributes/ssoe.rb
```diff
@@ -91,7 +92,11 @@ def mhv_icn
       end
 
       def mhv_correlation_id
-        safe_attr('va_eauth_mhvien')
+        safe_attr('va_eauth_mhvuuid') || safe_attr('va_eauth_mhvien')
```
annaswims: Just double checking that it's `va_eauth_mhvien` IEN and not ICN

patrickvinograd: Yep, mhvien, yet another alias for mhv account id aka mhv uuid aka mhv correlation id.

PR #4064, opened 2020-03-30
Attach ICNs of people on CARMA Submission

kevinmirc:
Fixes: https://github.com/department-of-veterans-affairs/va.gov-team/issues/5894

## Description of change
When a 1010cg is submitted, we must attach the ICN of each person on the form to the payload that's sent to CARMA. These additions loop through each person of the 1010cg and attempt to find their ICN (raises client error when veteran not found).

As part of this work, helper methods were added to the claim model, claim validation has been moved to the controller, and the Form1010cg was re-written from a singleton in an Object-Oriented style.

### Considerations
- Currently `MVI::Service#find_profile` requires a valid `UserIdentity`. I am building a `UserIdentity` in order to use this existing `MVI` method. I'm open to suggestions on how to refactor this or write a different implementation.
- With these changes `POST /v0/caregivers_assistance_claim` will now be making anywhere from `0..6` external requests before sending a response (`2..4` MVI lookups, `1` call to get CARMA auth token, `1` call to submit data to CARMA). I think is still okay for handling the request synchronously, but I can follow up this work with a PR to do the processing in the background if needed.

## Testing

- [x] Specs for changes in CaregiversAssistanceClaimsController
- [x] Specs for new helper methods on CaregiversAssistanceClaim
- [x] Specs for new work performed by Form1010cg::Service

- [x] Attach ICN when found for person on 10-10cg
- [x] Ignore not-found MVI searches for PC, SC_1, SC_2
- [x] Raise Error when ICN not found for Veteran.
  - [x] Error makes it back to the client 4xx

#### Applies to all PRs


app/services/form1010cg/service.rb
```diff
@@ -3,25 +3,85 @@
...
 
+    def assert_veteran_status
+      raise_unprocessable if icn_for('veteran') == NOT_FOUND
+    end
+
+    def build_metadata
+      claim.form_subjects.each_with_object({}) do |namespace, acc|
+        icn = icn_for(namespace)
+        acc[namespace.to_sym] = {
+          icn: icn == 'NOT_FOUND' ? nil : icn
```
johnpaulashenfelter, line 36: Maybe reuse your `NOT_FOUND` constant? Or just use a string here and remove the constant everywhere else

app/services/form1010cg/service.rb
```diff
@@ -3,25 +3,85 @@
...
         submitted_at: carma_submission.submitted_at
       )
     end
 
+    def assert_veteran_status
+      raise_unprocessable if icn_for('veteran') == NOT_FOUND
+    end
+
+    def build_metadata
+      claim.form_subjects.each_with_object({}) do |namespace, acc|
```
johnpaulashenfelter, line 33: Not sure what acc means here -- and is there a word other than "namespace" that describes what you're iterating through? Even if its a `form_subject` it might be clearer.

spec/controllers/v0/caregivers_assistance_claims_controller_spec.rb
```diff
@@ -8,53 +8,11 @@
...
-          'title' => 'Missing parameter',
-          'detail' => 'The required parameter "form", is missing',
-          'code' => '108',
-          'status' => '400'
-        }
-      )
-    end
-
     context 'when Flipper :allow_online_10_10cg_submissions is' do
       context 'disabled' do
```
johnpaulashenfelter, line 12: Nice use of context -- it will read really well with `rspec -f doc`

spec/controllers/v0/caregivers_assistance_claims_controller_spec.rb
```diff
@@ -64,38 +22,114 @@
...
+              expect(res_body['errors']).to be_present
+              expect(res_body['errors'].size).to eq(2)
+              expect(res_body['errors'][0]['title']).to include("did not contain a required property of 'veteran'")
+              expect(res_body['errors'][0]['code']).to eq('100')
+              expect(res_body['errors'][0]['status']).to eq('422')
+              expect(res_body['errors'][1]['title']).to include(
+                "did not contain a required property of 'primaryCaregiver'"
+              )
+              expect(res_body['errors'][1]['code']).to eq('100')
+              expect(res_body['errors'][1]['status']).to eq('422')
```
johnpaulashenfelter, line 99: We've now got the super_diff gem in the app -- it may or may not let you condense this a bit if you're interested in taking a look.

spec/models/saved_claim/caregivers_assistance_claim_spec.rb
```diff
@@ -3,21 +3,171 @@
...
+      }
+
+      # Required properties for specific form_subjects
+      if form_subject == :primaryCaregiver
+        data['vetRelationship'] = %w[Father Mother Son Daughter].sample
+        data['medicaidEnrolled'] = [true, false].sample
+        data['medicareEnrolled'] = [true, false].sample
+        data['tricareEnrolled'] = [true, false].sample
+        data['champvaEnrolled'] = [true, false].sample
+      end
```
johnpaulashenfelter, line 31: Sample runs the risk of breaking reproducibility in a spec here. If the value really doesn't matter just set it to a default.

spec/models/saved_claim/caregivers_assistance_claim_spec.rb
```diff
@@ -3,21 +3,171 @@
...
+
+      data['plannedClinic'] = %w[740 568A4 550].sample if form_subject == :veteran
```
johnpaulashenfelter, line 33: Opps -- ditto the sample comment.

app/services/form1010cg/service.rb
kevinmirc: Oops. I used `acc` (accumulator) out of habit (it's a common practice for this same function in javascript `Array.prototype.reduce`).

 I think `metadata` would make more sense here. I will also use `form_subject` instead of `namespace`. Thanks!

spec/controllers/v0/caregivers_assistance_claims_controller_spec.rb
kevinmirc: This is tricky because I have to use an RSpec `include` matcher for this objects `title` property (because of the uuid present in the string).

I'm not seeing an example or spec in super_diff that allows you to use a Rspec `include` inside an `eq` matcher like so:
```rb
expect(res_body['errors'][0]).to eq(
  'title' => include("did not contain a required property of 'veteran'"),
  'code' => '100',
  'status' => '422'
 )
```

app/services/form1010cg/service.rb
```diff
@@ -3,25 +3,85 @@
...
+
+      attributes = {
+        first_name: data['fullName']['first'],
+        middle_name: data['fullName']['middle'],
+        last_name: data['fullName']['last'],
+        birth_date: data['dateOfBirth'],
+        gender: data['gender'] == 'U' ? nil : data['gender'],
+        ssn: data['ssnOrTin'],
+        email: data['email'] || 'no-email@example.com',
+        uuid: SecureRandom.uuid,
```
annaswims, line 77: What are the consequences of sending a uuid to MVI that is not a user's uuid. Might MVI save this uuid and corrupt the user's record?

app/services/form1010cg/service.rb
```diff
@@ -3,25 +3,85 @@
...
+        loa: {
+          current: LOA::THREE,
```
annaswims, line 79: This seems like it could violate our ATO.

app/services/form1010cg/service.rb
kevinmirc: @annaswims The uuid never gets passed to the API request to MVI. I'm building a `UserIdentity` since `MVI::Service#find_profile` requires a valid `UserIdentity` and these are the requirements of a valid user identity.

The request to MVI will only send `first_name`, `middle_name`, `last_name`, `birth_date`, `ssn`, and `gender`.
```rb
    # lib/mvi/service.rb:198
    def message_user_attributes(user)
      given_names = [user.first_name]
      given_names.push user.middle_name unless user.middle_name.nil?
      profile = {
        given_names: given_names,
        last_name: user.last_name,
        birth_date: user.birth_date,
        ssn: user.ssn,
        gender: user.gender
      }
      MVI::Messages::FindProfileMessage.new(profile).to_xml
    end
```

A better alternative would be to create an `MVI::Service#find_profile_with_attributes` that would only require `first_name`, `middle_name`, `last_name`, `birth_date`, `ssn`, and `gender` vs and entire `UserIdentity`. This would avoid confusion and (more importantly) eliminate the risk of a future developer using this `UserIdentity` for other purposes (like a `put` or `patch`).

Thoughts?

jholton: HCA also has code that queries MVI to look up a user's ICN based on traits.
Check out [`MVI::AttrService`](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/mvi/attr_service.rb).
HCA [feeds it ](https://github.com/department-of-veterans-affairs/vets-api/blob/f6fcced922e40034eb4fb230c2f65a8876bb49c0/app/controllers/v0/health_care_applications_controller.rb#L37) an [`HCA::UserAttribute` object](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/hca/user_attributes.rb).
The `HCA::UserAttribute` object exists for this exact purpose: serves an object that "duck types" with the necessary attributes to facilitate an MVI traits search.
It was created in the `HCA` namespace because that was the first use for it.
Now that we have multiple use cases for it, we should probably move it to another namespace (i.e. `MVI`) and re-use it here so we keep things "DRY", and hopefully reduce the likelihood of running into confusing scenarios like this again.

kevinmirc: Looks like the `MVI` lib already has a model for requiring the proper attributes in a search: `MVI::Messages::FindProfileMessage` (in _/lib/mvi/messages/find_profile_message.rb_). It even contains the validation required.

@jholton, @annaswims, What are your thoughts on this refactor:

Implement a `MVI::Service#find_profile` and a `MVI::Service#find_profile_by_user_identity`. 

- `#find_profile` requires a valid `MVI::Messages::FindProfileMessage`, `MVI::Messages::FindProfileMessageIcn`, or `MVI::Messages::FindProfileMessageEdipi`. These models already exist. Any consumers of the service is responsible for building a valid message for the `#find_profile` call.

- `#find_profile_by_user_identity` would require a valid `UserIdentity` and converts it to a valid `MVI::Messages::FindProfileMessage` | `MVI::Messages::FindProfileMessageIcn` | `MVI::Messages::FindProfileMessageEdipi`, then calls the new `#find_profile` method.

kevinmirc: I will do a followup PR to make this change. Here's the ticket: https://github.com/department-of-veterans-affairs/va.gov-team/issues/9083

kevinmirc: @jholton I was working on this change and realized that the `MVI::AttrService` is tightly coupled with HCA, and having CARMA use it wouldn't make sense. It has it's own Faraday client and configuration that specifies HCA (https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/mvi/attr_configuration.rb).

I'm wondering if I should (1) merge the `MVI::AttrService`'s features into `MVI::Service` so all other consumers can use the same generic client, or (2) follow suit and make a CARMA MVI service with its own Faraday client and config.

I'm leaning towards Number 1 but I don't what the reason reasons for `MVI::AttrService` having its own client other than having that service name "**HCAMVI**" show up in logs and metrics separately from "**MVI**". Is this the behavior we want?

PR #4066, opened 2020-03-30
Appeals API - Remove Pre-Pivot Code/Routes and Update Documentation

edmangimelli:
Resolves https://vasdvp.atlassian.net/browse/DASH-159

Resolves https://vasdvp.atlassian.net/browse/DASH-167

Adds the following headers to _HLR Create_ endpoint:
```
X-VA-First-Name *
X-VA-Middle-Initial
X-VA-Last-Name *
X-VA-Birth-Date *
X-VA-File-Number
X-VA-Service-Number
X-VA-Insurance-Policy-Number
```
(`*` = required)

Removes `Claimant` field (specifying a non-veteran claimant is not part of MVP).

Makes veteran's address a required field.

Removes unused Caseflow endpoints (from lib, controllers, tests, routes, etc).

Minor description / formatting fixes.

Fixes and adds to tests for changes.

Double check paper form for fields we're missing.


bastosmichael:
Are we forgoing Swagger Blocks for Appeals, feels like we should have used the DSL to maintain parity with Forms, Benefits Intake and Benefits Claims API's?

edmangimelli:
> Are we forgoing Swagger Blocks for Appeals, feels like we should have used the DSL to maintain parity with Forms, Benefits Intake and Benefits Claims API's?

Housing the documentation in vets-api is a newish decision, but in January (?) we brought it in and decided to skip swagger blocks for the sake of simplicity.

Like you suggested via Zoom, let's revisit this next time we make some documentation changes. As much as I hate writing yaml, I don't immediately see the benefit of rewriting 1100+ lines in swagger blocks. I was actual able to reuse / compose quite a bit with plain old Open API.

I definitely _do_ feel the pain of using both JSON Schema and Open API --if swagger blocks smoothes that over, I'm all in.

modules/appeals_api/config/routes.rb
```diff
@@ -5,14 +5,7 @@
...
 
   namespace :v0, defaults: { format: 'json' } do
-    resources :appeals, only: [:index] do
-      collection do
-        get 'higher_level_reviews/:uuid', to: 'appeals#show_higher_level_review'
-        post 'higher_level_reviews', to: 'appeals#create_higher_level_review'
-        get 'intake_statuses/:intake_id', to: 'appeals#show_intake_status'
-      end
-    end
-
```
edmangimelli, line 15: @camerontesterman I think these were put into v0 by mistake --does that seem right?

edmangimelli: (or they're no longer relevant)

camerontesterman: yeah they were added by mistake, should've been removed.

edmangimelli: thanks!

PR #4069, opened 2020-03-31
Rails 6 upgrade

erluti:
https://github.com/department-of-veterans-affairs/va.gov-team/issues/7241

## Description of change
* Resolve gem and test issues related to updating Rails version to 6.0.2.
* Resolved deprecation warning for using `update` instead of `update_attributes`.
* Addressed deprecation warning related to `ActiveSupport::OrderedOptions` not allow non-symbol keys (starting in 6.1).  I was able to fix this in most places, except we pass a `ActiveSupport::OrderedOptions` into `Redis.new` which is accessing values with quotes.

## Notes
[apivore](https://github.com/department-of-veterans-affairs/apivore) and [active_model_serializers](https://github.com/department-of-veterans-affairs/active_model_serializers) were using versions that weren't marked compatible to Rails 6.0.2, so they were forked to remove that restriction.
Related tickets for changing these dependencies:
* [replace apivore](https://github.com/department-of-veterans-affairs/va.gov-team/issues/7572)
* [replace active_model_serializers](https://github.com/department-of-veterans-affairs/va.gov-team/issues/465)


spec/fixtures/education_benefits_claims/1990/kitchen_sink.spl
```diff
@@ -114,7 +114,7 @@ Active Duty    From Active Duty
 2013-04-22     2013-06-14         Army Reserve
                          Service Status:   ACTIVE DUTY
                          Involuntary Call: yes
-Benefits to Apply to: chapter33, chapter1606,
+                         Benefits to Apply to: chapter33, chapter1606,
```
erluti: this is what the template should be generating, but before Rails 6, this wan't including these leading whitespaces. https://github.com/department-of-veterans-affairs/vets-api/blob/15c3a5cd53a6b1b7c4a28e8c5653cf3c67737f47/app/workers/education_form/templates/tours_of_duty.erb

spec/request/attachments_request_spec.rb
```diff
@@ -56,7 +56,7 @@
 
         expect(response).to be_successful
         expect(response.headers['Content-Disposition'])
-          .to eq('attachment; filename="noise300x200.png"')
+          .to eq("attachment; filename=\"noise300x200.png\"; filename*=UTF-8''noise300x200.png")
```
erluti: this is the new default value for ['Content-Disposition'] in Rails 6

omgitsbillryan:
What kind of changes did you make to `apivore` & `active_model_serializers`?

Using forks seems like adding on tech debt. My gut tells me that we probably want to remove our dependence on these two gems before upgrading to rails 6.

erluti:
> What kind of changes did you make to `apivore` & `active_model_serializers`?
> 
> Using forks seems like adding on tech debt. My gut tells me that we probably want to remove our dependence on these two gems before upgrading to rails 6.

The only change is removing the Rails version restriction and updated a call to prevent deprecation warnings in `active_model_serializer`.

@johnpaulashenfelter recommended this approach because pulling those out is a bigger lift.  The plan is to follow up by removing those.

config/application.rb
```diff
@@ -5,36 +5,38 @@
...
   class Application < Rails::Application
-    # Settings in config/environments/* take precedence over those specified here.
-    # Application configuration should go into files in config/initializers
-    # -- all .rb files in that directory are automatically loaded.
-
-    # Set Time.zone default to the specified zone and make Active Record auto-convert to this zone.
-    # Run "rake -D time" for a list of tasks for finding time zone names. Default is UTC.
-    # config.time_zone = 'Central Time (US & Canada)'
+    # Initialize configuration defaults for originally generated Rails version.
+    config.load_defaults 5.2
```
johnpaulashenfelter, line 29: I had to ask/lookup the info about this: https://prathamesh.tech/2019/08/22/new-framework-defaults-in-rails/

app/models/account.rb
```diff
@@ -28,8 +28,8 @@ class Account < ApplicationRecord
   # Required for configuring mixed in ActiveRecordCacheAside module.
   # Redis settings for ttl and namespacing reside in config/redis.yml
   #
-  redis REDIS_CONFIG['user_account_details']['namespace']
-  redis_ttl REDIS_CONFIG['user_account_details']['each_ttl']
+  redis REDIS_CONFIG[:user_account_details][:namespace]
+  redis_ttl REDIS_CONFIG[:user_account_details][:each_ttl]
```
omgitsbillryan, line 32: Same comment here WRT moving this into a kind of "Rails 6 preparation PR".

modules/appeals_api/appeals_api.gemspec
```diff
@@ -19,7 +19,7 @@ Gem::Specification.new do |s|
   s.files = Dir['{app,config,db,lib}/**/*', 'MIT-LICENSE', 'Rakefile', 'README.rdoc']
   s.test_files = Dir['spec/**/*']
 
-  s.add_dependency 'rails', '~> 5.2.3'
+  s.add_dependency 'rails', '>= 5.2.3'
```
jholton, line 22: I _feel_ like all these module `rails` depencies should just specify `'~> 6.0.2'` since they're not going to be able to use any deprecated/removed Rails 5 features anyway.  Am I being nitpicky? (Or am I wrong?)

config/spring.rb
```diff
@@ -1,5 +1,11 @@
 # frozen_string_literal: true
 
-Spring.watch 'config/application.yml'
-Spring.watch 'config/settings.yml'
-Spring.watch 'config/settings.local.yml'
+Spring.watch(
+  '.ruby-version',
+  '.rbenv-vars',
+  'tmp/restart.txt',
+  'tmp/caching-dev.txt',
```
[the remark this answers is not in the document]
erluti: these were added with the `rails update` command

modules/appeals_api/appeals_api.gemspec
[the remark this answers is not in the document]
erluti: This requirement was stopping the bundle update from working, so I started to change it to `'~> 6.0.2'` but then realized that whoever came along next would have the same issue.  I think I'm more on the side of stripping the version completely since they are part of vets-api.  But I also see the logic in making it match what was in the main bundle.

app/models/account.rb
erluti: These changes are because of deprecation warnings.  So if we were to split them out, would it make more sense to be _after_ the Rails 6 change?

config/application.rb
```diff
@@ -5,36 +5,38 @@
...
 require 'action_controller/railtie'
 require 'action_mailer/railtie'
+# require "action_mailbox/engine"
+# require "action_text/engine"
 # require "action_view/railtie"
+# require "action_cable/engine"
 # require "sprockets/railtie"
 require_relative '../lib/http_method_not_allowed'
 require_relative '../lib/statsd_middleware'
+require 'rails/test_unit/railtie'
```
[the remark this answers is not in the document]
erluti: these came from the rails update command.  they are newly supported in rails 6 and should be uncommented to use them.

app/models/account.rb
omgitsbillryan: Oh interesting. Ya, that could work too. 

Not a big deal either way, maybe just a nice-to-have WRT smaller PR's.

modules/appeals_api/appeals_api.gemspec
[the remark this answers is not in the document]
johnpaulashenfelter: I'm not sure there's a need for *any* rails dependency is there? i mean clearly this is only going to be used in this rails app, but its a bit redundant and an extra step at every upgrade. 

I would guess at one point these said `~> 4.2` back in the beginning :)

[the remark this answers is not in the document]
jholton: >  I think I'm more on the side of stripping the version completely since they are part of vets-api.

> I'm not sure there's a need for any rails dependency is there?

even better idea.

PR #4070, opened 2020-04-01
VAOS: Fix undefined provider in appointments serializer

kreek:
## Description of change
Fixes an issue where the appointment serializer was trying to call `.map` on a nil provider (`vvs_appointment[:providers][:provider]`). Using the example response that caused the error also revealed some fields we didn't have in our json schema. Note the fields aren't present in all items in the response so I set the spec check to strict: false which disables the default behavior of assuming all fields are required even if they're not required in the schema.

## Testing
Unit tests

No longer throws an undefined method error.

#### Applies to all PRs


modules/vaos/spec/request/appointments_request_spec.rb
```diff
@@ -171,6 +171,16 @@
           end
         end
       end
+
+      context 'with a response that threw a map undefined error' do
+        it 'no longer throws the error' do
```
jholton, line 176: Just a wording nitpick.  Instead of addressing the bug that prompted this work, we can describe what is being tested.  Something like
```ruby
  context 'with an optional attribute(s) missing' do
    it 'successfully parses the response' do
```

PR #4087, opened 2020-04-02
Removing unecessary data after Error or Establishment

bastosmichael:
## Description of change
Fixes Ticket https://vasdvp.atlassian.net/secure/RapidBoard.jspa?rapidView=140&projectKey=DASH&modal=detail&selectedIssue=DASH-154&quickFilter=393&assignee=5e386d8f6061160ca042d953

## Testing
Update to Rspec test

- Removes PII and PHI data when successfully established or errored

#### Applies to all PRs


bastosmichael:
Interesting perspective, so according to our current production customer on Claims v0 Tim once they try to auto establish a claim and it fails they don't attempt to fix the error and resubmit. If they did they would have to change the payload somehow and that would change the md5 hash which would then create a new auto established claim all together. I can see the benefit of maintaining errored values for testing later on but in most cases resubmission would entail a new model anyhow and any reattempt at running it against EVSS would mean a change happening in EVSS itself and not on the payload sent to us. I can remove the errored nulls at least for now.

PR #4088, opened 2020-04-02
Appeals API - HLR Submission Persistence

edmangimelli:
resolves https://vasdvp.atlassian.net/browse/DASH-153

creates a model for temporarily storing HLR Create submissions (stores headers, and json body encrypted)


modules/appeals_api/db/migrate/20200402145412_hlr_submission.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class HlrSubmission < ActiveRecord::Migration[5.2]
+  def change
+    enable_extension 'uuid-ossp'
+    enable_extension 'pgcrypto'
+
+    create_table :hlr_submissions, id: :uuid do |t|
```
charleystran, line 8: this table should be called appeals_api_hlr_submissions or better yet appeals_api_higher_level_reviews

modules/appeals_api/db/migrate/20200402145412_hlr_submission.rb
```diff
@@ -0,0 +1,19 @@
...
+      t.integer :status, default: 0
+      t.string :encrypted_json
```
bastosmichael, line 10: This needs to be form_data

modules/appeals_api/db/migrate/20200402145412_hlr_submission.rb
```diff
@@ -0,0 +1,19 @@
...
+      t.string :encrypted_json_iv
+      t.string :encrypted_headers
```
bastosmichael, line 12: This needs to be auth_headers

modules/appeals_api/db/migrate/20200402145412_hlr_submission.rb
```diff
@@ -0,0 +1,19 @@
...
+      t.string :encrypted_headers_iv
+      t.string :encrypted_pdf
```
bastosmichael, line 14: this needs to be file_data

edmangimelli:
>This needs to be form_data
>This needs to be auth_headers
>This needs to be file_data

Just out of curiosity, are these have-to's for implementation reasons, or for sharing conventions with Claims?

modules/appeals_api/db/migrate/20200402145412_hlr_submission.rb
edmangimelli: i like it (hlr -> higher_level_review). but here's to hoping we'll never need to make a join table --i've run into problems with long table names before

edmangimelli:
> Can we rename the table to just higher_level_reviews? That would put it in line with names of the claims_api tables

to me it seems likely that we'll want an actual HigherLevelReviews model, though. 
i'd rather not have one model for both saving/encrypting user submissions (which we want to clean up regularly) and representing HigherLevelReviews with whatever business logic we want to attach to them. I would opt for two smaller models.
thoughts?

bastosmichael:
I think one persisted model per appeals type is perfectly fine, also @charleystran I think the reason I wanted to persist the PDF itself was for debugging purposes, what happens if an issue happens with a bad pdf and we can't reproduce it because we don't have the original pdf that was generated to see for ourselves? Your call on this since you're leading this charge but something to consider on the issue of PDF persistence.

charleystran:
@bastosmichael I think persisting the PDF should be its own ticket if we want to do it because we need to also create jobs to cleanup old pdf's.

PR #4094, opened 2020-04-03
Improve configuration and messaging for dealing with Sidekiq Enterprise in Gemfile

johnpaulashenfelter:
There has been a slow but steady trickle of improvements to how we handle Sidekiq Enterprise, which requires authentication, in our Gemfile. Recent improvements have helped remove some of the configuration burden (#4060). This PR allows for both local and global bundler config files to be supported and adds an informational message to the bundler output when the Sidekiq Enterprise credentials are missing.

## Testing

Used global, app-level, and ENV var to test the error and happy paths for generating the Gemfile.


Gemfile
```diff
@@ -174,12 +174,16 @@ end
...
+  unless (Bundler::Settings.new(Bundler.app_config_path)['enterprise.contribsys.com'].nil? ||
+          Bundler::Settings.new(Bundler.app_config_path)['enterprise.contribsys.com']&.empty?) &&
          ENV.fetch('BUNDLE_ENTERPRISE__CONTRIBSYS__COM', '').empty?
     source 'https://enterprise.contribsys.com/' do
       gem 'sidekiq-ent'
       gem 'sidekiq-pro'
     end
+  else
+    puts "No credentials found to install Sidekiq Enterprise. This is fine for local development but you may not check in this Gemfile.lock."
+    sleep 3
```
jholton, line 186: Is this just so the message lingers in front of the user?

Gemfile
```diff
@@ -174,12 +174,16 @@ end
...
-          Bundler::Settings.new['enterprise.contribsys.com']&.empty?) &&
+  unless (Bundler::Settings.new(Bundler.app_config_path)['enterprise.contribsys.com'].nil? ||
+          Bundler::Settings.new(Bundler.app_config_path)['enterprise.contribsys.com']&.empty?) &&
          ENV.fetch('BUNDLE_ENTERPRISE__CONTRIBSYS__COM', '').empty?
     source 'https://enterprise.contribsys.com/' do
       gem 'sidekiq-ent'
       gem 'sidekiq-pro'
     end
+  else
+    puts "No credentials found to install Sidekiq Enterprise. This is fine for local development but you may not check in this Gemfile.lock."
```
jholton, line 185: Just a little more verbosity:
```suggestion
    puts "No credentials found to install Sidekiq Enterprise. This is fine for local development but you may not check in this Gemfile.lock with the Sidekiq gems removed"
```

Gemfile
johnpaulashenfelter: Yeah -- otherwise its at the TOP of the list of gems and scrolls off screen. Don't love it, can dump it all together.

Gemfile
johnpaulashenfelter: Took your advice

Gemfile
```diff
@@ -174,9 +174,12 @@ end
 group :production do
   # sidekiq enterprise requires a license key to download but is only required in production.
   # for local dev environments, regular sidekiq works fine
-  unless (Bundler::Settings.new['enterprise.contribsys.com'].nil? ||
-          Bundler::Settings.new['enterprise.contribsys.com']&.empty?) &&
-         ENV.fetch('BUNDLE_ENTERPRISE__CONTRIBSYS__COM', '').empty?
+  if (Bundler::Settings.new(Bundler.app_config_path)['enterprise.contribsys.com'].nil? ||
+      Bundler::Settings.new(Bundler.app_config_path)['enterprise.contribsys.com']&.empty?) &&
+     ENV.fetch('BUNDLE_ENTERPRISE__CONTRIBSYS__COM', '').empty?
+    puts 'No credentials found to install Sidekiq Enterprise. This is fine for local development but you may not check in this Gemfile.lock with any Sidekiq gems removed. The README file in this directory contains more information.'
```
jholton, line 180: We can add some color.  This will output the text in yellow:
```suggestion
    Bundler.ui.warn 'No credentials found to install Sidekiq Enterprise. This is fine for local development but you may not check in this Gemfile.lock with any Sidekiq gems removed. The README file in this directory contains more information.'
```

Gemfile
```diff
@@ -174,9 +174,12 @@ end
...
+    sleep 3
```
jholton, line 181: I'm leaning toward just removing the `sleep`.  Adding time to startup does not improve the developer experience.  I think outputting the message in yellow is a decent alternative for grabbing attention.

johnpaulashenfelter: I'm ok with that -- though i'm think error more than warn.

FWIW sleep only happens if you don't have the configuration but good pickup on the `Builder.ui`

johnpaulashenfelter: Ran it -- works great.

```
$ bundle
No credentials found to install Sidekiq Enterprise. This is fine for local development but you may not check in this Gemfile.lock with any Sidekiq gems removed. The README file in this directory contains more information.
Fetching gem metadata from https://rubygems.org/.........
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies.....
Using rake 13.0.1$ bundle
No credentials found to install Sidekiq Enterprise. This is fine for local development but you may not check in this Gemfile.lock with any Sidekiq gems removed. The README file in this directory contains more information.
Fetching gem metadata from https://rubygems.org/.........
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies.....
Using rake 13.0.1
```

jholton: I just tried with `Bundler.ui.error` and it outputs the message twice for some reason.

PR #2888, opened 2019-03-18
Improve PII Sanitizer, don't sanitize certain fields

saneshark:
## Description of change
ICN and EDIPI while PII, are just primary keys, and they're needed for correlating our logs with downstream partners. These fields cannot really be used to cause problems since you can't use an ICN or EDIPI to open a credit card for example. They're tremendously useful for debugging purposes.

## Testing done
specs

## Testing planned
none

#### Unique to this PR
Introduced an exceptions list. Since state gets triggered by regular expression, we will also check that whatever the key is, it is not in the exceptions list. There is a spec that tests that "state" is correctly filtered whereas relaystate, relayState, relay_state are not.

#### Applies to all PRs

- [ ] Appropriate logging
- [ ] Swagger docs have been updated, if applicable
- [ ] Provide link to originating GitHub issue, or connected to it via ZenHub
- [ ] Does not contain any sensitive information (i.e. PII/credentials/internal URLs/etc., in logging, hardcoded, or in specs)
- [ ] Provide which alerts would indicate a problem with this functionality (if applicable)

PR #4095, opened 2020-04-03
Filter eauth PII from sentry

annaswims:
## Description of change
Filter PII attributes related to sso before they are sent to sentry

Currently, all fields starting with `va_eauth_` are being filtered in sentry, but this change will allow us to see the non-pii fields and filter before the pii is sent to sentry.

## Testing
specs
## Acceptance Criteria 
@patrickvinograd or @ericbuckley, please take a look at the spec to confirm that the remaining fields are not PII

#### Unique to this PR
After this PR is merged remove `va_eauth_` from `
Additional field names to match against when scrubbing data. Separate multiple entries with a newline
va_eauth_
`  http://sentry.vfs.va.gov/settings/vets-gov/projects/platform-api-staging/
#### Applies to all PRs


lib/saml/user.rb
```diff
@@ -29,7 +29,7 @@ class User
     def initialize(saml_response)
       @saml_response = saml_response
       @saml_attributes = saml_response.attributes
-
+puts saml_attributes&.to_h
```
jholton, line 32: debug line?

ericbuckley:
@annaswims what about `va_eauth_uid` and `va_eauth_secid` should we be filtering those at as well?

lib/saml/user.rb
annaswims: Guilty.

annaswims:
va_eauth_{first|middle|last}name are being filtered already because the keys include `{first|middle|last}name`

annaswims:
@ericbuckley are va_eauth_uid and va_eauth_secid considered PII or are they just primary keys that are only useful if you have access to our database?

jholton:
> @ericbuckley are va_eauth_uid and va_eauth_secid considered PII or are they just primary keys that are only useful if you have access to our database?

This ☝️ specific question needs to be answered, but _if_ they are "just primary keys", I think the rationale presented in [this PR](https://github.com/department-of-veterans-affairs/vets-api/pull/2888#issuecomment-474173431) applies.
(Unfortunately, I don't see "SEC ID" mentioned in the doc that is linked to in the PR comment)

annaswims:
I know we've talked about id.me uuid and allow it to be logged and I'd consider sec_id to be the corresponding id from SSOe.  Only useful for identification purposes if someone has access to IAM data.

ericbuckley:
> @ericbuckley are va_eauth_uid and va_eauth_secid considered PII or are they just primary keys that are only useful if you have access to our database?

@annaswims I believe the va_eauth_secid is a unique identifier for the users MVI record, the va_eauth_uid value is a little more complicated, as depending on what credentials the user used to authenticate with this will change.  When logging in with ID.me credentials, it would be the users ID.me UUID, however when logging in with other credentials it could be something else.  Its probably safest to filter that value, however @patrickvinograd could probably provide more context.

PR #4100, opened 2020-04-03
Update rubocop gem and configuration

johnpaulashenfelter:
Updates rubocop to 0.81 and rubocop-rails to 2.5.0. Also adds configuration (defaults to false)
for the new cops in v0.80 and v0.81.


.rubocop.yml
```diff
@@ -329,3 +329,18 @@ RSpec/SubjectStub:
...
+  Enabled: false
+
+Style/HashEachMethods:
+  Enabled: false
+
+Style/HashTransformKeys:
+  Enabled: false
+
+Style/HashTransformValues:
+  Enabled: false
```
omgitsbillryan, line 346: Thoughts on enabling these cops but adding the exceptions to the `.rubocop_todo.yml` so that they have a better chance of being looked at some day?

johnpaulashenfelter: @omgitsbillryan honestly i didn't read enough about them to know what they're for yet. Intro'd in v0.80 and v0.81 in the new way rubocop is handling additions. Probably worth actively deciding our thoughts on each one so we don't have todos we don't want to ever do :)

johnpaulashenfelter: Mostly this was to get rid of the extra output

PR #4102, opened 2020-04-03
Remove dependabot

johnpaulashenfelter:
We've been trying dependabot for a few months. As of the latest round
of dependabot PRs, all of them include `Gemfile.lock` files that removed
the commercial Sidekiq gems. We can't merge these, so in this case it took
reviewing/closing/creating 15 new PRs which created a fair bit of work.

I checked our configuration on GitHub and we have enterprise.contribsys.com
configured, which must be why it *used* to work. No clear reason why it
no longer works.

There's no clear way to disable the configuration without removing it
altogether based on all the docs I've been able to find. The second best
option is removing it.


annaswims:
the root cause is https://github.com/department-of-veterans-affairs/vets-api/pull/4057 the sidekiq password is set in 
https://app.dependabot.com/accounts/department-of-veterans-affairs/config-variables

so, perhaps we can just fix the check?

LindseySaari:
Have we looked into Anna's comment at all? Also might be worth looking into alternatives such as: https://github.com/lpender/bummr

johnpaulashenfelter:
@LindseySaari @annaswims #4057 solves a ton of ongoing issues we had with folks on Docker. @jholton and I spent a lot of time beating that into submission and I'm not thrilled about refactoring it again for dependabot. 

It's not that hard to `bundle outdated --only-explicit` and update a gem. I'll call that out explicitly in the utility dev docs since its relevant

johnpaulashenfelter:
https://github.com/department-of-veterans-affairs/va.gov-team/pull/9491

annaswims:
there were two dependabots.  One configured through github  and one configured through dependabot. So, dependabot will likely still make security PR's.

PR #4104, opened 2020-04-06
Original claims bug fixes

SMLuthi:
## Description of change

Add extra gating to ITF controller that verifies the user has a valid `birls_id`. This is to account for an extreme, but possible, edge case.

Fix multiple instances of broken raising `ValidationErrors`. These errors were expecting a specific object but the instances in questions did not contain what is required... so it was throwing a 500 error instead. These have been replaced with more generic `UnprocessableEntity` errors that are more flexible.

The add person call in `in_progress_form_controller` will now raise an error if the add_person call responds with anything but `ok`.

The `clear_cache` logic has been removed due to a possible race condition when calling in_progress_form and, subsequently, the itf controller. It was verified that the proxy add was being made successfully but the user object did not reflect this change. Its hard to debug the exact reason but a potential issue is that the subsequent call to `find_profile` did not contain the new ids. Therefore, the call in the mvi model has been changed to, instead, assign the new ids to the profile directly and recache the profile afterwards. As a bonus, this will avoid making, yet, another call to MVI.

## Testing

spec tests
Will be tested in staging with staging user accounts missing the ids

#### Unique to this PR

- [x] Add extra gating to ITF controller
- [x] Fix broken ValidationErrors errors
- [x] Replace cache destruction with recaching and updating the profile ids directly.
- [x] Add throwing an error if the add_person call to mvi fails in in_progress_forms controller

#### Applies to all PRs


jholton:
> a potential issue is that the subsequent call to `find_profile` did not contain the new ids.

Do they appear in results from `find_profile` later?
Is the idea/theory that there is some kind of lag between submitting the user changes to MVI and the MVI service subsequently returning the changes in `find_profile` requests?

app/models/mvi.rb
```diff
@@ -129,9 +129,12 @@ def mvi_add_person
 
   private
 
-  def clear_cache
-    Mvi.delete(user.uuid)
-    @mvi_response = nil
+  def add_ids(response)
+    # set new ids in the profile and recache the response
+    profile.birls_id = response.mvi_codes[:birls_id] if response.mvi_codes[:birls_id].present?
+    profile.participant_id = response.mvi_codes[:participant_id] if response.mvi_codes[:participant_id].present?
```
jholton, line 135: Are `response.mvi_codes[:birls_id]` and `response.mvi_codes[:participant_id]` `nil` when not `present?`
Looking at the parser below, I think they'll always be `nil` or a stingified digit.
So here's a nitpicky syntax suggestion you can feel free to push back on:
```suggestion
    profile.birls_id = response.mvi_codes[:birls_id].presence
    profile.participant_id = response.mvi_codes[:participant_id].presence
```

lib/mvi/responses/add_parser.rb
```diff
@@ -56,11 +56,33 @@ def parse
...
+      def parse_ids(attributes)
+        codes = { other: [] }
+        attributes.each do |attribute|
+          if attribute[:code].include? 'BRLS'
+            codes[:birls_id] = sanitize_ids(attribute[:code])
+          elsif attribute[:code].include? 'CORP'
+            codes[:participant_id] = sanitize_ids(attribute[:code])
+          else
+            codes[:other].append(attribute)
+          end
```
jholton, line 74: Here's another nitpicky syntax suggestion you can feel free to push back on:
```suggestion
          case attribute[:code]
          when /BRLS/
            codes[:birls_id] = sanitize_ids(attribute[:code])
          when /CORP/
            codes[:participant_id] = sanitize_ids(attribute[:code])
          else
            codes[:other].append(attribute)
          end
```

spec/models/mvi_spec.rb
```diff
@@ -213,15 +214,24 @@
...
+      MVI::Responses::AddPersonResponse.new(
+        status: 'OK',
+        mvi_codes: {
+          birls_id: '1234567890',
+          participant_id: '0987654321'
+        }
+      )
+    end
 
     it 'clears user from cache and returns nil' do
```
jholton, line 229: Looks like this spec needs to be renamed.

lib/mvi/service.rb
```diff
@@ -110,7 +110,7 @@ def measure_info(user_identity)
     end
 
     def get_mvi_error_key(e)
-      error_name = e.body&.first&.[](:displayName)
+      error_name = e.body[:other]&.first&.[](:displayName)
```
jholton, line 113: Just looking at the diff....it looks like before `e.body` was not guaranteed to exist.  Is it now?
Should this be?  
```suggestion
      error_name = e.body&.[](:other)&.first&.[](:displayName)
```

SMLuthi: That is correct!

app/models/mvi.rb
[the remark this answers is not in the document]
SMLuthi: Thats a good question. I can put gating in there that verifies if the profile exists. However, for it to even get to that point the user will have had to make two successful calls to MVI... an orchestrated search (which is just a fancy find profile call) and the proxy add. Plus, the osearch is limited to using the icn of the veteran making the call... and I believe only `loa3` users have icn's in their profile.

SMLuthi: Sure, looks a little cleaner

lib/mvi/responses/add_parser.rb
SMLuthi: Case statement using regexes, I like it!

app/models/mvi.rb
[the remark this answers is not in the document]
jholton: Makes sense to me.  
Conversely to my first comment, I wouldn't want to add a guard clause if it's going to hide other problems.  i.e. if `profile` _should_ always be present by the time we get here, I want to see an error so we know that maybe there's something we should investigate.

SMLuthi:
@jholton So the whole test case was strange. The proxy add would be called from the SiP call that starts of the form, we verified that MVI received the request and created the ids, but the ITF call to vets-api immediately following SiP would fail due to the user not having the proper authorization (ie. missing the id's). We'd restart the form and then everything would go off without a hitch. The ids had been generated in the profile successfully at that point. I did a bunch of investigating around why it could be that way but I honestly couldnt find a good reason... On paper, the entire process looked perfect since it all was happening sequentially. Redis isnt async so the changes to destroy the cache should have propagated immediately forcing vets-api to make another call to MVI immediately. In the end, I figured that maybe MVI isnt reflecting its own changes immediately and that may be the issue (though its kind of a stretch). Either way, making these changes are good as it removes an extra find profile call we'll be forcing the api to make.

PR #4107, opened 2020-04-07
7056 pass lat long to ppms

darkbushido:
## Description of change
department-of-veterans-affairs/va.gov-team#7056
PPMS has added support for latitude/longitude in the address field. With this change, the number of mandatory fields has been reduced to `address`, `radius`, and `specialtycode1`.

Feature flags have been added to use the new search functionality and to suppress the unused parameters. 

## Testing
This will be tested on the review instance to verify that we are getting responses.

When this is rolled out to production this needs to have the feature flags enabled.

#### Unique to this PR

- [ ] Enable `facility_locator_ppms_location_query`

#### Applies to all PRs

### DIFF
--- a/lib/facilities/ppms/client.rb
+++ b/lib/facilities/ppms/client.rb
@@ -150,7 +150,11 @@ module Facilities
 
         projection = RGeo::Geographic::ProjectedWindow.new(regeo_factory, x_min, y_min, x_max, y_max)
         lat, lon = projection.center_xy
-        rad = (projection.height * longitude_degree_distance(lat)).round(2)
+        rad = [
+          (projection.height * latitude_degree_distance).round(2),
+          (projection.width * longitude_degree_distance(lat)).round(2)
+        ].max
+        
         {
           latitude: lat,
           longitude: lon,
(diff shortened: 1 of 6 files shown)


spec/request/v0/facilities/ccp_request_spec.rb
```diff
@@ -4,9 +4,11 @@
 
 RSpec.describe 'Community Care Providers', type: :request, team: :facilities do
   around do |example|
-    VCR.use_cassette('facilities/va/ppms', match_requests_on: %i[path query], allow_playback_repeats: true) do
-      example.run
-    end
+    VCR.insert_cassette('facilities/va/ppms', match_requests_on: %i[path query], allow_playback_repeats: true)
+    VCR.insert_cassette('facilities/va/ppms_new_query', match_requests_on: %i[path query], allow_playback_repeats: true)
```
jholton, line 8: What is the difference between the cassettes `ppms` and `ppms_new_query`?  
Is it that `ppms_new_query` sends lat/long params?  
If so, can we rename it to reflect their difference?  Something like `ppms_lat_long_query`?

darkbushido: it's named that way because it's a 'new' query,
I would like the file to go away when I remove the feature flag.

lib/facilities/ppms/client.rb
```diff
@@ -127,11 +127,47 @@ def deduplicate_response_arrays!(response)
         response
       end
 
+      EARTH_RADIUS = 3_958.8
+
+      def regeo_factory
```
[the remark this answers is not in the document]
darkbushido: There are only 2 hard things in Computer Science:
```
0. Cache Invalidation
1. Naming Conventions
7. Asynchronous Callbacks
2. Off-by-one errors
```

PR #4112, opened 2020-04-08
Rename Appeals::Service to Caseflow::Service

bastosmichael:
## Description of change
The motivation here is that teams in va.gov share the vet-api repo with us and are buildng services to consume the lighthosue api.  Renaming that service to be more explicit that it connects to Caseflow and not the Appeals API will help future developers use the right namespaces.

References ticket https://vasdvp.atlassian.net/secure/RapidBoard.jspa?rapidView=140&projectKey=DASH&modal=detail&selectedIssue=DASH-84&quickFilter=393

Also includes devops PR https://github.com/department-of-veterans-affairs/devops/pull/6805

#### Unique to this PR

#### Applies to all PRs


lib/caseflow/service.rb
```diff
@@ -2,30 +2,30 @@
...
+  #   caseflow_response = Caseflow::Service.new.get_appeals(user)
   #
   class Service < Common::Client::Base
     include SentryLogging
     include Common::Client::Monitoring
 
-    configuration Appeals::Configuration
+    configuration Caseflow::Configuration
 
     STATSD_KEY_PREFIX = 'api.appeals'
```
kfrz, line 18: I think we have some specs that may test for this string

bastosmichael: Fixed those tests.

PR #4117, opened 2020-04-08
Remove SSL configuration and certs from HCA configuration.

jholton:
## Description of change
department-of-veterans-affairs/va.gov-team#7764
Removing SSL config from HCA service configuration.
SSL connection negotiation is now handled by fwdproxy.

## Testing
Specs
Tested submission of HCA application manually in review instance (staging environment)
Tested hitting `HCA::Service.new.health_check` manually in production.  
No errors were observed.

#### Unique to this PR

- [x] Remove references to certs in `HCA::Configuration`

#### Applies to all PRs

### DIFF
--- a/lib/hca/configuration.rb
+++ b/lib/hca/configuration.rb
@@ -17,13 +17,6 @@ module HCA
       'HCA'
     end
 
-    # Allow connection to be used without certificates present
-    # SSL configuration provided by fwdproxy
-    # @return [Boolean]
-    def allow_missing_certs?
-      true
-    end
-
     def connection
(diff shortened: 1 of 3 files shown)


lib/hca/configuration.rb
```diff
@@ -34,20 +17,8 @@ def service_name
       'HCA'
     end
 
-    def ssl_options
-      ssl = {
-        verify: true
```
jholton, line 39: This option is unnecessary because the [default value is `true`](https://github.com/lostisland/faraday/blob/d77c9efee9b12763f685f46df470eb22351154f5/lib/faraday/adapter/net_http.rb#L210)

PR #4119, opened 2020-04-09
Appeals hlr pdf fillable

charleystran:
This PR generates a PDF from the from the higher level reviews endpoint.


edmangimelli:
I thought we weren't wanting to use the HLR abbreviation in the names of constants, or were we only concerned with the higher level review model/table?

modules/appeals_api/app/workers/appeals_api/hlr_pdf_submit_job.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+require 'sidekiq'
+require 'appeals_api/hlr_pdf_constructor'
+
+module AppealsApi
+  class HlrPdfSubmitJob
```
edmangimelli, line 7: should we use the HLR abbeviation here?

modules/appeals_api/lib/appeals_api/hlr_pdf_constructor.rb
```diff
@@ -0,0 +1,122 @@
+# frozen_string_literal: true
+
+module AppealsApi
+  class HlrPdfConstructor
```
edmangimelli, line 4: should we use the HLR abbeviation here?

modules/appeals_api/app/workers/appeals_api/hlr_pdf_submit_job.rb
charleystran: No, I will change these. the abbreviation a bad idea.

modules/appeals_api/app/workers/appeals_api/hlr_pdf_submit_job.rb
```diff
@@ -0,0 +1,18 @@
...
+    include Sidekiq::Worker
+
+    def perform(higher_level_review_id)
+      pdf_constructor = AppealsApi::HlrPdfConstructor.new(higher_level_review_id)
+      pdf_constructor.fill_pdf
+      # set status to pending upload
+      HigherLevelReview.update(higher_level_review_id, status: 'processing')
```
edmangimelli, line 14: is this comment irrelevant now? (now that we're using 'processing')

modules/appeals_api/lib/appeals_api/hlr_pdf_constructor.rb
```diff
@@ -0,0 +1,122 @@
...
+    PDF_TEMPLATE = Rails.root.join('modules', 'appeals_api', 'config', 'pdfs')
+
+    def initialize(higher_level_review_id)
+      @higher_level_review = AppealsApi::HigherLevelReview.find(higher_level_review_id)
+      veteran
+    end
```
edmangimelli, line 10: why pass in the id and not the HigherLevelReview object?

modules/appeals_api/app/workers/appeals_api/hlr_pdf_submit_job.rb
charleystran: yea, I need to update the comment good catch, I was putting in the comments to mark the points until we got the central mail piece done

modules/appeals_api/lib/appeals_api/hlr_pdf_constructor.rb
```diff
@@ -0,0 +1,122 @@
...
+        middle_name: auth_headers['X-VA-Middle-Initial'],
+        last_name: auth_headers['X-VA-Last-Name'],
+        ssn: auth_headers['X-VA-SSN'],
+        va_file_number: auth_headers['X-VA-File-Number'],
+        service_number: auth_headers['X-VA-Service-Number'],
+        insurance_policy_number: auth_headers['X-VA-Insurance-Policy-Number'],
+        birth_date: auth_headers['X-VA-Birth-Date'],
+        address_line_1: form_data.dig('veteran', 'address', 'addressLine1'),
+        address_line_2: form_data.dig('veteran', 'address', 'addressLine2'),
+        city: form_data.dig('veteran', 'address', 'city'),
```
edmangimelli, line 32: I think we need to have a game plan for addressLine3

modules/appeals_api/lib/appeals_api/hlr_pdf_constructor.rb
```diff
@@ -0,0 +1,122 @@
...
+        state: form_data.dig('veteran', 'address', 'stateCode'),
+        country: form_data.dig('veteran', 'address', 'country'),
+        zip: form_data.dig('veteran', 'address', 'zipCode'),
+        zip_last_4: form_data.dig('veteran', 'address', 'zip_last_4'),
```
edmangimelli, line 36: this is not a field in the JSON

modules/appeals_api/lib/appeals_api/hlr_pdf_constructor.rb
charleystran: I am generally opposed to passing persisted objects between different classes. While unlikely in this case, when using jobs, it is possible for the underlying database record to change between steps and this makes sure this class starts with the exact DB record.

modules/appeals_api/app/workers/appeals_api/hlr_pdf_submit_job.rb
charleystran: fixed

modules/appeals_api/app/workers/appeals_api/hlr_pdf_submit_job.rb
charleystran: updated

modules/appeals_api/lib/appeals_api/hlr_pdf_constructor.rb
charleystran: fixed

modules/appeals_api/lib/appeals_api/hlr_pdf_constructor.rb
charleystran: there is no address 3 field on the PDF, the only 2 fields are these 
FieldType: Text
FieldName: F[0].#subform[2].CurrentMailingAddress_NumberAndStreet[0]
FieldNameAlt: 9. Current Mailing Address. Enter Number and Street.
FieldFlags: 25165824
FieldJustification: Left
FieldMaxLength: 30
---
FieldType: Text
FieldName: F[0].#subform[2].CurrentMailingAddress_ApartmentOrUnitNumber[0]
FieldNameAlt: 9. Enter Apartment or Unit Number.
FieldFlags: 25165824
FieldJustification: Left
FieldMaxLength: 5

We should document what those 2 fields expect

charleystran: As for zip-4, maybe we should add to the payload as an optional field since its on the form?

modules/appeals_api/lib/appeals_api/hlr_pdf_constructor.rb
charleystran: this has now been added

modules/appeals_api/lib/appeals_api/hlr_pdf_constructor.rb
bastosmichael: yes zip 4 should always be optional or we run the risk of people needing to do a lookup to get the remaining 4 every time (and they might put down the wrong zip4 anyhow).

charleystran: I added it as optional. The dig will make sure it doesnt error on nil

modules/appeals_api/lib/appeals_api/higher_level_review_pdf_constructor.rb
```diff
@@ -0,0 +1,122 @@
...
+        country: form_data.dig('veteran', 'address', 'country'),
+        zip: form_data.dig('veteran', 'address', 'zipCode'),
+        zip_last_4: form_data.dig('veteran', 'address', 'zip_last_4'),
+        benefit_type: form_data['benefitType'],
+        same_office: form_data['sameOffice'],
+        informal_conference: form_data['informalConference'],
+        conference_times: form_data['informalConferenceTimes'],
+        issues: included
+      )
+    end
```
edmangimelli, line 43: Do you think it worthwhile to make a Veteran model that can contain this business logic?

bastosmichael: we can refactor later but we already have a few veteran models that have this sort of information that we can bringin and update etc.

PR #4120, opened 2020-04-09
statsd latency measurement for SAML authentication

ericbuckley:
Reworking the existing `LoginRedirectApplication` redis store to be a generic `SAMLRequestTracker`.  Its new purpose is to 
a) store a redirect url if the user needs to be redirected after authentication 
b) store a created at timestamp so we can track how long the SAML authentication process is taking 
c) have an open ended `payload` attribute so we can add more information if needed in the future.

Also updating the v0/sessions_controller to use the SAMLRequestTracker as well, so we can get a baseline for how long its taking our current users to authenticate.

closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/7322

## Testing
- updated/added unit tests
- locally tested a SAML authentication workflow to make sure redirects are working and the latency time is logged

#### Applies to all PRs


app/controllers/v1/sessions_controller.rb
```diff
@@ -133,12 +134,12 @@ def user_login(saml_response)
...
         set_cookies
         after_login_actions
-        redirect_to url_service.login_redirect_url(saml_uuid: user_session_form.saml_uuid)
+        redirect_to url_service(user_session_form.saml_uuid).login_redirect_url
         if location.start_with?(url_service.base_redirect_url)
           # only record success stats if the user is being redirect to the site
           # some users will need to be up-leveled and this will be redirected
           # back to the identity provider
-          login_stats(:success, saml_response)
+          login_stats(:success, saml_response, user_session_form)
```
patrickvinograd, line 142: If we're only using the saml_uuid in login_stats, could consider only passing that as the parameter :shrug:

patrickvinograd:
My reading of this is that it is robust to a response coming back and no `tracker` being found; might be worth encoding that as a spec where you wipe out the tracker object between request and response and make sure there is a reasonable fallback behavior.

app/controllers/v1/sessions_controller.rb
ericbuckley: `login_stats` uses the `user_session_form` in the failure block

app/controllers/v0/sessions_controller.rb
```diff
@@ -133,23 +134,27 @@ def user_logout(saml_response)
       end
     end
 
-    # rubocop:disable Metrics/MethodLength
+    def success_stat_tags(saml_response, *tags)
```
erluti, line 137: This method does a lot of string-building and is accessing a lot of methods on `@current_user` and `saml_response` and since it's only used in one place I thought this might be better as a variable.  But when I plug it into `login_stats`, the rubocop warning comes back for MethodLength.  Which lead me to thinking maybe just all the successful login information could be it's own method:
```
    def login_stats_success(saml_response, tracker)
      success_tags = [
        "loa:#{@current_user.loa[:current]}",
        "idp:#{@current_user.identity.sign_in[:service_name]}",
        "context:#{saml_response.authn_context}",
        VERSION_TAG
      ]
      StatsD.increment(STATSD_LOGIN_NEW_USER_KEY, tags: [VERSION_TAG]) if request_type == 'signup'
      # track users who have a shared sso cookie
      StatsD.increment(STATSD_LOGIN_SHARED_COOKIE, tags: success_tags) if cookies.key?(Settings.sso.cookie_name)
      StatsD.increment(STATSD_LOGIN_STATUS, tags: success_tags + ['status:success'])
      StatsD.measure(STATSD_LOGIN_LATENCY, tracker.age, tags: success_tags)
    end
```

Then below becomes:
```
def login_stats(status, saml_response, user_session_form = nil)
      tracker = url_service(user_session_form&.saml_uuid).tracker
      case status
      when :success
        login_stats_success(saml_response, tracker)
        callback_stats(:success, saml_response)
      when :failure
[...]
```
What do you think about this approach?

erluti: Similarly in `app/controllers/v1/sessions_controller.rb` if you like it

lib/saml/url_service.rb
```diff
@@ -190,14 +189,23 @@ def add_query(url, params)
...
-        uuid: uuid, redirect_application: @redirect_application
+    # Initialize a new SAMLRequestTracker, if a valid previous SAML UUID is
+    # given, copy over the redirect and created_at timestamp.  This is useful
+    # for a user that has to go through the upleveling process.
+    def initialize_tracker(params, previous_saml_uuid: nil)
+      previous = previous_saml_uuid && SAMLRequestTracker.find(previous_saml_uuid)
+      redirect = previous&.payload_attr(:redirect) || Settings.ssoe.redirects[params[:application]]
+      # if created_at is set to nil (meaning no previous tracker to use), it
+      # will be initialized to the current time when it is saved
+      @tracker = SAMLRequestTracker.new(
```
erluti, line 200: Instead of setting an instance variable in the method, it could return the `SAMLRequestTracker`.  I think it would add some clarity to the initializer what this is doing by calling `@tracker = initialize_tracker(params, previous_saml_uuid: previous_saml_uuid)`.  If this method was named something like `build_saml_request_tracker_from_params` future readers wouldn't need to dig into the implementation unless they specifically cared.  This is more style, I suppose.

app/controllers/v0/sessions_controller.rb
ericbuckley: yeah that works.  The only reason I broke it out in the first place was to get around the robocop warnings

lib/saml/url_service.rb
ericbuckley: originally I did just that, but then I started reading other parts of the code and noticed that `initialize_query_params` (called from the constructor) was setting the instance variable rather than returning it.  So for consitency between the two methods, I stoped returning the object.  Do you think it be better to convert both methods to returning values rather than setting instance variables?

erluti: Yeah, I think so, but I didn't want to comment on code you didn't write.  If you're up for it, I endorse it!

ericbuckley: I ended up just moving the `initialize_query_params` logic to the `initialize` method because it was really simple and the function name was slightly misleading.  Let me know what you think of the changes https://github.com/department-of-veterans-affairs/vets-api/pull/4120/commits/8cd6f4126b0e0262636da084c6ffc588fcdbe447

ericbuckley:
@erluti are there any other changes you wanted to see to this PR?

PR #4124, opened 2020-04-10
Create client for Lighthouse Facilities API

nfasulo:
## Description of change

This PR creates a basic client to make requests against the Lighthouse Facilities API. This work is driven by the extraction of the Facilities API out of vets-api and into into own codebase (https://github.com/department-of-veterans-affairs/lighthouse-facilities). This PR is the first step toward moving the Facility Locator to consume the Facility API instead of interacting with the related models directly.

The code for this client is structured with an eye towards adding additional clients for Lighthouse APIs that vet-api might need to consume in the future.

Instead of supplying a single key for all of vet-api, any team that needs access will need to pass an api key to client. A key for development can be obtained from the Lighthouse [developer portal](https://developer.va.gov/).

For [API-191](https://vajira.max.gov/browse/API-191). Creating keys for the Facility Locator and adding them to ansible templates will be handled with [API-192](https://vajira.max.gov/browse/API-192).

## Testing

Tested against Lighthouse's lab environment

#### Unique to this PR

- [x] client can connect to new Facilities API
- [x] client exposes methods for returning a facility by id and querying with params

#### Applies to all PRs


lib/lighthouse/facilities/configuration.rb
```diff
@@ -0,0 +1,33 @@
...
+      def service_name
+        'LighthouseFacilities'
+      end
+
+      def connection
+        Faraday.new(base_path, headers: base_request_headers, request: request_options, ssl: ssl_options) do |conn|
+          conn.use :breakers
+          conn.use :instrumentation, name: 'lighthouse.facilities.request.faraday'
+
+          # Uncomment this if you want curl command equivalent or response output to log
```
rtravitz, line 22: I know this comment block shows up in a good number of clients, but I think it's safe to drop here.

spec/lib/lighthouse/facilities/client_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+  describe '#get_by_id' do
+    it 'returns a facility' do
+      VCR.use_cassette('/lighthouse/facilities', match_requests_on: %i[path query]) do
+        r = Lighthouse::Facilities::Client.new(Settings.lighthouse.api_key).get_by_id('vha_358')
+        expect(r).to have_attributes(vha_358_attributes)
+      end
+    end
+  end
+
+  describe '#get_facilities' do
```
rtravitz, line 104: Should we have a test for the error cases described in the documentation?

lib/lighthouse/facilities/facility.rb
```diff
@@ -0,0 +1,42 @@
...
+      attribute :feedback, Object
+      attribute :access, Object
+      attribute :mobile, Boolean
+      attribute :active_status, String
+      attribute :visn, String
+      attribute :operating_status, Object
+      attribute :facility_type_prefix, String
+      attribute :unique_id, String
+
+      def initialize(fac)
```
rtravitz, line 29: A test that this init is inserting all of the attributes correctly might be helpful.

lib/lighthouse/facilities/client.rb
```diff
@@ -0,0 +1,30 @@
...
+      def initialize(api_key)
+        self.headers = { 'apikey' => api_key }
+      end
+
+      def get_by_id(id)
+        response = perform(:get, "/va_facilities/v0/facilities/#{id}", nil, headers)
+        Lighthouse::Facilities::Response.new(response.body, response.status).new_facility
+      end
+
+      def get_facilities(params)
```
rtravitz, line 24: A usage example of what params are available would be nice for usability here!

lib/lighthouse/facilities/facility.rb
nfasulo: The tests for `get_by_id` and `get_facilities` both use 
`expect(r).to have_attributes(vha_358_attributes)` 
to verify the attributes of the returned object. Are you thinking of something more than that like tests for the model itself?

spec/lib/lighthouse/facilities/client_spec.rb
nfasulo: Added tests for 400, 401. and 404 errors.

lib/lighthouse/facilities/client.rb
nfasulo: Added method signatures

config/settings.yml
```diff
@@ -484,6 +484,8 @@ va_mobile:
 
 lighthouse:
   api_key: fake_key
+  facilities:
```
darkbushido, line 487: Shouldn't the facilities endpoint use its own api-key instead of sharing the one that other endpoints are using?

lib/lighthouse/facilities/client.rb
```diff
@@ -0,0 +1,45 @@
...
+
+require 'common/client/base'
+require 'lighthouse/facilities/response'
+
+module Lighthouse
+  module Facilities
+    # Documentation located at:
+    # https://developer.va.gov/explore/facilities/docs/facilities
+    class Client < Common::Client::Base
+      attr_accessor :headers
```
darkbushido, line 11: When will the headers be accessed outside of the client?

lib/lighthouse/facilities/client.rb
```diff
@@ -0,0 +1,45 @@
...
+
+      configuration Lighthouse::Facilities::Configuration
+
+      def initialize(api_key)
```
darkbushido, line 15: this needs to call super to pick up any additional initializations in the base class.

lib/lighthouse/facilities/facility.rb
```diff
@@ -0,0 +1,42 @@
...
+      attribute :hours, Object
+      attribute :services, Object
+      attribute :feedback, Object
+      attribute :access, Object
+      attribute :mobile, Boolean
+      attribute :active_status, String
+      attribute :visn, String
+      attribute :operating_status, Object
+      attribute :facility_type_prefix, String
+      attribute :unique_id, String
```
darkbushido, line 27: these should really be alphabetized.

lib/lighthouse/facilities/facility.rb
darkbushido: this needs to call super

darkbushido: I really wish the serializers lived in the serializer directory.

lib/lighthouse/facilities/configuration.rb
```diff
@@ -0,0 +1,31 @@
...
+      end
+
+      def service_name
+        'Lighthouse_Facilities'
+      end
+
+      def connection
+        Faraday.new(base_path, headers: base_request_headers, request: request_options) do |conn|
+          conn.use :breakers
+          conn.use :instrumentation, name: 'lighthouse.facilities.request.faraday'
```
darkbushido, line 21: 👍

lib/lighthouse/facilities/response.rb
```diff
@@ -0,0 +1,36 @@
...
+
+module Lighthouse
+  module Facilities
+    class Response < Common::Base
+      attribute :body, String
+      attribute :status, Integer
+      attribute :parsed_json, String
+      attribute :data, Object
+
+      def initialize(body, status)
```
darkbushido, line 13: super...

lib/lighthouse/facilities/response.rb
```diff
@@ -0,0 +1,36 @@
...
+        self.body = body
+        self.status = status
+
+        self.parsed_json = parse_json
+        self.data = parsed_json['data']
+      end
+
+      def parse_json
```
darkbushido, line 21: ok, Why?
this doesn't get reused and while I do like putting work in methods this is a one-liner your abstracting that gets used as a one-liner in most of the app.

lib/lighthouse/facilities/response.rb
```diff
@@ -0,0 +1,36 @@
...
+        JSON.parse(body)
+      end
+
+      def get_facilities_list
+        data.map do |facility|
+          Lighthouse::Facilities::Facility.new(facility)
```
darkbushido, line 27: `Facilities::Facility` ya I know naming conventions in CS are hard.

spec/lib/lighthouse/facilities/client_spec.rb
```diff
@@ -0,0 +1,151 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Lighthouse::Facilities::Client do
+  let(:params) do
+    {
+      bbox: [60.99, 10.54, 180.00, 20.55] # includes the Phillipines and Guam
```
darkbushido, line 8: Does this need a bounding box? the current facilities lookup gets a center point and radius.

spec/lib/lighthouse/facilities/client_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+      active_status: 'A',
+      visn: '21',
+      operating_status: { 'code' => 'NORMAL' },
+      facility_type_prefix: 'vha',
+      unique_id: '358'
+    }
+  end
+
+  it 'is an Facilities::Client object' do
+    facilities_client = described_class.new('fake_key')
```
darkbushido, line 81: since this is going to be used often why not abstract this into a let statement

spec/lib/lighthouse/facilities/client_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+    expect(facilities_client).to be_an(Lighthouse::Facilities::Client)
```
darkbushido, line 82: What are we testing here? how rspec describes a class?

spec/lib/lighthouse/facilities/client_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+      expect do
+        Lighthouse::Facilities::Client.new('fake_key').get_facilities(params)
+      end.to raise_error(Common::Exceptions::GatewayTimeout)
+    end
+  end
+
+  context 'with a bad API key' do
+    it 'returns a 401 error' do
+      VCR.use_cassette('/lighthouse/facilities_401', match_requests_on: %i[path query]) do
+        expect { Lighthouse::Facilities::Client.new('bad_key').get_by_id('vha_358') }
```
darkbushido, line 98: is there a reason this creates a new facilities client with the api key of bad_key?
since your loading a tape you could use `facilities_client`

lib/lighthouse/facilities/client.rb
darkbushido: is this just to make testing easier?

config/settings.yml
nfasulo: The original idea was to expose the api key when initializing the client so that multiple teams (with separate keys in the config) could use the facilities endpoint. That feels like overkill at this point, so I am going change this to just use a single api key for the client.

spec/lib/lighthouse/facilities/client_spec.rb
nfasulo: Just a sanity check that the client is being created. I stole it from the PPMS client spec.

lib/lighthouse/facilities/facility.rb
nfasulo: Serializer directory? Do you mean you wish this was in app/models?

spec/lib/lighthouse/facilities/client_spec.rb
nfasulo: Doesn't the facilities locator use a bounding box when you search for a state, city, or zip?

lib/lighthouse/facilities/response.rb
nfasulo: This was a bit of copypasta. I have simplified this init.

lib/lighthouse/facilities/facility.rb
darkbushido: https://github.com/department-of-veterans-affairs/vets-api/tree/master/app/serializers

spec/lib/lighthouse/facilities/client_spec.rb
darkbushido: PPMS yes, but for VA facilities no

spec/lib/lighthouse/facilities/client_spec.rb
darkbushido: Ya, I'd rather not cargo cult from that.

lib/lighthouse/facilities/facility.rb
nfasulo: I'm confused. I added this model so that the response is serializable, but am assuming serialization will take place at the controller. Am I making this model do too much?

spec/lib/lighthouse/facilities/client_spec.rb
nfasulo: I have added the distance from the requested point to facilities when that data is returned.

lib/lighthouse/facilities/facility.rb
darkbushido: ok, it looked like you were following the active support serializer format.

PR #4127, opened 2020-04-13
Upgrade Sentry integration gem to latest version

kfrz:
## Description of change

The sentry-raven gem is currently locked to 2.9.0. It could be upgraded to the latest 2.13.0. 

See the [changelog](https://github.com/getsentry/raven-ruby/blob/master/changelog.md#2130) which includes some bugfixes.

## Testing

- [x] Tested events on development Sentry instance via review instance. [1](sentry.vfs.va.gov/vets-gov/platform-api-development/issues/115479) [2](http://sentry.vfs.va.gov/vets-gov/platform-api-development/issues/115581/)

- [x] CI passes
- [x] Sentry events are sent without issue in review instance 
- [x] sentry-raven is at latest version (2.13.0)
#### Applies to all PRs


annaswims:
Is there a reason to upgrade the gem to a version that is newer than the server version?  I thought that their docs generally suggested not to do to that.

kfrz:
@annaswims I was trying to find the correlation, but `2.13.0` is the most updated minor version and what installs when unpinned. Seems to work. 

Can you link to the doc where it warns against that? I tried to search but can't find it.

Gemfile
```diff
@@ -97,7 +97,7 @@ gem 'rgeo-geojson'
 gem 'ruby-saml'
 gem 'rubyzip', '>= 1.3.0'
 gem 'savon'
-gem 'sentry-raven', '2.9.0' # don't change gem version unless sentry server is also upgraded
+gem 'sentry-raven'
```
johnpaulashenfelter, line 100: You might consider locking to < 3.0 or to ~>2.13 so you make it clear you can't/won't go to 3.0 if there's a reason preventing it.

kfrz:
@annaswims found it [here:](https://docs.sentry.io/server/upgrading/)

image

PR #4130, opened 2020-04-13
Namespace existing VAOS controllers as v0 so FHIR can be v1

kreek:
## Description of change
The current VAOS endpoints hang off of /v0/voas, however, the controllers are namespace to VAOS without a version. This PR places the current controllers under v0 so the upcoming FHIR work can go under v1. 

The current routes will not change. There's only one commit but Github is showing most as moved but some as deleted and moved 🤷‍♂️  

UPDATE from @saneshark - Picking up where this left off. Tracking here: https://github.com/department-of-veterans-affairs/va.gov-team/issues/7823

## Testing
Added routing unit tests, local integration tests

Current controllers live under app/controllers/vaos/v0, existing routes do not change.

#### Applies to all PRs


config/routes.rb
```diff
@@ -314,7 +314,7 @@
     mount VeteranConfirmation::Engine, at: '/veteran_confirmation'
   end
 
-  mount VAOS::Engine, at: '/v0/vaos'
+  mount VAOS::Engine, at: '/'
```
saneshark, line 317: I already suspect that this is going to cause grief for code reviewers. ughh... we might have to go:

/vaos/v0 

because I'm pretty sure '/' isn't going to fly.

saneshark: I definitely think we should have the ability to namespace v1 without introducing a new module / engine, but I also see value in having the root level authority happen in the main config/routes.rb not in the modules/vaos/config/routes.rb 

@johnpaulashenfelter I'm going to defer to you on guidance here on how you want us to proceed. We can probably do /v0/vaos/fhir if we have to, but ultimately this problem will surface in the future and it would be a good time to reconcile how we do routing now to better support versioning from within the modules -- ideally without having to introduce a new module / engine.

saneshark: @kreek I think it would be helpful to introduce some comprehensive routing specs for existing routes as a separate PR. Lets get that merged until we can get questions addressed here. Those routing specs will be helpful in ensuring we don't inadvertently brake anything based on whatever approach VSP agrees.

johnpaulashenfelter: @saneshark I'm just thinking through this, and i wonder a bit about just going with `/vaos/whatever_your_convention_needs_to_be` so the numbering is in the module, not the API.

Want to talk to identity a little since they have a similar issue of versioning, but they're also a shared resource.

kreek: @johnpaulashenfelter there was some discussion about this in slack and that was the direction we were leaning https://dsva.slack.com/archives/CBU0KDSB1/p1586891137363400

saneshark:
Almost all of the changes comprise adding `module V0` and calling objects with that namespace.

It's a lot of files modified, but it's important to note that save for routing_specs none of the specs really had to be changed, and we have temporary legacy route rewrite in place that we can remove as soon as the frontend switches over to using these routes.

kreek:
LGTM, @saneshark I'm still on as the author though so I can't approve it 😐

saneshark:
Thanks @kfrz and yes, I apologize about the gnarly nature of this PR. If there was a good way to break it out into chunks I would have done that. Avoided making similar namespace changes where it wasn't 100% needed, any future namespacing for app/services can be performed in separate PRs (if necessary).

saneshark:
Also: here is a ticket I just opened to follow-up on reverting the "rewrite" logic. https://github.com/department-of-veterans-affairs/va.gov-team/issues/8620

PR #4132, opened 2020-04-14
Remove dry-types and dry-struct gems

erluti:
## Description of change
Remove `dry-types` and `dry-struct` from gemfile.   While [looking into upgrading them](https://github.com/department-of-veterans-affairs/va.gov-team/issues/6388), they were not used.  So this is to remove them.

## Testing
Searched code and found no uses of the file using `dry-types` and `dry-struct`.  With the gems and files removed, test suite still passes.


omgitsbillryan:
Probably want feedback from @kreek on this one.

omgitsbillryan:
What about [`resource.rb.erb` which extends `Common::Resource`](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/generators/module/templates/app/models/resource.rb.erb#L6)?

erluti:
> What about [`resource.rb.erb` which extends `Common::Resource`](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/generators/module/templates/app/models/resource.rb.erb#L6)?

Apparently these are used in the new modules generator.  https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/commands/va/va_command.rb

Unless we want to pull out the generator, I'll end up closing this.

kreek:
The generator used that new resource model, based on dry-rb's types/structs, to move away from virtus which no longer being maintained. It wasn't really the resource that some had an issue with but more how dry-rb implements forms. Or their version of forms which is dry-schemas. I personally like it but others felt it moved too far out of Rails land. 

Long story short, the generator uses the dry-struct based resource but it doesn't have to. It could be converted to a virtus based model and then the dry-types gems could be removed.

johnpaulashenfelter:
Maybe the solution is to pull that generator out into its own gem, assuming there's any value in it at all. Not sure how many people use it on new projects.

erluti:
The generator is currently generating classes that depend on these gems, so I don't think making it external has any impact on wether we support these depedencies.

erluti:
Made a ticket to look at this again later https://app.zenhub.com/workspaces/vsp-5cedc9cce6e3335dc5a49fc4/issues/department-of-veterans-affairs/va.gov-team/8461

PR #4139, opened 2020-04-15
wip - Duplicate redis app data

omgitsbillryan:
Draft as part of the work to migrate redis from 3.2.4 to 5.0.5.

This will:
- initialize a second redis instance
- duplicate all data that extends from `Common::RedisStore`.


annaswims:
"current" and "other" are tricky names for the redis instances, especially if the goal is to transition from one to the other.

omgitsbillryan:
I had always intended on renaming that "other redis" and the best I could come up with was, "secondary" 🤷 . Open to other suggestions.

Also, there's more place in code that call `Redis` directly as opposed to inheriting `RedisStore`. I'm looking to see if maybe there's a more elegant way to accomplish this with some kind of callback or subscription model...

spec/lib/common/models/redis_store_spec.rb
```diff
@@ -62,39 +65,48 @@
...
+
+      expect(Redis.current).to receive(:set).once.with(expected_key, expected_val)
+      expect(SECONDARY_REDIS).to receive(:set).once.with(expected_key, expected_val)
       subject.save
     end
   end
 
   describe '#update' do
     it 'updates only user the user attributes passed in as arguments' do
-      expect(subject).to receive(:save).once
```
omgitsbillryan, line 75: Had to remove this because this implicitly creates a method stub for `:save` - which meant that the data was never actually stored into Redis... which is a problem because I added an expectation below where I read directly from Redis.

lib/common/models/redis_store.rb
```diff
@@ -81,13 +84,16 @@ def self.keys
     end
 
     def self.delete(redis_key = nil)
+      secondary_redis.del(redis_key)
       redis_namespace.del(redis_key)
```
johnpaulashenfelter, line 88: This, and all of these, are really a transaction right? If one gets updated and one doesn't you're now out of sync.

And not that its _likely_ but if one succeeds and one fails, the method fails but maybe some data has been written.

The slippery slope here is you end up writing database-style WAL or use locking, both of which are not for the faint of heart.

omgitsbillryan:
There are a few locations that instantiate their own Redis & Redis::Namespace instances. I think a separate PR is warranted (even if we weren't about to migrate) to consolidate these.

AFAI understand, Sentinel is a tool available for creating replicas, monitoring nodes and performing automatic failover in the event that the master node goes down. I _think_ AWS EC abstracts all of this away for us. I can see in the console that in our cluster groups have 2 nodes: 1 primary and 1 replica. I haven't been able to find any documentation that suggests using Sentinel is even possible when using AWS EC.

WRT transactions, data syncing and error handling - do have you any idea how often writes fail? I would expect to see application error(s) in Sentry if a write ever failed. The fact that we don't have any error handling in place for the writes we do currently, makes me think that we (probably 🤷 ) also wouldn't need to worry about it for the vagov EC.

WRT performance, I think adding a few ms is not too big of a deal. We're only duplicating the write-based commands (`del`, `set`, `expire`) which make up ~25% of calls to redis according to cloudwatch. As a mitigation, we could start off by only duplicating `redis_store` that has the longest `ttl` first since it's pointless to duplicate data that has a 60 minute ttl since it will long be expired by the time we officially flip over. This would add complexity but might give us a more gradual approach to make sure performance is still ok.

Sorry for writing a wall of text 😬 , but there were a lot of points.

lib/common/models/redis_store.rb
omgitsbillryan: Has a `set` or `del` _ever_ failed so far to date that you know of? If it had, I would expect to see an error somewhere in sentry or maybe logs.

johnpaulashenfelter:
Digging more into the details here after some work on migrating Redis instances all week 😄 

Seeding a new Redis cluster with an existing production Redis cluster (1.5gb in my tests) looks like it takes 15-30 minutes, likely on the lower side of that range. That's taking a live snapshot from a running Redis cluster with zero downtime (though small performance hit) and restoring to a completely new Redis instance. Both take _roughly_ the same amount of wall clock time and it varies depending on a million reasons, but 30m is a good enough estimate.


The GET action now acts like a two-tier cache -- read from new cluster, if not found try the old cluster before returning not found. If found in the old cluster, update the new cluster so there's a hit next time

Pros: Far fewer dual reads
Cons: potentially adds reads of old data that was deleted/changed during the downtime after the cluster was snapshotted

The SET action now just becomes write to the new cluster. 

Pros: New data in the new db
Cons: Not sure

The DEL action now can become a no-op I think?

Pros: Don't need to delete what's not there!
Cons: Old cluster keeps data longer than necessary, though I guess you could delete it?

etc.

With the seeded data, the calls to the old cluster should drop off a *lot* more quickly I'd think. It would be straight-forward to monitor that to see how things work.

Thoughts?

johnpaulashenfelter:
Also MONITOR is an interesting idea in here, but there specific solution (SQS, 300rpm) seems like its not tenable for us volume-wise.

omgitsbillryan:
So, 15-30 minutes to create snapshot from old Redis cluster, then 15-30 minutes to apply snapshot to the new Redis cluster, for a total of 30-60 minutes - am I following that properly?

After that, it looks like you're describing a strategy where we use the new cluster and migrate any old data in real time - I had the same thought when I took this on originally. You mention `get`, `set`, & `del`, but the full list of `Redis` method calls made in `vets-api` is `[get, set, del, expire, keys, exists, ttl]`, so we'd need a strategy for all of them. Here's some notes I had made for each:

```
Strategy: Use the new Redis and migrate old data in real time

get:
  - get from new
  - if nil then get from old

set:
  - set in the new

del:
  - if del new == 1 return 1
  - return del old
  - notes: del returns 1 if key exists

expire:
  - if expire new == true then return true
  - return expire old
  - notes: expire() returns false if the key doesnt exist

keys:
  - (get old and new keys).unique

exists:
  - return exists?(new) if true
  - else return exists?(old)

ttl:
  - return new ttl if not nil
  - return old ttl
```

You're right that calls to the old cluster would quickly drop off, but the downside is the added complexity + risk of bugs in code w/ those goofy if-else conditions. Also, I don't think we need to be concerned with performance. I _think_ (🤞) that simply writing _everything_ to multiple Redis instances will be nearly unnoticed, expecially if we `.pipeline` all our calls to the duplicate Redis instance. ([see `.pipelined`](https://dev.to/erikaheidi/pipeline-all-the-things-redis-performance-boost-at-application-level-3blb)).

omgitsbillryan:
Closing this wip. We ended up going with a _slightly_ different approach 👉 https://github.com/department-of-veterans-affairs/vets-api/pull/4210

PR #4141, opened 2020-04-16
Updating Reports Mailer

bastosmichael:
## Description of change
Defined in Ticket https://vasdvp.atlassian.net/browse/DASH-27

## Original issue(s)

- Adding per consumer and per status counts and rates to the report

Screenshot from 2020-04-20 15-08-30


ryanlink:
@bastosmichael would it be possible/difficult to group these by consumer as well, or output to csv so we can sort as needed?

PR #4146, opened 2020-04-16
Added COVID19 Chatbot Token service to vets API

shaunhubbard-va:
## Description of change
This change set will add a service endpoint to issue tokens for the COVID 19 chatbot.

## Original issue(s)
department-of-veterans-affairs/covid19-chatbot/issues/91

Three new values were added to `settings.yml`, the two secrets vary by environment and their key will be entered into credstash.

An `rspec` file was added in order to test the new endpoint `rpec/controllers/v0/coronavirus_chatbot_controller_spec.rb`, and two cassettes were made in order to facilitate responses from the external service.


config/routes.rb
```diff
@@ -288,6 +288,8 @@
         defaults: { feature: feature }
       )
     end
+
+    resources :coronavirus_chatbot, only: :create
```
karlbrown-va, line 292: should the route be more like coronoavirus_chatbot_tokengen or something? given that this endpoint is mainly just returning the valid token as per my understanding.

shaunhubbard-va: how about `/v0/coronoavirus_chatbot/tokens` so that we can keep it railsy?

karlbrown-va: sure - I take it tokens is better than token re: rails conventions?

shaunhubbard-va: typically resources are pluralized as to reflect the real world, depening on inflector conventions. TL;DR yes

config/settings.yml
```diff
@@ -506,3 +506,8 @@ vet_verification:
 github_stats:
   username: github-stats-rake
   token: fake_token
+
+coronavirus_chatbot:
+  directline_uri: northamerica.directline.botframework.com
```
johnpaulashenfelter, line 511: Alphabetical if you don't mind please :)

config/settings.yml
```diff
@@ -506,3 +506,9 @@ vet_verification:
 github_stats:
   username: github-stats-rake
   token: fake_token
+
+# Settings for the coronavirus chatbot tokens
+coronavirus_chatbot:
+  app_secret: a_secret
```
karlbrown-va, line 512: consider calling this fake-secret

config/settings.yml
```diff
@@ -506,3 +506,9 @@ vet_verification:
...
+  directline_uri: northamerica.directline.botframework.com
+  webchat_secret: another_secret
```
karlbrown-va, line 514: same here, to match the pattern elsewhere

app/controllers/v0/coronavirus_chatbot/tokens_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+
+module V0
+  module CoronavirusChatbot
+    class TokensController < ApplicationController
+      skip_before_action :authenticate, only: [:create]
+
+      rescue_from 'V0::CoronavirusChatbot::TokensController::ServiceException', with: :service_exception_handler
+      rescue_from Net::HTTPError, with: :service_exception_handler
+
+      def create
```
karlbrown-va, line 11: what sort of authentication is on this endpoint? Can anyone hit it or only va.gov front end?

shaunhubbard-va: per line 6  in this same file, none

PR #4147, opened 2020-04-17
Add Veteran and Caregiver to CARMA::Submission::Metadata

kevinmirc:
Some preliminary work for sending ICNs with the CARMA submission of a 10-10CG


spec/lib/carma/models/submission/metadata_spec.rb
```diff
@@ -16,16 +98,70 @@
...
+        ]
+      )
     end
   end
 
   describe '#to_request_payload' do
     it 'can receive :to_request_payload' do
-      metadata = described_class.new claim_id: 123
-      expect(metadata.to_request_payload).to eq(
+      example_1 = described_class.new claim_id: 123
```
erluti, line 115: If you have two examples, I think these should be two separate tests.
`it 'can receive :to_request_payload'`(example_2) and maybe `it 'can receive :to_request_payload with only claimId'`(example_1) or something like that?

lib/carma/models/submission/metadata.rb
```diff
@@ -4,12 +4,42 @@ module CARMA
...
+                            :primary_caregiver,
+                            :secondary_caregiver_one,
+                            :secondary_caregiver_two
 
         def initialize(args = {})
           @claim_id = args[:claim_id]
+
+          self.veteran = args[:veteran] || {}
+          self.primary_caregiver = args[:primary_caregiver] || {}
+          self.secondary_caregiver_one = args[:secondary_caregiver_one] if args[:secondary_caregiver_one]
```
erluti, line 25: I think it would be good to add a specific test for one of these cases, because this is a little unclear.
If you are worried that `args[:secondary_caregiver_one]` can be nil, maybe the setter logic should just set it to nil when the value passed to secondary_caregiver_one is nil.  But if `args[:secondary_caregiver_one]` can sometimes be `false`, it would make more sense if I could see a test with the scenario where `false` is this value.

lib/carma/models/submission/metadata.rb
```diff
@@ -4,12 +4,42 @@ module CARMA
...
+          self.secondary_caregiver_two = args[:secondary_caregiver_two] if args[:secondary_caregiver_two]
+        end
+
+        def veteran=(args = {})
```
erluti, line 29: I think we should avoid using the name `args`.  It doesn't do a good job of describing what the arguments for the method are.  In the case of an initializer sometimes arguments are just a grab bag and "args" can make sense, but here it should be something like "veteran_data" or "veteran_hash"

erluti: Also, default values for args doesn't make sense for an `=` method, so I would delete the ` = {}`

lib/carma/models/submission/metadata.rb
kevinmirc: The related specs are located here: `spec/lib/carma/models/submission/metadata_spec.rb:51` "::new initializes with defaults".

At the end of the day, I want newly instantiated Metadata models to _always_ to have a Veteran and Caregiver model in the `veteran` and `primaryCaregiver` namespace but default `nil` for the `secondaryCarevier*` namespaces. So I think it makes sense to keep that logic in the initialize method (since it's initialization logic).

I've instead updated the arg check to be more explicit, to show that I'm checking for nil rather than any falsy value. Let me know if that clears up this nil check, or if there is a way to make my intentions more readable.

kevinmirc:
@erluti any way you can mark the code climate error as "won't fix"?

lib/carma/models/submission/metadata.rb
erluti: I think making it more explicit is a big improvement.  But I think it makes the setters better/safer if that check is in there.  Currently someone can call `metadata.seconday_caregiver_one = nil` and it will blow up.  So if that logic is in the setter, then you don't need the checks in the initializer which as a bonus knocks out the cognitive complexity check.
I think I'd write the new setter like this, but there's other valid approaches.
```
def secondary_caregiver_one=(sc_one_metadata_hash)
          if sc_one_metadata_hash.nil?
            @secondary_caregiver_one = nil
          else
            @secondary_caregiver_one = Metadata::Caregiver.new(sc_one_metadata_hash)
          end
        end
```

kevinmirc: That's a good point. I've moved the nil check to the setters and added a spec for setting it to nil.

lib/carma/models/submission/metadata.rb
```diff
@@ -4,12 +4,42 @@ module CARMA
...
+        def veteran=(veteran_data_hash)
+          @veteran = Metadata::Veteran.new(veteran_data_hash)
+        end
+
+        def primary_caregiver=(pc_metadata_hash)
+          @primary_caregiver = Metadata::Caregiver.new(pc_metadata_hash)
+        end
+
+        def secondary_caregiver_one=(sc_one_metadata_hash)
+          @secondary_caregiver_one = Metadata::Caregiver.new(sc_one_metadata_hash) unless sc_one_metadata_hash.nil?
```
erluti, line 38: I think this could have the weird side effect that if it's set to a value, and then later someone tries to set it as nil it won't change the value.  But I don't imagine that's a use-case for this

kevinmirc: Good catch. I've updated it.

kevinmirc:
Thanks for the review @erluti !

PR #4151, opened 2020-04-17
move flipper sign in link to top

annaswims:
## Description of change
Move the sign in prompt from the bottom to the top of the page in the flipper admin screen so that it is more visible to users who need to log in or are not identity-verified.
before
Screen Shot 2020-04-17 at 10 41 01 AM
After
Screen Shot 2020-04-17 at 10 45 35 AM


johnpaulashenfelter:
@annaswims thanks so much for the before and after

PR #4157, opened 2020-04-20
Adding fields to va facilities response

darkbushido:
the new fields are :operating_status

## Description of change

Adding :operating_status to the facilities serializer
Until we switch over to using the lighthouse API instead of active record this is going to grab the single field we need from lighthouse.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#7452

This is adding the fields as `nil` until we switch over to the lighthouse client.

### DIFF
--- a/spec/request/v0/facilities/va_request_spec.rb
+++ b/spec/request/v0/facilities/va_request_spec.rb
@@ -49,6 +49,97 @@ RSpec.describe 'VA GIS Integration', type: :request, team: :facilities, vcr: vcr
     expect(response.body).to be_a(String)
     json = JSON.parse(response.body)
     expect(json['data']['id']).to eq('vha_648A4')
+    expect(json).to include(
+      {
+        "data" => {
+          "id" => "vha_648A4",
+          "type" => "va_facilities",
+          "attributes" => {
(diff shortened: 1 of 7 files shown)


spec/request/v0/facilities/va_request_spec.rb
```diff
@@ -2,9 +2,20 @@
...
+  match_requests_on: %i[path query],
+  allow_playback_repeats: true,
+  record: :none
+}
+
+RSpec.describe 'VA GIS Integration', type: :request, team: :facilities, vcr: vcr_options do
   include SchemaMatchers
 
+  before do
+    Flipper.enable(:facility_locator_pull_operating_status_from_lighthouse, true)
```
erluti, line 16: I removed this from the spec and it still passes.  I even tried `Flipper.disable(:facility_locator_pull_operating_status_from_lighthouse)` and it still passes.  Is this necessary?  Should there be a test specifically for when it's enabled/disabled?

darkbushido: this is a fast hack to get it out asap. I should be verifying that the hash key has a value instead of the hash key existing.

PR #4159, opened 2020-04-20
Fixed typo in call to Covid Chatbot settings

shaunhubbard-va:
## Description of change
There was a typo in a call to settings.

## Original issue(s)
N/A

None.

An assert was added to prevent this as a future regression


spec/controllers/v0/coronavirus_chatbot/tokens_controller_spec.rb
```diff
@@ -33,7 +35,7 @@
 
         expect(token_data['userId']).to eq user_id
         expect(token_data['locale']).to eq locale
-        expect(token_data['directLineURI']).to eq Settings.coronavirus_chatbot.directLine_uri
+        expect(token_data['directLineURI']).to eq directline_uri
```
johnpaulashenfelter, line 38: Is the token data really capitalized here? don't have a lot of context

shaunhubbard-va: @johnpaulashenfelter for context. `https://github.com/department-of-veterans-affairs/covid19-chatbot/blob/master/server.js#L110` The whole of this service was lifted from that file.

PR #4167, opened 2020-04-21
7826 VAOS FHIR service

kreek:
## Description of change
Adds a VAOS FHIR based service that will perform calls and pass through the response (not parse and re-serialize) responses to the FE. It's methods will match the terminology use in the FHIR DTSU 2 REST spec http://hl7.org/fhir/dstu2/http.html. Currently only read is implemented. 

It will catch errors and remap them to vets-api service exceptions.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#7826

* Adds a new error key to exceptions.en.yml

### Testing done
units tests, local integration tests

### Testing planned
review/staging instance integration tests


modules/vaos/app/services/vaos/v1/fhir_service.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+require_relative '../base_service'
+require 'common/exceptions/external/backend_service_exception'
+
+module VAOS::V1
+  class FHIRService < VAOS::BaseService
+    STATSD_KEY_PREFIX = 'api.vaos.fhir'
+
+    def perform(http_method, path, params = nil)
```
kreek, line 10: I originally had this method named `make_request` as `perform` is private in the parent class. Don't know if that's more or less confusing than having it named the same thing. This way kind of breaks encapsulation but in this instance it's meant to be called with dynamic http methods and paths.

kreek:
@saneshark still can't seem to add AJ to vets-api PRs, his github username is `ajmagdub`

modules/vaos/app/services/vaos/v1/fhir_service.rb
kreek: I'd also like to protect it with a hash of resources (paths), maybe we can load and cache the conformance doc and confirm that resource is available?

kreek: Actually since we're not doing dynamic routes having a hash of resources probably isn't necessary.

saneshark: might prove useful in route globbing ruleset to have, but that should only be needed temporarily.

saneshark: you could do 

private_method :perform if you don't want to  break encapsulation after defining the method.

saneshark: but what's wrong with the existing perform method? just need distinct headers added in right?

modules/vaos/app/services/vaos/v1/fhir_service.rb
```diff
@@ -0,0 +1,24 @@
...
+    end
+
+    private
+
+    def config
+      VAOS::V1::FHIRConfiguration.instance
+    end
+
+    def headers
+      super.merge!('Content-Type' => 'application/json+fhir')
```
saneshark, line 21: Why `merge!` instead of `merge` -- i know it only gets called once, but it seems more functional in terms of non idempotence.

modules/vaos/spec/services/v1/fhir_service_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+require 'rails_helper'
+
+describe VAOS::V1::FHIRService do
+  subject { VAOS::V1::FHIRService.new(user) }
+
+  let(:user) { build(:user, :vaos) }
+
+  before { allow_any_instance_of(VAOS::UserService).to receive(:session).and_return('stubbed_token') }
+
+  describe '#make_request' do
```
saneshark, line 12: here I think you're calling attention to a method that you decided to change to #perform instead.

modules/vaos/spec/services/v1/fhir_service_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+    context 'with an invalid http method' do
+      it 'raises a no method error' do
+        expect { subject.perform(:let, 'Organization/353830') }.to raise_error(
+          NoMethodError
+        )
+      end
```
saneshark, line 18: NITPICK: I don't know if you need to test this, since it's hopefully already tested by the inherited parent class. Maybe just test that subject.is_an_instance_of(Common::Client...)?

modules/vaos/spec/services/v1/fhir_service_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+    end
+
+    context 'when VAMF returns a 404' do
+      it 'raises a backend exception with key VAOS_404' do
+        VCR.use_cassette('vaos/fhir/404', match_requests_on: %i[method uri]) do
+          expect { subject.perform(:get, 'MissingResource') }.to raise_error(
+            Common::Exceptions::BackendServiceException
+          ) { |e| expect(e.key).to eq('VAOS_404') }
+        end
```
saneshark, line 27: Question: I'm a little confused why we need to have a distinct VAOS_404 exception. Are we not satisfied with the generic description returned for 404 errors?

modules/vaos/spec/services/v1/fhir_service_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+              'address' => [
+                {
+                  'line' => ['2002 Holcombe Blvd.'],
+                  'city' => 'Houston',
+                  'state' => 'TX',
+                  'postalCode' => '77030-4298'
+                }
+              ]
+            }
+          )
```
saneshark, line 61: NITPICK: Should this live in a fixture or something instead? I think it might be more reusable to have it live in a fixture resource elsewhere.

saneshark: Isn't this implicitly expected by using the VCR cassette you have recorded? LOL Maybe we're content verifying that the response status is 200, and you could reference the body payload YAML.read(...) instead?

config/locales/exceptions.en.yml
```diff
@@ -522,6 +522,12 @@ en:
         code: 'VAOS_403'
         detail: 'This user is forbidden from performing requests on the upstream server'
         status: 403
+      VAOS_404:
+        <<: *external_defaults
+        title: Not Found
+        code: 'VAOS_404'
+        detail: 'The resource could not be found'
+        status: 404
```
saneshark, line 530: Question, why was this needed?

kreek: might be another way around this but without the specific 404 entry here it was being mapped to a 400 by default

modules/vaos/app/services/vaos/v1/fhir_service.rb
```diff
@@ -0,0 +1,49 @@
...
+      #
+      def read(resource_type, id)
+        unless RESOURCES.include?(resource_type)
+          raise Common::Exceptions::InvalidFieldValue.new('resource_type', resource_type)
+        end
+
+        perform(:get, "#{resource_type}/#{id}", nil, headers)
+      end
+
+      # TODO: #create http://hl7.org/fhir/dstu2/http.html#create
```
erluti, line 34: Super nitpicky, but I think it's better to delete these TODOs and just have a well-written ticket on the backlog.

PR #4168, opened 2020-04-21
Inbound SSOe

ericbuckley:
Users that come to VA.gov after having already logged into eauth.va.gov, will be able to automatically log in.  However the cached SAML response we get back from SSOe (when users had logged in with MHV or DSLogon credentials) will have 
 a different `authnContext` to account for, so we need to make some changes to accomdate for the 'urn:oasis:names:tc:SAML:2.0:ac:classes:Password' authnContext.

closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/5417

- [x] unit tests for the new authnContext
- [x] unit tests for inbound saml user attributes


ericbuckley:
@patrickvinograd what do you think about me removing the "email" changes from this PR (at least until https://github.com/department-of-veterans-affairs/va.gov-team/issues/8305 is resolved) so we can at least merge the authnContext changes?

lib/saml/user_attributes/ssoe.rb
```diff
@@ -96,7 +99,16 @@ def mhv_correlation_id
       end
 
       def mhv_account_type
-        safe_attr('va_eauth_mhvassurance')
+        val = safe_attr('va_eauth_mhvassurance')
+        val ||= safe_attr('va_eauth_credentialassurancelevel') if csid == 'mhv'
+        # FIXME: for inbound SSOe what number indicates Advanced/Basic?
```
patrickvinograd, line 104: Both Basic and Advanced would map to `credentialassurancelevel=1`

lib/saml/user_attributes/ssoe.rb
```diff
@@ -53,7 +54,9 @@ def birth_date
       end
 
       def email
-        safe_attr('va_eauth_emailaddress')
+        val = safe_attr('va_eauth_emailaddress')
+        val ||= safe_attr('va_eauth_commonname')
```
patrickvinograd, line 58: I'm surprised by  this line. I know you have the guard on the next line to check if it's an email address. And I know one of the captured payloads for inbound SSO had an email address here but I'm not 100% confident that that represents a user-supplied email address (vs. e.g. coercing an MHV username into an email address with a phony domain) We should validate that  with an additional inbound MHV user SAML payload.

lib/saml/user_attributes/ssoe.rb
patrickvinograd: So we have a loss of fidelity here...BUT, there's a separate thread with the MHV team going where we would do away with a lot of  the account-level-conditional logic so 🤞 that this wouldn't be an issue.

patrickvinograd: I'm wondering if we can just return nil here for inbound. The mhv_account_type is used in a few places:
- to populate the `sign_in:account_type` object which is used for the MHV shared cookie which we don't use in SSOe.
- to populate `user.mhv_account_type` which is used for various service authorization checks, but that has a fallback to getting the account type from the MHV API.
- wherever else sign_in might be serialized into the user profile for front end, I haven't checked this part.

lib/saml/user_attributes/ssoe.rb
ericbuckley: I only made that change because of the value I saw in the `ssoe_inbound_mhv_premium` factory.  I'll remove the changes if thats not guarenteed to be a verfied email

lib/saml/user_attributes/ssoe.rb
ericbuckley: If we return `nil` here for MHV inbound users, `loa_highest` will always evaluate to 1.  Is that value only used for upleveling or does the vets-website code use that value for something?

ericbuckley: @patrickvinograd looks like vets-website is using the account type here https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/platform/site-wide/cta-widget/helpers.js#L37-L52

lib/saml/user_attributes/ssoe.rb
```diff
@@ -11,6 +11,7 @@ class SSOe
                                    uuid idme_uuid sec_id mhv_icn mhv_correlation_id mhv_account_type
                                    dslogon_edipi loa sign_in multifactor].freeze
       IDME_GCID_REGEX = /^(?<idme>\w+)\^PN\^200VIDM\^USDVA\^A$/.freeze
+      INBOUND_AUTHN_CONTEXT = 'urn:oasis:names:tc:SAML:2.0:ac:classes:Password'
```
kfrz, line 14: should this have a `.freeze` appended?

ericbuckley: good idea, ill add it

ericbuckley: actually thats giving me a linting error.  @kfrz are ruby constant strings by default frozen?

`lib/saml/user_attributes/ssoe.rb:14:31: C: Style/RedundantFreeze: Do not freeze immutable objects, as freezing them has no effect.`

PR #4170, opened 2020-04-22
remove carma stub and hit endpoint

kevinmirc:
Resolves: https://github.com/department-of-veterans-affairs/va.gov-team/issues/7789

## Description of change
- [x] Standup a review instance to test the connection with CARMA.
- [x] Send additional header on CARMA submissions
- [x] Use real CARMA requests on Staging
- [x] Make updates to support version 5.0.0 of `vets-json-scheam`

- [x] App settings for review instances and Staging have been set up
- [x] This PR depends on https://github.com/department-of-veterans-affairs/vets-api/pull/4185 being merged because they make support of `vets-json-schema` of version < 5.0.0.
  - The dependent branch was merged into this pr.


jholton:
Why does #4185 need to be part of this?
Have you verified that pulling in the 686C-674 and 526 schema changes won't break anything?

config/routes.rb
```diff
@@ -70,7 +70,7 @@
     resource :hca_attachments, only: :create
 
     # Excluding this feature until external service (CARMA) is connected
-    resources :caregivers_assistance_claims, only: :create
+    resources :caregivers_assistance_claims, only: :create unless Rails.env.production?
```
jholton, line 73: Just so I'm clear, the CARMA functionality hasn't launched in production, right?

jholton:
Notes:  This PR contains the author's `CARMA` functionality and the contents of PR #4185, which removes `vic` functionality that was never launched in production.
Combining these PRs is not ideal, but is necessary due to the order of changes and versioning of `vets_json_schema`.

To answer my question above:
> Have you verified that pulling in the 686C-674 and 526 schema changes won't break anything?

I received confirmation via Slack [here](https://dsva.slack.com/archives/CLY6Q69RV/p1588097021261200) and [here](https://dsva.slack.com/archives/C5AGLBNRK/p1588097919076500) that including those schema changes should not hurt anything.

config/routes.rb
kevinmirc: Yes, that's correct.

spec/fixtures/pdf_fill/10-10CG/kitchen_sink.json
```diff
@@ -1,103 +1,95 @@
 {
-  "veteran": {
```
erluti, line 2: is this just changing the tabs from 2 spaces to 4?  looking in the rest of this directory, it looks like `overflow.json` is 4 but `simple.json` is 2.  Looking around, it looks like the other pdf_fills are using 2 space tabs.  So I think this should be using 2 spaces.

spec/request/caregivers_assistance_claims_request_spec.rb
```diff
@@ -80,7 +80,7 @@
     context 'when unauthenticated' do
       it_behaves_like 'any invalid submission'
 
-      timestamp = Date.parse('2020-03-09T06:48:59-04:00')
+      timestamp = DateTime.parse('2020-03-09T06:48:59-04:00')
```
erluti, line 83: Why was this changed?  Should it be the Railsy `Time.zone`?

kevinmirc: This change was just to retain the time information all together. I realized this test was comparing a `Date` to `DateTime`.

kevinmirc: This change was needed for this following expectation:
```
expect(DateTime.parse(res_body['data']['attributes']['submittedAt'])).to eq timestamp
```

The test would pass before because I was using `CARMA::Client#create_submission_stub`, which generates a DateTime of `#now`. This would make the test above pass because it would compare a `Date` with time 0 to a `DateTime` with time 0.

But once I stopped using the stub, and started using the VCR cassette, the time info needed. Because I was then comparing `Date` with time 0 and `DateTime` with time `06:48:59-04:00`.

jholton:
Just to keep on with the `vets-json-schema` versioning issue...is there a PR to update the FE to the latest version?

kevinmirc:
@jholton Here's the related UI PR https://github.com/department-of-veterans-affairs/vets-website/pull/12310/files

PR #4176, opened 2020-04-22
add config.hosts variable for rails 6

wyattwalter:
## Description of change

I don't know if this is a good way to do this, but it looks like Rails 6 added a `config.hosts` variable that we need to set. We already were doing this through `Settings.virtual_hosts` in config, so just doing this addresses the immediate problem. I'd defer to anyone else's knowledge of Rails if there's a better way here.

## Original issue(s)

Tested in a review instance so far to great success.


jholton:
Closing for now in favor of #4178.
Will revisit for production/other env enforcement.

PR #4182, opened 2020-04-23
consolidate Redis initialzation to one place

omgitsbillryan:
## Description of change
Consolidates our initialization of Redis. 

### Additional Context
We were defining redis like this in multiple places:
```ruby
REDIS_CONFIG = Rails.application.config_for(:redis).freeze
Redis.current = Redis.new(REDIS_CONFIG[:redis])
```

When really, it should only happen once in an initializer.

## Original issue(s)
Connects: https://github.com/department-of-veterans-affairs/va.gov-team/issues/6865

This PR is basically a preceding PR to duplicate our redis data across multiple Redis instances. Even if that work falls through, this refactor still cleans up our global instantiation of Redis.

_EDIT:_ This is also connected to https://github.com/department-of-veterans-affairs/va.gov-team/issues/6262 👈 thx @erluti


omgitsbillryan:
I had to rename `config/initializers/redis.rb` to `01_redis.rb` to ensure it runs first. Reason being, other initializers make use of the instance of redis and initializers run in alphabetical order.

PR #4186, opened 2020-04-24
[Settings.yml] Update Lighthouse dev environment

ncksllvn:
## Description of change

Following up to https://github.com/department-of-veterans-affairs/devops/pull/6651, this PR updates some references to the Lighthouse dev environment, which should now use `sandbox-api` instead of `dev-api`.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/7441
https://github.com/department-of-veterans-affairs/va.gov-team/issues/8369

Changes to settings.yml


kelsonic:
image

Looks like there's a few places where we're still referencing `dev-api.va.gov`, not sure if we want to update those as well 🙂

rtravitz:
> Looks like there's a few places where we're still referencing `dev-api.(va|vets).gov`, not sure if we want to update those as well 🙂

(Saw the edit late, but still posting in case it's helpful context for anyone.)

There's a (not very intuitive) split in vets-api between Lighthouse URLs and those used by other teams in the repo. Lighthouse should look like `https://dev-api.va.gov/services/*`.

ncksllvn:
@kelsonic haha that's just the regular dev-api, whereas Lighthouse has `services` as the first piece of its path. vets-website shouldn't be hitting Lighthouse directly because that would expose the secret API key

ncksllvn:
All of the other references I see to `dev-api.va.gov/services` are in cassettes so I think we're okay, although I do see Jenkins saying this PR fails

ncksllvn:
@department-of-veterans-affairs/vsp-tools-backend sorry for the noise on this one- but it's ready for review finally

config/settings.yml
```diff
@@ -109,7 +109,7 @@ vet360:
     timeout: 30
     mock: false
   address_validation:
-    hostname: 'dev-api.va.gov'
+    hostname: 'sandbox-api.va.gov'
```
ncksllvn, line 112: This was elusive because i was always looking for dev-api.va.gov/services

ncksllvn: @lihanli for you awareness the Lighthouse dev env changed URLs

omgitsbillryan:
For better or worse, some of these settings don't exist in our prod `settings.yml` file (`decision_review` for example). This means you're effectively updating prod settings. Is this intentional? 

I'm totally OOL on this one btw 😉 , so apologies if that's a dumb question.

Oh, also, I still see references to `dev-api` in the review instance settings YAML file - does it need updating there too?

ncksllvn:
> For better or worse, some of these settings don't exist in our prod settings.yml file (decision_review for example). This means you're effectively updating prod settings. Is this intentional?

It isn't directly intentional, but even if we have a prod Vets-API method that for some reason points to a dev Lighthouse API, we still need to update that Lighthouse URL to use the new sandbox URL. `decision_review` still needs a prod connection setup as it hasn't been launched yet. CC: @Mottie @annaswims 

> Oh, also, I still see references to dev-api in the review instance settings YAML file - does it need updating there too?

Hm I don't know anything about that one. If there are instances of `dev-api.va.gov/services/` then yes we should!

PR #4188, opened 2020-04-24
686C-674 prefill

lihanli:
## Description of change

prefill the 686c-674 form. dependents data doesn't need to be prefilled for now

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/7883

tested with unit tests


app/models/form_profiles/va21686c674.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    super
+  end
+
+  def metadata
+    {
+      version: 0,
+      prefill: true,
+      # TODO: get correct returnUrl
+      returnUrl: '/veteran-information'
```
lihanli, line 32: @cohnjesse @micahchiang  hey do you guys know what the form specific path of the first page of this form is? i need to know for the returnUrl field of the save in progress form. for example in 1990 form the returnUrl is "/applicant/information" https://staging.va.gov/education/apply-for-education-benefits/application/1990/applicant/information

micahchiang: @lihanli - path for the first page of the form is `/686-options-selection`

app/models/form_profiles/va21686c674.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+class FormProfiles::VA21686c674 < FormProfile
```
kfrz, line 3: It'd be super neat if we had class-level documentation that gave an overview of *what* the form is for, *where* it can be found on VA.gov (or which product) and any other prerequisites (LOA3, , MHV Premium, etc).

Not a blocker for this PR but just a suggestion.

kfrz: Form class name is `VA21686c674`, in configuration mapping it's referred to as `686C-674`, and in spec it's `v686_c_674` & `686C-674` 

Which is it, why, and can we normalize? 

When working with a multitude of forms (or anything) that are named obscurely, it's important to be able to recognize them quickly.

While I realize this is a problem beyond the scope of this particular PR, I'm wondering if we can set a good example - or at least clarify why they have different designations depending on the context.

spec/models/form_profile_spec.rb
```diff
@@ -73,6 +73,33 @@
     }
   end
 
+  let(:v686_c_674_expected) do
```
kfrz, line 76: These `let` blocks are gross and should be extracted to fixtures at some point.

lihanli:
> In your description you mention
> 
> > dependents data doesn't need to be prefilled for now
> 
> Is there a document-able reason why not - or is it just in effort to narrowly scope the work here?
> 
i was told on slack by @kathleencrawford it wasn't needed, not sure what the reason was, maybe she has more information

app/models/form_profiles/va21686c674.rb
lihanli: > Form class name is `VA21686c674`, in configuration mapping it's referred to as `686C-674`, and in spec it's `v686_c_674` & `686C-674`
> 
> Which is it, why, and can we normalize?
> 
> When working with a multitude of forms (or anything) that are named obscurely, it's important to be able to recognize them quickly.
> 
> While I realize this is a problem beyond the scope of this particular PR, I'm wondering if we can set a good example - or at least clarify why they have different designations depending on the context.

i renamed the class to match the other prefill classes. i agree standardizing the naming in another PR would be a good idea. classes and variables in ruby have to start with a letter and classes have to be capitalized in a certain way in order for rails autoloading and eagerloading to work, so it's not possible to have the same naming for everything, but maybe the same standard conversion can be used for everything. `v#{form_code}.underscore` for the variable name and `v#{form_code}.underscore.camelize` for the class name.

PR #4194, opened 2020-04-27
update binstubs

jholton:
## Description of change
As a follow-on action to updating to Rails 6, I've regenerated binstubs.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#8343

- [x] excluded the `bin` directory form rubocop enforcement
- [x] regenerate binstubs
- [x] remove `bin/update` file.  It's been removed from Rails.


bin/rails
```diff
@@ -1,14 +1,10 @@
...
-
 begin
-  load File.expand_path('spring', __dir__)
+  load File.expand_path('../spring', __FILE__)
 rescue LoadError => e
   raise unless e.message.include?('spring')
 end
-
 APP_PATH = File.expand_path('../config/application', __dir__)
 $LOAD_PATH.unshift(File.expand_path('../lib', __dir__))
```
jholton, line 8: This line was manually added (or left as-is in this case).  The fact that this is necessary makes me think we should revisit our autoloading logic.

erluti:
How did you generate these?  I've never really tried to generate binstubs before and googled up `bundle install --binstubs` and got a whole bunch of new files created

jholton:
> How did you generate these? I've never really tried to generate binstubs before and googled up `bundle install --binstubs` and got a whole bunch of new files created

I generated them manually:
`bundle binstubs bundler`
`bundle binstubs rspec-core`
For Rails:
`rails app:update:bin`
`bundle binstubs spring`
Then
`bundle exec spring binstub --all` to add `spring` references in the other executables.  [source](https://github.com/rails/spring).

PR #4202, opened 2020-04-29
Add feature flag for sso auto-login/logout

hinzed1127:
## Description of change
This adds a feature flag specifically to toggle automatic establishment/disconnection of a `vets-api` session based on a user's SSOe session status. This is intended to be temporary so debugging this feature isn't a blocker to SSO as a whole.


ericbuckley:
@hinzed1127 would you want to just call it ssoe_inbound, so we include both auto login and auto logout functionality?

hinzed1127:
@ericbuckley good point, updated 👍

PR #4204, opened 2020-04-30
Separate Sidekiq Redis URL from caching/primary Redis URL

johnpaulashenfelter:
## Description of change

Our AWS architecture currently uses a single Elasticache Redis instance for both ephemeral
caching data and more mission-critical/persistent Sidekiq async jobs and scheduled tasks. The recommended configuration for Sidekiq includes a dedicated Redis instance (https://github.com/mperham/sidekiq/wiki/Using-Redis#multiple-redis-instances) which is exactly what this PR sets the stage to support.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#8149

By default, the settings file _uses the same Redis url for both caching and Sidekiq_ so on the surface, nothing has changed. Configuration of a 2nd instance requires changes to the `settings.yml` files created as part of the deployment process.

Developers can set up two local Redis instances by using a `settings.local.yml` file and using another local Redis database (eg /1) or potentially an entirely separate local Redis instance. 

```
redis_sidekiq:
  url: redis://localhost:6379/1
```

Docker containers will use the default settings (Sidekiq and caching both using the configured single Redis instance) but can be overridden in a similar way.

Tests continue to run fine because this is almost entirely a configuration change. Local testing showed jobs queued into the Redis instance specified by the URL, and could be verified through SidekiqUI or a Redis visualization tool (eg Redis-commander).

Once devops repos have been updated with new configuration, this will be tested on staging before a coordinated rollout on production.

### DIFF
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -3,7 +3,7 @@
 Sidekiq::Enterprise.unique! if Rails.env.production?
 
 Sidekiq.configure_server do |config|
-  config.redis = REDIS_CONFIG[:redis]
+  config.redis = REDIS_CONFIG[:sidekiq]
   # super_fetch! is only available in sidekiq-pro and will cause
   #   "undefined method `super_fetch!'"
   # for those using regular sidekiq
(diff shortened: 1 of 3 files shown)


johnpaulashenfelter:
Configuration for devops repo merged in https://github.com/department-of-veterans-affairs/devops/pull/6868

omgitsbillryan:
Is there / has there been something holding this back from merging?

PR #4207, opened 2020-04-30
VAOS - urgent More 404 routing issues

saneshark:
## Description of change
Found a few more 404 errors. This should be the last time this is an issue, because I took my own advice and introduced comprehensive specs for every single one of the 22 request types, and tested them both for /v0/vaos and /vaos/v0

Sorry for the grief on this one -- still need this urgently to avoid blowing up production.

All 44 specs are passing. I'm satisfied. 

## Original issue(s)
https://github.com/department-of-veterans-affairs/vets-api/pull/4130/files

Nothing not already described above.

### DIFF
--- a/modules/vaos/spec/routing/vaos_routing_spec.rb
+++ b/modules/vaos/spec/routing/vaos_routing_spec.rb
@@ -3,6 +3,30 @@
 require 'rails_helper'
 
 RSpec.describe 'routes for Session', type: :routing do
+  it 'routes to the appointments index' do
+    expect(get('/v0/vaos/appointments')).to route_to(
+      format: :json,
+      controller: 'vaos/v0/appointments',
+      action: 'index'
+    )
+  end
+
+  it 'routes to the appointments create' do
+    expect(post('/v0/vaos/appointments')).to route_to(
+      format: :json,
+      controller: 'vaos/v0/appointments',
+      action: 'create'
+    )
+  end
+
+  it 'routes to the appointments cancel' do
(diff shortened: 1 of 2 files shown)


modules/vaos/spec/routing/vaos_routing_spec.rb
```diff
@@ -2,83 +2,196 @@
...
-  it 'routes to the appointment requests index' do
-    expect(get('/v0/vaos/appointment_requests')).to route_to(
-      format: :json,
-      controller: 'vaos/v0/appointment_requests',
-      action: 'index'
-    )
-  end
+RSpec.describe 'VAOS routing configuration', type: :routing do
+  # Temporarily supporting '/v0/vaos' but also '/vaos/v0'
+  ['/v0/vaos', '/vaos/v0'].each do |partial_route|
```
erluti, line 7: I think it might be less complex and easier to pull out later if you just copied this file and renamed it something like `v0_vaos_routing_spec.rb`

erluti: I guess it could make maintenance a pain, how long will these temp routes be around?

saneshark: not long, no more than 1-2 weeks

saneshark: I do want to stress a little bit of urgency on this one which is making me nervous -- we have to get this in before the production deploy -- that or revert the last two PRs that have since been merged. Sorry for not testing these properly.

PR #4210, opened 2020-05-01
duplicate redis data to secondary redis

omgitsbillryan:
## Description of change
Duplicate *most of our redis data to a secondary redis instance running in parallel. 

The data _not_ duplicated is:
- breakers
- sidekiq data
- rack request rate limiting

### Additional Context

Redis data will be duplicated in real time. The idea is to let this run in code as long as our longest ttl data ([30 days according to redis.yml](2592000)), at which time we can switch the redis server from secondary to primary. User data is what we're most concerned with, which is why it's safer/easier to let the "utility" kinds of data (like breakers), expire and start regenerate itself.

One bonus is that user data also gets separate from sidekiq data - a side NTH we've wanted for a while.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/6865

The devops PR to update `Settings.yml` for the envs: https://github.com/department-of-veterans-affairs/devops/pull/6854

## Testing
I've tested this locally by running 2 Redis instances in docker - versions `3.2.4` & `5.0.5`. 

image


omgitsbillryan:
I've also audited our usage of Redis in vets-api 
https://dsva.slack.com/archives/C0MQ281DJ/p1588268094116400

I found our complete list of redis commands to be: `[get, set, del, expire, keys, exists, ttl]`

config/settings.yml
```diff
@@ -358,6 +358,7 @@ claims_api:
 redis:
   host: localhost
   port: 6379
```
johnpaulashenfelter, line 360: Any objection to changes this to a single URL 

```
redis:
  url: localhost:6379
```

and the corresponding changes in deployment files? It can be a lot easier to work with multiple Redis instances in development particularly if you can use the /0../15 syntax to access additional databases.

That would let you do something like

```
redis:
  url: localhost:6379

redis_secondary
  url: localhost:6379/1

redis_sidekiq:
  url: localhost:6379/2

etc
```

Means you could also just cut/paste Elasticache URLs directly instead of having to break it into URL and port :)

johnpaulashenfelter: FWIW I'm already using `url` in the PR for separating Redis out but thought updating here made sense to keep them a little more in sync. Happy to help w/ devops PRs if that's a help

config/settings.yml
```diff
@@ -358,6 +358,7 @@ claims_api:
...
+  duplicate_to_secondary: false
```
johnpaulashenfelter, line 361: Is having the 2nd set of URLs (`secondary_redis`) just as effective as the flag? Seems like its one less thing to worry about if you look for the config instead of the flag _and_ the config.

config/initializers/01_redis.rb
```diff
@@ -4,4 +4,32 @@
...
+    super
+  end
+
+  def expire(key, seconds)
+    @secondary_redis.expire(key, seconds)
+    super
+  end
+end
+
+Redis.current = if Settings.redis&.duplicate_to_secondary
```
johnpaulashenfelter, line 30: Can this be `if Settings&.redis_secondary` or similar?

johnpaulashenfelter:
@omgitsbillryan do you think anyone from lighthouse might have some insight on some of this data /cc @charleystran @bastosmichael

config/initializers/rack_attack.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class Rack::Attack
-  Rack::Attack.cache.store = Rack::Attack::StoreProxy::RedisStoreProxy.new(Redis.current)
+  Rack::Attack.cache.store = Rack::Attack::StoreProxy::RedisStoreProxy.new(Redis.new(REDIS_CONFIG[:redis]))
```
omgitsbillryan, line 4: This data will regenerate itself, so it's not worth the trouble to migrate it in real time. Instead, we can just flip it over to the new Redis whenever we want and let the old data die.

config/settings.yml
omgitsbillryan: Opened the ops PR here: https://github.com/department-of-veterans-affairs/devops/pull/6877

config/settings.yml
omgitsbillryan: Seems 👍 to me. My only preference would be that we do this work in a separate PR to minimize the moving parts.

omgitsbillryan: Whipped up a draft PR here: https://github.com/department-of-veterans-affairs/vets-api/pull/4219

config/initializers/rack_attack.rb
johnpaulashenfelter: Agreed 👍

config/settings.yml
omgitsbillryan: ☝️ PR merged 👍

config/initializers/rack_attack.rb
erluti: What's the trouble of migrating it?  When `RedisDuplicator` is pulled out, we'll want these to be `Redis.current` again.  So if it's just the additional monetary cost of unnecessary duplication, can we just let it ride until we're ready to remove `RedisDuplicator`?  I like using a hack to duplicate redis, but I don't like making two places that need to also be reverted.

omgitsbillryan: In `RedisDuplicator` we're only overriding `set`, `del`, & `expire` because that's all we use in `vets-api` to "save-and-remove" data in Redis. OTOH, `breakers`, uses `[zrem, zadd, incr, expire]`. OTOOH, I haven't yet bothered looking into which Redis calls `Rack::Attack::StoreProxy::RedisStoreProxy` uses.

I could try to encompass breakers and rack attack stuff, but I think that `RedisDuplicator` would get more complex, when really we care about user data a lot more than the utility data of breakers and rack attack.

omgitsbillryan: Didn't want to leave rack attack using `Redis.current` with the way `RedisDuplicator` is coded right now, since rack attack might be using `del`, `set,` and/or `expire`, which would leave us with some partial data state.

erluti: That all makes sense.  If the "remove duplicator" ticket (whatever that is) could have a note about these places that aren't using `Redis.current`, that'd be nice.  Thanks for answering

omgitsbillryan: Donezo - https://github.com/department-of-veterans-affairs/va.gov-team/issues/6865#issuecomment-626831951

PR #4211, opened 2020-05-01
VAOS FHIR operation outcome model and serializer

kreek:
## Description of change
FHIR error responses are wrapped in [operation outcomes](https://www.hl7.org/fhir/DSTU2/operationoutcome.html) this PR adds an operation outcome model and serializes it with different exception types within it's issue field. e.g.

```json
{
    "resourceType": "Organization",
    "id": "987654",
    "text": {
        "status": "generated",
        "div": "<div xmlns=\"http://www.w3.org/1999/xhtml\"><p>Received an an invalid response from the upstream server</p></div>"
    },
    "issue": [
        {
            "severity": "error",
            "code": "VAOS_502",
            "details": {
                "text": "Received an an invalid response from the upstream server"
            },
            "diagnostics": "Klass"
        }
    ]
}
```

The http error code in the response header will be the same level as the code in the issue. 'VOAS_502' => 502.

## Original issue(s)
https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/7828


modules/vaos/app/serializers/vaos/v1/operation_outcome_serializer.rb
```diff
@@ -0,0 +1,73 @@
...
+          serialize_information(issue)
+        end
+      end
+
+      def serialize_base_error(issue)
+        issue.errors.map do |error|
+          {
+            severity: 'error',
+            code: error.code,
+            details: error.detail,
```
jbalboni, line 48: Isn't details meant to be an object? http://hl7.org/fhir/DSTU2/operationoutcome.html#resource says it's a CodeableConcept.

modules/vaos/app/serializers/vaos/v1/operation_outcome_serializer.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+module VAOS
+  module V1
+    # http://hl7.org/fhir/DSTU2/operationoutcome.html
```
johnpaulashenfelter, line 5: THIS IS SUPER HELPFUL, thank you

modules/vaos/spec/serializers/v1/operation_outcome_serializer_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+require_relative '../../support/fixture_helper'
+
+describe VAOS::V1::OperationOutcomeSerializer do
+  describe 'json serialization' do
+    let(:resource_type) { :Organization }
+    let(:id) { '987654' }
+
+    context 'with a resource of Organization and a backend service exception' do
+      let(:issue) { Common::Exceptions::BackendServiceException.new('VAOS_502', source: 'Klass') }
+      let(:operation_outcome) { VAOS::V1::OperationOutcome.new(resource_type: resource_type, id: id, issue: issue) }
```
johnpaulashenfelter, line 13: These `let` statements are really the setup for the single example you follow with. Personally i think its clearer when the setup is in the example, but at least this is consistent in the spec.

modules/vaos/spec/serializers/v1/operation_outcome_serializer_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+        )
+      end
+    end
+
+    context 'with an informational message' do
+      let(:issue) do
+        {
+          detail: 'Additional information may be found in the conformance doc'
+        }
+      end
```
johnpaulashenfelter, line 38: Seems like this could be a single line. Did it start out as more and just get reduced to this?

saneshark:
@kreek this looks great to me, but how are we proposing breaking out execution to raise one of these types of errors? Will it all be serialized at the controller level and we raise standard exceptions and override the ApplicationController#rescue_from block?

kreek:
@saneshark we'll override one part of the ApplicationController rescue_from block https://github.com/department-of-veterans-affairs/vets-api/pull/4225

PR #4218, opened 2020-05-01
[8509] Add "GET /v0/debts" Endpoint

camerontesterman:
## Description of change
Adds the `GET /v0/debts` endpoint, which uses the debts service to show a veteran their debt details and history.

This endpoint uses the `debts` service which integrates with the VA Debt Management Center's Debt API.

## Original issue(s)
[department-of-veterans-affairs/va.gov-team#8509](https://github.com/department-of-veterans-affairs/va.gov-team/issues/8509)


app/controllers/v0/debts_controller.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module V0
+  class DebtsController < ApplicationController
+    def index
+      render json: service.get_letters(fileNumber: @current_user.ssn)
```
LindseySaari, line 6: Any chance ssn could be blank? + a spec for this case if it's applicable

PR #4219, opened 2020-05-04
refactor how we configure redis

omgitsbillryan:
## Description of change
Changes how we consume redis config from the Settings.yml file. This PR also requires that we update the relevant dev, staging, & prod settings.yml files in the devops repo.

EDIT:
Do not merge this PR until the ops PR is merged: https://github.com/department-of-veterans-affairs/devops/pull/6868

EDIT #2 (5-5-2020): ☝️ that PR is merged, this is good to go 👍 

## Original issue(s)
None.

## Testing Performed
Tested the connection in staging using the new `Redis.new(url: URL)` syntax:

image

This work spawned off PR comments thread here: https://github.com/department-of-veterans-affairs/vets-api/pull/4210#discussion_r418670148

### DIFF
--- a/config/redis.yml
+++ b/config/redis.yml
@@ -1,6 +1,6 @@
 development: &defaults
   redis:
-    url: <%= Settings.redis_app_data.url %>
+    url: <%= Settings.redis.app_data.url %>
   # DO NOT CHANGE BELOW TTL (We have agreement with MHV on this for SSO)
   session_store:
     namespace: vets-api-session
--- a/config/redis.yml
+++ b/config/redis.yml
@@ -1,7 +1,4 @@
 development: &defaults
-  redis:
-    host: <%= Settings.redis.host %>
-    port: <%= Settings.redis.port %>
   # DO NOT CHANGE BELOW TTL (We have agreement with MHV on this for SSO)
   session_store:
     namespace: vets-api-session
--- a/config/redis.yml
+++ b/config/redis.yml
@@ -1,6 +1,6 @@
 development: &defaults
   redis:
-    url: <%= Settings.redis_url %>
+    url: <%= Settings.redis_app_data.url %>
   # DO NOT CHANGE BELOW TTL (We have agreement with MHV on this for SSO)
   session_store:
     namespace: vets-api-session
--- a/config/redis.yml
+++ b/config/redis.yml
@@ -1,4 +1,6 @@
 development: &defaults
+  redis:
+    url: <%= Settings.redis_url %>
   # DO NOT CHANGE BELOW TTL (We have agreement with MHV on this for SSO)
   session_store:
     namespace: vets-api-session
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -353,7 +353,8 @@ claims_api:
   disability_claims_mock_override: false
   schema_dir: config/schemas
 
-redis_url: 'redis://localhost:6379'
+redis_app_data:
+  url: redis://localhost:6379
 
 # Settings for GovDelivery (email delivery)
 govdelivery:
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -353,11 +353,7 @@ claims_api:
   disability_claims_mock_override: false
   schema_dir: config/schemas
 
-# Settings for Redis
-# TODO(knkski): Move all redis settings here?
-redis:
-  host: localhost
-  port: 6379
+redis_url: 'redis://localhost:6379'
 
 # Settings for GovDelivery (email delivery)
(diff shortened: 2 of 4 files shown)


config/settings.yml
```diff
@@ -353,11 +353,7 @@ claims_api:
   disability_claims_mock_override: false
   schema_dir: config/schemas
 
-# Settings for Redis
-# TODO(knkski): Move all redis settings here?
-redis:
-  host: localhost
-  port: 6379
+redis_url: 'redis://localhost:6379'
```
johnpaulashenfelter, line 356: This needs corresponding PRs in devops for the 4 environments as well, right @omgitsbillryan

johnpaulashenfelter: Saw it in the description -- my bad

config/redis.yml
```diff
@@ -1,7 +1,4 @@
 development: &defaults
-  redis:
-    host: <%= Settings.redis.host %>
-    port: <%= Settings.redis.port %>
```
johnpaulashenfelter, line 4: I'm not adverse to leaving the redis configuration(s) in the `redis.yml` file, but I'm also fine moving it all to the single value in the settings file like this PR does.

config/settings.yml
omgitsbillryan: Yep. Probably even 2.

First, to add `redis_url`. Then once, dev, staging, & prod have all deployed with the `vets-api` code in this PR, a 2nd PR to devops that deletes the old setting.

omgitsbillryan: Oh! This will also affect anyone with a review instance who decides to merge the latest master of `vets-api` into their branch 🤔 ... I suppose it's probably ok to make an announcement in #vfs-platform-support

config/redis.yml
omgitsbillryan: Good point. I think I'll put this back, just to minimize the change.

johnpaulashenfelter: If you replace `host` and `port` with `url` from the settings.redis_url, then you can still use `REDIS_CONFIG[:redis]`.

config/settings.yml
```diff
@@ -353,11 +353,7 @@ claims_api:
   disability_claims_mock_override: false
   schema_dir: config/schemas
 
-# Settings for Redis
-# TODO(knkski): Move all redis settings here?
-redis:
-  host: localhost
-  port: 6379
+redis_url: 'redis://localhost:6378'
```
johnpaulashenfelter, line 356: 6379 right?

omgitsbillryan:
I think it failed on a flapping spec:
```
12:03:31  Failed examples:
12:03:31  
12:03:31  rspec ./modules/vba_documents/spec/request/v0/uploads_request_spec.rb:118 # VBA Document Uploads Endpoint #download /v0/uploads/{id} returns a 200 with content-type of zip
```

I re-kicked it off.

omgitsbillryan:
Do not merge this PR until https://github.com/department-of-veterans-affairs/devops/pull/6868 is merged

omgitsbillryan:
This PR is now good to go 👍  since the relevant devops PR with settings has been merged:

https://github.com/department-of-veterans-affairs/devops/commit/a20f55ca049e60b98175a65360bc9efe077b9a56

config/settings.yml
```diff
@@ -353,11 +353,14 @@ claims_api:
...
   schema_dir: config/schemas
 
-# Settings for Redis
-# TODO(knkski): Move all redis settings here?
 redis:
   host: localhost
   port: 6379
+  app_data:
+    url: redis://localhost:6379
+    # secondary_url: localhost:6378
```
johnpaulashenfelter, line 361: Still has the wrong port number, right? and needs the `redis://`

omgitsbillryan: Updated the `redis://` protocol prefix.

The host and port aren't "wrong" per-se (since it's commented out), and I dont think this should ever be defined in `settings.yml` since it's overidden in the various prod, staging, dev `settings.local.yml` files.

omgitsbillryan: I'm open to removing it altogether.

I left it in because of how we removed the enabled/disabled flag. W/o that flag and/or this comment, a local developer would have no way of knowing that they can define `redis.app_data.secondary_url` to enable redis duplication.

omgitsbillryan: Could also update the value in the comment to something very obviously wrong, like:
```YAML
secondary_url: redis://fake_secondary_redis:9999
```

I've followed this pattern in the past since it makes things very clear & searchable should a developer accidentally run into it.

PR #4224, opened 2020-05-05
[Lighthouse Forms] Update migration script to account for removed forms

ncksllvn:
## Description of change

The Lighthouse service VA Forms generates a local DB via a `form_reloader` script that scrapes va.gov/vaforms/ to build a local DB. However, sometimes forms are deleted entirely off of va.gov/vaforms/. Currently, this change isn't reflected at all in the Lighthouse Forms DB. There isn't a "deleted" flag. However, there is a "valid_pdf" flag. This PR makes use of that by flipping the "valid_pdf" to false for all forms that are in the internal DB but not found during the latest scrape. It also exposes `valid_pdf` for each form model in `/services/va_forms/v0/forms` (Lighthouse) and `/v0/forms` (Vets-API standard route). During the next scrape, this should set about 50-100 forms to `valid_pdf: false` because a bunch of forms were removed off of `/vaforms` recently.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/8647
https://github.com/department-of-veterans-affairs/va.gov-team/issues/8648

- This change is specifically visible when the build job that runs `form_reloader` is executed

- Will need to make sure I didn't break anything in Dev/Staging
- This should set `valid_pdf` to false for about 50-100 forms that were removed off of va.gov/vaforms a couple weeks ago

### DIFF
--- a/modules/va_forms/app/workers/va_forms/form_reloader.rb
+++ b/modules/va_forms/app/workers/va_forms/form_reloader.rb
@@ -20,9 +20,11 @@ module VaForms
     def mark_stale_forms
       processed_form_names = @processed_forms.map { |f| f['form_name'] }
       missing_forms = VaForms::Form.where.not(form_name: processed_form_names)
-      # rubocop:disable Rails/SkipsModelValidations
-      missing_forms.update_all(valid_pdf: false)
-      # rubocop:enable Rails/SkipsModelValidations
+      missing_forms.find_each do |form|
+        form.update(valid_pdf: false)
+      rescue
+        Rails.logger.warn "VA Forms failed to mark form as invalid: #{form.form_name}"
+      end
     end
 
     def load_page(current_page: 0)
--- a/modules/va_forms/app/workers/va_forms/form_reloader.rb
+++ b/modules/va_forms/app/workers/va_forms/form_reloader.rb
@@ -22,6 +22,7 @@ module VaForms
       missing_forms = VaForms::Form.where.not(form_name: processed_form_names)
       # rubocop:disable Rails/SkipsModelValidations
       missing_forms.update_all(valid_pdf: false)
+      # rubocop:enable Rails/SkipsModelValidations
     end
 
     def load_page(current_page: 0)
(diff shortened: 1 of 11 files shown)


modules/va_forms/app/workers/va_forms/form_reloader.rb
```diff
@@ -8,8 +8,20 @@ class FormReloader
...
     def perform
       load_page(current_page: 0)
+      mark_stale_forms
+    end
+
+    def mark_stale_forms
+      processed_form_names = @processed_forms.map { |f| f['form_name'] }
+      missing_forms = VaForms::Form.where.not(form_name: processed_form_names)
+      # rubocop:disable Rails/SkipsModelValidations
+      missing_forms.update_all(valid_pdf: false)
```
ncksllvn, line 24: I need some help on this, because I know disabling rubocop is probably not the best thing to do. Basically, as this line suggests, I just need to update all of these forms to `valid_pdf: false`. If I loop through `missing_forms`, then do -

```
form.valid_pdf = false
form.save
``` 

The save fails. It seems to be failing validation, but I just pulled the forms during the `Form.where.not(`. So, why would that fail validation?

bastosmichael: Unfortunately update_all is way faster because it doesn't instantiate an active record object for each record and instead deals directly with the database. In your case, since you need validations and callbacks, you'll need to instantiate the objects and so you're best bet is iterating in batches of 1000 and performing the update as originally shown. Such as:

`VaForms::Form.where.not(form_name: processed_form_names).find_each { |u| u.update(valid_pdf: false) }`

ncksllvn:
@bastosmichael I was gonna rely on `update_sha256` to always set the `valid_pdf` value for any of the new forms. Do you think I should default `valid_pdf` in `parse_form_row`?

bastosmichael:
I think that's fine, given that the state of creating a form means that it's valid at the moment it's created and only becomes invalid at subsequent checks and updates later down the pipeline.

modules/va_forms/app/workers/va_forms/form_reloader.rb
ncksllvn: Thanks @bastosmichael! Updated ✅

PR #4227, opened 2020-05-06
Creating base64 tempfile for Claims API

bastosmichael:
## Description of change
Accepts base64 string attachment and decodes into PDF.

## Original issue(s)
https://vasdvp.atlassian.net/browse/DASH-196


modules/claims_api/app/controllers/claims_api/base_form_controller.rb
```diff
@@ -47,7 +47,28 @@ def auth_headers
...
+        if document.is_a?(String)
+          decode_document(document)
+        else
+          document
+        end
+      end.compact
+    end
+
+    def decode_document(document)
+      match = document.match(/\;base64\,(.+?)$/)
```
charleystran, line 60: Should we be doing a match here? this could be very slow on large files. If we dont want to do a start with, maybe we could take the first x chars and do the match on that vs the whole document

bastosmichael: Good point, You will not see any commas, colons, or double quotes in a Base64 encoded string so maybe we do a document.split(',').last and it should give us the same base64 string we need.

modules/claims_api/app/controllers/claims_api/base_form_controller.rb
```diff
@@ -47,7 +47,28 @@ def auth_headers
...
+      if match
+        base64 = match.captures.first
+        decoded_data = Base64.decode64(base64)
+        filename = "temp_upload_#{Time.zone.now.to_i}.pdf"
+        temp_file = Tempfile.new(filename, encoding: 'ASCII-8BIT')
+        temp_file.write(decoded_data)
+        temp_file.close
+        ActionDispatch::Http::UploadedFile.new(filename: filename,
+                                               type: 'application/pdf',
+                                               tempfile: temp_file)
```
edmangimelli, line 70: do we need to worry about cleaning up these temp files?

modules/claims_api/app/controllers/claims_api/base_form_controller.rb
edmangimelli: i like the _taking the first x chars to check if it's base64_ approach 👍

modules/claims_api/app/controllers/claims_api/base_form_controller.rb
bastosmichael: The ActionDispatch::Http::UploadedFile does eventually hit a .close state when the controller finishes the process the same way it would if it were just an uploaded binary file which basically closes out the temp file as well. https://api.rubyonrails.org/v6.0.2.1/classes/ActionDispatch/Http/UploadedFile.html#method-i-close

modules/claims_api/app/controllers/claims_api/base_form_controller.rb
bastosmichael: Done

modules/claims_api/app/controllers/claims_api/base_form_controller.rb
edmangimelli: i'm thinking more of an `unlink` (which will delete the file, rather than just close it)

edmangimelli: oh nm.
>Uploaded files are temporary files whose lifespan is one request. When the object is finalized Ruby unlinks the file, so there is no need to clean them with a separate maintenance task.

bastosmichael: :+1:

PR #4230, opened 2020-05-06
Expose redis port for Docker development

johnpaulashenfelter:
## Description of change

Our Docker container for development exports the Postgresql port on a non-standard local port so it can be accessed during development. We do not do the same for Redis until now.

This PR exposes Redis using a similar manner to what we currently do with Postgresql -- adding a zero to the standard port number `6379` + `0` => `63790`
  
## Original issue(s)
None, but noticed as part of working on #4229

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,6 +2,8 @@ version: '3.4'
 services:
   redis:
     image: redis:5.0-alpine
+    ports:
+      - "63790:6379"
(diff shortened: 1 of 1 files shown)


johnpaulashenfelter:
Don't know if anyone will use this, but at least its now consistent.

Will make testing the redis container a bit easier for sure, with all the upcoming redis work.

PR #4234, opened 2020-05-06
Cleanup Rails framework defaults files for Rails 6

johnpaulashenfelter:
## Description of change

We have two files containing overrides for defaults in Rails to help with migrations, one for Rails 5 and one for Rails 6. We are loading the defaults for Rails 5.2 and using `new_framework_defaults_6.0.rb` to override with the old values as is customary for an upgrade.

This PR is beginning to pick apart those configuration to find which are still necessary

- [x] Start by loading defaults for 5.2
- [x] Work through old (rails <6) defaults in `new_framwork_defaults.rb`
- [x] Remove Rails6 defaults for components that are unused, active storage, active job, actioncable, etc
- [ ] ActionDispatch issues that need some investigation

## Original issue(s)
department-of-veterans-affairs/va.gov-team#131


erluti:
I hadn't made a PR until now, but I've this work here https://github.com/department-of-veterans-affairs/vets-api/pull/4128

johnpaulashenfelter:
Closed in favor of #4250

PR #4235, opened 2020-05-06
Handles and validates multi-value fields in SSOe

patrickvinograd:
SSOe may provide comma-separated multivalue attributes for some fields.
This makes the parsing code robust to that, and also enforces validation
on the MHV ID, EDIPI, and ICN fields to ensure that there are no
conflicting values.

## Description of change
- Parses comma-separated values in the mhvien and dodedipnid attributes from SSOe SAML
- Validates that there is one and only one value present for MHV ID, EDIPI, and ICN across all the fields where those values may appear.
- Prevents login if above is not true.
- Adds new "auth failed" codes that can be consumed by frontend (TBD) to present a more specific message to the user. Content for those messages will be borrowed from the current MHV-specific errors when multiple MHV IDs are present, and then submitted to appropriate website/content teams for review.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#8820

Added robust specs; I do have a staging user who I can use to verify this against originally-reported issue but that user doesn't work against localhost (because not present in MHV/SSOe environments other than staging).

### DIFF
--- a/app/controllers/v1/sessions_controller.rb
+++ b/app/controllers/v1/sessions_controller.rb
@@ -55,6 +55,10 @@ module V1
         redirect_to url_service.login_redirect_url(auth: 'fail', code: auth_error_code(saml_response.error_code))
         callback_stats(:failure, saml_response, saml_response.error_instrumentation_code)
       end
+    rescue SAML::UserAttributeError => e
+      log_message_to_sentry(e.message, {}, {}, :warning)
+      redirect_to url_service.login_redirect_url(auth: 'fail', code: e.code)
+      callback_stats(:failure, saml_response, e.tag)
     rescue => e
       log_exception_to_sentry(e, {}, {}, :error)
       redirect_to url_service.login_redirect_url(auth: 'fail', code: '007') unless performed?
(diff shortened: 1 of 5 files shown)


app/controllers/v1/sessions_controller.rb
```diff
@@ -55,6 +55,10 @@ def saml_callback
         redirect_to url_service.login_redirect_url(auth: 'fail', code: auth_error_code(saml_response.error_code))
         callback_stats(:failure, saml_response, saml_response.error_instrumentation_code)
       end
+    rescue SAML::UserAttributeError => e
+      log_message_to_sentry(e.message, {}, {}, :warning)
+      redirect_to url_service.login_redirect_url(auth: 'fail', code: e.code)
```
johnpaulashenfelter, line 60: I think you want `log_message_to_sentry(e.message, :warning, {}, {})`

Apparently the method signature for log_message and log_exception are different :/

johnpaulashenfelter: We also might want to add a tag here. Checking w/ our sentry team

patrickvinograd: Good catch, thank you. Fixed in latest commit.

johnpaulashenfelter: Awesome. I'll be working to fix those methods because that kind of surprise shouldn't happen.

PR #4238, opened 2020-05-07
Decision Reviews API - Add Sidekiq Job for Updating HigherLevelReview Statuses Based on Central Mail Statuses

edmangimelli:
resolves https://vajira.max.gov/browse/API-747

Periodically poll Central Mail for the status of HigherLevelReviews that have been uploaded (based off of how VBA Documents does this)

Acceptance Criteria:
- [x] make sidekiq jobs
- [x] tests


modules/appeals_api/app/models/appeals_api/higher_level_review.rb
```diff
@@ -154,6 +197,8 @@ def contestable_issues
       form_data&.dig('included')
     end
 
+    ###########
```
johnpaulashenfelter, line 200: Artifact of a merge?

modules/appeals_api/app/models/appeals_api/higher_level_review.rb
```diff
@@ -17,7 +56,11 @@ def past?(date)
     attr_encrypted(:form_data, key: Settings.db_encryption_key, marshal: true, marshaler: JsonMarshal::Marshaller)
     attr_encrypted(:auth_headers, key: Settings.db_encryption_key, marshal: true, marshaler: JsonMarshal::Marshaller)
 
-    enum status: { pending: 0, processing: 1, submitted: 2, established: 3, error: 4 }
+    enum status: {
+      pending: 0, submitted: 1, processing: 2, error: 3, uploaded: 4, received: 5, success: 6, vbms: 7, expired: 8
```
bastosmichael, line 60: So I just now realized that we are doing numerical enums here for status instead of actually writing the status in the model, might be too late to change it but I've found Benefits Intake and Claims where we need to query the model itself and do analysis on the data without being able to reference the model itself for a particular state every time. Perfect example of this is how we do the unsuccessful report mailer job and how storing the physical status as an enum option instead of storing the physical string status itself https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/vba_documents/app/workers/vba_documents/report_unsuccessful_submissions.rb#L23 I understand all of the data storage arguments that can be made for storing a numerical integer vs a searchable string but those kind of database related storage savings only apply when you expect your tables to have millions upon millions of rows (not something we are going to be facing here) and there's also a database and code readability issue to be concerned with here that future future data scientists may be pulling this data directly from the DB (like I had to do for Zach to analyize all Benefits Submissions over the last 2 years) irregardless of the model and that this kind of layered abstraction is unnecessary in today's more nosql duplication is okay approach to data storage.

edmangimelli: Yeah, I could see it be annoying if you were to query the DB _outside_ of Rails. It would be nicer if there were a separate DB table (HigherLevelReviewStatuses) over strings IMO, but strings are dead simple. I'm fine with switching this field. Mind if I do it in this PR?

bastosmichael: Please do it in this PR but keep it simple, both Claims and Benefits Intake have an approach you can take or you can create your own but I just want to make sure we prevent the abstraction that's all, I would prefer not leaving this for later and us potentially going live with something like this would make it much harder to migrate away from it later.

modules/appeals_api/app/workers/appeals_api/higher_level_review_pdf_submit_job.rb
```diff
@@ -19,25 +19,24 @@ def generate_pdf(higher_level_review_id)
...
+      higher_level_review.update!(status: 'processing')
       pdf_constructor.stamp_pdf(pdf_path, higher_level_review.consumer_name)
     end
 
     def upload_to_central_mail(higher_level_review_id, pdf_path)
       higher_level_review = AppealsApi::HigherLevelReview.find higher_level_review_id
       metadata = {
-        'veteranFirstName' => higher_level_review.auth_headers['first_name'],
-        'veteranLastName' => higher_level_review.auth_headers['last_name'],
-        'source' => higher_level_review.auth_headers['consumer_name'],
```
edmangimelli, line 31: i think this consumer line was a typo

bastosmichael: The line isn't a type as much as it's just referencing the wrong header here, the header might not be consumer_name or might be something different.
https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/appeals_api/app/controllers/appeals_api/v0/appeals_controller.rb#L46

```
'source' => higher_level_review.auth_headers['consumer_name'],
```

PR #4242, opened 2020-05-07
[Bugfix] For user attributes, first look at the identity provider, then MVI

annaswims:
## Description of change
We were troubleshooting an disability claim with a veteran and we were unable to retrieve a veteran's rated disability from evss because we were sending a nil value to evss for gender. The veteran had a gender in MVI, but gender was nil from the identity provider (id.me).  

https://github.com/department-of-veterans-affairs/vets-api/pull/1666 changed the order of presedence for the data. 

MHV gives us a limited subset of the attributes that the other identity providers so we were using `mhv_icn.present?` as a way to determine if the user logged in with MHV (and therefore we'd need to look up all the attributes in MVI). But now, there's no reason not to use the identity provider's data first, and MVI's data second (note: we only do MVI lookups for LOA3 users).
## Original issue(s)

https://github.com/department-of-veterans-affairs/va.gov-team/issues/8908

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -76,27 +76,27 @@ class User < Common::RedisStore
   end
 
   def middle_name
-    identity.middle_name || (mhv_icn.present? ? mvi&.profile&.given_names.to_a[1..-1]&.join(' ').presence : nil)
+    identity.middle_name.presence || mvi&.profile&.given_names.to_a[1..-1]&.join(' ').presence
   end
 
   def last_name
-    identity.last_name || (mhv_icn.present? ? mvi&.profile&.family_name : nil)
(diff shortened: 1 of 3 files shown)


annaswims:
we should probably do the same for middle name, but I don't think it matters nearly as much.  when I fall back to the middle name in MVI 30+ test fail because many of our identity mocks don't have middle names and our factory has a faker middle name

annaswims:
@patrickvinograd I don't think this affects you, but I figured I'll tag you just in case.

annaswims:
If you'd rather, I could just make the change for gender, but it made sense to me to make the change for all attributes.

kfrz:
LGTM. FWIW I certainly don't mind cleaning up related code in simple changesets.

PR #4245, opened 2020-05-08
user email is now optional

ericbuckley:
The prominent change is making email optional in the `UserIdentity` class, this will allow access for inbound SSOe users that do not have an email address on file.

There are a few cases in the vets-api where `user.email` is expected to always have a value.  To remedey this we've introduced a new `common_name` attribute that will be set to the `va_eauth_commonname` value when authenticated via SSOe and the user's email when not using SSOe.

In the cases where an email address is optional, we've updated the code to use `user.email` or an empty string.  In the cases where its required, we've found that the caller just needs a unique identifier, so we've updated those cases to use the new `common_name` attribute.

closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/8305

### DIFF
--- a/lib/evss/jwt.rb
+++ b/lib/evss/jwt.rb
@@ -30,7 +30,6 @@ module EVSS
 
         # custom claims
         assuranceLevel: @user.loa[:current],
-        email: @user.email,
         firstName: @user.first_name,
         middleName: @user.middle_name,
         lastName: @user.last_name,
@@ -39,7 +38,8 @@ module EVSS
         prefix: '', # TODO: i thought we had these in mvi...
         suffix: '',
         correlationIds: @user.mvi&.profile&.full_mvi_ids
-      }
+        # only include the email address if its not nil
+      }.merge({ email: @user.email }.compact)
     end
   end
 end
--- a/app/controllers/v0/post911_gi_bill_statuses_controller.rb
+++ b/app/controllers/v0/post911_gi_bill_statuses_controller.rb
@@ -37,7 +37,7 @@ module V0
         raise EVSS::GiBillStatus::ExternalServiceUnavailable
       when EVSS::GiBillStatus::GiBillStatusResponse::KNOWN_ERRORS[:vet_not_found]
         log_vet_not_found(@current_user, response.timestamp)
-        raise Common::Exceptions::RecordNotFound, @current_user.email
+        raise Common::Exceptions::RecordNotFound, @current_user.common_name
       when EVSS::GiBillStatus::GiBillStatusResponse::KNOWN_ERRORS[:invalid_auth]
         # 403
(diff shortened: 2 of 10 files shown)


annaswims:
I tink there are a lot of places that we're calling user.email, assuming that we'll get a value.  Would it make sense  for user.email to fall back to the email address in vets360/VA profile if we don't get one from the identity provider?

Samara-Strauss:
> I think there are a lot of places that we're calling user.email, assuming that we'll get a value. Would it make sense for user.email to fall back to the email address in vets360/VA profile if we don't get one from the identity provider? Maybe @Samara-Strauss would have thoughts on that.

I want to say yes to this, but I guess it depends on the reason/use case(s).

ericbuckley:
@annaswims this is still a work in progress, and I'll make sure to update the PR with a description before moving it to review, but here is the original issue that @patrickvinograd created to outline where in vets-website/vets-api the users email address was being used https://github.com/department-of-veterans-affairs/va.gov-team/issues/8305

patrickvinograd:
Missing the use in `/lib/evss/jwt.rb`. Guidance from EVSS team was to omit the `email` claim from the JWT rather than supplying an empty value.

ericbuckley:
> Missing the use in `/lib/evss/jwt.rb`. Guidance from EVSS team was to omit the `email` claim from the JWT rather than supplying an empty value.

As the code stands it should be submitting a null value, `{"email": null, ....}`.  But you're saying we need to remove the attribute altogether, correct?

kfrz:
This looks good as far as I can tell, only thing I didn't see addressed here is this use case:

> Form526OptInsController - accepts email as input (unknown if prefilled on frontend)

Did you verify this isn't prefilled on frontend? 

No explicit test cases need to be added?

ericbuckley:
> This looks good as far as I can tell, only thing I didn't see addressed here is this use case:
> 
> > Form526OptInsController - accepts email as input (unknown if prefilled on frontend)
> 
> Did you verify this isn't prefilled on frontend?
> 
> No explicit test cases need to be added?

@kfrz just confirmed that it the email in the form is initialized to the empty string and not any email value from the user's profile https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/applications/disability-benefits/526EZ/components/create526EmailForm.jsx#L16

ericbuckley:
@patrickvinograd do you want to see any more changes to this PR before I merge?

lib/evss/jwt.rb
```diff
@@ -39,7 +38,8 @@ def payload
         prefix: '', # TODO: i thought we had these in mvi...
         suffix: '',
         correlationIds: @user.mvi&.profile&.full_mvi_ids
-      }
+        # only include the email address if its not nil
+      }.merge({ email: @user.email }.compact)
```
patrickvinograd, line 42: 👍

PR #4248, opened 2020-05-08
enable enforcement of CSRF protection.

jholton:
## Description of change
enable enforcement of CSRF protection.  Rails will now raise `ActionController::InvalidAuthenticityToken` for CSRF token violations

## Original issue(s)
department-of-veterans-affairs/va.gov-team-sensitive#44
department-of-veterans-affairs/va.gov-team-sensitive#42
department-of-veterans-affairs/va.gov-team-sensitive#61


jholton:
> Mind writing up your findings when using `LoggingForgeryStrategy` just in case some poor soul finds themselves reading about this PR some number of months from now?

Good idea.  Before this PR, `LoggingForgeryStrategy` just logged CSRF violations to Sentry ([here](http://sentry.vfs.va.gov/vets-gov/platform-api-production/issues/113226/events/latest/)) without actually raising an error or taking action on a user's session.

This allowed us to identify endpoints that were being called without the FE providing a CSRF token and fix those calls.
It also allowed us to track down any less obvious scenarios (e.g. FE's first call to BE is a POST.  Thus no CSRF token exists yet.  Thus call fails) and remedy those.

We've gotten the number of violations per day down to a small amount. The ones that still exist look legitimate.

PR #4250, opened 2020-05-08
use rails 6.0 defaults

erluti:
Defaults for Rails 6.0 are enabled and working for our app.

https://app.zenhub.com/workspaces/vsp-5cedc9cce6e3335dc5a49fc4/issues/department-of-veterans-affairs/va.gov-team/8047


johnpaulashenfelter:
@erluti You should be able to use the configuration for Rails6 and disable zeitwork:

```
config.load_defaults "6.0"
config.autoloader = :classic
```

at least according to the docs (https://guides.rubyonrails.org/configuring.html#rails-general-configuration -- have to scroll down, can't link directly).

You could then clean up most of the frameworks default file like in #4234

spec/rakelib/support/cookies_spec.rb
```diff
@@ -10,19 +10,20 @@
   let(:session) { Session.create(uuid: user.uuid, token: 'abracadabra') }
 
   describe '#api_session_header' do
-    # much of this code comes from here: https://stackoverflow.com/a/51579296
+    # much of this code comes from here: https://gist.github.com/wildjcrt/6359713fa770d277927051fdeb30ebbf
     def decrypt_session_cookie(cookie)
-      salt = Rails.application.config.action_dispatch.authenticated_encrypted_cookie_salt
-      encrypted_cookie_cipher = 'aes-256-gcm'
+      config = Rails.application.config
+      salt   = config.action_dispatch.authenticated_encrypted_cookie_salt
```
johnpaulashenfelter, line 16: Gotta be honest I'd rather see one line here -- guessing this was for rubocop? I'd ignore it when it makes things HARDER to read

spec/rakelib/support/cookies_spec.rb
```diff
@@ -10,19 +10,20 @@
...
+      encrypted_cookie_cipher = config.action_dispatch.encrypted_cookie_cipher || 'aes-256-gcm'
 
-      key_generator = ActiveSupport::KeyGenerator.new(Rails.application.secrets.secret_key_base, iterations: 1000)
+      key_generator = ActiveSupport::KeyGenerator.new(Rails.application.secret_key_base, iterations: 1000)
```
johnpaulashenfelter, line 19: Did this key just move?

erluti: Apparently this is a Rails helper for finding the secret_key_base. 
 https://api.rubyonrails.org/classes/Rails/Application.html#method-i-secret_key_base

TLDR this method checks `ENV`, then the `credentials.yml`, and finally the old-style `secrets.yml`.  Byebug showed me these values were the same here

spec/rakelib/support/cookies_spec.rb
erluti: config is reused in a couple places, but the salt only is used once so I just moved this line to where it's used

PR #4252, opened 2020-05-11
[8659] MDOT Auth

camerontesterman:
## Description of change
This PR modifies the MDT service to handle authentication with the DLC service. During the initial `GET /supplies` the data that is returned by the DLC includes a header called `va_api_key` which is a JWT that is encoded with the DLC system patient id.

This JWT is stored within redis after being retrieved from the initial GET, and is later used during the POST when the veteran submits their order.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/8659

## Related Mockdata Updates
https://github.com/department-of-veterans-affairs/vets-api-mockdata/pull/132


lib/mdot/client.rb
```diff
@@ -51,16 +52,25 @@ def get_supplies
     #
     def submit_order(request_body)
       with_monitoring_and_error_handling do
-        perform(:post, @supplies, request_body, headers).body
+        perform(:post, @supplies, request_body, submission_headers).body
```
LindseySaari, line 55: Just to verify, this POST request only needs the VA_API_KEY token and no longer needs the headers referenced on line 61, correct?

camerontesterman: yes, that is correct.

lib/mdot/token.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module MDOT
+  class Token < Common::RedisStore
+    redis_store REDIS_CONFIG[:mdot][:namespace]
+    redis_ttl REDIS_CONFIG[:mdot][:each_ttl]
+    redis_key :uuid
+
```
LindseySaari, line 8: May want to add validation for the uuid "validates :uuid, presence: true"

camerontesterman: good suggestion - added

PR #4256, opened 2020-05-11
VAOS FHIR read organization

kreek:
## Description of change
Adds a controller for the Organization resource and implements the FHIR read (rails show) method. Also adds StatsD metrics to the FHIR Service.

## Original issue(s)
https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/9041

## Testing planned
Review instance and staging integration tests


saneshark:
This looks really really good. But I think we can remove schema validations. We are no longer really concerned with JSON API schemas and everything else is a straight pass through anyway. It follows the FHIR spec, why would we want or need to reverify? 

If we need or want schema validation in the future, I wonder if there is a gem we could leverage in the future. Writing out schemas for something that follows a very detailed and specific specification readily available at fhir.org seems like we're overtesting IMHO.

config/settings.yml
```diff
@@ -494,7 +494,7 @@ bgs:
   ssl_verify_mode: 'peer'
 
 va_mobile:
-  url: 'https://veteran.apps.va.gov/'
+  url: 'https://veteran.apps.va.gov'
```
kreek, line 497: @saneshark this is the trailing slash to causes local specs and the base path in fhir configuration out of sync. In staging and production the foward proxy does not have a trailing slash. The v0 VOAS services as they use this base url (not a base path, ie `https://veteran.apps.va.gov` vs `https://veteran.apps.va.gov/vsp/v1/`) don't suffer from the issue as Faraday strips extra slashes immediately after the domain name.

PR #4262, opened 2020-05-13
Upgrade danger gem

johnpaulashenfelter:
## Original issue(s)


johnpaulashenfelter:
@LindseySaari not really. v7 broken lots of things for BitBucket's new infrastructure. 8 moved on from there -- idk why that ended up going so far semantically.

PR #4265, opened 2020-05-13
properly label user and user_identity in mvi-related code

annaswims:
## Original issue(s)

### DIFF
--- a/app/models/mvi.rb
+++ b/app/models/mvi.rb
@@ -139,7 +139,7 @@ class Mvi < Common::RedisStore
 
   def response_from_redis_or_service
     do_cached_with(key: user.uuid) do
-      mvi_service.find_profile(user)
+      mvi_service.find_profile(user.identity)
     end
   end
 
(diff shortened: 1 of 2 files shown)


app/models/mvi.rb
```diff
@@ -139,7 +139,7 @@ def add_ids(response)
 
   def response_from_redis_or_service
     do_cached_with(key: user.uuid) do
-      mvi_service.find_profile(user)
+      mvi_service.find_profile(user.identity)
```
annaswims, line 142: find_profile takes a user_identity, but it happens to work because of all of the delegations that happen in user. 
Screen Shot 2020-05-13 at 4 59 20 PM

lib/mvi/service.rb
```diff
@@ -195,15 +195,15 @@ def message_edipi(user)
       MVI::Messages::FindProfileMessageEdipi.new(user.dslogon_edipi).to_xml
     end
 
-    def message_user_attributes(user)
```
annaswims, line 198: now that `find_profile` is being called with a `user_identity`, `create_profile` and  `message_user_attributes` will too.
Screen Shot 2020-05-13 at 5 01 59 PM

app/models/mvi.rb
annaswims: by using user_identity to look up things in mvi we'll make it possible to fallback to mvi values in user

PR #4266, opened 2020-05-13
Decision Reviews API - don't run batch job unless there are hlrs to work on

edmangimelli:


modules/appeals_api/app/workers/appeals_api/higher_level_review_upload_status_batch.rb
```diff
@@ -27,5 +27,9 @@ def higher_level_review_ids
     def slice_size
       (higher_level_review_ids.length / 5.0).ceil
     end
+
+    def enabled
+      Settings.decision_review.higher_level_review_updater_enabled
```
bastosmichael, line 32: Does this setting actually exist and I guess the better question is why does it exist?

edmangimelli: it does!
https://github.com/department-of-veterans-affairs/vets-api/blob/master/config/settings.yml#L139

edmangimelli: and the reason why it does is because this job was pretty much a copy of vba_document's job

PR #4275, opened 2020-05-14
Revise AppInfo to get ref from env

kfrz:
## Description of change

We used to have the current SHA of the release available in the initializer. This was broken when infra moved to Docker.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#9131

- Associated `devops` [pull request](https://github.com/department-of-veterans-affairs/devops/pull/6945)

### DIFF
--- a/config/initializers/app_info.rb
+++ b/config/initializers/app_info.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
 module AppInfo
-  GIT_REVISION = Env.fetch('GIT_REVISION', 'ABC1234')
+  GIT_REVISION = ENV.fetch('GIT_REVISION', 'MISSING_GIT_REVISION')
   GITHUB_URL   = 'https://github.com/department-of-veterans-affairs/vets-api'
 end
--- a/config/initializers/app_info.rb
+++ b/config/initializers/app_info.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
 module AppInfo
-  GIT_REVISION = `git rev-parse HEAD`&.chomp
+  GIT_REVISION = ENV['GIT_REVISION']
   GITHUB_URL   = 'https://github.com/department-of-veterans-affairs/vets-api'
 end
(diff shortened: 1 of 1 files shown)


config/initializers/app_info.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
 module AppInfo
-  GIT_REVISION = `git rev-parse HEAD`&.chomp
+  GIT_REVISION = ENV['GIT_REVISION']
```
johnpaulashenfelter, line 4: I think you might want to try it with a default `ENV.fetch('GIT_REVISION', 'MISSING_GIT_REVISION') or similar.

kfrz: I think as per discussion in Slack this is fine to expect a value here - if it's `nil`/blank that means the BRD cycle is problematic and likely this would be the least of our concerns

kfrz:
Not a lot is built on this code (nor likely will be). The only thing it's used for is setting the header `X-Git-Sha` and response for https://api.va.gov/v0/status

config/initializers/app_info.rb
johnpaulashenfelter: Thought about this some more -- its going to be blank in development and test too. Let's put some kind of default in.

kfrz: It won't be blank, lower envs are part of the build/release process, unless I'm misunderstanding what we're talking about when we say build/release.

The image ref tag *has* to be present, I think, else Ansible doesn't know what to build.

Anyways I added a default value feel free to change it if there's something more useful.

johnpaulashenfelter: At least until we don't have ansible 😄 

Updated.

johnpaulashenfelter:
@kfrz did we have to add `git` to the docker image to fix the original problem? Likely can remove it now.

config/initializers/app_info.rb
omgitsbillryan: NBD either way IMO, they a default val or `nil` will both work just fine.

PR #4277, opened 2020-05-14
Default flipper toggle to off in development

kfrz:
## Description of change

The current behavior is for features to be [enabled by default](https://github.com/department-of-veterans-affairs/vets-api/blob/master/config/initializers/flipper.rb) in development environments.

This is causing problems for developers, because the SSOe flag is defaulted to enabled but SSOe is not ready for developers to use locally. See [this DSVA Slack thread](https://dsva.slack.com/archives/CBU0KDSB1/p1589216158357600).

## Original issue(s)
department-of-veterans-affairs/va.gov-team#8519

Features are now still enabled by default in development but can be disabled by default in a given environment by adding a key to feature in `config/features.yml` like:

```yaml
feature_name
  default_enabled_environments:
    development: false
```

Features can also be toggled in development by visiting `/flipper`


config/initializers/flipper.rb
```diff
@@ -42,10 +42,12 @@
 Flipper::UI.configuration.feature_creation_enabled = false
 # Make sure that each feature we reference in code is present in the UI, as long as we have a Database already
 FLIPPER_FEATURE_CONFIG['features'].each_key do |feature|
+  feature_config = FLIPPER_FEATURE_CONFIG['features'][feature]
+
   unless Flipper.exist?(feature)
     Flipper.add(feature)
     # default features to enabled for development and test only
```
LindseySaari, line 49: Can you update the comment -- "default features to enabled for development"

annaswims:
This looks good to me, but I'd lean towards updating the config to default to "true" for all of the features that are enabled in staging.  That way a FE dev who occasionally starts vets-api sees behavior that's closer to what they'd expect.

kfrz:
@annaswims your comment implies a reversal of the logic from the original approach, now to individually configure each feature to match https://api.va.gov/flipper/features we'll need to add an override key:value to each feature in `config/features.yml`

I don't have the cycles to go through and manually update config to match production right now - we'll either have to merge the code as originally intended or not merge it at all right now. Either way requires a circle back on the ticket.

config/initializers/flipper.rb
```diff
@@ -42,10 +42,13 @@
 Flipper::UI.configuration.feature_creation_enabled = false
 # Make sure that each feature we reference in code is present in the UI, as long as we have a Database already
 FLIPPER_FEATURE_CONFIG['features'].each_key do |feature|
```
annaswims, line 44:
```suggestion
FLIPPER_FEATURE_CONFIG['features'].each do |feature, feature_config|
```
if you change this line you can get rid of line 45

annaswims:
I suspect tests will pass after master is merged and the CVE is ignored.

config/initializers/flipper.rb
```diff
@@ -42,10 +42,13 @@
...
-  unless Flipper.exist?(feature)
-    Flipper.add(feature)
-    # default features to enabled for development and test only
-    Flipper.enable(feature) if Rails.env.development? || Rails.env.test?
+  feature_config = FLIPPER_FEATURE_CONFIG['features'][feature]
+
+  Flipper.add(feature) unless Flipper.exist?(feature)
+
+  # default features to be enabled in development and test unless override set in config/features.yml
+  if feature_config.dig('default_enabled_environments', Rails.env) && Rails.env.in?('development', 'test')
```
erluti, line 50: it looks like this was previously enabling all of the features for test, but with no keys for test in `features.yml`, test will not enable any of these.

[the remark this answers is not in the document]
erluti: Or the key could be more explicit, like `dev_disabled_by_default` and it could be a bit more like the original method:
```
Flipper.enable(feature) if (Rails.env.development? && !feature_config['dev_disabled_by_default']) || Rails.env.test?
```

config/features.yml
```diff
@@ -1,107 +1,158 @@
...
 features:
   dashboard_show_covid19_alert:
     actor_type: user
+    default_enabled_environments:
     description: >
       Allows for showing a special COVID19 alert on the My VA Dashboard
   facility_locator_show_community_cares:
     actor_type: user
+    default_enabled_environments:
+      development: true
```
mpelzsherman, line 20: We are no longer using `facility_locator_show_community_cares`.

Can you please replace it with this?

```yml
  facilities_ppms_suppress_community_care:
    actor_type: user
    description: Hide ppms community care searches
    default_enabled_environments:
      development: false
```

mpelzsherman:
Any ETA on this? It would save us some time flipping off features every time we do a build!

kfrz (commented):
I think this accomplishes the goals - @mpelzsherman @annaswims ready for final review to make sure it meets your expectations

annaswims (approved):
happy

 YES!

config/initializers/flipper.rb
```diff
@@ -41,12 +41,11 @@
...
-  unless Flipper.exist?(feature)
-    Flipper.add(feature)
-    # default features to enabled for development and test only
-    Flipper.enable(feature) if Rails.env.development? || Rails.env.test?
-  end
+FLIPPER_FEATURE_CONFIG['features'].each do |feature, feature_config|
+  Flipper.add(feature) unless Flipper.exist?(feature)
+
+  # default features to enabled for test and those explicitly set for development
+  Flipper.enable(feature) if Rails.env.test? || (Rails.env.development? && feature_config['enable_in_development'])
```
ericboehs, line 48: It seems we shouldn't be trying to enable a feature unless it exists. Is there a reason to exclude this from `.exists?` conditional?

config/features.yml
```diff
@@ -1,11 +1,13 @@
 ---
-# Add a new feature toggle here to ensure that it is initialized in all environments.
-# Features are defaulted to enabled in development and test environments and disabled in all others.
+# Add a new feture toggle here to ensure that it is initialized in all environments.
+#
+# Features are enabled by default in test environments, and disabled by default in other environments.
+# Set a `enable_in_development` to 'true' to override this behavior for development.
```
ericboehs, line 5: This line sounds like it was written for the commit rather than this file as it looks like it's already set to `true` for all features. Should the copy read `false` here? 😕

config/initializers/flipper.rb
erluti: this was an oversight, good catch!

LindseySaari (approved):
💯

erikphansen:
Why was `enable_in_development: true` set for only some of the existing feature flags? Profile 2.0 stopped showing up locally after this change. If I hadn't seen one mention of this change in one VSA channel I probably would have had no idea why things were behaving strangely locally.

PR #4279, opened 2020-05-15
shorten dangerbot error messages

omgitsbillryan:
## Description of change
Updates our ruby-danger configuration:
- now **fails** instead of **warns** when files are changed in `/app` & `/db` within the same PR
- Places the file summary of each check into a collapsable element
- Some wording changes
  - adds a few links on migration and deployment best practices

How the comment looks... [before](https://github.com/department-of-veterans-affairs/vets-api/pull/4238#issuecomment-626061940) & [after](https://github.com/department-of-veterans-affairs/vets-api/pull/4272#issuecomment-628694668)

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/9043

ruby-danger is not required to pass for a PR to be able to merge... so even though a PR might fail because it creates migrations and app changes together, the PR author is will able to merge to PR - a decision I think we want to change.

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -11,13 +11,31 @@ lines_of_code = filtered_changed_files.sum { |_file, changes| (changes[:insertio
 
 if lines_of_code > MAX_PR_SIZE
   msg = <<~HTML
-    You changed `#{lines_of_code}` LoC. This exceeds our desired maximum of `#{MAX_PR_SIZE}`. Big PRs are difficult to review and often become stale. Consider breaking this PR up into smaller ones.
+    You changed `#{lines_of_code}` LoC. This exceeds our desired maximum of `#{MAX_PR_SIZE}`.
 
-    **Calculation Summary**
+    <details><summary>File Summary</summary>
+
+    #### Included Files
+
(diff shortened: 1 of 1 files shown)


omgitsbillryan:
All of my testing was done on [this PR](https://github.com/department-of-veterans-affairs/vets-api/pull/4272) and the code just copied to here.

Dangerfile
```diff
@@ -28,14 +46,31 @@ db_files  = all_touched_files.select { |filepath| filepath.include? "db/" }
...
+    - #{app_files.collect { |filepath| "#{filepath}" }.join("\n- ")} 
+
+    </details>
+
+    Database migrations do not run automatically with vets-api deployments. Application code must always be backwards compatible with the DB, both before and after migrations have been run. For more info: 
+
+    - [guidance on safe db migrations](https://github.com/ankane/strong_migrations#checks)
+    - [`vets-api` deployment process](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/engineering/deployment.md)
+
+  HTML
```
omgitsbillryan, line 69: The diff looks big, but all I did was update this section to display the message using heredoc consistent with how we do it for max PR size. It helps with readability and easily allows us to include the collapsable `<details><summary>File Summary</summary>` html.

Also changed the wording around.

PR #4283, opened 2020-05-15
remove sidekiq-instrument

erluti:
## Description of change
In order to upgrade sidekiq to 6, we can't use sidekiq-instrument.
This PR pulls out of https://github.com/enova/sidekiq-instrument the features we use and puts them in our own codebase to use and maintain as we see fit.
I also introduce tests for the stats we are tracking.

## Original issue
department-of-veterans-affairs/va.gov-team#8953


spec/rails_helper.rb
```diff
@@ -94,9 +94,9 @@ def self.all_matches
 Sidekiq::Testing.fake!
 Sidekiq::Testing.server_middleware do |chain|
   chain.add Sidekiq::SemanticLogging
-  # chain.add Sidekiq::Instrument::ServerMiddleware
   chain.add Sidekiq::ErrorTag
 end
+require 'sidekiq/api'
```
johnpaulashenfelter, line 99: This seems to stand out -- not clear why it has to be in all the specs

erluti: i was just putting it with the other `require 'sidekiq/testing'` but i can put it in the only test that uses it

LindseySaari:
Are we blocked by Sidekiq Zeitwerk mode on this/close for the time being?

erluti:
> Are we blocked by Sidekiq Zeitwerk mode on this/close for the time being?

This PR is good to go.  It's still `Sidekiq 5`, but sidekiq-instrument is no longer supported and will block us from ever doing `Sidekiq 6`.

omgitsbillryan:
If there's some technical limitation to using sidekiq < 6, and you didn't modify any of the code, I'd think that this code wouldn't work for the same reason that the gemspec includes: `'sidekiq', '>= 4.2', '< 6'`.

So,

1. Do you know exactly `sidekiq-instrument` requires sidekiq < 6?
2. Did you modify any of the code when porting it to our codebase?

erluti:
> 1. Do you know exactly `sidekiq-instrument` requires sidekiq < 6?
Sadly because it required sidekiq < 5 until it didn't.  Almost 3 years ago `sidekiq-instrument` went to version `0.3.0` with this commit https://github.com/enova/sidekiq-instrument/commit/6c74ff1cbcf99141f389ef3f17db530724ee7297 that says it's relaxing the requirements for sidekiq and will not be tested.
> 2. Did you modify any of the code when porting it to our codebase?
The last commit to the repo was more than 2 years ago to support RubyGems and this lack of support is part of the inspiration to remove the gem dependency.
My approach here was to write tests for `SidekiqStatsJob` that tests the StatsD values we were generating.  When those tests were passing for our current codebase, I add the `perform`method from the [sidekiq-instrument worker class](https://github.com/enova/sidekiq-instrument/blob/master/lib/sidekiq/instrument/worker.rb) and included Sidekiq::Worker like that base class does.  The tests created for our original were still passing.
Then I needed to pull over the [sidekiq-instrument middleware](https://github.com/enova/sidekiq-instrument/tree/master/lib/sidekiq/instrument/middleware).  I modified those to not require the [mixin](https://github.com/enova/sidekiq-instrument/blob/master/lib/sidekiq/instrument/mixin.rb) that only generated queue names because it supported using a `statsd_metric_name` method that we never used.  I copied over the associated tests for those middleware libraries to maintain that my versions accomplished what the originators had specc'ed them to do.
Finally there are also some changes due to our linting rules.

Overall I think that this makes it clearer what stats we are creating and makes it easier to look up where those values come from and remove/change them if we ever wanted in addition to removing the requirements of an unmaintained gem.

omgitsbillryan:
I should have been more specific, I was trying to poke specifically at any changes to make this code sidekiq 6 compliant. W/o that, we're basically 🤞 that this works with sidekiq 6 - basically the same thing as forking the repo and changing the gemspec to allow sidekiq 6.

The code in this PR (and the `sidekiq-instrument` gem) requires an understanding of Sidekiq classes and meta programming that most devs on `vets-api` don't readily have, myself included. I'm not sure how much of a win it is to remove our reliance on an unmaintained gem in favor of code in `vets-api` that will be largely ignored and not understood. I'd prefer to have that stuff abstracted away in a gem personally.

omgitsbillryan:
Side note, we might also consider what looks to be a more up to date sidekiq statsd instrumentation lib https://github.com/phstc/sidekiq-statsd

PR #4291, opened 2020-05-18
[VA Forms] Fix for scrape not processing a page

ncksllvn:
## Description of change

We noticed that the count of valid form on `/v0/forms` does not match 471 as indicated on `/vaforms/search_action.asp`. From comparing the data sets of `v0/forms` with those on `v.gov/vaforms`, we found that page 2 of the results was being skipped during the scrape job.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/8647

- I ran this script locally and outputted the scraped form names. It matched 471 and included forms that are missing from `v0/forms`.


modules/va_forms/app/workers/va_forms/form_reloader.rb
```diff
@@ -13,7 +13,7 @@ def initialize
     end
 
     def perform
-      load_page(current_page: 0)
+      load_page(current_page: -1)
```
ncksllvn, line 16: Indexing on `vaforms/search_action.asp` starts at 0, so the first page should be `-1` in order to account for the increment inside of `load_page`. Otherwise, we increment to 1, then hit the `unless current_page == 1` condition, which removes the query params to get the first page, but the next iteration of the loop goes to page 3 instead of page 2.

modules/va_forms/app/workers/va_forms/form_reloader.rb
```diff
@@ -44,6 +43,7 @@ def load_page(current_page: 0)
       next_button = doc.css('input[name=Next10]')
       last_page = next_button.first.attributes['disabled'].present?
       parse_page(doc)
+      current_page += 1
```
ncksllvn, line 46: Indexing on `vaforms/search_action.asp` starts at `0`. The way this was written with the increment at the top and the `unless current_page == 1` would cause page 2 to be skipped. It would go -> Page 1 (`unless current_page == 1` would omit the parameters, so `current_page` of 1 results in page 1) -> Page 3 (when `current_page` is 2) -> ...

bastosmichael: Are you counting down or up? Wouldn't this technically never stop looping if the current page was never set to zero?

modules/va_forms/app/workers/va_forms/form_reloader.rb
```diff
@@ -26,9 +26,8 @@ def mark_stale_forms
     end
 
     def load_page(current_page: 0)
-      current_page += 1
       params = {}
-      unless current_page == 1
+      unless current_page.zero?
         params = {
```
bastosmichael, line 31: Are you counting down or up? Wouldn't this technically never stop looping if the current page was never set to zero?

ncksllvn: It increments on line 46, This `unless` block just omits the `params` values if we're on page zero because page zero is the default page. I don't think it's necessary at all but that's how the script was written

bastosmichael: @charleystran wrote the original implementation so I want to have his eyes on why this unless is in place and how it increments before we may a change like this, he might have better context on the logic. I only wrote the original proof of concept before pagination was introduced.

charleystran: The unless was in place in order to send blank parameters for the starting page so we would mimic the current url. This change is ok and fixes the issue that was found.

bastosmichael: :+1: :shipit:

PR #4263, opened 2020-05-13
Integrate fhir healthcare service read request

ajmagdub:
## Description of change
Add controller, routes, parameter validations, and mock VCR Cassette for specs for the FHIR Service to support HealthCareService resource.
 
## Original issue(s)
[VAOS + FHIR - Search Healthcare Service Resource](https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/9095)

PR #4294, opened 2020-05-19
9094 VAOS FHIR query organization

kreek:
## Description of change
Add the index endpoint used to query FHIR Organization resources. Updates the FHIRService to take a resource in its constructor and dynamically select the correct resource in the controller.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#9094


modules/vaos/app/controllers/vaos/v1/base_controller.rb
```diff
@@ -18,7 +18,11 @@ class BaseController < VAOS::V0::BaseController
       private
 
       def fhir_service
-        VAOS::V1::FHIRService.new(current_user)
+        VAOS::V1::FHIRService.new(current_user, controller_name.singularize.capitalize.to_sym)
+      end
+
+      def query_string
+        request.fullpath.split('?').last
```
LindseySaari, line 25: Would the query string ever be blank?

kreek: Good call, so far getting a list of resources is always a query but that could change.

saneshark: # File actionpack/lib/action_controller/request.rb, line 333
```ruby
    def query_string
      @env['QUERY_STRING'].present? ? @env['QUERY_STRING'] : (@env['REQUEST_URI'].split('?', 2)[1] || '')
    end
```

LindseySaari: Awesome 👍 Could you implement the edge case here and in the tests as well?

modules/vaos/app/services/vaos/v1/fhir_service.rb
```diff
@@ -9,33 +9,47 @@ module V1
...
-    #   service = VAOS::V1::FHIRService.new(user)
-    #   response = service.read(:Organization, 987654)
+    #   service = VAOS::V1::FHIRService.new(user, :Organization)
+    #   response = service.read(987654)
     #
     class FHIRService < VAOS::BaseService
       STATSD_KEY_PREFIX = 'api.vaos.fhir'
       RESOURCES = %i[Appointment HealthcareService Location Organization Patient Schedule Slot].freeze
 
+      def initialize(user, resource_type)
```
saneshark, line 19: NITPICK: if we make user, the second argument, defaulted to nil, then some of these actions that don't require user could later be refactored to not have it. Alternately, consider making it a "named parameter,"

```ruby
def initialize(resource_type, user: nil)
```

LindseySaari:
Note: There may be some merge conflicts with this PR: https://github.com/department-of-veterans-affairs/vets-api/pull/4263

LindseySaari:
Let me know when this is ready for re-review

PR #4297, opened 2020-05-19
10-10CG Updates - Default veteran status, remove ICN for veteran when sending to CARMA , Fix MVI Lookup

kevinmirc:
**Description**
If the veteran status cannot be confirmed, do not send CARMA the ICN of the veteran on the submission (requested by the Caregiver Program and CARMA).

**Technical Summary**
- [x] Update the `isVeteran` flag to default to `false` vs `null` (requested by CARMA team).
- [x] If the veteran status cannot be confirmed, on a Form 10-10cg Submission, do not pass the ICN for the veteran to CARMA.
- [x] `MVI::Service` does not raise an error on bad requests. Read the response's `error` attribute vs trying to rescue an exception.

### DIFF
--- a/app/services/form1010cg/service.rb
+++ b/app/services/form1010cg/service.rb
@@ -92,7 +92,7 @@ module Form1010cg
 
       case response.status
       when 'OK'
-        return @cache[:icns][form_subject] = response.profile.icn.split('V').first
+        return @cache[:icns][form_subject] = response.profile.icn
       when 'NOT_FOUND'
         return @cache[:icns][form_subject] = NOT_FOUND
       end
--- a/app/services/form1010cg/service.rb
+++ b/app/services/form1010cg/service.rb
@@ -92,7 +92,7 @@ module Form1010cg
 
       case response.status
       when 'OK'
-        return @cache[:icns][form_subject] = response.profile.icn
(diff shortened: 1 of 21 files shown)


app/services/form1010cg/service.rb
```diff
@@ -83,38 +80,43 @@ def build_metadata
...
       return cached_icn unless cached_icn.nil?
 
-      begin
-        response = mvi_service.find_profile(build_user_identity_for(form_subject))
-      rescue MVI::Errors::RecordNotFound
+      response = mvi_service.find_profile(build_user_identity_for(form_subject))
+
+      case response.status
+      when 'OK'
+        return @cache[:icns][form_subject] = response.profile.icn
```
LindseySaari, line 92: Noticed that you removed the safe navigation operators response&.profile&.icn, would response.profile ever be nil?

kevinmirc: Since we're doing a search for a profile, and the status is "OK", there should be a profile there. Also had a safeguard on ICN, but found out that an ICN should always be present on an MVI profile.

LindseySaari: Ok great, looks good

PR #4299, opened 2020-05-19
Add patient info to profile

patrickvinograd:
## Description of change
Provides indicators for conditions around MHV access eligibility in the user profile:
- va_patient - whether the user has registered identifiers at any VA treating facility
- mhv_account_state - whether the user has correlated MHV account identifiers in their MVI profile.

These are used by the frontend to gate access to MHV, in conjunction with https://github.com/department-of-veterans-affairs/vets-website/pull/12624. That PR eliminates the calls to create/upgrade MHV accounts when SSOe is used, in favor of sending the user to MHV to register/upgrade their account there. 

Eventually, these new profile attributes can replace the separate call to the mhv_account API endpoint, and all the complexity of tracking mhv account state can be eliminated. vets-api will do some basic eligibility/soundness checks, and then send the user to MHV to modify their account if needed.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#8405

No config changes/feature flags. This is using/aggregating some existing data from MVI.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -107,6 +107,14 @@ class User < Common::RedisStore
     identity.mhv_account_type || MhvAccountTypeService.new(self).mhv_account_type
   end
 
+  def mhv_account_state
+    return 'DEACTIVATED' if (va_profile.mhv_ids.to_a - va_profile.active_mhv_ids.to_a).any?
+    return 'MULTIPLE' if va_profile.active_mhv_ids.size > 1
+    return 'NONE' if mhv_correlation_id.blank?
+
+    'OK'
+  end
+
   def loa
     identity&.loa || {}
   end
--- a/app/services/users/profile.rb
+++ b/app/services/users/profile.rb
@@ -72,9 +72,7 @@ module Users
         multifactor: user.multifactor,
         verified: user.loa3?,
         sign_in: user.identity.sign_in,
-        authn_context: user.authn_context,
-        va_patient: user.va_patient?,
-        mhv_account_state: user.mhv_account_state
+        authn_context: user.authn_context
       }
     end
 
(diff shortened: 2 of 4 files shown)


LindseySaari:
Can you update specs for these changes?

app/models/user.rb
```diff
@@ -107,6 +107,14 @@ def mhv_account_type
     identity.mhv_account_type || MhvAccountTypeService.new(self).mhv_account_type
   end
 
+  def mhv_account_state
+    return 'DEACTIVATED' if (va_profile.mhv_ids.to_a - va_profile.active_mhv_ids.to_a).any?
+    return 'MULTIPLE' if va_profile.active_mhv_ids.size > 1
+    return 'NONE' if mhv_correlation_id.blank?
```
omgitsbillryan, line 113: Wouldn't it be possible for all of these scenarios to evaluate to `true`? - in which case, is it correct to favor `'DEACTIVATED'` just because it's the first guard clause?

(disclaimer: i know very little about MHV)

patrickvinograd: Yes it's possible, at least for DEACTIVATED and MULTIPLE to both be true. The remediation is the same - need to contact the help desk and get your accounts straightened out, so I think preferring DEACTIVATED is fine.

ericbuckley: would `mhv_ids` or `active_mhv_ids` ever be nil?

patrickvinograd:
Specs added

app/models/user.rb
patrickvinograd: I don't see how - they are populated from an Array.select which should always return an empty or populated array.

omgitsbillryan:
Whoops, I'm late. Looks like swagger would need to be updated too, no?

ericbuckley:
> Whoops, I'm late. Looks like swagger would need to be updated too, no?

@omgitsbillryan I just created this PR to fix https://github.com/department-of-veterans-affairs/vets-api/pull/4309

PR #4300, opened 2020-05-19
8720 v1 facilities api

darkbushido:
## Description of change
Lighthouse has created a new api that they would like us to use.
This removes the switch on the V0 facilities api and creates a new V1 facilities api that talks to the new lighthouse backend

## Original issue(s)
department-of-veterans-affairs/va.gov-team#8720

This removes the hack to switch the V0 api from active record to lighthouse
since this is new ground there isn't any feature flag
This does not have a SWAGGER doc. the current implementation of swagger doesn't support V1 namespacing well.
This api takes the request, passes it on to lighthouse with the needed api key
it then parses the response makes, adds the pagination metadata and links that point to this endpoint instead of passing the raw lighthouse response back.


app/controllers/facilities_controller.rb
```diff
@@ -12,4 +12,76 @@ def validate_params
   rescue ArgumentError
     raise Common::Exceptions::InvalidFieldValue.new('bbox', params[:bbox])
   end
+
+  private
+
```
LindseySaari, line 17: Not sure that this code belongs in the controller, could you implement a helper/module instead for this to keep the controller clean?

app/controllers/facilities_controller.rb
```diff
@@ -12,4 +12,77 @@ def validate_params
...
+      pagination: {
+        current_page: paginated_obj.current_page,
+        prev_page: paginated_obj.previous_page,
+        next_page: paginated_obj.next_page,
+        total_pages: paginated_obj.total_pages
+      }
+    }
+  end
+
+  def generate_previus_page_link(paginated_obj)
```
LindseySaari, line 54: spelling issue, it looks consistent where it's called, but just want to point out

app/controllers/facilities_controller.rb
darkbushido: I could do that. when this gets reused and we are killing off active model serializer this should move into application_controller.rb
https://medium.com/@anMagpie/rails-api-with-fastjson-from-netflix-e7bf4f4e5e7b

app/controllers/facilities_controller.rb
darkbushido: I can't spell my way out of a wet paper bag.

PR #4301, opened 2020-05-19
bam: Correctly accept more than one topic arn

benniemosher:
This PR fixes an issues where VBA Documents musts work with both VAGOV and VETSGOV SNS topic names.


charleystran:
@LindseySaari, @benniemosher helped us on the lighthouse side get this taken care of as a side effect of the sandbox migration.

PR #4302, opened 2020-05-19
Form 526: Remove invalid chars from treatmentCenterName

SarahJaine:
## Description of change

**_Please note: this is my second time writing Ruby so please let me know if my way of doing this was odd._**
Description from ticket:
About 500 form 526 submissions were rejected over the past week due to characters that EVSS doesn't like, for example:
```
 {
      "treatmentCenterName": "VA Medical Center, Miami Florida",
      "treatmentDateRange": {
        "from": "[REDACTED DATE]"
      },
```
was rejected by EVSS, due to the comma.
and we got bac this response from evss:
`{message="[{"key"=>"form526.treatments[0].center.name.Pattern" "text"=>"must match "([a-zA-Z0-9"\/&\(\)\-'.#]([a-zA-Z0-9(\)\-'.# ])?)+$"" "severity"=>"ERROR"}]"}`

## Original issue(s)
closes department-of-veterans-affairs/va.gov-team/issues/9121

- [x] added regex gsub method and strip to remove any invalid characters from `treatmentCenterName`
- Instead of copying the regex from the error, I simplified it into a _stricter_ protocol.  I thought this would be easier to read, while leaving intact any important characters that we could expect in a facility name.
- [x] changed preexisting test to ensure invalid characters were removed from `treatmentCenterName`


lib/evss/disability_compensation_form/data_translation_all_claim.rb
```diff
@@ -387,7 +387,7 @@ def translate_treatments
             'endDate' => approximate_date(treatment['treatmentDateRange']['to']),
             'treatedDisabilityNames' => treatment['treatedDisabilityNames'],
             'center' => {
-              'name' => treatment['treatmentCenterName']
+              'name' => treatment['treatmentCenterName'].gsub(/[^a-zA-Z0-9 .()#&'"-]+/) { '' }.strip.to_s
```
LindseySaari, line 390: You could also write the gsub comma replacement like: gsub(/[^a-zA-Z0-9 .()#&'"-]+/, '').strip.to_s (not sure that you need a block for a simple comma replacement). Both do the same thing though 😸

LindseySaari: Do you forsee any other data containing invalid comma characters? Another suggestion could be to make this into a method for reuse

SarahJaine: @LindseySaari Awesome, I didn't know you could do that. I guess I didn't read far enough into the gsub examples.  To your second point, there are over 100 invalid text errors possible for this form. I have a feeling many are corrected via front end validation.  If we do need regex checks for mult fields I'll pull this out to a method in a future PR.

annaswims, line 390:
```suggestion
              'name' => treatment['treatmentCenterName'].gsub(/[^a-zA-Z0-9 .()#&'"-]+/, '').strip
```

I think the `to_s` is harmless but unnecessary.

Also, the more I look at EVSS' regex, the less I understand their intent, but your regex will strip out the correct characters so 🤷‍♀️

annaswims: I'd lean towards not reusing this, because their regex is goofy and I hope they don't use it widely (we don't know if they do).

SarahJaine: Ah, yeah that returns a string.  When I was playing around with other regex methods, one of them didn't return a string.  I guess I didn't recheck the return type when I settled on gsub. Thanks!

SarahJaine: About the original EVSS regex, yeah, that thing is nearly illegible. I have no idea what their intent is beyond stripping those characters. The implication that they're looping through the two different capture groups makes my head hurt.  I think if anything, our regex will be stricter then theirs, which is preferable.

LindseySaari: Sounds good!

SarahJaine:
Released as [vets-api/v0.0.624](https://github.com/department-of-veterans-affairs/vets-api/releases/tag/vets-api%2Fv0.0.624) in [vets-api release job 671](http://jenkins.vfs.va.gov/job/releases/job/vets-api/671/console) which was [deployed May 20, 2020 7:28:28 PM](http://jenkins.vfs.va.gov/job/deploys/job/vets-api-worker-vagov-prod/274/)

PR #4304, opened 2020-05-20
Additional redirect parameter 

patrickvinograd:
## Description of change
Adds support for an additional parameter `to` that can be specified along with `application` in the signin process. The to parameter is intended as a relative path that gets tacked on to the redirect URL derived from `application`, to construct a complete redirect URL. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#8587

No feature flags or additional settings.

Allowing our system to redirect based on user input presents some risk, as described in https://cheatsheetseries.owasp.org/cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.html

We mitigate  this by:
- Not letting the caller specify the entire URL, but only the path part - the base URL/domain is derived from the `application` parameter whose configuration we control.
- Ensuring that we join the base URL and `to` parameter with a `/` to avoid domain extension
- Stripping any CRLF characters to avoid header splitting attacks.

### DIFF
--- a/lib/saml/url_service.rb
+++ b/lib/saml/url_service.rb
@@ -56,7 +56,7 @@ module SAML
       if auth == 'success'
         return verify_url if user.loa[:current] < user.loa[:highest]
 
-        redirect_target = build_tracked_url
+        redirect_target = @tracker&.payload_attr(:redirect)
         return redirect_target if redirect_target.present?
       end
 
@@ -71,16 +71,6 @@ module SAML
       end
     end
     # rubocop:enable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
-
-    def build_tracked_url
-      return nil unless @tracker&.payload_attr(:redirect)
-
-      prefix = @tracker&.payload_attr(:redirect)
-      prefix = prefix.delete_suffix('/')
-      suffix = @tracker&.payload_attr(:to) || ''
-      suffix = suffix.delete_prefix('/').gsub('\r\n', '')
-      [prefix, suffix].join('/')
-    end
 
     def logout_redirect_url
       "#{base_redirect_url}#{LOGOUT_REDIRECT_PARTIAL}"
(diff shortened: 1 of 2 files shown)


patrickvinograd:
Needs tests

ericbuckley:
@patrickvinograd what do you think about just adding a new `build_redirect` method to encapsulate all the logic and storing it as one attribute in the SAMLRequestTracker rather than 2?

```
diff --git a/lib/saml/url_service.rb b/lib/saml/url_service.rb
index 0ef2f9b6d..9487b6deb 100644
--- a/lib/saml/url_service.rb
+++ b/lib/saml/url_service.rb
@@ -189,12 +189,21 @@ module SAML
       end
     end

+    def build_redirect
+      default = Settings.ssoe.redirects[params[:application]]
+      return default unless default && params[:to]
+
+      url = URI.parse(default)
+      url.path = params[:to]
+      return url.to_s
+    end
+
     # Initialize a new SAMLRequestTracker, if a valid previous SAML UUID is
     # given, copy over the redirect and created_at timestamp.  This is useful
     # for a user that has to go through the upleveling process.
     def initialize_tracker(params, previous_saml_uuid: nil)
       previous = previous_saml_uuid && SAMLRequestTracker.find(previous_saml_uuid)
-      redirect = previous&.payload_attr(:redirect) || Settings.ssoe.redirects[params[:application]]
+      redirect = previous&.payload_attr(:redirect) || build_redirect
       # if created_at is set to nil (meaning no previous tracker to use), it
       # will be initialized to the current time when it is saved
       SAMLRequestTracker.new(
```

patrickvinograd:
I like that. Using `uri.path` seems cleaner, though will need to catch `URIError` and figure out what to do with it. Seems like URI.parse does not mess with encoding/decoding which is nice.

patrickvinograd:
I found that using the `URI`  class to parse/assemble the URL was more trouble than it was worth - because of error handling as well as the potential for the nested "to" query parameter.

lib/saml/url_service.rb
```diff
@@ -53,9 +53,12 @@ def base_redirect_url
 
     # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
     def login_redirect_url(auth: 'success', code: nil)
-      return verify_url if auth == 'success' && user.loa[:current] < user.loa[:highest]
+      if auth == 'success'
+        return verify_url if user.loa[:current] < user.loa[:highest]
 
-      return @tracker&.payload_attr(:redirect) if @tracker&.payload_attr(:redirect)
```
ericbuckley, line 58: for the cerner redirect, if the user authentication fails, do we want to show them the va.gov error page or redirect them back to cerner for them to show an error?

patrickvinograd: I...don't know. Question for @bridgethapner if we've specified that

patrickvinograd: My concern would be that Cerner would try to send them back to the sign-in page again and they'd be stuck in a loop with no real feedback

patrickvinograd: I think we should deploy this as-is, wherein it will show an error message, while we iterate with Cerner et al about whether we should redirect them back to any particular destination.

ericbuckley: I'm ok with that, I just wanted to make sure it's intentional to redirect cerner users to the va.gov failure page.

PR #4306, opened 2020-05-20
Add partial SSN to form20-0996

SarahJaine:
## Description of change

Backend's "Save in progress" (SIP) response to /v0/in_progress_forms/20-0996 should include last 4 of SSN per the FE design. I believe the exclusion of this field was an oversight in the original form implementation that is now being corrected.

## Original issue(s)
closes department-of-veterans-affairs/va.gov-team#9311

- [x] add last four of SSN to SIP
- [x] add test for partial SSN to form attribute spec
- This seems way too easy to be right but here we are.


annaswims:
looks pretty much like https://github.com/department-of-veterans-affairs/vets-api/pull/4143/files

annaswims:
looks like we'd want to make a change to vets-json-schema like this one https://github.com/department-of-veterans-affairs/vets-json-schema/commit/9b5769e3bb45c511faef521003fee451724d6b18

SarahJaine:
> looks like we'd want to make a change to vets-json-schema like this one [department-of-veterans-affairs/vets-json-schema@9b5769e](https://github.com/department-of-veterans-affairs/vets-json-schema/commit/9b5769e3bb45c511faef521003fee451724d6b18)

That's a good point. Yeah the json schema will definitely need to be updated. I've never really understood the way the schema validation works, do the PRs for the BE changes and the json schema need to go through in a particular order?

SarahJaine:
> > looks like we'd want to make a change to vets-json-schema like this one [department-of-veterans-affairs/vets-json-schema@9b5769e](https://github.com/department-of-veterans-affairs/vets-json-schema/commit/9b5769e3bb45c511faef521003fee451724d6b18)
> 
> That's a good point. Yeah the json schema will definitely need to be updated. I've never really understood the way the schema validation works, do the PRs for the BE changes and the json schema need to go through in a particular order?

@annaswims I spoke with @Mottie about this.  He informed me that the json schema validation is run on the BE for data submitted by the user to the BE. Since the SIP data is going the opposite direction (from the BE to the FE) it should not be included in the json schema.

PR #4308, opened 2020-05-21
Validating JSON object instead of JSON content

bastosmichael:
## Description of change
One of our consumers sent us an array of json objects instead of just an actual json object which we never validated against so this PR fixes that and ensures we fail invalid non json objects for metadat in the future.

## Original issue(s)
https://vajira.max.gov/browse/API-946

### DIFF
--- a/modules/vba_documents/app/workers/vba_documents/upload_processor.rb
+++ b/modules/vba_documents/app/workers/vba_documents/upload_processor.rb
@@ -107,9 +107,9 @@ module VBADocuments
         raise VBADocuments::UploadError.new(code: 'DOC102',
                                             detail: 'Non-numeric or invalid-length fileNumber')
       end
-    rescue JSON::ParserError
+    rescue JSON::ParserError, NoMethodError
       raise VBADocuments::UploadError.new(code: 'DOC102',
-                                          detail: 'Invalid JSON content')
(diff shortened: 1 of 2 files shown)


modules/vba_documents/app/workers/vba_documents/upload_processor.rb
```diff
@@ -107,9 +107,9 @@ def validate_metadata(metadata_input)
         raise VBADocuments::UploadError.new(code: 'DOC102',
                                             detail: 'Non-numeric or invalid-length fileNumber')
       end
-    rescue JSON::ParserError
+    rescue JSON::ParserError, NoMethodError
```
edmangimelli, line 110: is the `NoMethodError` to catch if `metadata` isn't a hash?
if that's the case, what do you think about just testing for that after the parse?
`.is_a? Hash`
or even test that is reponds to `:keys` or `:[]`

edmangimelli: I guess my thought is that I wouldn't want catching NoMethodError to obscure some other problem (but I'm still just guessing at the intention)

bastosmichael: ya I like that approach much better, let me refactor some things one sec.

PR #4309, opened 2020-05-21
adding swagger properties for new va profile attributes

ericbuckley:
Adding swagger properties for the two (`va_patient` and `mhv_account_state`) new va profile attributes.

refs https://github.com/department-of-veterans-affairs/vets-api/pull/4299


app/swagger/requests/user.rb
```diff
@@ -99,6 +99,8 @@ class User
                       key :type, :string
                     end
                   end
+                  property :va_patient, type: :boolean
+                  property :mhv_account_state, type: :string, enum: %w[OK DEACTIVATED MULTIPLE NONE], example: 'OK'
```
omgitsbillryan, line 103: Thoughts on adding a `description:`? `MULTIPLE` & `NONE` in particular don't seem immediately obvious to me what they mean.

app/swagger/requests/user.rb
```diff
@@ -99,6 +99,15 @@ class User
                       key :type, :string
                     end
                   end
+                  property :va_patient, type: :boolean
+                  property :mhv_account_state,
+                           type: :string,
+                           enum: %w[OK DEACTIVATED MULTIPLE NONE],
+                           example: 'OK',
+                           description: 'DEACTIVATED: user has MHV ids that are not active; '\
```
ericbuckley, line 107: @patrickvinograd is this swagger description accurate?

patrickvinograd, line 107:
```suggestion
                           description: 'DEACTIVATED: user has at least one MHV id that is not active; '\
```

patrickvinograd: Minor phrasing update, otherwise, descriptions look accuate

PR #4311, opened 2020-05-21
Allow CaregiverAssistanceClaims to generate a PDF now.

kevinmirc:
## Description of change
Allow a `CaregiversAssistanceClaim` to receive `:to_pdf` now that it has the capability.

### DIFF
--- a/spec/models/saved_claim/caregivers_assistance_claim_spec.rb
+++ b/spec/models/saved_claim/caregivers_assistance_claim_spec.rb
@@ -3,45 +3,15 @@
 require 'rails_helper'
 
 RSpec.describe SavedClaim::CaregiversAssistanceClaim do
-  let(:build_claim_data_for) do
-    lambda do |form_subject, &mutations|
-      data = {
-        'fullName' => {
-          'first' => Faker::Name.first_name,
-          'last' => Faker::Name.last_name
-        },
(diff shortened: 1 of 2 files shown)


spec/models/saved_claim/caregivers_assistance_claim_spec.rb
```diff
@@ -3,45 +3,15 @@
...
-      mutations&.call data
+  describe '#to_pdf' do
+    it 'converts form to pdf' do
+      claim = build(:caregivers_assistance_claim)
+      expected_file_path = 'tmp/pdfs/my_claim.pdf'
 
-      data
-    end
-  end
+      expect(PdfFill::Filler).to receive(:fill_form).with(claim).once.and_return(expected_file_path)
```
LindseySaari, line 11: Isn’t this behavior encapsulated in the test below on line 13 (claim.to_pdf) since the to_pdf method calls PdfFill::Filler.fill_form(claim)? I guess it doesn't hurt to call out in a separate expectation though

kevinmirc: Not sure if I see your point. Yes, `claim.to_pdf` encapsulates the logic above. I want to stub it though to make sure I'm not actually executing `PdfFill::Filler#fill_form` and that `#to_pdf` calls it properly.

PR #4293, opened 2020-05-19
Decision Reviews API - Update `decision_review` Settings

edmangimelli:
update decision_review settings in settings.yml.
1) make them match more closely to what's in the devops repo
2) correct version number
3) remove unused api key field

Acceptance Criteria
- [x] tests pass

PR #4313, opened 2020-05-21
Appeals API - Pull Our Settings Out of `decision_review`

edmangimelli:
This comment https://github.com/department-of-veterans-affairs/vets-api/pull/4293#discussion_r427291436
prompted me to look at `lib/decision_review` and realize that, at some point, we started putting our settings under `decision_review` ***which is not owned by the Lighthouse Appeals API team***. My best guess is that the va.gov front-end team (which will eventually be consuming the Appeals API) owns `decision_review` in `settings.yml`

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -133,9 +133,12 @@ mdot:
 # Settings for Decision Reviews
 decision_review:
   prefill: true
-  url: https://sandbox-api.va.gov/services/appeals/v3/decision_review
+  url: https://sandbox-api.va.gov/services/appeals/v1/decision_review
   api_key: abcd1234abcd1234abcd1234abcd1234abcd1234
   mock: false
+
+# Settings for modules/appeals_api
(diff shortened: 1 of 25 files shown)


config/settings.yml
```diff
@@ -133,9 +133,12 @@ mdot:
...
   prefill: true
-  url: https://sandbox-api.va.gov/services/appeals/v3/decision_review
+  url: https://sandbox-api.va.gov/services/appeals/v1/decision_review
   api_key: abcd1234abcd1234abcd1234abcd1234abcd1234
   mock: false
+
+# Settings for modules/appeals_api
+modules_appeals_api:
   higher_level_review_updater_enabled: true
   schema_dir: config/schemas
```
edmangimelli, line 143: `modules_appeals_api` isn't a beautiful name, but there's certainly no mistaking what it is (this isn't the first time we've collided with the other team)

PR #4314, opened 2020-05-21
Only call EVSS - PCIU endpoints for LOA3 users with an edipi

annaswims:
## Description of change
Only call EVSS PCIU endpoints for LOA3 users with and EDIPI.  EVSS requires an EDIPI 

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/8676

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -47,15 +47,15 @@ class User < Common::RedisStore
   end
 
   def pciu_email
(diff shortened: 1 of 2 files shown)


annaswims:
It seems like we should cache these calls, but at least we can reduce the number of calls with this PR.

PR #4315, opened 2020-05-21
fix for when active_mhv_ids is nil

ericbuckley:
When `va_profile.active_mhv_ids` is `nil` the `mhv_account_state` method can result in an error.

http://sentry.vfs.va.gov/vets-gov/platform-api-staging/issues/120906/?query=is%3Aunresolved

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -109,7 +109,7 @@ class User < Common::RedisStore
 
   def mhv_account_state
     return 'DEACTIVATED' if (va_profile.mhv_ids.to_a - va_profile.active_mhv_ids.to_a).any?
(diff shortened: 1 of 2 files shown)


ericbuckley:
I tried creating a test for this use case using the following mvi stub, but for some reason even when explicitly setting the value to `nil`, it always evaluated to the empty string.  This probably has something to do with how factories work, but I was unable to figure it out.

```
      let(:mvi_profile) do
        build(:mvi_profile_response,
              :missing_attrs)
      end
      let(:mhv_user) { build(:user, :mhv) }
```

patrickvinograd:
Try adding `       allow_any_instance_of(MVI::Models::MviProfile).to receive(:active_mh    v_ids).and_return(nil)` in the `before do` block of a new test case

PR #4319, opened 2020-05-22
reworking session controllers statsd init

ericbuckley:
New tags have been added to track differences between v0 and v1 sessions, however the current statsd initialization code doesn't address all the new combinations of tags in use.  This PR addresses that, so we can accurately track increases in grafana.

refs https://github.com/department-of-veterans-affairs/va.gov-team/issues/9428

### DIFF
--- a/app/controllers/v0/sessions_controller.rb
+++ b/app/controllers/v0/sessions_controller.rb
@@ -136,12 +136,7 @@ module V0
 
     def login_stats_success(saml_response, user_session_form = nil)
       tracker = url_service(user_session_form&.saml_uuid).tracker
-      tags = [
-        "loa:#{@current_user.loa[:current]}",
-        "idp:#{@current_user.identity.sign_in[:service_name]}",
-        "context:#{saml_response.authn_context}",
-        VERSION_TAG
-      ]
+      tags = ["context:#{saml_response.authn_context}", VERSION_TAG]
       StatsD.increment(STATSD_LOGIN_NEW_USER_KEY, tags: [VERSION_TAG]) if request_type == 'signup'
       StatsD.increment(STATSD_LOGIN_SHARED_COOKIE, tags: tags) if cookies.key?(Settings.sso.cookie_name)
       StatsD.increment(STATSD_LOGIN_STATUS, tags: tags + ['status:success'])
@@ -156,9 +151,7 @@ module V0
       when :failure
         StatsD.increment(STATSD_LOGIN_STATUS,
                          tags: ['status:failure',
-                                "idp:#{params[:type]}",
                                 "context:#{saml_response.authn_context}",
-                                "error:#{user_session_form.error_instrumentation_code}",
                                 VERSION_TAG])
         callback_stats(:failure, saml_response, user_session_form.error_instrumentation_code)
       end
--- a/app/controllers/v1/sessions_controller.rb
+++ b/app/controllers/v1/sessions_controller.rb
@@ -12,6 +12,8 @@ module V1
     REDIRECT_URLS = %w[signup mhv dslogon idme mfa verify slo].freeze
 
     STATSD_SSO_NEW_KEY = 'api.auth.new'
+    STATSD_SSO_NEW_FORCEAUTH = 'api.auth.new.forceauth'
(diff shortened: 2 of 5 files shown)


app/controllers/v0/sessions_controller.rb
```diff
@@ -136,12 +136,7 @@ def user_logout(saml_response)
 
     def login_stats_success(saml_response, user_session_form = nil)
       tracker = url_service(user_session_form&.saml_uuid).tracker
-      tags = [
-        "loa:#{@current_user.loa[:current]}",
```
LindseySaari, line 140: Could you explain the removal of the idp and loa tags? The reason isn't obvious to me based on the PR description

ericbuckley: I added the `loa` and `idp` tags a [couple of months ago](https://github.com/department-of-veterans-affairs/vets-api/pull/3762).  They're redundant as both can be inferred from the `context` tag, however at the time I thought it would be easier to query on.

After reading the [StatsD Metrics Guide](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/engineering/backend/sending-metrics-using-statsd.md#why-do-we-increment-a-bunch-of-metrics-by-0-in-our-initializers) yesterday it became apparent that too many tags can have performance implications in Prometheus and it does make initialization more complex.  So in light of this, and knowing the values are redundant, I removed them.

LindseySaari: Ok thanks for the info! Just wanted to make sure we weren't breaking any grafana boards by removing them! ➕ 💯 for removing tags in relation to better performance

app/controllers/v1/sessions_controller.rb
```diff
@@ -30,7 +32,15 @@ def new
       raise Common::Exceptions::RoutingError, params[:path] unless REDIRECT_URLS.include?(type)
 
       StatsD.increment(STATSD_SSO_NEW_KEY,
-                       tags: ["context:#{type}", "forceauthn:#{force_authn?}", VERSION_TAG])
+                       tags: ["context:#{type}", VERSION_TAG])
+      if force_authn?
```
LindseySaari, line 36: Since we’re tracking STATSD_SSO_NEW_FORCEAUTH and STATSD_SSO_NEW_INBOUND, do we still want to track STATSD_SSO_NEW_KEY as a separate metric? I see that STATSD_SSO_NEW_KEY used to have a tag of forceauthn, but no longer does.

Could you also add specs for the new STATSD_SSO_NEW_FORCEAUTH and STATSD_SSO_NEW_INBOUND metrics please?

ericbuckley: All 3, or at least the ability to track the 3 scenarios, is going to be needed for our reporting.  We want to know when
a) a user initiates a new authentication request
b) a user initiates a new authentication request using forceAuthN
c) a user initiates a new authentication request using inbound SSOe (auto login)

`b` and `c` are optional, but `a` will happen every time.  I think we can do this with 1 stat and 2 tags or just 3 different stats.  I'm not sure what the preferred way for measuring this is, but I'm open to suggestions.

LindseySaari: Ah ok, thanks for the explanation. I think the three stat approach is fine.  My only other comment is just some additional specs for the inbound and forceauth metrics. Thanks!

ericbuckley: sure thing.  the specs have been updated here https://github.com/department-of-veterans-affairs/vets-api/pull/4319/files#diff-b6ff071088096f7d49f49f8b428ab449L86-R137

ericbuckley:
> Looks good after [CVE-2020-11077](https://github.com/advisories/GHSA-w64w-qqph-5gxm) issue is addressed from CI and Code Climate complexity issue

@LindseySaari I just [merged in master](https://github.com/department-of-veterans-affairs/vets-api/pull/4319/commits/bf0a8fde3a1c52a45380b25fcfe524a888065291) to remove the puma vulnerability. Is the PR ok to merge now?

LindseySaari:
Looks like a CodeClimate complexity issue, but after that, looks good 👍

PR #4320, opened 2020-05-22
Update native install instructions to support Ubuntu 20.04

benniemosher:
This PR is the steps that I had to take to get vets-api running on Ubuntu 20.04. I did not use the 18.04 instructions because it has you use a script that installs things I don't need like `vlc`.


benniemosher:
> Nice! Thanks for this, would you mind updating the ruby version in the Alternative (Ubuntu 18.04 LTS) instructions as well? I see it still references v 2.4.5

This should be good to go @LindseySaari!

PR #4329, opened 2020-05-26
Inbound SSOe force needed authentication

ericbuckley:
When a user logs in via SSOe and is missing an ID.me UUID value, we should abort the authentication process and redirect them back to VA.gov with the `force-needed` flag set.

refs https://github.com/department-of-veterans-affairs/va.gov-team/issues/5416


ericbuckley:
@patrickvinograd / @hinzed1127 note that if we use this design to set the 'force-auth' auth parameter then we will need an additional change to the FE to append `?inbound=true` to the request whenever we initiate an auto login.  This has the advantage though that a user will never land on an authentication failed page when auto logging in.

lib/saml/errors.rb
```diff
@@ -11,6 +11,9 @@ class UserAttributeError < StandardError
     MHV_ICN_MISMATCH = { code: '103',
                          tag: :mhv_icn_mismatch,
                          message: 'MHV credential ICN does not match MPI record' }.freeze
+    IDME_UUID_MISSING = { code: '104',
```
hinzed1127, line 14: jw are these codes derived from anything we're getting back from other services or just for internal tracking?

patrickvinograd: these don't map to anything else - they are basically the contract between backend and frontend so frontend knows what error page to display to users.

PR #4331, opened 2020-05-26
9365 va file number endpoint

digitaldrk:
## Description of change
This endpoint exists to ~retrieve~ verify a Veteran has a VA File number. The VA File number is necessary to successfully submit a 686 form which we are working on in another ticket. Rather than ask the veteran to provide their VA File number and possibly introduce bad data, retrieve the number, verify it exists, and return a true false for the FE.

If the user does not have a VA File number but does have a participan_id, the VA would like to know for data integrity reasons. We will log the occurrence to Sentry and then store relevant information temporarily in `PersonalInformationLog`.

## Original issue(s)
* https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/9365

* https://github.com/department-of-veterans-affairs/vets-api/pull/2888

    * No settings should be necessary in a 'production-like' environment. A proxy is needed to run locally
    * No feature flag
    * Basic error logging was added
    * no metrics were added
    * Swagger docs were updated to document this endpoint
    * no concerns

This PR includes a spec for the service response and the controller response


app/serializers/lighthouse/people/va_file_number_serializer.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class Lighthouse::People::VaFileNumberSerializer < ActiveModel::Serializer
```
patrickvinograd, line 3: Why the "Lighthouse" in the module name? What is the connection to Lighthouse?

patrickvinograd:
I wouldn't say their are "no PII concerns" IIRC a VA File Number is sometimes == to somebody's SSN? I may be wrong about that, maybe there are APIs that accept a file number or SSN equivalently. But if so, then you want to document clearly in swagger, etc that this is a sensitive piece of PII, it's not just some innocuous VA-only identifier.

app/services/bgs/people_service.rb
```diff
@@ -0,0 +1,57 @@
...
+module BGS
+  class PeopleService
+    include SentryLogging
+    class VaFileNumberNotFound < StandardError; end
+
+    def initialize(current_user)
+      @current_user = current_user
+    end
+
+    def find_person_by_ptcpnt_id
```
patrickvinograd, line 12: why abbreviate `ptcpnt` vs. `participant`?

patrickvinograd:
This seems like it should have some authorization in front of it (see examples under `app/policies` - i.e. only allowing access if the user has a participant id?) Otherwise there are going to be lots of false positives of people who are not known to VBA, are not veterans, etc. Maybe if you're only invoking this endpoint from a specific application that already has such checks, but best practice is for the API to be self-contained it its authorization, not to rely on how it's used.

app/serializers/lighthouse/people/va_file_number_serializer.rb
digitaldrk: The data source we're using for this PR leverages the lighthouse_bgs gem. When I saw there was already a Lighthouse module, seemed to make sense. Do you have a suggestion/preference?

app/services/bgs/people_service.rb
digitaldrk: Good point. I have changed it to participant

app/serializers/lighthouse/people/va_file_number_serializer.rb
patrickvinograd: The lighthouse_bgs gem was named thusly because it was a fork of another Ruby bgs gem, forked because the Lighthouse API team wanted to add some capabilities and not conflict with other uses of that gem (which was a dubious choice IMO). 

There's nothing in this class specific to Lighthouse, you're not part of the Lighthouse program, it just seems like arbitrary name cruft that will confuse future developers.

digitaldrk:
> This seems like it should have some authorization in front of it (see examples under `app/policies` - i.e. only allowing access if the user has a participant id?) Otherwise there are going to be lots of false positives of people who are not known to VBA, are not veterans, etc. Maybe if you're only invoking this endpoint from a specific application that already has such checks, but best practice is for the API to be self-contained it its authorization, not to rely on how it's used.

Great Idea, thank you for the suggestion. I have added authorization to the endpoint.

digitaldrk:
> I wouldn't say their are "no PII concerns" IIRC a VA File Number is sometimes == to somebody's SSN? I may be wrong about that, maybe there are APIs that accept a file number or SSN equivalently. But if so, then you want to document clearly in swagger, etc that this is a sensitive piece of PII, it's not just some innocuous VA-only identifier.

@patrickvinograd My initial understanding with PII concerns were centered around persistence but I understand now that we cannot show PII. You are correct that VA File Number is SSN. For posterity, my understanding is it's not SSN for those Veterans born before 1960.

After speaking with one of the FE devs on my team, we determined it was ok to send a boolean indicating the Veteran does/does not have a VA file number. I have renamed files and route more accurately convey this. WDYT

app/serializers/lighthouse/people/va_file_number_serializer.rb
digitaldrk: Understood, thank you for the explanation. Renamed and placed outside of Lighthouse namespace.

Gemfile.lock
```diff
@@ -124,7 +124,6 @@ PATH
 
 GEM
   remote: https://rubygems.org/
-  remote: https://enterprise.contribsys.com/
```
johnpaulashenfelter, line 127: We use this for sidekiq-ent and pro. You can get the credentials for contribsys from https://github.com/department-of-veterans-affairs/va.gov-team-sensitive/blob/master/platform/engineering/sidekiq-enterprise-setup.md or just reset Gemfile.lock

johnpaulashenfelter: Resetting this will help get the Danger message for amount of changes in a better place :)

johnpaulashenfelter:
@patrickvinograd are you ok with the PII situation?

Gemfile.lock
digitaldrk: @johnpaulashenfelter Thank you for the tip. I wondered about that. I have reverted the gemfile.lock back to it's original state.

app/controllers/v0/profile/valid_va_file_numbers_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+
+        render(
+          json: valid_va_file_number?(response),
+          serializer: ValidVaFileNumberSerializer
+        )
+      end
+
+      private
+
+      def valid_va_file_number?(service_response)
```
erluti, line 20: Conventionally methods that end with `?` return a boolean value.  Maybe this could be named something like `valid_va_file_number_json` or ``valid_va_file_number_data` (or something better)?

config/initializers/inflections.rb
```diff
@@ -2,6 +2,7 @@
 
 ActiveSupport::Inflector.inflections do |inflect|
   inflect.acronym 'AWS'
+  inflect.acronym 'BGS'
```
erluti, line 5: yes! thanks for adding this in!

app/policies/bgs_policy.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+BGSPolicy = Struct.new(:user, :bgs) do
+  def access?
+    if user.icn.present? && user.ssn.present? && user.participant_id.present?
+      log_success
```
erluti, line 6: I think this method is clearer without the private methods
```
  def access?
    if user.icn.present? && user.ssn.present? && user.participant_id.present?
      StatsD.increment('api.bgs.policy.success') if user.loa3?
      true
    else
      StatsD.increment('api.bgs.policy.failure') if user.loa3?
      false
    end
  end
```
I'm worried the return values for `access?` aren't clear because they are buried in the private methods

app/controllers/v0/profile/valid_va_file_numbers_controller.rb
digitaldrk: Ah, I could see how that would be misleading. I figured it would be ok bc the hash value will be true/false. I will make the change

app/policies/bgs_policy.rb
digitaldrk: Cool with me.

Just a FWIW, I 'borrowed heavily' from this file that uses the private method approach. Would you like to maintain parody or have me make the change you requested?

https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/policies/evss_policy.rb

digitaldrk: @erluti I made your requested changes but code climate wasn't a fan. Not sure how to proceed 🤔 .

image

erluti: I think writing clear code is more important than matching code somewhere else.
I tested it against the linter, I didn't think about CC.
My first thought is that this method basically returns the value of the if-statement, so I think pulling that back might help.  And then have a method just for the statsd logging
```
def access?
  accessible = user.icn.present? && user.ssn.present? && user.participant_id.present?
  increment_statsd_for_loa_user(accessible)
  accessible
end

def increment_statsd_for_loa_user(accessible)
  return unless user.loa3?
  if accessible
    StatsD.increment('api.bgs.policy.success')
  else
    StatsD.increment('api.bgs.policy.failure')
  end
end
```

erluti: Looks like my suggestion hit a linting issue.  I'm also surprised it scored `increment_statsd_for_loa_user` as a 6 for cognitive complexity.  I'm okay with that if you are, but the linting issue will block the pipeline.

app/policies/bgs_policy.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+BGSPolicy = Struct.new(:user, :bgs) do
+  def access?
+    accessible = user.icn.present? && user.ssn.present? && user.participant_id.present?
+    increment_statsd_for_loa_user(accessible) if user.loa3?
+    accessible
+  end
+
+  def increment_statsd_for_loa_user(accessible)
```
erluti, line 10: i'd drop `for_loa_user` from this method name if you are pulling out the loa3 check.
(also pull in master to stop the security warning)

app/policies/bgs_policy.rb
digitaldrk: > thanks for responding to my feedback

@erluti Thank you for suggesting them. I didn't see your approval until after I made a change to get past code climate. I'm past CC now. I'll work on the security issue that popped up.

Link for your convenience:

https://github.com/department-of-veterans-affairs/vets-api/pull/4331/commits/cc4ce05bf22ec33aac7b2ad825374840236f8918

app/policies/bgs_policy.rb
digitaldrk: Good call

Gemfile.lock
digitaldrk: @johnpaulashenfelter Checking in with you on this. The PR says it's blocked by this requested change.

PR #4332, opened 2020-05-27
Adding extra page for Contestable issues

bastosmichael:
## Description of change
Add a page in the PDF for additional Contestable Issues (over 6).

## Original issue(s)
https://vajira.max.gov/browse/API-746


bastosmichael:
I've created the tech debt ticket here https://vajira.max.gov/browse/API-1016 for our team to tackle later on, thanks.

PR #4334, opened 2020-05-27
Allow regex for allowed CORS origin domains

jholton:
## Description of change
Allows for configuration of allowed CORS origin domains using regex

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/9445

A default allowed value of `https://www.va.gov` is provided in `settings.yml`.
This will ensure that `Settings.web_origin_regex` is never `nil` and thus the evaluation of `Regexp.new(Settings.web_origin_regex)` won't raise an exception.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -50,7 +50,8 @@ module VetsAPI
     # CORS configuration; see also cors_preflight route
     config.middleware.insert_before 0, Rack::Cors, logger: (-> { Rails.logger }) do
       allow do
-        origins { |source, _env| Settings.web_origin.split(',').include?(source) }
+        regex = Regexp.new(Settings.web_origin_regex)
+        origins { |source, _env| Settings.web_origin.split(',').include?(source) || source.match?(regex) }
         resource '*', headers: :any,
                       methods: :any,
                       credentials: true,
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -3,6 +3,7 @@ virtual_hosts: ["127.0.0.1", "localhost"] # Safe host names
 
 # For CORS requests; separate multiple origins with a comma
 web_origin: http://localhost:3000,http://localhost:3001,http://127.0.0.1:3000,http://127.0.0.1:3001,null
+web_origin_regex: \Ahttps?:\/\/www\.va\.gov\z
 
 # Settings for SAML authentication
 saml:


config/application.rb
```diff
@@ -50,7 +50,8 @@ class Application < Rails::Application
     # CORS configuration; see also cors_preflight route
     config.middleware.insert_before 0, Rack::Cors, logger: (-> { Rails.logger }) do
       allow do
-        origins { |source, _env| Settings.web_origin.split(',').include?(source) }
+        regex = Regexp.new(Settings.web_origin_regex)
+        origins { |source, _env| Settings.web_origin.split(',').include?(source) || source.match?(regex) }
```
jholton, line 54: This is now evaluating the fully qualified domain strings in `Settings.web_origin` and the regex domains in `Settings.web_origin_regex`
Any merit to converting all domains to the regex version and eliminating the `Settings.web_origin` strings? Or just leave both options?

config/settings.yml
```diff
@@ -3,6 +3,7 @@ virtual_hosts: ["127.0.0.1", "localhost"] # Safe host names
 
 # For CORS requests; separate multiple origins with a comma
 web_origin: http://localhost:3000,http://localhost:3001,http://127.0.0.1:3000,http://127.0.0.1:3001,null
+web_origin_regex: \Ahttps?:\/\/www\.va\.gov\z
```
jholton, line 6: added as a default value.  Environment specific values will be added to the env specific `settings.local.yml` files in the `devops` repo

config/application.rb
thilton-oddball: Yeah, I think it might be a little cleaner and simpler to have them all in the regex. The regex format that is being used is slightly different than what I am used to and I don't know how advanced you can get with a regex in a yaml file, but if you add them all in, the regex might look something like this (I think): 

web_origin_regex: \Ahttps?:\/\/(localhost:300\d|127\.0\.0\.1:300\d|www\.va\.gov)\z

regex = Regexp.new(Settings.web_origin_regex)
origins { |source, _env| source.nil? || source.match?(regex) }

jholton: Will merge as-is for now.  Converting it over will have to be a multi-step task:
* merge this
* change configs to use regex
* then we can remove the `Settings.web_origin` if we want.

gunsch: drive-by comment: it might be "simpler" to have everything in one place, but I do worry about making that the default since there are more ways this can be implemented incorrectly with regexes than exact hostname strings. (and failures to do it correctly could be a potential security concern)

PR #4336, opened 2020-05-28
fix regex in rake form526:errors

annaswims:
## Description of change
the regex in `form526.rake` was using `\\` to escape characters instead of `\`, so nothing matched.  Therefore, most errors weren't being included in this report. 

And, while I was in the neighborhood I got rid of all single-letter variables.

## Original issue(s)

this is a rake task that is occasionally run by developers to report on errors.  It does not change data.

### DIFF
--- a/rakelib/form526.rake
+++ b/rakelib/form526.rake
@@ -123,7 +123,7 @@ namespace :form526 do
     # errors in a message. Each error will have a `key` and a `text` key. The
     # following regex will group all key/text pairs together that are present in
     # the string.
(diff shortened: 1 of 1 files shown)


annaswims:
@SarahJaine once this is merged and we run `rake form526:errors[2020-05-26,2020-05-27] ` we'll see a report that includes exact error counts by participant. 

````
form526.treatments[0].center.name.Pattern: must match \
*****************
Unique Participant ID count: 20
submission_id:       participant_id:      created_at:
XXXXXXXXX              XXXXXXXXX             2020-05-26 11:11:16 UTC

PR #4345, opened 2020-05-29
require sidekiq-pro and sidekiq-enterprise if they're available

annaswims:
## Description of change
I wanted to run form 526 locally, but it uses `Sidekiq::Batch`, which is a pro feature.  I have sidekiq-pro/sidekiq-enterprise installed, but I was seeing an error (something like Sidekiq::Batch method not found). ~~Probably because we're not eager loading in dev  So, I'm requiring them if they're available..~~ because those gems are in the prod group.
## Original issue(s)

before or with no pro/ent gems:
Screen Shot 2020-05-29 at 11 28 56 AM
after (I verified the is a pro banner, but no enterprise banner):
Screen Shot 2020-05-29 at 11 24 09 AM
Bonus irrelevant image:


johnpaulashenfelter:
@annaswims I think it might also make sense to move the gems to the main/default group instead of `:production` so you can run it in development and test as well _if they are installed_. Then you likely avoid the loading problem as well.

This came up with @charleystran not long ago as well and its a pretty big blocker until you realize the environment rails is running in matters for that gem. 

Our currently situation almost certainly means we're not testing the places we use enterprise or pro features as well.

If moving the gems to the default group solves your problem, let's do that!

johnpaulashenfelter:
Thanks @annaswims -- going to take a quick look to make sure docker doesn't complain about this and see where things end up.

annaswims:
I've updated the PR to put the gems in the default env and to use them in tests when they're available.

spec/jobs/central_mail/submit_form4142_job_spec.rb
```diff
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
 require 'rails_helper'
-require 'sidekiq/testing'
```
annaswims, line 4: redundant. 
`require 'sidekiq/testing'
Sidekiq::Testing.fake!` is in rails_helper

annaswims:
Edit: this PR opened a can of worms for how we test code that uses`Sidekiq::Batch`, a Pro feature.  When testing with `Sidekiq::Testing.fake!`,   `Sidekiq::Batch.new.jobs do ...` does not return job id's. Looking at that lead me to see that we've got redundant includes and were defining `Sidekiq::Batch` twice

PR #4360, opened 2020-06-02
[9681] Fix MDT Eligibility Response

camerontesterman:
## Description of change
The response from the DLC that helps determine eligibility is upcased. This change downcases it to conform to the virtus model.

This change also adds a DLC field - vetEmail, to both show the veteran the email that they have on file for them from the DLC, and allow them to edit it as well.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/9681

## Testing
Has been tested on a review instance


lib/mdot/response.rb
```diff
@@ -30,11 +32,8 @@ def determine_eligibility
       eligibility = MDOT::Eligibility.new
 
       supplies.each do |supply|
-        group = supply.product_group.split.last.to_sym
-
-        if eligibility.attributes.key?(group)
-          eligibility.attributes = { group => supply.next_availability_date <= Time.zone.today }
-        end
+        group = supply.product_group.downcase.to_sym
```
thilton-oddball, line 35: Based on what it was previously, does not doing a split on product_group and grabbing the last element have the potential to cause issues? Not familiar with what a typical value would be so just trying to understand.

camerontesterman: So the product_groups were changed a little bit. Prior to this they were like "hearing aid batteries" or "hearing aid accessories" so we would have to pull out the last word and symbolize it, but now they are like "BATTERIES" so doing the split is no longer necessary.

PR #4362, opened 2020-06-02
change routing error code from 411 to 404

jholton:
## Description of change
change routing error code from 411 to 404

## Original issue(s)

### DIFF
--- a/config/locales/exceptions.en.yml
+++ b/config/locales/exceptions.en.yml
@@ -116,7 +116,7 @@ en:
         <<: *defaults
         title: Not found
         detail: "There are no routes matching your request: %{path}"
-        code: 411
+        code: 404
         status: 404
       unknown_format:
         <<: *defaults


config/locales/exceptions.en.yml
```diff
@@ -116,7 +116,7 @@ en:
         <<: *defaults
         title: Not found
         detail: "There are no routes matching your request: %{path}"
-        code: 411
```
jholton, line 119: @saneshark I know it's been 4 years, but do you remember why this was set as 411 instead of 404?

saneshark: code and http status are distinct -- CODE should be unique to detail.

saneshark: I characterize code as a set of exception codes that VA.gov can raise / return, totally distinct from HTTP Status code. In fact, we have codes of "SM101" etc for that very reason as well with no direct correlation to HTTP Status.

jholton: Thanks for the clarification.  How/where i the `code` used?

jholton:
Not merging after receiving clarification of use of `code`

config/locales/exceptions.en.yml
saneshark: exceptions.en.yml file deliberately done with future localization in mind, so that a type of error would have the same code, but you could in theory have a spanish, english, japanese version if you wanted. And of course, any API consumer would then have a means of reliably overriding the VA.gov specific language should they want to as well.

PR #4363, opened 2020-06-02
add EVSS/AWS/ReferenceData to breakers

annaswims:
## Description of change
configure breakers so that we can see stats for EVSS reference data stats in [Backend Service Report ](http://grafana.vfs.va.gov/d/000000050/backend-service-report?orgId=1)

https://github.com/department-of-veterans-affairs/va.gov-team/issues/8387

### DIFF
--- a/config/initializers/breakers.rb
+++ b/config/initializers/breakers.rb
@@ -45,6 +45,7 @@ services = [
   EVSS::PCIUAddress::Configuration.instance.breakers_service,
   EVSS::GiBillStatus::Configuration.instance.breakers_service,
   EVSS::Dependents::Configuration.instance.breakers_service,
+  EVSS::ReferenceData::Configuration.instance.breakers_service,
   Gibft::Configuration.instance.breakers_service,
   VIC::Configuration.instance.breakers_service,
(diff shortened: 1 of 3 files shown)


annaswims:
@omgitsbillryan it looks like you've done some work in #4032 and before that, but has EVSS::ReferenceData always been  [disabled](https://github.com/department-of-veterans-affairs/devops/search?q=reference_data_service&unscoped_q=reference_data_service)?  We're looking at using it for a new endpoint for discharge location.

omgitsbillryan:
As far as I can remember, this has always been disabled. I remember setting up this integration with the idea that getting a list of countries & states would be the first simple API to use as proof of concept. I think the "reference data" endpoints switched over to using JWT auth over the weird `va_auth_header_*` values that I'm not even sure exist anymore.

annaswims:
well, that answers one question I had. I was going to ask if the countries/states endpoints have been used before.  We're looking at getting separation location data from EVSS, which lead me to wonder how often this service was used.
https://github.com/department-of-veterans-affairs/va.gov-team/issues/9735

omgitsbillryan:
The list of countries we get from EVSS was being used by Letters last I knew.

PR #4364, opened 2020-06-03
fix for callback stats being sent before upleveling

ericbuckley:
- Callback stats should be sent every time out IDP sends back a SAML response.  However in the case where a user needs to be upleveled to LOA3, these stats we're not being sent.

- Updating the tag permutations on the `STATSD_LOGIN_STATUS` metrics to match those of `STATSD_SSO_NEW_KEY`.  We want to create grafana graphs of success rate, and having the tags match exactly on these stats gives us a better glance at whats happening.

- tackling some tech debt in the `V1::SessionController::saml_callback` function, by consolidating the 3 different failure blocks into 1.


app/controllers/v0/sessions_controller.rb
```diff
@@ -108,10 +109,12 @@ def user_login(saml_response)
         after_login_actions
         redirect_to url_service(user_session_form.saml_uuid).login_redirect_url
```
patrickvinograd, line 110: Can we just move `callback_stats(:success, saml_response)` to here, out of the conditional, and remove both the new else block and the callback_stats call in login_stats_success?

ericbuckley: I tried addressing all the calls to `callback_stats` in this commit, but may have gotten a little  carried away.  Let me know what you think, I can revert or apply the same changes to v0.  https://github.com/department-of-veterans-affairs/vets-api/pull/4364/commits/5bf2e03189e9c20303c575b9fc3ced7ff018edf8

lib/saml/url_service.rb
```diff
@@ -213,11 +213,12 @@ def build_redirect(params)
     def initialize_tracker(params, previous_saml_uuid: nil)
       previous = previous_saml_uuid && SAMLRequestTracker.find(previous_saml_uuid)
       redirect = previous&.payload_attr(:redirect) || build_redirect(params)
-      inbound = params[:inbound] || nil
+      inbound = previous&.payload_attr(:inbound) || params[:inbound] || nil
```
erluti, line 216: Can `params[:inbound]` be `false`?  Otherwise the `|| nil` is unnecessary.

app/controllers/v1/sessions_controller.rb
```diff
@@ -171,34 +158,27 @@ def user_logout(saml_response)
...
     end
 
-    def login_stats(status, saml_response, user_session_form = nil)
+    def login_stats(status, _saml_response, user_session_form)
+      tracker = url_service(user_session_form&.saml_uuid).tracker
+      type = tracker.payload_attr(:type)
       case status
       when :success
-        login_stats_success(saml_response, user_session_form)
+        tags = ["context:#{type}", VERSION_TAG]
```
erluti, line 173: `tags` here and in the failure case have thee same value, these could move out of the case and be defined after `type`

app/controllers/v1/sessions_controller.rb
```diff
@@ -113,12 +111,10 @@ def saml_settings(options = {})
       SAML::SSOeSettingsService.saml_settings(options)
     end
 
-    def auth_error_code(code)
-      if code == '005' && validate_session
-        UserSessionForm::ERRORS[:saml_replay_valid_session][:code]
-      else
-        code
-      end
+    def saml_error(form)
```
erluti, line 114: what do you think about calling this something like `raise_saml_error`?  It's not obvious below when this is called that it's going to interrupt control flow

lib/saml/url_service.rb
ericbuckley: yeah re-reading that I agree, `|| nil` doesn't add anything

ericbuckley:
@erluti I made some changes based on your comments.  How does it look now?

PR #4369, opened 2020-06-08
fix for partial birthdates

ericbuckley:
it's possible that only a partial birthdate will comeback from the SSOe
SAML response, for example the MVI record may only have the user's birth
year and not the month or day.  When we only have a partial birth date
we should return nil.

fixes: http://sentry.vfs.va.gov/vets-gov/platform-api-production/issues/130015/


ericbuckley:
> > for example the MVI record may only have the user's birth year and not the month or day.
> 
> How common is this? Is there any merit to trying to salvage what we _do_ get back from mvi as opposed to returning `nil`?
> 
> LGTM either way, but wondering what the consequences of this are.

Over the past 6 days, we've seen it happen once over the ~1900 authentication requests we processed, so it's not common.  

As for returning the data we have, I opted to return `nil` because I thought it would be safer that way.  If we return a year, and the FE application somewhere assumes this value is a date, I thought that might raise an error.   However, I have little context for knowing where on the FE this value is used, so if someone knows more about this I'd be open to returning a difference value.

omgitsbillryan:
I think `nil` is the safe way to go. I might even argue for some kind `ActiveModel::Validations` code to validate incoming data from MVI - but that's outside the scope of your PR here.

PR #4370, opened 2020-06-08
Add endpoint for EVSS 

annaswims:
## Description of change
For form526 BDD we need to add separation location, we can get that data from the EVSS reference data endpoint intakesites.  

Previously, there were two other endpoints that were calling EVSS reference data endpoints through an old/unused authentication scheme so I stripped that out. That means we no longer need the sockisify gem.

## Original issue
department-of-veterans-affairs/va.gov-team#9735


app/controllers/v0/apidocs_controller.rb
```diff
@@ -173,7 +173,6 @@ class ApidocsController < ApplicationController
       Swagger::Schemas::ConnectedApplications,
       Swagger::Schemas::Dependents,
       Swagger::Schemas::DismissedStatus,
-      Swagger::Schemas::Form526::RatedDisabilities,
```
annaswims, line 176: duplicate

config/settings.yml
```diff
@@ -194,9 +195,6 @@ evss:
     cert_path: ~
     key_path: ~
     root_ca: ~
-  jwt:
```
annaswims, line 197: unused

config/settings.yml
```diff
@@ -444,10 +442,6 @@ betamocks:
   cache_dir: /cache
   services_config: config/betamocks/services_config.yml
 
-faraday_socks_proxy:
```
annaswims, line 447: unused - TODO: check if I need to remove from devops repo.

spec/request/address_request_spec.rb
```diff
@@ -166,18 +167,5 @@
         end
       end
     end
-
-    context 'with a 401 malformed token response', vcr: { cassette_name: 'evss/reference_data/401_malformed' } do
```
erluti, line 170: how come this test/cassette was removed?

annaswims: There used to be header/JWT code that was unique to this controller, now it uses the same header that the rest of the EVSS calls use.

annaswims:
this pr will fix the EVSS connection in review instances https://github.com/department-of-veterans-affairs/devops/pull/7107

spec/request/swagger_spec.rb
```diff
@@ -517,6 +517,13 @@
         end
       end
 
+      it 'supports getting separation_locations' do
+        expect(subject).to validate(:get, '/v0/disability_compensation_form/separation_locations', 401)
```
annaswims, line 521: Reconsider this test

annaswims, line 521: @SarahJaine What I missed earlier when we were talking was the `headers` argument.  The 401 response is unauthenticated.

PR #4375, opened 2020-06-09
 logs service exception to sentry

shefalinayak:
## Description of change

TokensController has a custom exception handler for token parsing and network errors. We are now logging the exceptions to Sentry.

## Original issue(s)
[department-of-veterans-affairs/covid19-chatbot#258]

- Adds Sentry error logging


app/controllers/v0/coronavirus_chatbot/tokens_controller.rb
```diff
@@ -61,7 +61,8 @@ def bearer_token
       end
 
       def service_exception_handler(exception)
-        logger.error "An error occured with the microsoft service that issues chatbot tokens #{exception.message}"
+        logger.error "An error occurred with the Microsoft service that issues chatbot tokens #{exception.message}"
+        log_exception_to_sentry(exception, {}, {}, :error)
```
johnpaulashenfelter, line 65: I noticed this weird method signature during some work before vacation and started a refactor branch. Nearly every other method just uses `log_exception_to_sentry(exception)` so would suggest you do the same here.

Also the `log_exception_to_sentry` method is going to trigger a log entry (though with a different message) so maybe you want to add a tag or extra context instead of the `logger.error` text?

PR #4379, opened 2020-06-10
updates to use zeitwerk loader

philipwritescode:
# DO NOT MERGE
✋✋✋✋✋✋✋ 🛑 🛑 🛑 🛑 🛑 🛑 🛑 🛑 🛑 🔥🔥🔥🔥🔥🔥
Due to the size of these changes, they are going to be broken into multiple PRs.  This will remain as a baseline to make sure nothing changed here is missed.
The [issue](department-of-veterans-affairs/va.gov-team#9534) is going to be used as an epic and new tickets to break out this work should be available there.

## Description of change
[Rails 6 introduced the new "zeitwerk" loader for auto-loading files in rails.](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html). It depends on using standard Ruby file and folder naming conventions.  So to use `zeitwerk` changes had to be made to better conform with Ruby on Rails standards.
* files moved/renamed to match Module and Class names
  * `common/exceptions` now uses the `Internal` and `External` namespaces
  * files defining multiple classes had to be broken up
  * EVSS::PCIUAddress got it's own file to define the namespace and the `build_address` method was moved here in order to unwind some interdependent files
* inflections added to support files and classes based on acronyms (this resulted in class and file name changes for consistency)
  * inflecting VA resulted in `lib/pdf_fill/extras_generator.rb`'s use of `humanize` capitalizing VA when it had lowered it before.  So `spec/fixtures/pdf_fill/21P-527EZ/overflow_extras.pdf` and `spec/fixtures/pdf_fill/21P-530/overflow_extras.pdf` were updated to reflect the now-capitalized VA
* `require_dependency` removed
* `lib` directory is no longer in the auto-load path
  * `requires` needed for including `lib` files were added to appropriate files
* notably, `MVI` lib was clashing with the autoloaded `MVI` model, so `MVI` lib became `MasterVeteranIndex`
* various changes resulted in triggering some linting rules (usually related to line length), resulting in some style changes

## Original issue(s)
department-of-veterans-affairs/va.gov-team#9534

Apologies in advance for the size of the PR.  These changes were discovered working through the resulting errors and failing tests from enabling `zeitwerk`.

### DIFF
--- a/app/mailers/spool_submissions_report_mailer.rb
+++ b/app/mailers/spool_submissions_report_mailer.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'reports/uploader'
+
 class SpoolSubmissionsReportMailer < ApplicationMailer
   REPORT_TEXT = 'Spool submissions report'
   RECIPIENTS = %w[
--- a/app/mailers/year_to_date_report_mailer.rb
+++ b/app/mailers/year_to_date_report_mailer.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'reports/uploader'
+
 class YearToDateReportMailer < ApplicationMailer
   REPORT_TEXT = 'Year to date report'
 


omgitsbillryan:
Would it at all be possible/practical to break this up into smaller PRs? I'd like to review the code itself, but it's kind of impossible with the size it is right now.

erluti:
> Would it at all be possible/practical to break this up into smaller PRs? I'd like to review the code itself, but it's kind of impossible with the size it is right now.

Practical is the big question.  The work was done in the order issues were discovered.  I think parts of it could possibly be broken into smaller pieces, but I think those smaller PRs would have to be regenerated by hand because there's no obvious set of commits to move over.  It wouldn't surprise me if it took a whole sprint or more to split this work into 5-10 PRs.  I like the increased confidence of having a series of PRs to review.  But is it worth the time?  I think there's an additional risk that some of these pieces would require additional changes to be added in `classic` autoloader mode, that would then be removed when the final PR to turn on `zeitwerk` was added.
All that to say: 🤷 
I'd be happy to discuss pros and cons as a team to decide what the best path forward would be.

erluti:
I thought some more about what breaking this into multiple PRs would look like.  I think 5-10 was quite the underestimate.
* there would be a PR for each of the inflections (23)
• a PR for renaming the `MVI` lib to `MasterVeteranIndex`
• a PR for adding Internal/External namespacing to exceptions in lib
• There are other namespace issues related to modules matching file paths that would be a handful of separate PRs
• removing of autoloading the lib directory would be it's own PR and would touch a lot of the files this giant PR already touches
• finally a PR for zeitwerk, where there's a bunch of adding adding requires and removing requires_dependency

Making each of these changes visible to review independently would be a big win, but I think it would take a lot of time.

omgitsbillryan:
> there would be a PR for each of the inflections (23)

I'd be happy to just have just 1 PR for this. Even though it would exceed 250 LoC, it would still be manageable to review since only one logical thing would be happening in it.

Any code that can be pulled into it's own PR is a win to me, even if you can't break all of it up. Are there any low hanging fruit that could be easily pulled out that would at lease _reduce_ the size of this PR?

PR #4380, opened 2020-06-11
Add Optional Base64 encoding blurb

mlvidart:
Update description with optional Base64 instructions and links to cURL and raw Base64 examples for individual PDFs. 

The team decided to call out this option in the description because it works differently from the benefits-intake API and it generated some confusion among users. The proposed change addresses the issue. 

@bastosmichael feel free to change write-up if we need to add any more details.


ryanlink:
I've clarified a few paragraphs of the content here with the commit above - @bastosmichael please confirm accuracy.  thanks!

PR #4382, opened 2020-06-11
form 526 BDD - Good enough for FE dev

annaswims:
## Original issue(s)

### DIFF
--- a/app/workers/evss/disability_compensation_form/submit_form526_cleanup.rb
+++ b/app/workers/evss/disability_compensation_form/submit_form526_cleanup.rb
@@ -4,7 +4,7 @@ module EVSS
   module DisabilityCompensationForm
     class SubmitForm526Cleanup < Job
       include Sidekiq::Worker
-
+      # TODO: AEC
       FORM_ID = '21-526EZ'
       STATSD_KEY_PREFIX = 'worker.evss.submit_form526_cleanup'
 
--- a/lib/evss/disability_compensation_form/data_translation_all_claim.rb
+++ b/lib/evss/disability_compensation_form/data_translation_all_claim.rb
@@ -575,7 +575,8 @@ module EVSS
 
       def application_create_date
         # Application create date is the date the user began their application
-        @acd ||= InProgressForm.where(form_id: VA526ez::FORM_ID, user_uuid: @user.uuid)
+        # TODO AEC
+        @acd ||= InProgressForm.where(form_id: FormProfiles::VA526ez::FORM_ID, user_uuid: @user.uuid)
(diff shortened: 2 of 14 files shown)


app/workers/evss/disability_compensation_form/submit_form526_cleanup.rb
```diff
@@ -4,7 +4,7 @@ module EVSS
   module DisabilityCompensationForm
     class SubmitForm526Cleanup < Job
       include Sidekiq::Worker
-
+      # TODO: AEC
       FORM_ID = '21-526EZ'
```
Mottie, line 8: `21-526EZ` is version 1, and shouldn't be in use any more.

ndsprinkle: @Mottie I'm pretty sure version 2 uses this ID as well. They share IDs.

annaswims: I verified that all the in progress 526 forms have id 21-526EZ (before we implement bed)

lib/evss/disability_compensation_form/data_translation_all_claim.rb
```diff
@@ -575,7 +575,10 @@ def greater_itf_date?
 
       def application_create_date
         # Application create date is the date the user began their application
-        @acd ||= InProgressForm.where(form_id: VA526ez::FORM_ID, user_uuid: @user.uuid)
+        # TODO AEC
+        @acd ||= InProgressForm.where(form_id: [FormProfiles::VA526ez::FORM_ID, FormProfiles::VA526ezbdd::FORM_ID],
```
annaswims, line 579: This is awkward, but we don't have the InProgressForm id at this point.  Maybe we could just get the FE to send the InProgressForm id on submit.

Mottie: Could you just check the service dates to determine if it's BDD or a normal submission?

annaswims:
I made an issue to do more broad testing of BDD once we get all of the pieces ready.
https://github.com/department-of-veterans-affairs/va.gov-team/issues/10368

lib/evss/disability_compensation_form/data_translation_all_claim.rb
annaswims: @SarahJaine had a talk about this yesterday.  I think we're good here.  Most likely, there'll only be one form InProgressForm and in all cases, the one they're submitting will be the most recently edited one.

PR #4383, opened 2020-06-12
Updating homelessSituationType and homelessnessRiskSituationType enums

bastosmichael:
## Description of change
We needed to add standard enums to the homelessness section of the 526 because previously it was just considered an open field. Also added homelessnessRisk factors.

## Original issue(s)
https://vajira.max.gov/browse/API-1081


modules/claims_api/config/schemas/526.json
```diff
@@ -294,7 +293,33 @@
               ],
               "properties": {
                 "homelessSituationType": {
-                  "type": "string"
+                  "type": "string",
+                  "default": "OTHER",
+                  "enum": ["FLEEING_CURRENT_RESIDENCE",
```
charleystran, line 298: Correct me if I am wrong here, but I thought the point of this was to use the translation keys here https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/evss/disability_compensation_form/data_translation_all_claim.rb#L13 and not the values?

modules/claims_api/spec/fixtures/form_526_evss.json
```diff
@@ -43,13 +43,21 @@
         ]
```
edmangimelli, line 43: on line 26 of this file, the value "Air Force Reserve" is being used.
that should be "Air Force Reserves" (plural) to match the enum

edmangimelli: also, this fixture is missing the "standardClaim" field, which is required by modules/claims_api/config/schemas/526.json

modules/claims_api/spec/fixtures/form_526_evss.json
```diff
@@ -43,13 +43,21 @@
...
+    "homelessness": {
+      "pointOfContact": {
+        "pointOfContactName": "John Doe",
+        "primaryPhone": "555-555-5555"
+      },
+      "currentlyHomeless": {
+        "homelessSituationType": "FLEEING_CURRENT_RESIDENCE",
+        "otherLivingSituation": "other living situations"
+      }
+    },
```
edmangimelli, line 55: the "homelessness" object needs to be moved under the "veteran" top-level object. that's why it looks like it's not obeying the `"required": [ "pointOfContact" ],` constraint

modules/claims_api/spec/fixtures/form_526_json_api.json
```diff
@@ -16,6 +16,16 @@
...
+      "homelessness": {
+        "pointOfContact": {
+          "pointOfContactName": "John Doe",
+          "primaryPhone": "555-555-5555"
+        },
+        "currentlyHomeless": {
+          "homelessSituationType": "FLEEING_CURRENT_RESIDENCE",
+          "otherLivingSituation": "other living situations"
+        }
+      },
```
edmangimelli, line 28: same situation. needs to be moved under "veteran"

modules/claims_api/spec/fixtures/form_526_evss.json
```diff
@@ -43,13 +43,21 @@
         ]
       }
     ],
+    "homelessness": {
+      "pointOfContact": {
+        "pointOfContactName": "John Doe",
+        "primaryPhone": "555-555-5555"
```
edmangimelli, line 49:
```suggestion
          "primaryPhone": {
            "areaCode": "555",
            "phoneNumber": "555-5555"
          }
```

modules/claims_api/spec/fixtures/form_526_json_api.json
```diff
@@ -16,6 +16,16 @@
         }
       },
       "claimantCertification": true,
+      "homelessness": {
+        "pointOfContact": {
+          "pointOfContactName": "John Doe",
+          "primaryPhone": "555-555-5555"
```
edmangimelli, line 22:
```suggestion
          "primaryPhone": {
            "areaCode": "555",
            "phoneNumber": "555-5555"
          }
```

modules/claims_api/spec/fixtures/form_526_json_api.json
bastosmichael: Thanks for catching this.

modules/claims_api/config/schemas/526.json
bastosmichael: Fixed

modules/claims_api/spec/fixtures/form_526_evss.json
bastosmichael: Fixed

PR #4388, opened 2020-06-15
JSON responses are deduplicated based on ID

darkbushido:
## Description of change
fixing an issue where PPMS gives the same ProviderIdentifier for several different facilities. We use the ProviderIdentifier as the ID causing the extra responses to be deduplicated because of the ID. The API calls that do not have unique ProviderIdentifiers will use an MD5 hash of all of the attributes for the ID

## Original issue(s)
department-of-veterans-affairs/va.gov-team#9049

* This can be turned on with the feature flag `facility_locator_ppms_forced_unique_id`


erluti:
Is it possible to add a test for some data that used to comeback incorrectly having the correct list with this new way of generating ids?

darkbushido:
That was added `spec/request/v0/facilities/ccp_request_spec.rb`

app/models/provider.rb
```diff
@@ -29,6 +30,14 @@ def <=>(other)
     self.Miles <=> other.Miles
   end
 
+  def hexdigest
+    Digest::MD5.hexdigest(attributes.except(:ProviderHexdigest).to_a.join('|'))
```
f1337, line 34: Copied from Slack:

Could we use a SHA digest instead of MD5? I’m not really up-to-speed on the performance constraints.

In newer Ruby versions, SHA-1 is as fast or faster than MD5. And SHA-256 is only 2x slower: https://gist.github.com/morimori/1330095

app/serializers/provider_serializer.rb
```diff
@@ -4,7 +4,12 @@ class ProviderSerializer < ActiveModel::Serializer
   type 'cc_provider'
 
   def id
-    "ccp_#{object.ProviderIdentifier}"
+    if Flipper.enabled?(:facility_locator_ppms_forced_unique_id)
+      obj_id = object.ProviderHexdigest || object.ProviderIdentifier
+      "ccp_#{obj_id}"
+    else
+      "ccp_#{object.ProviderIdentifier}"
+    end
```
f1337, line 12: A minor re-factoring suggestion, but non-blocking:

```suggestion
obj_id = object.ProviderHexdigest if Flipper.enabled?(:facility_locator_ppms_forced_unique_id)
obj_id ||= object.ProviderIdentifier
"ccp_#{obj_id}"
    end
```

erluti:
> That was added `spec/request/v0/facilities/ccp_request_spec.rb`

I didn't see any new expectations.  I thought there might be one that specifically aligned with the ticket.

spec/request/v0/facilities/ccp_request_spec.rb
```diff
@@ -279,7 +279,57 @@
               }
             },
             {
-              'id' => 'ccp_1447660816',
+              'id' => 'ccp_1c8a8512f5daf1046e950a6ee2f3af8a350d0f3281f6fd85a60494e11c78fcce',
```
darkbushido, line 282: This is an entry that was previously unavailable due to the deduplication by ID

spec/request/v0/facilities/ccp_request_spec.rb
```diff
@@ -279,7 +279,57 @@
...
+                'name' => 'NextCare Urgent Care Ocotillo',
+                'phone' => nil,
+                'pos_codes' => '20',
+                'pref_contact' => nil,
+                'specialty' => [],
+                'unique_id' => '1043371826'
+              }
+            },
+            {
+              'id' => 'ccp_efa5302e3e4a62554562b8a8617d9a67255b8fcb2cb1052e8fe10c859186baff',
```
darkbushido, line 307: This is an entry that was previously unavailable due to the deduplication by ID

spec/request/v0/facilities/ccp_request_spec.rb
```diff
@@ -279,7 +279,57 @@
...
+                'name' => 'MinuteClinic LLC',
+                'phone' => nil,
+                'pos_codes' => '17',
+                'pref_contact' => nil,
+                'specialty' => [],
+                'unique_id' => '1629245311'
+              }
+            },
+            {
+              'id' => 'ccp_bb40a8b55b4b30c42d28efa923d8508da71f06a1d9a5eb5d0343ab9a9e998bc8',
```
darkbushido, line 332: This is an entry that was previously unavailable due to the deduplication by ID

spec/request/v0/facilities/ccp_request_spec.rb
```diff
@@ -304,7 +354,57 @@
               }
             },
             {
-              'id' => 'ccp_1871782490',
+              'id' => 'ccp_f4e0d537b1279f239622a3c635bc4a1c9b0cfab9465c3552b366d8fa5c1fe83d',
```
darkbushido, line 357: This is an entry that was previously unavailable due to the deduplication by ID

spec/request/v0/facilities/ccp_request_spec.rb
```diff
@@ -304,7 +354,57 @@
...
+                'name' => 'Take Care Health Arizona PC',
+                'phone' => nil,
+                'pos_codes' => '17',
+                'pref_contact' => nil,
+                'specialty' => [],
+                'unique_id' => '1992993570'
+              }
+            },
+            {
+              'id' => 'ccp_1376b2f242026ed068af34f4e24531f1eab3a5043c402d84bcde4e69804387c4',
```
darkbushido, line 382: This is an entry that was previously unavailable due to the deduplication by ID

spec/request/v0/facilities/ccp_request_spec.rb
```diff
@@ -327,6 +427,31 @@
                 'specialty' => [],
                 'unique_id' => '1871782490'
               }
+            },
+            {
+              'id' => 'ccp_14162e717eb34eb738c02bfc294c66eb99fc98fd7e285aa6a8cc627128eb55d5',
```
darkbushido, line 432: This is an entry that was previously unavailable due to the deduplication by ID

spec/request/v0/facilities/ccp_request_spec.rb
```diff
@@ -304,7 +354,57 @@
...
+                'name' => 'MinuteClinic LLC',
+                'phone' => nil,
+                'pos_codes' => '17',
+                'pref_contact' => nil,
+                'specialty' => [],
+                'unique_id' => '1629245311'
+              }
+            },
+            {
+              'id' => 'ccp_0b766787f4c9c9e5b40d3b45191f9aef28bfad6b568b1104870a41d2d0ab168f',
```
darkbushido, line 407: This is an entry that was previously unavailable due to the deduplication by ID

darkbushido:
This does not have any new tests but an existing test shows the bug, I have commented on each entry in the test expectation that was previously unavailable due to the deduplication based on ID. This is also not a new bug and would be an issue in Any response where the ID is not unique.

config/features.yml
```diff
@@ -30,6 +30,9 @@ features:
   facility_locator_ppms_location_query:
     actor_type: user
     description: Use the Lat/Long instead of an address
+  facility_locator_ppms_forced_unique_id:
+    actor_type: user
+    description: Use an MD5 sum for the ID on PPMS Place of Service Calls
```
f1337, line 35: Do we want to update this to read SHA256?

app/serializers/provider_serializer.rb
darkbushido: so you really don't like `obj_id = object.ProviderHexdigest || object.ProviderIdentifier`

erluti: personally, I think the `if ... else` structure is clearer and I'm fine with the `||`.  But I think as a matter of style both are acceptable

PR #4384, opened 2020-06-12
Create endpoint for MVI proxy add

SarahJaine:
## Description of change

Create new endpoint to make orchestrated search call to MVI to add users who are missing `participant_id` and `birls_id` or just `participant_id`.  This functionality was previously a before action on the `in_progress_forms` endpoint.  For reasons detailed in the attached ticket, this functionality needed to be moved up in the user flow.  The plan now if for the FE to call this new endpoint before the `intent_to_file` endpoint if the user has `ADD_PERSON` in the services list.

## Original issue(s)
First PR for department-of-veterans-affairs/va.gov-team#9856
Second PR for this issue is https://github.com/department-of-veterans-affairs/vets-api/pull/4390

## Things to know about this PR

**Feedback requested:**
- I'm not at all attached to the name of this endpoint, it's kind of stupid.  Can you think of something more apt?
- Check out all my comments on specific lines of code below

Changes
- [x] remove proxy add from `in_progress_forms` endpoint
- [x] removes proxy add description from  swagger docs for `in_progress_forms` endpoint
- [x] create new swagger docs for this endpoint
- [x] moves/renames evss `access_original_claims?` policy to mvi `access_add_person?` policy 
- [x] write new tests for `access_add_person?` policy
- [x] rename `ORIGINAL_CLAIMS` attribute in services list

PR #4390, opened 2020-06-15
Remove MVI proxy add from InProgressFormsController

SarahJaine:
## Description of change

**Merge after https://github.com/department-of-veterans-affairs/vets-api/pull/4384 is merged and associated front end work, tracked in https://github.com/department-of-veterans-affairs/va.gov-team/issues/10237, is complete.**

In https://github.com/department-of-veterans-affairs/vets-api/pull/4384 we added a new endpoint, `POST /v0/mvi_users/{id}` that contains the mvi proxy add action.  After the front end changes their processes to call the new endpoint explicitly, we can remove the proxy add actions from the `InProgressFormsController`.

## Original issue(s)
Second PR for department-of-veterans-affairs/va.gov-team#9856


app/swagger/requests/in_progress_forms.rb
```diff
@@ -55,13 +55,7 @@ class InProgressForms
         operation :get do
           extend Swagger::Responses::AuthenticationError
 
-          key :summary,     'Get form data'
```
erluti, line 58: do you mean for this swagger to no longer have a `summary`?

SarahJaine: yep, since the summary and description are now the same I thought I'd return this to just having a description

PR #4391, opened 2020-06-16
Temporarily ignore CVE-2020-8184 in `rack` while waiting on fix

johnpaulashenfelter:
We are blocked from upgrading `rack` because of a dependency conflict
with Sidekiq5 (which is in the long process of being upgraded in #XXXX)

While we get a fix in for this specific issue in rack through a patch,
we want to avoid breaking the build for the next day or so until a fix
is implemented. This PR temporarily ignores the vulnerability check for
this CVE.

## Original issue(s)


johnpaulashenfelter:
Exception approved here: https://github.com/department-of-veterans-affairs/va.gov-team-sensitive/issues/127

PR #4395, opened 2020-06-16
Debts letter downloader

lihanli:
## Description of change

Add class for listing and downloading a user's PDF debt letters from VBMS

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/9217

tested in rails console on dev


spec/lib/debts/letter_downloader_spec.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require Rails.root.join('modules', 'claims_api', 'spec', 'support', 'fake_vbms.rb')
+
+RSpec.describe Debts::LetterDownloader do
+  let(:file_number) { '796330625' }
+  let(:letter_downloader) { described_class.new(file_number) }
```
erluti, line 8: conventionally i think this would be called `subject`, which would make the tests easier to follow IMO

erluti, line 8: conventionally i think this would be the subject instead of a `let`, which would make the tests easier to follow IMO

PR #4398, opened 2020-06-16
Move lib/facilities models to app/models/facilities

LindseySaari:
## Description of change
In an effort to break the Zeitwerk PR into smaller conceptual based changes, this PR moves lib/facilities models to app/models/facilities. The lib facilities are all inheriting from BaseFacility which is an ActiveRecord model, but are in the lib directory. This PR moves them to app/models/facilities so they are autoloaded with the rest of rails in Zeitwerk mode.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#10195

- [x] Specs pass


erluti:
Can we add the `VHA` and `VC` to this PR?  In the zeitwerk PR the facility files are the only other changes in it.  Adding the inflections don't cause any other changes

PR #4399, opened 2020-06-16
set vaos token cache ttl from token exp value

kreek:
# Description of change
We currently cache the VAOS JWT token for a hard coded 15 minutes. This PR updates the TTL to the value from the expiration in the token itself (also 15 minutes but that could change).

## Original issue(s)


modules/vaos/spec/services/jwt_spec.rb
```diff
@@ -18,7 +18,7 @@
       expect(subject.token).to be_a(String).and match(jwt_regex)
     end
 
-    context 'decoded payload' do
+    context 'with a decoded payload' do
```
erluti, line 21: i think it's clearer when there's a "with", thanks for adding this

modules/vaos/app/services/vaos/user_service.rb
```diff
@@ -7,7 +7,11 @@ def session
       return cached.token if cached
 
       token = get_session_token
-      SessionStore.new(account_uuid: user.account_uuid, token: token).save
+      session_store = SessionStore.new(account_uuid: user.account_uuid, token: token)
+      session_store.save
+      ttl = get_ttl(token)
```
erluti, line 12: I don't think we lose any clarity if we skip this unnecessary assignment and just call `session_store.expire(get_ttl(token))`.

kreek:
@saneshark yes it will be in two parts. Still debugging session refresh with the BE team.

PR #4400, opened 2020-06-17
10197 zeitwerk eagerload tranactional email mailer

LindseySaari:
## Description of change
TransactionalEmailMailer makes available a list of mailers that depend on it. In order for that to function in autoload mode, special code is required to eager load decedents 

** Removed previous HCA inflection changes from this PR after the inflections [revert](https://github.com/department-of-veterans-affairs/vets-api/pull/4456)

## Original issue(s)
department-of-veterans-affairs/va.gov-team#10197

- [x] Specs pass


johnpaulashenfelter:
@LindseySaari it seems like all the HCA mailer and related work together would make a simpler PR, even if its a little bigger. If its going to end up `HCAFoo` everywhere, this seems like regressing to go to `HcaFoo`

app/workers/transactional_email_analytics_job.rb
```diff
@@ -1,5 +1,13 @@
 # frozen_string_literal: true
 
+unless Rails.application.config.eager_load
+  # constantize all the mailers so TransactionalEmailMailer.descendants has something
+  # inspired by https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#single-table-inheritance
+  Dir['app/mailers/*.rb']
+    .collect { |mailer| %r{app/mailers/(.*)\.rb}.match(mailer)[1] }
+    .map { |x| x.camelize.constantize }
+end
```
johnpaulashenfelter, line 9: This really makes my 🕷️ sense tingle. 

Does this come out once the zeitwerk flag is flipped?

erluti: It's sort of the opposite case, the zeitwerk loader needs this because TransactionalEmailAnalyticsJob uses `descendants` which doesn't autoload unnamed files.  So when it's loaded it needs to eager-load it's descendants.

erluti: This is the method that uses the meta descendants
https://github.com/department-of-veterans-affairs/vets-api/blob/56ecc229ba6ec8f51e460dc50bfa12ce6eebcb8e/app/workers/transactional_email_analytics_job.rb#L45

erluti: I guess alternatively we could explicitly list the TransactionalEmailAnalyticsJobs in a constant.  Just in my original zeitwerk PR I was trying to make minimal changes.  As an isolated PR it might make sense to change it up.

johnpaulashenfelter: I realized part of the root of the tingling here is that the mailers aren't really following Rails norms. Instead of inheriting, traditionally you'd have each different email as a method in the container (eg ApplicationMailer, TransactionalMailer) and then all of these mailer controllers start to hew to the convention.

The transactional analytics job i need to dig into more. It seems like there's no real need to dynamically ask "what emails need the analytics" when you can tell it "these jobs need analytics". Solves both problems but now we're getting way closer to rewriting code :/

erluti: I've never written a mailer before, but maybe we could get that rewrite into some of this tech debt stuff?

LindseySaari:
> @LindseySaari it seems like all the HCA mailer and related work together would make a simpler PR, even if its a little bigger. If its going to end up `HCAFoo` everywhere, this seems like regressing to go to `HcaFoo`

I'm on it! Reverting the change and I think the HCA inflection makes sense here. I'm uncovering a lot of interconnections as I breakout the giant Zeitwerk PR. 😄

app/workers/transactional_email_analytics_job.rb
LindseySaari: I'm a ➕ 1 on re-writing as part of tech debt 👍 I'm not familiar with the mailers either and this probably requires some digging and better understanding overall

omgitsbillryan: > we could explicitly list the TransactionalEmailAnalyticsJobs in a constant

I'm in favor of this ☝️ solution because it's simplicity. Perhaps, we could use this Dir/collect/map loop in a spec instead to verify that the new constant we make contains all of the "descendents" we expect.

app/workers/transactional_email_analytics_job.rb
```diff
@@ -36,6 +36,19 @@ def perform
     end
   end
 
+  # constantize all the mailers so TransactionalEmailMailer.descendants has something
+  def self.all_mailers
```
erluti, line 40: I think https://github.com/department-of-veterans-affairs/vets-api/blob/56ecc229ba6ec8f51e460dc50bfa12ce6eebcb8e/app/workers/transactional_email_analytics_job.rb#L45 should use `all_mailers` instead of `descendants`.  When I suggested this approach it was motivated by getting away from using `descendants`.

app/workers/transactional_email_analytics_job.rb
```diff
@@ -36,6 +36,19 @@ def perform
...
+    [
+      FailedClaimsReportMailer,
+      SpoolSubmissionsReportMailer,
+      TransactionalEmailMailer,
+      ApplicationMailer,
+      DirectDepositMailer,
+      HcaSubmissionFailureMailer,
+      YearToDateReportMailer
```
jholton, line 48: I haven't been following closely the changes to `TransactionalEmail` related classes, so apologies if I'm missing something, but:
"Transactional" emails are supposed to be emails that are generated as a result of a "transaction" with an end user.  They are not reports or other mass mailers to stakeholders; they are emails sent to a single end user (e.g. veteran).
Maybe I didn't choose the clearest name; naming things is hard.
The two mailers that are subclasses of `TransactionalEmailMailer` are `DirectDepositMailer` and `HCASubmissionFailureMailer`.
These emails are generated as results of form submissions.
The bodies of the emails include the end user's unique Google Analytics tracking ID.
To complete the picture for GA, this job runs nightly and reports the delivery status of each email to GA.
This job is irrelevant for any mailer class other than `DirectDepositMailer` and `HcaSubmissionFailureMailer`.

I tried to set this up so it'd be easy to add additional "transactional" mailers in the future:
- Create a new mailer
- make it a subclass of `TransactionalEmailMailer`
- define the same constants you see in `DirectDepositMailer` and `HcaSubmissionFailureMailer`
- create your email view (make sure to include the users' GA tracking ID)
 Done.  Now, you have automatic GA tracking and nightly reporting to GA for you end user destined emails.

jholton: Just realized there's a new `TransactionalEmailMailer` subclass: `SchoolCertifyingOfficialsMailer`

PR #4401, opened 2020-06-17
10173 zeitwerk internal external namespacing

LindseySaari:
## Description of change
Zeitwerk mode uncovered many namespacing issues in relation to the vets-api codebase. In particular, many of our Common Exceptions were namespaced within Internal/External namspaces, but were missing the namespacing logic to include ::Internal/::External across the codebase. In order for Zeitwerk mode to autoload correctly, the namespacing updates are necessary.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#10173


erluti:
👍  looks good to me

saneshark:
Just a thought, but why not move "Internal" and "External" into the root folder, and instead use inheritance? There is probably enough difference between internal and external that inheritance might serve a practical purpose, elucidate the same distinctions between error classes, and not require making changes across so many files.

Otherwise VAOS changes LGTM.

erluti:
> Just a thought, but why not move "Internal" and "External" into the root folder, and instead use inheritance? There is probably enough difference between internal and external that inheritance might serve a practical purpose, elucidate the same distinctions between error classes, and not require making changes across so many files.

@saneshark I don't understand what you think they should inherit from?  Do you mean introduce a new `Internal` and `External` exception class for things to inherit from?  I don't think we test for which kind of error it is anywhere, so I thought the decision came down to using Internal/External names spacing or dropping it completely, but I erred on the side of preserving the distinction via a namespace.  I felt like it makes code clearer where they are used.

saneshark:
What I mean by introducing inheritance instead of namespacing is that you could move them all into the root level folder and still preserve some semblance of distinction between the two. There might even be some opportunity to distinguish and abstract out common logic from the two via inheritance. But it's just a though... not necessarily even a good one. 😂

PR #4402, opened 2020-06-17
Update to Rails 6.0.3.2

johnpaulashenfelter:
## Description of change

New Rails gem release to solve CVE-2020-8185, no in bundler-audit as of yet but the patch is available.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#10369

Will verify in CI and staging, but except no surprises.


johnpaulashenfelter:
@erluti nope - this is a completely different CVE than the other two :/

PR #4403, opened 2020-06-17
API 644 Improve MVI Documentation in Vets API

lukemarkwordtlibertyits:
## Description of change
Added description of how MVI differs in Development and Production. To be included in the Veteran Confirmation Swagger declaration. This was requested so customers would better understand how user verification differs in Development and Production environments.

## Original issue(s)
https://vajira.max.gov/browse/API-644


Gemfile.lock
```diff
@@ -124,7 +124,6 @@ PATH
 
 GEM
   remote: https://rubygems.org/
-  remote: https://enterprise.contribsys.com/
```
erluti, line 127: be careful not to check in these gemfile changes

lukemarkwordtlibertyits: I saw that. Definitely my bad. I am working on removing it as we speak.

PR #4407, opened 2020-06-17
add more context to EVSS gender error

annaswims:
## Description of change
In http://sentry.vfs.va.gov/vets-gov/platform-api-production/issues/116861/events/e03b48e06b74481d800f42663229dee7/ we see veterans where user.gender is nil/blank/ or something other than 'M' or 'F' this will help us understand whether the issue is with the identity provider or MVI. 

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/10400

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -84,7 +84,7 @@ class User < Common::RedisStore
   end
 
   def gender
-    identity.gender || (mhv_icn.present? ? mvi&.profile&.gender : nil)
+    identity.gender.presence || (mhv_icn.present? ? mvi&.profile&.gender : nil)
   end
 
   def birth_date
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -84,7 +84,7 @@ def last_name
   end
 
   def gender
-    identity.gender || (mhv_icn.present? ? mvi&.profile&.gender : nil)
+    identity.gender.presence || (mhv_icn.present? ? mvi&.profile&.gender : nil)
```
SarahJaine, line 87: Excellent!

lib/evss/disability_compensation_auth_headers.rb
```diff
@@ -30,15 +30,18 @@ def eauth_json
     end
 
     def gender
-      case @user.gender
```
annaswims, line 33: User.gender can only fall back to MVI when mhv_icn is present because we might need  to use gender to look up a person in MVI, but here we're free to fall back to MVI's gender.
``` ruby
  def gender
    identity.gender || (mhv_icn.present? ? mvi&.profile&.gender : nil)
  end
```
`

PR #4408, opened 2020-06-17
temporarily skip CVE-2020-8185

annaswims:
## Description of change
CVE - https://groups.google.com/forum/#!searchin/rubyonrails-security/CVE-2020-8185%7Csort:date/rubyonrails-security/pAe9EV8gbM0/6JpbOZUSBAAJ

untrusted users can run migrations. I think we're fine to skip in the short term as we wouldn't have migrations in master that aren't ok to run.  

I'm making this PR to unblock myself but I don't intend to do the follow-up work.

### DIFF
--- a/rakelib/security.rake
+++ b/rakelib/security.rake
@@ -14,7 +14,7 @@ task security: :environment do
 
   puts 'running bundle-audit to check for insecure dependencies...'
   exit!(1) unless ShellCommand.run('bundle-audit update')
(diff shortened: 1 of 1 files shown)


erluti:
We've already updated to 6.0.3.2, so this shouldn't be necessary
https://github.com/department-of-veterans-affairs/vets-api/pull/4402

PR #4411, opened 2020-06-18
10178 zeitwerk mvi rename

LindseySaari:
## Description of change
MVI is a Rails model as well as a namespace in our lib directory that utilities the Master Veteran Index external Service. In the [Zeitwerk PR](https://github.com/department-of-veterans-affairs/vets-api/pull/4379/files), we removed the autoloading of the lib directory. In Zeitwerk mode, if the lib MVI name-spaced classes are referenced before the MVI model, the model isn't defined as inheriting from Common::RedisStore.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#10178


annaswims:
If you're going to go to the trouble of renaming MVI, do you care that it's not called MVI anymore?🤷‍♀️
It's MPI/MasterPersonIndex now I think

LindseySaari:
> If you're going to go to the trouble of renaming MVI, do you care that it's not called MVI anymore?🤷‍♀️
> It's MPI/MasterPersonIndex now I think

@annaswims Yes, this crossed my mind as well! We may have to circle back on this because we're trying to get Zeitwerk over the line as it's a current blocker for our sidekiq, etc upgrades.

spec/lib/master_veteran_index/configuration_spec.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
```
omgitsbillryan, line 1: I'm assuming this file was just a move / rename?

omgitsbillryan:
> In Zeitwerk mode, if the lib MVI name-spaced classes are referenced before the MVI model, the model isn't defined as inheriting from Common::RedisStore.

I don't quite understand the "the model isn't defined as inheriting from Common::RedisStore" part. What does the mvi model get defined as then? Maybe before & after using the rails console might help me understand.

erluti:
> > In Zeitwerk mode, if the lib MVI name-spaced classes are referenced before the MVI model, the model isn't defined as inheriting from Common::RedisStore.
> 
> I don't quite understand the "the model isn't defined as inheriting from Common::RedisStore" part. What does the mvi model get defined as then? Maybe before & after using the rails console might help me understand.

Loading `class MVI < Common::RedisStore` establishes the class the way we intend to use it.  To put something in it's names space, you would do this:
```
class MVI
  class NewNestedClass
  end
end
```

But if `NewNestedClass` is loaded first, the actual `MVI` class will fail to load because you can't retroactively have it inherit from `Common::RedisStore`.

app/controllers/v0/mvi_users_controller.rb
```diff
@@ -10,15 +10,15 @@ def submit
...
+          source: 'MVIUsersController'
         )
       end
 
       # Scenario indicates serious problems with user data
       if @current_user.birls_id.nil? && @current_user.participant_id.present?
         raise Common::Exceptions::UnprocessableEntity.new(
           detail: 'No birls_id while participant_id present',
-          source: 'MviUsersController'
+          source: 'MVIUsersController'
```
saneshark, line 21: NITPICK, this is mostly just a heads up, but this is totally going to cause any instances of these errors to be grouped differently in Sentry I would think. If we plan on evaluating MVI vs MPI it might not hurt to make that evaluation as part of these changes -- if nothing else to save some grief in having errors grouped in Sentry differently yet again later.

That being said, I don't know with what frequency that type of thing happens.

app/controllers/v0/profile/personal_informations_controller.rb
```diff
@@ -21,7 +21,7 @@ class PersonalInformationsController < ApplicationController
       #   }
       #
       def show
-        response = Mvi.for_user(@current_user).profile
+        response = MVI.for_user(@current_user).profile
```
saneshark, line 24: I don't know how I feel about this. This makes MVI look like a constant or Module instead of a class. I think traditionally even class names that are an abbreviation are done as Titlecase.

app/models/mvi.rb
```diff
@@ -1,13 +1,16 @@
...
+require 'master_veteran_index/orch_search_service'
+require 'master_veteran_index/responses/find_profile_response'
+require 'master_veteran_index/responses/add_person_response'
+require 'master_veteran_index/service'
 
 # Facade for MVI. User model delegates MVI correlation id and VA profile (golden record) methods to this class.
 # When a profile is requested from one of the delegates it is returned from either a cached response in Redis
 # or from the MVI SOAP service.
-class Mvi < Common::RedisStore
+class MVI < Common::RedisStore
```
saneshark, line 13: same comment as before -- this makes the class appear to be a constant, is it really necessary?

app/controllers/v0/mvi_users_controller.rb
erluti: I think this is was an oversight, I think you're right we should leave this alone

app/controllers/v0/profile/personal_informations_controller.rb
kreek: There is a `JSON` class 😄 maybe it could be MVIRecord? (alternately in the MVI docs they use 'Person' e.g. the soap methods are called FindPerson, UpdatePerson...)

erluti:
We're looking into the naming decisions here.  Combined with the fact that MVI is now MPI (Master Persons Index), we're planning on closing this and creating a new PR using the new name.

PR #4416, opened 2020-06-18
User specified SAML authn request

ericbuckley:
To support inbound ssoe integration, we will need the Authn context to match that of what the user had previously signed in with.  We are working with the IAM team on updating the `/keepalive` endpoint to return this data, if the user has an active SSOe session.  Thus from the FE we can pass this custom "authn" context to the API endpoint and use it to construct the SAML Request.

refs https://github.com/department-of-veterans-affairs/va.gov-team/issues/10411


app/controllers/v1/sessions_controller.rb
```diff
@@ -74,28 +73,20 @@ def metadata
...
-        url_service.verify_url
+      when 'custom'
+        raise Common::Exceptions::ParameterMissing, 'authn' if params[:authn].blank?
+
+        url_service.custom_url params[:authn]
       when 'slo'
         url_service.ssoe_slo_url # due to shared url service implementation
       else
-        raise Common::Exceptions::RoutingError, params[:path]
+        url_service.public_send("#{type}_url")
```
patrickvinograd, line 87: This will _probably_ throw a static analyzer error in CI because you're invoking public_send based on user input (even though user input is constrained by that routing error check above)

ericbuckley: I refactored this method because CI told me the complexity, or maybe line length, was too much with the addition of the "custom" case statement.  Right now CI is passing, but can you think of another way of doing this?

spec/controllers/v1/sessions_controller_spec.rb
```diff
@@ -237,6 +237,35 @@ def expect_logger_msg(level, msg)
...
+              .to trigger_statsd_increment(described_class::STATSD_SSO_NEW_KEY,
+                                           tags: ['context:custom', 'version:v1'], **once)
+            expect(response).to have_http_status(:found)
+            expect(response.location)
+              .to be_an_idme_saml_url('https://pint.eauth.va.gov/isam/sps/saml20idp/saml20/login?SAMLRequest=')
+              .with_relay_state('originating_request_id' => nil, 'type' => 'custom')
+              .with_params('clientId' => '123123')
+          end
+
+          it 'raises exception on missing parameter' do
```
erluti, line 252: i think it would be nice if this test's name was explicitly saying the missing parameter is `authn`.
```
it 'raises exception when missing authn paramater' do
```

app/controllers/v1/sessions_controller.rb
erluti: I think the best thing to do would have a heavy-duty refactoring of UrlService, since this method is already wrapping calls to that object anyway.  But that's out-of-scope for what's going on here and for what it's worth, I think it's clearer to have the rubocop exception and a giant case-statement than to use metaprogramming.

spec/controllers/v1/sessions_controller_spec.rb
ericbuckley: will change

app/controllers/v1/sessions_controller.rb
```diff
@@ -76,6 +75,8 @@ def metadata
 
     # rubocop:disable Metrics/CyclomaticComplexity
     def redirect_url(type)
+      raise Common::Exceptions::RoutingError, type unless REDIRECT_URLS.include?(type)
```
erluti, line 78: I like that you moved this to the top of the method

PR #4418, opened 2020-06-19
10165 - AAID BB BGS and CARMA inflections

LindseySaari:
## Description of change
Defines AAID, BB, BGS and CARMA inflections for VA acronyms across the codebase. These inflections are needed for the autoloading of filenames and namespaces for Zeitwerk mode.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#10165


lib/carma/models/metadata.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+require_relative 'base'
+require_relative 'caregiver'
+require_relative 'veteran'
+
+module CARMA
+  module Models
+    class Metadata < Base
```
erluti, line 9: i think this is so much clearer without being nested in classes

config/initializers/inflections.rb
```diff
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
 ActiveSupport::Inflector.inflections do |inflect|
+  inflect.acronym 'AAID'
```
kfrz, line 4: J/C, What's AAID? Not in [`acronyms`](https://github.com/department-of-veterans-affairs/acronyms)

PR #4419, opened 2020-06-19
10165 use inflection for S-acronyms

erluti:
## Description of change
Introducing an inflection for 'SAML', 'SFTP', 'SM', and 'SSO' as acronyms. These acronym will support zeitwerk filenames and module namespaces.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#1065

## Additional Notes
`SFTPWriter` needs to be defined as a module space for zeitwerk, so making that change along with the acronym inflection.


config/initializers/inflections.rb
```diff
@@ -13,7 +13,11 @@
   inflect.acronym 'MDOT'
   inflect.acronym 'PagerDuty'
   inflect.acronym 'PPIU'
+  inflect.acronym 'SAML'
+  inflect.acronym 'SFTP'
+  inflect.acronym 'SM'
   inflect.acronym 'SSOe'
+  inflect.acronym 'SSO'
```
omgitsbillryan, line 20: I don't see any code updated as a result of this new inflection - what gives??

erluti: Zeitwerk tries to link filenames to class names, so `sso_whatever` should have `SsoWhatever` as the classname unless you have the inflection.
However I can't find a real example of that here, and I tried my old zeitwerk PR without this inflection and it doesn't seem to cause issues.  So I'm going to take it out.

PR #4420, opened 2020-06-19
add JWT, NCA and MHV inflections

LindseySaari:
Description of change
Defines AAID, BB, BGS and CARMA inflections for VA acronyms across the codebase. These inflections are needed for the autoloading of filenames and namespaces for Zeitwerk mode.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#10165

- [x] Specs pass
- [x] Rakefile rake tasks function properly


LindseySaari:
@erluti I may need to break out the inflection migration changes to a separate PR.

erluti:
> @erluti I may need to break out the inflection migration changes to a separate PR.

In my original zeitwerk PR attempt, I didn't change any of the migration files.  I was thinking that migrations are already run in production and capitalization won't really matter when people re-run them.

PR #4424, opened 2020-06-22
don't ignore CVE-2020-8184 in bundle-audit; fix is merged

f1337:
Now that https://github.com/department-of-veterans-affairs/vets-api/pull/4394 has been merged, `bundle-audit` does not need to ignore CVE-2020-8184.


erluti:
Looks like the build is still failing because it thinks the only solution is to upgrade rack

f1337:
huh, well then we'll leave this un-merged until we upgrade rack.

johnpaulashenfelter:
Closed in favor of #4499

PR #4426, opened 2020-06-22
Integration with VANotify, to send Form526 confirmation email

philherbert:
### Changes in behaviour

On successful submission of Form 526, we make a request to VANotify containing the veteran's email address and the ID of the template used for Form 526 confirmation emails. On VANotify, this will then queue a job to send an email (currently using GovDelivery).

### Dependencies introduced

We added a new gem to the `Gemfile` - `notifications-ruby-client`. This is used to make requests to VANotify.

## Original issue(s)
department-of-veterans-affairs/notification-api#45

  * Yes, we've added a `vanotify` block to `settings.yml`, containing (fake values for) the Notification API url, API Key, and ID for the confirmation email template.
  * These do indeed vary by environment - we've added configuration for the development and review environments in this PR: department-of-veterans-affairs/devops#7145
  * The API key is in credstash, under `vets_api.<env>.vanotify_api_key`
  * Yes, it is `form526_confirmation_email`
  * No Sentry logging, but more logging will be coming in future stories: department-of-veterans-affairs/notification-api#86 (for error handling), and department-of-veterans-affairs/notification-api#78 & department-of-veterans-affairs/notification-api#74 (general monitoring)
  * We do use a general `Rails.logger.error` in `Form526ConfirmationEmailJob`
  * Not yet, see the above links to monitoring stories
  * No, we didn't change any APIs, and the API integration with VANotify is done using a Ruby client
  * We do obtain the veteran's email address in `form526_submission.rb`, with `form['form526']['form526']['veteran']['emailAddress']`. This is not explicitly logged, but we'll double check once it's deployed.

## Testing
We've added unit tests to cover all of our code changes, but we're primarily going to test on the Review instance, so we can verify that the configuration allows successful integration.

**UPDATE**: we have tested it on the review instance and it works as expected


philherbert:
Screen Shot 2020-06-24 at 14 12 52

.gitignore
```diff
@@ -63,3 +63,5 @@ dump.rdb
 /.project
 .generators
 .rakeTasks
+
+.talismanrc
```
kfrz, line 67: I'm not certain we're using Talisman in our git hooks for this proejct, can this be added to your local `~/.gitignore.global` ?

config/settings.yml
```diff
@@ -480,20 +480,23 @@ flipper:
     - kmircovich@governmentcio.com
     - lauren.alexanderson@va.gov
     - lindsey.hattamer@oddball.io
+    - ltran@thoughtworks.com
```
kfrz, line 483: NIT: I'd prefer to see Flipper admins set up in a separate PR, but I don't think there's a hard and fast rule so this is fine.

spec/jobs/form526_confirmation_email_job_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+require 'rails_helper'
+require 'notifications/client'
+
+RSpec.describe Form526ConfirmationEmailJob, type: :worker do
+  before { Sidekiq::Worker.clear_all }
+
+  describe '#perform' do
+    let(:notification_client) { double('Notifications::Client') }
+
+    context 'defaults' do
```
kfrz, line 12: Context blocks should use "with $CONTEXT" or "when $CONTEXT" so the spec block reads more naturally:

'with defaults' it 'sends a confirmation email'
or -

'when user is unauthorized' it 'does not send an email' etc.

kfrz, line 12:
```suggestion
    context 'with default attributes' do
```

philherbert: Good to know! Thanks for the suggestion, we've approved it

config/settings.yml
philherbert: Fair enough, we'll be aware of that for future changes! Thanks.

.gitignore
philherbert: Good spot, we've just removed it! Thanks.

PR #4428, opened 2020-06-22
Insert 2122 signatures

charleystran:
## Description of change
This PR adds the base64 encoded signature images to the a 21-22 form, and fills out the form with structured data.


modules/claims_api/app/models/claims_api/power_of_attorney.rb
```diff
@@ -16,6 +17,70 @@ class PowerOfAttorney < ApplicationRecord
...
+    end
+
+    def convert_base64_data_to_image(signature)
+      path = "tmp/#{signature}_#{id}_signature_b64.png"
+
+      File.open(path, 'wb') do |f|
+        f.write(Base64.decode64(form_data.dig('signatures', signature)))
+      end
+      path
+    end
```
edmangimelli, line 41: I think I've already mentioned this, but I'm worried that two requests using the same id could try to be writing to the same file. We don't need `#{SecureRandom.hex}` in this filename?

modules/claims_api/app/models/claims_api/power_of_attorney.rb
```diff
@@ -16,6 +17,70 @@ class PowerOfAttorney < ApplicationRecord
...
+      pdf_path = Rails.root.join('modules', 'claims_api', 'config', 'pdf_templates', "21-22A-#{page}.pdf")
+      stamp_path = "#{::Common::FileHelpers.random_file_path}.pdf"
+
+      inserted = []
+      Prawn::Document.generate(stamp_path, margin: [0, 0]) do |pdf|
+        y_representative_coords = page == 1 ? 118 : 216
+        y_veteran_coords = page == 1 ? 90 : 322
+        pdf.image representative_signature, at: [35, y_representative_coords], height: 20
+        inserted << 'representative'
+        pdf.image veteran_signature, at: [35, y_veteran_coords], height: 20
```
edmangimelli, line 53: should this be followed by:
```
        inserted << 'veteran'
```
?

modules/claims_api/app/models/claims_api/power_of_attorney.rb
```diff
@@ -16,6 +17,70 @@ class PowerOfAttorney < ApplicationRecord
...
+    def convert_base64_data_to_image(signature)
+      path = "tmp/#{signature}_#{id}_signature_b64.png"
+
+      File.open(path, 'wb') do |f|
+        f.write(Base64.decode64(form_data.dig('signatures', signature)))
+      end
+      path
+    end
+
+    def insert_signatures(page, veteran_signature, representative_signature)
```
edmangimelli, line 43: non-blocking:  seems like a good place to use kwargs to me.

modules/claims_api/app/models/claims_api/power_of_attorney.rb
```diff
@@ -53,6 +118,18 @@ def external_uid
       source_data.present? ? source_data['icn'] : Settings.bgs.external_uid
     end
 
+    def signature_image_paths
+      @signature_image_paths ||= {}
+    end
+
+    def create_signature_image(signature_type)
+      path = "/tmp/#{signature_type}_#{id}_signature.png"
```
edmangimelli, line 126: same concern https://github.com/department-of-veterans-affairs/vets-api/pull/4428/files#r444194792

modules/claims_api/app/models/claims_api/power_of_attorney.rb
charleystran: No this wont be an issue. Actually based on the current way ITF works, if the headers are not changed, this process would never run a second time for a veteran. When we do change that, this method gives us a reliable name without having to persist another field in the database. In your case above, we would want a new file to be written and used.

modules/claims_api/app/models/claims_api/power_of_attorney.rb
charleystran: I might clean this up. The purpose is really for adding to the errors of which signature is bad since the Prawn error doesnt include any detail. so if the inserted array has rep in it and an error is thrown, that means the veteran file is the bad one.

modules/claims_api/lib/claims_api/stamp_signature_error.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module ClaimsApi
+  class StampSignatureError < StandardError
+    attr_accessor :code
+    attr_accessor :detail
+
+    def initialize(message: nil, detail: nil)
+      message = message if message.present?
```
edmangimelli, line 9: say what?

modules/claims_api/lib/claims_api/stamp_signature_error.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module ClaimsApi
+  class StampSignatureError < StandardError
+    attr_accessor :detail
+
+    def initialize(message: nil, detail: nil)
+      super(message)
```
edmangimelli, line 8: 👍

PR #4429, opened 2020-06-22
VAOS refresh session token

kreek:
## Description of change
VAOS sessions are extended via an endpoint. This PR updates the base service to run a background job when another response is returned. To avoid unnecessary calls session refresh is locked to a minute (the longest a VAOS response could take).

## Original issue(s)

Logs when a new session is created and when it's refreshed.


saneshark:
Interesting approach with the 2 minute window. Why not an asynchronous API call in a worker that runs more regularly?

kreek:
> Interesting approach with the 2 minute window. Why not an asynchronous API call in a worker that runs more regularly?

I like that, was trying to keep calls down as this will literally double our calls to the BE but it would ensure the session is always pushed out with each request.

modules/vaos/app/services/vaos/session_service.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+# frozen_string_literal
```
jholton, line 3: Typo/redundant?

modules/vaos/app/services/vaos/user_service.rb
```diff
@@ -2,37 +2,103 @@
...
+
+    def update_session_token(account_uuid)
+      cached = cached_by_account_uuid(account_uuid)
+      if cached
+        url = '/users/v2/session/jwts'
+        response = perform(:get, url, nil, refresh_headers(account_uuid))
+        new_token = response.body[:jws]
+        Rails.logger.info('VAOS session updated',
+                          {
+                            account_uuid: account_uuid,
```
saneshark, line 27: NITPICK:

I was going to suggest moving this into save_session!  but I think you're using the same method for refresh and first time creation, so i guess this works. Trying to think if maybe save_session could save_and_log_session but accepts parameters to inform which type of logging to do. 

in other words:

```ruby
def save_session!(action = 'update')
   ... do stuff
   if success 
      "VAOS session #{action} successful"
   else
      "VAOS session #{action} failed"
  end
end

PR #4430, opened 2020-06-23
Facility locator forcing legacy urgent care to pos

darkbushido:
when you select a provider and enter the code for urgent care we are now
forcing it to use the POS locator

## Description of change
People are able to search for urgent care using two different paths.
 * The legacy path is to select the provider and under specialty to select `urgent care`
   * This calls the PPMS provider endpoint with the specialty code of `261QU0200X`
 * The new path is to select `urgent care` and community care providers
    * This calls the PPMS Place of Service endpoint
This checks if the specialty code is `261QU0200X` and makes the call to the Place of service endpoint
I had to switch from the case statement to the if statement format because this is getting a bit ugly and was failing the complexity linter.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#5128

* Feature flag `:facility_locator_ppms_legacy_urgent_care_to_pos_locator`

unit tests to verify that we are making the POS call instead of the provider call


app/controllers/v0/facilities/ccp_controller.rb
```diff
@@ -36,6 +29,37 @@ def services
...
+      provider_search
+    when 'cc_pharmacy'
+      provider_search('services' => ['3336C0003X'])
+    when 'cc_urgent_care'
+      api.pos_locator(search_params)
+    end
+  end
+
+  def ppms_search_with_legacy_urgent_care
+    if search_params[:type] == 'cc_provider' && search_params['services'] == ['261QU0200X']
```
erluti, line 52: minor optimization: you could put the `search_params['services'] == ['261QU0200X']` predicate first so you don't have to evaluate `search_params[:type] == 'cc_provider'` twice when it's true

app/controllers/v0/facilities/ccp_controller.rb
```diff
@@ -36,6 +29,37 @@ def services
 
   private
 
+  def ppms_search
+    if Flipper.enabled?(:facility_locator_ppms_legacy_urgent_care_to_pos_locator)
```
erluti, line 33: does there need to be a test for when this flipper is disabled that shows the `261QU0200X` result is the same as `cc_provider`?

app/controllers/v0/facilities/ccp_controller.rb
darkbushido: I can't, `type` has precedence. so if they pass cc_pharmacy I don't want them to get a POS search

app/controllers/v0/facilities/ccp_controller.rb
darkbushido: I removed the test that showed how it worked without flipper. if you need it I can put it back.

app/controllers/v0/facilities/ccp_controller.rb
erluti: sorry, I meant reversing them in this line like:
```
search_params['services'] == ['261QU0200X'] && search_params[:type] == 'cc_provider' 
```
so when `search_params['services'] == ['261QU0200X']` is false it won't immediately check `search_params[:type] == 'cc_provider'` again on the next line

app/controllers/v0/facilities/ccp_controller.rb
erluti: As long as it's a valid code path I think it's worth keeping

app/controllers/v0/facilities/ccp_controller.rb
darkbushido: I think the performance hit on this is minor. we already have the params hash. I think your proposed change makes it more difficult to read. I think it was breaking the complexity checker but originally it was a case statement with an if  inside the `cc_provider` case.

PR #4431, opened 2020-06-23
Pv saml post binding

patrickvinograd:
## Description of change
Updates v1 session controller to use SAML POST Binding. This is done at request of IAM SSOe team who believes doing so is more in line with intent for their stack and will resolve some outstanding bugs. 

The hard-working "URLService" was unfortunately forked since now it needs to generate SAML POST request form parameters instead of a straight up URL. I think this can be brought back together by the consuming classes being aware of the need to render a form vs. a redirect depending on the scenario. 

Testing in local and dev shows no regression relative to using the SAML Redirect binding. Jury is still out on whether it resolves the suspected list of issues.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#10468

### DIFF
--- a/app/controllers/v1/sessions_controller.rb
+++ b/app/controllers/v1/sessions_controller.rb
@@ -9,7 +9,6 @@ require 'saml/responses/logout'
 module V1
   class SessionsController < ApplicationController
     skip_before_action :verify_authenticity_token
-    # before_action :enable_post_view
 
     REDIRECT_URLS = %w[signup mhv dslogon idme custom mfa verify slo].freeze
 
--- a/app/controllers/v1/sessions_controller.rb
+++ b/app/controllers/v1/sessions_controller.rb
@@ -127,7 +127,7 @@ module V1
     end
 
     def render_login(type, previous_saml_uuid = nil)
-      post_params, login_url = login_params(type, previous_saml_uuid)
+      login_url, post_params = login_params(type, previous_saml_uuid)
       renderer = ActionController::Base.renderer
       renderer.controller.prepend_view_path(Rails.root.join('lib', 'saml', 'templates'))
       result = renderer.render template: 'sso_post_form',
--- a/app/controllers/v1/sessions_controller.rb
+++ b/app/controllers/v1/sessions_controller.rb
@@ -8,11 +8,9 @@ require 'saml/responses/logout'
 
 module V1
   class SessionsController < ApplicationController
-    include ActionController::Rendering
-    include ActionView::Layouts
 
     skip_before_action :verify_authenticity_token
-    before_action :enable_post_view
+    # before_action :enable_post_view
 
     REDIRECT_URLS = %w[signup mhv dslogon idme custom mfa verify slo].freeze
 
@@ -130,9 +128,15 @@ module V1
       end
     end
 
+    # TODO: Replace ActionView::Base with ActionController::Base.renderer
+    # once we figure out how to override the view_paths for that renderer
     def render_login(type, previous_saml_uuid = nil)
       post_params, login_url = login_params(type, previous_saml_uuid)
-      render template: 'sso_post_form', locals: { url: login_url, params: post_params }
+      renderer = ActionView::Base.with_view_paths([Rails.root.join('lib','saml','templates')])
+      result = renderer.render template: 'sso_post_form', 
+        locals: { url: login_url, params: post_params },
+        format: :html
+      render body: result, content_type: "text/html"
     end
 
     # rubocop:disable Metrics/CyclomaticComplexity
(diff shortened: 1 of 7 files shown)


app/controllers/v1/sessions_controller.rb
```diff
@@ -138,15 +121,47 @@ def user_login(saml_response)
...
-        # only record login stats if the user is being redirect to the site
-        # some users will need to be up-leveled and this will be redirected
-        # back to the identity provider
-        login_stats(:success, saml_response, user_session_form)
+      helper = url_service(user_session_form.saml_uuid)
+      if helper.should_uplevel?
+        render_login('verify')
+      else
+        redirect_to helper.login_redirect_url
+        if location.start_with?(helper.base_redirect_url)
```
ericbuckley, line 129: I think this if block and the one on line 125 are testing the same thing.  Now that there is an else block I think this can be removed

app/controllers/v1/sessions_controller.rb
```diff
@@ -237,12 +252,16 @@ def originating_request_id
     end
 
     def url_service(previous_saml_uuid = nil)
-      SAML::URLService.new(saml_settings,
-                           session: @session_object,
-                           user: current_user,
-                           params: params,
-                           loa3_context: LOA::IDME_LOA3,
-                           previous_saml_uuid: previous_saml_uuid)
+      SAML::PostURLService.new(saml_settings,
```
ericbuckley, line 255: why not reuse SAML::URLService?  Whats different about PostURLService?

app/controllers/v1/sessions_controller.rb
ericbuckley: actually I just realized there is a bug with this if statement that makes the if on line 125 preferred.  When we redirect users to cerner, this block won't get executed.

app/controllers/v1/sessions_controller.rb
patrickvinograd: Instead of returning a URL, it's returning the SSOe URL and a list of form parameters (but not jammed together into a query string),that are then used in this controller to render the auto-submitting form HTML template. 

But it's messy because _some_ of the interactions (such as SLO) are still a simple redirect vs. a render, so there's not really a clean interface to URLService/PostURLService.

app/controllers/v1/sessions_controller.rb
patrickvinograd: Ooh, good catch.

app/controllers/v1/sessions_controller.rb
patrickvinograd: There's so much overlapping logic I'd like to bring those classes back together...thinking. about how to do that.

lib/saml/post_url_service.rb
```diff
@@ -0,0 +1,89 @@
+# frozen_string_literal: true
+
+module SAML
+  # This class is responsible for providing the requests for the various
+  # SSO and SLO endpoints
+  # It provides a similar interface to URLService, but for most endpoints
+  # it returns an SSO URL and form request parameters for use in a SAML POST
+  # submission, instead of a self-contained redirect URL.
+  class PostURLService < URLService
+   # rubocop:disable Metrics/ParameterLists
```
kfrz, line 10:
```suggestion
    # rubocop:disable Metrics/ParameterLists
```

Rubocop indent warn

lib/saml/post_url_service.rb
```diff
@@ -0,0 +1,89 @@
+# frozen_string_literal: true
+
+module SAML
+  # This class is responsible for providing the requests for the various
+  # SSO and SLO endpoints
+  # It provides a similar interface to URLService, but for most endpoints
+  # it returns an SSO URL and form request parameters for use in a SAML POST
+  # submission, instead of a self-contained redirect URL.
```
kfrz, line 8:
```suggestion
  # This class is responsible for providing the requests for the various SSO and SLO endpoints.
  # It provides a similar interface to {SAML::URLService}, but for most endpoints it returns an SSO URL and 
  # form request parameters for use in a SAML POST submission, instead of a self-contained redirect URL.
  #
  # @see SAML::URLService
  #
```

Suggestion for adding links to class references for YARD purposes:

image

lib/saml/post_url_service.rb
```diff
@@ -0,0 +1,89 @@
...
+    def login_redirect_url(auth: 'success', code: nil)
+      if auth == 'success'
+        # if the original auth request specified a redirect, use that
+        redirect_target = @tracker&.payload_attr(:redirect)
+        return redirect_target if redirect_target.present?
+      end
+
+      # if the original auth request specified inbound ssoe and authentication
+      # failed, set 'force-needed' so the FE can silently fail authentication and NOT
+      # show the user an error page
```
kfrz, line 51: Could this be part of the method signature w/o major lift to refactor?

Otherwise, maybe this doc-comment can go above the method declaration, so it's visible in the generated YARD doc for the method.

PR #4435, opened 2020-06-23
Add bddQualified field to 526

SarahJaine:
## Description of change

Add required bddQualified boolean field to form 526 application so service member can apply for disability benefits prior to separation from the military.

Logic for field: 
- Release from Active Duty (RAD) is calculated by user submission not EMIS data
- If form is submitted >180 days prior to RAD date: BE will return failure because FE should gate this. It's better for the user to simply wait to submit.
- If form is submitted < 90 days prior to RAD date: bddQualified = false.  Form is still submitted EVSS will accept the form and per Barry the form will be processed as "Quick Start," a version of 526.
- If form is submitted 180-90 days prior to RAD date: bddQualified = true.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#10031


lib/evss/disability_compensation_form/data_translation_all_claim.rb
```diff
@@ -604,6 +641,36 @@ def itf
...
+      def translate_bdd
+        user_supplied_rad_date
+        bdd_qualified?
+      end
+
+      def user_supplied_rad_date
+        # Retrieve the most recent Release from Active Duty (RAD) date from user supplied service periods
+
+        recent_service_period = translate_service_periods.sort_by { |episode| episode['activeDutyEndDate'] }.reverse[0]
+        @user_supplied_rad_date = recent_service_period['activeDutyEndDate'].in_time_zone(EVSS_TZ).to_date
```
erluti, line 658: it doesn't look like instance variable is being accessed anywhere else, so instead of setting it this method could just return the value and be used in the `bdd_qualified?` method.  And if that's the case, `translate_bdd` is the same as `bdd_qualified?` and one of those method names could be dropped.  I tried it out locally and all the tests still passed.  (I went with `bdd_qualified?` because it's used as a boolean and I figured that made sense.)
```
      def user_supplied_rad_date
        # Retrieve the most recent Release from Active Duty (RAD) date from user supplied service periods

        recent_service_period = translate_service_periods.sort_by { |episode| episode['activeDutyEndDate'] }.last
        recent_service_period['activeDutyEndDate'].in_time_zone(EVSS_TZ).to_date
      end

      def bdd_qualified?
        # To be bdd_qualified application should be submitted 180-90 days prior to Release from Active Duty (RAD) date.
        # Applications < 90 days prior to release can be submitted but only with value as false.
        days_until_release = user_supplied_rad_date - @form_submission_date

        if days_until_release > 180
          raise Common::Exceptions::UnprocessableEntity.new(
            detail: 'User may not submit BDD more than 180 days prior to RAD date',
            source: 'DataTranslationAllClaim'
          )
        end
        days_until_release >= 90
      end
```

SarahJaine: Yeah, I think you're right.  I started our with something significantly more complex, but as I paired it down I didn't refactor.  This is much nicer.  Thank you!

PR #4437, opened 2020-06-23
VBMS debt letters controller

lihanli:
## Description of change
Add controller and swagger documentation for getting a user's PDF debt letters

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/9215

swagger documentation updated


app/controllers/v0/debt_letters_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+
+module V0
+  class DebtLettersController < ApplicationController
+    def index
+      render(json: service.list_letters)
+    end
+
+    def show
+      send_data(
+        service.get_letter(params[:id]),
```
lihanli, line 11: @camerontesterman do you think that we should verify that a debt letter is in the user's folder (by calling `list_letters` again), before allowing them to download the letter?

lib/debts/letter_downloader.rb
```diff
@@ -38,5 +40,13 @@ def list_letters
         )
       end
     end
+
+    private
+
+    def verify_letter_in_folder(document_id)
+      raise Common::Exceptions::Unauthorized unless list_letters.map do |letter|
```
erluti, line 47: using `map` and `include?` is iterating over this list more than once.  `any` accepts a block and still returns a bool and IMO is a little clearer
```
raise Common::Exceptions::Unauthorized unless list_letters.any? { |letter| letter[:document_id] == document_id }
```

spec/lib/debts/letter_downloader_spec.rb
```diff
@@ -24,24 +24,42 @@ def get_vbms_fixture(path)
...
-
     before do
-      expect(vbms_client).to receive(:send_request).with(
-        request_double
-      ).and_return(
-        OpenStruct.new(
-          document_id: document_id,
-          content: content
-        )
+      # rubocop:disable RSpec/SubjectStub
```
erluti, line 28: did you try making `request_double` into a `list_request` and `letter_request` for each time of request you'd want to stub?  That way you don't have to stub the object under test.

spec/lib/debts/letter_downloader_spec.rb
```diff
@@ -8,69 +8,62 @@
 
   let(:file_number) { '796330625' }
   let(:vbms_client) { FakeVbms.new }
-  let(:request_double) do
+
+  def stub_request(request_name, args, return_val)
```
erluti, line 12: I like this method.  Small suggestion, name it something like `stub_vbms_client_request`

PR #4443, opened 2020-06-24
update decision review

annaswims:
## Description of change
since we wrote the code to make a call to Lighthouse (https://sandbox-api.va.gov/services/appeals/v1/decision_reviews/contestable_issues) some headers and attribute names have changed. This updates the schema and the specs.
The 200-response spec now includes a real recorded response.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#9817

cc: @Mottie

### DIFF
--- a/app/swagger/swagger/schemas/appeals/contestable_issues.rb
+++ b/app/swagger/swagger/schemas/appeals/contestable_issues.rb
@@ -11,7 +11,7 @@ module Swagger
             items type: :object do
               key :description, 'A contestable issue (to contest this, you include it as a RequestIssue'\
                                 ' when creating a HigherLevelReview, SupplementalClaim, or Appeal)'
-              property :type, type: :string, enum: %w[ContestableIssue]
+              property :type, type: :string, enum: %w[contestableIssue]
               property :attributes, type: :object do
                 property :ratingIssueId do
                   key :type, :string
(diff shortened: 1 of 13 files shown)


lib/decision_review/service.rb
```diff
@@ -75,7 +75,7 @@ def get_contestable_issues(user)
 
     def request_headers(user)
       {
-        'veteranId' => user.ssn,
+        'x-va-ssn' => user.ssn,
         'receiptDate' => Time.current.strftime('%Y-%m-%d')
```
edmangimelli, line 79:
```suggestion
        'x-va-receipt-date' => Time.current.strftime('%Y-%m-%d')
```

edmangimelli: https://github.com/department-of-veterans-affairs/vets-api/blob/623e4284fdb573fc11bd3de18abfd563928748fa/modules/appeals_api/app/swagger/appeals_api/v1/contestable_issues_controller_swagger.rb#L19

edmangimelli: you could also use `Time.current.strftime('%F')`

edmangimelli:
🍾  looks good!

app/swagger/swagger/schemas/appeals/contestable_issues.rb
```diff
@@ -11,7 +11,7 @@ class ContestableIssues
             items type: :object do
               key :description, 'A contestable issue (to contest this, you include it as a RequestIssue'\
                                 ' when creating a HigherLevelReview, SupplementalClaim, or Appeal)'
-              property :type, type: :string, enum: %w[ContestableIssue]
```
annaswims, line 14: we probably want to make corresponding changes to https://github.com/department-of-veterans-affairs/vets-json-schema/blob/master/src/schemas/20-0996/schema.js

annaswims:
[confirmed ](http://grafana.vfs.va.gov/d/zGS4XzkMk/backend-to-frontend-app-mapping?orgId=1&var-data_source=Prometheus%20(Production)&var-backend_controller=v0%2Fcontestable_issues&var-frontend_app=All&from=now-7d&to=now)that this endpoint is not yet used.

annaswims:
I legitimately redid the VCR cassette for 200_contestable_issues.yml and edited the others.  It's worth noting that the other VCR cassettes look like they might not be real recorded requests.

PR #4450, opened 2020-06-25
Add appeals metadata controller

charleystran:
## Description of change
This PR adds the metadata controller to the appeals api. Currently the controller just runs the healthcheck for caseflow,  but its supported for v0 and v1 of the api.

### DIFF
--- a/modules/appeals_api/app/controllers/appeals_api/metadata_controller.rb
+++ b/modules/appeals_api/app/controllers/appeals_api/metadata_controller.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+require 'appeals_api/health_checker'
+
+module AppealsApi
+  class MetadataController < ::ApplicationController
+    skip_before_action :verify_authenticity_token
+    skip_after_action :set_csrf_header
+    skip_before_action(:authenticate)
+
+    def healthcheck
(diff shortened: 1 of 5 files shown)


modules/appeals_api/app/controllers/appeals_api/metadata_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+    private
+
+    def healthy_service_response
+      {
+        data: {
+          id: 'appeals_healthcheck',
+          type: 'appeals_healthcheck',
+          attributes: {
+            healthy: true,
+            date: Time.zone.now.to_formatted_s(:iso8601)
```
edmangimelli, line 29: this is a full timestamp. is that what we want? or just a date?

modules/appeals_api/app/controllers/appeals_api/metadata_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+               status: :service_unavailable
+      end
+    end
+
+    private
+
+    def healthy_service_response
+      {
+        data: {
+          id: 'appeals_healthcheck',
```
edmangimelli, line 25: i've been using `id: nil`, personally, for dummy resources. Idk if there's been any agreement on what should be used when something is not a db-backed resource

PR #4460, opened 2020-06-30
prevent iconv spec failures

erluti:
## Description of change
Drop usage of iconv, and use ActiveSupport instead.
Removal of right single-quotation marks (unicode 2019) is handled explicitly.

## Original issue
Specs fail due to docker container using an explicit encoding.
related https://app.zenhub.com/workspaces/vsp-5cedc9cce6e3335dc5a49fc4/issues/department-of-veterans-affairs/va.gov-team/7250


app/workers/education_form/forms/base.rb
```diff
@@ -64,14 +64,25 @@ def initialize(app)
...
+    #   Workaround is to sub the ASCII apostrophe, though other non-UTF-8 chars might break specs
+    #
     # Convert the JSON/OStruct document into the text format that we submit to the backend
+    # rubocop:enable Style/AsciiComments
     def format
       @applicant = @form
       # the spool file has a requirement that lines be 80 bytes (not characters), and since they
       # use windows-style newlines, that leaves us with a width of 78
       wrapped = word_wrap(parse_with_template_path(@record.form_type), line_width: 78)
+      wrapped = wrapped.gsub('’', "'")
```
johnpaulashenfelter, line 83: This can go in the `en.yml` file under the transliteration key right?

I hesitate to simply make the spec pass -- if there's a right single quote, there can be a left. Same for double quotes.

I *really* wish we had a better idea if this work was even necessary for the output file.

omgitsbillryan: > I hesitate to simply make the spec pass -- if there's a right single quote, there can be a left. Same for double quotes.

Agreed. Is there a way to just default unknown transliterations to some default character, like `?` for example, or `□`

erluti: It defaults them to `?` automatically, but that fails the spec

erluti: I tried using `en.yml` but it didn't fix the specs https://dsva.slack.com/archives/CQHBJ5U06/p1593465483117100?thread_ts=1593450526.102700&cid=CQHBJ5U06

omgitsbillryan: [Looks like it defaults to `?` for replacements](https://github.com/rails/rails/blob/fbe2433be6e052a1acac63c7faf287c52ed3c5ba/activesupport/lib/active_support/inflector/transliterate.rb#L62).

Don't bother with the gsub imo, just allow the default `?` to occur.

erluti: If this approach is okay, I think we could be a bit more aggressive with quote characters.
I didn't link the other ticket because I wasn't sure if this was ok with everybody.  I'll link it

erluti: > [Looks like it defaults to `?` for replacements](https://github.com/rails/rails/blob/fbe2433be6e052a1acac63c7faf287c52ed3c5ba/activesupport/lib/active_support/inflector/transliterate.rb#L62).
> 
> Don't bother with the gsub imo, just allow the default `?` to occur.

Update the fixtures to show `?`s instead of apostrophes?

omgitsbillryan: I'm happy either way - trying our best to replace quotes or just letting them default to `?`.

omgitsbillryan:
@erluti is this waiting on anything? Can we merge it now?

PR #4462, opened 2020-07-01
fix inbound autologin tracking

ericbuckley:
## Description of change
Last week we created a new session type, `custom`, that is only being used to initiate auto logins when we know the user has an active SSOe session.  Additionally, we decided to set the forceAuthN to true for every SAML request that is NOT an auto login.

vets-website also released some changes today that stopped sending these parameters through as they are no longer needed, however we forgot to also update vets-api to remove any tracking code based on these query parameters.  Notably when authentication fails we need to send the query parameters 'auth=force-needed` back to the FE so the FE knows to silently fail and not show an error screen to the user with auto login fails.

This PR removes the inbound/force tracking codes and updates the `force-needed` logic to check the type rather than the query param.


app/controllers/v1/sessions_controller.rb
```diff
@@ -52,6 +49,7 @@ def ssoe_slo_callback
     end
 
     def saml_callback
+      Rails.logger.info('SSOe SAML received')
```
patrickvinograd, line 52:
```suggestion
      Rails.logger.info('SSOe: SAML Response')
```

app/controllers/v1/sessions_controller.rb
```diff
@@ -133,6 +123,7 @@ def render_login(type, previous_saml_uuid = nil)
                                locals: { url: login_url, params: post_params },
                                format: :html
       render body: result, content_type: 'text/html'
+      Rails.logger.info('SSOe SAML sent')
```
patrickvinograd, line 126:
```suggestion
      Rails.logger.info('SSOe: SAML Request')
```

patrickvinograd: I'd like to also be logging authnContext of both the request and response, but both of those are more involved and could be a separate PR. 

For request, I think it would mean also returning a separate authnContext from the helper methods, or else logging that info directly in the helper, since otherwise it's buried in the form parameters.  

For response, we can log it in the callback_stats method, or anywhere after we've parsed the response, but then we lose the benefit of logging right at the top of the saml_callback method.

patrickvinograd: Just trying to. get apple-apples comparisons. SAML Request/responses can always be in terms of authnContexts; logical logins can be in terms of the "type" from the `new` request.

ericbuckley: I think thats worth doing now.  For the response I think its ok to do it after parsing, we always have the nginx `/v1/sessions/callback` records to compare it to

ericbuckley: @patrickvinograd does this work https://github.com/department-of-veterans-affairs/vets-api/pull/4462/commits/5c072e7250dc67e71209e947e3d81ef223ef84a1

lib/saml/url_service.rb
```diff
@@ -169,7 +169,7 @@ def build_sso_url(link_authn_context)
       new_url_settings = url_settings
       new_url_settings.authn_context = link_authn_context
       saml_auth_request = OneLogin::RubySaml::Authrequest.new
-      save_saml_request_tracker(saml_auth_request.uuid)
+      save_saml_request_tracker(saml_auth_request.uuid, link_authn_context)
```
patrickvinograd, line 172: Needs to be duplicated in post_url_service :/

PR #4453, opened 2020-06-26
Jp/remove flapping spec from coverage

johnpaulashenfelter:
## Description of change

Excluding skipped spec from codecoverage calculation.
## Original issue(s)
department-of-veterans-affairs/va.gov-team#0000

PR #4464, opened 2020-07-01
add rspec-retry, set retries=3 for uploads_request_spec, to unblock production deploys

f1337:
##  Description of change

There is a flapping/blinky spec, which appears to be triggered by an encoding error in the Origami PDF library. We have yet to determine the root cause (it's a difficult one), and in the meantime, we want to unblock production deploys.

We (the BE tools team) identified two immediate ways to unblock deploys:
* Retry the failing spec
* Disable the failing spec

We have an [open PR](https://github.com/department-of-veterans-affairs/vets-api/pull/4453) for disabling the failing spec, but doing so means we lose test coverage for that API endpoint. I don't mean the coverage report, but the loss of actual coverage.

This PR is for the retry option. Introducing `rspec-retry` could, if never removed, theoretically lead to it being used more often, which is not ideal. The trade-off is that we get to retain spec coverage on the API while we investigate further.

## Original issue(s)
department-of-veterans-affairs/va.gov-team/issues/7305

### DIFF
--- a/modules/vba_documents/spec/request/v1/uploads_request_spec.rb
+++ b/modules/vba_documents/spec/request/v1/uploads_request_spec.rb
@@ -6,7 +6,7 @@ require_relative '../../support/vba_document_fixtures'
 require_dependency 'vba_documents/object_store'
 require_dependency 'vba_documents/multipart_parser'
 
-RSpec.describe 'VBA Document Uploads Endpoint', type: :request do
+RSpec.describe 'VBA Document Uploads Endpoint', type: :request, retry: 3 do
   include VBADocuments::Fixtures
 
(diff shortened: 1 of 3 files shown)


modules/vba_documents/spec/request/v1/uploads_request_spec.rb
```diff
@@ -126,6 +126,7 @@
         expect(response.status).to eq(404)
       end
     end
+
     it 'returns a 200 with content-type of zip' do
```
johnpaulashenfelter, line 130: @f1337 i _think_ its always this spec that fails, at least that's the one I've seen a lot. Don't know if it matters or not but seems like the retry should be as close to the source as possible.

PR #4469, opened 2020-07-01
VAOS refresh session token fix

kreek:
## Description of change
This fixes a PR that was rolled back because background jobs weren't firing and a test was failing in CI. 

VAOS sessions don't automatically extend. PR adds logic to run a background job that extend the VAOS session. Uses 1 min distributed lock via redis so that multiple api calls, especially when initializing the app for a user, don't result in multiple jwt refreshes.

## Original issue(s)

* adds an entry to config/redis.yml
* updates the the vaos gemspec to include sidekiq (now matches other modules that use sidekiq)
* adds CloudWatch logs around session creation and update

Tested locally and on a review instance, will do integration tests on staging.


modules/vaos/vaos.gemspec
```diff
@@ -6,15 +6,23 @@ $LOAD_PATH.push File.expand_path('lib', __dir__)
...
-Gem::Specification.new do |spec|
-  spec.name        = 'vaos'
-  spec.version     = VAOS::VERSION
-  spec.authors     = ['Alastair Dawson']
-  spec.email       = ['alastair.j.dawson@gmail.com']
-  spec.homepage    = 'https://api.va.gov'
-  spec.summary     = 'VAOS API'
-  spec.description = 'VOAS (VA Online Scheduling) is part of VAMF (VA Mobile Framework) and allows veterans to make and manage appointments for care'
-  spec.license     = 'MIT'
+Gem::Specification.new do |s|
```
erluti, line 9: why change this variable to `s`?

kreek: It was to match the way the other modules do it.

PR #4471, opened 2020-07-02
Appeals API - nest /contestable_issues under /higher_level_reviews

edmangimelli:
go from `/decision_reviews/contestable_issues` to `/decision_reviews/higher_level_reviews/contestable_issues/{benefit_type}`

Now that we're tackling a new decision review type (NOD), `/decision_reviews/contestable_issues` cannot remain hardwired for returning HLR contestable issues specifically (HLR-compensation contestable issues, to be exact).

Nesting `/contestable_issues` under `/higher_level_reviews` is the most _REST_ -ful solution, prevents polluting our headers, and allows us to continue making additive changes (and push a major version change further down the road).

continues https://github.com/department-of-veterans-affairs/caseflow/pull/14618
to resolve https://vajira.max.gov/browse/API-1320

Acceptance Criteria:
- [x] tests pass
- [x] documentation updated and tested


lib/caseflow/service.rb
```diff
@@ -76,8 +81,8 @@ def healthcheck
 
     private
 
-    def request_headers(additional_headers = {})
-      DEFAULT_HEADERS.merge(additional_headers)
+    def authorized_perform(method, path, params, additional_headers = nil, options = nil)
+      perform(method, path, params, DEFAULT_HEADERS.merge(additional_headers || {}), options)
```
edmangimelli, line 85: i'd prefer this to match the arity of `perform` (going against code climate's suggestion)

bastosmichael: I don't have an issue with this if we are the only ones really using this lib for now.

PR #4474, opened 2020-07-02
GIDS: Split out search timeouts

zurbergram:
## Description of change
Because of the addition of a fuzzy search to 2 GIDS endpoints in https://github.com/department-of-veterans-affairs/gibct-data-service/pull/677 the timeout for search for institutions or vet tec programs needs to be increased.

These changes break out the 2 searches to have their own configuration and configurable settings `gids.search.read_timeout` and `gids.search.open_timeout`.  The current timeout settings remain in place for other calls to GIDS.

## Original issue(s)

* uses `gids.search.read_timeout` or `gids.search.open_timeout` settings if they exist

## Testing
This will be tested in Staging as currently when the feature flag `gibct_search_enhancements` is enabled the CT has timeout issues when searching for an institution or VET TEC program


app/models/gids_redis.rb
```diff
@@ -19,6 +19,10 @@ def method_missing(name, *args)
       self.rest_call = name
       self.scrubbed_params = args.first
       response_from_redis_or_service.body
+    elsif search_respond_to?(name)
+      self.rest_call = name
```
LindseySaari, line 23: Since both conditionals assign self.rest_call and self.scrubbed_params, I think the repeated code can be moved out and refactored

app/models/gids_redis.rb
```diff
@@ -36,7 +44,17 @@ def response_from_redis_or_service
     end
   end
 
+  def response_from_redis_or_search_service
+    do_cached_with(key: rest_call.to_s + scrubbed_params.to_s) do
+      gi_search_service.send(rest_call, scrubbed_params)
```
LindseySaari, line 49: Wondering if this can be refactored/cleaned up somehow. This code is almost identical to response_from_redis_or_service minus the use of gi_search_service vs  gi_service

app/models/gids_redis.rb
zurbergram: setting these properties regardless of conditional path

app/models/gids_redis.rb
zurbergram: refactored

app/models/gids_redis.rb
zurbergram: changed

PR #4475, opened 2020-07-02
Send PDF to CARMA

kevinmirc:
Resolves https://github.com/department-of-veterans-affairs/va.gov-team/issues/8871

Send the PDF version of the 10-10CG as an attachment to the 10-10CG CARMA Case (Submission).

## Description of change
- Add methods to `CARMA::Client` for sending attachments
  - [x] Add VCR Cassettes
  - [x] Add stub method

- Add `CARMA::Models::Attachments` and `CARMA::Models::Attachment`
  - [x] Should build a proper request payload
  - [x] Should contain documentation
  - [x] Should support POA and 10-10CG doc types

- Update `Form1010cg::Service` to send attachments after submission
  -  [x] Failure should not prevent user from receiving a successful response

## Other changes
- [x] Can generate PDF of claim with guid as file name (vs the claim.id)
- [x] Convert CARMA::Client::Client to a singlton
  - Token Refresh work will be done as followup: https://github.com/department-of-veterans-affairs/va.gov-team/issues/6686
- [x] Reorganize CARMA VCR Cassettes
  - Split them up by requests (one request per cassettes)
  - Rename directories


LindseySaari:
Is this ready for Review?

kevinmirc:
@LindseySaari Yes, ready for review. Thanks!

lib/carma/models/attachment.rb
```diff
@@ -0,0 +1,126 @@
...
+          # property:   VersionData
+          # value:      The base64 encoded binary of the pdf file content.
+          # comments:   The base64 encoded binary of the pdf file content.
+          #             service call response(carmacase.id).
+          # examples:   'JVBERi0xLjMKJcTl8uXrp<.....rest of the base64 ecoded pdf file content>'
+          'VersionData' => as_base64
+        }
+      end
+
+      def to_h
```
LindseySaari, line 104: Maybe a nitpick, but the to_h method naming here may be confusing with ruby to_h

LindseySaari:
Can you explain the "Convert CARMA::Client::Client to a singleton" changes please?

app/services/form1010cg/service.rb
```diff
@@ -32,14 +34,56 @@ def initialize(claim)
...
+        )
+
+        carma_attachments.add(CARMA::Models::Attachment::DOCUMENT_TYPES['10-10CG'], file_path)
+
+        carma_attachments.submit!
+        submission.attachments = carma_attachments.to_h
+      rescue
+        # Regardless of the reason, we should not raise an error when sending attachments fails.
+        # If we made it this far, there is a submission that exists in CARMA, so the user should
+        # get a sucessful response whether attachments reach CARMA or not.
```
LindseySaari, line 80: I guess i'm not following why one wouldn't care if the attachments don't successfully submit to CARMA

lib/carma/client/client.rb
```diff
@@ -43,6 +38,34 @@ def create_submission_stub(_payload)
...
+        with_monitoring do
+          client.post(
+            '/services/data/v47.0/composite/tree/ContentVersion',
+            payload,
+            'Content-Type': 'application/json'
+          ).body
+        end
+      end
+
+      def upload_attachments_stub(_payload)
```
LindseySaari, line 52: Can this stub be moved into the test suite? Not sure if there's a specific reason it needs to live here

LindseySaari: Ah, I see it's used with Flipper & stub_carma_responses currently

lib/carma/models/attachment.rb
kevinmirc: That was kind of my intention, would you suggest `#to_hash` instead, to avoid confusion?

app/services/form1010cg/service.rb
kevinmirc: The PDF we're sending to CARMA is just the submission, itself, as a PDF. This is to follow the current conventions of CARMA: every case has the PDF of the submission attached. The people who process the claims prefer to have the PDF as a reference.

So the end-user doesn't know an attachment is being sent with the submission at all. We also don't want to have an error bubble up to the response, because that will mislead the user to thinking the submission failed when it didn't.

I can update the code comment to be more descriptive of the business requirement here.

lib/carma/client/client.rb
kevinmirc: Will add a comment here for clarity.

app/services/form1010cg/service.rb
LindseySaari: Ok, makes sense now. An updated comment would be great!

lib/carma/models/attachment.rb
LindseySaari: Yeah I think so, it's more clear to follow I think

app/services/form1010cg/service.rb
```diff
@@ -78,9 +78,16 @@ def submit_attachment # rubocop:disable Metrics/MethodLength
...
+        # The end-user doesn't know an attachment is being sent with the submission at all. The PDF we're
+        # sending to CARMA is just the submission, itself, as a PDF. This is to follow the current
+        # conventions of CARMA: every case has the PDF of the submission attached.
+        #
+        # Regardless of the reason, we shouldn't raise an error when sending attachments fails.
+        # It's non-critical and we don't want the error to buble up to the response,
+        # misleading the user to think thier claim was not submitted.
+        #
+        # If we made it this far, there is a submission that exists in CARMA.
+        # So the user should get a sucessful response, whether attachments reach CARMA or not.
```
LindseySaari, line 90: 👍

lib/carma/client/client.rb
```diff
@@ -49,6 +50,7 @@ def upload_attachments(payload)
         end
       end
 
+      # Used for Feature Flipper :stub_carma_responses
```
LindseySaari, line 53: 👍

kevinmirc:
Thanks for the review!

PR #4476, opened 2020-07-02
Pv saml request statsd

patrickvinograd:
## Description of change
Adds a statsd metric for SAML requests, for matching up against SAML callbacks.

* Will be able to add to SiteAuthentication or SSOe Launch dashboards.

### DIFF
--- a/app/controllers/v0/apidocs_controller.rb
+++ b/app/controllers/v0/apidocs_controller.rb
@@ -122,6 +122,7 @@ module V0
       Swagger::Requests::BackendStatuses,
       Swagger::Requests::BB::HealthRecords,
       Swagger::Requests::BurialClaims,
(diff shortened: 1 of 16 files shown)


ericbuckley:
@patrickvinograd I added a small fix for a bug I noticed in testing the error page https://github.com/department-of-veterans-affairs/vets-api/pull/4476/commits/e6608c949265af1be6484a14a2ecd884cb7decc6

patrickvinograd:
> LGTM; would we want to add the same stat to v0?

In principle yes, but it will be kind of a hassle since the upleveling for v0 is embedded in the URLService class. We can refactor that to do the same 'should_uplevel? check' as v1 does but still need a way of getting the requested authnContext back from URLService.

PR #4485, opened 2020-07-06
BDD: Default values isVaEmployee and standardClaim

SarahJaine:
## Description of change

EVSS's xsd for the 526 claim form shows the fields isVaEmployee and standardClaim as optional/not required.  However, we noticed upon submission of BDD 526 claims, EVSS was returning errors telling us that both these fields are conditionally required under some circumstances for the BDD form.  To simplify this we're defaulting both values to `false`. 

## Original issue(s)
closes department-of-veterans-affairs/va.gov-team#10861

### DIFF
--- a/spec/support/disability_compensation_form/all_claims_fe_submission.json
+++ b/spec/support/disability_compensation_form/all_claims_fe_submission.json
@@ -114,7 +114,6 @@
         }
       ]
     },
-    "standardClaim": false,
     "isTerminallyIll": false,
     "vaTreatmentFacilities": [
       {
--- a/lib/evss/disability_compensation_form/data_translation_all_claim.rb
+++ b/lib/evss/disability_compensation_form/data_translation_all_claim.rb
@@ -45,7 +45,7 @@ module EVSS
       #
       def translate
         output_form['claimantCertification'] = true
-        output_form['standardClaim'] = input_form['standardClaim']
+        output_form['standardClaim'] = input_form['standardClaim'] || false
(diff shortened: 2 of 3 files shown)


spec/support/disability_compensation_form/all_claims_fe_submission.json
```diff
@@ -114,7 +114,6 @@
         }
       ]
     },
-    "standardClaim": false,
```
LindseySaari, line 117: is standardClaim optional?

SarahJaine: So, what I was going to say is: The data processing looks like FE => BE form translation => EVSS.`standardClaim` is not always passed from the FE to the BE.  So this PR is to force the value to `false` if it was not provided in the FE data.  EVSS requires a value in some circumstances.
But now that I'm looking at the form, the question that creates the data for this field is required. (See screenshot below.) I'm going to take out the changes that have to do with setting a default value for `standardClaim` because I think it's redundant.

Screen Shot 2020-07-07 at 10 07 33 AM

LindseySaari: Sounds good, just tag as Ready for Review when you're all set.

SarahJaine:
After some discussion with @annaswims, it seems more logical to move this to this communication futher up the stack.  By adding the fields as required in vets-json-schema we'll communicate this need with the FE.  Work is being completed in https://github.com/department-of-veterans-affairs/vets-json-schema/pull/450

PR #4486, opened 2020-07-07
Attach Veteran's Status to 10-10CG Submissions

kevinmirc:
Fixes: https://github.com/department-of-veterans-affairs/va.gov-team/issues/9832

**Description**
Attach Veteran's Status to 10-10CG Submissions.

**Considerations**
This will likely have a merge conflict with my other PR: https://github.com/department-of-veterans-affairs/vets-api/pull/4475.

### DIFF
--- a/app/services/form1010cg/service.rb
+++ b/app/services/form1010cg/service.rb
@@ -70,8 +70,7 @@ module Form1010cg
         }
       end
 
-      # Set the veteran status on the :veteran namespace of metadata
-      metadata[:veteran][:is_veteran] = false
+      metadata[:veteran][:is_veteran] = is_veteran('veteran')
 
       metadata
     end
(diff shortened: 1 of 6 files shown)


app/services/form1010cg/service.rb
```diff
@@ -112,9 +111,8 @@ def is_veteran(form_subject) # rubocop:disable Naming/PredicateName
       return @cache[:veteran_statuses][form_subject] = false if icn == NOT_FOUND
 
       response = EMIS::VeteranStatusService.new.get_veteran_status(icn: icn)
-      raise response.error if response.error?
 
-      is_veteran = response&.items&.first&.title38_status_code == 'V1'
+      is_veteran = response.try(:items)&.first&.title38_status_code == 'V1'
```
LindseySaari, line 115: Curious as to why .try is used instead of &. here

kevinmirc: I was assuming that response could either be an `EMIS::Responses::Response` or some EMIS Error object. But on second look, it will always be an `EMIS::Responses::Response` and therefore will always have an `#items` method. 

I'll update to use `&`. Thanks for pointing this out.

PR #4487, opened 2020-07-07
Render "Backend Service Outage" on invalid veteran status for 10-10CG submissions

kevinmirc:
Fixes: https://github.com/department-of-veterans-affairs/va.gov-team/issues/10583

### DIFF
--- a/app/controllers/v0/caregivers_assistance_claims_controller.rb
+++ b/app/controllers/v0/caregivers_assistance_claims_controller.rb
@@ -5,6 +5,8 @@ module V0
   class CaregiversAssistanceClaimsController < ApplicationController
     skip_before_action(:authenticate)
 
+    rescue_from ::Form1010cg::Service::InvalidVeteranStatus, with: :backend_service_outage
+
     def create
       return service_unavailable unless Flipper.enabled?(:allow_online_10_10cg_submissions)
 
@@ -26,6 +28,10 @@ module V0
 
     def service_unavailable
       render nothing: true, status: :service_unavailable, as: :json
+    end
+
+    def backend_service_outage
+      render_errors Common::Exceptions::ServiceOutage.new(nil, detail: 'Backend Service Outage')
     end
   end
 end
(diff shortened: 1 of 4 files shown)


app/controllers/v0/caregivers_assistance_claims_controller.rb
```diff
@@ -5,6 +5,8 @@ module V0
   class CaregiversAssistanceClaimsController < ApplicationController
     skip_before_action(:authenticate)
 
+    rescue_from ::Form1010cg::Service::InvalidVeteranStatus, with: :backend_service_outage
```
LindseySaari, line 8: Wondering if you could extend the rescue_from behavior from the application_controller.rb to rescue_from Common::Exceptions::ServiceOutage instead of raising InvalidVeteranStatus https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/controllers/application_controller.rb#L115

Render errors is called here: https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/controllers/application_controller.rb#L126

On a side note: Does Backend Service Outage make sense here when an ICN isn't found?

LindseySaari: I think I answered my question as to why Backend Service Outage Exception is being raised from the attached ticket: " we need to ensure that the same error appears regardless of why the application was rejected " -- Makes sense.

Still curious on the first question though regarding the rescue_from behavior.

kevinmirc: Yes, the backend outage error is for ambiguity.

Regarding the **rescue_from** logic:

The `ApplicationController` seems out of scope for this. I could see it being moved to the `ApplicationController` if multiple controllers needed to handle this error. This is a special case, so I did not want to fill the application-wide error handling logic with a one-off exception.

I chose to use the custom error `Form1010cg::Service::InvalidVeteranStatus` for clarity. It's easier to interpret the source of the error. If I were to raise a `Common::Exceptions::ServiceOutage` instead, it could mislead a developer to thinking an external API call raised an error (if not looking at the stack trace). With this error, it's explicitly written in the controller that "InvalidVeteranStatus renders as a BackendServiceOutage."

LindseySaari: Ok, great! WRT clarity, I think this makes sense! LGTM

PR #4494, opened 2020-07-08
Update 10203 prefill

dmccurdybah:
## Description of change
Updated the 10203 form profile / prefill handler.  The existing form profile was a placeholder and all updates were made to match the actual 10203 form.

The 10203 isn't live yet, so any updates will only affect our testing in staging.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#10549

- The 10203 form is behind the `edu_benefits_stem_scholarship` feature flag.

- I updated the an existing GIDS VCR cassette in order to make the it match data in an existing EVSS cassette. It was a small change to an institution ID and I'm familiar with GIDS.

- Tested locally and QA reviewed.

- Mockdata PR: https://github.com/department-of-veterans-affairs/vets-api-mockdata/pull/144


app/models/form_profiles/va10203.rb
```diff
@@ -1,11 +1,88 @@
...
+    service = EVSS::GiBillStatus::Service.new(user)
+    service.get_gi_bill_status
+  rescue => e
+    Rails.logger.error "Failed to retrieve GiBillStatus data: #{e.message}"
+    {}
+  end
+
+  def initialize_entitlement_information(gi_bill_status)
+    return {} if gi_bill_status == {}
+    return {} if gi_bill_status.remaining_entitlement.blank?
```
LindseySaari, line 61: You could make the 'return {}' a one liner by combining the if's with an || or maybe it's more readable the current way?

dmccurdybah: Sounds good.  No preference here!

PR #4495, opened 2020-07-08
add Digital Signature for 21-22 paragraph

ryanlink:
Suggested changes by @darduda 

Documentation updates for the Benefits Claims API's new 21-22 endpoint


modules/claims_api/app/swagger/claims_api/description/v1.md
```diff
@@ -63,6 +63,24 @@ By default, the API generates PDFs of form 21-526 based on incoming data, and th
 
 If an authorized representative is filing an original claim, rather than the Veteran (i.e. the OAuth token is not the Veteran’s), use the [PUT /forms/526/{id}](#operations-Disability-upload526Attachment) endpoint to upload a scanned PDF of the form signed in ink by the Veteran.
 
+### 21-22 (Power of Attorney) and Digital Signatures
+
+The endpoint accepts a JSON payload via POST. Full URL, including query parameters.
```
ryanlink, line 68: This is on the endpoint description as well.  Don't all endpoints require a full URL?

darduda: Yes, I think. I suppose I'm not sure what the "Full URL, including query parameters" is supposed to mean in this context. I'm okay with removing that or changing it, if we can't figure out a good reason why it's called out here. Otherwise, looks good.

PR #4496, opened 2020-07-08
Temporarily disable SSOe Flipper toggles for `development` environments

jholton:
## Description of change
Temporarily disable SSOe Flipper toggles for `development` environments until SSOe is working there.
This will ensure that running `vets-api` locally and in review instances will use the `v0` (non-SSOe) login path.

## Original issue(s)


config/initializers/flipper.rb
```diff
@@ -47,6 +47,10 @@
     # default features to enabled for development and test only
     Flipper.enable(feature) if Rails.env.development? || Rails.env.test?
```
jholton, line 48: Alternatively, we could change this line to 
` Flipper.enable(feature) if (Rails.env.development? || Rails.env.test?) && !feature.match(/ssoe/)`
but that wouldn't help local users who already have the toggle set in their local DB.

PR #4497, opened 2020-07-08
Create MVI::Models::MviUserAttributes

kevinmirc:
Fixes: https://github.com/department-of-veterans-affairs/va.gov-team/issues/9083
Context here: https://github.com/department-of-veterans-affairs/vets-api/pull/4064#discussion_r424115957

**Description**
Since there are now two places in the codebase where we do MPI Profile searches with only a user's attributes (vs a UserIdentity), we need to create a UserAttributes object in the MVI lib that other services can use for their search.

**Summary**
- [x] Move `HCA::UserAttributes` to MVI lib.
- [x] Update Form1010cg::Service to use `MviUserAttributes` instead of `UserIdentity`

### DIFF
--- a/app/services/form1010cg/service.rb
+++ b/app/services/form1010cg/service.rb
@@ -138,7 +138,11 @@ module Form1010cg
       cached_icn = @cache[:icns][form_subject]
       return cached_icn unless cached_icn.nil?
 
-      response = mvi_service.find_profile(build_user_identity_for(form_subject))
+      response = mvi_service.find_profile(
+        build_mvi_user_attributes_for(
+          form_subject
+        )
+      )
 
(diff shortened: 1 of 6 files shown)


annaswims:
Is this still in progress? If so, I think that @jholton should be a reviewer before it's merged.

kevinmirc:
I'm closing this PR until we can find a path forward. The current `MVI::AttrService` is bound to HCA. So I'd either need to create another `AttrService` for CARMA, or make the HCA's `AttrService` more generic. Not sure which pattern is preferred.

More info is in our original thread: https://github.com/department-of-veterans-affairs/vets-api/pull/4064#discussion_r452275225

PR #4498, opened 2020-07-08
Logs diagnostics about sessions without ID.me UUID

patrickvinograd:
## Description of change
We want to consider consuming ID.me UUIDs from the accounts table but first are going to collect data about what percentage of inbound users this would help. 

No PII concerns, data is logged to cloudwatch and contains no user-specific identifiers.

### DIFF
--- a/app/controllers/v1/sessions_controller.rb
+++ b/app/controllers/v1/sessions_controller.rb
@@ -180,9 +180,10 @@ module V1
     # would be resolved by an account lookup, before we implement that
     def log_missing_uuid_info(exception)
       return if exception&.identifier.blank?
+
       accounts = Account.where(icn: exception.identifier)
-      Rails.logger.info("SSOe: Account UUID mapping NOT FOUND") if accounts.blank?
-      Rails.logger.info("SSOe: Account UUID mapping FOUND - #{accounts.size} entries") unless accounts.blank? 
+      Rails.logger.info('SSOe: Account UUID mapping NOT FOUND') if accounts.blank?
+      Rails.logger.info("SSOe: Account UUID mapping FOUND - #{accounts.size} entries") if accounts.present?
     end
 
     def new_stats(type)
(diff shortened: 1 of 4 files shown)


app/controllers/v1/sessions_controller.rb
```diff
@@ -175,6 +176,15 @@ def user_logout(saml_response)
       end
     end
 
+    # Diagnostic logging to determine what percentage of these issues
+    # would be resolved by an account lookup, before we implement that
+    def log_missing_uuid_info(exception)
+      return if exception&.identifier.blank?
+      accounts = Account.where(icn: exception.identifier)
```
ericbuckley, line 183: would attempting to match by icn OR secid increase our odds?
```
accounts = Account.where(icn: exception.attributes[:mhv_icn]).or(Account.where(sec_id: exception.attributes[:sec_id])
```

lib/saml/errors.rb
```diff
@@ -19,11 +19,14 @@ class UserAttributeError < SAMLError
...
 
-    def initialize(message:, code:, tag:)
+    attr_reader :identifier
+
+    def initialize(message:, code:, tag:, identifier: nil)
       @code = code
       @tag = tag
       @level = :warning
       @context = {}
+      @identifier = identifier
```
ericbuckley, line 29: Do we need to filter this in sentry?

app/controllers/v1/sessions_controller.rb
patrickvinograd: Technically yes but I just checked production and there are 12 accounts with a secid but no ICN. Whereas there are 2,192,808 accounts with an ICN.

lib/saml/errors.rb
patrickvinograd: I don't think so; ICN is fair game to log, like ID.me UUID and secid. But we should go through and bring https://github.com/department-of-veterans-affairs/vets-api/pull/4095 to completion.

app/controllers/v1/sessions_controller.rb
ericbuckley: I was thinking about the case where the secid would be in the SAML Response, but not the mhv_icn.

PR #4499, opened 2020-07-09
Update Sidekiq to enable rack upgrade

johnpaulashenfelter:
Sidekiq 5.2.8 had blocked upgrading the rack gem that mitigated CVE-2020-8184.
The be-tools team applied the recommended patch, but with *several* `rack` CVEs
either mitigated through monkeypatching or deemed low-risk, the tech debt was piling up.

The Sidekiq team released a point release of the older 5.x branch that relaxed the
overerly tight dependency on `rack` versions so these CVE fix releases could be applied.

## Description of change

Sidekiq upgraded to support `rack` upgrade and CVE mitigation

## Original issue(s)
department-of-veterans-affairs/va.gov-team#9057


rakelib/security.rake
```diff
@@ -14,7 +14,7 @@ task security: :environment do
 
   puts 'running bundle-audit to check for insecure dependencies...'
   exit!(1) unless ShellCommand.run('bundle-audit update')
-  audit_result = ShellCommand.run('bundle-audit check --ignore CVE-2020-8161 CVE-2020-8184')
+  audit_result = ShellCommand.run('bundle-audit check')
```
LindseySaari, line 17: Thx!

PR #4503, opened 2020-07-10
Inject ID.me UUID from accounts table

patrickvinograd:
# Description of change
Detects a SAML payload that is missing an ID.me UUID, and attempts to look one up (based on the presented ICN) in the Accounts table. If a unique mapping is found, injects the idme_uuid into the User/UserIdentity before establishing a session.

There may be users who have ICNs that map to multiple ID.me UUIDs. For now, we are ignoring those until we can decide how to handle them correctly.

We found that ~80% of inbound SSOe sessions did not have an ID.me UUID present.
Of those, between 80-85% of a brief sample were present in the accounts table, and by doing this injection we let those users have auto-login without needing to re-authenticate.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/11096

## Feedback Requested
Should I remove the diagnostic logging from the sessions controller, and/or replace it with a log message indicating whether or not we were able to do this injection?

### DIFF
--- a/app/controllers/v1/sessions_controller.rb
+++ b/app/controllers/v1/sessions_controller.rb
@@ -178,6 +178,8 @@ module V1
 
     # Diagnostic logging to determine what percentage of these issues
     # would be resolved by an account lookup, before we implement that
+    # TODO: Remove this method after we're confident in the UUID injection
+    # performed in UserSessionForm
     def log_missing_uuid_info(exception)
       return if exception&.identifier.blank?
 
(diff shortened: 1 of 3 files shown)


patrickvinograd:
> LGTM as for "Should I remove the diagnostic logging from the sessions controller, and/or replace it with a log message indicating whether or not we were able to do this injection?" -- Do you need the diagnostic logging anymore? I think a message indicating whether or not the injection was successful would be helpful as well.

I added a log line when the injection is successful; I left the diagnostic logging in for now, with a TODO to remove it once we're comfortable with how this is working.

PR #4504, opened 2020-07-10
PDF Generation for Form 21-686C

kathleencrawford:
## Description of change

Adding the ability to generate a PDF for the form 21-686C.  This PR utilizes existing PDF generation code, while adding an implementation for the 21-686C.  The form 21-686C allows the user to add or remove dependents to their benefits.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#8759

There are no additional settings or logging added for this PR.  The PDF generation functionality already exists, this PR just adds form 21-686C to the existing functionality.

Regarding Code Climate issues...yes, there are a lot of lines of code in one file and lots of "similar blocks" of code.  This PDF is 15 pages long, it is a very complex form, that is why that one file is so big.  Most of it is just mapping the input fields from the PDF (there just happen to be so many of them on this form).  

## Testing

- [x] spec updated
- [x] local testing


lib/pdf_fill/forms/form_helper.rb
```diff
@@ -43,6 +45,7 @@ def split_postal_code(address)
         return if address.blank?
 
         postal_code = address['postalCode']
+        postal_code = address['zip_code'] if postal_code.blank?
```
digitaldrk, line 48: Does this work?

`postal_code = address['postalCode'] || address['zip_code']`

if so you could use that pattern in other areas like this :)

lib/pdf_fill/forms/va686c674.rb
```diff
@@ -0,0 +1,1833 @@
...
+              'last' => {
+                key: 'form1[0].#subform[18].CHILDLastName[%iterator%]',
+                limit: 18,
+                question_num: 14,
+                question_suffix: 'C',
+                question_text: 'PREVIOUS MARRIAGE HISTORY >  PREVIOUS SPOUSE LAST NAME'
+              }
+              # 'suffix' => {
+              # @TODO
+              # } # end of suffix
```
digitaldrk, line 460: Do you still need this?

lib/pdf_fill/forms/va686c674.rb
```diff
@@ -0,0 +1,1833 @@
...
+              },
+              'last' => {
+                key: 'veteranSpouse.previousSpouse.lastName[%iterator%]',
+                limit: 18,
+                question_num: 15,
+                question_suffix: 'C',
+                question_text: 'PREVIOUS MARRIAGE HISTORY >  SPOUSES PREVIOUS SPOUSE LAST NAME'
+              },
+              'suffix' => {
+                # @TODO not in the form
```
digitaldrk, line 580: Need the TODO?

lib/pdf_fill/forms/va686c674.rb
```diff
@@ -0,0 +1,1833 @@
...
+            }, # end of start location
+            'reason_marriage_ended' => {
+              'death' => { key: 'spouse_marriage_history.reason_marriage_ended.death[%iterator%]' },
+              'divorce' => { key: 'spouse_marriage_history.reason_marriage_ended.divorce[%iterator%]' },
+              'annulment' => { key: 'spouse_marriage_history.reason_marriage_ended.annulment[%iterator%]' },
+              'other' => { key: 'spouse_marriage_history.reason_marriage_ended.other[%iterator%]' }
+            },
+            'reason_marriage_ended_other' => {
+              key: 'spouse_marriage_history.reason_marriage_ended_other[%iterator%]',
+              # limit: 12, # @TODO THIS BREAKS
```
digitaldrk, line 631: Marking this one for visibility. Looks like it's not needed

lib/pdf_fill/forms/form_helper.rb
kathleencrawford: I've updated this as suggested - thanks!

lib/pdf_fill/forms/va686c674.rb
kathleencrawford: This is due to some differences between the front end and paper form, we still need to make a determination on how to handle these use cases.  I left the TODO comments in there as markers for when a decision is made and this is revisited.  I've added comments to hopefully make this a bit more clear.

lib/pdf_fill/forms/va686c674.rb
kathleencrawford: See above note about comments - thanks :)

lib/pdf_fill/forms/va686c674.rb
kathleencrawford: This has been removed, thanks!

PR #4508, opened 2020-07-10
Revert "Update vets-json-schema"

SarahJaine:
We need to quickly revert this before this change is deployed to production today.  I was under the impression that the FE is always sending us this field. Through staging testing we discovered that it is possible to not answer this question.  This schema mismatch will make some form526 submissions fail.

Reverts department-of-veterans-affairs/vets-api#4502


SarahJaine:
Reverting this change would make it so that BDD claims no longer fail.  However, BDD claims are still gated out of production by the flipper ui.

PR #4513, opened 2020-07-11
Benefits Intake API - Fix Swagger Errors

edmangimelli:
resolves https://vajira.max.gov/browse/API-1425

### DIFF
--- a/modules/vba_documents/app/swagger/vba_documents/v0/controller_swagger.rb
+++ b/modules/vba_documents/app/swagger/vba_documents/v0/controller_swagger.rb
@@ -64,6 +64,7 @@ module VbaDocuments
           end
 
           response 400 do
+            key :description, 'Document upload failed'
             content 'application/xml' do
               schema do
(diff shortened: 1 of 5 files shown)


modules/vba_documents/app/swagger/vba_documents/v0/controller_swagger.rb
```diff
@@ -92,7 +93,10 @@ class ControllerSwagger
             key :description, 'ID as returned by a previous create upload request'
             key :required, true
             key :example, '6d8433c1-cd55-4c24-affd-f592287a7572'
-            key :type, :string
+            schema do
+              key :type, :string
+              key :format, :uuid
```
bastosmichael, line 98: Does this not render correctly in swagger UI?

edmangimelli: type has to be nested under schema.
my changes work locally (against dev portal running locally)

PR #4511, opened 2020-07-10
Add form526 bdd tests

annaswims:
## Description of change
Add more specs for form 526 BDD 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#10368

PR #4514, opened 2020-07-13
526: Remove V1 Claims Code and Configurations

annaswims:
## Description of change
we no longer use the form526 v1/ Increase only disability form so we can delete the related code. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#7812


annaswims:
https://github.com/department-of-veterans-affairs/vets-api/pull/4511 adds back some of the coverage that was reduced by this PR

app/models/form526_submission.rb
```diff
@@ -44,21 +44,17 @@ class Form526Submission < ApplicationRecord
...
+  def start
     workflow_batch = Sidekiq::Batch.new
     workflow_batch.on(
       :success,
       'Form526Submission#perform_ancillary_jobs_handler',
       'submission_id' => id
     )
     jids = workflow_batch.jobs do
-      klass.perform_async(id)
+      EVSS::DisabilityCompensationForm::SubmitForm526AllClaim.perform_async(id)
```
LindseySaari, line 57: Nice refactor here... I think it's more visually appealing

spec/request/disability_compensation_form_request_spec.rb
```diff
@@ -157,7 +141,7 @@
 
     context 'with invalid json body' do
       it 'returns a 422' do
-        post '/v0/disability_compensation_form/submit', params: { 'form526' => nil }.to_json, headers: headers
+        post '/v0/disability_compensation_form/submit_all_claim', params: { 'form526' => nil }.to_json, headers: headers
```
LindseySaari, line 144: Were there not specs covering the ‘/v0/disability_compensation_form/submit_all_claim’ endpoint to test 422? I see you’re replacing /submit with /submit_all_claim but I would expect a spec to just be deleted

annaswims:
I see I've dropped coverage in app/workers/evss/disability_compensation_form/submit_form526.rb and will make sure to add it back.

spec/request/disability_compensation_form_request_spec.rb
annaswims: there wasn't a spec for submit_all_claim - 422 before. There should've but there was't.  These two endpoints share a ton of code because essentailly `submit_all_claim` does everything `submit` does and more... but perhaps with a slightly different json strucutre.

LindseySaari: Ok great, thanks for the reply! Looks good

PR #4519, opened 2020-07-13
update swagger /request and /response directory nesting

LindseySaari:
# Description of change
app/swagger/requests/address.rb defines Swagger::Requests::Address but Zeitwerk mode expects it to be nested within another /swagger directory. There needs to be a nested swagger directory to support that namespacing (ie app/swagger/swagger/requests/address.rb) in both /requests and /responses. This PR moves both the /requests and /responses directories into the /swagger/swagger directory

## Original issue(s)
department-of-veterans-affairs/va.gov-team#11175


omgitsbillryan:
What about moving the entire `app/swagger` directory into `/lib` instead? Would that allow us to avoid the swagger/swagger redundancy?

erluti:
> What about moving the entire `app/swagger` directory into `/lib` instead? Would that allow us to avoid the swagger/swagger redundancy?

I like this idea.  We would then need to add a lot of `requires` in places since lib won't autoload anymore, so I didn't do it in my original PR.  I also didn't know if there were any other impacts of having the swagger generators in a different directory because I didn't dig into how the swagger generators work.

Maybe it could be a tech debt ticket to come back and move if it's complicated.

LindseySaari:
> What about moving the entire `app/swagger` directory into `/lib` instead? Would that allow us to avoid the swagger/swagger redundancy?

I agree with Philip here. The app (and swagger) directory is autoloaded where as lib is not. This would result in many additional requires across the codebase. **Note: We had a custom solution in place to autoload lib, but this is no longer the case after our changes WRT Zeitwerk mode

PR #4520, opened 2020-07-13
Add a fake clamav scanner for local development

johnpaulashenfelter:
## Description of change

Add a fake ClamAV file and configuration example for local development to avoid the complexity of installation and configuration required by ClamAV for native OSX and Linux/Ubuntu installations. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#11176

* New docs in the native install about running the fake scanner
* Tested using local installs of vets-website and vets-api with user `vets.gov.user+17@gmail.com` completing form at http://localhost:3001/pension/application/527EZ/introduction and uploading _multiple_ files.

### DIFF
--- a/bin/fake_clamdscan
+++ b/bin/fake_clamdscan
@@ -1,0 +1,3 @@
+#!/bin/bash
+echo "Fake scanning file with fake_clamdscan script"
+exit 0
--- a/docs/setup/native.md
+++ b/docs/setup/native.md
@@ -49,6 +49,35 @@ To start, fetch this code:
    ```
 1. Run `bin/setup` to setup the database and start the server.
 
+## Settings and configuration
+
+We use the `config` gem to manage settings in the application. Local
+settings for each developer should be managed in your own local `config/settings.local.yml` file, which
+by default can override the standard configuration *and* is excluded from
+source control so the settings can persist.
+
+There is a basic configuration file you can copy as a starting point in the `config` directory:
+
+```bash
+cp config/settings.local.yml.example config/settings.local.yml
+```
+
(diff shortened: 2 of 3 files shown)


docs/setup/native.md
```diff
@@ -74,16 +103,18 @@ All of the OSX instructions assume `homebrew` is your [package manager](https://
    - `brew install imagemagick`
 1. Install Poppler
    - `brew install poppler`
-1. Install ClamAV
+1. Install ClamAV or *use the fake scanner for development*
```
omgitsbillryan, line 106: I think we should default to using the fake clamav scanner. 99.9% of devs would prefer this. That means:

- use the fake one in `settings.yml`
- remove this "Install ClamAV" section altogether
  - maybe add some kind "optional steps" link to a separate markdown file on how to install Clam AV natively
- update the dev, stag, prod settings.yml files to override the fake clamav

bin/fake_clamdscan
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+echo "Fake scanning file with fake_clamdscan script"
+exit 0
```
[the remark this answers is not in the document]
LindseySaari: Agreed 💯

docs/setup/native.md
johnpaulashenfelter: settings.yml isn't the best place -- development.yml is for exactly this kind of thing (with the option of settings.local).

Since we *also* hardcode the setting in the `docker-compose` file, i think untangling _there_ makes more sense and already on my agenda :)

PR #4523, opened 2020-07-13
Add option to run ClamAV from `make up`

johnpaulashenfelter:
## Description of change

Provide a way to run clamav in the Docker container with the Makefile.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#11176

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,12 @@ else
     ENV_ARG	 := dev
 endif
 
+ifdef clam
+	FOREMAN_ARG := all=1
+else
+	FOREMAN_ARG := all=1,clamd=0,freshclam=0
+endif
+
 COMPOSE_DEV  := docker-compose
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -7,6 +7,12 @@ else
     ENV_ARG	 := dev
 endif
 
+ifdef clam
+	FOREMAN_ARG := all=1
```
LindseySaari, line 11: Just want to make sure that clamd & freshclam don't need to be set to 1

Makefile
```diff
@@ -7,6 +7,12 @@ else
...
+else
+	FOREMAN_ARG := all=1,clamd=0,freshclam=0
```
LindseySaari, line 13: Is this the same as web=1, cron=1, clamd=0, freshclam=0 ?

johnpaulashenfelter: Yep -- its just the config we had before (and similar to production)

PR #4526, opened 2020-07-14
don't reset session cookies for unauthenticated users

ericbuckley:
If an unauthenticated user makes an API call to any controller that uses
the authentication and sso concern, the user's session cookie will
reset.  This is problematic as some forms can be submitted by
unauthenticated users, thus having their CSRF cookie reset early will
prevent the form submission from succeeding.

### DIFF
--- a/app/controllers/concerns/authentication_and_sso_concerns.rb
+++ b/app/controllers/concerns/authentication_and_sso_concerns.rb
@@ -26,7 +26,7 @@ module AuthenticationAndSSOConcerns
     load_user
 
     if @session_object.nil?
-      Rails.logger.info('SSO: INVALID SESSION', sso_logging_info)
+      Rails.logger.debug('SSO: INVALID SESSION', sso_logging_info)
       clear_session
       return false
     end
@@ -48,7 +48,7 @@ module AuthenticationAndSSOConcerns
 
   # Destroys the users session in 1) Redis and the MHV SSO Cookie
   def clear_session
-    Rails.logger.info('SSO: ApplicationController#clear_session', sso_logging_info)
+    Rails.logger.debug('SSO: ApplicationController#clear_session', sso_logging_info)
 
     cookies.delete(Settings.sso.cookie_name, domain: Settings.sso.cookie_domain)
     @session_object&.destroy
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/authentication_and_sso_concerns.rb
```diff
@@ -45,15 +46,22 @@ def load_user
     @current_user = User.find(@session_object.uuid) if @session_object
   end
 
-  # Destroys the users session in 1) Redis, 1) the MHV SSO Cookie, 3) and the Session Cookie
-  def reset_session
-    Rails.logger.info('SSO: ApplicationController#reset_session', sso_logging_info)
+  # Destroys the users session in 1) Redis and the MHV SSO Cookie
+  def clear_session
+    Rails.logger.info('SSO: ApplicationController#clear_session', sso_logging_info)
```
jholton, line 51: Same concern as above about this being "noisy".  For `reset_session` the `sso_logging_info` still gets logged (and in fact gets logged twice as it currently is since `reset_session` calls `clear_session`)

app/controllers/concerns/authentication_and_sso_concerns.rb
```diff
@@ -26,7 +27,7 @@ def validate_session
 
     if @session_object.nil?
       Rails.logger.info('SSO: INVALID SESSION', sso_logging_info)
```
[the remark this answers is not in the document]
ericbuckley: want to make it a `debug` statement?

app/controllers/concerns/authentication_and_sso_concerns.rb
ericbuckley: Shall I just delete that logging statement?

app/controllers/concerns/authentication_and_sso_concerns.rb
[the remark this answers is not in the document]
jholton: I think I like the change to `debug` 👍

spec/controllers/application_controller_spec.rb
```diff
@@ -392,7 +392,7 @@ def append_info_to_payload(payload)
         it 'renders json error' do
           get :test_authentication
           expect(controller.instance_variable_get(:@session_object)).to be_nil
-          expect(session).to be_empty
+          expect(session).to_not be_empty
```
patrickvinograd, line 395:
```suggestion
          expect(session).not_to be_empty
```

PR #4530, opened 2020-07-14
Fix bgs healthcheck

charleystran:
## Description of change
This fixes the BGS healthcheck by using the correct renamed name for the BGS service


charleystran:
re-opening

PR #4532, opened 2020-07-15
Add StatsD events to Form1010cg Submissions

kevinmirc:
Fixes: https://github.com/department-of-veterans-affairs/va.gov-team/issues/5672

**Description**
Add StatsD increments for Form 10-10cg Submissions.

### DIFF
--- a/app/services/form1010cg/service.rb
+++ b/app/services/form1010cg/service.rb
@@ -9,18 +9,17 @@ module Form1010cg
     attr_accessor :claim, # SavedClaim::CaregiversAssistanceClaim
                   :submission # Form1010cg::Submission
 
-    STATSD_KEY_PREFIX = 'api.form1010cg'
+    STATSD_KEY_PREFIX = 'api.form1010cg.submission'
     NOT_FOUND         = 'NOT_FOUND'
 
     def self.metrics
-      prefix = STATSD_KEY_PREFIX + '.submission'
       OpenStruct.new(
-        attempt: prefix + '.attempt',
-        success: prefix + '.success',
+        attempt: STATSD_KEY_PREFIX + '.attempt',
+        success: STATSD_KEY_PREFIX + '.success',
         failure: OpenStruct.new(
           client: OpenStruct.new(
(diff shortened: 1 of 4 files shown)


app/services/form1010cg/service.rb
```diff
@@ -11,6 +11,21 @@ class InvalidVeteranStatus < StandardError
 
     NOT_FOUND = 'NOT_FOUND'
 
+    def self.metrics
+      prefix = 'api.form1010cg'
```
LindseySaari, line 15: probably a nitpick, but most of the other service classes define a STATSD_KEY_PREFIX constant. STATSD_KEY_PREFIX= 'api.form1010cg'

LindseySaari: ^^ Just to keep this uniform across the codebase

kevinmirc: Good call! I've updated this.

app/services/form1010cg/service.rb
```diff
@@ -9,7 +9,22 @@ class InvalidVeteranStatus < StandardError
     attr_accessor :claim, # SavedClaim::CaregiversAssistanceClaim
                   :submission # Form1010cg::Submission
 
-    NOT_FOUND = 'NOT_FOUND'
+    STATSD_KEY_PREFIX = 'api.form1010cg'
+    NOT_FOUND         = 'NOT_FOUND'
+
+    def self.metrics
+      prefix = STATSD_KEY_PREFIX + '.submission'
```
LindseySaari, line 16: Since submission is always appended, does it make sense to move it up to the STATSD_KEY_PREFIX constant or are you planning on adding additional metrics in the future e.g. api.form1010cg.something_else?

kevinmirc: There may be a status check in the future, but I can move it to STATSD_KEY_PREFIX and extract it when/if there are more metric types.

kevinmirc:
Thanks, @LindseySaari! The StatsD docs say that metrics are retained for only 15 days by default, how do I go about extending this retention period?

PR #4537, opened 2020-07-16
use the new drupal cms gql api - API-1454

charleystran:
## Description of change
This PR changes out the VA Form scraper to use the new drupal graphql endpoint to pull forms data


modules/va_forms/app/workers/va_forms/form_reloader.rb
```diff
@@ -6,82 +6,66 @@ module VaForms
...
-      form.last_revision_on = parse_date(line.css('td:nth-child(4)').text) if revision_string.present?
-      form.pages = line.css('td:nth-child(5)').text
-      form_url = url.starts_with?('http') ? url.gsub('http:', 'https:') : get_full_url(url)
-      form.url = Addressable::URI.parse(form_url).normalize.to_s
-      form = update_sha256(form)
-      form.save if current_sha256 != form.sha256
+    def build_and_save_form(form)
+      va_form = VaForms::Form.find_or_initialize_by form_name: form['fieldVaFormName']
+      current_sha256 = va_form.sha256
+      va_form.form_name = form['fieldVaFormName']
```
edmangimelli, line 52:
```suggestion
```

modules/va_forms/app/workers/va_forms/form_reloader.rb
```diff
@@ -107,7 +91,7 @@ def update_sha256(form)
     end
 
     def get_full_url(url)
-      "#{BASE_URL}/vaforms/#{url.gsub('./', '')}" if url.starts_with?('./va') || url.starts_with?('./medical')
+      "#{FORM_BASE_URL}/vaforms/#{url.gsub('./', '')}" if url.starts_with?('./va') || url.starts_with?('./medical')
```
edmangimelli, line 97:
```suggestion
      raise ArgumentError, 'url must start with ./va or ./medical' unless url.starts_with?('./va') || url.starts_with?('./medical')

      "#{FORM_BASE_URL}/vaforms/#{url.gsub('./', '')}"
```
if this method returns a nil (which could happen as it's currently written: `x if y`), the `Addressable::URI.parse` line will raise an exception. should we just raise the exception here so we know what the problem was?

modules/va_forms/app/workers/va_forms/form_reloader.rb
charleystran: added this in a different commit

modules/va_forms/app/workers/va_forms/form_reloader.rb
charleystran: I dont think so, the intent here is to make sure the method is used properly for these url patterns. moving the exception to the URI parse may loose the intent during future work of the class. Perhaps renaming it to something like expand_va_url would be more clear of intent

modules/va_forms/app/workers/va_forms/form_reloader.rb
```diff
@@ -6,82 +6,69 @@ module VaForms
...
+      query = File.read(Rails.root.join('modules', 'va_forms', 'config', 'graphql_query.txt'))
+      body = { query: query }
+      response = connection.post('graphql', body.to_json)
+      forms_data = JSON.parse(response.body)
+      processed_forms = []
+      forms_data.dig('data', 'nodeQuery', 'entities').each do |form|
+        va_form = build_and_save_form(form)
+        processed_forms << va_form
+      rescue
+        next
```
edmangimelli, line 21: do we wanna do any kinda logging here? i'm worried that they may send us something formatted in a way we don't expect, and build_and_save throw starts throwing exceptions that are silently ignored. if build_and_save failed for all of them, we'd notice, but if it was just a 1/4 or a 1/3 we might not (exs: uri normalization fails, date parsing fails)

modules/va_forms/app/workers/va_forms/form_reloader.rb
edmangimelli: >moving the exception to the URI parse
currently the exception is at the URI parse

modules/va_forms/app/workers/va_forms/form_reloader.rb
charleystran: we could log to Sentry maybe. Since its a background job no one is actively looking at it. The goal of next was to not allow one bad form the break the loop

edmangimelli: >The goal of next was to not allow one bad form the break the loop
for sure. I'd be down for just sending the exception to sentry, and, if we see a bunch, take a peek and see if we need to change something up b/c the data coming in is goofier than we thought

charleystran: added

modules/va_forms/app/workers/va_forms/form_reloader.rb
```diff
@@ -5,83 +5,76 @@
...
+      body = { query: query }
+      response = connection.post('graphql', body.to_json)
+      forms_data = JSON.parse(response.body)
+      processed_forms = []
+      forms_data.dig('data', 'nodeQuery', 'entities').each do |form|
+        va_form = build_and_save_form(form)
+        processed_forms << va_form
+      rescue => e
+        log_message_to_sentry(
+          "#{form['fieldVaFormNumber']} faild to import into forms forms database",
```
edmangimelli, line 23:
```suggestion
          "#{form['fieldVaFormNumber']} failed to import into forms database",
```

PR #4545, opened 2020-07-17
Enable SSOe for localhost development

patrickvinograd:
## Description of change
Enables SSOe login in localhost environments using an agreed-upon SAML configuration. Requires no SP cert/key, so does not use SAML encryption or signing. 

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/7092

Adds settings.yml defaults under saml_ssoe stanza. Corresponding devops PR overrides these for dev/staging/prod/review, where signing/encryption are expected/required:
https://github.com/department-of-veterans-affairs/devops/pull/7289

Will test against localhost once corresponding SSOe configuration is enabled.


LindseySaari:
Are there any specs that can be added to verify these changes?

patrickvinograd:
Added specs around the config loading.
*Note that the devops PR should be merged first:* 
department-of-veterans-affairs/devops#7289

patrickvinograd:
Devops PR is merged so this is safe to merge once approved.

spec/lib/saml/ssoe_settings_service_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      end
+    end
+
+    context 'with encryption configured' do
+      before do
+        Settings.saml_ssoe.certificate = 'foobar'
+        Settings.saml_ssoe.request_signing = false
+        Settings.saml_ssoe.response_signing = false
+        Settings.saml_ssoe.response_encryption = true
+      end
```
omgitsbillryan, line 51: I think this can leave `Settings` in a random & undesirable state which may cause failures in other specs which, for better or worse, rely on the `Settings` being a certain way.

Might want to use [`with_settings`](https://github.com/department-of-veterans-affairs/vets-api/blob/master/spec/rails_helper.rb#L37), or just be sure to save the state of `Settings` before the spec runs and return the Settings to that state after the spec has completed.

config/ssoe_idp_int_metadata_isam.xml
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
omgitsbillryan, line 1: Not related to this PR, just for my own learning / posterity - why are we providing a hardcoded metada file when ruby-saml [generates one for us](https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/controllers/saml_controller.rb)?

patrickvinograd: This is the IDP metadata.

spec/lib/saml/ssoe_settings_service_spec.rb
patrickvinograd: Good call, I'll switch to with_settings

PR #4546, opened 2020-07-18
686c-674 Form processing module

digitaldrk:
### Regarding the size of this PR
Apologies for the size of this PR. This is part of a large feature and was sliced off of a larger branch my team has/had going. After consulting with a member of VSP, it was suggested that we ship this `/lib` code and then work our way towards the FE with more PR's.

### What this PR accomplishes
This processes the [686c-674](https://www.vba.va.gov/pubs/forms/VBA-21-686c-ARE.pdf) form submission allowing Veterans to add dependents to their benefits. To add some context about the form, it allows users to add children, report a death, report a divorce, add a student, add a child over 18 attending school, marriages and various derivations of those events and more. Given the scope of this form, this code could end up processing 100's of data attributes taking up to 2 minutes to complete. Ultimately, the code in this PR will run within a job.

### BGS
This PR uses calls we have added to the [bgs-ext gem](https://github.com/department-of-veterans-affairs/bgs-ext/pull/19) to interact with the services we need within BGS.

**Auth** is accomplished using settings in [settings.yml](https://github.com/department-of-veterans-affairs/vets-api/blob/master/config/settings.yml#L506). We send along auth params in each request.

### Flowchart of calls

- Green ovals are the clases/methods in this PR
- Blue diamonds are calls made to BGS

686c-674_lib_submission_process

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/11260

    - Settings should be already configured per environment we are already making use of BGS calls in production
     - This PR's code is not accessible to the user, more PR's to come will open that up. We will be feature flagging when that happens
    - We have added Sentry logging. Each call will retry 3 times sending messages to Sentry about each attempt. If all attempts fail, an error is sent to Sentry. Both messages and errors include a tag for our team and user `icn`
    - No Prometheus/Grafana metrics added
     - Not applicable to this PR
    - None at this time


lib/bgs/dependents.rb
```diff
@@ -0,0 +1,229 @@
...
+
+      @dependents << serialize_dependent_result(
+        participant,
+        'Child',
+        'Other',
+        { 'type': '674' }
+      )
+    end
+
+    def format_674_info
```
LindseySaari, line 159: Could these formatting methods be pulled out? It may clean up this service class a bit

digitaldrk: Sure thing. Did you have an implementation in mind? Maybe a helper/formatter module?

lib/bgs/service.rb
```diff
@@ -0,0 +1,197 @@
...
+        address['military_postal_code'] = address.delete('state_code')
+        address['military_post_office_type_code'] = address.delete('city')
+      end
+
+      create_address(@proc_id, participant_id, address)
+    end
+
+    def notify_of_service_exception(error, method, attempt = nil, status = :error)
+      msg = "Unable to #{method}: #{error.message}: try #{attempt} of #{MAX_ATTEMPTS}"
+      context = { icn: @user[:icn] }
```
LindseySaari, line 171: Is this a PII concern?

lib/bgs/student_school.rb
```diff
@@ -0,0 +1,142 @@
...
+      with_multiple_attempts_enabled do
+        service.vnp_child_student.child_student_create(
+          child_student_params
+        )
+      end
+    end
+
+    # rubocop:disable Metrics/AbcSize
+    # rubocop:disable Metrics/MethodLength
+    def child_school_params
```
LindseySaari, line 38: To me it seems like the parameter mapping doesn't belong in the service class. What about creating a model/models to clean this up a bit?

lib/bgs/service.rb
digitaldrk: Good eye. My understanding is no. We have implemented this and shipped elsewhere. I asked this question a while back here are resources I was provided. LMK if you still feel it's a concern.

https://github.com/department-of-veterans-affairs/vets-api/pull/2888

https://dsva.slack.com/archives/CBU0KDSB1/p1591299574037900?thread_ts=1591294003.033000&cid=CBU0KDSB1

LindseySaari: Ok awesome, just caught up on the above thread. Looks like icn doesn't cause any PII concern and is useful for debugging

LindseySaari:
Another thought: You could break this PR down even further into smaller chunks for each new service class introduced. Just a suggestion, but open to thoughts and opinions. It may be conceptually easier to review smaller PRs.

lib/bgs/student_school.rb
digitaldrk: @LindseySaari I wasn't _quite_ sure if this is what you were suggesting but I didn't want to come back empty handed. What do you think about this?

https://github.com/department-of-veterans-affairs/vets-api/pull/4546/files#diff-44bb8d03cb5942b139547c36a9525c68

lib/bgs/dependents.rb
digitaldrk: Great suggestion. I like this much better, thank you.

digitaldrk:
> Another thought: You could break this PR down even further into smaller chunks for each new service class introduced. Just a suggestion, but open to thoughts and opinions. It may be conceptually easier to review smaller PRs.

I hear you, thank you for diving into this thing.

Where do you think we are in the review process? My concern is that we would add to the timeline. That we would push code that would get reviewed and then we would have to rework a lot of code we had waiting based on that review, (hope I'm making sense).

omgitsbillryan:
+1 to smaller PR's. Could we break the PR up by service + models necessary for that service?

From what I've read so far, the code looks clean, but logistically I keep having leave and come back to this PR and I forget where I left off 🤯 . I'll come back to this later this evening, but any work you can do to break this up would be a big help.

digitaldrk:
> +1 to smaller PR's. Could we break the PR up by service + models necessary for that service?
> 
> From what I've read so far, the code looks clean, but logistically I keep having leave and come back to this PR and I forget where I left off 🤯 . I'll come back to this later this evening, but any work you can do to break this up would be a big help.

@omgitsbillryan 

Understood. I'll look into how I can break this up. Thank you for looking into it. I will be available later this evening and be available in any capacity that would be of help.

lib/bgs/benefit_claim.rb
```diff
@@ -0,0 +1,65 @@
...
+      end
+    end
+
+    # rubocop:disable Metrics/MethodLength
+    def benefit_claim_params
+      {
+        file_number: @veteran[:file_number],
+        ssn: @user[:ssn],
+        ptcpnt_id_claimant: @user[:participant_id],
+        benefit_claim_type: '1',
```
omgitsbillryan, line 41: This might read nicer if we created a frozen var for the params that are just constants and then merge with the dynamic params. e.g.

```ruby
BENEFIT_CLAIM_PARAM_CONSTANTS = {
  benefit_claim_type: '1',
  payee: '00',
  # other constants
}.freeze

def benefit_claim_params
  {
    file_number: @veteran[:file_number],
    # other dynamic params
  }.merge(BENEFIT_CLAIM_PARAM_CONSTANTS)
end
```

lib/bgs/dependents.rb
```diff
@@ -0,0 +1,156 @@
...
+      @dependents
+    end
+
+    # rubocop:enable Metrics/CyclomaticComplexity
+    # rubocop:enable Metrics/PerceivedComplexity
+
+    private
+
+    def add_children
+      @dependents_application['children_to_add'].each do |child_info|
```
omgitsbillryan, line 36: Is it at all possible for `@dependents_application['children_to_add']` to be `nil`? If so, is there any way to safely recover?

lib/bgs/service.rb
```diff
@@ -0,0 +1,197 @@
...
+        )
+      end
+    end
+
+    def service
+      @service ||= BGS::Services.new(
+        external_uid: @user[:icn],
+        external_key: @user[:external_key]
+      )
+    end
```
omgitsbillryan, line 61: Would it be possible to abstract this to a separate class that the other services (marriage, form686c, etc) don't inherit from? It would make it more obvious that these are external service calls.

```ruby
bgs_service.create_person
```

It wasn't obvious to me at first glance that all of these external calls were being made.

omgitsbillryan: It would also cut down the size of this class.

lib/bgs/form686c.rb
```diff
@@ -0,0 +1,93 @@
...
+
+module BGS
+  class Form686c < Service
+    def initialize(user)
+      @user = user.with_indifferent_access
+    end
+
+    def submit(payload)
+      proc_id = create_proc_id_and_form
+      veteran = VnpVeteran.new(proc_id: proc_id, payload: payload, user: @user).create
```
omgitsbillryan, line 11: I'm noticing a number of service calls, like this one, that chain together multiple external service calls. For example here, `create_proc_id_and_form` itself, calls 2 other external services, which then also might go on and call even more external services. 

I think it'd be clearer if the `service` classes were "dumb" and better adhered to the single-responsibility principle. In this case, submit would require that some kind of a `VnpVeteranCreateResponse` object be passed in.

With dumb services, the chaining logic could be better encapsulated in one place and not scattered through the service classes themselves.

lib/bgs/form686c.rb
```diff
@@ -0,0 +1,93 @@
...
+
+      all_dependents = dependents + marriages
+
+      VnpRelationships.new(proc_id: proc_id, veteran: veteran, dependents: all_dependents, user: @user).create
+      process_674(proc_id, dependents, payload)
+    end
+
+    def process_674(proc_id, dependents, payload)
+      dependents.each do |dependent|
+        if dependent[:type] == '674'
```
omgitsbillryan, line 75: I have no idea what a dependent of type 674 means. Can we encapsulate this into a named boolean function? e.g.

```ruby
def is_foo_bar_denendent? #
  dependent[:type] == '674'
end
```

lib/bgs/benefit_claim.rb
```diff
@@ -0,0 +1,65 @@
...
+        payee: '00',
+        end_product: @veteran[:benefit_claim_type_end_product],
+        end_product_code: '130DPNEBNADJ',
```
omgitsbillryan, line 44: Same comment here about using a frozen hash constant then merging the dynamic params with it.

lib/bgs/dependents.rb
```diff
@@ -0,0 +1,156 @@
+# frozen_string_literal: true
+
+module BGS
+  class Dependents < Service
+    include Helpers::Formatters
+
+    def initialize(proc_id:, payload:, user:)
+      @proc_id = proc_id
+      @payload = payload
```
omgitsbillryan, line 9: What form is `payload` supposed to take here? Maybe it could be a model that we could validate before proceeding? Because it looks like this service expects this object to contain a lot of things.

lib/bgs/helpers/formatters.rb
```diff
@@ -0,0 +1,114 @@
+# frozen_string_literal: true
```
omgitsbillryan, line 1: "Helper" usually means a code smell. All of the logic in here looks like serialization logic. I think if we had some better encapsulation around the BGS response object models, we might then also be able to encapsulate most/all of this serialization logic inside either the model or a dedicated serializer for the model.

omgitsbillryan:
Are any of your services hooked up to "Breakers"? It's an open-circuit design pattern all of our external services adhere to which helps `vets-api` protect itself against external services going down and hogging threads + causing response times to spike.

- [breakers initialization](https://github.com/department-of-veterans-affairs/vets-api/blob/master/config/initializers/breakers.rb#L35-L65)
- [breakers grafana dashboard](http://grafana.vfs.va.gov/d/000000022/breakers?orgId=1)
- [an example of hooking a service up to breakers](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/evss/documents_service.rb#L26-L28)

...Or does the BGS services gem handle this kind of logic?

digitaldrk:
> Are any of your services hooked up to "Breakers"? It's an open-circuit design pattern all of our external services adhere to which helps `vets-api` protect itself against external services going down and hogging threads + causing response times to spike.
> 
> * [breakers initialization](https://github.com/department-of-veterans-affairs/vets-api/blob/master/config/initializers/breakers.rb#L35-L65)
> * [breakers grafana dashboard](http://grafana.vfs.va.gov/d/000000022/breakers?orgId=1)
> * [an example of hooking a service up to breakers](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/evss/documents_service.rb#L26-L28)
> 
> ...Or does the BGS services gem handle this kind of logic?

No Idea on breakers. Haven't heard of that. I will look into it, thank you for the resource.

lib/bgs/service.rb
```diff
@@ -0,0 +1,197 @@
...
+      )
+
+      raise exception
+    end
+
+    def format_date(date)
+      return nil if date.nil?
+
+      Date.parse(date).to_time.iso8601
+    end
```
omgitsbillryan, line 195: Probably nitpicky, but this is duplicate code 

https://github.com/department-of-veterans-affairs/vets-api/pull/4546/files#diff-39a641783cd9a07c7cf96358929c6680.

Maybe create a `lib/bgs/date_formatter.rb` mixin?

omgitsbillryan:
All of my other comments are mostly superficial. My only real worry is WRT how long each of these service calls can potentially take and not being protected should the external service go down.

This PR looks to just create the services and not yet use them, so maybe this comment jumps the gun... just be mindful when using these services to make them asynchronous using sidekiq if/when possible.

digitaldrk:
> All of my other comments are mostly superficial. My only real worry is WRT how long each of these service calls can potentially take and not being protected should the external service go down.
> 
> This PR looks to just create the services and not yet use them, so maybe this comment jumps the gun... just be mindful when using these services to make them asynchronous using sidekiq if/when possible.

Sage advice @omgitsbillryan FYI a 'large' form submission takes up to 2 min(!) we have code coming soon that will indeed run this in a job.

Thank you for your time and the great feedback.

digitaldrk:
Working on breaking this PR down.

lib/bgs/dependents.rb
digitaldrk: `payload` is a massive hash. The form allows users to add unlimited dependents in some workflows.

charleystran:
@omgitsbillryan  following up on your comments about breakers. The BGS gem (which spun out of the Caseflow BGS gem) does not use faraday so it cant use breakers. Its all wrapped around Savon as a SOAP client.

lib/bgs/form686c.rb
digitaldrk: I may not be understanding your comment completely so apologies if I'm way off.

The idea I had was to try to bundle some of the calls into a kind of "Bounded Context". Where we have for example, a representation of a veteran in BGS with `VnpVeteran` after a culmination of calls. `Vnp` prefix is probably mis-leading 🤔 

It's my humble opinion that unfortunately, we are shouldering logic we shouldn't have to; many of these calls could/should be lumped together on the service side. So there is indeed heavy chaining going on

omgitsbillryan:
Looking at the smaller broken up PR's is it ok to close this one?

I like that all the smaller PR's are linked to this one so we don't lose that flow chart 😄 .

PR #4550, opened 2020-07-20
Correctly configure breakers for EVSS Disability Compensation

annaswims:
## Description of change

I'm configuring breakers for EVSS Disability Compensation  calls for breakers because it seems like it was overlooked and I'd like the metrics on the [backend service report](http://grafana.vfs.va.gov/d/000000050/backend-service-report?orgId=1) dashboard.  

## Original issue(s)
department-of-veterans-affairs/va.gov-team#8387

I haven't yet stopped to think about how breakers errors will work with our janky error handling/retries.  Will do that before merging


spec/jobs/evss/disability_compensation_form/submit_form526_all_claim_spec.rb
```diff
@@ -78,6 +78,17 @@
       end
     end
 
+    context 'with a breakers outage' do
+      it 'runs the retryable_error_handler and raises a ' do
```
edmangimelli, line 82: raises wha??

spec/lib/evss/disability_compensation_form/service_spec.rb
```diff
@@ -87,5 +91,22 @@
         expect { subject.submit_form526(valid_form_content) }.to raise_error(Common::Exceptions::GatewayTimeout)
       end
     end
+
+    context 'with an breakers error' do
```
edmangimelli, line 95:
```suggestion
    context 'with a breakers error' do
```

edmangimelli:
the failing test is pretty baffling
```
rspec ./modules/claims_api/spec/requests/v1/disability_compensation_request_spec.rb:211 # Disability Claims  #526 form 526 validation Breakers outages are recorded (investigating) is logged to PersonalInformationLog
```
why wouldn't v0 be failing too?
I wonder if this because these tests are being run in parallel? (the forced outage is being ended early because another test (v0 in this case) ended it)

PR #4551, opened 2020-07-20
Fix EVSS::DisabilityCompensationForm retry handling

annaswims:
## Description of change
Something changed on the [April 28th deploy of evss ](https://dsva.slack.com/archives/C8R3JS8BU/p1588171812160400)
Screen Shot 2020-04-29 at 10 43 56 AM

And after that point errors were no longer categorized as retriable.  I think that the issue is that EVSS started returning a http status of 200 where they previously returned a status of 500.  I think I may have been partially to blame for this because I previously mentioned to them that I thought 500 was a strange status for "EVSS is working as expected but an upstream server is having issues" or "there is a data problem".  

We were rescuing `EVSS::DisabilityCompensationForm::ServiceException` (thrown when there was a 500) but we're  getting a `EVSS::ErrorMiddleware::EVSSError` because we had a 200 response.

After this PR ` service.submit_form526(form_data)` will throw a `EVSS::DisabilityCompensationForm::ServiceException` for 500 responses or 200 responses with error bodies.

@edmangimelli @bastosmichael I suspect this is affecting Lighthouse as well.

 
## Original issue(s)


app/workers/evss/disability_compensation_form/submit_form526.rb
```diff
@@ -62,17 +63,12 @@ def service(_auth_headers)
       # @param error [EVSS::DisabilityCompensationForm::ServiceException]
       #
       def retry_form526_error_handler!(error)
-        if (error.key == 'evss.external_service_unavailable' && ep_code_valid?(error)) ||
```
annaswims, line 65: moved the logic to service_exception.rb

annaswims:
@department-of-veterans-affairs/va-api-engineers

The goal of this PR is that regardless of whether EVSS sends a http 200 or 500 response if EVSS returns a body with an errors hash we raise a `EVSS::DisabilityCompensationForm::ServiceException`.  However, I see that 

`service(auth_headers).submit_form526(form_data`) was rescuing from [::EVSS::DisabilityCompensationForm::ServiceException](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/app/workers/claims_api/claim_establisher.rb#L20)
while `service.validate_form526(auto_claim.to_internal)`  was rescuing from [ EVSS::ErrorMiddleware::EVSSError](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/app/controllers/claims_api/base_disability_compensation_controller.rb#L27)

does [this commit](https://github.com/department-of-veterans-affairs/vets-api/pull/4551/commits/98720423ebf0c84f71c5547181154e6f85fbb19f) that rescues from either seem ok to you?

lib/evss/disability_compensation_form/service_exception.rb
```diff
@@ -35,6 +35,14 @@ class ServiceException < EVSS::ServiceException
         PIFInUse: 'evss.disability_compensation_form.pif_in_use'
       }.freeze
 
+      # Retry if any upstream external service unavailability exceptions (unless it is caused by an invalid EP code)
+      # and any PIF-in-use exceptions are encountered.
+
+      def retryable?
+        (@key == 'evss.external_service_unavailable' && only_has_retriable_message_texts?) ||
+          (@key == 'evss.disability_compensation_form.pif_in_use')
```
omgitsbillryan, line 43:
```suggestion
        (@key == ERROR_MAP[:serviceError] && only_has_retriable_message_texts?) ||
          (@key == ERROR_MAP[: PIFInUse])
```

omgitsbillryan: OTOH, it's odd that `ERROR_MAP` contains multiple keys with the same value - `serviceError` & `ServiceException`. Not sure which is appropriate to use.

lib/evss/disability_compensation_form/service.rb
```diff
@@ -66,8 +66,12 @@ def validate_form526(form_content)
       private
 
       def handle_error(error)
-        if error.is_a?(Common::Client::Errors::ClientError) && error.status != 403 && error.body.is_a?(Hash)
-          save_error_details(error)
+        # Common::Client::Errors::ClientError is raised from Common::Client::Base#request after it rescues
+        # Faraday::ClientError.  EVSS::ErrorMiddleware::EVSSError is raised from EVSS::ErrorMiddleware when
+        # there is a 200-response with an error message in the body
+        if ((error.is_a?(Common::Client::Errors::ClientError) && error.status != 403) ||
+           error.is_a?(EVSS::ErrorMiddleware::EVSSError)) && error.body.is_a?(Hash)
```
omgitsbillryan, line 73: 😱 😱 😱 

🤯🤯🤯

annaswims: Your emojis make a great point and I don't disagree.  The whole point of this PR is "make this behave just like it did in April before EVSS made a change to their responses".  I see that many people have tried to refactor EVSS error handling over the years and each one makes it just a tad more complicated. It seemed like my two choices were "mega refactor, again" or "add another condition".  I'd happily take suggestions for another way.  It's hard though, because lighthouse also uses this code.

Also for context, EVSS' sustainment contract ends this week and I'm not aware of what will happen next.

omgitsbillryan: Sorry for the lack of context, I'm in full agreement that, "add another condition" is the way to go here.

I'm pretty sure I once took a crack at trying to undo this. After multiple weeks, I eventually gave up on a huge PR.

edmangimelli:
> @department-of-veterans-affairs/va-api-engineers
> 
> The goal of this PR is that regardless of whether EVSS sends a http 200 or 500 response if EVSS returns a body with an errors hash we raise a `EVSS::DisabilityCompensationForm::ServiceException`. However, I see that
> 
> `service(auth_headers).submit_form526(form_data`) was rescuing from [::EVSS::DisabilityCompensationForm::ServiceException](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/app/workers/claims_api/claim_establisher.rb#L20)
> while `service.validate_form526(auto_claim.to_internal)` was rescuing from [ EVSS::ErrorMiddleware::EVSSError](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/app/controllers/claims_api/base_disability_compensation_controller.rb#L27)
> 
> does [this commit](https://github.com/department-of-veterans-affairs/vets-api/pull/4551/commits/98720423ebf0c84f71c5547181154e6f85fbb19f) that rescues from either seem ok to you?

without knowing this code too well, this seems pretty reasonable.
Do you think we change the line in `track_526_validation_errors `
```
        key = error['key'].gsub(/\[(.*?)\]/, '')
```
to
```
        key = error['key']&.gsub(/\[(.*?)\]/, '')
```
i don't know how suspicious we should be of what's returned to us (but maybe we want this to fail so we know when they've changed things up on us)

PR #4553, opened 2020-07-20
[11363] Camelize List Debt Letters Response

camerontesterman:
## Description of change
Camelizes the list debt letters response that is sent to the front end.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/11363

### DIFF
--- a/lib/debts/letter_downloader.rb
+++ b/lib/debts/letter_downloader.rb
@@ -35,8 +35,9 @@ module Debts
       end
 
       debts_records.map do |debts_record|
-        debts_record.marshal_dump.slice(
-          :document_id, :doc_type, :type_description, :received_at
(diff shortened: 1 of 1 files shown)


lib/debts/letter_downloader.rb
```diff
@@ -35,8 +35,9 @@ def list_letters
       end
 
       debts_records.map do |debts_record|
-        debts_record.marshal_dump.slice(
-          :document_id, :doc_type, :type_description, :received_at
+        record = debts_record.marshal_dump.deep_transform_keys { |key| key.to_s.camelize(:lower) }
```
LindseySaari, line 38: Could you use a serializer instead? Then I think olive branch will take care of the camel casing if the X-Key-Inflection header is set to camel.

PR #4556, opened 2020-07-20
notification-api#71/email-parameters

marisahoenig:
## Description of change
Upon submitting 526EZ form using vets web pages the user should receive an email to confirm submission.  This change will add the full name, submit date, and claim id as personalization to the email.

## Original issue(s)
department-of-veterans-affairs/notification-api#71

   * **Yes**. The feature flag, form526_confirmation_email was added by a previous PR.
   * We obtain the veteran's full name, but this is not logged.

## Testing
We've tested our changes on the review instance, and we get the email with the information that we added upon form526 submission.


app/models/form526_submission.rb
```diff
@@ -121,15 +121,22 @@ def perform_ancillary_jobs
...
     if submission.form526_job_statuses.all?(&:success?)
-      submission.send_form526_confirmation_email if Flipper.enabled?(:form526_confirmation_email, @current_user)
+      submission.send_form526_confirmation_email if Flipper.enabled?(:form526_confirmation_email)
       submission.workflow_complete = true
       submission.save
     end
   end
 
   def send_form526_confirmation_email
+    user = User.find(user_uuid)
```
annaswims, line 131: There's a pretty solid chance we won't have the user in redis by this point.  The user is only persisted in redis when they have an active session (they are logged out after 30 minutes of inactivity or when they click log out).   In cases  where EVSS or and upstream partner is having an outage and we retry for a while we almost certainly won't have the user object in redis.

"#{auth_headers['va_eauth_firstName']} #{auth_headers['va_eauth_lastName']}" should get you what you want though.

saman-moshafi-va: We need first, middle and last names plus suffix as this is what is displayed on the web page, once a submission succeeds.  They are not all in auth_headers.  Modifying auth_headers also seems to be invasive as many other services and components seems to use them.

ffafara-tw: @annaswims As Saman mentioned we would prefer to have full name (with middle name and suffix). 
I see 3 options:
1. add middle name and suffix to auth headers -> I'm hesitant to modify auth headers that are going to be sent to EVSS. I don't want us to accidentally break EVSS integration
2. Propagate full name as part of options to perform_ancillary_jobs_handler and workflow_complete_handler -> Is it safe to pass full name that way from privacy perspective (Full name is PII)? Would the options be persisted in some sidekiq jobs db? Is there a way to encrypt full name inside options?
3. Add FullName as encrypted attribute to Form526Submission

Any recommendations / other ideas?

ffafara-tw: @gunsch @f1337 Your security input on option 2 above will be greatly appreciated! There is also a slack thread about it here: https://app.slack.com/client/T03FECE8V/CBU0KDSB1/thread/CBU0KDSB1-1596043470.273400

philherbert: Hello @annaswims! We have implemented option 2 in @ffafara-tw 's comment above - getting the user and full name synchronously on submission, and passing them down to the sidekiq jobs. Could you please take a look at 2ba4ccf? Thanks!

app/models/form526_submission.rb
```diff
@@ -60,6 +61,11 @@ def start
     jids.first
   end
 
+  def get_full_name
+    user = User.find(user_uuid)
```
annaswims, line 65: There's still a fairly high chance that get_full_name will return nil. `User.find` will only return a value if the user has an active session

philherbert: Ah, I see! Is it possible to submit the form without having an active session?

PR #4558, opened 2020-07-21
Claims API - Remove Unused Method from BaseDisabilityCompensationController

edmangimelli:
the `service` method in app/controllers/claims_api/base_disability_compensation_controller.rb is not being used, as far as I can tell (checked parents and children)

### DIFF
--- a/modules/claims_api/app/controllers/claims_api/base_disability_compensation_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/base_disability_compensation_controller.rb
@@ -62,17 +62,5 @@ module ClaimsApi
         StatsD.increment STATSD_VALIDATION_FAIL_TYPE_KEY, tags: ["key: #{key}"]
       end
     end
-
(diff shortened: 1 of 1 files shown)


modules/claims_api/app/controllers/claims_api/base_disability_compensation_controller.rb
```diff
@@ -62,17 +62,5 @@ def track_526_validation_errors(errors)
         StatsD.increment STATSD_VALIDATION_FAIL_TYPE_KEY, tags: ["key: #{key}"]
       end
     end
-
-    def service(auth_headers)
-      if Settings.claims_api.disability_claims_mock_override && !auth_headers['Mock-Override']
```
bastosmichael, line 67: Don't remove this, This method isn't being used in the base controller because it is meant to be used in the above controllers that inherit from base controller. If I'm wrong show me but don't just remove this method.

PR #4560, opened 2020-07-21
user loa3 and loa1 tests with a camel-inflected schema

erluti:
## Description of change
The goal is to have the `/v0/user` endpoint test the camel-inflected version of the response.
To do this a rake task was added to convert json schemas into a camelized version.  This task was used to create one for `user_loa3.json` and `user_loa1.json`.  The `SchemaMatchers` module was updated to test against the camelized files instead of the snake json we already have.  Then the new matcher was used to test`/v0/user` "with camel inflection".

## Original issue(s)
department-of-veterans-affairs/va.gov-team#10976


spec/request/user_request_spec.rb
```diff
@@ -6,6 +6,28 @@
 RSpec.describe 'Fetching user data', type: :request do
   include SchemaMatchers
 
+  context 'GET /v0/user in camel - when an LOA 3 user is logged in' do
```
erluti, line 9: copied from the context below, but changed to `in camel`

spec/support/schemas_camelized/user_loa3.json
```diff
@@ -0,0 +1 @@
...:{"type":["string",null]},"zip":{"type":["string",null]},"lastSignedIn":{"type":["string",null]},"loa":{"type":"object","required":["current","highest"],"properties":{"current":{"type":"integer"},"highest":{"type":"integer"}}},"multifactor":{"type":"boolean"},"verified":{"type":"boolean"},"signIn":{"type":"object","properties":{"serviceName":{"type":"string"}}},"authnContext":{"type":"string"}}},"vaProfile":{"type":["object","null"],"required":["status"],"oneOf":[{"properties":{"status":{"type":["string"]}}},{"properties":{"status":{"type":"string"},"birthDate":{"type":"string"},"familyName":{"type":"string"},"gender":{"type":"string"},"isCernerPatient":{"type":"boolean"},"facilities":{"type":"array"},"givenNames":{"type":"array"},"vaPatient":{"type":"boolean"},"mhvAccountState":{"type":"string"}}}]},"veteranStatus":{"type":["object","null"],"required":["status"],"oneOf":[{"properties":{"status":{"type":["string"]}}},{"properties":{"isVeteran":{"type":"boolean"},"servedInMilitary":{"type":"boolean"},"status":{"type":"string"}}}]},"vet360ContactInformation":{"type":["object","null"]}}}}},"meta":{"type":"object","required":["errors"],"properties":{"errors":{"type":["array",null]}}}}}
```
erluti, line 1: this file was generated by the above raketask

spec/request/user_request_spec.rb
erluti: adding the `headers: { 'X-Key-Inflection' => 'camel' }` was the only difference between the `before`-blocks

spec/request/user_request_spec.rb
```diff
@@ -6,6 +6,28 @@
...
+      mhv_account = double('MhvAccount', creatable?: false, upgradable?: false, account_state: 'upgraded')
+      allow(MhvAccount).to receive(:find_or_initialize_by).and_return(mhv_account)
+      allow(mhv_account).to receive(:user_uuid).and_return(mhv_user.uuid)
+      allow(mhv_account).to receive(:terms_and_conditions_accepted?).and_return(true)
+      allow(mhv_account).to receive(:needs_terms_acceptance?).and_return(false)
+      allow(mhv_account).to receive(:user=).and_return(mhv_user)
+      create(:account, idme_uuid: mhv_user.uuid)
+      sign_in_as(mhv_user)
+      get v0_user_url, params: nil, headers: { 'X-Key-Inflection' => 'camel' }
+    end
```
f1337, line 23: Could we refactor this and the nearly identical block below into a common context + before block, using a variable for the key-inflection header?

rakelib/camelize_file.rake
```diff
@@ -0,0 +1,55 @@
...
+
+class SchemaCamelizer
+  attr_reader :original_path
+
+  def initialize(schema_path)
+    @original_path = schema_path
+    raw_schema = File.read(schema_path)
+    raw_schema.gsub!(/"required": \[(["\w*"\,? ?\n?]*)\]/) do
+      # rubocop:disable Style/PerlBackrefs
+      keys = $1.split(',').map(&:strip).map { |key| camelizer.call(key.gsub('"', '')) }
```
LindseySaari, line 26: What is this block doing?

rakelib/camelize_file.rake
```diff
@@ -0,0 +1,55 @@
...
+  def camel_path
+    @camel_path ||= original_path.gsub('schemas', 'schemas_camelized')
+  end
+
+  def save!
+    raise 'expected spec/support/schemas to be original path!' if original_path == camel_path
+
+    File.open(camel_path, 'w') { |file| file.write(camel_schema.to_json) }
+  end
+end
```
f1337, line 55: Could we break this into two files, and put the supporting ruby class under `support/`?

rakelib/camelize_file.rake
```diff
@@ -0,0 +1,55 @@
...
+    schema_path = Rails.root.join('spec', 'support', 'schemas', "#{args[:json_schema_file]}.json")
+    raise IOError, "No json-schema file at #{schema_path}" unless File.exist? schema_path
+
+    transformer = SchemaCamelizer.new(schema_path.to_s)
+    transformer.save!
+    print "Saved camelized schema to #{transformer.camel_path}\n"
+  end
+end
+
+class SchemaCamelizer
```
f1337, line 18: Some YARD docs for the class & its public methods would be awesome.

rakelib/camelize_file.rake
erluti: i added some comments because there is a lot going on

spec/request/user_request_spec.rb
erluti: I don't like giant `before`-blocks and I especially don't like using `lets` to pass in values to a `before`-block.  So I thought duplicating the `before`-block was preferable to that approach.  I'll take another look and see if I can refactor it in a way that's clearer

erluti: I hated it but, I went ahead and used the existing context

PR #4561, opened 2020-07-21
Clean up old migrations

johnpaulashenfelter:
## Description of change

Many of our old migrations are no longer runnable between changes to Rails (migration Rails version numbers), the addition of linters (zero_downtime_migrations), and some issues with live-updating data in a migration controlled by feature flags.

The migrations now run successfully
## Original issue(s)
department-of-veterans-affairs/va.gov-team#11450

The resulting schema is _slightly_ different, possibly a result of pg11 differences?

The schema will be checked against production to verify its content.


annaswims:
Why fix old migrations when people could just run `db:setup` ?

db/migrate/20170815232409_clean_up_education_benefits_data_transfer.rb
```diff
@@ -2,16 +2,16 @@ class CleanUpEducationBenefitsDataTransfer < ActiveRecord::Migration[4.2]
...
-      )
+    #   education_benefits_claim.saved_claim = SavedClaim::EducationBenefits.form_class(form_type).new(
+    #     encrypted_form: education_benefits_claim.read_attribute(:encrypted_form),
+    #     encrypted_form_iv: education_benefits_claim.read_attribute(:encrypted_form_iv)
+    #   )
 
-      education_benefits_claim.save!
-    end
+    #   education_benefits_claim.save!
+    # end
```
johnpaulashenfelter, line 14: There is a flipper feature flag in this model, and at this point in the migration there *are no tables* yet for flipper, which was added recently. They *shouldnt* be here at all or be a little cleaner and easier to use, but this is pragmatic

johnpaulashenfelter:
@annaswims this was mostly to make sure the migrations and the schema match up (basically, they do). Next step is to remove them completely which is likely what this PR turns into.

db/production--generated.diff
```diff
@@ -0,0 +1,16 @@
+13c13
+< ActiveRecord::Schema.define(version: 2020_05_08_201027) do
+---
+> ActiveRecord::Schema.define(version: 2020_06_17_105905) do
+17d16
+<   enable_extension "pg_stat_statements"
```
johnpaulashenfelter, line 6: This was done i *believe* directly in prod for @annaswims

db/orig-production.diff
```diff
@@ -0,0 +1,22 @@
...
+152c152
+<     t.uuid "auto_established_claim_id", null: false
+---
+>     t.uuid "auto_established_claim_id"
+162,163c162,163
+<     t.index ["lay_term"], name: "index_disability_contentions_on_lay_term", using: :gin
+<     t.index ["medical_term"], name: "index_disability_contentions_on_medical_term", using: :gin
+---
+>     t.index ["lay_term"], name: "index_disability_contentions_on_lay_term", opclass: :gin_trgm_ops, using: :gin
+>     t.index ["medical_term"], name: "index_disability_contentions_on_medical_term", opclass: :gin_trgm_ops, using: :gin
```
johnpaulashenfelter, line 22: I think these diffs relate to https://github.com/rails/rails/issues/34493

db/orig-production.diff
```diff
@@ -0,0 +1,22 @@
...
+88c89
+<     t.index ["name"], name: "index_base_facilities_on_name", using: :gin
+---
+>     t.index ["name"], name: "index_base_facilities_on_name", opclass: :gin_trgm_ops, using: :gin
+143d143
+<     t.string "signature_errors", default: [], array: true
+152c152
+<     t.uuid "auto_established_claim_id", null: false
+---
+>     t.uuid "auto_established_claim_id"
```
johnpaulashenfelter, line 16: This is a bit of a concern

db/orig-production.diff
```diff
@@ -0,0 +1,22 @@
...
+---
+> ActiveRecord::Schema.define(version: 2020_05_08_201027) do
+16a17
+>   enable_extension "pg_stat_statements"
+88c89
+<     t.index ["name"], name: "index_base_facilities_on_name", using: :gin
+---
+>     t.index ["name"], name: "index_base_facilities_on_name", opclass: :gin_trgm_ops, using: :gin
+143d143
+<     t.string "signature_errors", default: [], array: true
```
johnpaulashenfelter, line 12: This is a bit of a concern

johnpaulashenfelter:
FWIW https://edgeguides.rubyonrails.org/active_record_migrations.html#old-migrations

db/orig-production.diff
johnpaulashenfelter: Its from migrations from last month that seem not to have been run?

db/production--generated.diff
annaswims: I think that's correct. I recall asking someone in ops to do it.   Not that I especially like having differences, but if developers want to enable it locally they have to change their postgres config and restart their db.

f1337:
> FWIW https://edgeguides.rubyonrails.org/active_record_migrations.html#old-migrations

Aye, I was about to comment and ask about this. Why not update schema.rb and whatnot for some of these old ones, versus fixing them?

f1337:
> @annaswims this was mostly to make sure the migrations and the schema match up (basically, they do). Next step is to remove them completely which is likely what this PR turns into.

+1 for updating schema and removing the migrations vs fixing them.

PR #4565, opened 2020-07-21
check for existing user when setting up mfa

ericbuckley:
when authenticating to enroll in multifactor auth, an existing
user session should be present, however because Redis is ephemeral,
its possible that the existing user record will be deleted, thus
we need to check for its existence

fixes https://github.com/department-of-veterans-affairs/va.gov-team/issues/11598

### DIFF
--- a/app/models/user_session_form.rb
+++ b/app/models/user_session_form.rb
@@ -22,7 +22,11 @@ class UserSessionForm
     @user_identity = UserIdentity.new(normalized_attributes)
     @user = User.new(uuid: @user_identity.attributes[:uuid])
     @user.instance_variable_set(:@identity, @user_identity)
-    if saml_user.changing_multifactor?
+    if saml_user.changing_multifactor? && existing_user.present?
+      # when authenticating to enroll in multifactor auth, an existing
+      # user session should be present, however because Redis is ephemeral,
+      # its possible that the existing user record will be deleted, thus
+      # we need to check for its existence
       @user.mhv_last_signed_in = existing_user.last_signed_in
       @user.last_signed_in = existing_user.last_signed_in
     else


omgitsbillryan:
> however because Redis is ephemeral, its possible that the existing user record will be deleted

Isn't it also possible that the user record never existed in the first place? Seems like this code gets called during the saml callback but before the user and session get created.

ericbuckley:
> > however because Redis is ephemeral, its possible that the existing user record will be deleted
> 
> Isn't it also possible that the user record never existed in the first place? Seems like this code gets called during the saml callback but before the user and session get created.

I had a similar thought but wasn't sure its possible.  The first check in the if block is to see if the authentication call is for a user enrolling in multifactor auth, as far as I can tell we ONLY present options to the user to enroll in multifactor if they are a) logged in and b) don't have it enabled.  Thus I think the first condition would mean that some user record exists or previously did.

omgitsbillryan:
I see. Seems like if

```
saml_user.changing_multifactor?==true && existing_user.present?==false
```

might indicate that something has gone wrong and warrants an exception? 🤷  At the least, we need a spec to exercise this case.

ericbuckley:
> I see. Seems like if
> 
> ```
> saml_user.changing_multifactor?==true && existing_user.present?==false
> ```
> 
> might indicate that something has gone wrong and warrants an exception? 🤷 At the least, we need a spec to exercise this case.

If we stored the User record in the database I would agree, but because Redis records can be dropped for various reasons I'm not sure this is the case.  However, I'm willing to defer to someone that understands the history of the `app/models/user.py`

omgitsbillryan:
The only ways I'm aware of that `user` and `session` get removed from redis is if:

1. the user logs out (we forcibly destroy `session` and `user`)
2. [the TTL of 30 minutes](https://github.com/department-of-veterans-affairs/vets-api/blob/master/config/redis.yml#L5-L7) expires and redis clears them away

ericbuckley:
> The only ways I'm aware of that `user` and `session` get removed from redis is if:
> 
>     1. ~the user logs out (we forcibly destroy `session` and `user`)~ EDIT: I just went down a code-review rabbit hole, and it looks like we no longer forcibly destroy these - so the only way go away is if they expire from Redis
> 
>     2. [the TTL of 30 minutes](https://github.com/department-of-veterans-affairs/vets-api/blob/master/config/redis.yml#L5-L7) expires and redis clears them away

What if Redis runs low on memory, do you know which key eviction policy is setup?

omgitsbillryan:
[Looks like the default is `volatile-lru`](https://aws.amazon.com/premiumsupport/knowledge-center/oom-command-not-allowed-redis/), and I can't find anywhere that we set it explicity. Looking at cloudwatch metrics though, we average ~25% memory usage and have never gotten above 40%. I think it's highly unlikely that our redis elasticache would evict any of our user or session data.

ericbuckley:
> [Looks like the default is `volatile-lru`](https://aws.amazon.com/premiumsupport/knowledge-center/oom-command-not-allowed-redis/), and I can't find anywhere that we set it explicity. Looking at cloudwatch metrics though, we average ~25% memory usage and have never gotten above 40%. I think it's highly unlikely that our redis elasticache would evict any of our user or session data.

thats good to know, my only other thought then is what if it takes the user more than 30 mins to enroll in multifactor?

omgitsbillryan:
Bumping. Only holdup to me is the lack of a spec that exercises this conditional. Mind adding one?

ericbuckley:
@jholton did you have any time last week to look at the nginx logs for users running into this error?

f1337:
If there are no objections, I'm going to submit a spec for this, in order to unblock it.

jholton:
I'm researching instances of the error occurring so we can (hopefully) better understand why it's happening at all.
I'll update in a day or so.

ericbuckley:
closing in favor of https://github.com/department-of-veterans-affairs/vets-api/pull/4671

PR #4566, opened 2020-07-22
camel schema tranformer improvements

erluti:
## Description of change
~https://github.com/department-of-veterans-affairs/vets-api/pull/4560 was created to test a camelized version of a schema in tests.  Used appeals to test to see if this approach is repeatable.  Some schemas can reference other schemas, so added support to convert reference schemas at the same time when camelizing a schema.~

Was originally created to test camelizing, but it turns out that it's already a camelized schema.  But these improvements will be good to merge into 4560

## Original issue(s)
department-of-veterans-affairs/va.gov-team#10976

## of note
This merges in with that original PR


spec/request/appeals_request_spec.rb
```diff
@@ -101,6 +110,15 @@
           expect(response).to match_response_schema('appeals')
         end
       end
+
+      it 'returns a successful camel-inflected response' do
+        VCR.use_cassette('caseflow/appeals_null_eta') do
+          get appeals_endpoint
+          expect(response).to have_http_status(:ok)
+          expect(response.body).to be_a(String)
+          expect(response).to match_camelized_response_schema('appeals')
```
erluti, line 119: Surprised that this test doesn't fail without the key-inflection, I checked the appeals schema and it's already camel'ed.  So actually adding these tests and camelized schemas is not helpful for this spec

PR #4567, opened 2020-07-22
Client-side PDF Generation for 10-10CG

kevinmirc:
**Description**
Allow a user to generate a completed 10-10CG PDF for the purposes of mailing it into the VA if they are not qualified to submit digitally.

**Summary**
- Added endpoint for PDF generation and download
- Moved PDF Fill utilities to its own helper (to be used in other specs)
- Add a StatsD metric to count the number of PDF downloads (and to compare to qualification failures)


kevinmirc:
Thanks, @omgitsbillryan!

PR #4572, opened 2020-07-23
fix login status failure stats

ericbuckley:
After reviewing the `user_login` function in the `V1::SessionController` it was noted that the `UserAttributeErrors` are raised when we initialize the `UserSessionForm`, not when we invoke the `valid?` method.  Thus any of these validation errors immediately bubble up to the `saml_callback` method and don't log `login_stats`.  Adding a begin/rescue block around this code will allow us to issue the login stats before moving back to the `saml_callback` method.

Additionally changes were made the tags associated to `api.auth.login.failure` so we can also track the error issued when a failure happens.

fixes https://github.com/department-of-veterans-affairs/va.gov-team/issues/11595

### DIFF
--- a/app/controllers/v1/sessions_controller.rb
+++ b/app/controllers/v1/sessions_controller.rb
@@ -103,7 +103,7 @@ module V1
       begin
         user_session_form = UserSessionForm.new(saml_response)
         raise_saml_error(user_session_form) unless user_session_form.valid?
-      rescue SAML::SAMLError => e
+      rescue SAML::UserAttributeError => e
         login_stats(:failure, saml_response, e)
         raise
       end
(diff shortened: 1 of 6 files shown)


app/controllers/v1/sessions_controller.rb
```diff
@@ -101,10 +100,12 @@ def authenticate
...
 
     def user_login(saml_response)
-      user_session_form = UserSessionForm.new(saml_response)
-      unless user_session_form.valid?
-        login_stats(:failure, saml_response, user_session_form)
-        raise_saml_error(user_session_form)
+      begin
+        user_session_form = UserSessionForm.new(saml_response)
+        raise_saml_error(user_session_form) unless user_session_form.valid?
+      rescue SAML::SAMLError => e
```
jholton, line 106: Exactly which error are we accounting for here? Is this to catch `SAML::UserAttributeError`s from `UserSessionForm#normalize_saml`?
Are there any other `SAML::SAMLError`s that might occur?

ericbuckley: yeah thats a good point, it should really only be `SAML::UserAttributeError` that could be raised on line 104.  I'll update the `rescue`

PR #4573, opened 2020-07-23
removing redirect logic from url_service

ericbuckley:
The FE now handles the redirect logic needed for cerner so no need for
having it in the API anymore

refs https://github.com/department-of-veterans-affairs/va.gov-team/issues/13584


lib/saml/url_service.rb
```diff
@@ -57,10 +57,6 @@ def base_redirect_url
     def login_redirect_url(auth: 'success', code: nil)
       if auth == 'success'
         return verify_url if user.loa[:current] < user.loa[:highest]
-
-        # if the original auth request specified a redirect, use that
-        redirect_target = @tracker&.payload_attr(:redirect)
-        return redirect_target if redirect_target.present?
       end
```
jholton, line 60: combine the `if` statements?
```ruby
return verify_url if user.loa[:current] < user.loa[:highest] && auth == 'success'
```

ericbuckley: @jholton changed in 77a4ef9

jholton (approved):
love to see more red than green!

config/settings.yml
```diff
@@ -41,10 +41,6 @@ sso:
   cookie_secure: false
   cookie_domain: localhost
 
-ssoe:
-  redirects:
-    myvahealth: https://ehrm-va-test.patientportal.us.healtheintent.com/
```
kfrz, line 46: Any corresponding devops PRs needed here, or credstash removals?

kfrz (approved):
Just one comment, LGTM!

config/settings.yml
ericbuckley: thanks for reminding me.  the following block should be removed after this is deployed to production.  I'll update the linked issue with another task to do so

https://github.com/department-of-veterans-affairs/devops/blob/9e8b921829471b070c356d027dcd3dd8866f4af6/ansible/deployment/config/vets-api/prod-settings.local.yml.j2#L28-L30

ericbuckley: updated. https://github.com/department-of-veterans-affairs/va.gov-team/issues/13584

PR #4575, opened 2020-07-23
Feature Branch for moving ITF Service from EVSS to BGS

bastosmichael:
## Description of change
Feature branch for migrating Intent to File from EVSS to BGS

## Original issue(s)
https://vajira.max.gov/browse/API-1230


modules/claims_api/app/swagger/claims_api/forms/form_0966_response_swagger.rb
```diff
@@ -23,38 +23,70 @@ class Form0966ResponseSwagger
...
+                key :type, :integer
+                key :example, '234'
+                key :description, I18n.t('claims_api.field_descriptions.participant_claimant_id')
+              end
+              property :participant_vet_id do
+                key :type, :integer
+                key :example, '234'
+                key :description, I18n.t('claims_api.field_descriptions.intent_to_file_id')
+              end
+              property :signature_indicated do
```
bastosmichael, line 79: I thought we removed this from the BGS call, do we still want to include it if we aren't sending it upstream anymore?

modules/claims_api/app/swagger/claims_api/forms/form_0966_v0_example.json
```diff
@@ -0,0 +1,10 @@
+{
+  "type": "form/0966",
+  "attributes": {
+    "type": "compensation",
+    "participant_claimant_id": 123456789,
+    "participant_vet_id": 987654321,
+    "received_date": "2015-01-05T17:42:12.058Z",
+    "signature_indicated": "YES"
```
bastosmichael, line 8: I would recommend removing.

modules/claims_api/app/swagger/claims_api/forms/form_0966_v1_example.json
```diff
@@ -0,0 +1,10 @@
+{
+  "type": "form/0966",
+  "attributes": {
+    "type": "compensation",
+    "participant_claimant_id": 123456789,
+    "participant_vet_id": 987654321,
+    "signature_indicated": "YES"
```
bastosmichael, line 7: Recommend removing.

modules/claims_api/config/locales/en.yml
```diff
@@ -2,3 +2,17 @@ en:
...
+  claims_api:
+    field_descriptions:
+      expiration_date: Date the ITF expires, this 1 year from the received_date
+      intent_to_file_id: Primary key ID assigned by the VA for the ITF
+      journal_date: Date the ITF was entered into the VA Journal
+      journal_object_id: The VA application that took the last action on the ITF
+      participant_claim_id: Participant ID or SSN of the claimant
+      participant_vet_id: Participant ID or SSN of the Veteran
+      received_date: Date the Intent to File was received at the VA
+      signature_indicated: Indicator that the signature was added
```
bastosmichael, line 14: Recommend removing.

modules/claims_api/app/swagger/claims_api/v1/form_0966_controller_swagger.rb
```diff
@@ -129,8 +130,9 @@ class Form0966ControllerSwagger
                 property :data do
                   key :type, :object
                   key :required, [:attributes]
-                  key :example, type: 'form/0966', attributes: { type: 'compensation' }
+                  key :example, JSON.parse(File.read(EXAMPLE_PATH))
                   property :attributes do
+                    key :required, %i[participant_claimant_id participant_vet_id signature_indicated]
```
bastosmichael, line 135: I don't think these should be required, only optional and we should be pulling participant_vet_id from MVI when not included.

modules/claims_api/app/swagger/claims_api/forms/form_0966_v1_example.json
```diff
@@ -0,0 +1,10 @@
+{
+  "type": "form/0966",
+  "attributes": {
+    "type": "compensation",
+    "participant_claimant_id": 123456789,
+    "participant_vet_id": 987654321,
```
bastosmichael, line 6: We should still retain a minimum payload that only includes type and nothing else so that we can do a participant id look up for the vet at least.

charleystran: this was been fixed

modules/claims_api/app/swagger/claims_api/v1/form_0966_controller_swagger.rb
charleystran: this was been fixed

modules/claims_api/config/locales/en.yml
charleystran: this was been fixed

modules/claims_api/app/swagger/claims_api/forms/form_0966_v1_example.json
charleystran: this was been fixed

modules/claims_api/app/swagger/claims_api/forms/form_0966_v0_example.json
charleystran: this was been fixed

modules/claims_api/app/swagger/claims_api/forms/form_0966_response_swagger.rb
charleystran: this was been fixed

modules/claims_api/app/controllers/claims_api/v0/forms/intent_to_file_controller.rb
```diff
@@ -9,20 +9,26 @@ class IntentToFileController < ClaimsApi::BaseFormController
...
-        before_action :check_future_type, only: [:submit_form_0966]
 
         FORM_NUMBER = '0966'
 
         def submit_form_0966
-          response = itf_service.create_intent_to_file(form_type)
-          render json: response['intent_to_file'],
+          bgs_response = bgs_service.intent_to_file.insert_intent_to_file(intent_to_file_options)
+          render json: bgs_response,
                  serializer: ClaimsApi::IntentToFileSerializer
```
edmangimelli, line 18: will bgs_response ever be nil? we'll probably get a 500 if we pass nil to the IntentToFileSerializer

modules/claims_api/app/controllers/claims_api/v1/forms/intent_to_file_controller.rb
```diff
@@ -10,21 +10,26 @@ class IntentToFileController < BaseFormController
...
         before_action :validate_json_schema, only: %i[submit_form_0966 validate]
-        before_action :check_future_type, only: [:submit_form_0966]
 
         FORM_NUMBER = '0966'
         def submit_form_0966
-          response = itf_service.create_intent_to_file(form_type)
-          render json: response['intent_to_file'],
+          response = bgs_service.intent_to_file.insert_intent_to_file(intent_to_file_options)
+          render json: response,
                  serializer: ClaimsApi::IntentToFileSerializer
```
edmangimelli, line 19: same as above --i think we should check that response is truthy.
also noticed that v0 has `bgs_response` and v1 has `response`

modules/claims_api/app/controllers/claims_api/v1/forms/intent_to_file_controller.rb
```diff
@@ -10,21 +10,26 @@ class IntentToFileController < BaseFormController
...
         def active
-          response = itf_service.get_active(active_param)
-          render json: response['intent_to_file'],
-                 serializer: ClaimsApi::IntentToFileSerializer
+          bgs_response = bgs_service.intent_to_file.find_intent_to_file_by_ptcpnt_id_itf_type_cd(
+            target_veteran.participant_id,
+            ClaimsApi::IntentToFile::ITF_TYPES[active_param]
+          )
+          if bgs_response.present?
+            render json: bgs_response&.first,
```
edmangimelli, line 28: if `bgs_response` is present, i don't think we need the safe navigation (`&.`)

modules/claims_api/app/controllers/claims_api/v0/forms/intent_to_file_controller.rb
```diff
@@ -9,20 +9,26 @@ class IntentToFileController < ClaimsApi::BaseFormController
...
         def active
-          response = itf_service.get_active(active_param)
-          render json: response['intent_to_file'],
-                 serializer: ClaimsApi::IntentToFileSerializer
+          bgs_response = bgs_service.intent_to_file.find_intent_to_file_by_ptcpnt_id_itf_type_cd(
+            target_veteran.participant_id,
+            ClaimsApi::IntentToFile::ITF_TYPES[active_param]
+          )
+          if bgs_response.present?
+            render json: bgs_response&.first,
```
edmangimelli, line 27: if `bgs_response` is present, i don't think we need the safe navigation (`&.`)

modules/claims_api/app/serializers/claims_api/intent_to_file_serializer.rb
```diff
@@ -6,5 +6,27 @@ class IntentToFileSerializer < ActiveModel::Serializer
...
     attribute :status
+
+    type :intent_to_file
+
+    def id
+      object[:intent_to_file_id]
+    end
+
+    def type
+      ClaimsApi::IntentToFile::ITF_TYPES.key(object[:itf_type_cd])
```
edmangimelli, line 17: I've never seen `key` before. TIL!

modules/claims_api/app/swagger/claims_api/forms/form_0966_response_swagger.rb
```diff
@@ -23,38 +23,70 @@ class Form0966ResponseSwagger
...
+                  burial
+                  pension
+                ]
+              end
+
+              property :status do
+                key :type, :string
+                key :example, 'active'
+                key :description, I18n.t('claims_api.field_descriptions.status')
+              end
```
edmangimelli, line 61: is there a status enum?

modules/claims_api/app/swagger/claims_api/forms/form_0966_response_swagger.rb
```diff
@@ -23,38 +23,70 @@ class Form0966ResponseSwagger
...
-              key :example, 'active'
-              key :description, 'Says if the Intent to File is Active or Expired'
+              property :participant_claimant_id do
+                key :type, :integer
+                key :example, '234'
+                key :description, I18n.t('claims_api.field_descriptions.participant_claimant_id')
+              end
+              property :participant_vet_id do
+                key :type, :integer
+                key :example, '234'
```
edmangimelli, line 76:
```suggestion
                key :example, 234
```

modules/claims_api/app/swagger/claims_api/forms/form_0966_response_swagger.rb
```diff
@@ -23,38 +23,70 @@ class Form0966ResponseSwagger
...
+                key :description, I18n.t('claims_api.field_descriptions.intent_to_file_id')
+              end
 
-            property :status do
-              key :type, :string
-              key :example, 'active'
-              key :description, 'Says if the Intent to File is Active or Expired'
+              property :participant_claimant_id do
+                key :type, :integer
+                key :example, '234'
```
edmangimelli, line 71:
```suggestion
                key :example, 234
```

modules/claims_api/app/swagger/claims_api/forms/form_0966_response_swagger.rb
```diff
@@ -23,38 +23,70 @@ class Form0966ResponseSwagger
...
+
+              property :status do
+                key :type, :string
+                key :example, 'active'
+                key :description, I18n.t('claims_api.field_descriptions.status')
+              end
+
+              property :intent_to_file_id do
+                key :type, :integer
+                key :example, '234'
```
edmangimelli, line 65:
```suggestion
                key :example, 234
```

modules/claims_api/app/swagger/claims_api/forms/form_0966_response_swagger.rb
```diff
@@ -23,38 +23,70 @@ class Form0966ResponseSwagger
...
+                key :description, I18n.t('claims_api.field_descriptions.participant_claimant_id')
+              end
+              property :participant_vet_id do
+                key :type, :integer
+                key :example, '234'
+                key :description, I18n.t('claims_api.field_descriptions.intent_to_file_id')
+              end
+              property :signature_indicated do
+                key :type, :string
+                key :example, '234'
```
edmangimelli, line 81: this is a good example of a `signature_indicated` string? (i have no idea, just double-checking)

modules/claims_api/app/swagger/claims_api/forms/form_0966_v0_example.json
```diff
@@ -0,0 +1,10 @@
+{
+  "type": "form/0966",
+  "attributes": {
+    "type": "compensation",
+    "participant_claimant_id": 123456789,
+    "participant_vet_id": 987654321,
```
edmangimelli, line 6: i don't think this matters, but, above, I see that this is usually the same number

modules/claims_api/app/swagger/claims_api/forms/form_0966_v1_example.json
edmangimelli, line 6: same question as above

modules/claims_api/app/swagger/claims_api/forms/form_0966_response_swagger.rb
edmangimelli: also, is this an enum too?

modules/claims_api/app/swagger/claims_api/forms/form_0966_response_swagger.rb
```diff
@@ -23,38 +23,70 @@ class Form0966ResponseSwagger
...
-              key :description, 'Says if the Intent to File is Active or Expired'
+              property :participant_claimant_id do
+                key :type, :integer
+                key :example, '234'
+                key :description, I18n.t('claims_api.field_descriptions.participant_claimant_id')
+              end
+              property :participant_vet_id do
+                key :type, :integer
+                key :example, '234'
+                key :description, I18n.t('claims_api.field_descriptions.intent_to_file_id')
```
edmangimelli, line 77: should this be `claims_api.field_descriptions.participant_vet_id` ?

modules/claims_api/config/locales/en.yml
```diff
@@ -2,3 +2,17 @@ en:
   date:
     formats:
       va_form: "%m/%d/%Y"
+  claims_api:
+    field_descriptions:
+      expiration_date: Date the ITF expires, this 1 year from the received_date
```
edmangimelli, line 7:
```suggestion
      expiration_date: Date the ITF expires, this is 1 year from the received_date
```

modules/claims_api/app/swagger/claims_api/forms/form_0966_response_swagger.rb
bastosmichael: Fixed

modules/claims_api/app/swagger/claims_api/forms/form_0966_response_swagger.rb
bastosmichael: Fixed

modules/claims_api/app/swagger/claims_api/forms/form_0966_response_swagger.rb
bastosmichael: Fixed

modules/claims_api/app/controllers/claims_api/v1/forms/intent_to_file_controller.rb
bastosmichael: Done

modules/claims_api/app/controllers/claims_api/v0/forms/intent_to_file_controller.rb
bastosmichael: Done

modules/claims_api/app/controllers/claims_api/v0/forms/intent_to_file_controller.rb
bastosmichael: It won't be nil during the actual submit, meaning we'll get something back even if it's an error from their end, but we should never pass nil if validate_json_schema and validate_json_format does it's job.

modules/claims_api/app/controllers/claims_api/v1/forms/intent_to_file_controller.rb
bastosmichael: Like the comment above because it's the submit this is one of those instances where we want a hard fail because it's actually pushing the request upstream.

modules/claims_api/app/swagger/claims_api/forms/form_0966_v0_example.json
bastosmichael: it should be fine for example purposes.

modules/claims_api/app/swagger/claims_api/forms/form_0966_v1_example.json
bastosmichael: it should be fine for example purposes.

modules/claims_api/app/swagger/claims_api/forms/form_0966_v0_example.json
bastosmichael: we want this to be an integer after all.

modules/claims_api/app/swagger/claims_api/forms/form_0966_v1_example.json
bastosmichael: we want this to be an integer after all.

modules/claims_api/app/controllers/claims_api/v0/forms/intent_to_file_controller.rb
```diff
@@ -9,20 +9,28 @@ class IntentToFileController < ClaimsApi::BaseFormController
...
-                 serializer: ClaimsApi::IntentToFileSerializer
+          bgs_response = bgs_service.intent_to_file.find_intent_to_file_by_ptcpnt_id_itf_type_cd(
+            target_veteran.participant_id,
+            ClaimsApi::IntentToFile::ITF_TYPES[active_param]
+          )
+          if bgs_response.present?
+            bgs_active = bgs_response.detect do |itf|
+              itf.present? && itf[:itf_status_type_cd] == 'Active' && itf[:exprtn_dt].to_datetime > Time.zone.now
+            end
+            render json: bgs_active, serializer: ClaimsApi::IntentToFileSerializer
```
charleystran, line 30: Doesnt this risk passing nil to the serializer? You could get a valid BGS response for a veteran who is found and not have an ITF for that compensation type. Or it doesnt meet the date parameters. Then nil is passed to the serializer and the result would be a 500. In that case I think a 404 would be a more accurate response.

bastosmichael: It'll filter by type yes but good catch on the bgs_active state, refactored it so that it shouldn't be an issue any further. Thanks.

PR #4576, opened 2020-07-23
Add 686c-674 failure mailer

digitaldrk:
## Description of change
This PR adds a mailer and testing to verify all dynamic data is sent through.

The email will get fired if a user's 686c-674 submission fails. [Code for the 686c-674 form is moving through PR](https://github.com/department-of-veterans-affairs/vets-api/pull/4546) and being developed currently.

## Original issue(s)
https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/11339

    * Mailer should already be configured for each env
    * Not on this code particularly but the FE will use Flipper
    * There are sentry alerts in other code associated with this process that will fire, they have tags for our team
    * N/A
    * N/A
    * No concerns

### DIFF
--- a/app/mailers/dependents_application_failure_mailer.rb
+++ b/app/mailers/dependents_application_failure_mailer.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
 class DependentsApplicationFailureMailer < TransactionalEmailMailer
-  TEMPLATE = 'hca_submission_failure'
-
   def build(user_hash)
     @user_hash = user_hash
     template = File.read('app/mailers/views/dependents_application_failure.erb')
--- a/app/mailers/dependents_application_failure_mailer.rb
+++ b/app/mailers/dependents_application_failure_mailer.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+class DependentsApplicationFailureMailer < TransactionalEmailMailer
+  TEMPLATE = 'hca_submission_failure'
+
+  def build(user_hash)
+    @user_hash = user_hash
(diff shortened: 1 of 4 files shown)


app/mailers/dependents_application_failure_mailer.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class DependentsApplicationFailureMailer < ApplicationMailer
+  def build(user_hash)
+    @user_hash = user_hash
```
omgitsbillryan, line 5: Does this instance variable assignment need to exist? Seems like the plain old input param `user_hash` would suffice.

app/mailers/dependents_application_failure_mailer.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class DependentsApplicationFailureMailer < ApplicationMailer
+  def build(user_hash)
```
omgitsbillryan, line 4: Might be a good idea to validate the input at the beginning of `def build(user_hash)` just in case `@user_hash['first_name']`, `@user_hash['last_name']` or `@user_hash['email']` are `nil`. Something like :
```ruby
def build(user_hash)
  user = User.new(user_hash)
  raise "Bad user #{user.errors}" unless user.valid?
```

Probably nitpicky, but my 2 cents.

digitaldrk: @omgitsbillryan I think I'd rather the veteran get the email without the personalization than raise and them not get it. WDYT? 🤔

PR #4583, opened 2020-07-23
using a browser safe ready function to post the saml request

ericbuckley:
some versions of IE 11 don't work with addEventListner.  Using a
function closure at the end of the page should work for all browsers

fixes https://github.com/department-of-veterans-affairs/va.gov-team/issues/11596

### DIFF
--- a/lib/saml/templates/sso_post_form.html.erb
+++ b/lib/saml/templates/sso_post_form.html.erb
@@ -25,9 +25,9 @@
   <% end %>
 
   <script>
-    document.addEventListener("DOMContentLoaded", function(event) {
+    (function() {
       document.getElementById("saml-form").submit();
-    });
+    })();
   </script>
 </body>
 </html>


ericbuckley:
I was unable to test this when deployed to dev because CAG wouldn't boot up for me.  However Patrick was able to test and noted it worked for him on his GFE https://dsva.slack.com/archives/GTUSXTBNE/p1595532238287500

patrickvinograd:
Confirmed, without this fix both Johnny and I observed a blank page on IE on GFEs due to compatibility mode. I deployed it to dev and re-tested and it worked, submitting the form and POSTing to SSOe. 

Because the SAML form is entirely self-contained it appears safe to use this IIFE syntax to execute the method inline, and not use any more complex "page ready" logic.

patrickvinograd:
But, might want to have frontend review group weigh in since it's a JS issue.

PR #4585, opened 2020-07-24
[686] Call VA file number and create PDF job

kathleencrawford:
## Description of change

This PR gets the user's VA file number and then kicks off the PDF generation job, using the veteran information and payload from the front end form submission.  The generated PDF is then uploaded to VBMS.  

## Original issue(s)
department-of-veterans-affairs/va.gov-team#11258

Sentry logging has been added in case the PDF generation or upload fails.  No additional settings or configuration have been added.  

## Testing
- [x] spec created
- [x] local testing


app/services/bgs/dependent_service.rb
```diff
@@ -2,13 +2,60 @@
...
+      external_key = @user.common_name || @user.email
+
+      @service ||= BGS::Services.new(
+        external_uid: @user.icn,
+        external_key: external_key
       )
+    end
+
+    def add_va_file_number_to_payload
+      va_file_number = service.people.find_person_by_ptcpnt_id(@user.participant_id)
```
omgitsbillryan, line 35: What does `service.people` do? Is that also making another service call behind the covers? I'm concerned with someone calling `submit_686c_form` thinking that they're just making one service call, when in reality it could be many more.

omgitsbillryan: Also the abstraction reads a bit funky to me. Shouldn't "find person" return a "person" of some kind? e.g. 
```ruby
bgs_person = service.people.find_person_by_ptcpnt_id(@user.participant_id)
```

app/workers/vbms/submit_dependents_pdf_job.rb
```diff
@@ -0,0 +1,28 @@
...
+
+module VBMS
+  class SubmitDependentsPDFJob
+    include Sidekiq::Worker
+    include SentryLogging
+
+    # Generates PDF for 686c form and uploads to VBMS
+    def perform(saved_claim_id, veteran_info)
+      claim = SavedClaim::DependencyClaim.find(saved_claim_id)
+      claim.format_and_upload_pdf(veteran_info)
```
omgitsbillryan, line 11: I think it might be good practice to validate that the `claim` you've found is valid. It's not very likely, but it is possible that the `dependency_claim.rb` file changes in code & gets redeployed, and this job ends up finding an "old" form of the DependencyClaim.

Are there any attributes that you know are mandatory that you could check for first to fail early just in case they aren't there?

digitaldrk:
> Some comments, but I'm not quite done. Will continue in the morning.

Awesome @omgitsbillryan thank you for getting us going on this one.

app/services/bgs/dependent_service.rb
digitaldrk: Great question. We are calling `service.people` with the intent to get the user's `va_file_number`, I agree that the naming of the variable is confusing. We were so laser focused on getting the va_file_number that we named it that. I will change the variable name to better convey what that is.

To add some context, once we get that 'lib/bgs' code in that you've been helping us with, we will be passing that va_file_number into both the PDF job and a form submit job. The va_file_number is necessary for both the pdf and the form submission. Our thinking was that it's best to do that call one time to mitigate risk of failure than call it in both jobs.

app/workers/vbms/submit_dependents_pdf_job.rb
digitaldrk: Yeah, that's a good idea. We will get a list together and see about adding validation.

digitaldrk: We added some basic validation for the form data getting passed in. LMK if this was what you had in mind.


Link for your convenience:
https://github.com/department-of-veterans-affairs/vets-api/pull/4585/files#diff-a7e4458dfa7cec528b9bdbb7a0cb353eR15

spec/workers/vbms/submit_dependents_pdf_job_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      'veteran_information' => {
+        'birth_date' => '1809-02-12',
+        'full_name' => {
+          'first' => 'WESLEY', 'last' => 'FORD', 'middle' => nil
+        },
+        'ssn' => '796043735',
+        'va_file_number' => '796043735'
+      }
+    }
+  end
```
omgitsbillryan, line 20: If you defined this as a model class, you could then easily define a factory for it.

app/services/bgs/dependent_service.rb
```diff
@@ -2,13 +2,60 @@
...
-        external_uid: current_user.icn,
-        external_key: current_user.common_name
+    include SentryLogging
+
+    def initialize(user)
+      @user = user
+    end
+
+    def get_dependents
+      service.claimants.find_dependents_by_participant_id(@user.participant_id, @user.ssn)
```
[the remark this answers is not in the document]
omgitsbillryan: Also, just to make you aware - you can use [`cache_aside`](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/common/models/concerns/cache_aside.rb) or use [`redis_store`](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/common/models/redis_store.rb) for caching BGS responses.

I'm not very familiar with the BGS service gem, so I apologize if this is already being done there.

app/services/bgs/dependent_service.rb
```diff
@@ -2,13 +2,60 @@
...
+    end
+
+    def submit_686c_form(claim)
+      va_file_number_with_payload = add_va_file_number_to_payload
+
+      VBMS::SubmitDependentsPDFJob.perform_async(claim.id, va_file_number_with_payload)
```
omgitsbillryan, line 18: You're calling the variable as, "va_file_number_with_payload" but when you read it from the queue (and in your rspec) you call it "vet_info". I don't think the abstraction is quite right.

I'm not certain what the right approach is, but took a quick crack at one...
```ruby
bgs_person = service.people.find_person_by_ptcpnt_id(@user.participant_id)
vet_info = VetInfo.new(@user, bgs_person)
VBMS::SubmitDependentsPDFJob.perform_async(claim.id, vet_info.to_686c_form_hash)

### vet_info.rb
class VetInfo
  def initialize(user, bgs_person)
    @user = user
    @bgs_person = bgs_person
  end

  def to_686c_form_hash
    {
      'veteran_information' => {
        'full_name' => {
          'first' => @user.first_name,
          'middle' => @user.middle_name,
          'last' => @user.last_name
        },
        'ssn' => @user.ssn,
        'va_file_number' => @bgs_person[:file_nbr],
        'birth_date' => @user.birth_date
      }
    }
  end
end
```

PR #4598, opened 2020-07-28
9557 vaos fhir create appointment resource

ajmagdub:
## Description of change
VAOS + FHIR - Create Appointment Resource

## Original issue(s)
[VFT Zenhub issue #9557](https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/9557)

Add Appointment post Rails route, Appointment Controller create method and rspec test for Appointment creation.

### DIFF
--- a/modules/vaos/app/services/vaos/v1/fhir_service.rb
+++ b/modules/vaos/app/services/vaos/v1/fhir_service.rb
@@ -49,10 +49,10 @@ module VAOS
       # The interaction is preformed by and HTTP POST command.
       # http://hl7.org/fhir/dstu2/http.html#create
       #
-      # @request ActionDispatch Request containing POST request values.
+      # @body JSON string containing POST request values in the body key.
       #
-      def create(request)
-        perform(:post, @resource_type.to_s, request.body.read)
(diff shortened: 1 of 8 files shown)


modules/vaos/app/services/vaos/v1/fhir_service.rb
```diff
@@ -45,11 +45,21 @@ def search(query_string)
         query_string.blank? ? perform(:get, @resource_type.to_s) : perform(:get, "#{@resource_type}?#{query_string}")
       end
 
+      # The create interaction creates a resource based on the FHIR resource passed in the request body.
+      # The interaction is preformed by and HTTP POST command.
+      # http://hl7.org/fhir/dstu2/http.html#create
+      #
+      # @request ActionDispatch Request containing POST request values.
+      #
+      def create(request)
```
kreek, line 54: Could `create()` take a hash rather than the entire request? (move the body.read up to the controller). Most likely that it's always a request, but in the future the service might be called by something other than a controller.

PR #4601, opened 2020-07-28
[686c] Adds BGSDependents models for 686c form submission

digitaldrk:
## Description of change

This PR begins adding models for the 686c form submission. These models will be mostly responsible for serializing data that we are sending/receiving to/from BGS.

## Original issue(s)
https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/11260


app/models/bgs_dependents/veteran.rb
```diff
@@ -0,0 +1,48 @@
...
+      ]
+
+      if dependents_application['current_marriage_information']
+        vet_info << ['martl_status_type_cd', dependents_application.dig('current_marriage_information', 'type')]
+      end
+
+      vet_info.to_h
+    end
+
+    def veteran_response(participant, person, va_file_number, address, end_product)
```
omgitsbillryan, line 29: Not sure if you have access to the code climate report, it doesn't like that this method takes 5 params:
image

app/models/bgs_dependents/veteran.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+module BGSDependents
+  class Veteran < Base
+    def initialize(proc_id, user)
+      @user = user
+      @proc_id = proc_id
+    end
+
+    def formatted_params(payload, user)
```
omgitsbillryan, line 10: Is this `user` param different from the `@user` instance var?

app/models/bgs_dependents/relationship.rb
```diff
@@ -0,0 +1,28 @@
...
+        begin_dt: format_date(dependent[:begin_date]),
+        end_dt: format_date(dependent[:end_date]),
+        marage_state_cd: dependent[:marriage_state],
+        marage_city_nm: dependent[:marriage_city],
+        marage_trmntn_state_cd: dependent[:divorce_state],
+        marage_trmntn_city_nm: dependent[:divorce_city],
+        marage_trmntn_type_cd: dependent[:marriage_termination_type_code],
+        mthly_support_from_vet_amt: dependent[:living_expenses_paid_amount]
+      }
+    end
```
omgitsbillryan, line 26: It's kind of unorthodox to me that a rails model wouldn't have any attributes.

Here's an example that uses a model initializer to de-serialize a response object into model attributes - [link](https://github.com/department-of-veterans-affairs/vets-api/pull/4574/files#diff-a8bf1190d48d7f74fe39404ce0ed02d7R30-R47) that might be applicable here instead.

digitaldrk: I hear you, that's great thank you for sharing that resource.

PR #4608, opened 2020-07-29
[686] Add spouse model

digitaldrk:
## Description of change

This PR is one of a few that are adding models for the 686c form submission. These models will be mostly responsible for serializing data that we are sending/receiving to/from BGS.

## Original issue(s)


app/models/bgs_dependents/marriage.rb
```diff
@@ -0,0 +1,71 @@
...
+
+    def spouse_attributes
+      marriage_info = {
+        'ssn': @spouse_information['ssn'],
+        'birth_date': @spouse_information['birth_date'],
+        'ever_married_ind': 'Y',
+        'martl_status_type_cd': marital_status,
+        'vet_ind': spouse_is_veteran,
+        'lives_with_vet': lives_with_vet,
+        'alt_address': alt_address
```
omgitsbillryan, line 51: `def alt_address` is defined as a public method on this class which overrides the attribute, so someone who calls `<an instance of BGSDependents::Marriage>.alt_address` will always step into the `def alt_address` logic of `@dependents_application.dig('does_live_with_spouse', 'address')` - is this intentional?

I think what you want here is to get rid of `def alt_address` and just initialize the attribute instead:
```ruby
'alt_address': @dependents_application.dig('does_live_with_spouse', 'address')
```
so that expensive `dig` logic only happens once in initialization.

app/models/bgs_dependents/marriage.rb
```diff
@@ -0,0 +1,71 @@
...
+      @dependents_application.dig('does_live_with_spouse', 'address')
+    end
+
+    def address(marriage_info)
+      dependent_address(
+        @dependents_application,
+        marriage_info['lives_with_vet'],
+        marriage_info['alt_address']
+      )
+    end
```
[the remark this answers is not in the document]
omgitsbillryan: It even seems like `:address` could be an attribute of this model 🤷

app/models/bgs_dependents/marriage.rb
```diff
@@ -0,0 +1,71 @@
...
+        'alt_address': alt_address
+      }.merge(@spouse_information['full_name']).with_indifferent_access
+
+      marriage_info.merge!({ 'va_file_number': @spouse_information['va_file_number'] }) if spouse_is_veteran == 'Y'
+
+      marriage_info
+    end
+
+    def lives_with_vet
+      @dependents_application.dig('does_live_with_spouse', 'spouse_does_live_with_veteran')
```
omgitsbillryan, line 60: Curious about your usage of `dig` rather than referencing the hash directly? Is `@dependents_application['does_live_with_spouse']['spouse_does_live_with_veteran']` not guaranteed to exist?

app/models/bgs_dependents/marriage.rb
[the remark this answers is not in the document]
digitaldrk: > Why does this method accept a marriage_info input param
You are correct, thank you for pointing that out. It's left-over from a refactor.

> I'm also not understanding why it needs to be defined in base.rb. Looks like it could just be written as a one-liner
We have the `dependent_address ` method in base because it's used by another class as well (other class is in another PR). If I try the one-line approach, it goes beyond our line length constraint.

> It even seems like :address could be an attribute of this model
Yeah, that is a great idea, silly I didn't think of it 🤦

app/models/bgs_dependents/marriage.rb
digitaldrk: > def alt_address is defined as a public method on this class which overrides the attribute..

Sorry, this PR has some silly mistakes like that, I have/had tunnel vision 🤦 

> I think what you want here is to get rid of def alt_address and just initialize the attribute instead

I like that idea. I'm fairly certain all this `alt_address`/`lives_with_veteran` business was for the `address` in the end. I've just done address as an attribute like you suggested and placed the result of the alt/vet address in there bc I think that's the real goal with the lives w/ vet/alt address business. WDYT?

app/models/bgs_dependents/marriage.rb
digitaldrk: I am paranoid of calling a method on `nil`, after researching and speaking w/ our FE team. I feel confident in `does_live_with_spouse => spouse_does_live_with_veteran ` being present.

app/models/bgs_dependents/marriage.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+module BGSDependents
+  class Marriage < Base
```
kfrz, line 4: Please document class attributes and methods with [YARD](https://yardoc.org/)

app/models/bgs_dependents/base.rb
```diff
@@ -2,6 +2,12 @@
 
 module BGSDependents
   class Base < Common::Base
```
kfrz, line 4: Please document class attributes and methods with [YARD](https://yardoc.org/)

app/models/bgs_dependents/base.rb
```diff
@@ -2,6 +2,12 @@
...
+    def dependent_address(dependents_application, lives_with_vet, alt_address)
```
kfrz, line 5: Prefer [keyword args](https://thoughtbot.com/blog/ruby-2-keyword-arguments) over positional args

spec/models/bgs_dependents/base_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      'country_name' => 'USA',
+      'address_line1' => 'Alternative LN',
+      'city' => 'Stuart',
+      'state_code' => 'FL',
+      'zip_code' => '21122'
+    }
+  end
+
+  describe '#dependent_address' do
+    it "returns the vets's address" do
```
kfrz, line 31:
```suggestion
    it 'returns the vet\'s address' do
```

spec/models/bgs_dependents/marriage_spec.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+RSpec.describe BGSDependents::Marriage do
```
kfrz, line 4:
```suggestion
require 'rails_helper'

RSpec.describe BGSDependents::Marriage do
```

spec/models/bgs_dependents/marriage_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+  let(:address_output) do
+    {
+      'country_name' => 'USA',
+      'address_line1' => '8200 Doby LN',
+      'city' => 'Pasadena',
+      'state_code' => 'CA',
+      'zip_code' => '21122'
+    }
+  end
```
kfrz, line 48: Prefer factories over fixtures

app/models/bgs_dependents/marriage.rb
omgitsbillryan: > I am just paranoid of calling a method on nil

It's probably better to fail fast than to keep marching along like things are fine. In that vein, I'll go broken record on you, and recommend some active model validations, so that once initialization is complete, you can simply call `raise 'we didnt get all the stuff we were expecting' unless self.valid?`

kathleencrawford:
@kfrz I think I've made all the requested changes, please let me know if I missed anything or misinterpreted anything.  Thanks!

app/models/bgs_dependents/base.rb
kfrz: Thanks, nicely documented!

erikphansen:
I changed which ticket this was linked to.

kathleencrawford:
Oops, sorry about that - thanks @erikphansen, that was a typo on my part.

PR #4617, opened 2020-07-29
[686] [lib] Add Student School

kathleencrawford:
## Description of change

## Original issue(s)


lib/bgs/service.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module BGS
+  class Service
+    # include BGS::Exceptions::BGSErrors
```
LindseySaari, line 5: Can this comment be removed?

app/models/bgs_dependents/child_school.rb
```diff
@@ -7,8 +7,7 @@ class ChildSchool < Base
     attribute :program_information, Hash
     attribute :current_term_dates, Hash
 
-    def initialize(dependents_application, proc_participant)
```
LindseySaari, line 10: Is the proc_participant no longer needed? Just trying to understand what's going on here

app/models/bgs_dependents/child_student.rb
```diff
@@ -7,8 +7,7 @@ class ChildStudent < Base
     attribute :student_networth_information, Hash
     attribute :student_expected_earnings_next_year, Hash
 
-    def initialize(dependents_application, proc_participant)
```
LindseySaari, line 10: Here also (same as above)

lib/bgs/service.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+
+    def bgs_auth
+      {
+        jrn_dt: Time.current.iso8601,
+        jrn_lctn_id: Settings.bgs.client_station_id,
+        jrn_status_type_cd: 'U',
+        jrn_user_id: Settings.bgs.client_username,
+        jrn_obj_id: Settings.bgs.application,
+        ssn: @user[:ssn] # Just here to make the mocks work
```
f1337, line 26: If this is only for test/development, can we wrap this SSN injection in a conditional?

lib/bgs/service.rb
kathleencrawford: I've removed this comment, thank you for pointing that out.

app/models/bgs_dependents/child_school.rb
kathleencrawford: We've been doing a lot of refactoring due to breaking up a larger PR into multiple smaller PRs.  This functionality has been moved to `/lib/bgs/student_school.rb` in the `submit_params` method.

app/models/bgs_dependents/child_student.rb
kathleencrawford: (Same as above) We've been doing a lot of refactoring due to breaking up a larger PR into multiple smaller PRs. This functionality has been moved to /lib/bgs/student_school.rb in the submit_params method.

lib/bgs/student_school.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    def submit_params(child_object)
+      {
+        vnp_proc_id: @proc_id,
+        vnp_ptcpnt_id: @vnp_participant_id
+      }.merge(child_object.params_for_686c)
+    end
+
+    def bgs_service
+      @service ||= BGS::Service.new(@user)
```
omgitsbillryan, line 39: Is `BGS::Services` & `BGS::Service` different? I guess this is really a question about the BGS client lib itself.

digitaldrk: It is different. The gem changed names at not too long ago. I was hesitant to change naming conventions I saw in the lib folder. This is confusing though. I suppose we could rename this folder to something like `form_686_submission` so it would read like `Form686Submission::Service.new(@user)` 🤔


https://github.com/department-of-veterans-affairs/vets-api/pull/4445

WDYT?

lib/bgs/student_school.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    def create
+      bgs_service.create_child_school(submit_params(child_school))
+      bgs_service.create_child_student(submit_params(child_student))
+    end
+
+    private
+
+    def child_school
+      @child_student = BGSDependents::ChildSchool.new(
```
omgitsbillryan, line 20: The `@child_student` instance variable gets overwritten in `def child_student` - is that intentional? or maybe this variable needs to be renamed to `@child_school`?

omgitsbillryan: Also, why does this need to exist as an instance variable at all?

omgitsbillryan: Also also, `def child_school` & `def child_student` look pretty thin. Why not delete these private methods and just add inline instantiations in `def create`? i.e.

```ruby
child_school = BGSDependents::ChildSchool.new(@dependents_application)
child_student = BGSDependents::ChildStudent.new(@dependents_application)
```

lib/bgs/student_school.rb
```diff
@@ -0,0 +1,42 @@
...
+    def initialize(proc_id:, vnp_participant_id:, payload:, user:)
+      @user = user
+      @proc_id = proc_id
+      @vnp_participant_id = vnp_participant_id
+      @dependents_application = payload['dependents_application']
+    end
+
+    def create
+      bgs_service.create_child_school(submit_params(child_school))
+      bgs_service.create_child_student(submit_params(child_student))
```
omgitsbillryan, line 14: I would have expected that `BGS::StudentSchool.new.create` would return a "school", but as written it returns whatever `bgs_service.create_child_student` returns (presumably a "child student"? 🤷 ). Just want to make sure this is intentional.

lib/bgs/student_school.rb
```diff
@@ -0,0 +1,42 @@
...
+        @dependents_application
+      )
+    end
+
+    def submit_params(child_object)
+      {
+        vnp_proc_id: @proc_id,
+        vnp_ptcpnt_id: @vnp_participant_id
+      }.merge(child_object.params_for_686c)
+    end
```
omgitsbillryan, line 36: What do you think about moving this logic into the `BGSDependents::ChildStudent` and/or `BGSDependents::ChildSchool` class? Similar to the respective `def params_for_686c` methods.

lib/bgs/student_school.rb
digitaldrk: Good point @omgitsbillryan . We actually are not concerned with what gets returned, we have no need for it in subsequent calls. 

Perhaps `.call` is better?

lib/bgs/service.rb
kathleencrawford: Done, added a check for env

lib/bgs/student_school.rb
kathleencrawford: Thanks for the recommendation, I've updated as suggested.

lib/bgs/student_school.rb
kathleencrawford: Moved these to the models as suggested.

app/models/bgs_dependents/child_school.rb
kathleencrawford: Not to make it even more confusing, but due to some more refactoring and feedback suggestions, we've added this back in the `params_for_686c` method.

app/models/bgs_dependents/child_student.rb
kathleencrawford: Not to make it even more confusing, but due to some more refactoring and feedback suggestions, we've added this back in the params_for_686c method.

digitaldrk:
> LGTM, but I do have some serious concerns that some of the logic is not well encapsulated and is difficult to understand. I left a number of comments.

Thank you very much everyone for the great comments but also, allowing us to move forward. We have made suggested changes.

PR #4619, opened 2020-07-30
[686] [lib] Add Veteran

kathleencrawford:
## Description of change

## Original issue(s)

### Additional Considerations
I realize that this is over the recommended LOC, but I really didn't change that many files and the spec file itself is 138 lines.


app/models/bgs_dependents/base.rb
```diff
@@ -2,10 +2,51 @@
...
+        suffix_nm: payload['suffix'],
+        brthdy_dt: format_date(payload['birth_date']),
+        birth_state_cd: payload['place_of_birth_state'],
+        birth_city_nm: payload['place_of_birth_city'],
+        file_nbr: payload['va_file_number'],
+        ssn_nbr: payload['ssn'],
+        death_dt: format_date(payload['death_date']),
+        ever_maried_ind: payload['ever_married_ind'],
+        vet_ind: payload['vet_ind'],
+        martl_status_type_cd: 'Married'
```
f1337, line 21: I assume this hard-coding is intentional?

lib/bgs/service.rb
```diff
@@ -0,0 +1,71 @@
...
+      )
+    end
+
+    def find_benefit_claim_type_increment
+      service.data.find_benefit_claim_type_increment(
+        {
+          ptcpnt_id: @user[:participant_id],
+          bnft_claim_type_cd: '130DPNEBNADJ',
+          pgm_type_cd: 'CPL',
+          ssn: @user[:ssn] # Just here to make the mocks work
```
f1337, line 43: If this is only for testing/development, could we put a condition here to only inject this variable in those environments?

lib/bgs/service.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+
+    def bgs_auth
+      {
+        jrn_dt: Time.current.iso8601,
+        jrn_lctn_id: Settings.bgs.client_station_id,
+        jrn_status_type_cd: 'U',
+        jrn_user_id: Settings.bgs.client_username,
+        jrn_obj_id: Settings.bgs.application,
+        ssn: @user[:ssn] # Just here to make the mocks work
```
f1337, line 61: If this is only for testing/development, could we put a condition here to only inject this variable in those environments?

app/models/bgs_dependents/base.rb
kathleencrawford: Good catch.  This has been updated with some logic to populate "Never Married", "Married", or "Separated"

lib/bgs/service.rb
kathleencrawford: Done, added a check for env to determine whether or not to pass ssn.

lib/bgs/service.rb
kathleencrawford: Done, added a check for env to determine whether or not to pass ssn.

app/models/bgs_dependents/veteran.rb
```diff
@@ -52,5 +52,19 @@ def veteran_response(participant, va_file_number, address, end_product)
...
+    def marital_status(dependents_application)
+      return 'Never Married' if dependents_application.dig('veteran_was_married_before') == false
+
+      if dependents_application.dig('does_live_with_spouse', 'spouse_does_live_with_veteran') == false
+        return 'Separated'
+      end
+
+      return 'Married' if dependents_application.dig('does_live_with_spouse', 'spouse_does_live_with_veteran') == true
+
+      nil
```
omgitsbillryan, line 67:
```suggestion
      spouse_lives_with_vet = dependents_application.dig('does_live_with_spouse', 'spouse_does_live_with_veteran')
      return nil if spouse_lives_with_vet.nil?
      spouse_lives_with_vet ? 'Married' : 'Separated'
```

No need to call dig twice. I think the logic can be cleaned up a bit as well.

lib/bgs/vnp_veteran.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    def vet_address(participant_id)
+      address_params = vnp_veteran.create_address_params(@proc_id, participant_id, @veteran_info)
+
+      bgs_service.create_address(address_params)
+    end
+
+    def vnp_veteran
+      @vnp_veteran ||= BGSDependents::Veteran.new(@proc_id, @user)
+    end
```
omgitsbillryan, line 37: Maybe we want to rename this to something else? "veteran" maybe?
```ruby
def veteran
  @veteran ||= BGSDependents::Veteran.new(@proc_id, @user)
end
```

I say this because the name of this class itself is `vnp_veteran`. It's confusing to me that the class `VnpVeteran` would have an instance variable called `@vnp_veteran` that is actually of a different class 🤯

lib/bgs/vnp_veteran.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    def person_params(participant)
+      vnp_veteran.create_person_params(@proc_id, participant[:vnp_ptcpnt_id], @veteran_info)
+    end
+
+    def vet_address(participant_id)
+      address_params = vnp_veteran.create_address_params(@proc_id, participant_id, @veteran_info)
+
+      bgs_service.create_address(address_params)
+    end
```
omgitsbillryan, line 33: I'm not seeing a clear abstraction that this method is encapsulating. Thoughts on deleting this method & moving:
```ruby
address_params = vnp_veteran.create_address_params(@proc_id, participant_id, @veteran_info)
bgs_service.create_address(address_params)
```

directly up into `def create`?

lib/bgs/vnp_veteran.rb
```diff
@@ -0,0 +1,43 @@
...
+    def create
+      participant = bgs_service.create_participant(@proc_id, nil)
+      claim_type_end_product = bgs_service.find_benefit_claim_type_increment
+      va_file_number = bgs_service.get_va_file_number
+      bgs_service.create_person(person_params(participant))
+      bgs_service.create_phone(@proc_id, participant[:vnp_ptcpnt_id], @veteran_info)
+      address = vet_address(participant[:vnp_ptcpnt_id])
+
+      vnp_veteran.veteran_response(participant, va_file_number, address, claim_type_end_product)
+    end
```
omgitsbillryan, line 21: How long do you think this will take to finish executing? Seems like it could be a while with all those service calls.

lib/bgs/service.rb
```diff
@@ -34,6 +34,36 @@ def create_address(address_params)
...
+    end
+
+    def find_benefit_claim_type_increment
+      increment_params = {
+        ptcpnt_id: @user[:participant_id],
+        bnft_claim_type_cd: '130DPNEBNADJ',
+        pgm_type_cd: 'CPL'
+      }
+
+      increment_params.merge!(user_ssn) unless Rails.env.production?
```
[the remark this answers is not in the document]
digitaldrk: @omgitsbillryan We are only concerned with sending the `ssn` if the app is being run locally in a dev's laptop to access mock data.

We were asked to add the conditional to account for this.
https://github.com/department-of-veterans-affairs/vets-api/pull/4619#discussion_r465723705

lib/bgs/vnp_veteran.rb
digitaldrk: @omgitsbillryan Oh yeah, it will take a while.

This is the challenging part of this feature. We have to daisy chain various calls so that we can use ids/values from some to run others.

For example, we are creating a veteran here, We have to use their their `address_id`, `participant_id`, `claim_type_end_product`, and `va_file_number` in subsequent calls where we associate them with their dependents and create a benefit record.

digitaldrk: All of this will run in a job. PR will be coming in soon

lib/bgs/service.rb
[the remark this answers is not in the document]
digitaldrk: Ah, @kathleencrawford informed me that we use mock data in one of our aws env's. Nice catch, we have a fix.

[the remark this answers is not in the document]
omgitsbillryan: Gotcha 👍 . J/w - what do you mean by access mock data?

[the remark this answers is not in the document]
omgitsbillryan: Are you referring to [vets-api-mockdata](https://github.com/department-of-veterans-affairs/vets-api-mockdata)?

lib/bgs/vnp_veteran.rb
digitaldrk: Agreed, that is 🤯 .

lib/bgs/vnp_veteran.rb
omgitsbillryan: Once the job is created, you might be able to delete this class entirely and just put the logic from `def create` into the job.

You'll also have to consider what happens when the job completes - right now at the end of `def create` it calls `vnp_veteran.veteran_response` which looks to be serializing a response. Since jobs don't respond to http requests, you'll probably need to instead save the response into Redis to be fetched in a subsequent GET call to a different endpoint.

omgitsbillryan: I'm a bit nervous that `def create` could be a landmine for another developer down the road who doesn't realize it could hang a request for a long time. Once the job is created, you might be able to delete this class entirely and just put the logic from `def create` into the job. That way, there won't be any mistaken uses of this logic.

app/models/bgs_dependents/veteran.rb
```diff
@@ -52,5 +52,15 @@ def veteran_response(participant, va_file_number, address, end_product)
...
+
+    private
+
+    def marital_status(dependents_application)
+      spouse_lives_with_vet = dependents_application.dig('does_live_with_spouse', 'spouse_does_live_with_veteran')
+
+      return nil if spouse_lives_with_vet.nil?
+
+      spouse_lives_with_vet ? 'Married' : 'Separated'
+    end
```
omgitsbillryan, line 64: What about the never married guard clause? Looks like you blew it away when addressing [my earlier comment](https://github.com/department-of-veterans-affairs/vets-api/pull/4619#discussion_r466554173).
```ruby
return 'Never Married' if dependents_application.dig('veteran_was_married_before') == false
```

spec/lib/bgs/vnp_veteran_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+          'state' => 'California',
+          'city' => 'Slawson'
+        },
+        'marriage_type' => 'OTHER',
+        'marriage_type_other' => 'Some Other type',
+        'view:marriageTypeInformation' => {}
+      }
+    }
+  end
+  let(:formatted_payload) do # This is here for the mocks since they receive formatted params
```
omgitsbillryan, line 31: What are "the mocks" in this context? As someone without context, I'm not entirely sure what this comment is saying. Could you expand a little on this comment?

spec/lib/bgs/vnp_veteran_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+      participant_id: user_object.participant_id,
+      ssn: user_object.ssn,
+      first_name: user_object.first_name,
+      last_name: user_object.last_name,
+      external_key: user_object.common_name || user_object.email,
+      icn: user_object.icn
+    }
+  end
+  let(:all_flows_payload) { FactoryBot.build(:form_686c_674) }
+  let(:current_marriage_details) do
```
omgitsbillryan, line 17: `current_marriage_details` isn't used anywhere - what gives? Can we delete it?

app/models/bgs_dependents/veteran.rb
omgitsbillryan: Is it concerning that blowing away the guard clause didn't cause any specs to fail? 😬

digitaldrk: I am concerned about that too 😆 . Working on that context.

lib/bgs/service.rb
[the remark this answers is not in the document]
digitaldrk: Yep. That's what I'm referring to 👍

lib/bgs/vnp_veteran.rb
digitaldrk: Your concerns are warranted, this is a a wild feature (to me anyway).

I'd like to take a quick step back and make sure we're on the same page, hope you don't mind.

The concept in this PR of creating a 'veteran' is part of many more calls that use the result from this 'veteran'. All of which are run in a single job.

I made a flow chart when we submitted the original big PR that this one derives from. It's a little out dated since we've made changes but I think it still accurately conveys the calls that we do to BGS.

https://user-images.githubusercontent.com/4297274/88064107-d9266280-cb38-11ea-83c2-241b884b2a81.png

spec/lib/bgs/vnp_veteran_spec.rb
digitaldrk: It's for some rspec/test mocks, you're right it is confusing, I had to read the code to figure out what I meant. After thinking about it, I decided to remove the comment since it caused us more confusion than help.

spec/lib/bgs/vnp_veteran_spec.rb
digitaldrk: Nice 👌 Like, left over from test(s) that were necessary before refactor/PR feedback.

PR #4625, opened 2020-07-30
Add Backend Exception Handling for Debts

camerontesterman:
## Description of change
* Adds exception handling to the Debts service. 
* Adds a 400 Bad Request exception for debts within the exceptions yml file.
* Adds JSON schema validation for Debts responses.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/11791

Tested and verified locally.

### DIFF
--- a/config/locales/exceptions.en.yml
+++ b/config/locales/exceptions.en.yml
@@ -384,6 +384,11 @@ en:
         code: 'CASEFLOWSTATUS502'
         detail: 'Received an invalid response from the upstream server'
         status: 502
+      DEBTS400:
+        <<: *external_defaults
+        title: 'Bad Request'
+        code: 'DEBTS400'
+        detail: 'Received a bad request response from the upstream server'
       GI404:
         <<: *external_defaults
         title: 'Record not found'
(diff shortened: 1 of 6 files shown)


config/locales/exceptions.en.yml
```diff
@@ -384,6 +384,11 @@ en:
         code: 'CASEFLOWSTATUS502'
         detail: 'Received an invalid response from the upstream server'
         status: 502
+      DEBTS400:
+        <<: *external_defaults
+        title: 'Bad Request'
+        code: 'DEBTS400'
+        detail: 'Received a bad request response from the upstream server'
```
f1337, line 391: Should we also include `status: 400`?

lib/debts/service.rb
```diff
@@ -2,10 +2,57 @@
...
+        yield
+      end
+    rescue => e
+      handle_error(e)
+    end
+
+    def save_error_details(error)
+      Raven.tags_context(
+        external_service: self.class.to_s.underscore
+      )
```
f1337, line 30: Do we want to include tests to ensure we're sending the expected details to Raven?

config/locales/exceptions.en.yml
camerontesterman: good catch! added.

lib/debts/service.rb
camerontesterman: yeah that's a good idea. I'll add that and then resolve this convo.

camerontesterman: Added

PR #4628, opened 2020-07-31
MVI troubleshooting

annaswims:
## Description of change
I don't understand why we see issues in Sentry where the user has an ICN but when we try to look up the user in MVI they are not found or, at the very least we're getting an `InvalidRequestError` from MVI.  It be nice if we could log those responses, but I'm not sure if we can because of PII.  So, I'm logging what I can.

http://sentry.vfs.va.gov/vets-gov/platform-api-production/issues/39378/events/7245658/

## Original issue(s)
department-of-veterans-affairs/va.gov-team#11814


lib/mvi/service.rb
```diff
@@ -141,29 +141,24 @@ def build_exception(key, error)
...
-          log_console_and_sentry('MVI Invalid Request', :error)
+          log_message_to_sentry('MVI Invalid Request', :error)
         end
       when MVI::Errors::FailedRequestError
-        log_console_and_sentry('MVI Failed Request', :error)
+        log_message_to_sentry('MVI Failed Request', :error)
       end
     end
 
-    def log_console_and_sentry(message, sentry_classification = nil)
```
annaswims, line 162: this method was pointless because `log_message_to_sentry` already writes to the rails log.

lib/mvi/service.rb
```diff
@@ -141,29 +141,24 @@ def build_exception(key, error)
...
 
-    def mvi_error_handler(user_identity, e)
+    def mvi_error_handler(user_identity, e, source = '')
       case e
       when MVI::Errors::DuplicateRecords
-        log_console_and_sentry('MVI Duplicate Record', :warn)
+        log_message_to_sentry('MVI Duplicate Record', :warn)
       when MVI::Errors::RecordNotFound
-        log_console_and_sentry('MVI Record Not Found')
+        Rails.logger.info('MVI Record Not Found')
```
f1337, line 149: Just double-checking that it was intentional to use `Rails.logger` here, without also calling `log_message_to_sentry`?

Also, I don't see a spec example for the `Rails.logger` use-case. Is that something you want to include?

lib/mvi/service.rb
```diff
@@ -188,14 +183,17 @@ def create_profile_message(user_identity)
     # rubocop:enable Layout/LineLength
 
     def message_icn(user)
+      Raven.tags_context(mvi_find_profile: 'icn')
```
f1337, line 186: Could we add some test coverage for this?

lib/mvi/service.rb
```diff
@@ -188,14 +183,17 @@ def create_profile_message(user_identity)
...
       MVI::Messages::FindProfileMessageIcn.new(user.mhv_icn).to_xml
     end
 
     def message_edipi(user)
+      Raven.tags_context(mvi_find_profile: 'edipi')
```
f1337, line 191: Could we add some test coverage for this?

lib/mvi/service.rb
```diff
@@ -188,14 +183,17 @@ def create_profile_message(user_identity)
...
       MVI::Messages::FindProfileMessageEdipi.new(user.dslogon_edipi).to_xml
     end
 
     def message_user_attributes(user)
+      Raven.tags_context(mvi_find_profile: 'user_attributes')
```
f1337, line 196: Could we add some test coverage for this?

lib/mvi/service.rb
annaswims: yeah, that's intentional.  `log_console_and_sentry` was just writing to the log unless there was a second argument.  And, this error is "we didn't find you in MVI" which is expected to happen often and not really an error.

PR #4629, opened 2020-07-31
STEM: Update SCO email content to reflect the final version recevied from EDU on 07/24 #11713

zurbergram:
## Description of change

Updated content of the school certifying official emails and fix an issue with its `ActionMailer::Preview`

## Original issue(s)
department-of-veterans-affairs/va.gov-team#11713


spec/mailers/previews/school_certifying_officials_mailer_preview.rb
```diff
@@ -13,6 +13,8 @@ def build
   private
 
   def recipients(facility_code)
+    return [] if facility_code.blank?
```
omgitsbillryan, line 16: What's going on here?

Side note, what happens when you send and email to an empty array of recipients?

zurbergram: Via https://guides.rubyonrails.org/action_mailer_basics.html#previewing-emails we can generate previews of emails for content testing locally by going to http://localhost:3000/rails/mailers/

zurbergram: This change makes it so that http://localhost:3000/rails/mailers/school_certifying_officials_mailer/build?facility_code=11000139 works without the `facility_code` param in the url

zurbergram: In the [sidekiq job](https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/workers/education_form/send_sco_email.rb#L94) it checks if there are recipients before attempting to actually send an email

omgitsbillryan: Having 0 recipients for an email seems like a noteworthy thing. Maybe incrementing a StatsD metric or raising and exception.... Granted it's completely unrelated to this PR, and I'm not an expert on that is and isn't exceptional on this sco feature.

zurbergram: Added StatsD metric to the sidekiq job

zurbergram:
> LGTM.
> 
> Unrelated note - are these preview mailer routes only meant for localhost use? / Are the routes disabled for prod, staging & dev?

I don't think these are available in staging or prod.  https://staging-api.va.gov/rails/mailers/ is not accessible

PR #4630, opened 2020-07-31
Add View Payments Service

digitaldrk:
## Description of change
This PR uses the BGS gem to obtain a User's payment history. This includes the returned payments and issued payments. We will be adding an endpoint for vets-website to use to display this information.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#11828

**Note This does update the `bgs_ext` gem to allow access to the payment history service. I would appreciate a glance at the gemfile.lock file to make sure it looks good to go**


app/services/bgs/payment_service.rb
```diff
@@ -0,0 +1,44 @@
...
+  class PaymentService
+    include SentryLogging
+
+    def initialize(current_user)
+      @current_user = current_user
+    end
+
+    def payment_history
+      # rubocop:disable Rails/DynamicFindBy
+      response = service.payment_history.find_by_ssn(@current_user.ssn)
```
f1337, line 13: Does the following not work?

```suggestion
      response = service.payment_history.find_by(ssn: @current_user.ssn)
```

digitaldrk: @f1337 Good question. This is not an ActiveRecord query. Rubocop thinks it is though, that's why I had to add the disable.

Here is what we're actually calling: https://github.com/department-of-veterans-affairs/bgs-ext/blob/8e994a59b2abe4204dde14147e33a5ac98a3e7c2/lib/bgs/services/payment_history.rb#L10

PR #4636, opened 2020-08-03
Update 10-10CG PDF Generation: remove signiture fill, write label of …

kevinmirc:
Fixes: https://github.com/department-of-veterans-affairs/va.gov-team/issues/11536
Fixes: https://github.com/department-of-veterans-affairs/va.gov-team/issues/11429

Dependent on: https://github.com/department-of-veterans-affairs/vets-json-schema/pull/466


lib/pdf_fill/forms/va1010cg.rb
```diff
@@ -681,6 +646,14 @@ def merge_veteran_last_treatment_facility_helper
...
         }
       end
+
+      def merge_planned_facility_label_helper
+        target_facility_code = @form_data.dig 'veteran', 'plannedClinic'
+        caregiver_facilities = VetsJsonSchema::CONSTANTS['caregiverProgramFacilities'].values.flatten
+        selected_facility = caregiver_facilities.find { |facility| facility['code'] == target_facility_code }
+        display_value = selected_facility.nil? ? nil : "#{selected_facility['code']} - #{selected_facility['label']}"
+        @form_data['helpers']['veteran']['plannedClinic'] = display_value
+      end
```
f1337, line 656: Do we have spec coverage for this change?

kevinmirc: Yes. This is covered in the `PdfFill::Filler#fill_form` spec, [seen here](https://github.com/department-of-veterans-affairs/vets-api/blob/master/spec/lib/pdf_fill/filler_spec.rb#L44), which will assert that the "Planned Facility" field matches what's expected (the same value that's in the fixture).

PR #4638, opened 2020-08-03
Make "allow_online_10_10cg_submissions" flipper cookie based

kevinmirc:
Fixes: https://github.com/department-of-veterans-affairs/va.gov-team/issues/11835

**Description**:
This feature flag is used on the front end to redirect users who are not allowed to do a 10-10cg Submission. This will be used (initially) to allow specified users to fill out the form and (later) to do a percentage-based release to the public.

**Summary**:
- Make the "allow_online_10_10cg_submissions" feature flipper cookie-based.
- Remove the effect of "allow_online_10_10cg_submissions" flipper to prevent submission on the backend.


kevinmirc:
Closing for now since the release strategy for this form is still being determined. We may end up making two toggles, one for a whitelist of users and one for percent of visitors.

kevinmirc:
We've decided to use cookie-based feature toggle for beta users and our progressive public rollout strategy.

PR #4641, opened 2020-08-03
10203 STEM applicant emails

dmccurdybah:
## Description of change
Added a new mailer that sends a confirmation email for the 22-10203 form.  This is a new form and is still behind a feature flag. The applicant email is required in the schema and the email is always sent on submit.

## Original issue(s)
department-of-veterans-affairs/va.gov-team/11809

- Implemented behind the `edu_benefits_stem_scholarship` feature flag


app/mailers/views/stem_applicant_confirmation.html.erb
```diff
@@ -0,0 +1,46 @@
...
+    <p>
+      <b>Your claim was sent to:</b>
+      <br />
+      Eastern Region
+      <br />
+      VA Regional Office
+      <br />
+      P.O. Box 4616
+      <br />
+      Buffalo, NY 14240-4616
```
shawkey-daniel, line 32: Is this handy in the data? I don't think it will change, but theoretically it could. If it is handy, it would be nice if it was dynamic, otherwise not necessary.

dmccurdybah: Yeah, I think the address is a constant value so I'll try to pull that in

app/mailers/stem_applicant_confirmation_mailer.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+class StemApplicantConfirmationMailer < TransactionalEmailMailer
+  SUBJECT = 'VA Rogers STEM Scholarship, Application Confirmation'
+  GA_CAMPAIGN_NAME = 'stem_applicant_confirmation-10203-submission-notification'
+  GA_DOCUMENT_PATH = '/email/form'
```
f1337, line 6:
```suggestion
  GA_DOCUMENT_PATH = '/email/form'.freeze
```

I'll concede it's extra-paranoid, but freezing path strings means we're less susceptible to path-traversal vulnerabilities, should another layer of security fail.

dmccurdybah: image

It looks like rubocop doesn't like that

f1337: oh, it's already immutable, my bad

PR #4642, opened 2020-08-03
[686] [lib] Add Dependents

kathleencrawford:
## Description of change

## Original issue(s)

### Other Considerations
I realize this PR is over 250 LoC, but this is part of a much larger PR that we've already broken up and specs account for 250+ lines.


lib/bgs/dependents.rb
```diff
@@ -0,0 +1,173 @@
...
+    def initialize(proc_id:, payload:, user:)
+      @proc_id = proc_id
+      @payload = payload
+      @dependents = []
+      @dependents_application = @payload['dependents_application']
+      @user = user
+    end
+
+    # rubocop:disable Metrics/PerceivedComplexity
+    # rubocop:disable Metrics/CyclomaticComplexity
```
f1337, line 14: I hesitate to be that person, but, I agree with Rubocop here. The complexity behind each private method used is begging to be wrapped in distinct objects.

I don't think we need to block the merge on this, but for the sake of future maintenance and debugging, I think we need to immediately follow-up with these methods broken out.

lib/bgs/dependents.rb
```diff
@@ -0,0 +1,173 @@
...
+
+    def report_deaths
+      @dependents_application['deaths'].each do |death_info|
+        death = BGSDependents::Death.new(death_info)
+        relationship_types = death.relationship_type(death_info)
+
+        next if relationship_types[:family] == 'Child' # BGS does not support child death at this time
+
+        formatted_info = death.format_info
+        death_info['location']['state_code'] = death_info['location'].delete('state')
```
f1337, line 61: Is there a reason for `.delete('state')` over `['state']`?

lib/bgs/dependents.rb
```diff
@@ -0,0 +1,173 @@
...
+        participant = bgs_service.create_participant(@proc_id)
+        bgs_service.create_person(person_params(death, participant, formatted_info))
+        # I think we need the death_location instead of creating an address
+        # There is no support in the API for death location
+        # create_address(@proc_id, participant[:vnp_ptcpnt_id], death_info['location'])
```
f1337, line 66: Can we remove this commented-out code, or will this be used in the future?

lib/bgs/dependents.rb
kathleencrawford: Thanks for your feedback and suggestion.  I agree with you...and plan on iterating and refactoring this in the future so will take that into consideration.  Thank you for not blocking this merge for now.

lib/bgs/dependents.rb
kathleencrawford: The `.delete('state')` will replace the `state` key with a new `state_code` key (instead of leaving both `state` and `state_code` keys).

lib/bgs/dependents.rb
kathleencrawford: I've modified this comment - removed the old/unused code but left a comment that we will need to add functionality there once BGS implements it.

PR #4645, opened 2020-08-04
Urgent Care Mashup

darkbushido:
this will make requests to both lighthouse and PPMS and merge the results

## Description of change

Creating a new endpoint that will combine the urgent care results from PPMS and Lighthouse

## Original issue(s)
department-of-veterans-affairs/va.gov-team#10318

### DIFF
--- a/app/controllers/v1/facilities/ccp_controller.rb
+++ b/app/controllers/v1/facilities/ccp_controller.rb
@@ -42,17 +42,6 @@ class V1::Facilities::CcpController < FacilitiesController
     @api ||= Facilities::PPMS::Client.new
   end
 
-  def pagination_params
-    hsh = super
-    page = Integer(hsh[:page] || 1)
-    per_page = Integer(hsh[:per_page] || 1)
-    total_entries = page * per_page + 1
-    hsh.compact.transform_values!(&:to_i)
-    hsh.merge(
-      total_entries: total_entries
-    )
-  end
-
   def ppms_params
     params.permit(
       :address,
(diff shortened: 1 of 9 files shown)


app/controllers/v1/facilities/ccp_controller.rb
```diff
@@ -39,7 +39,7 @@ def specialties
   private
 
   def api
-    @api ||= Facilities::PPMS::Client.new
+    Facilities::PPMS::Client.new
```
f1337, line 42: Why remove the instance var cache?

darkbushido: No reason, it wasnt re-added

app/controllers/v1/facilities/va_ccp_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+    lighthouse_api_results = lighthouse_api.get_facilities(
+      lighthouse_params.merge(
+        type: :health,
+        services: ['UrgentCare']
+      )
+    )
+
+    providers_facilities = PPMS::ProviderFacility.new
+    providers_facilities.providers = ppms_api_results
+    providers_facilities.facilities = lighthouse_api_results
```
f1337, line 21: Could we refactor lines 6-21 into a command or class? The code is very clean & readable (thank you! 🙇), but the merging of results seems like a `PPMS::ProviderFacility` concern, more than a controller concern.

f1337:
We can override the dangerbot on this one. The size in this case was not indicative of complexity.

app/controllers/v1/facilities/va_ccp_controller.rb
darkbushido: Yes but No. PPMS::ProviderFacility is more of a dummy object that makes fast_jsonapi 'include' all of the results from each resource. This is an odd limitation I ran into, the response will only use one Serializer and I could not find another way to get both objects into the response. I will move the logic into the ProviderFacility. it will make it a bit different from how the other facility controllers handle it.

PR #4650, opened 2020-08-05
One-Time Rake Task for Benefits Intake API

edmangimelli:
resolves https://vajira.max.gov/browse/API-1784

> An upstream DB is out of sync. That upstream partner has asked us if we could update some statuses manually, for the consumers' sake. 2020-08-03


modules/vba_documents/lib/tasks/one_time_status_update_20200803.rake
```diff
@@ -0,0 +1,2210 @@
...
+      80ecfbd4-d3eb-4bf5-b8f3-cc0596b81cfa
+      80fcebbb-8602-4274-8280-b71a28bb4045
+      4bc95949-3abe-4270-a032-e48f11345cc0
+      4c005633-0d87-4a46-b85f-da22f5fca862
+      4c461012-5355-46d0-9077-88a14979fb59
+      4c78d2b1-96b7-44f5-b383-b788064fa5cf
+      4cb4132b-6cac-4976-a424-1902c50ead1a
+      4ce8cc8b-928e-45ba-b2c4-c16ba58d3af4
+      4d839061-de4c-43c8-aaaf-12a4515172d4
+      4db6305811e2700f66ee09fe
```
bastosmichael, line 2092: These don't belong to us at all, just remove anything that doesn't look like a standard UUID and we can turn these over to the va.gov side.

edmangimelli: when this rake task runs, it will print out a list of bad uuids (as well as other errors). that's why i left them in

edmangimelli: it's also a nice record of what they gave us

modules/vba_documents/lib/tasks/one_time_status_update_20200803.rake
```diff
@@ -0,0 +1,2210 @@
+# frozen_string_literal: true
+
+# rubocop:disable Metrics/BlockLength
+namespace :vba_documents do
```
johnpaulashenfelter, line 4: I would love to see this in the `:tmp` namespace personally -- easier to be sure to remove it later since intention is clear. But 🤷 not going to push too hard.

bastosmichael: I actually agree, maybe a :vba_documents followed by a :tmp so that we know what this was intended to do but still mark it as a temp rake task.

edmangimelli: @johnpaulashenfelter , what's easier to cleanup later:
```
namespace :vba_documents do
  namespace :temp do
```
or
```
namespace :temp do
  namespace :vba_documents do
```
or is either one good?

johnpaulashenfelter: Personally i like temp as the namespace but both communicate intent. Since this is down in a module my normal assumption of having a `tmp.rake` is a bit less strong.

PR #4666, opened 2020-08-06
adding logout type for v1 session controller

ericbuckley:
Similar to the existing 'slo' type, however /v1/sessions/logout/new will
only purge the users VA.gov session and leave the SSOe session unchanged

closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/11640


spec/controllers/v1/sessions_controller_spec.rb
```diff
@@ -401,6 +401,31 @@ def expect_logger_msg(level, msg)
...
+
+      context 'logout routes' do
+        around do |example|
+          Settings.sso.cookie_enabled = true
+          example.run
+          Settings.sso.cookie_enabled = false
+        end
+
+        context 'routes /sessions/logout/new to SessionsController#new with type: #logout' do
+          it 'redirects' do
```
jholton, line 420: It 'redirects' where?

ericbuckley: I'll add more context

app/controllers/v1/sessions_controller.rb
```diff
@@ -179,15 +178,17 @@ def saml_response_logging(saml_response)
...
+      if type == 'slo'
+        url = url_service.ssoe_slo_url
+        # due to shared url service implementation
+        # clientId must be added at the end or the URL will be invalid for users using various "Do not track"
+        # extensions with their browser.
+        redirect_to params[:client_id].present? ? url + "&clientId=#{params[:client_id]}" : url
       else
-        Rails.logger.info('SLO callback response could not resolve logout request for originating_request_id '\
-          "'#{originating_request_id}'")
+        ssoe_slo_callback
```
patrickvinograd, line 191: Won't this call to ssoe_slo_callback result in the frontend redirecting to `va.gov/logout`?  That seems like it will make it hard to seamlessly get the session synchronization working on the frontend, without bouncing the user through that logout screen.

ericbuckley: That's a good point, it will redirect the user back to the `va.gov/logout` view.  If we [remove the logout check](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/platform/site-wide/user-nav/containers/AutoSSO.jsx#L31), AutoSSO will render on that page and issue an auto login if the conditions match.  What do you think?

patrickvinograd: At that point the user still won't be at the page they were trying to navigate to (say the new user with mismatched session is coming back from MHV, which linked them to some specific health feature page). 

Since the `v1/sessions/logout/new" method doesn't need to involve SAML, can we invoke it as an XHR request in this scenario? In that case we could even ignore the redirect response. 

The part I'm not sure about is whether there's any "logout" processing that needs to happen on the frontend.

ericbuckley: I see what you're saying now.  I worry a little about an XHR request because it has the potential of letting the user's react session state get out of sync with their API session, especially if we ignore the response.

Would it make more sense to ONLY hit this new url if the user had NO SSOe session, and in the case where they do have an SSOe session (but also a mismatch transaction id), then we just issue an auto login call immediately (which should be the equivalent of performing a logout+login).

patrickvinograd: That seems plausible to me, but obviously warrants testing to make sure there are no unintended consequences. 

One thing I'm wondering if we should try to better mitigate is the possibility of the frontend making a /v0/user call while in the mismatched session state. I don't know if we can totally avoid that especially across all permutations of multiple tabs 🤔

patrickvinograd: I remember we changed some of the `reset_session` logic and would want to make sure that decision holds up with this change in mind.

ericbuckley: I just ran the following test on dev.va.gov

- turned off inbound SSOe (to prevent any auto logouts)
- logged in with `vets.gov.user+9` [tab 1]
- logged out using `int.eauth.va.gov/pkmslogout` [tab 2]
- logged into int.eauth.va.gov `using vets.gov.user+262` [tab 2]
- manually logged in on dev using `dev-api.va.gov/v1/sessions/custom/new?authn=...` [tab 1]

When I ran the last step, it effectively logged me out as user #9 and logged me in as user #262 in one step.

ericbuckley: @patrickvinograd if the above sounds good, we can close this PR along with https://github.com/department-of-veterans-affairs/vets-website/pull/13759, in favor of https://github.com/department-of-veterans-affairs/vets-website/pull/13866

ericbuckley:
closed in favor of department-of-veterans-affairs/vets-website#13866

PR #4668, opened 2020-08-07
[686] Updating controller create action

kathleencrawford:
## Description of change

This PR updates the controller and swagger docs for 686c form submission. 

## Original issue(s)

department-of-veterans-affairs/va.gov-team#11282

### Other considerations
Regarding the `va686c674.rb` file, this file is used to generate the PDF.  It looks like I made a lot of changes, but (for the most part) all I did was move one section of the hash into a different section.


app/services/bgs/dependent_service.rb
```diff
@@ -14,6 +14,13 @@ def get_dependents
 
     def submit_686c_form(claim)
       bgs_person = service.people.find_person_by_ptcpnt_id(@user.participant_id)
+
+      # rubocop:disable Rails/DynamicFindBy
+      bgs_person = service.people.find_by_ssn(@user.ssn) if bgs_person.nil?
```
f1337, line 19: I think I left a similar comment on another PR, but does this not work:
```suggestion
      bgs_person = service.people.find_by(ssn: @user.ssn) if bgs_person.nil?
```

digitaldrk: @f1337 @kathleencrawford and I are on the same team. I wanted to link our conversation here for convenience and so you know we are listening to and value your input.

https://github.com/department-of-veterans-affairs/vets-api/pull/4630#discussion_r470093738

PR #4669, opened 2020-08-07
STEM: Update 'Auto Email Sent to SCO:' in Spool File #12205

zurbergram:
## Description of change

Make sure value is always "No" or "Yes" at https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/workers/education_form/templates/10203.erb#L58

## Original issue(s)
department-of-veterans-affairs/va.gov-team#12205


zurbergram:
> Could we include test coverage for the new behavior?

Are you referencing the new function in `app/models/saved_claim.rb` ?

PR #4670, opened 2020-08-07
11585 add camel inflected schemas to ppiu request

erluti:
## Description of change
Adds camelCase schemas for ppiu request specs.  Also camel version of requests and responses to compare.
Adds a new rake task for non-schema-files.

## Original issue(s)


lib/tasks/camelize_file.rake
```diff
@@ -26,4 +26,30 @@ namespace :camelize_file do
...
+      print "Saved camelized json and its references:\n"
+      saved_files.each do |save_path|
+        print " - #{save_path}\n"
+      end
+    end
+    if transformer.unchanged_schemas.any?
+      print "These json files were already camelized (or perhaps have only one word keys?):\n"
+      print "  [#{transformer.unchanged_schemas.join(', ')}]\n"
+    end
+  end
```
omgitsbillryan, line 54: Can we just delete the `:schema` task?

Seems like the only difference between `:json` and `:schema` is that `:schema` defaults the path of the file to `/spec/support/schemas`. If someone wanted to convert a schema, they could still just use this `:json` task, right? They'd just have to provide the full path as the param input.

omgitsbillryan:
Could we make a separate PR dedicated to the new `:json` task? Usages of this new task could then be their own PR's as well.

erluti:
> Could we make a separate PR dedicated to the new `:json` task? Usages of this new task could then be their own PR's as well.

Made a separate PR here: https://github.com/department-of-veterans-affairs/vets-api/pull/4682
Hopefully the description gives enough context to answer your questions

lib/tasks/camelize_file.rake
erluti: There's also a difference in the output path.  I tried to give more context in the breakout PR, let me know if you have more questions or it's unclear https://github.com/department-of-veterans-affairs/vets-api/pull/4682

PR #4674, opened 2020-08-07
11585 add camel inflected schemas to alternate phone request

erluti:
## Description of change
adding a camel-inflected tests and version of alternate_phone_request

## Original issue(s)


spec/request/alternate_phone_request_spec.rb
```diff
@@ -19,6 +20,14 @@
           expect(response).to match_response_schema('phone_number_response')
         end
       end
+      it 'matches the alternate phone schema when camel-inflected' do
+        VCR.use_cassette('evss/pciu/alternate_phone') do
+          get '/v0/profile/alternate_phone', headers: inflection_header
```
johnpaulashenfelter, line 25: Shouldn't the headers here be the default headers *and* the X-Key-Inflection? Maybe headers.merge(inflection_header) or similar?

erluti: I was copying the test above it and it doesn't use any headers on its request.  Is there some magic that applies them automatically?

johnpaulashenfelter: i bet they just never got added

erluti: I don't want to create inconsistencies with the parallel tests and it looks like there's a lot of tests in there not using the headers option.

PR #4682, opened 2020-08-10
Task for camelizing any json

erluti:
## Description of change
The `camelize_file:schema` task was created to support converting files in `spec/support/schemas` to a camelCase version to support a modified version of [SchemaMatchers](https://github.com/department-of-veterans-affairs/vets-api/blob/8c781e409c02895d143fb4f1711ff3f844070209/spec/support/matchers/schema_matchers.rb).  That matcher is used to match a response against a schema explicitly in the `spec/support/schemas` directory, so no pathing is used in the matcher and `camelize_file:schema` is meant to parallel that and puts the output in the file `spec/support/schemas_camelized`.
This new task, `camelize_file:json` was needed because some tests use support json files as fixtures for testing input and output from endpoints.  Those fixture files don't fit into a neat format like the other, so this task supports an input file and then puts the output in the same directory, but adds `in_camel` to the end of the filename.  These files aren't used with any helpers, but loaded manually in the relevant tests.

## Original issue(s)
https://github.com/department-of-veterans-affairs/vets-api/pull/4670#issuecomment-670678189


erluti:
> LGTM.
> 
> Seems like the only difference between `:json` & `:schema` is the input and output file paths. Why not make just one task that takes in `input_file` & `output_file` argument params?

mainly this way is less typing in console, lol.  I personally get annoyed moving/renaming files because you can't even autocomplete the output path.  Maybe if I come back to add a third task I'll revisit how this is structured

PR #4683, opened 2020-08-10
Migrating from EVSS to BGS on PowerOfAttorneyVerifier

bastosmichael:
## Description of change
During yesterday's production VSO test we discovered that the two new users Diane was trying to add did not have EDIPI and so EVSS was failing because of it. We need to dig into a work around for this on 2122 prod submissions.

http://sentry.vfs.va.gov/vets-gov/platform-api-production/issues/165351/?query=is%3Aunresolved#extra

The solution may end up just be pulling PoA from BGS instead of EVSS when doing comparisons etc.

## Original issue(s)
https://vajira.max.gov/browse/API-1830


modules/veteran/app/models/veteran/user.rb
```diff
@@ -35,8 +30,11 @@ def set_attributes(attributes_hash)
       attributes_hash.each { |key, value| send("#{key}=", value) if respond_to?("#{key}=") }
     end
 
-    def auth_headers
-      @auth_headers ||= EVSS::AuthHeaders.new(@user).to_h
+    def bgs_service
+      @bgs_service ||= BGS::Services.new(
+        external_uid: Settings.bgs.external_uid,
```
charleystran, line 35: These need to be the ICN and email of the veteran. I think they said we could just pass PID for both also

bastosmichael: Updated to include the PID.

bastosmichael:
We are still blocked until we hear back from BGS for Production access to Claimant Services

bastosmichael:
No longer blocked by BGS production access, ready for review.

modules/claims_api/app/controllers/claims_api/application_controller.rb
```diff
@@ -127,14 +127,12 @@ def veteran_from_headers(with_gender: false)
...
-                  vet.loa = {
-                    current: header('X-VA-LOA').try(:to_i),
-                    highest: header('X-VA-LOA').try(:to_i)
-                  }
+                  { current: header('X-VA-LOA').try(:to_i), highest: header('X-VA-LOA').try(:to_i) }
                 end
       vet.mvi_record?
       vet.gender = header('X-VA-Gender') || vet.mvi.profile&.gender if with_gender
       vet.edipi = header('X-VA-EDIPI') || vet.mvi.profile&.edipi
+      vet.participant_id = header('X-VA-PID') || vet.mvi.profile&.participant_id
```
charleystran, line 135: Can we change this to try over & to be consistent with line 130? Would like to avoid mixing styles.

bastosmichael: Line 130 is stylistically different from 135 because it is both totally different in scope (this one is passing a single string value while the other is passing a hash) and rubocop method size prevents me from stylistically changing lines 133 through 135 to match the desired results, considering how much data would have to be passed to refactored methods keeping the same style of 133 and 134 made the most sense.

modules/claims_api/spec/services/unsynchronized_evss_claim_service_spec.rb
```diff
@@ -8,22 +8,20 @@
   let(:auth_headers) { EVSS::AuthHeaders.new(user).to_h }
```
edmangimelli, line 8:
```suggestion
```

edmangimelli (approved):
looks good! 🎉

hinzed1127 (approved):

PR #4686, opened 2020-08-10
Api 1135 claims itf via bgs

charleystran:
## Description of change
This PR moves the ITF changes for the claims API to use BGS over EVSS. EVSS only allows compensation ITF but BGS also supports survivor and pension.


modules/claims_api/app/controllers/claims_api/base_form_controller.rb
```diff
@@ -63,5 +63,28 @@ def decode_document(document)
...
+        participant_vet_id: target_veteran.participant_id,
+        received_date: Time.zone.now.strftime('%Y-%m-%dT%H:%M:%S%:z'),
+        signature_indicated: 'Y',
+        submitter_application_icn_type_code: ClaimsApi::IntentToFile::SUBMITTER_CODE,
+        ssn: target_veteran.ssn
+      }
+    end
+
+    def itf_service
+      EVSS::IntentToFile::Service.new(target_veteran)
```
bastosmichael, line 87: Shouldn't this be BGS service?

charleystran: unused. Moved to the base controller before we decided on the /active changes. Removed now

PR #4687, opened 2020-08-11
[686] Updating form 686 PDF to include countries outside US

kathleencrawford:
## Description of change

We recently made some front-end UI changes to the form, to account for locations outisde the US.  This PR is to update the PDF any time a country outside the US is selected.  (Previously, we were not getting country from the front-end.)

## Original issue(s)
department-of-veterans-affairs/va.gov-team#11822

This PR includes one file used for the PDF generation, and then a few generated PDF fixtures for spec testing.  No additional documentation or swagger docs were added, as this just adds on to existing functionality.

## Testing

- [x] local testing


lib/pdf_fill/forms/va686c674.rb
```diff
@@ -1586,10 +1638,16 @@ def merge_child_helpers
...
 
-          # extract postal code
+          # extract country: FE uses 3 char country codes, but pdf expects 2 char country code
+          child['place_of_birth']['country'] = extract_country(child['place_of_birth'])
+
+          # extract postal code and country
           unless child['does_child_live_with_you']
             child['child_address_info']['address']['zip_code'] =
               split_postal_code(child.dig('child_address_info', 'address'))
+            # @TODO TEST THIS 20200804
```
f1337, line 1648: Does this TODO mean there is still testing required?

lib/pdf_fill/forms/va686c674.rb
```diff
@@ -485,15 +485,22 @@ class Va686c674 < FormBase
...
-                question_suffix: 'E',
+                question_suffix: 'I',
                 question_text: 'PREVIOUS MARRIAGE HISTORY > STATE'
+              },
+              'country' => {
+                key: 'veteran_marriage_history.start_location.country[%iterator%]',
+                limit: 2,
+                question_num: 14,
+                question_suffix: 'J',
+                question_text: 'PREVIOUS MARRIAGE HISTORY > COUNTRY'
```
f1337, line 503: Could we (or should we) cover this change with a spec? Same question for the other hash additions in this file.

lib/pdf_fill/forms/va686c674.rb
f1337: Related, should we test this postal-code/country split?

lib/pdf_fill/forms/va686c674.rb
```diff
@@ -1545,6 +1589,10 @@ def merge_previous_marriage_helpers
             # 'annulment' => select_radio_button(reason_marriage_ended == 'ANNULMENT'),
             'other' => select_radio_button(reason_marriage_ended == 'Other')
           }
+
+          # extract country: FE uses 3 char country codes, but pdf expects 2 char country code
+          spouse['start_location']['country'] = extract_country(spouse['start_location'])
+          spouse['end_location']['country'] = extract_country(spouse['end_location'])
```
f1337, line 1595: Could we (should we) add spec coverage for the 3-char => 2-char conversion behavior here? Same for the other additions of `extract_country` in this file.

lib/pdf_fill/forms/va686c674.rb
kathleencrawford: Thank you!  This comment has been removed.  The split postal code is tested when the filler_spec is run and the overflow.pdf gets generated.  You can see the 9 digit test postal code here: https://github.com/department-of-veterans-affairs/vets-api/blob/d33be24b41da7cc55f0e864697c38b9c7ac08897/spec/fixtures/pdf_fill/686C-674/overflow.json#L201

lib/pdf_fill/forms/va686c674.rb
kathleencrawford: There is a filler_spec that generates 3 versions of the PDF (simple, kitchen_sink, and overflow).  I've updated the fixtures to include some locations with countries outside the US, while leaving some inside the US.  The location for marriage history with country outside the US is here: https://github.com/department-of-veterans-affairs/vets-api/blob/d33be24b41da7cc55f0e864697c38b9c7ac08897/spec/fixtures/pdf_fill/686C-674/overflow.json#L69-L73

lib/pdf_fill/forms/va686c674.rb
kathleencrawford: Same comment as above.  The tests are covered in filler_spec when the PDFs get generated based off the data in the payload.  Please see screenshots for examples of how the 3 digit country code is being extracted (and tested) as 2 digit country codes in the PDF:

PreviousMarriageStartLocation


PreviousMarriageEndLocation

lib/pdf_fill/forms/va686c674.rb
kathleencrawford: Here is a screenshot of how that postal code is then generated on the PDF:

SplitPostalCode

PR #4688, opened 2020-08-11
[686] Change User hash to User object

digitaldrk:
## Description of change
This is an adjustment to support a passing in a user object in instead of a hash for the 686c form.

## Original issue(s)

### DIFF
--- a/app/models/bgs_dependents/veteran.rb
+++ b/app/models/bgs_dependents/veteran.rb
@@ -7,7 +7,6 @@ module BGSDependents
     attribute :first_name, String
     attribute :middle_name, String
     attribute :last_name, String
-    attribute :external_key, String
     attribute :icn, String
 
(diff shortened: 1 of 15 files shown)


app/models/bgs_dependents/veteran.rb
```diff
@@ -55,6 +57,19 @@ def veteran_response(participant, va_file_number, address, end_product)
 
     private
 
+    def user_veteran_attributes
+      external_key = @user.common_name || @user.email
```
f1337, line 61: Forward-thinking, non-blocking comment: Do we want to move `external_key` to a method, or do we not foresee re-use of this logic?

digitaldrk: I'm glad you pointed this out. We don't even need the external key since we merged another branch that passed in the user object. Great shout out!

PR #4690, opened 2020-08-11
Pact Authentication

LindseySaari:
## Description of change
For POC, we need to test an authenticated endpoint to verify that PACT can successfully test endpoints that require authenticated users. This requires stubbing the session along with mvi, edipi and vet360 profiles (similar to our test suite). The pact_uri points to a specific path for the user contract & endpoints vs the broker url.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#11565

## Testing Done
- [x] Specs pass
- [x] Green rake task verification result


spec/support/authenticated_session_helper.rb
```diff
@@ -1,12 +1,14 @@
...
-  def sign_in(user = FactoryBot.build(:user, :loa3), token = nil, raw = false)
+  def sign_in(user = FactoryBot.build(:user, :loa3), token = nil, raw = false, pact = false)
     user = user.persisted? ? user : User.create(user)
     token ||= 'abracadabra'
     session_object = Session.create(uuid: user.uuid, token: token)
     session_options = { key: 'api_session', secure: false, http_only: true }
-    if raw
+    if pact
+      session_object
+    elsif raw
```
f1337, line 11: Not for today, because it would introduce side-effects: But someday™, I think it'd be neat of this method `yield if block_given?` instead of passing raw & pact flags.

PR #4691, opened 2020-08-11
11585 add camel inflected schemas to disability compensation form

erluti:
## Description of change
Adding camelCase version of tests and schema for the disability compensation form request spec.

## Original issue(s)

### DIFF
--- a/spec/support/schemas/rated_disability_base.json
+++ b/spec/support/schemas/rated_disability_base.json
@@ -3,7 +3,7 @@
   "required": [
     "decision_code",
     "decision_text",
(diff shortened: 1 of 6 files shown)


spec/support/schemas/rated_disability_base.json
```diff
@@ -3,7 +3,7 @@
   "required": [
     "decision_code",
     "decision_text",
-		"diagnostic_code",
+    "diagnostic_code",
```
erluti, line 6: this file was previously using tabs on this line (which consequently broke my camelizing script), so i switched it to spaces

PR #4698, opened 2020-08-12
11585 schema camelizer supports nested directories

erluti:
## Description of change
Some of the tests that use schema matchers refer to schemas in a directory nested under `spec/support/schemas` and this change is so the `SchemaCamelizer` supports those schema files.
Also cleaned up the error message around saving when the output path is not different.

## Original issue(s)

### DIFF
--- a/spec/lib/tasks/support/schema_camelizer_spec.rb
+++ b/spec/lib/tasks/support/schema_camelizer_spec.rb
@@ -146,7 +146,8 @@ describe SchemaCamelizer do
       File.open(schema_file_in_weird_location, 'w') { |file| file.write(JSON.pretty_generate(schema)) }
 
       subject = SchemaCamelizer.new(schema_file_in_weird_location)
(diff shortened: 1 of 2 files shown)


spec/lib/tasks/support/schema_camelizer_spec.rb
```diff
@@ -146,8 +147,17 @@ def create_source_schema(name, hash)
       File.open(schema_file_in_weird_location, 'w') { |file| file.write(JSON.pretty_generate(schema)) }
 
       subject = SchemaCamelizer.new(schema_file_in_weird_location)
-      exception_text = 'expected to move from a schemas directory to a schemas_camelized directory!'
+      exception_text = 'expected `#camel_path` (tmp/camel_schema_tests/bad_location.json) ' \
```
erluti, line 150: the linter didn't like the line length, so i put this string on two lines

PR #4703, opened 2020-08-12
[12160] eFolders MVP

camerontesterman:
## Description of change
This change adds an efolder service which will allow the veteran to view and download copies of their documents. This service will be exposed via a `/v0/efolder` route which connects to two methods (`#index` & `#show`) within the efolder controller which is also included in this change.

As mentioned above, two routes have been added:

- `GET /v0/efolder`
- `GET /v0/efolder/{:id}`

The first route will return a JSON response which contains all of the documents available for the veteran to view and download. The second route allows the veteran to download a copy of the document.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/12159
https://github.com/department-of-veterans-affairs/va.gov-team/issues/12160


spec/controllers/v0/efolder_controller_spec.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
```
erluti, line 3: can this be moved to a request spec instead of a controller spec?  That's the recommended approach by the rails and rspec core teams

erluti (changes requested):
Fairly minor change.  I think the test as written will port quickly to a request spec

erluti (approved):
LGTM
Thanks for changing that test!

PR #4709, opened 2020-08-13
Add failure mailer for 686 form submission

digitaldrk:
## Description of change
The 686 form runs many calls to BGS in order to complete its objective of adding dependents to a veteran's benefits. We would like to inform the Veteran that their submission failed. We run retries in almost every call. We then raise an exception if all retries are exhausted.

The goal of this PR is to capture exceptions that have occured during the submission process (in the /lib/bgs directory) and fire a mailer informing the veteran that they need to either revisit the form or get in touch w/ someone at the VA to resolve their issue.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#12265

### DIFF
--- a/app/workers/bgs/submit_form686c_job.rb
+++ b/app/workers/bgs/submit_form686c_job.rb
@@ -11,15 +11,23 @@ module BGS
 
     def perform(user_uuid, saved_claim_id, vet_info)
       user = User.find(user_uuid)
+      claim = valid_claim(saved_claim_id, vet_info)
+
+      BGS::Form686c.new(user).submit(claim.parsed_form)
+    rescue
+      DependentsApplicationFailureMailer.build(user).deliver_now if user.present?
+    end
+
+    private
+
+    def valid_claim(saved_claim_id, vet_info)
       claim = SavedClaim::DependencyClaim.find(saved_claim_id)
 
       raise Invalid686cClaim unless claim.valid?(:run_686_form_jobs)
 
       claim.add_veteran_info(vet_info)
 
-      BGS::Form686c.new(user).submit(claim.parsed_form)
-    rescue
-      DependentsApplicationFailureMailer.build(user).deliver_now if user.present?
+      claim
     end
   end
 end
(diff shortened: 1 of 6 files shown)


spec/mailers/dependents_application_failure_mailer_spec.rb
```diff
@@ -3,28 +3,17 @@
...
+      mailer = described_class.build(user).deliver_now
 
       expect(mailer.subject).to eq("We can't process your dependents application")
-      expect(mailer.to).to eq([user_object.email])
+      expect(mailer.to).to eq([user.email])
       expect(mailer.body.raw_source).to include(
-        "Hello #{user_object.first_name} #{user_object.last_name}",
-        'You recently submitted an application to add or remove dependents from your VA benefits (VA Form 28-686c/674)'
+        "Dear #{user.first_name} #{user.last_name}",
+        "We're sorry. Something went wrong when we tried to submit your application to add or remove a dependent"
```
f1337, line 16: This is a tiny, non-blocking suggestion:

If we make use of Rails built-in i18n, we can use the i18n variables in the ERB template and the spec, such that if we ever have to update the string again, we won't have to update the spec.

app/workers/bgs/submit_form686c_job.rb
```diff
@@ -11,13 +11,23 @@ class Invalid686cClaim < StandardError; end
 
     def perform(user_uuid, saved_claim_id, vet_info)
       user = User.find(user_uuid)
+      claim = valid_claim(saved_claim_id, vet_info)
+
+      BGS::Form686c.new(user).submit(claim.parsed_form)
+    rescue
+      DependentsApplicationFailureMailer.build(user).deliver_now if user.present?
```
f1337, line 18: Are we sure we want to `deliver_now`? Is there a performance cost? I don't even know if we have a mail queue, so it's possible this question is philosophical.

digitaldrk: @f1337 Ah yes, I am open to suggestions.

A search through the project did not return `#deliver_later`, just `#deliver_now`. Also, I figured since we're firin' this in a job, it shouldn't be too much of an issue.

So with the above in mind, I went with `#deliver_now`. This is my first mailer on this project so I could just be naive to what's available.

spec/mailers/dependents_application_failure_mailer_spec.rb
digitaldrk: That's awesome. Great suggestion.

PR #4711, opened 2020-08-13
Pact Forms

LindseySaari:
## Description of change
For POC, we need to test the vets-api forms endpoints to verify that PACT CDCT can be successful with endpoints that relate to forms. This requires stubbing the session along with mvi, edipi and vet360 profiles (similar to our test suite) and making use of VCR cassettes. The session and user stubbing was refactored into a method for provider state usage. The previous implementation was stubbing auth for every test, which is not necessary (or accurate) because some endpoints don’t require an authorized user.

## Docker Notes
Additionally, the below db config in test.yml was required for a green pact status for the docker workflow

```
test_database_url: postgis://postgres:password@postgres:5432/vets_api_test?pool=4
```

## Original issue(s)
department-of-veterans-affairs/va.gov-team#11564

## Testing Done
- [x] Specs pass 
- [x] Green rake task verification result
- [x] broker matrix tagged with branch name and git sha


spec/factories/in_progress_forms.rb
```diff
@@ -237,5 +237,156 @@
...
+          "mothersMaidenName": 'Smith',
+          "privacyAgreementAccepted": true,
+          "testUploadFile": {
+            "filePath": '/src/platform/testing/',
+            "fileName": 'example-upload.png',
+            "fileTypeSelection": '7'
+          }
+        }.to_json
+      end
+    end
```
f1337, line 390: We can address this in a follow-up PR, but IMO this file is getting a bit long. We should probably break each form type out into its own file. Maybe `factories/in_progress_forms/hca_form.rb`.

PR #4719, opened 2020-08-13
[VA Forms] Add field form_details_url

ncksllvn:
## Description of change

This PR adds a property "form_details_url" to the VaForm model of the VA Forms API. This will be info pages covering each form on www.va.gov. An example of a legacy form detail page is https://www.va.gov/vaforms/form_detail.asp?FormNo=10EZ%20(pdf). We are modernizing them.

The logic is simple. The GraphQL lookup always returns the form's detail page under `form.entityUrl.path`. However, the page has to be published as well. Therefore, `form.entityPublished` is checked first. Otherwise, we just leave it null.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/12027


spec/support/schemas/va_forms/form.json
```diff
@@ -29,13 +29,14 @@
...
-            "deleted_at": { "type": ["string", "null"] }, 
-            "related_forms": { "type": ["array", "null"], "items": { "type": "string"}  }, 
+            "form_usage": { "type": ["string", "null"] },
+            "form_tool_intro": { "type": ["string", "null"] },
+            "form_tool_url": { "type": ["string", "null"] },
+            "form_details_url": { "type": ["string", "null"] },
+            "form_type": { "type": ["string", "null"] },
+            "language": { "type": ["string", "null"] },
+            "deleted_at": { "type": ["string", "null"] },
+            "related_forms": { "type": ["array", "null"], "items": { "type": "string"}  },
```
ncksllvn, line 39: IDK why the diff shoes all these lines changed. I only added `"form_details_url": { "type": ["string", "null"] },`

ncksllvn:
This isn't ready for review. I just want to use the review instance.

spec/support/schemas/va_forms/form.json
f1337: It's likely something like a whitespace change.

modules/va_forms/app/workers/va_forms/form_reloader.rb
```diff
@@ -72,6 +65,23 @@ def build_and_save_form(form)
...
+        form_type: form['fieldVaFormType'],
+        form_usage: form.dig('fieldVaFormUsage', 'processed'),
+        form_tool_intro: form['fieldVaFormToolIntro'],
+        form_tool_url: form.dig('fieldVaFormToolUrl', 'uri'),
+        deleted_at: form.dig('fieldVaFormDeletedDate', 'value'),
+        related_forms: form['fieldVaFormRelatedForms'].map { |f| f.dig('entity', 'fieldVaFormNumber') },
+        benefit_categories: map_benefit_categories(form['fieldBenefitCategories'])
+      }
+      mapped[:form_details_url] = "#{FORM_BASE_URL}#{form.dig('entityUrl', 'path')}" if form['entityPublished']
+      mapped
```
ncksllvn, line 82: Robocop made me separate out this block from `build_and_save_form`. Something about too many lines in one function.

ncksllvn: Also, I used `FORM_BASE_URL`, but that constant is technically for the default domain where PDFs are stored, rather than where the form detail pages (HTML) live. Initially I thought about typing out `www.va.gov` again but decided to just reuse the var anyway.

modules/va_forms/app/workers/va_forms/form_reloader.rb
```diff
@@ -72,6 +65,23 @@ def build_and_save_form(form)
...
+        language: form.dig('langcode', 'value'),
+        form_type: form['fieldVaFormType'],
+        form_usage: form.dig('fieldVaFormUsage', 'processed'),
+        form_tool_intro: form['fieldVaFormToolIntro'],
+        form_tool_url: form.dig('fieldVaFormToolUrl', 'uri'),
+        deleted_at: form.dig('fieldVaFormDeletedDate', 'value'),
+        related_forms: form['fieldVaFormRelatedForms'].map { |f| f.dig('entity', 'fieldVaFormNumber') },
+        benefit_categories: map_benefit_categories(form['fieldBenefitCategories'])
+      }
+      mapped[:form_details_url] = "#{FORM_BASE_URL}#{form.dig('entityUrl', 'path')}" if form['entityPublished']
```
bastosmichael, line 81: Is there ever a situation where the form detail page might become unpublished?

ncksllvn: No, at least not for the foreseeable future. Once published, we would expect it to remain published. However, they're going to be published in batches before then so many forms (all of them are unpublished right now) will have `entityPublished == false`

swirtSJW: Not sure if this figures in, but it is possible that a form could be deleted (which would result in the form detail being unpublished and replaced with another form (it would get a new row id that would end up with a different form detail page that would be published later.  There is nothing technical in the CMS that keeps a form detail page from being unpublished.

ncksllvn:
To get past the danger bot I had to move my schema changes into a separate PR, https://github.com/department-of-veterans-affairs/vets-api/pull/4734

PR #4720, opened 2020-08-13
Adds API support for the COVID Research Volunteer Form

LeakyBucket:
## Description of change
This change introduces a new module for the handling of COVID Research related data and business logic.  This initial PR adds a controller for receiving form submissions and logic to validate those against a `vets_json_schema` definition.

There are two significant bodies of logic here: the controller, and a small service class.  The controller has a single action, `create`, which accepts a `POST` with the `JSON` payload for a form submission.  The service currently validates the payload against the schema in `vets_json_schema`.  The controller then returns an _unprocessable_ or _accepted_ response.

It is worth noting that authentication is not required or desired for this specific form.  It is desired that the volunteer process be open to anyone.

The first iteration is only intended to return a success message if the form is valid and errors otherwise.  The next iterations will include data delivery (to `genISIS`) as well as an email notification of receipt.

## Original issue(s)

There are some basic Prometheus metrics being collected with this PR.  Total submissions and validation failures currently, this is expected to expand with the introduction of the back-end integration.

Is there anything special I need to do to strip PII from the logs (in the JSON body)?


f1337:
Hi Glen, would you mind reorganizing the commits or breaking into 2 PRs: 1 commit/PR for the engine boilerplate, and 1 for the covid logic?

LeakyBucket:
@f1337 sure thing! That felt a little weird to me since the boiler plate really does nothing but if it helps you out then 👍

PR #4724, opened 2020-08-14
add more graceful handling of 526 error

charleystran:
## Description of change
This PR makes a more graceful handling of unknown issues during the claims api 526 upload endpoint

### DIFF
--- a/modules/claims_api/app/controllers/claims_api/base_disability_compensation_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/base_disability_compensation_controller.rb
@@ -73,7 +73,7 @@ module ClaimsApi
 
     def unprocessable_response
       {
-        errors: [{detail: 'An unknown error occurred. Please retry the request'}]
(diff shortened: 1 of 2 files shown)


modules/claims_api/app/controllers/claims_api/base_disability_compensation_controller.rb
```diff
@@ -68,5 +70,19 @@ def track_526_validation_errors(errors)
         StatsD.increment STATSD_VALIDATION_FAIL_TYPE_KEY, tags: ["key: #{key}"]
       end
     end
+
+    def unprocessable_response
+      {
+        data: {
```
bastosmichael, line 76: This should be an error and not a data field. https://discuss.jsonapi.org/t/clarification-on-errors-during-get-requests-404-and-partial-response/254

charleystran: updated

bastosmichael:
Looks good but need to get the tests to pass

PR #4741, opened 2020-08-17
Api-204 remove va_facilities module

nathanbwright:
## Description of change
This PR finishes [API-204](https://vajira.max.gov/browse/API-204).
It removes the va_facilities module, which previously served the Facilities API for Lighthouse. This API was split out [into a separate service](https://github.com/department-of-veterans-affairs/lighthouse-facilities). This Rails Engine has not served traffic in several months.
The remaining models and tables can be removed at some point in the future after v0 of the facilities controllers under `/app` are fully deprecated.

## Outstanding questions
- The `ZCTA` constant (which uses the `2018_Gaz_zcta_national.txt`) is still in use in the main app here: https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/models/facilities_query/zip_query.rb#L8 We can just move creating the `ZCTA` constant into the main app's initializers dir (like I have in the PR), but where should the `.txt` file live?
- I'm assuming that we need to keep the `geojson` mime type (that was added in the inital va_facilities PR), is that correct? https://github.com/department-of-veterans-affairs/vets-api/commit/cfd79ba48a17545f5b340217a7fa0e038a8b394f#diff-2d1aa170c0b961bdfa652bdacc112877
- https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/serializers/va_facility_serializer.rb#L4 and https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/serializers/va_suggested_facility_serializer.rb#L4 reference `type 'va_facilities'` does this need to change somehow?
- What should happen to these hard coded urls that reference the `va_facilities` routes? 
https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/lighthouse/facilities/client.rb#L20
https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/lighthouse/facilities/client.rb#L34
- Is the swagger file referencing the new va_facilities api? https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/swagger/swagger/schemas/va_facilities.rb or 

I am making the assumption that `va_facilities` != `facilities`, is that a correct assumption?


nathanbwright:
Closed this because of an issue with running jenkins and danger CI checks from my forked version of the repo. Will create a new PR from a branch in the main repo (now that I have the correct permissions).

PR #4745, opened 2020-08-18
Adding security for form data and Redis

LeakyBucket:
## Description of change

This adds encryption/decryption for the form data as well as some serialization to make it easier to manage the encrypted data.

This logic is intended to be applied before a form is put into Redis (for Sidekiq) and then again by the Sidekiq worker when processing a form before delivering it.

## Original issue(s)

There are definitely potential PII and PHI concerns with this PR.  This is intended specifically to encrypt both PII and PHI before inserting it into Redis.  From my testing I should be able to get out what I put in.  However it does mean that the `iv` is stored with the encrypted data in Redis.

My understanding is that this is functionally equivalent to how it is handled in Postgres so this is probably ok but a focused check there would be appreciated.

The basic intended flow for data would be:

  1. Encryption
  2. Encoding (base64)
  3. Serialization
  4. Store in Redis (Sidekiq enqueue)
  5. Pull from Redis (Sidekiq worker)
  6. Decode
  7. Decrypt
  8. Deliver

The Sidekiq worker and service for performing delivery are not yet in the application, but will be in the next PR.


modules/covid_research/app/services/covid_research/volunteer/form_crypto_service.rb
```diff
@@ -0,0 +1,40 @@
...
+        # AttrEncrypted won't fire without explicit self
+        self.submission = form_data
+
+        {
+          form_data: encrypted_submission,
+          iv: encrypted_submission_iv
+        }
+      end
+
+      def encrypt_and_encode(form_data)
```
f1337, line 30: A comment and/or yardoc here would be helpful. Perhaps something like:
```suggestion
      # @param form_data [String] unencrypted form data
      # @return [Hash] a Hash with encrypted and Base64-encoded submission and iv
      def encrypt_and_encode(form_data)
```

modules/covid_research/app/services/covid_research/volunteer/form_crypto_service.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module CovidResearch
+  module Volunteer
+    class FormCryptoService
+      extend AttrEncrypted
+      attr_encrypted :submission, key: Settings.db_encryption_key
+
+      def decrypt_redis_format(redis_format)
```
f1337, line 9: A comment and/or yardoc here would be helpful. Perhaps something like:
```suggestion
      # @param redis_format [Hash] a Hash with encrypted and Base64-encoded submission and iv
      # @return [String] decrypted form data
      def decrypt_redis_format(redis_format)
```

modules/covid_research/app/services/covid_research/volunteer/form_crypto_service.rb
```diff
@@ -0,0 +1,40 @@
...
+        decrypt_form(redis_format.form_data, redis_format.iv)
```
f1337, line 10: Do we want to add any `Base64.decode`-ing after this? It seems a bit asymmetrical to have plain strings for input, but then output Base64-encoded strings.

LeakyBucket:
@f1337 and I suppose @jholton not sure who to address this update to 😆.

The initial requests/suggestions were in fact very good suggestions/requests.  I have implemented documentation for the new encryption related code.  However during that process I started to dislike the way the two classes interacted and shared partial behaviors etc.  With that in mind I have also refactored the code to what I hope is simpler and cleaner.

PR #4747, opened 2020-08-18
Api 204 remove facilities module

nathanbwright:
## Description of change
This PR finishes [API-204](https://vajira.max.gov/browse/API-204).
It removes the va_facilities module, which previously served the Facilities API for Lighthouse. This API was split out [into a separate service](https://github.com/department-of-veterans-affairs/lighthouse-facilities). This Rails Engine has not served traffic in several months.
The remaining models and tables can be removed at some point in the future after v0 of the facilities controllers under `/app` are fully deprecated.

### Outstanding questions
- The `ZCTA` constant (which uses the `2018_Gaz_zcta_national.txt`) is still in use in the main app here: https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/models/facilities_query/zip_query.rb#L8 We can just move creating the `ZCTA` constant into the main app's initializers dir (like I have in the PR), but where should the `.txt` file live?
- I'm assuming that we need to keep the `geojson` mime type (that was added in the inital va_facilities PR), is that correct? https://github.com/department-of-veterans-affairs/vets-api/commit/cfd79ba48a17545f5b340217a7fa0e038a8b394f#diff-2d1aa170c0b961bdfa652bdacc112877
- https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/serializers/va_facility_serializer.rb#L4 and https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/serializers/va_suggested_facility_serializer.rb#L4 reference `type 'va_facilities'` does this need to change somehow?
- What should happen to these hard coded urls that reference the `va_facilities` routes? 
https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/lighthouse/facilities/client.rb#L20
https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/lighthouse/facilities/client.rb#L34
- Is the swagger file referencing the new va_facilities api? https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/swagger/swagger/schemas/va_facilities.rb or 

I am making the assumption that `va_facilities` != `facilities`, is that a correct assumption?


nfasulo:
>     * The `ZCTA` constant (which uses the `2018_Gaz_zcta_national.txt`) is still in use in the main app here: https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/models/facilities_query/zip_query.rb#L8 We can just move creating the `ZCTA` constant into the main app's initializers dir (like I have in the PR), but where should the `.txt` file live?

I think it would be make sense to put it under `/lib/facilities`.

>     * I'm assuming that we need to keep the `geojson` mime type (that was added in the inital va_facilities PR), is that correct? [cfd79ba#diff-2d1aa170c0b961bdfa652bdacc112877](https://github.com/department-of-veterans-affairs/vets-api/commit/cfd79ba48a17545f5b340217a7fa0e038a8b394f#diff-2d1aa170c0b961bdfa652bdacc112877)

I think you are safe to remove the `geojson` mime type. I don't see any other references to `geojson` in the codebase and the non-Lighthouse Facilities endpoint returns json regardless of the the Accept header you pass in.

>     * https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/serializers/va_facility_serializer.rb#L4 and https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/serializers/va_suggested_facility_serializer.rb#L4 reference `type 'va_facilities'` does this need to change somehow?

Odd though it is that v0 and v2 non-Lighthouse Facilities endpoint are returning the JSON API type "va_facilities" while v1 is returning "facilities", I wouldn't mess with that as there is the possibility that it could affect a consumer of the API. 

>     * What should happen to these hard coded urls that reference the `va_facilities` routes?
>       https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/lighthouse/facilities/client.rb#L20
>       https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/lighthouse/facilities/client.rb#L34

Nothing needs to be done with these. `Lighthouse::Facilities::Client` is pointing to the new implementation of the Lighthouse Facilities API in [this repo](https://github.com/department-of-veterans-affairs/lighthouse-facilities) that has replaced the version in vets-api.

>     * Is the swagger file referencing the new va_facilities api? https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/swagger/swagger/schemas/va_facilities.rb or

This swagger file is for the non-Lighthouse Facilities endpoint. The new Lighthouse Facilities API implementation's swagger file is generated in its repo.

> I am making the assumption that `va_facilities` != `facilities`, is that a correct assumption?

There's not a satisfying an answer to this question in general, but if you have a specific example you want to talk about I might be able to provide to clarity. `facilities` _could_ be used to indicate that the object includes non-VA facilities like community care providers, but I don't have confidence that there is a strict separation between the names `va_facilities` and `facilities`.

nathanbwright:
> > ```
> > * I'm assuming that we need to keep the `geojson` mime type (that was added in the inital va_facilities PR), is that correct? [cfd79ba#diff-2d1aa170c0b961bdfa652bdacc112877](https://github.com/department-of-veterans-affairs/vets-api/commit/cfd79ba48a17545f5b340217a7fa0e038a8b394f#diff-2d1aa170c0b961bdfa652bdacc112877)
> > ```
> 
> I think you are safe to remove the `geojson` mime type. I don't see any other references to `geojson` in the codebase and the non-Lighthouse Facilities endpoint returns json regardless of the the Accept header you pass in.
> 

I see a couple of other places in the app that reference geojson related stuff... So, maybe I'll just leave it for the time being?

PR #4758, opened 2020-08-19
10203 school country

dmccurdybah:
## Description of change
Add a new `schoolCountry` field to the 10203 form.  This is returned as part of the form prefill and displayed on the spool file.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#12553

- 10203 is not live
- Updated vets-json-schema version to 6.7.3.  This adds a required field to the 10203 form so has to be delivered with test updates.


spec/fixtures/education_benefits_claims/10203/minimal.spl
```diff
@@ -57,7 +57,8 @@ Program Name: Degree Name
 
 Provide the location where you plan to or will start training:
 School Name
-Test, TN
+Test,
```
thilton-oddball, line 60: Don't know that it is a big thing but do you want the comma displaying if there is no state?

dmccurdybah: Thanks for checking, yes, that's expected for input into TIMS

PR #4759, opened 2020-08-20
Conditionally sign the 10-10cg on PDF generation

kevinmirc:
**Description**
We generate a VA Form 10-10CG PDF in two cases:
1) When an end-user is not able to submit their form digitally, we allow them to download the completed form to mail in.
2) When an electronic form is submitted successfully, we generate the PDF to attach to the resulting Salesforce Case (in CARMA).

The PDF generation needs to accept runtime options to conditionally sign the form. We want the PDFs going to CARMA to be signed and dated, and we want the PDF end-users are generating to not be signed so they can apply a signature manually before mailing it in.

**Summary**
- Refactor the `PdfFill#fill_form` spec to allow more options to be passed.
- Update `PdfFill#fill_form` spec to accept a fixtures path for any given form test and options pair
- Allow `PdfFill#fill_form` to accept fill_options and pass those to `va{FORM_ID}#merge_fields`
- Update the two places the 10-10CG is generates, and pass it `fill_options: { sign: true | false }`
- Update the PDF signatures to have a DateTime with the timezone
- Update signature to include an identifier that it was electronically signed (/es/)


kevinmirc:
Thanks @thilton-oddball!

PR #4763, opened 2020-08-20
capture invalid saml response

ericbuckley:
Consolidating exception handling in `saml_callback` to a new function to reduce complexity and adding functionality to capture the SAML response payload when an exception is raised.

closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/12485


app/controllers/v1/sessions_controller.rb
```diff
@@ -243,6 +237,23 @@ def callback_stats(status, saml_response = nil, failure_tag = nil)
       end
     end
 
+    # rubocop:disable Metrics/ParameterLists
```
patrickvinograd, line 240: This refactor drops the `      unless performed?` guard clause that was in the second rescue block. Are we confident that's no longer needed?

app/controllers/v1/sessions_controller.rb
```diff
@@ -243,6 +237,23 @@ def callback_stats(status, saml_response = nil, failure_tag = nil)
...
+    def handle_callback_error(exc, status, response, level = :error, context = {},
+                              code = '007', tag = nil)
+      log_message_to_sentry(exc.message, level, extra_context: context)
+      redirect_to url_service(response&.in_response_to).login_redirect_url(auth: 'fail', code: code)
+      login_stats(:failure, response, exc) unless response.nil?
+      callback_stats(status, response, tag || code)
+      PersonalInformationLog.create(
+        error_class: exc,
```
jholton, line 248: This attribute is called `error_class`, but it's just a String.  More importantly, it's how you will identify your `PersonalInformationLog` entries.  So, if you know what value will end up in the database, and you are sure it will be consistent, it's probably fine.  Otherwise, you might consider just changing this to a string literal so you can more easily/reliably retrieve your `PersonalInformationLog` entries.

PR #4764, opened 2020-08-20
Update vets-json-schema

dmccurdybah:
## Original issue(s)


dmccurdybah:
Closing because change will require update to 10203 which will be made in https://github.com/department-of-veterans-affairs/vets-api/pull/4758

PR #4771, opened 2020-08-22
Adding/Updating BGS Required Fields

kathleencrawford:
## Description of change

During integration testing, we discovered some fields that are required by BGS for certain calls.  These fields include:
`
child_prevly_married_ind, jrn_lctn_id
`
## Original issue(s)
department-of-veterans-affairs/va.gov-team#12730

No additional settings or logging was added.  This PR just adds a few fields that were previously thought to be optional but we learned are required.


kathleencrawford:
thank you @thilton-oddball !

PR #4774, opened 2020-08-24
Remove unnecessary upsert gem

annaswims:
## Description of change
Upsert was added to vets-api in https://github.com/department-of-veterans-affairs/vets-api/pull/2418 before rails added upsert https://github.com/rails/rails/pull/35077 

Upsert was only used in Form526JobStatus

### DIFF
--- a/app/workers/evss/disability_compensation_form/job_status.rb
+++ b/app/workers/evss/disability_compensation_form/job_status.rb
@@ -29,7 +29,7 @@ module EVSS
             error_message: error_message,
             updated_at: Time.now.utc
           }
-          Form526JobStatus.upsert({ job_id: jid }, values)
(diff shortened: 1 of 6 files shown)


app/workers/evss/disability_compensation_form/job_status.rb
```diff
@@ -29,7 +29,7 @@ def job_exhausted(msg)
             error_message: error_message,
             updated_at: Time.now.utc
           }
-          Form526JobStatus.upsert({ job_id: jid }, values)
```
annaswims, line 32: the upsert gem and  [Activerecord.upsert ](https://edgeapi.rubyonrails.org/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-upsert)take different arguments

PR #4778, opened 2020-08-25
Add BDD tag to form 526 metrics and fix bugs

annaswims:
## Description of change
This ticket started with the goal of adding BDD metrics to form526, but then expanded to fix 526 job stats bugs.

It doesn't make sense to raise `GatewayTimeout` regardless of the real error.  We want to know what the real issue is if a job is exhausted. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#12376


spec/jobs/evss/disability_compensation_form/submit_form526_all_claim_spec.rb
```diff
@@ -71,9 +74,13 @@
         VCR.use_cassette('evss/disability_compensation_form/submit_form_v2') do
           subject.perform_async(submission.id)
           expect_any_instance_of(EVSS::DisabilityCompensationForm::Metrics).to receive(:increment_retryable).once
-          expect(Form526JobStatus).to receive(:upsert).twice
```
annaswims, line 74: I don't have a good understanding of why, but this test line changed the behavior.  With it, Form526JobStatus was not saved.  🤷‍♀️ I think lines 79-83 cover this just as wel.

spec/models/form526_job_status_spec.rb
```diff
@@ -10,7 +10,7 @@
       {
         form526_submission_id: form526_submission.id,
         job_id: jid,
-        job_class: EVSS::DisabilityCompensationForm::SubmitForm526.class.name.demodulize,
```
annaswims, line 13: EVSS::DisabilityCompensationForm::SubmitForm526.class.name.demodulize
=> "Class" , which is not correct.

app/workers/evss/disability_compensation_form/job_status.rb
```diff
@@ -15,28 +15,26 @@ module JobStatus
         # @param msg [Hash] The message payload from Sidekiq
         #
         def job_exhausted(msg)
-          submission_id = msg['args'].first
-          jid = msg['jid']
-          error_message = msg['error_message']
-          klass = self.class.name.demodulize
```
annaswims, line 21: `self.class.name.demodulize` is always `Class` 😞

app/workers/evss/disability_compensation_form/job_status.rb
```diff
@@ -15,28 +15,26 @@ module JobStatus
...
-
           values = {
-            form526_submission_id: submission_id,
-            job_id: jid,
-            job_class: klass,
+            form526_submission_id: msg['args'].first,
+            job_id: msg['jid'],
+            job_class: msg['class'].demodulize,
             status: Form526JobStatus::STATUS[:exhausted],
-            error_class: nil,
```
annaswims, line 28: No idea why we'd want this to be `nil` when we could add useful info.

app/workers/evss/disability_compensation_form/job_status.rb
```diff
@@ -15,28 +15,26 @@ module JobStatus
...
-            error_message: error_message,
+            error_class: msg['error_class'],
+            error_message: msg['error_message'],
             updated_at: Time.now.utc
           }
           Form526JobStatus.upsert(values, unique_by: :job_id)
 
           Rails.logger.error(
```
annaswims, line 29: I can't imagine anyone reading this log. Delete?

app/workers/evss/disability_compensation_form/submit_form526.rb
```diff
@@ -48,7 +48,7 @@ def response_handler(response)
       def retryable_error_handler(error)
         # update JobStatus, log and metrics in JobStatus#retryable_error_handler
         super(error)
-        raise EVSS::DisabilityCompensationForm::GatewayTimeout, error.message
```
annaswims, line 51: I don't understand why this choice was made originally.  I'd think we'd want to raise whatever the error really is.

rakelib/form526.rake
```diff
@@ -124,7 +124,9 @@ namespace :form526 do
     )
 
     submissions.find_each do |submission|
-      submission.form526_job_statuses.where('error_class is not null').each do |job_status|
+      job_statuses = submission.form526_job_statuses.where.not(status: [Form526JobStatus::STATUS[:try],
```
annaswims, line 127: we were missing out on errors because error_class was nil for exhausted jobs.

spec/models/form526_job_status_spec.rb
thilton-oddball: If you add .new before .class (EVSS::DisabilityCompensationForm::SubmitForm526.new.class.name.demodulize) does it give you what you want?

app/workers/evss/disability_compensation_form/job_status.rb
thilton-oddball: If the above suggestion works (adding .new before .class) it might work the same here.

annaswims: yeah, `msg['class'].demodulize` works too.

spec/models/form526_job_status_spec.rb
annaswims: yeah, that works too.  I just want to test that the value is `SubmitForm526` here.

PR #4780, opened 2020-08-25
Added prefill files

cohnjesse:
## Description of change
[Original Ticket](https://github.com/department-of-veterans-affairs/va.gov-team/issues/12669)

This PR is to add the code for prefill on the form 28-8832.


app/controllers/v0/profile/valid_va_file_numbers_controller.rb
```diff
@@ -6,13 +6,15 @@ class ValidVaFileNumbersController < ApplicationController
...
+        render json: {
+    "data": {
+        "id": "",
+        "type": "valid_va_file_number",
+        "attributes": {
+            "valid_va_file_number": true
+        }
+    }
+}.to_json
       end
```
kfrz, line 18: Looks like some formatting issues here w/ indentation.

app/models/form_profiles/va288832.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+class FormProfiles::VA288832 < FormProfile
+
+  class FormAddress
```
kfrz, line 5:
```suggestion
class FormProfiles::VA288832 < FormProfile
  class FormAddress
```

kfrz, line 5: Please add YARDoc class-level documentation and documentation for any public interface.

config/settings.yml
```diff
@@ -92,16 +92,19 @@ edu:
 dependents:
   prefill: true
 
+vre_counseling:
+  prefill: true
+
```
kfrz, line 97: Looks like all the other changes in this file were victim to an editor's auto-fixing quotes.

Our [`rubocop.yml`](https://github.com/department-of-veterans-affairs/vets-api/blob/master/.rubocop.yml#L33) enforces singe-quote style.

kfrz (changes requested):
Is this build expected to pass at any particular time? What's the rollout plan, otherwise?

app/models/form_profiles/va288832.rb
cohnjesse: Since we are only making use of existing infrastructure I heavily changed this code

app/controllers/v0/profile/valid_va_file_numbers_controller.rb
cohnjesse: This was not supposed to be in the PR, I removed it

app/models/form_profiles/va288832.rb
cohnjesse: Since we are only making use of existing infrastructure I heavily changed this code

config/settings.yml
cohnjesse: ouch, good catch. Updated this.

cohnjesse:
@kfrz the build on this is now passing

digitaldrk (approved):
This looks good to me as long as VSP is ok with the small quote changes in the config/settings.yml file

LindseySaari:
I guess I’d prefer that you only include the changes in settings.yml that are relevant to this PR

LindseySaari (approved):
👍 Going to approve the quote refactoring since it aligns with our current rubocop configurations

cohnjesse:
@LindseySaari I reverted the settings.yml file back to what it was before I changed all of the quotes. I just added my change.

PR #4783, opened 2020-08-26
Claims API Documentation Updates

hinzed1127:
## Description of change

This makes a number of corrections/clarifications to the swagger documentation for the Claims API, including:

- fixing a typo
- removing inaccurate details mentioning OAuth in Claims v0 docs
- making sure response examples are accurate and conform to the JSON:API spec
 
## Original issue(s)

https://vajira.max.gov/browse/API-2094


modules/claims_api/app/swagger/claims_api/v1/form_526_controller_swagger.rb
```diff
@@ -140,7 +140,11 @@ class Form526ControllerSwagger
             key :description, '526 response'
             content 'application/json' do
               schema do
-                key :'$ref', :ClaimsIndex
+                key :type, :object
```
hinzed1127, line 143: @bastosmichael before I go too deep on updating these, I wanna make sure I'm not adding `data` properties to the top level of any responses that don't need it. It was unclear looking through the JSON API spec if that formatting applies to `POST`/`PUT` requests in addition to `GET`

modules/claims_api/app/swagger/claims_api/v0/form_526_controller_swagger.rb
```diff
@@ -521,6 +521,7 @@ class Form526ControllerSwagger
 
           response 200 do
             key :description, 'upload response'
+            # TODO is there anything that should be shown as an example response body?
```
hinzed1127, line 524: the example response is empty, not sure if that's an oversight

modules/claims_api/app/swagger/claims_api/forms/form_2122_example.json
```diff
@@ -1,59 +1,54 @@
...
-        "lastName": "VSOLast",
-        "organizationName": "I Help Vets LLC",
-        "address": {
-          "numberAndStreet": "123 NotReal Pkwy",
-          "city": "sometown",
-          "state": "WI",
-          "country": "US",
-          "zipFirstFive": "54703"
-        }
+{
```
hinzed1127, line 1: The diff for this and the other `json` file is deceiving. All I did was strip out the top-level `data` properties, since defining them in the swagger blocks adds them us already

bastosmichael: Should we be keeping the full payload in the example and remove the extra layer of data form the swagger section that's calling this file? Would be confusing for someone to see the form 2122 example here but it's incorrectly formatted?

modules/claims_api/app/swagger/claims_api/forms/form_526_example.json
```diff
@@ -1,130 +1,126 @@
...
-              "areaCode": "123",
-              "phoneNumber": "1231234"
-            }
-          },
-          "currentlyHomeless": {
-            "homelessSituationType": "fleeing",
-            "otherLivingSituation": "other living situation"
-          }
-        }
+  "type": "form/526",
```
bastosmichael, line 2: Same here might need to keep this payload in tact and instead change the swagger section instead?

modules/claims_api/app/swagger/claims_api/v0/form_526_controller_swagger.rb
bastosmichael: It is an oversight and it looks like no one ever completed the TODO, if you hit the upload end point you get back the ClaimsShow schema with additional files listed. https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb#L47

modules/claims_api/app/swagger/claims_api/v1/form_526_controller_swagger.rb
bastosmichael: yes the data element applies to both the POST and PUT as well as the GET. JSON API spec treats the whole thing as a data model. The reason we return the ClaimsIndex instead of a 526 specific payload is because of the PII and PHI nature of the 526 payload itself and ClaimsIndex is what they get back when they check for status anyhow.

modules/claims_api/app/swagger/claims_api/v1/form_526_controller_swagger.rb
```diff
@@ -338,13 +346,15 @@ class Form526ControllerSwagger
             key :type, :string
           end
 
-          parameter do
+          request_body do
```
bastosmichael, line 349: I like this but make sure it renders properly on the portal's version of Swagger UI.

modules/claims_api/app/swagger/claims_api/v0/form_526_controller_swagger.rb
hinzed1127: no worries, I added that TODO as I was going through, so it wasn't previously there, just in my audit

- [ ] Add `ClaimsShow` schema as example response

modules/claims_api/app/swagger/claims_api/v1/form_526_controller_swagger.rb
hinzed1127: looks like it does ok

modules/claims_api/app/swagger/claims_api/forms/form_2122_example.json
hinzed1127: I don't think it's a problem to leave it out here. this is only used within the swagger code, and the end result is showing what's correct, so I wouldn't consider it misleading.

I'm having a hard time getting the swagger blocks to play nicely if I leave the data property in and getting everything to display properly, so maybe keeping the change but renaming this to `form_2122_example_data.json` or something to be a tad more explicit is a good middle ground?

modules/claims_api/app/swagger/claims_api/v0/form_2122_controller_swagger.rb
```diff
@@ -352,7 +360,103 @@ class Form2122ControllerSwagger
...
+                    key :'$ref', :ErrorModel
+                  end
+                end
+              end
+            end
+          end
+        end
+      end
+
+      swagger_path '/forms/2122/active' do
```
hinzed1127, line 390: this endpoint was missing from the v0 docs. from what I could tell, the logic is similar enough to v1 that the responses should be the same

modules/claims_api/app/swagger/claims_api/forms/form_2122_response_swagger.rb
```diff
@@ -148,7 +148,7 @@ class Form2122ResponseSwagger
             property :attributes do
               key :type, :object
               key :description, 'Required by JSON API standard'
-              key :required, %i[poa_code poa_first_name poa_last_name]
+              key :required, %i[serviceOrganization poaCode]
```
hinzed1127, line 151: updated this to the minimum payload properties

PR #4786, opened 2020-08-26
Education spool file region cleanup

dmccurdybah:
## Description of change
Moved states from `CENTRAL` to `EASTERN` because all submissions are now routed to only the Buffalo or Muskogee region.  Updated the comment to make clear that the empty list is required for historical reporting purposes.

This change doesn't modify any functionality but should make it clearer that there are only two possible regions for submissions.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#12818


shawkey-daniel:
I don't think this matters much, but I think it may be helpful that no one would assume it is a mistake that central is not used.

PR #4787, opened 2020-08-26
10170 drop autoloading of the lib directory

erluti:
Until this PR is merged, `config/application.rb` manually added the `/lib` directory to the autoloading path.  This is non-standard Rails configuration and has lead to some practices that make it difficult to switch over to the zeitwerk loader.  This change is to remove that code and make the needed changes to support the standard configuration.

## Description of change
This code was removed from `config/application.rb`
```
paths_name = Rails.env.development? ? 'autoload' : 'eager_load'
config.public_send("#{paths_name}_paths") << Rails.root.join('lib')
config.eager_load_paths << Rails.root.join('app')
```
Missing `require`s were added back in to include files from the lib directory.
Other changes to support zeitwerk include separating classes into their own files if there are in a shared file and separating modules from class name definitions.  `Vet360::Service::STATSD_KEY_PREFIX` had to be moved to `Vet360::Stats::STATSD_KEY_PREFIX` to prevent cyclical dependencies.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#10170


erluti:
> Looks good to me, just want to make sure that you reviewed the rake files as well. Looks like there's a few requires added in vets360.rake and connectivity.rake currently. I think I _may_ have added these, but want to be sure that there's no breaking changes to other rake tasks.

We had a different task for that, department-of-veterans-affairs/va.gov-team#10185.  But I bet you're right, they would probably break and we couldn't do it separately.  Thanks for pointing that out!

erluti:
@LindseySaari here's a PR to add in missing `require`s from the rakelib directory https://github.com/department-of-veterans-affairs/vets-api/pull/4794

LindseySaari:
> @LindseySaari here's a PR to add in missing `require`s from the rakelib directory #4794 and merge into this branch

Approved 💯 ✔️

PR #4789, opened 2020-08-27
[DRAFT] Update HLR for Lighthouse Changes

edmangimelli:
DRAFT


config/routes.rb
```diff
@@ -323,7 +324,7 @@
   end
 
   mount VAOS::Engine, at: '/vaos'
-  mount CovidResearch::Engine, at: '/covid-research'
```
annaswims, line 326: undo

spec/lib/decision_review/higher_level_review/get_contestable_issues/request_spec.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+require_relative '../../../../../lib/decision_review/request.rb'
```
annaswims, line 3:
```suggestion
require 'decision_review/request'
```

edmangimelli:
I liked a lot of things about this approach, but, alas, it was not meant to be.
💐 🎺

PR #4790, opened 2020-08-27
This translates boolean values to strings

LeakyBucket:
## Description of change
I overlooked the string requirement for QuestionValue.  This change maps true/false responses to ‘Yes’/‘No’ accordingly.

## Original issue(s)

This is a very small and self contained change.  I don't think there is really anything significant about it.

### DIFF
--- a/modules/covid_research/spec/serializers/covid_research/genisis_serializer_spec.rb
+++ b/modules/covid_research/spec/serializers/covid_research/genisis_serializer_spec.rb
@@ -27,5 +27,13 @@ RSpec.describe CovidResearch::GenisisSerializer do
     it 'translates the json payload to a list of key value pairs' do
       expect(output['FormQuestions']).to eq(expected)
     end
+
+    it 'translates true to "Yes"' do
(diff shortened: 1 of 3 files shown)


modules/covid_research/spec/serializers/covid_research/genisis_serializer_spec.rb
```diff
@@ -24,8 +32,22 @@
...
       expect(output['FormQuestions']).to eq(expected)
     end
+
+    it 'translates true to "Yes"' do
+      expect(output['FormQuestions'].first['QuestionValue']).to eq 'No'
+    end
+
+    it 'translates false to "No"' do
+      expect(output['FormQuestions'][5]['QuestionValue']).to eq 'Yes'
+    end
```
thilton-oddball, line 51: Should the Yes and No on the above tests be switched?

LeakyBucket: Doh, good catch!  Just pushed up a fix for that.

PR #4792, opened 2020-08-27
12659 add va inflection

erluti:
As part of the zeitwerk update (department-of-veterans-affairs/va.gov-team#9534) inflections are needed to support our use of acronyms in class and file names.  `VA` is obviously used in many places and without inflection our filenames would need to become `v_a_something` for `VASomething`.  Adding this inflection also discovered a place where the `humanize` method was turning text from `VA` to `va` because it wasn't inflected.

## Description of change
- `VA` inflection added to inflections file
-  classnames written as `VaSomething` become `VASomething`
- Two pdf fixtures added to support humanized `VA`
- inflect `VAR` because https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/vaos/app/services/vaos/cc_supported_sites_service.rb uses uses a response from VAOS service that uses the acronym
- patches `olive_branch` so that uninflected (deprecated) keys including `VA` are also returned.  more details here https://github.com/department-of-veterans-affairs/vets-api/pull/5133

## Original issue(s)
department-of-veterans-affairs/va.gov-team#12659


LindseySaari:
Could you add some info to the description about the olive branch patch and link to the PR you merged in?

LindseySaari:
I'm wondering if some of these changes will affect the way that errors are grouped in Sentry. For [example](http://sentry.vfs.va.gov/organizations/vsp/issues/5887/?query=is%3Aunresolved+VAFileNumberNotFound&statsPeriod=14d), will BGS::PeopleService::VaFileNumberNotFound be grouped separately from BGS::PeopleService::VAFileNumberNotFound? Maybe we looked into this already, but just wanted to check.

LindseySaari (approved):
It doesn't seem like the Sentry error groupings are of high concern so i'm going to give a 👍

erluti:
> I'm wondering if some of these changes will affect the way that errors are grouped in Sentry. For [example](http://sentry.vfs.va.gov/organizations/vsp/issues/5887/?query=is%3Aunresolved+VAFileNumberNotFound&statsPeriod=14d), will BGS::PeopleService::VaFileNumberNotFound be grouped separately from BGS::PeopleService::VAFileNumberNotFound? Maybe we looked into this already, but just wanted to check.

Chatting with Keifer he thinks errors are grouped by stack trace, so they should probably still group together

PR #4795, opened 2020-08-27
Form526 confirmation email - error handling

philherbert:
## Description of change

In order to correctly handle errors returned from the `notification-api` when sending a confirmation email for Form526ez, we have wrapped the `Notifications::Client` gem in a new service - `VaNotify::Service`, which inherits from all the proper places.

We want to use Betamocks, circuit breaking, and monitoring for the `VaNotify::Service`, but this requires that HTTP requests are made using the service's `Faraday` connection. The `Notifications::Client` gem doesn't support using custom middleware, so to achieve this, we have redefined one of the gem's internal methods, which now uses `Faraday` to make HTTP requests. See [this slack thread](https://dsva.slack.com/archives/CBU0KDSB1/p1598034961097300) for a discussion, and `VaNotify::Service#overwrite_client_networking` for the implementation.

We've also added tests for the service, which use `VCR`. New cassettes can be found in the `spec/support/vcr_cassettes/va_notify` directory.

## Original issue(s)
department-of-veterans-affairs/notification-api#86

  * Yes, a new field: `vanotify.mock`. This can vary by environment, but doesn't need to be present, as `VaNotify::Configuration` will assume that it's false by default.
  * Already covered by existing feature flag `form526_confirmation_email`.
  * Yes, in `Form526ConfirmationEmailJob#handle_errors`, we now log errors to Sentry.


erluti (approved):
LGTM
had a good chat in slack about the need a forked version of https://github.com/alphagov/notifications-ruby-client to avoid future monkey-patching and other tech debt

PR #4801, opened 2020-08-28
add dependency loading for form1010cg requests and services

kevinmirc:
`Form1010cg::Service` has more dependencies that were listed in this base branch. I've added those and also added those dependencies' dependencies, until the request spec for submitting a VA Form 10-10cg would pass.

### DIFF
--- a/spec/lib/mvi/service_spec.rb
+++ b/spec/lib/mvi/service_spec.rb
@@ -1,9 +1,7 @@
 # frozen_string_literal: true
 
 require 'rails_helper'
-require 'mvi/messages/add_person_message'
 require 'mvi/service'
-require 'mvi/responses/find_profile_response'
 
 describe MVI::Service do
   let(:user_hash) do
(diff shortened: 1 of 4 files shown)


spec/lib/mvi/service_spec.rb
```diff
@@ -1,9 +1,7 @@
 # frozen_string_literal: true
 
 require 'rails_helper'
-require 'mvi/messages/add_person_message'
```
erluti, line 4: why did these two get removed?

kevinmirc: The test suite would pass because they were required here. But when MVI::Service was used in runtime these deps were not loaded. So I moved them to the MVI Service and now these deps will be required in dev runtime and testing.

erluti: That's a good point.  I think I'll take another look at some of the other tests too.  Thanks!

kevinmirc:
Also note @erluti, when you pull from master on this base branch, `SavedClaim::CaregiversAssistanceClaims` will need a `require  pdf_fill/filler`.

I can merge if this all looks good to you.

erluti:
> I can merge if this all looks good to you.

It does!

PR #4806, opened 2020-08-28
SSOe SAML post tracker

ericbuckley:
Adding a new route to the v1 sessions controller to accept SAML tracking info and updating the SSOe SAML Post HTML form to send a request to this route immediately before submitting the form.  This should give us some data on whether some users are still having issues running the JS code in their browsers.

closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/13022


config/routes.rb
```diff
@@ -17,6 +17,7 @@
       to: 'v1/sessions#new',
       constraints: ->(request) { V1::SessionsController::REDIRECT_URLS.include?(request.path_parameters[:type]) }
   get '/v1/sessions/ssoe_logout', to: 'v1/sessions#ssoe_slo_callback'
+  get 'v1/sessions/tracker', to: 'v1/sessions#tracker'
```
jholton, line 20: Nitpick, for consistency with the other routes (even though further down in the file there are routes without it. 🤷 ):
```suggestion
  get '/v1/sessions/tracker', to: 'v1/sessions#tracker'
```

lib/saml/templates/sso_post_form.html.erb
```diff
@@ -26,7 +26,15 @@
...
     (function() {
-      document.getElementById("saml-form").submit();
+      var form = document.getElementById("saml-form");
+      var req = new XMLHttpRequest();
+      var qs = "saml_uuid=" + encodeURIComponent("<%= saml_uuid %>") +
+                "&authn=" + encodeURIComponent("<%= authn %>") +
+                "&type=" + encodeURIComponent("<%= type %>");
+      var url = location.origin + "/v1/sessions/tracker?" + qs;
+      req.open("GET", url);
+      req.send();
```
jholton, line 36: If a user is having trouble with the form submit in their browser, what are the chances that this JS won't work as expected either?

ericbuckley: @jholton do you mean how often do I think the XHR request will fire but the form submission will fail?

jholton: Yeah, that's what I meant. 

I guess my question is: how does this get us insight into if/which users are having trouble submitting the form?
Is it comparing logging of the SAML UUID in [`render_login`](https://github.com/department-of-veterans-affairs/vets-api/blob/bb9a54929eded71c34848f3844994b92fcc9b7b3/app/controllers/v1/sessions_controller.rb#L188) to the logging of SAML UUID in [`tracker`](https://github.com/department-of-veterans-affairs/vets-api/blob/bb9a54929eded71c34848f3844994b92fcc9b7b3/app/controllers/v1/sessions_controller.rb#L78)?

Should/can we log user agent info too to determine which browsers are having trouble?

ericbuckley: Right now we are assuming every `STATSD_SSO_SAMLREQUEST_KEY` issued is a request that should be showing up at eauth.va.gov.  However if there are any issues with this form loading, not neccessarlily submitting, the XHR request will fail to fire and the subsequent `STATSD_SSO_SAMLTRACKER_KEY` won't increment.

Basically if the `STATSD_SSO_SAMLREQUEST_KEY` doesn't match `STATSD_SSO_SAMLTRACKER_KEY` then we have an issue with the form loading, if they do match it just confirms that the problem is most likely downstream of us.

jholton (approved):

erluti (approved):
LGTM
I would like it if there was a comment with the tracker key or on the tracker action to clarify this is for tracking an issue, because presumably we'd want to roll this back this change once the problem is diagnosed.  Along those lines, could you be sure your team has a clean-up ticket referencing this PR so you can eventually roll it back?

ericbuckley:
> LGTM
> I would like it if there was a comment with the tracker key or on the tracker action to clarify this is for tracking an issue, because presumably we'd want to roll this back this change once the problem is diagnosed. Along those lines, could you be sure your team has a clean-up ticket referencing this PR so you can eventually roll it back?

@erluti I just created a tech debt GH issue to track the rollback.  https://github.com/department-of-veterans-affairs/va.gov-team/issues/13297

annaswims:
I believe this change prevents me from logging in unless I disable my ad blocker.  I'm wondering if you could do the same thing without the word `tracker`

Screen Shot 2020-09-17 at 10 51 08 AM

ericbuckley:
yeah of course, is there a setting in AdBlock that indicates any url with the word "track" in it is blocked?

annaswims:
AdBlock plus lets me subscribe to multiple lists, not sure which one is the issue though
My settings:

Screen Shot 2020-09-17 at 12 00 17 PM

[the lists](https://adblockplus.org/en/subscriptions)

PR #4814, opened 2020-09-01
[686] Add end date to 686 divorce

digitaldrk:
## Description of change
We are in a testing phase with our friends at BGS. They informed us we are not sending date of divorce and that we need to send it as an `end_date`. This PR does that.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#12425


lib/bgs/dependents.rb
```diff
@@ -88,6 +88,7 @@ def report_divorce
           divorce_state: formatted_info['divorce_state'],
           divorce_city: formatted_info['divorce_city'],
           divorce_country: formatted_info['divorce_country'],
+          end_date: formatted_info['end_date'],
```
erluti, line 91: should the expected hash in the spec include an end_date now?

digitaldrk: Yep, I'm glad you mentioned that bc it made me realize we needed another attribute. Added 👍

kathleencrawford (approved):
LGTM 👍

erluti (approved):
LGTM

PR #4817, opened 2020-09-01
allow unauthenticated users to access /v1/sessions/slo/new

ericbuckley:
While it doesn't make sense for an unauthenticated user to make a "logout request", the UX of this is sub-par as they will see an unauthorized 401 JSON response.  The need to fix [double logout problems](https://github.com/department-of-veterans-affairs/va.gov-team/issues/13054) is still needed, but in the meantime we can improve the UX for the user by just redirecting them back to the home page.

closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/12731


jholton (approved):

spec/controllers/v1/sessions_controller_spec.rb
```diff
@@ -265,7 +265,7 @@ def expect_logger_msg(level, msg)
       end
 
       context 'routes requiring auth' do
-        %w[mfa verify slo].each do |type|
+        %w[mfa verify].each do |type|
```
erluti, line 268: I think we're reducing test coverage here.  Can we also add a test for the `slo` redirect you've added? would it be similar to the test `it 'redirects on callback from external logout'`?

ericbuckley: yes it would, ill add a new test case for unauthenticated users issuing an `slo`

ericbuckley: @erluti just added https://github.com/department-of-veterans-affairs/vets-api/pull/4817/commits/a4521982ecc330a5adec59d20969cac134ce7a94

erluti (approved):
LGTM

PR #4821, opened 2020-09-01
health service without git commit history because of problem jenkins …

massrb:
…build with danger gem

Build without git commit history because mystery jenkins job has failed the build with a crash in danger gem

## Original issue(s)


jbritt1:
sorry, accidentally marked ready

PR #4823, opened 2020-09-02
Update HLR for Lighthouse Changes

edmangimelli:
Companion PRs:
https://github.com/department-of-veterans-affairs/vets-json-schema/pull/496
https://github.com/department-of-veterans-affairs/vets-website/pull/14200


lib/decision_review/service.rb
```diff
@@ -60,23 +46,40 @@ def get_higher_level_reviews_intake_status(uuid)
...
+    end
+
+    def get_contestable_issues_headers(user)
+      raise Common::Exceptions::Forbidden unless user.ssn
+
       {
-        'x-va-ssn' => user.ssn,
-        'x-va-receipt-date' => Time.current.strftime('%Y-%m-%d')
+        'X-VA-SSN' => user.ssn,
+        'X-VA-Receipt-Date' => Time.zone.now.strftime('%F')
```
annaswims, line 86: I know `'%F'` and `'%Y-%m-%d'` are functionally equivalent and that  `'%F'` is more concise, but I still prefer `'%Y-%m-%d'` because I can guess what it means without looking it up.

spec/lib/decision_review/service_spec.rb
```diff
@@ -9,231 +9,111 @@
...
-                }
-              }
-            ]
-          }
-
-          response = subject.post_higher_level_reviews(request.to_json)
-          expect(response).to be_accepted
-          expect(response).to be_an DecisionReview::Responses::Response
+    subject { described_class.new.post_higher_level_reviews(body: body.to_json, user: user) }
+    let(:body) { VetsJsonSchema::EXAMPLES['HLR-CREATE-REQUEST-BODY'] }
```
annaswims, line 14: YES! I really like using shared examples between BE & FE!

lib/decision_review/service.rb
```diff
@@ -29,24 +29,10 @@ class Service < Common::Client::Base
...
-    end
-
-    ##
-    # Retrieve a Higher Level Review intake status.
-    #
-    # @param uuid [uuid] The intake uuid provided from the response of creating a new Higher Level Review
-    # @return [DecisionReview::Responses::Response] Response object that includes the body,
-    #                                               status, and schema avalidations.
-    #
-    def get_higher_level_reviews_intake_status(uuid)
```
annaswims, line 46: v0/intake_statuses_controller.rb still calls `get_higher_level_reviews_intake_status`

app/controllers/v0/higher_level_reviews/contestable_issues_controller.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module V0
+  module HigherLevelReviews
+    class ContestableIssuesController < AppealsBaseController
+      def index
+        issues = decision_review_service.get_higher_level_review_contestable_issues(
+          user: current_user, benefit_type: params[:benefit_type]
```
annaswims, line 8: do we want to have an "allow list" for `benefit_type`? since we interpolate it later? 🤷‍♀️

lib/decision_review/service.rb
```diff
@@ -60,23 +46,44 @@ def get_higher_level_reviews_intake_status(uuid)
     def get_higher_level_reviews(uuid)
       with_monitoring_and_error_handling do
         raw_response = perform(:get, "higher_level_reviews/#{uuid}", nil)
-        DecisionReview::Responses::Response.new(raw_response.status, raw_response.body, 'review')
+        DecisionReview::Responses::Response.new(raw_response.status, raw_response.body, 'HLR-SHOW-RESPONSE-200')
```
annaswims, line 49: was there a reason you chose all caps for this string?  Seems unusual to me.

app/controllers/v0/higher_level_reviews/contestable_issues_controller.rb
edmangimelli: there's a schema for that :D

but Lighthouse will return a 422 if an invalid benefit_type is used. the service will treat that response like all _bad input_ responses. i felt like that was reasonable. should we treat benefit_type differently?

lib/decision_review/service.rb
edmangimelli: good catch. i just removed the controller 👍 
(it's no longer an endpoint for Lighthouse)

lib/decision_review/service.rb
edmangimelli: i like it 👍

spec/lib/decision_review/service_spec.rb
edmangimelli: https://emojis.slackmojis.com/emojis/images/1547582922/5197/party_blob.gif

lib/decision_review/service.rb
edmangimelli: the way vets_json_schema is set up is that it automatically upcases schema names.
yeah, i don't like it either

annaswims:
https://department-of-veterans-affairs.github.io/va-digital-services-platform-docs/api-reference/#/higher_level_reviews/showHigherLevelReview docs

app/swagger/swagger/requests/appeals/appeals.rb
```diff
@@ -18,125 +18,66 @@ class Appeals
...
               key :in, :body
-              key :description, 'Higher level review raw request data'
               key :required, true
-
-              schema do
-                key :'$ref', :HigherLevelReviewRequest
-              end
-            end
-
-            response 202 do
```
annaswims, line 95: does this mean that the only responses are 200 or 422?

edmangimelli: for the create? pretty much

annaswims:
TIL: 
 JSON.parse("") =>
JSON::ParserError: 784: unexpected token at ''

edmangimelli:
> TIL:
> JSON.parse("") =>
> JSON::ParserError: 784: unexpected token at ''

sorry, just now looking at this.
yeah, i've run into this problem before. can we go with what I have now and then in a separate PR i'll do what I did for Lighthouse?
https://github.com/department-of-veterans-affairs/vets-api/blob/5d337c7dc821345b9cba9ea29eab84d09184ca9e/modules/appeals_api/app/controllers/appeals_api/concerns/json_format_validation.rb

annaswims:
LGTM

annaswims (approved):
LGTM

lib/decision_review/service.rb
```diff
@@ -22,61 +18,90 @@ class Service < Common::Client::Base
 
     STATSD_KEY_PREFIX = 'api.decision_review'
 
+    HLR_CREATE_RESPONSE_SCHEMA = VetsJsonSchema::SCHEMAS['HLR-CREATE-RESPONSE-200']
+    raise unless HLR_CREATE_RESPONSE_SCHEMA.is_a? Hash
```
erluti, line 22: This is concerning because wouldn't it raise this exception when the DecisionReview Service is loaded?  It seems like it's checking that VetsJsonSchema has these required values and I think runtime is the wrong time to alert on that mismatch.  I think it would better serve as a test, so that way the pipeline would break if somebody tried to introduce a VetsJsonSchema that is missing these required schemas.  The test would also be more descriptive than this generic exception.

lib/decision_review/service.rb
```diff
@@ -89,44 +114,49 @@ def with_monitoring_and_error_handling
...
+      raise Common::Exceptions::SchemaValidationErrors, remove_pii_from_json_schemer_errors(errors)
+    end
+
+    def raise_schema_error_unless_200_status(status)
+      return if status == 200
+
+      raise Common::Exceptions::SchemaValidationErrors, ["expecting 200 status received #{status}"]
+    end
+
+    def remove_pii_from_json_schemer_errors(errors)
```
erluti, line 158: is this supposed to be `schemer`?

spec/lib/decision_review/service_spec.rb
```diff
@@ -9,232 +9,121 @@
...
-            'api.decision_review.post_higher_level_reviews.fail', tags: [
-              'error:CommonClientErrorsClientError', 'status:400'
-            ]
-          )
-          expect(StatsD).to receive(:increment).once.with('api.decision_review.post_higher_level_reviews.total')
-          expect { subject.post_higher_level_reviews({}) }.to raise_error(DecisionReview::ServiceException)
-        end
+    context '422 response' do
+      let(:body) { {} }
+      let(:exception) do
```
erluti, line 30: I found this use of let to convert the subject into an exception a bit confusing.  So I dug into a rspec matchers for expecting to raise_error, and found you can also check attributes.  So I think this context (and the other "bad response contexts) would be clearer if they were something like this:
```
context '422 response' do
      let(:body) { {} }

      it 'throws a DR_422 exception' do
        VCR.use_cassette('decision_review/HLR-CREATE-RESPONSE-422') do
          expect { subject }.to raise_error(
            an_instance_of(DecisionReview::ServiceException).and having_attributes(key: 'DR_422')
          )
        end
      end
    end
```

erluti (changes requested):
These are some great changes!  I really like this overall with a couple questions/suggestions.
Marked `request changes` because the runtime exception is concerning enough I want to hear more on that before I can approve

lib/decision_review/service.rb
edmangimelli: yep

lib/decision_review/service.rb
edmangimelli: tests would fail too, fwiw.
if these values aren't present, an exception will be thrown _somewhere_, and, if that's the situation, might as well make it obvious (hence raise)

spec/lib/decision_review/service_spec.rb
edmangimelli: this is pretty close to how i originally had it written, but, for some reason, it doesn't work (not entirely sure why)

edmangimelli: the raise_error bit works! i must have typoed originally, but looks like you can't use that matcher (have_attributes) when passing a block to expect.
considering that, i'm tempted to leave it as is (so i can easily test the exception)

lib/decision_review/service.rb
erluti: I think the problem is fundamentally that raising an exception here is this is testing a dependency.  Truly there should just be a test in VetsJsonSchema and we wouldn't need a test here.  But if this dependency needs to be tested, I think the test belongs in the test suite.
You are right though, that if these values come back nil the error message is enigmatic.  But again, I think a test solves this in that the description/context/name will be specific about what's wrong.

spec/lib/decision_review/service_spec.rb
erluti: I don't understand your comment.  The code I wrote above I tested on your branch and it worked.  It also covers everything you are testing about the exception.

lib/decision_review/service.rb
edmangimelli: >should just be a test in VetsJsonSchema and we wouldn't need a test here

i don't believe that's the case. this is catching an integration problem. a test might be passing in vets-json-schema, but, if vets-api is pointing to the wrong version of vets-json-schema (for instance), a schema could be missing (this has already happened to me a couple times)

spec/lib/decision_review/service_spec.rb
edmangimelli: Oh whoops! I was trying different things before i tested your snippet, and should have copied and pasted when i finally did. I was doing this:
```
    context '422 response' do
      let(:benefit_type) { 'apricot' }

      it 'throws a DR_422 exception' do
        VCR.use_cassette('decision_review/HLR-GET-CONTESTABLE-ISSUES-RESPONSE-422') do
          expect { subject }.to raise_error(DecisionReview::ServiceException).and have_attribute(key: 'DR_422')
        end
      end
    end
```
(for the bottommost 422 context (get contestable issues))
I've not seen 
```
raise_error(
  an_instance_of( x)
)
```
before
i'm used to:
```
raise_error SomeClass
```
nifty!

edmangimelli: it's beautiful now 😍

lib/decision_review/service.rb
erluti: Wouldn't the missing schema issue be related to having an old version of VetsJsonSchema?  There's no risk of advancing the schema and this disappearing, right?
I would still argue that integration is a category of testing, and would prefer an explicit test over a RuntimeError

edmangimelli: >There's no risk of advancing the schema and this disappearing, right?

not true. I made a schema go bye-bye recently (rename)

edmangimelli: what about a test _and_ what I have above?

scenario:

a schema gets renamed in vets-json-schema.
i think i've changed all instances of the name (including in my tests which specifically test for those schemas' presence)

i miss it in my controller.
now i'm getting a really strange backtrace.
15 min later i realize it's because i missed a rename

^ this happened. if a schema is nil and gets passed into JsonSchemer, the backtrace is super confusing (i might submit a PR to JsonSchemer actually. it looked like an oversight)

i could have used `fetch` and accomplished roughly the same thing, but i thought it'd be nice to ensure that it was a hash too

erluti:
```
    context 'schema constants' do
      it 'has HLR_CREATE_RESPONSE_SCHEMA hash' do
        expect(described_class::HLR_CREATE_RESPONSE_SCHEMA).to be_a Hash
      end
      it 'has HLR_SHOW_RESPONSE_SCHEMA hash' do
        expect(described_class::HLR_SHOW_RESPONSE_SCHEMA).to be_a Hash
      end
      it 'has HLR_GET_CONTESTABLE_ISSUES_RESPONSE_SCHEMA hash' do
        expect(described_class::HLR_GET_CONTESTABLE_ISSUES_RESPONSE_SCHEMA).to be_a Hash
      end
    end
```
I put these tests in `spec/lib/decision_review/service_spec.rb` and then set `HLR_SHOW_RESPONSE_SCHEMA` to nil and got two failures.  One the enigmatic one from JsonSchemer, but also one that makes it clear that there's an issue with the schema constant.
```
  1) DecisionReview::Service#get_higher_level_review 200 response returns a properly formatted 200 response
     Failure/Error: errors = JSONSchemer.schema(schema).validate(json).to_a
  2) DecisionReview::Service#create_higher_level_review schema constants has HLR_SHOW_RESPONSE_SCHEMA hash
     Failure/Error: expect(described_class::HLR_SHOW_RESPONSE_SCHEMA).to be_a Hash
       expected nil to be a kind of Hash
```

I think these tests give us the clear message that those constants have a bad value, which is unexpected/unacceptable, and would point the developer to the right point in the code.  They aren't susceptible to schema name changes, because we're checking the constant.  And if they rename the constant those tests would also fail because the constants name was changed.
My main concern about raising an exception in the main file here is that it fails when the file is loading, which would break the vets-api deploy. I think the right way to make sure changes aren't breaking features is through a test.

edmangimelli: i suppose i don't necessarily want to check the constants defined in the controller, I want to check the contents of the hash `VetsJsonSchema::SCHEMAS`

edmangimelli: made some updates! let me know 👍

erluti: Much better!  Thanks for looking at this again!

erluti (approved):

PR #4829, opened 2020-09-02
Gracefully handle PPMS Provider info response errors

darkbushido:
## Description of change
When searching for a provider we need to make an additional call per
provider so we know which name to display. rarely that additional call
fails causing the entire request to fail.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#13132

Adding Sentry logging so we can track when the additional request(s) fails.


app/controllers/v1/facilities/ccp_controller.rb
```diff
@@ -67,8 +67,15 @@ def ppms_search
 
   def provider_search(options = {})
     api.provider_locator(ppms_params.merge(options)).map do |provider|
-      prov_info = api.provider_info(provider['ProviderIdentifier'])
-      provider.add_details(prov_info)
+      begin
+        prov_info = api.provider_info(provider['ProviderIdentifier'])
+        provider.add_details(prov_info)
+      rescue => e
+        if Rails.env.production?
```
annaswims, line 74: you don't need the `Rails.env.production?` bit.  `log_exception_to_sentry` checks for `Settings.sentry.dsn.present?`

erluti (approved):
LGTM

PR #4831, opened 2020-09-03
Add support for controller in external requests

kfrz:
## Description of change
This PR adds support for calling controller labels in breakers metrics exports using the `Kernel#caller` method

Also automates instrumentation of all Faraday requests using `:instrumentation` middleware.

No tests yet, it's pretty bad and needs some adjusting.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#15482

### DIFF
--- a/config/initializers/statsd.rb
+++ b/config/initializers/statsd.rb
@@ -165,7 +165,6 @@ end
 StatsD.increment('shared.sidekiq.default.Facilities_InitializingErrorMetric.error', 0)
 
 ActiveSupport::Notifications.subscribe('facilities.ppms.request.faraday') do |_, start_time, end_time, _, payload|
-  duration = end_time - start_time
   measurement = case payload[:url].path
                 when /ProviderLocator/
                   'facilities.ppms.provider_locator'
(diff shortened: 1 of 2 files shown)


config/initializers/statsd.rb
```diff
@@ -183,3 +182,13 @@
...
   StatsD.measure('facilities.lighthouse', duration, tags: ['facilities.lighthouse'])
 end
+
+require 'active_support/notifications'
+
+ActiveSupport::Notifications.subscribe('request.faraday') do |name, start_time, end_time, _, env|
+  url = env[:url]
+  duration = end_time - start_time
+  http_method = env[:method].to_s.upcase
+  StatsD.measure('api.external.request.duration', duration, tags: ["host:#{url.host}"])
```
annaswims, line 192: this metric is pretty much a duplicate of job:breakers_http_request_duration:sum, that's displayed on http://grafana.vfs.va.gov/d/000000050/backend-service-report?orgId=1 I'd think. t

lib/breakers/statsd_plugin.rb
```diff
@@ -1,10 +1,30 @@
 # frozen_string_literal: true
 
+module CallChain
+  def self.caller_method(depth=1)
+    parse_caller(caller(depth+1).first).last
+  end
+
+  private
+
+  # Copied from ActionMailer
```
kfrz, line 10: Cool... but where in `ActionMailer`?

kfrz:
Might be able to use (already included) https://github.com/steveklabnik/request_store

config/initializers/statsd.rb
kfrz: Yes but this would actually automatically instrument any `faraday` requests with this detail rather than have to follow the facilities pattern as seen in lines above 185

PR #4832, opened 2020-09-03
Fix time-dependent test

annaswims:
## Description of change
I wrote a time dependent test.
spec/models/saved_claim/disability_compensation/form526_all_claim_bdd_spec.rb:23 will fail if it's not more than 90 days before Dec 1, 2020

http://jenkins.vfs.va.gov/blue/organizations/jenkins/testing%2Fvets-api/detail/PR-4827/2/pipeline


erluti (approved):
thanks for the quick fix

PR #4833, opened 2020-09-03
add require to facilities response

erluti:
Should resolve this error http://sentry.vfs.va.gov/vets-gov/platform-api-staging/issues/197187/


LindseySaari (approved):
👍

PR #4834, opened 2020-09-03
add missing requires for bgs and emis

LindseySaari:
## Description of change
After a recent change to remove autoloading of the lib directory, some requires were missing for bgs and emis lib directory files.

### DIFF
--- a/lib/emis/service.rb
+++ b/lib/emis/service.rb
@@ -1,12 +1,12 @@
 # frozen_string_literal: true
 
 require 'common/client/base'
-require 'emis/configuration'
-require 'emis/messages/edipi_or_icn_message'
-require 'emis/responses'
 require 'common/client/concerns/log_as_warning_helpers'
 require 'common/client/middleware/request/soap_headers'
 require 'common/client/middleware/response/soap_parser'
+require 'emis/configuration'
+require 'emis/messages/edipi_or_icn_message'
+require 'emis/responses'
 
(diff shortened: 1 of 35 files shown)


spec/lib/bgs/benefit_claim_spec.rb
```diff
@@ -2,6 +2,7 @@
 
 require 'rails_helper'
 require 'bgs/benefit_claim'
+require 'bgs/service'
```
erluti, line 5: so one thing I discovered is that a lot of these are showing up because the requires are actually missing from the code and not from the test.  `lib/bgs/benefit_claim` uses `BGS::Service` but doesn't have the `require` in place to use it.  So this should go there.

erluti: The other specs might have a similar issue

lib/emis/service.rb
```diff
@@ -1,12 +1,12 @@
 # frozen_string_literal: true
 
 require 'common/client/base'
-require 'emis/configuration'
-require 'emis/messages/edipi_or_icn_message'
-require 'emis/responses'
 require 'common/client/concerns/log_as_warning_helpers'
 require 'common/client/middleware/request/soap_headers'
 require 'common/client/middleware/response/soap_parser'
+require 'emis/configuration'
```
erluti, line 7: moved these for organizational/clarity

PR #4839, opened 2020-09-03
second attempt at PR 1.0 for health quest service; break commits up

massrb:
## Description of change

This is the first in a series of pull requests for the Health Quest service for the health experience team. There are 2000 lines of code so it is being attempted with multiple pull requests. The health quest service has code copied from VAOS in order to avoid impacting VAOS metrics. Kam Karshenas had recommended this approach. 

VAOS uses a setting in settings.yml like this
```
va_mobile:
  url: 'https://veteran.apps.va.gov'
  mock: false
  timeout: 15
```

The health quest service uses the same type of setting:

```
hqva_mobile:
  url: 'https://veteran.apps.va.gov'
  mock: false
  timeout: 15
```

These settings have no effect in a dev environment. For the other environments, the hqva_mobile setting will need to be set up.

## Original issue(s)


erluti:
It seems like the classes in /modules could be follow up PRs.  If this were stripped down to the module-generation/engine code, the SessionStore, Configuration, JwtWrapper could all be their own PRs with the Service following after those.  The goal is to have PRs that do "one thing"

modules/health_quest/spec/services/user_service_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+        it 'returns the session token' do
+          VCR.use_cassette('health_quest/users/post_session') do
+            session_token = subject.session(user)
+            expect(session_token).to be_a(String)
+          end
+        end
+
+        it 'makes a call out to the the health_quest user service once' do
+          VCR.use_cassette('health_quest/users/post_session') do
+            # LJG NOTE: this code was copied from VAOS and it does not fail rubocop, not sure why
```
erluti, line 66: there's an exception here https://github.com/department-of-veterans-affairs/vets-api/blob/960e5439eda33d813122d02ca934e8ce07538e42/.rubocop.yml#L317 but I think putting the exception in the code is better.  Remove your comment here and maybe add an issue to refactor this later.

modules/health_quest/app/services/health_quest/base_service.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module HealthQuest
+  class BaseService < Common::Client::Base
+    include Common::Client::Concerns::Monitoring
+    include SentryLogging
+
+    # LJG NOTE: statsd key for health quest different than VAOS
```
erluti, line 8: what does `LJG NOTE` mean?

modules/health_quest/app/services/health_quest/middleware/health_quest_logging.rb
```diff
@@ -0,0 +1,90 @@
...
+      # @return [Faraday::Env]
+      def call(env)
+        start_time = Time.current
+
+        @app.call(env).on_complete do |response_env|
+          log_tags = {
+            jti: jti(env, response_env),
+            status: response_env.status,
+            duration: Time.current - start_time,
+            # service_name: service_name || 'VAOS Generic', # Need to figure out a clean way to do this with headers
```
erluti, line 33: is this incomplete?

modules/health_quest/app/services/health_quest/user_service.rb
```diff
@@ -0,0 +1,100 @@
+# frozen_string_literal: true
+
+Faraday::Middleware.register_middleware health_quest_logging: HealthQuest::Middleware::HealthQuestLogging
+
+module HealthQuest
+  class UserService < HealthQuest::BaseService
```
erluti, line 6: I don't see the UserService being used anywhere.  Is this what this PR is adding?

erluti (commented):
I have some questions about this

modules/health_quest/spec/services/user_service_spec.rb
massrb: ok, I will leave the rubocop disable and remove my comment. Thanks. I started a different PR by the way

modules/health_quest/app/services/health_quest/base_service.rb
massrb: It's my initials. I don't have to put these comments in the code with my initials, but it makes it easy for me to grep these notes for my initials.

modules/health_quest/app/services/health_quest/middleware/health_quest_logging.rb
massrb: This code was copied from VAOS as Kam Karshenas had recommended copying the code despite that it is not ideal to do that. That same comment is also in the VAOS code

modules/health_quest/app/services/health_quest/user_service.rb
massrb: Yes, the whole set of changes would be 2084 lines or so, but I cut it down to 800+ lines and now I have another PR that is around 325 lines so I have to remove some things

modules/health_quest/app/services/health_quest/base_service.rb
erluti: I think it's better to keep notes like this out of the code when we can, I think everything we merge should be meaningful to everyone looking at the code

modules/health_quest/spec/services/user_service_spec.rb
erluti: Which PR do you want to work with moving forward?  Let's close the one we're not using

modules/health_quest/app/services/health_quest/base_service.rb
massrb: I removed that comment

modules/health_quest/spec/services/user_service_spec.rb
massrb: I closed the other PR

modules/health_quest/app/services/health_quest/middleware/health_quest_logging.rb
erluti: I think that's the case and point for striking commented code out.  I'd go ahead and remove this line since VAOS clearly never thought it was worth "figuring out a clean way to do this with headers."  Otherwise somebody else is going to cut and paste it and have the same question again, lol.

erluti (approved):
LGTM, I'd just like to scrap that legacy comment

PR #4843, opened 2020-09-03
statsd metrics with tags not matching those initialized restart the StatsD agent

ericbuckley:
If we increment a metric using tags that are different from the values used during initialization, the [statsd exporter raises an exception and triggers a restart](https://console.amazonaws-us-gov.com/cloudwatch/home?region=us-gov-west-1#logEventViewer:group=dsva-vagov-prod/var/log/prometheus/statsd_exporter.log;stream=deployment-vagov-prod-vets-api-server-20200903-190131-10.247.33.111;filter=restart;start=2020-09-02T19:54:38Z).  This will through off any prometheus queries that are attempting to calculate the rate of change over time, if triggered often.

[Slack thread that identified the issue](https://dsva.slack.com/archives/C30LCU8S3/p1599164567086300)

This PR addresses 3 problematic statsd metrics issues in the session controllers:
- `STATSD_SSO_CALLBACK_FAILED_KEY`: UserAttributeErrors raises would send the symbol code in, and thus needs to be prefixed with "error:"
- `STATSD_LOGIN_SHARED_COOKIE`: removing entirely because the tags didn't line up and its not being used for any queries
- `STATSD_SSO_SAMLREQUEST_KEY`: context was always being set to `nil` because it was referencing a different instance of `url_service`

### DIFF
--- a/app/controllers/v0/sessions_controller.rb
+++ b/app/controllers/v0/sessions_controller.rb
@@ -206,7 +206,7 @@ module V0
                          tags: ['status:failure',
                                 "context:#{saml_response.authn_context}",
                                 VERSION_TAG])
-        StatsD.increment(STATSD_SSO_CALLBACK_FAILED_KEY, tags: [failure_tag, VERSION_TAG])
+        StatsD.increment(STATSD_SSO_CALLBACK_FAILED_KEY, tags: ["error:#{failure_tag}", VERSION_TAG])
       when :failed_unknown
         StatsD.increment(STATSD_SSO_CALLBACK_KEY,
                          tags: ['status:failure', 'context:unknown', VERSION_TAG])
--- a/app/controllers/v0/sessions_controller.rb
+++ b/app/controllers/v0/sessions_controller.rb
@@ -206,7 +206,7 @@ module V0
                          tags: ['status:failure',
                                 "context:#{saml_response.authn_context}",
                                 VERSION_TAG])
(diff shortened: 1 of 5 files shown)


omgitsbillryan (approved):

app/controllers/v0/sessions_controller.rb
```diff
@@ -34,15 +33,16 @@ def new
 
       StatsD.increment(STATSD_SSO_NEW_KEY, tags: ["context:#{type}", VERSION_TAG])
       Rails.logger.info("SSO_NEW_KEY, tags: #{["context:#{type}", VERSION_TAG]}")
-      url = url_service.send("#{type}_url")
+      helper = url_service
```
erluti, line 36: the url_service method could instead be something like this so it can be used multiple times in an action without needing to be assigned to a variable
```
def url_service(previous_saml_uuid = nil)
      @url_service ||= SAML::URLService.new(saml_settings,
                           session: @session_object,
                           user: current_user,
                           params: params,
                           previous_saml_uuid: previous_saml_uuid)
    end
```

erluti (approved):
LGTM, only a minor suggestion and don't want to block this

app/controllers/v0/sessions_controller.rb
ericbuckley: thats a brilliant idea!  I just created this issue to make that change in the future https://github.com/department-of-veterans-affairs/va.gov-team/issues/13231

PR #4846, opened 2020-09-03
restore autoload lib code

erluti:
we want to rollback this change for the next build.  This PR just targets the code that manages autoloading the lib directory and restores that so /lib is autoloaded again.  The errors seen were related missing `requires`.


f1337 (approved):
Ship it! :shipit:
Guinea pigs let out of a pen and running like lemmings

PR #4847, opened 2020-09-03
rollback to when lib directory was auto-loaded

erluti:
revert of https://github.com/department-of-veterans-affairs/vets-api/pull/4787

rollback so this change isn't applied over the weekend.  There have been changes merged based on this work, and most of this work will have no impact on anything with explicit requires.

compare to https://github.com/department-of-veterans-affairs/vets-api/pull/4846 which doesn't remove:
* added requires
* separating of classes into unique files
* changes to prevent Vet360::Service::STATSD_KEY_PREFIX from creating a cyclical dependency


erluti:
Closed in favor of approach in PR #4846

PR #4848, opened 2020-09-03
#12967 - adding additional attributes to the address model

saneshark:
## Description of change
Updated the vet360 address model to include additional attributes for geolocation data needed as part of VAOS. This information exists in the vet360 contact information but was not getting passed up in the user model as part of the vet360_contact_information object. 

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/12967

The user object is effectively being changed to include 5 additional attributes for each of several potential addresses that could exist in the the vet360_contact_info addresses array.

I've tested locally that the new attributes come across as part of the user profile. I've additionally updated the address_spec and address factories to include these 5 needed attributes. I am surprised that there is no schema validation that is failing as part of my changes. Writing a schema for the user object though and all of its possible variations will be very challenging to do as part of this relatively simple PR. That's going to be an exercise for VSP to consider doing I think. 😂 

Also I am not sure which if any swagger docs need to be updated to include this additional set of attributes as being part of the address object in the future. Please let me know and I'd be happy to update those as needed.

NOTE ALSO: person.yml is the only VCR Cassette I could find that has these attributes. I am guessing the VCR cassettes previously recorded might be out of date since those attributes don't always exist. As part of my specs passing though I did not need to test against any mock data which is also a little bit surprising to me. :/


spec/lib/vet360/models/address_spec.rb
```diff
@@ -30,43 +40,43 @@
 
   describe 'validation' do
     context 'for any type of address' do
-      it 'address_pou is requred', :aggregate_failures do
+      it 'address_pou is requred' do
```
erluti, line 43: did you consider moving `:aggregate_failures` to the validation block instead of to the whole test suite?

erluti (commented):
I think the swagger you are looking for is https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/swagger/swagger/schemas/vet360/address.rb

grossb1:
Removed references to `bad_address_indicator`, it wasn't being returned in the `get_person` call

saneshark (commented):
@lihanli I was also one of the authors so I can't formally approve, but this LGTM.

lib/vet360/models/base_address.rb
```diff
@@ -36,8 +36,12 @@ class BaseAddress < Base
       attribute :created_at, Common::ISO8601Time
       attribute :effective_end_date, Common::ISO8601Time
       attribute :effective_start_date, Common::ISO8601Time
+      attribute :geocode_date, Common::ISO8601Time
+      attribute :geocode_precision, Integer
```
lihanli, line 40: how about making this field a string since we receive it as a string from their api and it could contain decimal places?

erluti:
I linked the swagger earlier, was anything going to be done with that?

lihanli:
i think `mailing_address` and `residential_address` needs to be updated here https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/swagger/swagger/schemas/vet360/contact_information.rb

lib/vet360/models/base_address.rb
saneshark: would float be better?

lihanli: float should work too

lihanli (approved):

saneshark:
> I linked the swagger earlier, was anything going to be done with that?

@erluti the swagger file you referenced is for address change requests. I've gone ahead and updated the one for contact-information as returned in the profile.

erluti (approved):

PR #4850, opened 2020-09-04
don't issue saml request tracking on logout

ericbuckley:
Issuing a SAML Request tracker on logout, `/sessions/slo/new`, is [creating more problems](https://console.amazonaws-us-gov.com/cloudwatch/home?region=us-gov-west-1#logs-insights:queryDetail=~(end~0~start~-3600~timeType~'RELATIVE~unit~'seconds~editorString~'filter*20*40message*20like*20*2fMalformed*20or*20empty*2f*0a*7c*20filter*20*40message*20like*20*2fslo*2f~isLiveTail~false~queryId~'2525f601-84db-4763-9366-9f2da1e23767~source~(~'dsva-vagov-prod*2fvar*2flog*2fprometheus*2fstatsd_exporter.log))) because the authn context is always `nil`.  That's expected, because to log a user out we don't need to issue a SAML Request.


jholton (approved):

patrickvinograd (approved):
LGTM, but what is the upshot of this for metrics - will we have any metrics for logout events with this change? 

I'm fine with merging this as is but would consider whether we want a separates statsd event that tracks logout events.

ericbuckley:
> LGTM, but what is the upshot of this for metrics - will we have any metrics for logout events with this change?
> 
> I'm fine with merging this as is but would consider whether we want a separates statsd event that tracks logout events.

The SAML Request/Response stats are going to be used to compare response rates with v1, to help identify which legs of the authentication workflow is seeing the largest decline.  For logout, we don't send a SAML request, so this stat doesn't make sense.

We are tracking new stats for logout, so if we want to measure logout success rates we could create a new stat (similar to LOGIN_SUCCESS) and add it i the saml logout callback function.

LindseySaari (approved):

PR #4852, opened 2020-09-04
Add missing service requires

erluti:
one of the most common missing `require` was references to the service in lib.  Searched for files referencing `::Service` and made sure they had the appropriate references.  Ran into some other missing requires related to common and configurations.


LindseySaari (approved):

erluti:
This PR isn't needed until https://github.com/department-of-veterans-affairs/vets-api/pull/4846 is reverted and lib-loading is no longer auto/eager by default

erluti:
lookout for incoming PRs with `HealthQuest::AppointmentService`

PR #4856, opened 2020-09-04
Strip colons from statsD tags

annaswims:
## Description of change
statsd tags use ':' as a delimiter, so strip them from the values.
## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/13265


lib/common/client/concerns/monitoring.rb
```diff
@@ -23,7 +23,8 @@ def increment_total(caller)
         end
 
         def increment_failure(caller, error)
```
erluti, line 25: any chance you could introduce a test for `increment_failure`?

erluti (commented):
LGTM, would love if you introduced a test though

lib/common/client/concerns/monitoring.rb
annaswims: yep! I added a test.

erluti: I'm not seeing any changes to `common/client/concerns/monitoring_spec.rb`, did you push it?

annaswims: oopsie! Now I pushed the commit. Thanks!

erluti (approved):
Thanks for adding a test!

PR #4857, opened 2020-09-08
Forms API - Include deleted forms when returning all forms

hinzed1127:
Dependency for https://github.com/department-of-veterans-affairs/va.gov-team/issues/11625 . We were previously filtering out non-`nil` values of `deleted_at` for the forms endpoint. This removes that and returns all forms

## Original issue(s)
[Forms API - Add option to include deleted forms in Search](https://vajira.max.gov/browse/API-1822)

- Swagger docs were missing the inclusion of `deleted_at` as a returned value. The description is WIP and will need approval before translation


templeman15:
@hinzed1127 @charleystran looks good to me!

The only thing that caught my attention was that no specs were updated. I'm not sure where the appropriate spec to update would be in this situation. `modules/va_forms/spec/models/form_spec.rb` does not have a spec for `.search` and `modules/va_forms/spec/requests/v0/forms_request_spec.rb` isn't testing for which specific forms are being returned.

modules/va_forms/config/locales/en.yml
```diff
@@ -7,6 +7,7 @@ en:
       benefit_categories: 'Listing of categories and match'
       benefit_category_name: 'Name of the benefit category of the form'
       benefit_category_description: 'Description of the benefit category of the form'
+      deleted_at: "The timestamp at which the form was deleted"
```
hinzed1127, line 10: 2 questions for this one:
1. is this a good enough description?
2. assuming it is, do I have to do anything to initiate translations of it?

jbalboni (approved):

erluti:
Merging in the latest master should get you past that security failure in the pipeline

spec/support/schemas/va_forms/forms.json
```diff
@@ -38,7 +38,14 @@
               "form_details_url": { "type": ["string", "null"] },
               "form_type": { "type": ["string", "null"] },
               "language": { "type": ["string", "null"] },
-              "related_forms": { "type": ["array", "null"], "items": { "type": "string"}  },
```
bastosmichael, line 41: Let's make sure we get @ncksllvn to review this since it impacts his code...

ncksllvn (approved):
LGTM! The only concern I have is if it’s okay for deleted forms to start being included in the response JSON without the client being aware of the change (like maybe through a query param “?includeDeletedForms=true” or something). I think we could update va.gov/find-forms pretty quickly though to do some handling.

hinzed1127:
looks like this just needs another quick glance by the @department-of-veterans-affairs/backend-review-group to be able to merge

bastosmichael (approved):
LGTM if Nick is good with it...

PR #4860, opened 2020-09-08
log more info when MVI add_person fails

annaswims:
## Description of change
We are seeing a high rate of error for MVI add person calls and I hope that logging the response will provide more info.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#12376


erluti (approved):
LGTM

PR #4861, opened 2020-09-08
Circle gem caching

LindseySaari:
## Description of change
This PR updates the gem caching for CircleCI jobs. The ruby orb is much slower and doesn't properly load the cache (instead it installs the gems). The lint job the orb took ~3 min and with the custom solution were seeing ~40-50 seconds.

Additionally trailing whitespace was removed from the config

## Original issue(s)
department-of-veterans-affairs/va.gov-team#13114


erluti (approved):
LGTM

PR #4864, opened 2020-09-08
[13212] Dependent debt filtering

camerontesterman:
## Description of change
This adds a method to determine whether any debts that are returned from the DMC are dependent debts. If any dependent debts are detected, the following occurs:

- `/v0/debts` returns as per usual, minus the dependent debts.
- `/v0/debts` has a new boolean (`has_dependent_debts`) added to the response which will be returned as `true`.
- `/v0/debt_letters` will return a `401`

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/13212

Tested and verified locally.


lib/debts/service.rb
```diff
@@ -6,26 +6,51 @@
...
     end
 
     private
 
+    def init_file_number
+      bgs_file_number = BGS::PeopleService.new(@user).find_person_by_participant_id[:file_nbr]
+      bgs_file_number.presence || @user.ssn
+    end
+
+    def sorted_debts
```
erluti, line 41: this is not a list of sorted debts, but debts with sorted histories.  I think this name is a little misleading and would probably do something like `debts_with_sorted_histories` myself

erluti (approved):
one minor nitpick, otherwise LGTM

lib/debts/service.rb
camerontesterman: Good call. Change has been made.

PR #4873, opened 2020-09-10
health quest service PR 2, break commits up into smaller pieces

massrb:
## Description of change

This is the second pull request for the health quest service. P Becker had mentioned other goals to make PRs even smaller, though this particular PR I already had all set up before that with tests and so on even though I am just pushing it now.

It has settings very similar to VAOS:

VAOS:

```
va_mobile:
  url: 'https://veteran.apps.va.gov'
  mock: false
  timeout: 15
```

Health Quest:

```
hqva_mobile:
  url: 'https://veteran.apps.va.gov'
  mock: false
  timeout: 15
 ```

## Original issue(s)

### CODE REVIEW

File: modules/health_quest/app/controllers/health_quest/v0/appointments_controller.rb [hunk 1]
```diff
@@ -0,0 +1,56 @@
...
+          appointment_service.get_appointments(type, start_date, end_date, pagination_params)
+      end
+
+      def each_serializer
+        "HealthQuest::V0::#{params[:type].upcase}AppointmentsSerializer".constantize
+      end
+
+      def validate_params
+        raise Common::Exceptions::ParameterMissing, 'type' if type.blank?
+        raise Common::Exceptions::InvalidFieldValue.new('type', type) unless %w[va cc].include?(type)
```
erluti, line 34: I think to be consistent with the other helpers, the InvalidFieldValue error should be in the `type` method and the line above should be `raise Common::Exceptions::ParameterMissing, 'type' if params[:type].blank?`
Then maybe this method would be named `validate_params_present`?

modules/health_quest/app/controllers/health_quest/v0/appointments_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+        HealthQuest::AppointmentService.new(current_user)
+      end
+
+      def appointments
+        @appointments ||=
+          appointment_service.get_appointments(type, start_date, end_date, pagination_params)
+      end
+
+      def each_serializer
+        "HealthQuest::V0::#{params[:type].upcase}AppointmentsSerializer".constantize
```
erluti, line 29: this should probably use `type` instead of `params[:type]`

erluti (commented):
Mainly want to know about `appointment_by_id`, but also a small refactor suggestion

modules/health_quest/app/controllers/health_quest/v0/appointments_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+      before_action :validate_params, only: :index
+
+      def index
+        res = each_serializer.new(appointments[:data], meta: appointments[:meta])
+        render json: res
+      end
+
+      private
+
+      def appointment_by_id
```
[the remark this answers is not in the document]
massrb: ok, sorry I will remove that. It belongs in the next pull request and not this one

modules/health_quest/app/controllers/health_quest/v0/appointments_controller.rb
massrb: ok, got it ..

massrb: Actually, I realized that there is no longer a CC type so since there is only one type, the type() method and some other things are not needed so I got rid of that method and simplified some things

erluti (approved):
Looks good!  I'm glad you pulled `type` since it's unnecessary

PR #4877, opened 2020-09-10
Add ssn and birthdate for dependents

digitaldrk:
## Description of change
While testing with our integration partner (BGS), we were informed that dependents' data should include SSN and DOB so that RBPS systems could easily look up existing dependents in their systems. This lowers the chance that a benefit claim does requires manual intervention helping Veterans efficiently adjust their benefits and saving the VA money.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#13273

    * negative
    * The the feature this touches is not available to the gen public yet
    * In this feature, yes but not in this PR
    * negative


erluti (approved):
LGTM

PR #4878, opened 2020-09-11
10-10CG Schema Change: Removing Gender

kevinmirc:
Fixes: https://github.com/department-of-veterans-affairs/va.gov-team/issues/13098

**Description**:
The 10-10CG will be removing the option to select gender as "Unknown" and gender will no longer be required. This PR supports those schema changes.

**Related PRs**
Schema: https://github.com/department-of-veterans-affairs/vets-json-schema/pull/502
UI: https://github.com/department-of-veterans-affairs/vets-website/pull/13993


spec/services/form1010cg/service_spec.rb
```diff
@@ -14,7 +14,6 @@
         },
         'ssnOrTin' => Faker::IDNumber.valid.remove('-'),
         'dateOfBirth' => Faker::Date.between(from: 100.years.ago, to: 18.years.ago).to_s,
-        'gender' => %w[M F].sample,
```
erluti, line 17: so this should have no gender key by default?

erluti (approved):
LGTM, minor question

spec/services/form1010cg/service_spec.rb
LinnJS: I think this is true since gender is no longer required

kevinmirc: Yes, gender should not be present in the minimalist example.

PR #4879, opened 2020-09-11
Add index to `accounts` table on `icn` field

jholton:
## Description of change
The `accounts` table can be queried on the `icn` field as part of the SAML login callback.
This field is not indexed, resulting in slow query times.

This PR adds an index to `accounts` table on `icn` field

## Original issue(s)
department-of-veterans-affairs/va.gov-team#13442


erluti (approved):
LGTM
TIL: `aglorithm: :concurrently` nice!

PR #4882, opened 2020-09-11
[API-1841] Update descriptions in docs by removing copy about feature being in t…

templeman15:
## Description of change
This PR removes the copy around the power of attorney only being in test environments. 

## Original issue(s)
[Update Claims API documentation to take out reference to POA only in Test Environments](https://vajira.max.gov/browse/API-1841)

Docs were updated to remove POA only being in test environments


modules/claims_api/app/swagger/claims_api/description/v0.md
```diff
@@ -12,7 +12,6 @@ Currently, the API supports automatic establishment of:
 
  - Application for Disability Compensation (21-526EZ)
  - Intent to File (21-0966)
- - Appointment of Representative (21-22 and 21-22a) **(in test environments only)**
```
erluti, line 15: here the whole line is struck, but in v1 only the part about being in test is removed.  Is this correct?

erluti (commented):
wanted to double check this is the correct change

modules/claims_api/app/swagger/claims_api/description/v0.md
templeman15: @erluti yes that is correct per my conversation with Kelly Lein due to power of attorney endpoints not being in docs on v0 all-together.

bastosmichael: We've tested V1 but due to the internal nature of V0 and is not Oauth based then there's still some work for us to do to successfully launch it in V0 because of BGS constraints.

bastosmichael (approved):
Make sure 2122 doesn't actually show up in documentation on V0 for now.

PR #4885, opened 2020-09-11
CircleCI simplecov report merger for parallel specs

LindseySaari:
## Description of change
When running parallel specs in CircleCI, we get partial simplecov reports in each container, which causes the Simplecov minimum coverage to fail. This PR fetches all coverage jsons generated by simplecov across the various parallel containers of the Circle build, and merges them. The coverage files are stored as artifacts in the spec job for each node and the workspace is then passed to the subsequent coverage job. The coverage job runs after the spec job and runs a rake task `bundle exec rake simplecov:report_coverage` to merge the results.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#13108

## Testing
- [x] The entire build passes in CircleCI
- [x] A simplecov report is generated per node and merged into one


.circleci/config.yml
```diff
@@ -32,18 +32,26 @@
...
 
+  defaults: &defaults
+    working_directory: ~/vets-api
+    executor: vets_api_docker_executor
+
   jobs:
     verify_pacts:
-      working_directory: ~/vets-api
+      <<: *defaults
       executor: vets_api_docker_executor
```
erluti, line 42: the defaults would also be pulling in the `executor` too, right?

LindseySaari: Yes, good catch! Forgot to remove

.circleci/config.yml
```diff
@@ -32,54 +32,88 @@
...
+      <<: *defaults
       steps:
         - checkout_and_install
         - setup_test_db
         - verify_stable_pacts
         - store_test_results:
             path: ./rspec
+        - store_artifacts:
+            path: ./ci
+        - store_artifacts:
```
erluti, line 50: can this key be duplicated

LindseySaari: Yes, there's no limit on the number of store_artifacts steps a job can run. Actually in this case, I just cleaned up a couple of the jobs that don't store any artifacts at all. Thanks for pointing this out!

lib/tasks/simplecov_parallel.rake
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'common/exceptions/base_error'
```
erluti, line 3: is this using this?

erluti (approved):
LGTM

lib/tasks/simplecov_parallel.rake
LindseySaari: Removing it 👍

PR #4887, opened 2020-09-11
SSOe SAML Tracker event changes

ericbuckley:
The SSOe SAML Tracker was added to give more clarity on how successful the SAML Post binding form is in submitting, however the numbers reported back are extremely low, and in some cases even lower than our SAML Response rates.  I believe the issue is that the XHR request is in some cases being canceled because the form submission is happening too quickly.  This PR addresses that, by only submitting the form once we receive acknowledgment from the XHR request that it has been received.

This will add a small delay to the user's experience, but should give us accurate metrics whether the SAML Post form JS block is loading properly.

This PR also fixes a param naming issue which is causing all the application logs to show up with SAML ids of `nil`.

closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/13409

### DIFF
--- a/app/controllers/v1/sessions_controller.rb
+++ b/app/controllers/v1/sessions_controller.rb
@@ -69,7 +69,7 @@ module V1
     end
 
     def tracker
-      id = pparams[:id]
+      id = params[:id]
       type = params[:type]
       authn = params[:authn]
       values = { 'id' => id, 'type' => type, 'authn' => authn }
--- a/app/controllers/v1/sessions_controller.rb
+++ b/app/controllers/v1/sessions_controller.rb
@@ -69,7 +69,7 @@ module V1
     end
 
     def tracker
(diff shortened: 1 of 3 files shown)


jholton (approved):
When it's merged and deployed to staging, I can test via IE in CAG

erluti (approved):

ericbuckley:
> When it's merged and deployed to staging, I can test via IE in CAG

@jholton I ran tests for the following browsers to a) make sure authentication still works b) double check that a SSOe SAML Tracker log entry is showing up.  So far everything has passed
- chrome on android
- safari on iOS
- firefox on mac
- safari on mac
- chrome on windows
- edge on windows
- ie 11 on windows

jholton:
I tested in IE 11.3866.14393.0 in a CAG session.  Worked as expected.

PR #4890, opened 2020-09-11
[API-1188] Create daily sidekiq job to find, catch and rerun "uploaded" statuses for when SNS fails

templeman15:
## Description of change
Setup a daily sidekiq job to find `UploadSubmissions` that are stuck in the `uploaded` status when SNS fails for some reason. The job runs the `UploadProcessor` worker.

## Original issue(s)
[Create daily sidekiq job to find, catch and rerun "uploaded" statuses for when SNS fails](https://vajira.max.gov/browse/API-1188)

### DIFF
--- a/config/sidekiq_scheduler.yml
+++ b/config/sidekiq_scheduler.yml
@@ -131,7 +131,7 @@ VBADocuments::ReportUnsuccessfulSubmissions:
   description: "Weekly report of unsuccessful submissions"
 
 VBADocuments::RunUnsuccessfulSubmissions:
-  cron: "0 0 * * MON-FRI America/New_York"
+  cron: "0 0 23 * * America/New_York"
   class: VBADocuments::RunUnsuccessfulSubmissions
   description: "Run VBADocuments::UploadProcessor for submissions that are stuck in uploaded status"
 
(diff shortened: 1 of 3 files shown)


config/sidekiq_scheduler.yml
```diff
@@ -130,6 +130,11 @@ VBADocuments::ReportUnsuccessfulSubmissions:
   class: VBADocuments::ReportUnsuccessfulSubmissions
   description: "Weekly report of unsuccessful submissions"
 
+VBADocuments::RunUnsuccessfulSubmissions:
+  cron: "0 0 * * MON-FRI America/New_York"
```
templeman15, line 134: Is this correct for running the job every day? @bastosmichael

bastosmichael: Looks good but make sure that the RunUnsuccessfulSubmissions gets triggered before ReportUnsuccessfulSubmissions time wise.

bastosmichael (changes requested):
Minor change to the timeline for when this job should run otherwise looks good.

config/sidekiq_scheduler.yml
templeman15: @bastosmichael updated to run daily at 2300 hours.

bastosmichael (approved):

PR #4898, opened 2020-09-14
13065 Mobile Auth

kreek:
## Description of change
To split these changes up there's a PR with the core vets-api `ApplicationController` changes [here](https://github.com/department-of-veterans-affairs/vets-api/pull/4914). Please look at that PR first.

Implements the authentication flow to the mobile engine. Adds an authenticate method to the mobile `ApplicationController` which uses the `IAMSSOeOAuth::SessionManager` to find and create sessions with it's own versions of the user, user_identity, and session classes (which all inherit from the core vets-api versions of those classes).

Also adds the initial version of the `/mobile/v0/user` endpoint. This will return more user traits once the mobile screens are finalized.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#13065

* Adds redis.yml entries for IAMUser, IAMUserIdentity, IAMSession classes
* The mobile auth design doc is [here](https://github.com/department-of-veterans-affairs/va.gov-team/blob/mobile-ssoe-design-doc/platform/engineering/design-docs/2020-08-12-mobile-ssoe-auth-design-doc.md)


app/models/iam_user_identity.rb
```diff
@@ -0,0 +1,50 @@
...
+  #
+  # @param iam_profile [Hash] the profile of the user, as they are known to the IAM SSOe service.
+  # @return [IAMUserIdentity] an instance of this class
+  #
+  def self.build_from_iam_profile(iam_profile)
+    loa_level = iam_profile[:fediamassur_level].to_i
+
+    identity = new(
+      email: iam_profile[:email],
+      expiration_timestamp: iam_profile[:exp],
```
jjulian, line 26: Can we confirm that the format of `exp` is compatible with the call to `redis_namespace.expireat`?

app/models/iam_user_identity.rb
```diff
@@ -0,0 +1,50 @@
...
+      first_name: iam_profile[:given_name],
+      iam_icn: iam_profile[:fediam_mviicn],
+      iam_edipi: iam_profile[:fediam_do_dedipn_id],
+      iam_sec_id: iam_profile[:fediamsecid],
+      last_name: iam_profile[:family_name],
+      loa: { current: loa_level, highest: loa_level },
```
jjulian, line 32: I think this is ok, but want to confirm. loa_level will always be "3" from IAM SSOe?

config/redis.yml
```diff
@@ -102,6 +102,15 @@ development: &defaults
...
     each_ttl: 3600 # 1 hour
+  iam_session:
+    namespace: iam-session
+    each_ttl: 1800
+  iam_user:
+    namespace: iam-user
+    each_ttl: 1800
+  iam_user_identity:
+    namespace: iam-user-identity
+    each_ttl: 1800
```
jjulian, line 113: Is this 30 minutes to match the access_token ttl?

modules/mobile/spec/spec_helper.rb
```diff
@@ -1,15 +0,0 @@
...
-
-require 'rspec/rails'
-
-ENGINE_RAILS_ROOT = File.join(File.dirname(__FILE__), '../')
-
-# Requires supporting ruby files with custom matchers and macros, etc,
-# in spec/support/ and its subdirectories.
-Dir[File.join(ENGINE_RAILS_ROOT, 'spec/support/**/*.rb')].sort.each { |f| require f }
-
-RSpec.configure { |config| config.use_transactional_fixtures = true }
```
jjulian, line 15: `spec_helper.rb` is not needed?

kreek: Realized we want to use the core vets-api spec_helper so this one's not needed.

config/redis.yml
kreek: Yes, but this is a default. The TTL will be set from the introspect `exp` value.

app/models/iam_user_identity.rb
kreek: 👍 I'll double check that with Perry.

app/models/iam_user_identity.rb
kreek: Redis namespace gem aliases all the Redis methods and expireat takes a unix timestamp 
https://github.com/redis/redis-rb/blob/master/lib/redis.rb#L430-L439

jjulian (approved):
Looks great!

kfrz:
Working to review this, though there are a few build failures it seems and merge resolutions required. 

Exciting stuff!

kreek:
@kfrz this is passing CI now, I did fix something but there was also a VBA Documents spec error which seems dependent on the run order.

erluti:
> a VBA Documents spec error which seems dependent on the run order.

This test flaps sometimes and we're trying to figure out why.  Do you have some more evidence for this?

kreek:
@erluti the failing VBA Document run

http://jenkins.vfs.va.gov/blue/organizations/jenkins/testing%2Fvets-api/detail/PR-4898/18/pipeline

modules/mobile/app/serializers/mobile/v0/user_serializer.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require 'fast_jsonapi'
+
+module Mobile
+  module V0
+    class UserSerializer
+      include FastJsonapi::ObjectSerializer
+
+      # TODO: return full set of user data, including latest appeals, claims, and appointments
```
kfrz, line 10: Is this `TODO` before the PR is merged, or is there a follow-up ticket for traceability purposes?

modules/mobile/spec/controllers/application_controller_spec.rb
```diff
@@ -0,0 +1,133 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require_relative '../rails_helper'
+
```
kfrz, line 5: Why is this necessary?

kfrz (approved):
This looks good to me, just a couple of nitpick/questions,  `TODO`s shouldn't be left in code committed, IMO, without an issue linked in the comment as well

Otherwise, :clap:

modules/mobile/app/serializers/mobile/v0/user_serializer.rb
kreek: this is for follow up, wanted a slimed down version to test auth before the full user endpoint goes in another PR

modules/mobile/spec/controllers/application_controller_spec.rb
kreek: I should rename the second one, the first one is the core 'rails_helper', the second is the module's 'rails_helper' which has Rspec configurations specific to this module, e.g. flipping the feature on, creating a signed in IAM user.

modules/mobile/app/serializers/mobile/v0/user_serializer.rb
kreek: added and issue # to the todo

PR #4902, opened 2020-09-15
[13357] Debt letter filename fix

camerontesterman:
##  Description of change
This changes the filename of the debt letters that can be downloaded by the veteran from `letter.pdf` to a more user friendly filename, which is the concatenation of the document's type description along with the date.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/13357

Tested and verified locally.

### DIFF
--- a/lib/debts/letter_downloader.rb
+++ b/lib/debts/letter_downloader.rb
@@ -29,6 +29,15 @@ module Debts
       ).content
     end
 
+    def file_name(document_id)
+      document = @client.send_request(
+        VBMS::Requests::FindDocumentVersionReference.new(@service.file_number)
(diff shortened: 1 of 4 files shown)


lib/debts/letter_downloader.rb
```diff
@@ -29,10 +29,13 @@ def get_letter(document_id)
       ).content
     end
 
+    def file_name(document_id)
```
erluti, line 32: is it possible to add a test for this new method?

camerontesterman: Done

erluti (approved):
LGTM

PR #4903, opened 2020-09-15
[API-2001] Set status to 'vbms' on various UploadSubmission GUIDs

templeman15:
## Description of change
This PR creates a rake task that sets various (709) GUIDs belonging to UploadSubmissions to be set to the `vbms` status. This is a clone w/some updates of the `modules/vba_documents/lib/tasks/one_time_status_update_20200803.rake` rake task.

## Original issue(s)
[Followup to orphaned "processing" statuses](https://vajira.max.gov/browse/API-2001)


bastosmichael (approved):
LGTM but pair with @hinzed1127 once this one is deployed on getting into the prod server via ssm and run this rake task since you don't have AWS access yet.

PR #4905, opened 2020-09-15
Update Sentry configuration to include environment

kfrz:
## Original issue(s)

### DIFF
--- a/config/initializers/app_info.rb
+++ b/config/initializers/app_info.rb
@@ -3,5 +3,4 @@
 module AppInfo
   GIT_REVISION = ENV.fetch('GIT_REVISION', 'MISSING_GIT_REVISION')
   GITHUB_URL   = 'https://github.com/department-of-veterans-affairs/vets-api'
-  VAGOV_ENV    = Settings.vsp_environment
 end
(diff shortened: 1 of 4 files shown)


config/initializers/app_info.rb
```diff
@@ -3,4 +3,5 @@
 module AppInfo
   GIT_REVISION = ENV.fetch('GIT_REVISION', 'MISSING_GIT_REVISION')
   GITHUB_URL   = 'https://github.com/department-of-veterans-affairs/vets-api'
+  VAGOV_ENV    = Settings.vsp_environment
```
omgitsbillryan, line 6: Is this sentry related or otherwise?

omgitsbillryan (approved):
LGTM

config/initializers/raven.rb
```diff
@@ -7,6 +7,7 @@
 
 Raven.configure do |config|
   config.dsn = Settings.sentry.dsn if Settings.sentry.dsn
+  config.environment = Settings.vsp_environment if Settings.vsp_environment
```
erluti, line 10: Is `~` an acceptable value here?  Lighthouse had some things break because somebody put in a filler value in settings that broke the development and staging deployments that didn't have the right `settings.local.yml` to go with it

PR #4906, opened 2020-09-15
Load user from redits and pass to flipper to enable UAT for specific …

ffafara-tw:
Configure form526_confirmation_email feature toggle in vets-api so that it can be used for UAT
[department-of-veterans-affairs/notification-api#175]

Co-authored-by: Phil Herber <pherbert@thoughtworks.com>

## Description of change

Load user from Redis and pass it to flipper, so that we can turn the toggle for specific users.
Since it's a background task it can happen that user session will not be present in Redis - in that case we will pass nil to flipper (as actor) and feature toggle will remain disabled.

## Original issue(s)
department-of-veterans-affairs/notification-api#175


kfrz (approved):
:+1: :shipit:

PR #4908, opened 2020-09-16
EVSS 500 error catch and fix

bastosmichael:
## Description of change
GDIT asked us to look up some ID's in Staging for test Claims V0 Submissions and found that the upstream 500 errors weren't being relayed downstream.

## Original issue(s)
https://vajira.max.gov/browse/API-2387

### DIFF
--- a/modules/claims_api/app/workers/claims_api/claim_establisher.rb
+++ b/modules/claims_api/app/workers/claims_api/claim_establisher.rb
@@ -3,10 +3,12 @@
 require 'sidekiq'
 require 'evss/disability_compensation_form/service_exception'
 require 'evss/disability_compensation_form/service'
+require 'sentry_logging'
 
 module ClaimsApi
   class ClaimEstablisher
     include Sidekiq::Worker
+    include SentryLogging
 
     def perform(auto_claim_id)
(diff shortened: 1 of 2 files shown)


templeman15:
@bastosmichael looks good to me. Maybe throwing in a spec for the new failing case?

modules/claims_api/app/workers/claims_api/claim_establisher.rb
```diff
@@ -14,17 +14,20 @@ def perform(auto_claim_id)
...
+    rescue ::EVSS::DisabilityCompensationForm::ServiceException => e
+      auto_claim.status = ClaimsApi::AutoEstablishedClaim::ERRORED
+      auto_claim.evss_response = e.messages
+      auto_claim.save
+      raise e
+    rescue ::Common::Exceptions::BackendServiceException => e
+      auto_claim.status = ClaimsApi::AutoEstablishedClaim::ERRORED
+      auto_claim.evss_response = [{ 'key' => e.status_code, 'severity' => 'FATAL', 'text' => e.original_body }]
+      auto_claim.save
+      raise e
```
hinzed1127, line 30: is there a way to just add in the sentry logging here (and in the other rescue block) so that way it doesn't get retried, like we were talking through before?

templeman15 (approved):
👍

PR #4909, opened 2020-09-16
hotfix syntax

jbritt1:
## Original issue(s)

Something is causing the deployment of vets-api-server to cycle in staging (and dev I think). Found a mention of an undefined method "environment". The suggestion that came with that error asks if the user meant "environments". Could be way off here, but just trying to connect the dots. Forgive me if this is not what is causing the issue being observed. 

image


benniemosher (approved):
lgtm

erluti (approved):
nice catch!

link to docs from Lindsey https://docs.sentry.io/platforms/ruby/configuration/options/#optional-settings

PR #4912, opened 2020-09-16
Allow single data fixture directory for 10-10cg PdfFill::Filler#fill_form spec

kevinmirc:
**Description**
There are two ways to generate a 10-10CG: signed and unsigned. The data fixtures for the spec both should be identical (which they are right now). This PR removed the duplicate fixtures and allows a single data fixture for multiple output PDF fixtures in the `PdfFill::Filler#fill_form` spec.


kfrz (approved):
:+1:

kfrz:
@kevinmirc could you link this to an existing issue for posterity/archival sake? 

Thanks!

kevinmirc:
Hi @kfrz,

Yes! I've updated the description to reference the PR this was a prerequisite for, and the related ticket. This was really just a refactor that could stand on its own. I didn't want it to clutter the changes following PR.

Thanks!

PR #4914, opened 2020-09-16
13065 ApplicationController Concern Refactor

kreek:
## Description of change
This PR is pre-work for the mobile auth PR [here](https://github.com/department-of-veterans-affairs/vets-api/pull/4898) (which includes these changes).

Moves the core vets-api `ApplicationController` logic into concerns so that they can be composed into the mobile `ApplicationController` (and any future controllers that don't use SAML auth). The lib `SentryLogging` module has a small refactor of duplicate code. The rest of the logic lives in four new concerns:
* `ExceptionHandling`: has the exception mapping and rails/sentry logging of the original errors
* `Headers`: has the`prepend_before_action` callback which sets the X-Git... headers
* `Instrumentation`: has the custom `append_info_to_payload` method
* `SentryControllerLogging`: the sentry logging that's specific to controllers

## Original issue(s)
department-of-veterans-affairs/va.gov-team#13065

* The mobile auth design doc is [here](https://github.com/department-of-veterans-affairs/va.gov-team/blob/mobile-ssoe-design-doc/platform/engineering/design-docs/2020-08-12-mobile-ssoe-auth-design-doc.md)


jholton (approved):
Nice organizing!

jjulian (approved):

app/controllers/application_controller.rb
```diff
@@ -9,19 +9,21 @@
```
erluti, line 9: move the above `require`s to the new concerns they belong with, please

kfrz (approved):
This is great!!! Very happy to see these changes. Looks good to me and looks like the specs are still passing. :+1: :zap: Thanks!

erluti (changes requested):
I don't want this comment to be missed https://github.com/department-of-veterans-affairs/vets-api/pull/4914#discussion_r491170522

erluti (approved):
Thanks for keeping those `require`s!

kreek:
thanks for looking at this again @erluti, I'll merge this after the deploy today and test on staging

PR #4915, opened 2020-09-16
simplify url_service calls

ericbuckley:
Cleaning up calls from the SessionControllers to `url_service`.  The first part of this PR addresses the referenced issue by creating an instance variable whenever the url_service function is invoked and using that value if defined rather than reinitializing a new instance of URLService.  The second part is removing the need to pass through the "previous_saml_uuid" variable when initializing URLService, and instead leaving it up to URLService to find that value, if available, in the params.

These changes will address 2 issues
- it will make the code slightly faster by reducing the need to reinitialize a URLService object everytime the `url_service` function is called (in addition to parsing XML, network latency is added to every initialization as we save data to Redis)
- the SessionControllers will be less prone to errors when developers add calls to url_service.  In the past we have ran into small bugs when a new call to url_service was needed and the `previous_saml_uuid` variable was omitted.

closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/13231


jholton (approved):

benniemosher (approved):

app/controllers/v1/sessions_controller.rb
```diff
@@ -130,70 +129,69 @@ def user_login(saml_response)
...
 
-        helper.custom_url params[:authn]
+        url_service(false).custom_url params[:authn]
       end
     end
     # rubocop:enable Metrics/CyclomaticComplexity
 
-    def saml_request_stats(tracker)
+    def saml_request_stats
+      t = url_service.tracker
```
erluti, line 185: is there a reason this was shortened to `t` instead of being `tracker`?

lib/saml/post_url_service.rb
```diff
@@ -31,12 +29,11 @@ def initialize(saml_settings, session: nil, user: nil, params: {},
         @type = JSON.parse(params[:RelayState])['type']
       end
       @query_params = {}
-      @tracker = initialize_tracker(params, previous_saml_uuid: previous_saml_uuid)
+      @tracker = initialize_tracker(params)
 
       Raven.extra_context(params: params)
       Raven.user_context(session: session, user: user)
     end
-    # rubocop:enable Metrics/ParameterLists
```
erluti, line 39: 🎉

lib/saml/url_service.rb
```diff
@@ -199,11 +196,19 @@ def add_query(url, params)
       end
     end
 
+    def previous_saml_uuid(params)
+      if params[:action] == 'saml_callback'
+        resp = SAML::Responses::Login.new(params[:SAMLResponse] || '', settings: @saml_settings)
```
erluti, line 201: we should add a `require_relative 'responses/login'` to this file

erluti (changes requested):
LGTM, only requesting changes because of the requires

app/controllers/v1/sessions_controller.rb
```diff
@@ -6,7 +6,6 @@
 require 'saml/responses/login'
 require 'saml/responses/logout'
 require 'saml/ssoe_settings_service'
-require 'saml/url_service'
```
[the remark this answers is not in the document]
ericbuckley: @erluti that require is on line 5

app/controllers/v1/sessions_controller.rb
ericbuckley: no good reason, I can revert

app/controllers/v1/sessions_controller.rb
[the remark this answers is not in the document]
erluti: Oops!  I didn't realize there were 5 lines above this.  Thanks for taking out the unneeded one!

erluti (approved):
thanks for the fixes!  looks good!

PR #4921, opened 2020-09-17
[13658] Fix MDT birthdate formatting issue

camerontesterman:
## Description of change 
Birthdates that do not have hyphens between the month & year and day & month will not pass validation when they are submitted to DLC's API. In order to ensure that this format is always followed, I've added a method that will parse whatever date is gathered from the `@current_user`'s `birth_date` instance variable and format it appropriately so that it will be accepted by DLC's API.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/13658

tested and verified locally

### DIFF
--- a/lib/mdot/client.rb
+++ b/lib/mdot/client.rb
@@ -74,9 +74,13 @@ module MDOT
         VA_VETERAN_MIDDLE_NAME: @user.middle_name,
         VA_VETERAN_LAST_NAME: @user.last_name,
         VA_VETERAN_ID: @user.ssn.last(4),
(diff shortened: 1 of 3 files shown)


kfrz (commented):
I think this looks good, are there unit tests for this method?

camerontesterman:
> I think this looks good, are there unit tests for this method?

Thanks for taking a look. Not direct tests, but I did alter the birth date to remove the hyphens from the user that is used in the get_supplies tests and am testing against the headers, if the birth date isn't formatted correctly this test would fail which would mean the private method isn't doing what it's supposed to do.

kfrz (approved):

PR #4922, opened 2020-09-17
Adding check for param type in 0966/active end point

bastosmichael:
## Description of change
During production testing I discovered this one bug popping up http://sentry.vfs.va.gov/vets-gov/platform-api-production/issues/196730/?query=is%3Aunresolved that was causing a 500 error in the /active end point because BGS can't not receive a type so needed a validation of the /active end point to prevent a request from coming through without a param type.

## Original issue(s)
https://vajira.max.gov/browse/API-1242


templeman15:
@bastosmichael thoughts on adding a spec for when no type param is passed?

ElleeB (commented):
LGTM

PR #4923, opened 2020-09-17
renaming the ssoe tracker url to trace

ericbuckley:
Renaming the `/v1/sessions/tracker` URL to `/v1/sessions/trace`.  Some browser ad blocker extensions (ie AdblockPlus) prevent XHR requests to urls with the word "track" in it.

I manually tested this locally with the following AdblockPlus settings enabled.  With the former URL, the form submission was blocked, however when using the `/v1/sessions/trace` URL, the POST SAML binding form was successfully submitted.
Screen Shot 2020-09-17 at 08 54 36

refs https://github.com/department-of-veterans-affairs/vets-api/pull/4806#issuecomment-694289957


jholton (approved):

kfrz (approved):
I think this is good - hopefully trackers won't block "trace" anytime soon :D

PR #4925, opened 2020-09-17
Add SaveInProgress id to metadata

annaswims:
## Description of change
we'd like to add the SIP id to the form metadata so we can better troubleshoot http://sentry.vfs.va.gov/vets-gov/website-production/issues/12188/

## Original issue(s)


kfrz (approved):
Code looks good, I can't think of any security implications that apply here. :ship:

PR #4926, opened 2020-09-17
Undo BDD changes

annaswims:
## Description of change
we decided that form526 bdd is one form (same as 526) after all so I'm reverting the changes made in  #4382 and #4491

## Original issue(s)


kfrz (approved):
:+1: Revert approved.

PR #4897, opened 2020-09-14
13281 Do not auto/eager-load lib directory

erluti:
Until this PR is merged, `config/application.rb` manually added the `/lib` directory to the auto/eager-load path. This is non-standard Rails configuration and has lead to some practices that make it difficult to switch over to the zeitwerk loader. This change is to remove that code and prepare the way for zeitwerk.

**Please test on your features locally!**

PR #4928, opened 2020-09-18
Require Libs for HLR

edmangimelli:
For https://github.com/department-of-veterans-affairs/vets-api/pull/4897


spec/lib/decision_review/service_spec.rb
```diff
@@ -1,7 +1,9 @@
 # frozen_string_literal: true
 
 require 'rails_helper'
+require 'common/exceptions/forbidden'
```
erluti, line 4: I think we should try to keep the requires in the specs to a minimum.  It looks like `Common::Exceptions::Forbidden` and `DecisionReview::ServiceException` are coming from the `DecisionReview::Service`, so that class should be requiring those.  I don't want to end up in a place where we don't spot a missing `require` because the test included what should have been in the class.

edmangimelli: blocker?

erluti: If you're going to make me sound all stern about it, yes.
Pretty easy to change in the github editor though

edmangimelli: whoops. I asked "blocker?" then immediately decided, meh, I'll just remove it, and removed my question. Sorry about the email notification 👍

erluti (approved):

PR #4929, opened 2020-09-18
alphabetize initializers/breakers

edmangimelli:
was originally doing this in the context of this pr: https://github.com/department-of-veterans-affairs/vets-api/pull/4928

but pulled it into it's own PR to investigate

discussion:
https://dsva.slack.com/archives/CBU0KDSB1/p1600431259007500

### DIFF
--- a/config/initializers/breakers.rb
+++ b/config/initializers/breakers.rb
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
 
-require 'bb/configuration'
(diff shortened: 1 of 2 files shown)


edmangimelli:
Decision came down to not mess with this load order 👍 
https://dsva.slack.com/archives/CBU0KDSB1/p1600435043008100?thread_ts=1600431259.007500&cid=CBU0KDSB1

PR #4934, opened 2020-09-21
PR1 for PGD service, PGD service here is dummied up a bit

massrb:
For the health quest module - this adds a PGD (patient generated data) service. The service is not actually functional but our front end developer wanted to see the endpoints.

## Original issue(s)

### DIFF
--- a/modules/health_quest/app/services/health_quest/pgd_service.rb
+++ b/modules/health_quest/app/services/health_quest/pgd_service.rb
@@ -8,10 +8,9 @@ module HealthQuest
     def get_pgd_resource(type, id = nil, pagination_params = {})
       # params = {}
       res_id = id || '333'
+      fhir_data = { id: res_id.to_s, text: "this is the #{type}" }
       {
-        # rubocop:disable Layout/LineLength
-        data: deserialized_resource({ my_attr: 'none/fake', fhir_data: { id: res_id.to_s, text: "this is the #{type}" } }, type),
-        # rubocop:enable Layout/LineLength
+        data: deserialized_resource({ my_attr: 'none/fake', fhir_data: fhir_data }, type),
         meta: pagination(pagination_params)
       }
       #      with_monitoring do
--- a/modules/health_quest/app/services/health_quest/pgd_service.rb
+++ b/modules/health_quest/app/services/health_quest/pgd_service.rb
@@ -5,7 +5,7 @@ Faraday::Middleware.register_middleware health_quest_logging: HealthQuest::Middl
 
 module HealthQuest
   class PGDService < HealthQuest::SessionService
-    def get_pgd_resource(type, id = nil, pagination_params = {}) 
+    def get_pgd_resource(type, id = nil, pagination_params = {})
       res_id = id || '333'
       fhir_data = { id: res_id.to_s, text: "this is the #{type}" }
       {
@@ -29,7 +29,7 @@ module HealthQuest
         pagination: {
           current_page: pagination_params[:page] || 0,
           per_page: pagination_params[:per_page] || 0,
-          total_pages: 0, 
+          total_pages: 0,
           total_entries: 0
         }
       }
--- a/modules/health_quest/app/services/health_quest/pgd_service.rb
+++ b/modules/health_quest/app/services/health_quest/pgd_service.rb
@@ -1,0 +1,62 @@
+# frozen_string_literal: true
+
+# rubocop:disable all
+
+Faraday::Response.register_middleware health_quest_errors: HealthQuest::Middleware::Response::Errors
+Faraday::Middleware.register_middleware health_quest_logging: HealthQuest::Middleware::HealthQuestLogging
+
+module HealthQuest
+  class PGDService < HealthQuest::SessionService
+    def get_pgd_resource(type, id=nil, pagination_params = {})
+
+      params = {}
+
+      {
+        data: deserialized_resource({my_attr: 'none/fake', fhir_data: {id: '333', text: "this is the #{type}"}}, type),
+        meta: pagination(pagination_params)
+      }
+#      with_monitoring do
+#        response = perform(:get, get_pgd_base_url(type, id), params, headers, timeout: 55)
+#        {
+#          data: deserialized_resource(response.body, type),
+#          meta: pagination(pagination_params)
+#        }
+#      end
+    end
+
+    private
+
+    def deserialized_resource(json_hash, type)
+      result = json_hash[:fhir_data]
+      return [] unless result
+
+      result.is_a?(Array) ? result.map { |el| OpenStruct.new(el) } : OpenStruct.new(result)
+    end
+
+    # TODO: need underlying APIs to support pagination consistently
+    def pagination(pagination_params)
+      {
+        pagination: {
+          current_page: pagination_params[:page] || 0,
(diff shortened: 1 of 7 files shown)


modules/health_quest/spec/services/pgd_service_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    it 'gets a PGD resource' do
+      expect(subject.get_pgd_resource(:questionnaire)[:data]).to eq(dummy_response.body[:data])
+    end
+  end
+
+  describe '#get_pgd_base_url' do
+    it 'gets a url' do
+      # ubocop:disable Layout/LineLength
+      expect(subject.send(:get_pgd_base_url, 'questionnaire')).to eq("/questionnaire/v1/patients/#{user.icn}")
+      # ubocop:enable Layout/LineLength
```
kfrz, line 29:
```suggestion
      url = subject.send(:get_pgd_base_url, 'questionnaire')
      expect(url).to eq("/questionnaire/v1/patients/#{user.icn}")
```

Perhaps some simple assignment like this is more readable and you can avoid skipping the cop

modules/health_quest/spec/services/pgd_service_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  let(:dummy_response) { double('fake_response', body: { data: expected_data }) }
+
+  before do
+    allow_any_instance_of(HealthQuest::UserService).to receive(:session).and_return(:dummy_session)
+    allow_any_instance_of(HealthQuest::SessionService).to receive(:perform).and_return(dummy_response)
+  end
+
+  describe '#get_pgd_resource' do
+    it 'gets a PGD resource' do
+      expect(subject.get_pgd_resource(:questionnaire)[:data]).to eq(dummy_response.body[:data])
```
kfrz, line 21:
```suggestion
      resource = subject.get_pgd_resource(:questionnaire)[:data]
      expect(resource).to eq(dummy_response.body[:data])
```

modules/health_quest/app/services/health_quest/pgd_service.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+# rubocop:disable all
```
kfrz, line 3: Is this a temporary move?

kfrz (changes requested):
If this is a WIP PR - prefer the WIP feature, unless there's a good reason not to. 

I'm unsure if this is ready for review/approval, sounds like not based on the main description.

modules/health_quest/app/services/health_quest/pgd_service.rb
massrb: Sorry, what happened is I had a PR that failed for no reason because danger/danger crashed so I redid a smaller PR and it looks like I pulled in some code that had the wrong rubocop disable

modules/health_quest/spec/services/pgd_service_spec.rb
massrb: applied this change

modules/health_quest/spec/services/pgd_service_spec.rb
massrb: applied this change

modules/health_quest/app/services/health_quest/pgd_service.rb
```diff
@@ -0,0 +1,51 @@
...
+        pagination: {
+          current_page: pagination_params[:page] || 0,
+          per_page: pagination_params[:per_page] || 0,
+          total_pages: 0,
+          total_entries: 0
+        }
+      }
+    end
+
+    def get_pgd_base_url(type, id = nil)
```
erluti, line 38: how is this method used?  It's in private but not used in the class

erluti (commented):
> For the health quest module - this adds a PGD (patient generated data) service. The service is not actually functional but our front end developer wanted to see the endpoints.

Since this is non-functional code, I feel like it should have a feature toggle on it.  I don't see how the FE is going to access this work.  Is there some incoming work that uses this service and exposes it to the FE?  That's probably where the feature toggle would go, I think.  That work should be included with this PR because I'm not comfortable merging unused and "unusable" code.  (I get it's purpose, but if something happened and no other work was ever done nobody would know why this code is here or what it's supposed to do.)

modules/health_quest/app/services/health_quest/pgd_service.rb
massrb: get_pgd_base_url() would be used subsequently when the service connects to another backend system. Since the service is nonfunctional but the front end person wants to use it just to verify the endpoint is there or dummy results are returned, then it isn't being used other than the spec tests check it but I will have to remove it from the PR. 

The way the feature toggle works is that in the policy method, if it's not enabled then when the endpoint gets accessed it will return not authorized. I am going to add the feature being checked for to be "show_healthcare_experience_questionnaire" as currently it is using "va_online_scheduling"

erluti: I don't understand how the flipper interacts with the PGDService you're introducing

massrb: This PR doesn't have an endpoint for anything PGD, I will have to add that in a subsequent PR. Currently there is no way that anything can interact with the PGD service because there is not yet an associated endpoint. The feature toggle using Flipper.enabled?(<feature name>) works with endpoints only by checking for authorization. So the flipper should be working with endpoints and not the internal service.

erluti: So you just described my concerns with merging this PR.  It's okay to make this part of a bigger PR that introduces the endpoint that uses this information.  Or better yet, create the endpoint pr first but return something like `render json: {data: 'dummy response'}` and then do this PR and connect them up.  But these changes aren't being used by anything and so wouldn't be a "viable" product (in a MVP sense).  So this puts the code-base at risk if the endpoint is never created to use this or (more likely) when the endpoint is created this code dosen't fit the way it needs to be used and has to be changed anyway.

massrb: ok, well I think we can just abandon the dummy data strategy for now. I will push up other changes

massrb: I have worked out a method to build a proxy to imitate an API using monteblank, that looks promising so the dummy data should not be needed. I have pushed up the service code without it returning dummy data

modules/health_quest/spec/services/pgd_service_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  before do
+    allow_any_instance_of(HealthQuest::UserService).to receive(:session).and_return(:dummy_session)
+    allow_any_instance_of(HealthQuest::SessionService).to receive(:perform).and_return(dummy_response)
+  end
+
+  describe '#get_pgd_resource' do
+    it 'gets a PGD resource' do
+      resource = subject.get_pgd_resource(:questionnaire)[:data]
+      # expect(resource).to eq(dummy_response.body[:data])
```
erluti, line 21: let's remove this commented out code

erluti (approved):
LGTM, just one note

kfrz (approved):

PR #4936, opened 2020-09-21
Turn `profile_show_profile_2.0` on by default

erikphansen:
## Description of change

This flag should have been turned on as part of https://github.com/department-of-veterans-affairs/vets-api/pull/4277

## Original issue(s)


kfrz (approved):
:+1: Good catch!

PR #4938, opened 2020-09-21
remove obsolete feature toggles

annaswims:
## Description of change
when a dev deletes a feature toggle in the code it should go away in the UI as well.  

For example in staging there's both ` show_526_wizard` and `show526_wizard`, due to a renaming.

for @Mottie


kfrz (approved):
Think this looks good. Did you test locally?

annaswims:
yes, it was tested locally.

PR #4941, opened 2020-09-21
added logging for type_of_care in appointment_requests_controller

grossb1:
## Description of change

Updated appointment request controller to log type_of_care_id

## Original issue(s)
https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/13760

Respective unit tests were updated to reflect this change

### DIFF
--- a/modules/vaos/app/controllers/vaos/v0/appointment_requests_controller.rb
+++ b/modules/vaos/app/controllers/vaos/v0/appointment_requests_controller.rb
@@ -57,7 +57,8 @@ module VAOS
           'VAOS AppointmentRequest',
           action: params[:action],
           type: params[:type].is_a?(String) ? params[:type].upcase : params[:type],
-          id: response[:data].try(:unique_id)
+          id: response[:data].try(:unique_id),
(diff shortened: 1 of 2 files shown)


saneshark:
We still want the ID too, but we also want the type_of_care_id. It looks like you removed ID.

grossb1:
Where did I remove Id?

kfrz (commented):
Just a gut-check: there's no correlated information in the logs that might allow someone to associate a given patient with type of care requested? Just want to be extra sure we're not logging more than we need to in that regard. 

cc @f1337

saneshark:
> Just a gut-check: there's no correlated information in the logs that might allow someone to associate a given patient with type of care requested? Just want to be extra sure we're not logging more than we need to in that regard.
> 
> cc @f1337

I think users account ID or UUID is always part of the logs, their remote IP address is also part of the logs. None of the details we're logging are any more specific to a user than those. That being said, all the reports that I pull deliberately exclude these additional attributes from the CSV files I generate.

saneshark (approved):
LGTM

PR #4942, opened 2020-09-21
[API 2569] Report All Upstream Status Dependencies in our Healthchecks - Appeals

templeman15:
## Description of change
This PR:

1) Updates the existing appeals API `healthcheck` endpoint to report the status of the API without any upstream dependencies
2) Creates a new endpoint `upstream_healthcheck`, which reports the status of the API with all the upstream dependencies. 

## Original issue(s)
[Report All Upstream Status Dependencies in our Healthchecks - Appeals](https://vajira.max.gov/browse/API-2569)


bastosmichael (approved):
Not sure @ryanlink wants us calling this the downstream healthcheck but I'm good with it if he is. Looks like you just need to improve test coverage.

ryanlink (changes requested):
Yeah can we not use downstream in the endpoint name please if the dependencies we're checking are actually upstream?

templeman15:
> Yeah can we not use downstream in the endpoint name please if the dependencies we're checking are actually upstream?

Renamed to `upstream_healthcheck`.

ryanlink (approved):

modules/appeals_api/app/controllers/appeals_api/metadata_controller.rb
```diff
@@ -41,40 +41,45 @@ def appeals_status
...
         }
-      }.to_json
+      }, status: health_checker.services_are_healthy? ? 200 : 503
     end
 
-    def unhealthy_service_response
+    private
+
+    def upstream_service_details(service_name, health_checker, time)
+      healthy = health_checker.send("#{service_name.snakecase}_is_healthy?")
```
erluti, line 71: you can avoid using metaprogramming  if this line was like `healthy = health_checker.healthy_service?(service_name)`
and then you had the method in health_checker like this:
```
    def healthy_service?(service)
      case service
      when /caseflow/i
        caseflow_is_healthy?
      else
        raise "AppealsApi::HealthChecker doesn't recognize #{service}"
      end
    end
```

modules/appeals_api/config/routes.rb
```diff
@@ -1,11 +1,14 @@
 # frozen_string_literal: true
 
-AppealsApi::Engine.routes.draw do
+AppealsApi::Engine.routes.draw do # rubocop:disable Metrics/BlockLength
```
erluti, line 3: This shouldn't be necessary because the modules routes files are supposed to ignore BlockLength
I'm looking into why the settings in `.rubocop.yml` aren't applying here

erluti (commented):
So glad you're including tests for the health checker!

modules/appeals_api/config/routes.rb
erluti: Ah ha!  AppealsApi has it's own rubocop.yml! `modules/appeals_api/.rubocop.yml`
update that to look like this:
```
Metrics/BlockLength:
  Exclude:
    - spec/**/*.rb
    - 'config/**/*.rb'
```
and then you don't need to disable rubocop here

modules/appeals_api/app/controllers/appeals_api/metadata_controller.rb
templeman15: Thanks for the suggestion, updated!

modules/appeals_api/config/routes.rb
templeman15: I was also wondering why this was needed here and not other routes files. Thanks and updated!

modules/appeals_api/app/controllers/appeals_api/metadata_controller.rb
```diff
@@ -68,7 +68,7 @@ def upstream_healthcheck
     private
 
     def upstream_service_details(service_name, health_checker, time)
-      healthy = health_checker.send("#{service_name.snakecase}_is_healthy?")
+      healthy = health_checker.healthy_service?(service_name.snakecase)
```
erluti, line 71: I'd drop the `snakecase` here since the regex in `healthy_service?` matches ignoring case and the service_names passed in are based on `AppealsApi::HealthChecker::SERVICES` I'd assume they wouldn't need to be massaged anyway

erluti (approved):

modules/appeals_api/app/controllers/appeals_api/metadata_controller.rb
templeman15: @erluti ready to merge? Thanks for your review.

PR #4943, opened 2020-09-21
13281 clean up requires in specs

erluti:
Based on this comment https://github.com/department-of-veterans-affairs/vets-api/pull/4897#issuecomment-696173520, all the requires used in specs were analyzed.  The result is mostly removing them when unneeded.  In a few places they were moved to the appropriate place for the `require`.


LindseySaari (approved):
Nice! 💯 Thanks for the cleanup! 🧼

PR #4944, opened 2020-09-22
Update features.yml with better VAOS dev defaults

jbalboni:
## Description of change
Now that features are off by default in dev, we need to set which VAOS flags we do want on by default. This makes those updates.


erluti (approved):
👍

PR #4945, opened 2020-09-22
Catching any BGS errors when submitting 0966

bastosmichael:
## Description of change
While testing burial I realized that we didn't catch any soap responses from BGS that errored out in submission outside of our validations when it errors at BGS.

## Original issue(s)
https://vajira.max.gov/browse/API-1242


templeman15 (approved):
Looks good!

PR #4947, opened 2020-09-22
Update VAOS logger to use request body/header to log JTI

grossb1:
## Description of change
This issue is looking to enhance our ability to coordinate with VAMF when tracking down issues a user may encounter when interacting with VAOS. Prior to this change the code was logging the JTI present on the response object.  This was problematic if a response wasn't present due to a down-stream error.  In this case, we would lose the ability to identify a user to VAMF because we wouldn't have a JTI.  The solution was to record the JTI present in the request body/header (as opposed to the response) so in the case of a down-stream error, we would be able to access the JTI associated with the request which could then be presented to VAMF to facilitate identifying the error.

The `token` variable was also memoized in the jwt_wrapper as an optimization.

Also, we noticed that we’re using two different variations of the header name `X-Vamf-Jwt` and `X-VAMF-JWT`.  We will fix this as part of a future PR and tag the ticket number that will address that in the comment.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/6837

Unit tests were updated to reflect this change


kfrz:
@grossb1 please use the WIP (draft) PR functionality for in-progress work.

Also your commits are empty :man_shrugging:

modules/vaos/spec/services/middleware/vaos_logging_spec.rb
```diff
@@ -10,13 +10,17 @@
...
+          all_other_uris_with_header_caps
+        ) { [status, { 'Content-Type' => 'text/plain', 'X-VAMF-JWT' => sample_jwt }, '{}'] }
         stub.post(user_service_uri) { [status, { 'Content-Type' => 'text/plain' }, sample_jwt] }
       end
     end
   end
 
   let(:sample_jwt) { read_fixture_file('sample_jwt.response') }
   let(:all_other_uris) { 'https://veteran.apps.va.gov/whatever' }
+  let(:all_other_uris_with_header_caps) { 'https://veteran.apps.va.gov/with_header_caps' }
```
saneshark, line 23: NITPICK: could call this `user_service_refresh_url` just to call attention to where and why we need this additional test.

modules/vaos/app/services/vaos/middleware/vaos_logging.rb
```diff
@@ -72,17 +72,24 @@ def user_session_request?(env)
...
         end
       rescue
         'unknown jti'
       end
+
+      # #x_vamf_headers identifies which X-Vamf-Header was set and returns the appropriate header value
+      #
+      # @param request_headers The set of request headers
+      # @return [String] the JWT set in the request headers
+      def x_vamf_headers(request_headers)
```
saneshark, line 90: This is good, but based on some conversation I had I think we'll be able to get rid of this in a future PR. That future PR will involve changing all of our other headers to use the capital version as well. Therefore it will mean basically every single VCR Cassette will need to be manually changed as well. As a result I think it will be best to tackle as a separate tech debt item in the future.

If we have a ticket number for that work we could link to it here.

saneshark (approved):
A few minor comments and nitpicks, but LGTM. We should do a quick test on staging and verify before the production deploy that everything is working as expected.

PR #4948, opened 2020-09-22
Add BGS Regional Office Lookup by Zip Code

kathleencrawford:
## Description of change

This PR adds the ability to look up a regional office location number based on the Veteran's zip code.  This location number is required when submitting dependency claims (Form 686c - adding or removing a dependent).  We are currently using the bgs_ext gem for integration with BGS.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#12763

There are no additional configuration or settings added for this feature.  The process flow for submitting the 686c was already in place, we are just adding this task as part of the overall workflow.  Sentry logging has been added as a warning in case the call to BGS fails to return a location number.  

- [x] Local testing


lib/bgs/service.rb
```diff
@@ -164,6 +164,16 @@ def user_ssn
       { ssn: @user.ssn }
     end
 
+    def get_regional_office_by_zip_code(zip_code, country, province, lob, ssn)
```
erluti, line 167: can you include a test for this method in `spec/lib/bgs/service_spec.rb`?

kathleencrawford: @erluti thank you for the suggestion - i've added a test for this as requested.

erluti (approved):
thanks for adding that test!

PR #4956, opened 2020-09-22
increment statsd on success of form526 confirmation email job

lingtran:
## Description of change
This is a small addition to the Form526ConfirmationEmailJob, which is to increment statsd on success. We've previously only incremented on failure.

## Original issue(s)
department-of-veterans-affairs/notification-api#137

Increment statsd on success for Form526ConfirmationEmailJob


kfrz (approved):
:+1: 

I don't like this having to be manually configured for each worker, but this is fine for now until we overhaul metrics a bit.

PR #4959, opened 2020-09-23
13349 military info doc

jperk51:
## Description of change
Api docs for Military Information mobile endpoint

## Original issue(s)
department-of-veterans-affairs/va.gov-team#13349

Mobile Api swagger doc update

N/A


kreek:
@thilton-oddball not that we don't value your input 😄 but I think that PRs that only affect engine code shouldn't need VSP backend approval? Do you know how we get that configured for the mobile engine?

erluti (approved):

modules/mobile/docs/openapi.yaml
```diff
@@ -19,11 +19,67 @@ paths:
...
\ No newline at end of file
+        branchOfService:
+          type: string
+          example: "United States Marine Corps"
+        beginDate:
+          type: string
+          example: "September 17, 1997"
+        endDate:
+          type: string
+          example: "December 31, 2002"
```
jjulian, line 85: Should we document our decision that dates are not to be parsed?

kreek: I think that's a good idea. Do you have thoughts on where that documentation would live? In the description for this endpoint? Or maybe the description field for the entire API?

kreek:
The yaml looks good, @jperk51 please generate the static html and include it in this PR so it's up-to-date

modules/mobile/docs/openapi.yaml
kreek: The more I think about this there's a reason ISO exists and we may be loading up a foot gun by pre-formatting dates. I think we should standardize on that.

jperk51: Sure, just to be clear on what will be happening, the current service history endpoint returns the dates in YYYY-MM-DD format, but the website displays them in the [Month] DD, YYYY format. In order to avoid having to do that translation in the app, and avoid needing to release new app versions for fixing/changing things like that, pre-formatting display strings can be useful. Let me know how/where I should go about documenting this decision, as long as everyone is on board with it, and I'll get on that asap

kreek: Sorry for the back and forth on this @jperk51 took a look at vets-api and lighthouse and they (for the most part) use ISO8601 date format so I think we should stick to that (profile uses ISO date exclusively). I understand wanting to keep as much business logic, including formatting, server side for the mobile app but in this case I think we're covered by the design system/content rules always using [Month] DD, YYYY so that's unlikely to change. Once we get to things like appointment times or other dates that need to be updated it will be easier to know that they're always ISO8601 when coming and going through the API. Please update the example to be YYYY-MM-DD.

kreek (approved):
@jperk51 if you want to add a description to the date fields that these are pre-formatted for display e.g.
```yaml
 beginDate:
  type: string
  description: ...
  example: "September 17, 1997"
```
 then this LGTM 🚀

jjulian (approved):

PR #4961, opened 2020-09-23
va.gov-team#13813: va.gov top level domain saml request cookie

benniemosher:
## Description of change

We are having some issues "tying" together our SSOe requests after they leave our system and enter other systems. We have agreed with the other teams to set a top-level domain cookie that will have a transaction_id (etc) in it to tie them together. This PR is my humble attempt at creating that cookie and testing it's existence.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#13813

- `settings.local.yml` files for each environment will need a to be updated


app/controllers/concerns/authentication_and_sso_concerns.rb
```diff
@@ -112,6 +112,24 @@ def set_sso_cookie!
...
+    Rails.logger.info('SSO: ApplicationController#set_sso_cookie!', sso_logging_info)
+
+    cookies[Settings.sso.saml_cookie_name] = {
+      value: sso_saml_cookie_content,
+      expires: nil, # NOTE: we track expiration as an attribute in "value." nil here means kill cookie on browser close.
+      secure: Settings.sso.cookie_secure,
+      httponly: true,
+      domain: Settings.sso.cookie_domain
+    }
+  end
```
f1337, line 131: Looking good! To improve test-ability, perhaps consider the following refactor:

```
def set_sso_saml_cookie!
    SSO::SAMLCookie.from(@session_object, @current_user)
end
```

and then in another file (maybe in lib?)

```
module SSO
    class SAMLCookie
        def self.from(session_object, current_user)
            # …{refactored from set_sso_saml_cookie}
        end

        def self.cookie_content
            …{refactored from sso_saml_cookie_content}
        end
    end
end
```

app/controllers/concerns/authentication_and_sso_concerns.rb
```diff
@@ -130,6 +148,17 @@ def sso_cookie_content
     }
   end
 
+  def sso_saml_cookie_content
+    return nil if @current_user.blank?
```
f1337, line 152: Whether you perform the refactor I suggested above or not, this could move up into the other method. I'll post a comment where.

app/controllers/concerns/authentication_and_sso_concerns.rb
```diff
@@ -112,6 +112,24 @@ def set_sso_cookie!
...
+    return unless Settings.sso.cookie_enabled &&
+                  @session_object.present? &&
+                  # if the user logged in via SSOe, there is no benefit from
+                  # creating a MHV SSO shared cookie
+                  !@current_user&.authenticated_by_ssoe
+
+    Rails.logger.info('SSO: ApplicationController#set_sso_cookie!', sso_logging_info)
+
+    cookies[Settings.sso.saml_cookie_name] = {
+      value: sso_saml_cookie_content,
```
f1337, line 125:
```suggestion
      value: (@current_user.present? ? sso_saml_cookie_content : nil),
```

benniemosher: done

app/controllers/concerns/authentication_and_sso_concerns.rb
benniemosher: done

app/controllers/v1/sessions_controller.rb
```diff
@@ -138,23 +138,49 @@ def user_login(saml_response)
...
+                         JSON.parse(cookies[Settings.ssoe_eauth_cookie.name])['transaction_id']
+                       else
+                         SecureRandom.uuid
+                       end
+
+      {
+        'timestamp' => Time.now.iso8601,
+        'transaction_id' => transaction_id,
+        'saml_request_id' => url_service.tracker&.uuid,
+        'saml_request_query_params' => url_service.query_params
```
jholton, line 180: @ericbuckley does this look right for `saml_request_query_params`?  Any "query_params" we should exclude?

app/controllers/v1/sessions_controller.rb
```diff
@@ -138,23 +138,49 @@ def user_login(saml_response)
...
+        value: saml_cookie_content.to_json,
+        expires: nil,
+        secure: Settings.ssoe_eauth_cookie.secure,
+        httponly: true,
+        domain: Settings.ssoe_eauth_cookie.domain
+      }
+    end
+
+    def saml_cookie_content
+      transaction_id = if current_user && url_service.should_uplevel?
```
ericbuckley, line 170: cool, so 1 transactionid for the entire authentication process, not just each leg of our SAML request/response trips.

config/settings.yml
```diff
@@ -30,18 +30,23 @@ saml_ssoe:
   response_signing: false
   response_encryption: false
 
+ssoe_eauth_cookie:
```
ericbuckley, line 33: I take it we'll need a PR to update the prod/staging/dev config files in the devops repo?

app/controllers/v1/sessions_controller.rb
jholton: Yeah, that's the idea based off of your comment in the associated issue:
> transaction id for user authentication (should generate a new value for this, unless we are upleveling in which case the previous transaction id should be used)

Does this make sense in that context?

config/settings.yml
jholton: Yeah.  I have those updates ready locally.  Just need to open the PR

app/controllers/v1/sessions_controller.rb
ericbuckley: 😄 I forgot that I wrote that!

ericbuckley (approved):
LGTM

jholton:
[Devops Settings PR](https://github.com/department-of-veterans-affairs/devops/pull/7695)

app/controllers/v1/sessions_controller.rb
```diff
@@ -138,23 +138,49 @@ def user_login(saml_response)
     end
 
     def render_login(type)
-      login_url, post_params = login_params(type)
+      login_url, @post_params = login_params(type)
```
erluti, line 141: why change this to an instance variable?

erluti (approved):
Approving, but if `@post_params` doesn't need to be an instance variable, let's remove that `@`

app/controllers/v1/sessions_controller.rb
jholton: Good catch.  That's left over from a previous refactoring that I forgot to revert.

PR #4963, opened 2020-09-23
modify staccato to use google analytics forward proxy

lihanli:
## Description of change
modify staccato library to send requests through the google analytics forward proxy instead of directly to google analytics

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/13856

tested in staging

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -572,6 +572,9 @@ github_stats:
   username: github-stats-rake
   token: fake_token
 
+google_analytics:
+  url: 'https://internal-dsva-vagov-staging-fwdproxy-1821450725.us-gov-west-1.elb.amazonaws.com:4473'
+
 # Settings for the coronavirus chatbot tokens
 coronavirus_chatbot:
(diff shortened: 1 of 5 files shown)


config/settings.yml
```diff
@@ -572,6 +572,9 @@ github_stats:
   username: github-stats-rake
   token: fake_token
 
+google_analytics:
```
erluti, line 575: settings.yml has a reference to `google_analytics_tracking_id` on line 457.  That should probably move in with this group, it's only used in a few places.

erluti: just occured to me that would require a corresponding change to devops.
hopefully the keys could be side-by-side if changing devops is too complicated

lihanli: thanks for the suggestion i made a new issue for it https://github.com/department-of-veterans-affairs/va.gov-team/issues/14017

erluti (approved):
thanks for creating that ticket

PR #1431, opened 2017-10-05
SAML Certificate Rollover

omgitsbillryan:
Turns our `/saml/metadata` endpoint from having 1 certificate, to 2.  Example:

```xml
<?xml version="1.0"?>
<md:EntityDescriptor xmlns:md="urn:oasis:names:tc:SAML:2.0:metadata" ID="_bbd79f27-f6d8-4379-b017-fe7200d31737" entityID="saml-rp.vetsgov.localhost">
  <md:SPSSODescriptor AuthnRequestsSigned="true" WantAssertionsSigned="false" protocolSupportEnumeration="urn:oasis:names:tc:SAML:2.0:protocol">
    <md:KeyDescriptor use="signing">
      <ds:KeyInfo xmlns:ds="http://www.w3.org/2000/09/xmldsig#">
        <ds:X509Data>
          <ds:X509Certificate>
[...]

PR #4965, opened 2020-09-24
multi-key saml support when decrypting authn responses

omgitsbillryan:
## Description of change
Adds support for a concurrent 2nd saml key for ID.me auth. 

## Why?
Our existing certificate is expiring soon, and our private key has not changed for a while, and our existing certificate still has "vetsgov" references in it... so we generated a brand new keypair. We don't want to incur any login downtime, so this PR makes `vets-api` support both keys for login. Once ID.me makes the new certificate "active", we can remove this logic and just use the new key + cert.

## Extra Context
`ruby-saml` has support for multiple certificates ([and it's what we used last time we rolled over](https://github.com/department-of-veterans-affairs/vets-api/pull/1431)), but that support requires both old and new cert to cryptographically chain back to the same private key. In our case, we have a _new_ private key.

## Original issue(s)

### DIFF
--- a/lib/saml/responses/login.rb
+++ b/lib/saml/responses/login.rb
@@ -10,11 +10,13 @@ module SAML
       def initialize(saml_response, options = {})
         super(saml_response, options)
       rescue OpenSSL::PKey::RSAError => e # "padding check failed." when decrypt fails
+        # :nocov: Temporary code only required during key rollovers using a new private key
         raise e unless Settings.saml.key_new
 
         duped_options = options.dup
         duped_options[:settings].private_key = Settings.saml.key_new
         super(saml_response, duped_options)
+        # :nocov:
       end
 
       def errors_message
--- a/lib/saml/responses/login.rb
+++ b/lib/saml/responses/login.rb
@@ -7,16 +7,14 @@ module SAML
     class Login < OneLogin::RubySaml::Response
       include SAML::Responses::Base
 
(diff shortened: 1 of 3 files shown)


lib/saml/responses/login.rb
```diff
@@ -7,6 +7,18 @@ module Responses
...
+      def initialize(saml_response, options = {})
+        super(saml_response, options)
+      rescue OpenSSL::PKey::RSAError => e # "padding check failed." when decrypt fails
+        # :nocov: Temporary code only required during key rollovers using a new private key
+        raise e unless Settings.saml.key_new
+
+        duped_options = options.dup
+        duped_options[:settings].private_key = Settings.saml.key_new
+        super(saml_response, duped_options)
+        # :nocov:
```
omgitsbillryan, line 19: I took a crack at writing a spec to cover this code instead but decided against it because:
1. it requires keys, both invalid and valid one and we [consciously removed our localhost key](url)
2. it would also require an encrypted authresponse, which again, we cannot decrypt (see 1)
3. the code is meant to be temporary - it should only exist for the duration of keyrollver... probably 1-2 days
4. I tried using rspec mocks, `allow(OneLogin::RubySaml::Response).to receive(:new).and_raise(OpenSSL::PKey::RSAError)` - but this is effectively mocking `super` which doesn't really work.

kfrz (approved):
:+1:

PR #4966, opened 2020-09-24
Bump vets-json-schema version

batemapf:
## Description of change
Implementing schema changes: https://github.com/department-of-veterans-affairs/vets-json-schema/pull/520.
Corresponding vets-website changes: https://github.com/department-of-veterans-affairs/vets-website/pull/14357

## Original issue(s)
department-of-veterans-affairs/va.gov-team#13900


settings.local.yml
```diff
@@ -0,0 +1,2 @@
+saml:
```
erluti, line 1: this file shouldn't be checked in

erluti (changes requested):

settings.local.yml
batemapf: ahhhhh that's it. totally forgot to ignore that one 🤦. already went ahead and created a clean PR over [here](https://github.com/department-of-veterans-affairs/vets-api/pull/4969).

PR #4970, opened 2020-09-24
Task/api 2170 vv utilize mock bgs

esubocz-va:
## Description of change
Veteran Verification has been edited to allow use of mock-bgs independent of other apps utilizing the BGS service. 

## Original issue(s)
https://vajira.max.gov/browse/API-2170

`mock_bgs` is a new configuration that can be set in the bgs section of your `settings.local.yml file`. it will default to `nil` if not set, and use the tunnel. so, no changes are required to any other apps. this will affect veteran verification only.

in order to interact with the mock-bgs service, you will need to have the app running locally. repo found here: https://github.com/department-of-veterans-affairs/health-apis-mock-benefits-gateway-services and the following changes will need to be made to your `settings.local.yml` file:
```
vet_verification:
  url: http://localhost:9090/v0
  mock_bgs: true
```

local testing has been done to ensure functionality remains the same when using the mock-bgs service. as well as that this causes no breaking changes for standard use of the tunnel.


modules/veteran_verification/app/services/bgs/disability_rating_service.rb
```diff
@@ -1,13 +1,59 @@
...
-        external_key: current_user.email
-      )
-      service.rating.find_rating_data(current_user.ssn)
+      if Settings.bgs.mock_bgs == true
+        options = {
+          wsdl: "#{Settings.bgs.url}/RatingServiceBean/RatingService/ratingRecord.wsdl",
+          soap_header: header(current_user),
+          endpoint: "#{Settings.bgs.url}/RatingServiceBean/RatingService"
+        }
+        @client ||= Savon.client(options)
```
erluti, line 15: I think there's too much going on in this section for just mocking a response.  It's also not very reusable for anywhere else we might want to mock BGS.  I think this should be lifted out into it's own class, perhaps `BGS::MockServices`?
So I think it would be neat if this method looked more like this:
```
 def get_rating(current_user)
      service = BGS.services_object(current_user)
      service.rating.find_rating_data(current_user.ssn)
end
```
then there would be a method in the `BGS` module space like this:
```
module BGS
  def self.services_object(current_user)
    if Settings.bgs.mock_bgs
      MockServices.new(current_user)
    else
      Services.new(
        external_uid: current_user.icn,
        external_key: current_user.email
      )
    end
  end
end
```
Then all this logic of instantiating a savon client and building a header and all this stuff can be tucked away in that class and keep this code simple.

esubocz-va: In the interest of single responsibility classes, I think you're right and I added a commit to define a mock-disability-rating-service separate from disability-rating-service code.  However, to be clear about mock-bgs, it is _only_ for disability rating data at this time - which is specific to veteran verification.  If it were to be expanded to mock other BGS endpoints, that is when we would like to move it to a more global level to prevent misunderstandings.

After discussing with my team (Pivot!) we believe the approach in my latest commit would be the best way to go. We are open to hearing your thoughts though.

kfrz: I'm curious - is there any reason not to sub-namespace the `bgs_disability_rating` service, such as in the `evss:` block of [`settings.yml`](https://github.com/department-of-veterans-affairs/vets-api/blob/master/config/settings.yml#L182)? 

Could be:

```yml
# ...
bgs:
  mock_disability_rating: false
# ...
```

kfrz: Oh I see, it's not data that can be mocked using betamocks, it calls an actual external service... Hmm. I don't think we have an established pattern for this yet, but I do second Philip's comment about tidying up the state

modules/veteran_verification/app/models/veteran_verification/disability_rating.rb
```diff
@@ -15,7 +15,11 @@ class DisabilityRating
     attribute :individual_ratings, Array
 
     def self.rating_service
-      BGS::DisabilityRatingService.new
+      if Settings.bgs.mock_bgs == true
```
beaugrantham-va, line 18: Should this be a setting under vet_verification or renamed to reflect that it's only mocking the disability rating service?

modules/veteran_verification/app/services/bgs/mock_disability_rating_service.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+require 'savon'
+require 'nokogiri'
+
+module BGS
+  class MockDisabilityRatingService
+    def get_rating(current_user)
+      options = {
+        wsdl: "#{Settings.bgs.url}/RatingServiceBean/RatingService/ratingRecord.wsdl",
```
beaugrantham-va, line 10: Same comment... if BGS is used by multiple modules, does this setting need to be local to vet_verification?

dbrown-va (approved):
Recommend enabling only in Dev for some additional testing before pointing Sandbox to the mock service.

PR #4971, opened 2020-09-24
11603 v1 facilities swagger

darkbushido:
## Description of change
Adding swagger docs for the V1 api

## Original issue(s)
department-of-veterans-affairs/va.gov-team#11603

* There weren't any v1 swagger docs before this
* I broke the testing up into smaller pieces, its something I think we should do to the V0 apivore tests.

Minimal rspec testing


kfrz (approved):
This looks good to me overall, danger-bot doesn't like the length but hard to avoid w/ swagger changes.

PR #4974, opened 2020-09-25
Prevent MPI searches where SSN not present, on 10-10CG submission

kevinmirc:
**Description**
We currently take each person on a 10-10CG submission and perform a profile search in MPI with their attributes. Now that SSNs are not required for caregivers, we search MPI without the SSN. This causes the `MVI::Service` to raise a `MVI::Errors::InvalidRequestError`, making the submission fail.

This PR prevents the search of a caregiver if their SSN was not provided.

**Summary**
- [x] Update `Form1010cg::Auditor` to allow three results for MPI search logging: `:found`, `:not_found`, `:skipped`
- [x] Return `NOT_FOUND` when `Form1010cg::Service#icn_for` is called for a form subject that lacks an SSN.


erluti (approved):
LGTM

kevinmirc:
Thanks, @erluti!

PR #4979, opened 2020-09-25
Allowing dashes in X-VA-SSN for Claims

bastosmichael:
## Description of change
Sapient Management Group is running into a 500 error because they are adding dashes to the SSN and we need to fix this so that it strips out any non numerical characters out before the look up happens properly instead of 500.

This is the CURL:
```
curl -X GET 'https://sandbox-api.va.gov/services/claims/v1/claims' \
--header 'X-VA-SSN: 796-13-0115' \
--header 'X-VA-First-Name: Tamara' \
--header 'X-VA-Last-Name: Ellis' \
--header 'X-VA-Birth-Date: 1967-06-19' \
--header 'Authorization: Bearer eyJraWQiOiJtUTRCM2hn....[redacted]'

This is the response:
{
   "errors": [

{        "title": "Internal server error",        "detail": "Internal server error",        "code": "500",        "status": "500"    }
]
}
```

## Original issue(s)
https://vajira.max.gov/browse/API-2626

### DIFF
--- a/modules/claims_api/app/controllers/claims_api/application_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/application_controller.rb
@@ -118,8 +118,8 @@ module ClaimsApi
 
     def veteran_from_headers(with_gender: false)
       vet = ClaimsApi::Veteran.new(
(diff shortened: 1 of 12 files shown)


templeman15 (approved):
Looks good 👍

PR #4980, opened 2020-09-26
Mobile API docs: update user, add address update

kreek:
## Description of change
Updates the user endpoint, and adds the upcoming address update (async), and update transaction (async check) endpoints. The user response includes services available to the user, and its meta field lists the complete set of services the API provides. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#13346

* Documentation only
* The rendered static HTML can be viewed [here](https://htmlpreview.github.io/?https://github.com/department-of-veterans-affairs/vets-api/blob/dc7f043bb0e6db2d03014eed88c89063c48be137/modules/mobile/docs/index.html) (although for the JS on the page to work you'll need to open the file locally).


modules/mobile/docs/openapi.yaml
```diff
@@ -1,52 +1,358 @@
...
+                        type: array
+                        example:
+                          - appeals
+                          - appointments
+                          - claims
+                          - directDepositBenefits
+                          - letters
+                          - militaryServiceHistory
+                          - userProfileUpdate
+        '401':
```
kreek, line 54: Looked through the OpenAPI docs/stack overflow and there doesn't seem to be a way to reuse these error codes (you can just ref the schema as below)

jjulian (approved):

PR #4983, opened 2020-09-28
caregivers google analytics event

lihanli:
## Description of change
add google analytics event for caregivers invalid veteran status error

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/12089

swagger docs updated

tested on staging


app/services/form1010cg/service.rb
```diff
@@ -108,7 +109,10 @@ def submit_attachment # rubocop:disable Metrics/MethodLength
     #
     # @return [nil]
     def assert_veteran_status
-      raise InvalidVeteranStatus if icn_for('veteran') == NOT_FOUND
+      if icn_for('veteran') == NOT_FOUND
+        log_invalid_veteran_status
```
kevinmirc, line 113: Rather than being in the `Form1010cg::Service`, I think all if this logic could live in `Form1010cg::Auditor`. It already has a method there to log and increment StatsD for the same event: when the veterans status is not valid.

Something like:
```rb
# app/services/form1010cg/auditor.rb
# ...
  # ...
    def record_submission_failure_client_qualification(claim_guid:, veteran_name:)
      increment self.class.metrics.submission.failure.client.qualification
      log 'Submission Failed: qualifications not met', claim_guid: claim_guid, veteran_name: veteran_name
      notify_ga_invalid_veteran_status
    end
    # ...
    private

    def get_subdomain
      # ...
    end

    def notify_ga_invalid_veteran_status
      # ...
    end
  # ...
# ...
```

`#record_submission_failure_client_qualification` can just accept the `ga_client_id` as an extra argument without the need to attach it to the claim.

For context, this record method is called here: https://github.com/department-of-veterans-affairs/vets-api/blob/59dbf80673ac8802ada3f8768538426420bd6fa4/app/controllers/v0/caregivers_assistance_claims_controller.rb#L8

kevinmirc (changes requested):
I like the changes to the test suites!

Only change I can suggest is moving the GA logging from the `Form1010cg::Service` to the `Form1010cg::Auditor`, and as a result removing it as an attribute of the `CaregiversAssistanceClaim`. The comment has more details, let me know what you think.

Lastly, what happens if the `x-google-client-id` header is not present? We want to make sure that the request can still get processed and no error is raised if it's not present or we fail to connect to GA (just leaving it to best effort).

app/services/form1010cg/service.rb
lihanli: > Lastly, what happens if the x-google-client-id header is not present?

how about i don't log the event if the client id is not present?

kevinmirc: Yes, that works 👍

kevinmirc (approved):

kevinmirc:
Thanks, @lihanli! Please wait to merge until after our public launch on 10/01/2020.

app/services/form1010cg/auditor.rb
```diff
@@ -49,9 +50,11 @@ def record_submission_failure_client_data(errors:, claim_guid: nil)
       log 'Submission Failed: invalid data provided by client', claim_guid: claim_guid, errors: errors
     end
 
-    def record_submission_failure_client_qualification(claim_guid:, veteran_name:)
+    def record_submission_failure_client_qualification(claim_guid:, veteran_name:, ga_client_id:)
       increment self.class.metrics.submission.failure.client.qualification
       log 'Submission Failed: qualifications not met', claim_guid: claim_guid, veteran_name: veteran_name
+
+      notify_ga_invalid_veteran_status(ga_client_id) if ga_client_id.present?
```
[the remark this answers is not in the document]
lihanli: i added a spec for it here https://github.com/department-of-veterans-affairs/vets-api/pull/4983/files#diff-9d676d3081e023f3b6a98eae2d9cb8b5R159, do you think we need another one in the controller spec?

spec/services/form1010cg/service_spec.rb
```diff
@@ -40,6 +40,10 @@
     end
   end
 
+  def expect_veteran_not_found
```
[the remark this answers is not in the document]
lihanli: thanks, i changed the method name

app/services/form1010cg/auditor.rb
```diff
@@ -73,6 +76,35 @@ def log_mpi_search_result(claim_guid:, form_subject:, result:)
...
+        action: 'Forms - Family Member Benefits',
+        label: 'caregivers-error',
+        document_location: "https://#{get_subdomain}.va.gov/" \
+          'family-member-benefits/apply-for-caregiver-assistance-form-10-10cg/review-and-submit'
+      )
+      event.custom_metrics['cg1'] = 'Family Member Benefits'
+      event.custom_metrics['cg3'] = 'Family Member Benefits'
+      event.add_custom_dimension('30', 'Modernized')
+      event.add_custom_dimension('57', INVALID_VET_STATUS_CODE)
+      event.track!
```
kevinmirc, line 105: I'm assuming `#track!` is where the actual API request happens. If the request fails, does this raise an error?

We wouldn't want to raise an error if this fails, so that a proper response can still be sent to the user. I'm thinking it may be best to wrap `event.track!` in a `begin` `rescue` statement and just return `true` if it works and `false` if it fails.

app/services/form1010cg/auditor.rb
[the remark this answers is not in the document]
erluti: yeah, that seems to be a valid code path in the controller and it's untested

app/services/form1010cg/auditor.rb
```diff
@@ -73,6 +77,52 @@ def log_mpi_search_result(claim_guid:, form_subject:, result:)
...
+    end
+
+    def notify_ga_invalid_veteran_status(ga_client_id)
+      event = create_ga_event(ga_client_id)
+
+      begin
+        event.track!
+        true
+      rescue => e
+        if Rails.env.production?
```
erluti, line 117: Why is this production check here?  I'm concerned this adds complexity to the code and puts us in a position where production behaves differently then our test and development environments.  I would recommend changing this rescue-block to be the desired behavior in production, and set up a test where `event.track!` raises an exception and the test shows the intended behavior for `notify_ga_invalid_veteran_status` in that situation.

lihanli: i already added a spec for the `event.track!` error state. i added the production check because if an error happened in a test, i want it to get raised instead of silently failing. for example if the http request `event.track!` sends no longer matched the vcr cassette, and the production check was removed, then the spec that checks that would pass even though it should fail because the error is getting rescued.

erluti: this method is returning false in that case though.  Isn't that something that can be checked in tests?  I think it's okay if Test A still passes because there's no exception, but there should be a Test B that fails specifically because this method is returning false.  Or maybe add an additional `expect` to the test to handle this condition.
Otherwise this code would need a test like `#notify_ga_invalid_veteran_status raises an exception when the environment is not production` which would be complex to write and bizarrely tests that the tests are tests.
Looks like `notify_ga_invalid_veteran_status` is a private method, so the test should probably be for `record_submission_failure_client_qualification`.  I'd probably move the tests around `notify_ga_invalid_veteran_status` to be for that method and have some expectations that the results of `event.track!` occurred.

lihanli: ok thanks, i made the changes

erluti (approved):
thanks for listening to feedback.  looks good!

PR #4987, opened 2020-09-28
add missing requires for PPMS::ProviderFacility

erluti:
resolves issue discovered here https://dsva.slack.com/archives/CBU0KDSB1/p1601328320190800

### DIFF
--- a/app/models/ppms/provider_facility.rb
+++ b/app/models/ppms/provider_facility.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+require 'facilities/ppms/client'
+require 'lighthouse/facilities/client'
+
 class PPMS::ProviderFacility < Common::Base
   attribute :facilities, Array
   attribute :id, String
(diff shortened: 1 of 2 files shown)


kfrz (approved):
Shouldn't this be caught with a simple unit test? It's a subclass of `Common::Base` model 

@darkbushido can you add a spec to make sure this model can be instantiated correctly? 


```rb
require 'rails_helper'

RSpec.describe PPMS::ProviderFacility, type: :model do
  it 'is a PPMS::ProviderFacility object' do
    expect(described_class.new).to be_a(PPMS::ProviderFacility)
  end

  describe 'provider_ids' do
    it 'returns the correct provider ids' do
      pending 'todo'
    end
  end

  describe 'facility_ids' do
    it 'returns the correct facility ids' do
      pending 'todo'
    end
  end
end
```

erluti:
> Shouldn't this be caught with a simple unit test? It's a subclass of `Common::Base` model

this is a great suggestion, but I wasn't sure how to set up and add a basic test because `PPMS::ProviderFacility.new` fails without some args that I'm not sure about

PR #4988, opened 2020-09-29
Small Prefill Refactor: Stop Passing Around User

edmangimelli:
Currently, `FormProfile` 's initializer takes in `form_id` but not `user`.

`user` typically is passed in via the `prefill` method, and `prefill` passes it on to the methods it relies on. This enforces a procedural flow and encourages unnecessary overriding of methods (to get user info at a different point in the flow).

Had to refactor `initialize_contact_information` a little bit to keep rubocop happy.


erluti (approved):
LGTM
Is there anywhere that uses "prefill" to override user data already put in the form?
What's an example of something that was problematic because user was being passed around like this?

edmangimelli:
> LGTM
> Is there anywhere that uses "prefill" to override user data already put in the form?
> What's an example of something that was problematic because user was being passed around like this?

@erluti
this refactor sets the stage for my next pr. by taking in user earlier, i don't have to override prefill in my subclass (va0996) like so many subclasses are already doing just to make a change to the [prefill yaml](https://github.com/department-of-veterans-affairs/vets-api/blob/7e65903bc76218786fce975a6dafc1048e2a9f20/config/form_profile_mappings/20-0996.yml)

edmangimelli:
i think these are examples of overriding prefill just to get access to user info:

https://github.com/department-of-veterans-affairs/vets-api/blob/a265b230317e02682f0dfd61876bec0a7586cac8/app/models/form_profiles/va686c674.rb

https://github.com/department-of-veterans-affairs/vets-api/blob/169018e9bc884c7c1bdda30d485776f599726ce7/app/models/form_profiles/va0994.rb

https://github.com/department-of-veterans-affairs/vets-api/blob/3feb34ed9b20c26618661b236ad4c92a5d42ad2b/app/models/form_profiles/va526ez.rb

PR #4993, opened 2020-09-29
notification-api#77 handling more error types

saman-moshafi-va:
## Description of change
in sending notifications make sure the handle_error method handles all exceptions. For example of the notification client fails to instantiate an exception with no status_code is raised.

## Original issue(s)
department-of-veterans-affairs/notification-api#77

Added one unit test so if NoArgument error is raised the expected behavior is observed

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -136,7 +136,6 @@ PATH
 
 GEM
   remote: https://rubygems.org/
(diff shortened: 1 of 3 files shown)


saman-moshafi-va:
There is no need for this PR anymore

PR #4996, opened 2020-09-29
sentry tracking for SSOe Post Binding Form

ericbuckley:
Adding Sentry error tracking to the SSOe SAML post binding form to get a better understanding if any of our users are encountering JS errors.

closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/14019


jholton (approved):
LGTM, but I've never worked with Sentry on the FE before.  I'd tag the current FE engineer on support rotation for review too.

ericbuckley:
> LGTM, but I've never worked with Sentry on the FE before. I'd tag the current FE engineer on support rotation for review too.

@AGarcia-Clarity have you used sentry CDN integration before?  If so could you give this a review?

AGarcia-Clarity:
> > LGTM, but I've never worked with Sentry on the FE before. I'd tag the current FE engineer on support rotation for review too.
> 
> @AGarcia-Clarity have you used sentry CDN integration before? If so could you give this a review?

@ericbuckley have you tested to see if the CSP interferes with it. Other than that it looks good.

AGarcia-Clarity (approved):

ericbuckley:
> > > LGTM, but I've never worked with Sentry on the FE before. I'd tag the current FE engineer on support rotation for review too.
> > 
> > 
> > @AGarcia-Clarity have you used sentry CDN integration before? If so could you give this a review?
> 
> @ericbuckley have you tested to see if the CSP interferes with it. Other than that it looks good.

Yes locally I ran a test.  I temporarily hard coded the Sentry DSN for dev and added an Exception to see if it would log correctly.  This event showed up when I viewed the page http://sentry.vfs.va.gov/vets-gov/platform-api-development/issues/218736/.

ericbuckley:
@bkjohnson would you be able to review this?  It's a change to the vets-api, however this template renders HTML and not JSON, and we need Sentry tracking to make sure our users aren't experiencing any JS errors.

bkjohnson:
Adding JS to the site from the `vets-api` doesn't seem great for a number of reasons (circumvents our linting, testing, etc), but I think this is more of an identity team concern than a frontend one.

ericbuckley:
> Adding JS to the site from the `vets-api` doesn't seem great for a number of reasons (circumvents our linting, testing, etc), but I think this is more of an identity team concern than a frontend one.

I agree it's not ideal, but using [JS to auto POST this HTML form seems to be the preferred way](https://en.wikipedia.org/wiki/SAML_2.0#HTTP_POST_Binding).  Long term it might be best to have vets-api redirect the user back to the FE and have the FE render and submit a similar HTML form, but I know there would be some extra overhead with that call to load the react environment.

For this PR, I wanted to just double-check that the Sentry tracking code looks ok.

PR #4997, opened 2020-09-29
Adding a timeout rescue with a status refresh check

bastosmichael:
## Description of change
Checking status refreshing if Faraday::TimeoutError persists...

## Original issue(s)
https://vajira.max.gov/browse/API-2488

### DIFF
--- a/modules/vba_documents/app/workers/vba_documents/upload_processor.rb
+++ b/modules/vba_documents/app/workers/vba_documents/upload_processor.rb
@@ -36,6 +36,8 @@ module VBADocuments
         response = submit(metadata, parts)
         process_response(response)
         log_submission(@upload, metadata)
+      rescue Faraday::TimeoutError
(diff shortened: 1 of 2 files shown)


modules/vba_documents/app/workers/vba_documents/upload_processor.rb
```diff
@@ -36,6 +36,8 @@ def download_and_process
         response = submit(metadata, parts)
         process_response(response)
         log_submission(@upload, metadata)
+      rescue Faraday::TimeoutError
+        VBADocuments::UploadSubmission.refresh_statuses!([@upload])
```
templeman15, line 40: Per our conversation on zoom, if you want to look into this actually being called that may be something to look into. Otherwise the PR looks good to me.

templeman15 (approved):
Look good to me!

erluti (approved):

PR #5002, opened 2020-09-30
added parent sites to direct booking call

grossb1:
## Description of change

Updated directing_booking_eligibility controller to accept `parent_sites` and pass them into the down-stream service call

From Kam's note below:
This was needed to address an issue with VCR wanting to default encoding of the request body to ASCII-8BIT for some bizarre reason, but then having a hard time because the payload includes UTF-8 characters, and then ending up with a base64encoded body that is not human-readable. I proposed we add this to rails_helper using VCR hooks because I suspect others might run into this aggravation as well:

```
VCR.configure do |c|
  c.before_record(:force_utf8) do |interaction|
    interaction.response.body.force_encoding('UTF-8')
  end
end
```

## Original issue(s)
https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/14063

Unit tests were added to reflect this update
Added a new vcr cassette to reflect this update


saneshark (approved):
LGTM

saneshark:
@omgitsbillryan I think this requires VSP review because of the addition we made to rails_helper.rb

grossb1:
The tags feature was Kam's idea!

PR #5005, opened 2020-09-30
[API 687] Update the documentation for original appeals status endpoint

templeman15:
## Description of change
This PR updates the documentation to display an array instead of an object for the data field in the appeals status API. 

## Original issue(s)
[Update the documentation for original appeals status endpoint](https://vajira.max.gov/browse/API-687)


bastosmichael (approved):
As long as you checked it against the developer portal I'm good with it.

PR #5006, opened 2020-09-30
add_parser has too much code copy-pasted

annaswims:
## Description of change
The MVI find profile response parser and the mvi add profile response parser shared a significant amount of code overlap.  

It looks like code was copy-pasted from `profile_parser` to `add-parser` [here](https://github.com/department-of-veterans-affairs/vets-api/pull/3632/files#diff-622e8b48f34f2d83fb6089237648d2ffR1
).

This PR cleans that up as a precursor to department-of-veterans-affairs/va.gov-team#11353 

Surely there's more to be done in reducing redundancy and increasing consistency, but this is a first step.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#11353

### DIFF
--- a/lib/mvi/responses/add_parser.rb
+++ b/lib/mvi/responses/add_parser.rb
@@ -1,23 +1,17 @@
 # frozen_string_literal: true
 
 require 'sentry_logging'
+require_relative 'parser_base'
 
 module MVI
   module Responses
     # Parses a MVI response and returns a MviProfile
-    class AddParser
+    class AddParser < ParserBase
       include SentryLogging
 
       ACKNOWLEDGEMENT_DETAIL_CODE_XPATH = 'acknowledgement/acknowledgementDetail/code'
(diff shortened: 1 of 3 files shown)


lib/mvi/responses/add_parser.rb
```diff
@@ -76,32 +49,16 @@ def parse_ids(attributes)
         attributes.each do |attribute|
           case attribute[:code]
           when /BRLS/
-            codes[:birls_id] = sanitize_ids(attribute[:code])
```
annaswims, line 79: this was one of the few places that have a tiny change to behavior.

lib/mvi/responses/parser_base.rb
```diff
@@ -0,0 +1,72 @@
+# frozen_string_literal: true
+
+class ParserBase
```
edmangimelli, line 3: is this worth putting under `module MVI` ?

edmangimelli (approved):
i like it!

PR #5007, opened 2020-09-30
13457 mobile military info endpoint

jperk51:
## Description of change
Set up mobile endpoint for military service history list. Created controller and serializers in the mobile module that will handle requests for military service history information leveraging the existing MilitaryInformation class

## Original issue(s)
department-of-veterans-affairs/va.gov-team#13457
Related docs: https://github.com/department-of-veterans-affairs/vets-api/blob/91ea93748cda3b6d7951770fbff92c8d01ed5e87/modules/mobile/docs/openapi.yaml#L24-L79

Created rspec in mobile module that tests this endpoint's responses.


jperk51:
We are working on created out own schema matcher in the mobile module. Once that is set up I can move the schema into our module so we aren't touching things outside of it.

modules/mobile/spec/request/military_information_request_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+
+require 'rails_helper'
+require_relative '../support/iam_session_helper'
+
+RSpec.describe 'military_information', type: :request do
+  include SchemaMatchers
+  describe 'GET /mobile/v0/military-service-history' do
+    context 'with a user who has a cached iam session' do
+      before { iam_sign_in }
+      let(:inflection_header) { {'X-Key-Inflection' => 'camel'} }
```
kreek, line 11: Figured this out when I updated my user request spec turns out just setting this isn't enough, you need to pass it into the headers as well 😄 

I added a helper method to the `IAMSessionHelper` class:
```ruby
  def iam_headers
    {
      'Authorization' => "Bearer #{access_token}",
      'X-Key-Inflection' => 'camel'
    }
  end
```

Which you then pass into the headers in the request spec call:
``` ruby
get '/mobile/v0/user', headers: iam_headers
```

kreek: I think once that's in we won't need `set_key_transform :camel_lower` in the serializer.

kreek (commented):
Few comments mainly from me steering you wrong on the need for a struct 😬 specs look great 😄  Once the tweaks go in I think this is ready to be a regular PR.

modules/mobile/app/serializers/mobile/v0/military_information_serializer.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+require 'fast_jsonapi'
+
+module Mobile
+  module V0
+    class MilitaryInformationSerializer
+      include FastJsonapi::ObjectSerializer
+      set_key_transform :camel_lower
```
kreek, line 9: I think you should be able to remove this now that the `iam_headers` include the olive branch key inflection

kreek (approved):
One minor comment but LGTM 🚀

jjulian (approved):

modules/mobile/app/serializers/mobile/v0/military_information_serializer.rb
LindseySaari: Was this comment addressed and ready for review?

modules/mobile/app/controllers/mobile/v0/military_information_controller.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require_dependency 'mobile/application_controller'
+
+module Mobile
+  module V0
+    class MilitaryInformationController < ApplicationController
```
LindseySaari, line 7: Is there an API doc PR that goes along with this new endpoint? If so, could you link in the PR body please?

modules/mobile/config/routes.rb
```diff
@@ -6,4 +6,8 @@
   namespace :v0 do
     get '/user', to: 'users#show'
   end
+
+  namespace :v0 do
+    get '/military-service-history', to: 'military_information#get_service_history'
```
LindseySaari, line 11: Can this route be moved underneath line 7?

modules/mobile/app/controllers/mobile/v0/military_information_controller.rb
jperk51: Added to the PR

modules/mobile/config/routes.rb
jperk51: Done

modules/mobile/app/serializers/mobile/v0/military_information_serializer.rb
jperk51: Yup, it was removed as requested.

LindseySaari (approved):
💯 Thanks for the changes!

PR #5010, opened 2020-09-30
Add code owner entry for the mobile API team

kreek:
## Description of change
Creates a code owner entry so that the [mobile API team](https://github.com/orgs/department-of-veterans-affairs/teams/mobile-api-team) owns code in the `modules/mobile` dir and associated spec support files. I copied the VAOS model where both the VAOS team and `backend-review-group` are code owners. Is this correct? We want the mobile team to own it and be able to approve for merge. Then also have the backend review group pinged for visibility, but not require their ✅ to merge.


jperk51 (approved):
Looks good to me

.github/CODEOWNERS
```diff
@@ -11,3 +11,9 @@ spec/support/*/vaos/                 @department-of-veterans-affairs/vfs-vaos @d
 
 lib/bgs @department-of-veterans-affairs/vfs-ebenefits @department-of-veterans-affairs/backend-review-group
 spec/bgs @department-of-veterans-affairs/vfs-ebenefits @department-of-veterans-affairs/backend-review-group
+
+modules/mobile/                      @department-of-veterans-affairs/mobile-api-team @department-of-veterans-affairs/backend-review-group
+lib/iam_ssoe_oauth                   @department-of-veterans-affairs/mobile-api-team @department-of-veterans-affairs/backend-review-group
```
erluti, line 16: anything outside of `modules/mobile` is shared code that VSP has to support, so making mobile-api-team code owners outside that directory isn't going to work.  Moving code that doesn't need to be shared with the platform into `modules/mobile` would be a good approach to own some of that stuff.

erluti (changes requested):

.github/CODEOWNERS
erluti: Did you want to own `spec/support/*/mobile/`?  Originally you were going to copy vaos and thought that was a good approach.  But I'll approve it without that if you want

kreek: @erluti we were going to try to have our spec support files also in the engine (still researching how to do this). But ya for now I'll add the mobile support dir back in.

erluti (approved):

PR #5016, opened 2020-10-01
updated logging to use the request instead of the response

grossb1:
## Description of change
This issue is looking to enhance our ability to coordinate with VAMF when tracking down issues a user may encounter when interacting with VAOS. Prior to this change the code was logging the JTI present on the response object.  This was problematic if a response wasn't present due to a down-stream error.  In this case, we would lose the ability to identify a user to VAMF because we wouldn't have a JTI.  The solution was to record the JTI present in the request body/header (as opposed to the response) so in the case of a down-stream error, we would be able to access the JTI associated with the request which could then be presented to VAMF to facilitate identifying the error.

The `token` variable was also memoized in the jwt_wrapper as an optimization.

Also, we noticed that we’re using two different variations of the header name `X-Vamf-Jwt` and `X-VAMF-JWT`.  We will fix this as part of a future PR and tag the ticket number that will address that in the comment.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/6837

Unit tests were updated to reflect this change

### DIFF
--- a/modules/vaos/app/services/vaos/jwt_wrapper.rb
+++ b/modules/vaos/app/services/vaos/jwt_wrapper.rb
@@ -14,7 +14,7 @@ module VAOS
     end
 
     def token
-      JWT.encode(payload, Configuration.instance.rsa_key, 'RS512')
+      @token ||= JWT.encode(payload, Configuration.instance.rsa_key, 'RS512')
     end
 
     private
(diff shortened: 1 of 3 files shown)


erluti (approved):
LGTM

saneshark:
LGTM too

saneshark (approved):

PR #5017, opened 2020-10-02
Add More Data to HLR Prefill

edmangimelli:
_As a user I want to verify my contact info / see the last four digits of my social security number and VA file number._

The FE needs to display extra data that won't be submitted with the form. This PR puts that extra data into a new top level property of the prefill.

AC:
- [x] tests pass

Notes:
It might seem like I should have added `street3` to the _prefill -> initialize_contact_information_ flow, adding `street3` as an attribute to `FormAddress` (that would have been a quicker solution).

I opted to not do that because some of the [mapping yamls](https://github.com/department-of-veterans-affairs/vets-api/tree/f19821602df400f0f79ccd780f5096e41f40c53a/config/form_profile_mappings) pull in the entire address, and adding `street3` to `FormAddress` would suddenly cause those forms to have a new field in their prefill (which might not pass their schema, or cause subtle problems)

There is some more refactoring here in the spirit of the [last refactor PR](https://github.com/department-of-veterans-affairs/vets-api/pull/4988). It's nudging the FormProfile class away from the

_overriding prefill **to kick off a unique set of `initialize_*` methods**_ pattern, and the

_changing/overriding a shared data structure (virtus classes)_ pattern

to having a method for each _node_ of data in the tree. This will allow more prefill updates to be just updates to the mapping yaml, instead of writing code in a subclass that is highly dependent on the internal structure of the parent class.

Example: you need more address info. Instead of overriding `prefill` or `initialize_contact_information` (or adding data to a shared structure like `FormAddress`, which would inadvertently change other forms' prefills), you can just change your yaml to `county: [vet360_contact_info, mailing_address, county]` or something similar.


app/models/form_profile.rb
```diff
@@ -232,40 +233,52 @@ def convert_vets360_address(address)
...
 
     FormContactInformation.new(opt)
   end
 
+  # doing this (below) instead of `@vet360_contact_info ||= Settings...` to cache nil too
+  def vet360_contact_info
+    return @vet360_contact_info if @vet360_contact_info_retrieved
+
+    @vet360_contact_info_retrieved = true
+    @vet360_contact_info = Settings.vet360.prefill &&
```
erluti, line 272: I think this is confusing because it reads like a boolean statement but it's actually setting a value when other values are present.  I think it would be easier to read if it were something like this:
```
  def vet360_contact_info
    return @vet360_contact_info if @vet360_contact_info_retrieved

    @vet360_contact_info_retrieved = true
    if Settings.vet360.prefill && user.vet360_id.present?
      @vet360_contact_info = Vet360Redis::ContactInformation.for_user(user)
    end
  end
```

erluti (commented):
looks good, one style suggestion

app/models/form_profile.rb
edmangimelli: i like it! much easier to grok 💯

edmangimelli: @erluti do you think it'd be good to add this final line, just to make it more clear that `@vet360_contact_info` will be returned? 
```
  def vet360_contact_info
    return @vet360_contact_info if @vet360_contact_info_retrieved

    @vet360_contact_info_retrieved = true
    if Settings.vet360.prefill && user.vet360_id.present?
      @vet360_contact_info = Vet360Redis::ContactInformation.for_user(user)
    end
    @vet360_contact_info # <-- HERE
  end
```

erluti: i'm always a fan of clarity!

erluti (approved):
looks good!  Master has a failing spec so when that's resolved you should be able to merge and have a passing pipeline

app/models/form_profile.rb
```diff
@@ -220,7 +220,8 @@ def initialize_identity_information
     )
   end
 
-  def convert_vets360_address(address)
+  def vet360_mailing_address_hash
```
annaswims, line 223: what about `def vet360_mailing_address_hash(include_street3 = false)` so that it's more clear to others who come along later that address line3 is even a thing?

edmangimelli: that hash is used as arguments to create a FormAddress object (via a FormContactInfo object).
so it could be added to `vet360_mailing_address_hash`, but would be ignored unless i add street3 as an attribute to a FormAddress.
Changing FormAddress would change the prefills for a handful of prefills which send the entire FormAddress object to the FE (suddenly all of those prefills would have an extra field). It's hard to say whether or not that would be a problem (could cause schemas to fail).
I decided to err on the side of caution and not change FormAddress.

PR #5020, opened 2020-10-02
Add DEPCHG to update calls

digitaldrk:
## Description of change
During testing with BGS, we found we need to send the `vnPProcTypeCd` in update calls in order for that attribute to persist in their system and create a valid claim.

We also would like to capture the procId associated with the claim submission to help them debug their systems.

## Original issue(s)

  * No additional changes
  * This feature is flagged, it is not available to the public
  * We added a message log to get the procId which is a unique "procedure ID" for the submission
  * No additional changes
  * No additional changes
  * No concerns

-->

### DIFF
--- a/lib/bgs/children.rb
+++ b/lib/bgs/children.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative 'service'
+
 module BGS
   class Children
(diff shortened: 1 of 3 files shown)


kathleencrawford (approved):
LGTM 👍

PR #5022, opened 2020-10-02
Update carrierwave to lastest

erluti:
Updates carrierwave gem to the latest.

`spec/uploaders/evss_claim_document_uploader_spec.rb` had a spec that was failing.  After investigating it was never working as intended, so it was removed.  Instead a spec for `ValidateFileSize` was added since it had no tests.


annaswims:
`rspec ./spec/uploaders/evss_claim_document_uploader_spec.rb:126 # EVSSClaimDocumentUploader#store! raises an error when the file is larger than 25 megabytes`

was a poorly written test. it was expecting an error, due to file size but instead the error raised was 
`CarrierWave::IntegrityError: You are not allowed to upload nil files, allowed types: pdf, gif, tiff, tif, jpeg, jpg, bmp, txt` which is a kind of `CarrierWave::UploadError`

annaswims:
There are breaking changes noted in the changelog.  Like - `extension_white_list ~> extension_whitelist`

https://github.com/carrierwaveuploader/carrierwave/blob/3356634d80d372c458030460978c3b63c006e042/CHANGELOG.md

annaswims:
I'm trying to standardize the error messages of file uploads in #5022 and added some tests that cover file upload size [here](https://github.com/department-of-veterans-affairs/vets-api/pull/5059/files#diff-a03e7121868faa25fc4afae420bff0eeR348) and

erluti:
> There are breaking changes noted in the changelog. Like - `extension_white_list ~> extension_whitelist`
> 
> https://github.com/carrierwaveuploader/carrierwave/blob/3356634d80d372c458030460978c3b63c006e042/CHANGELOG.md

didn't realize that one applied!  thanks!

erluti:
> I'm trying to standardize the error messages of file uploads in #5022 and added some tests that cover file upload size [here](https://github.com/department-of-veterans-affairs/vets-api/pull/5059/files#diff-a03e7121868faa25fc4afae420bff0eeR348) and

are you suggesting I try to copy some here or do you want to finish that PR before the upgrade is merged?

annaswims:
I was thinking I'd merge# 5059 first, since it adds a bit of coverage. It's good to go as soon as it gets an approval, I think.

annaswims:
(sorry, deleted my previous incorrect comment right after I wrote it.) 

I just wanted you to be aware.  I'm going to hold off merging mine until I've tested with the FE.

spec/uploaders/validate_file_size_spec.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe ValidateFileSize do
+  class ValidateFileSizeTest < CarrierWave::Uploader::Base
+    include ValidateFileSize
+
+    MAX_FILE_SIZE = 100
```
LindseySaari, line 9: To make this more uniform with Annas PR, can you make the MAX_FILE_SIZE a class method instead of a constant. See [here](https://github.com/department-of-veterans-affairs/vets-api/pull/5059/files#diff-5e5224c9764baaff6de004f124c5434cL10)

erluti: I feel like that's out of scope for this change.  It might even make sense as it's own PR

LindseySaari (approved):

spec/uploaders/validate_file_size_spec.rb
annaswims: So.... I think the solution is to get rid of the ValidateFileSize module altogether.  Carrierwave has [a size range method](https://github.com/carrierwaveuploader/carrierwave/blob/f15f4ba71dd2c78e8df494bf65938bbaf13960f9/lib/carrierwave/uploader/file_size.rb#L21)

LindseySaari: Oh, good thought Anna. We can probably break this work out into a separate PR

annaswims: Good call! FWIW, I ran the specs for our two branches together and this was the only test to fail.  I can fix the conflict after you merge.

erluti: I'll go forward with the merge here, then.

PR #5023, opened 2020-10-02
Fix: 10-10CG PDF not found when attempting to delete it after submission

kevinmirc:
Fixes: https://github.com/department-of-veterans-affairs/va.gov-team/issues/14273

**Description**:
When a user submits a 10-10CG, we generate a PDF of their submission and send it to CARMA as an attachment to the Case. There were, as of this time, two production submissions where the data and the attachment made it successfully to CARMA, but the user received a 500 error because we called `File.delete` on a file or directory that apparently didn't exist after it was sent to CARMA.

We are still not sure why the file cannot be found when it was generated just a few lines of code above where we try to delete it, but this fix will prevent any additional users from being told their submission failed when it actually didn't.

**Sentry**:
http://sentry.vfs.va.gov/vets-gov/platform-api-production/issues/221016/events/b8710ff80f0a4553bf655093d2d3a471/

### DIFF
--- a/app/services/form1010cg/service.rb
+++ b/app/services/form1010cg/service.rb
@@ -64,7 +64,7 @@ module Form1010cg
     # Will generate a PDF version of the submission and attach it to the CARMA Case.
     #
     # @return [Boolean]
-    def submit_attachment # rubocop:disable Metrics/MethodLength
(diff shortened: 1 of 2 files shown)


LindseySaari (approved):
👍

PR #5026, opened 2020-10-02
Moving into getting V0 working for POA

bastosmichael:
## Description of change
Turning Power of Attorney End Points for Claims V0 API

## Original issue(s)
https://vajira.max.gov/browse/API-1631


LindseySaari (approved):

modules/claims_api/app/controllers/claims_api/v0/forms/power_of_attorney_controller.rb
```diff
@@ -36,8 +36,6 @@ def submit_form_2122
           end
 
           ClaimsApi::PoaUpdater.perform_async(power_of_attorney.id)
-          data = power_of_attorney.form_data
-          ClaimsApi::PoaFormBuilderJob.perform_async(power_of_attorney.id) if data['signatureImages'].present?
```
hinzed1127, line 40: what's different about v0 from v1 that makes these removable here?

hinzed1127:
Are there any changes needed to address this line of things to do original issue? 

> Adding MVI lookup using the headers we accept in order to fetch participant ID

modules/claims_api/app/controllers/claims_api/v0/forms/power_of_attorney_controller.rb
bastosmichael: I checked and most of the work was already done in the corresponding parent controller due to ITF so I was pleasantly surprised how little was necessary to implement this properly after tests.

bastosmichael:
@hinzed1127 To answer your question, I checked and most of the work was already done in the corresponding parent controller due to ITF so I was pleasantly surprised how little was necessary to implement this properly after tests.

PR #5027, opened 2020-10-05
fix time-dependent spec

annaswims:
## Description of change
freeze time on BDD spec. BDD has a 3-month eligibility window.


erluti (approved):
thanks for the quick fix!

edmangimelli (approved):
lgtm

PR #5029, opened 2020-10-05
only initialize Setnry if its defined

ericbuckley:
some ad blockers are preventing the bundle from being loaded via the
sentry dsn, in that case we need to check that the Sentry variable is
defined or else the call to `Sentry.init` will result in an error and
prevent the user from submitting.

### testing
I manually tested this locally running AdblockPlus in Chrome.  With all the settings enabled, `/v1/sessions/idme/new` raised an error and the submission halted, however after introducing this change the submission went forward.

### DIFF
--- a/lib/saml/templates/sso_post_form.html.erb
+++ b/lib/saml/templates/sso_post_form.html.erb
@@ -29,7 +29,7 @@
 
   <script>
     (function() {
-      if ("<%= sentrydsn %>") {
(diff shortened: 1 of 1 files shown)


LindseySaari (approved):
👍

jholton (approved):

AGarcia-Clarity (approved):

PR #5031, opened 2020-10-05
upgrade shrine and aws-sdk-s3

erluti:
`shrine` had a vulnerability reported here https://github.com/department-of-veterans-affairs/vets-api/network/alert/Gemfile.lock/shrine/open

The newest `shrine` needs an updated `aws-sdk-s3`, so that is updated as well.

## to know about this PR
The `:logging` plugin is not supported in the latest shrine in favor of the `:instrumentation` plugin.  A separate issue will be made for that change-over


LindseySaari (approved):
Looks good as long as there's backwards compatibility support for logging, which looks like is covered by the instrumentation plugin 👍

Gemfile
```diff
@@ -138,7 +138,6 @@ group :test do
   gem 'rspec-retry'
   gem 'rspec_junit_formatter'
   gem 'rubocop-junit-formatter'
-  gem 'shrine-memory'
```
erluti, line 141: this is actually part of shrine 3

erluti:
New ticket for switching over to instrumentation https://app.zenhub.com/workspaces/vsp-5cedc9cce6e3335dc5a49fc4/issues/department-of-veterans-affairs/va.gov-team/14414

PR #5032, opened 2020-10-05
Task/api 2170 vv utilize mock bgs

esubocz-va:
## Description of change
Veteran Verification has been edited to allow use of mock-bgs independent of other apps utilizing the BGS service. 

## Original issue(s)
https://vajira.max.gov/browse/API-2170

`mock_bgs` is a new configuration that can be set in the bgs section of your `settings.local.yml file`. it will default to `nil` if not set, and use the tunnel. so, no changes are required to any other apps. this will affect veteran verification only.

in order to interact with the mock-bgs service, you will need to have the app running locally. repo found here: https://github.com/department-of-veterans-affairs/health-apis-mock-benefits-gateway-services and the following changes will need to be made to your `settings.local.yml` file:
```
vet_verification:
  url: http://localhost:9090/v0
  mock_bgs: true
```

local testing has been done to ensure functionality remains the same when using the mock-bgs service. as well as that this causes no breaking changes for standard use of the tunnel.


dbrown-va (approved):
Confirmed the additional spec addition to specify no mocking that caused the earlier master branch issue.

```
  before do
    Settings.vet_verification.mock_bgs = false
  end
```

modules/veteran_verification/app/services/bgs/mock_disability_rating_service.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+require 'savon'
+require 'nokogiri'
+
+module BGS
```
LindseySaari, line 6: I'm not seeing a spec for this service class? Is there a reason for not adding one?

esubocz-va: https://github.com/department-of-veterans-affairs/vets-api/blob/ad4fe80d65cf3eb50d00d9a5e4addc2c3e3f847f/modules/veteran_verification/spec/requests/mock_disability_rating_request_spec.rb

it's located here, next to the non-mocked service test.

spec/support/vcr_cassettes/bgs/rating_web_service/mock_rating_data.yml
```diff
@@ -0,0 +1,191 @@
+---
+http_interactions:
+- request:
+    method: get
+    uri: https://blue.qa.lighthouse.va.gov/mock-bgs/v0/RatingServiceBean/RatingService/ratingRecord.wsdl
```
beaugrantham-va, line 5: Does it matter that this uri is missing `/mock-bgs`?

PR #5033, opened 2020-10-05
Update Payment Summary to use a different BGS API

kathleencrawford:
## Description of change

This PR is for the View Payment History feature.  We were initially directed to use a BGS service to get payment history, but during implementation discovered that we should be using a different BGS API.  This PR updates the existing payment history functionality using the new API.  For reference, the name of the new BGS API being called is `retrievePaymentSummaryWithBDN`.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#14196

Swagger docs and spec tests have been updated.

## Testing

- [x] Local Testing
- [x] Spec coverage


app/controllers/v0/profile/payment_history_controller.rb
```diff
@@ -6,8 +6,9 @@ class PaymentHistoryController < ApplicationController
       before_action { authorize :bgs, :access? }
 
       def index
+        person = BGS::PeopleService.new(current_user).find_person_by_participant_id
         service = BGS::PaymentService.new(current_user)
```
digitaldrk, line 10: Since `person = BGS::PeopleService.new(current_user).find_person_by_participant_id` is a one-liner, maybe we should do that for this too:
```
service = BGS::PaymentService.new(current_user)
response = service.payment_history(person)
```

So it would be `response = BGS::PaymentService.new(current_user).payment_history(person)`

app/serializers/vet_payment_history_serializer.rb
```diff
@@ -10,11 +10,63 @@ def id
...
+    all_payments.each do |payment|
+      if payment.dig(:return_payment, :check_trace_number).present?
+        @returned_payments << {
+          'returned_check_issue_dt': payment[:payment_date],
+          'returned_check_cancel_dt': payment[:return_payment][:return_date],
+          'returned_check_amount': ActiveSupport::NumberHelper.number_to_currency(payment[:payment_amount]),
+          'returned_check_number': payment[:return_payment][:check_trace_number],
+          'returned_check_type': payment[:payment_type],
+          'return_reason': payment[:return_payment][:return_reason]
+        }
```
digitaldrk, line 41: I forget if we tried this or not but in case not, I wonder if we can pull this hash and the one below into their own methods 🤔 . 

If we do, I bet we can drop the cop disable.

app/serializers/vet_payment_history_serializer.rb
```diff
@@ -10,11 +10,63 @@ def id
     nil
   end
 
+  def initialize(object, options = nil)
+    @formatted_payments = []
+    @returned_payments = []
+    process_payments(object[:payments][:payment]) if object[:payments].present?
```
digitaldrk, line 16: This is confusing but not bc anything you did or didn't do. We may have already worked through this scenario.

IIRC object _could_ be this right? `{ payments: [], return_payments: [] }`

If this gets returned `{ payments: [], return_payments: [] }`, will the conditional work as intended?

kathleencrawford: Yes.  If `:payments = []`, `object[:payments].present?` will return `false`.

app/serializers/vet_payment_history_serializer.rb
kathleencrawford: Thanks!  I've moved the return payments into a separate method.

app/controllers/v0/profile/payment_history_controller.rb
kathleencrawford: Great idea!  Updated as suggested.

digitaldrk (approved):
😎

Gemfile.lock
```diff
@@ -993,4 +993,4 @@ RUBY VERSION
    ruby 2.6.6p146
 
 BUNDLED WITH
-   1.17.3
+   1.17.3
```
LindseySaari, line 996: Could you add the newline back here, or is this something that rubocop auto corrected?

app/serializers/vet_payment_history_serializer.rb
```diff
@@ -10,11 +10,70 @@ def id
...
+      pay_check_type: payment[:payment_type],
+      payment_method: get_payment_method(payment),
+      bank_name: payment.dig(:address_eft, :bank_name),
+      account_number: mask_account_number(payment[:address_eft])
+    }
+  end
+
+  def process_return_payment(payment)
+    @returned_payments << {
+      'returned_check_issue_dt': payment[:payment_date],
```
LindseySaari, line 54: Will payment[:return_payment] always exist?

app/serializers/vet_payment_history_serializer.rb
```diff
@@ -10,11 +10,70 @@ def id
...
+      'returned_check_cancel_dt': payment[:return_payment][:return_date],
+      'returned_check_amount': ActiveSupport::NumberHelper.number_to_currency(payment[:payment_amount]),
+      'returned_check_number': payment[:return_payment][:check_trace_number],
+      'returned_check_type': payment[:payment_type],
+      'return_reason': payment[:return_payment][:return_reason]
```
LindseySaari, line 59: Do you have any defaults for example if an attribute is nil or doesn't exist e.g. payment[:return_payment][:return_reason] where return_reason is nil?

Gemfile.lock
kathleencrawford: I don't know what happened here...github was being weird.  But it is fixed now.

app/serializers/vet_payment_history_serializer.rb
kathleencrawford: I'm not sure if I'm understanding your question correctly, so I'll do my best to answer/explain, but please clarify if I'm not answering your question.

So before we enter this method, it is using the "dig" method to determine whether or not [:return_payment][:check_trace_number] exists.  Using the "dig" method here prevents any errors in case [:return_payment] is nil (which it should not be, but just in case).  

Once  within this method, any of the return_payment attributes can be empty or nil.  
'returned_check_issue_dt'
'returned_check_cancel_dt'
'returned_check_amount'
'returned_check_number'
'returned_check_type'
'return_reason'

The FE is checking for nil values as well and putting something like "---" in the case of nil attributes.  

Does this help address your concern?

LindseySaari: Ah, thank you. I overlooked the dig method 👍

LindseySaari (approved):

app/serializers/vet_payment_history_serializer.rb
kathleencrawford: Thank you @LindseySaari!

PR #5043, opened 2020-10-06
Added email and phone fields to prefill

cohnjesse:
## Description of change
Since we now have prefill working on the CH36 form we wanted to add the email and phone number fields into the prefill so that the Veteran does not have to fill those out manually.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/12669


spec/models/form_profile_spec.rb
```diff
@@ -698,10 +698,12 @@
         'addressLine1' => street_check[:street],
         'addressLine2' => street_check[:street2],
         'city' => user.va_profile[:address][:city],
-        'stateCode' => user.va_profile[:address][:state],
-        'countryName' => 'USA',
-        'zipCode' => user.va_profile[:address][:postal_code][0..4]
-      }
+        'state' => user.va_profile[:address][:state],
```
LindseySaari, line 701: I see that you adjusted the stateCode, CountryName and zipCode mappings in [this PR](https://github.com/department-of-veterans-affairs/vets-api/pull/5041/files#diff-2289ed0baec2d282157fc61580876bb7R701-R703) and lines 701-703 revert those changes.

cohnjesse: @LindseySaari So I changed that because I am getting unit test failures with it the way it was on the build for this PR. I am not really sure what is causing the unit tests to fail.

spec/models/form_profile_spec.rb
```diff
@@ -699,9 +699,11 @@
         'addressLine2' => street_check[:street2],
         'city' => user.va_profile[:address][:city],
         'stateCode' => user.va_profile[:address][:state],
-        'countryName' => 'USA',
+        'countryName' => user.va_profile[:address][:country],
         'zipCode' => user.va_profile[:address][:postal_code][0..4]
-      }
+      },
+      'claimantPhoneNumber' => user.va_profile[:address][:state],
```
LindseySaari, line 705: It looks like you have the phone number and email address flip flopped.

LindseySaari: 'claimantPhoneNumber' => user.va_profile[:address][:state] should be 'claimantPhoneNumber' => us_phone and vise versa

LindseySaari: Oh, and phone number is set to state 'claimantPhoneNumber' => user.va_profile[:address][:state]

spec/models/form_profile_spec.rb
LindseySaari: It looks like you have a few typos in the spec. I commented inline

spec/models/form_profile_spec.rb
cohnjesse: Geez I am an idiot

LindseySaari (approved):
👍

PR #5044, opened 2020-10-07
Updates to `rake form526:errors`

edmangimelli:
1. sort results of rake form526:errors by number of errors descending
2. take in extra arg that can force json output


annaswims (approved):
Yes! Thank you!

LindseySaari (approved):

PR #5045, opened 2020-10-07
13350 letters api docs

jperk51:
## Description of change
Open API docs for Mobile Letters endpoints

## Original issue(s)
department-of-veterans-affairs/va.gov-team#13350

Added letter endpoint design to mobile API docs. No tests since this is a docs only change.


modules/mobile/docs/openapi.yaml
```diff
@@ -21,8 +21,194 @@ paths:
...
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Errors"
+        '504':
+          description: A response from an upstream service timed out.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Errors"
+  /v0/letters/beneficiary:
```
kreek, line 86: It's frustrating that letters isn't working on staging and is blocking us from testing if we can combine the letters list and beneficiary calls. There's a couple options; make them sequentially or in parallel. There's pros and cons to both, we should be waiting for the list before we know that a user has a beneficiary letter. But, you don't _need_ to know that so it could be faster to make the call in parallel and live with beneficary 404s.

kreek: looking at Grafana over the last couple days the letters call is taking 2.5 - 5ms and beneficiary is 2-2.5ms

LindseySaari (approved):

modules/mobile/docs/openapi.yaml
jperk51: Would it not be best to keep them as separate endpoints for both of those options? That way the app can just deal with how to fire the async calls. 

Aside: My opinion would be that we only fire the beneficiary request when the user selects that letter from the letter list, but that's with the assumption that call doesn't take several minutes to resolve.

kreek: That's definitely the simplest and it lets the app control how to retrieve things. There's not a lot of places we can make the API more efficient for mobile. One area is reducing the number of http requests.

The main issue is that api.va.gov because of its architecture will never be as fast as a top-tier private api hosted on AWS. Each call has an overhead. Once the request makes it into the VPC and gets to rails app making calls to backend services (especially if they're also host in AWS GovCloud) can happen fairly quickly vs doing the entire round trip again. Note the request latency listed above is from api.va.gov -> evss.

All that said separate requests might be 100% be the right choice 😃 just want to explore all the options.

kreek: This is more something to think about when we get to appointments, because va.gov currently makes like five requests just to build to initial UI.

kreek (approved):

modules/mobile/docs/openapi.yaml
jjulian: One of the benefits of this mobile api is to simplify and consolidate the existing endpoints where we can...I'd prefer to keep our plan of combining these two calls. If I'm reading this right, these calls take **_minutes_**, so this is a good candidate for (1) having the mobile app request this early, and (2) having the mobile app (or this api tier?) cache these results for a certain time period.

PR #5048, opened 2020-10-07
Api 204 remove stray file

nathanbwright:
This PR removes a stray file that I accidentally added with https://github.com/department-of-veterans-affairs/vets-api/pull/4747. The `2018_Gaz_zcta_national.txt` file should only live in `lib/facilities` (which it currently is at) and not be duplicated at rails root.


jholton (approved):
Thanks!

LindseySaari (approved):

PR #5052, opened 2020-10-08
Service history pact tests

lihanli:
## Description of change
add pact tests for profile service history endpoint

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/14234

tests are working locally


erikphansen:
@lihanli just a heads up that we merged the fix to the FE side of the Pact tests so we are calling the correct endpoint now.

spec/service_consumers/pact_helper.rb
```diff
@@ -42,9 +42,15 @@
     stub_mvi unless example.metadata[:skip_mvi]
     stub_emis unless example.metadata[:skip_emis]
     stub_vet360 unless example.metadata[:skip_vet360]
+
+    Redis.current.flushall
```
LindseySaari, line 46: Were you run into issues with existing Redis keys when running the pact tests?

lihanli: yes the `there are no service history records` spec won't pass unless redis is cleared because the `at least one entry in the service history exists` redis data will still exist

LindseySaari: I ran into a similar issue with db records and transactional fixtures 👍

LindseySaari (approved):
💯

PR #5057, opened 2020-10-08
convert Breakers `duration` value to ms before sending via StatsD

jholton:
## Description of change
convert Breakers `duration` value to ms before sending via StatsD

## Original issue(s)


jholton:
Updated Breakers gem to report in ms instead

PR #5058, opened 2020-10-08
[API 2880] NOD contestable issues endpoint (vets-api)

templeman15:
## Description of change
This PR:
 - Refactors HigherLevelReviews/ContestableIssuesController to use a common base class `BaseContestableIssuesController` 
- Implements `NoticeOfDisagreements/ContestableIssuesController` which inherits from base class
- Updates specs using shared examples
- Updates documentation for `NoticeOfDisagreements/ContestableIssuesController`

## Original issue(s)
[NOD contestable issues endpoint (vets-api)](https://vajira.max.gov/browse/API-2880)

### DIFF
--- a/modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/base_contestable_issues_controller.rb
+++ b/modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/base_contestable_issues_controller.rb
@@ -1,0 +1,83 @@
+# frozen_string_literal: true
+
+require 'caseflow/service'
+require 'common/exceptions'
+
+class AppealsApi::V1::DecisionReviews::BaseContestableIssuesController < AppealsApi::ApplicationController
+  skip_before_action(:authenticate)
+
+  EXPECTED_HEADERS = %w[X-VA-SSN X-VA-Receipt-Date].freeze
+
+  UNUSABLE_RESPONSE_ERROR = {
+    errors: [
+      {
+        title: 'Bad Gateway',
+        code: 'bad_gateway',
+        detail: 'Received an unusable response from Caseflow.',
+        status: 502
+      }
+    ]
+  }.freeze
+
+  def index
+    get_contestable_issues_from_caseflow
+    if caseflow_response_has_a_body_and_a_status?
+      render_response(caseflow_response)
+    else
+      render_unusable_response_error
+    end
+  end
+
+  private
+
+  attr_reader :caseflow_response, :backend_service_exception
+
+  def get_contestable_issues_from_caseflow
+    @caseflow_response = Caseflow::Service.new.get_contestable_issues headers: headers,
+                                                                      benefit_type: benefit_type,
+                                                                      appeal_type: appeal_type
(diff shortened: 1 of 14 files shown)


edmangimelli:
this is looking good 👍

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/base_contestable_issues_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+
+require 'caseflow/service'
+require 'common/exceptions'
+
+class AppealsApi::V1::DecisionReviews::BaseContestableIssuesController < AppealsApi::ApplicationController
+  skip_before_action(:authenticate)
+
+  EXPECTED_HEADERS = %w[X-VA-SSN X-VA-Receipt-Date].freeze
+
+  UNUSABLE_RESPONSE_ERROR = {
```
LindseySaari, line 11: This may look a bit cleaner as a class method instead of a constant

modules/appeals_api/spec/support/shared_examples_contestable_issues.rb
```diff
@@ -1,11 +1,9 @@
 # frozen_string_literal: true
 
-require 'rails_helper'
-
-describe AppealsApi::V1::DecisionReviews::HigherLevelReviews::ContestableIssuesController, type: :request do
+RSpec.shared_examples 'contestable issues index requests' do |options|
   let(:get_issues) do
     get(
-      '/services/appeals/v1/decision_reviews/higher_level_reviews/contestable_issues/compensation',
+      "/services/appeals/v1/decision_reviews/#{options[:appeal_type]}/contestable_issues/#{options[:benefit_type]}",
```
LindseySaari, line 6: I think single quotes is the preferred style

LindseySaari (approved):
One comment, but it looks like the code isn't new, rather copy/pasted into a new base controller, so i'll give a 👍

edmangimelli:
you _could_ change `appeal_type` to `decision_review_type`, if you want to be abundantly clear (since Caseflow calls an NOD an appeal).
non-blocking suggestion

edmangimelli (changes requested):
this looks good, but we're going to need to remove the `invalid benefit_type` 422 error response from the NOD contestableIssues documentation

modules/appeals_api/spec/requests/v1/notifce_of_disagreements/contestable_issues_controller_spec.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require_relative '../../../support/shared_examples_contestable_issues'
```
charleystran, line 4: could we use `AppealsApi::Engine.root.join` vs the `../`'s would be more readable in my opinion

templeman15: Updated!

edmangimelli (approved):
looks good!

PR #5059, opened 2020-10-09
Standardize error messages for file uploads

annaswims:
## Description of change
in preparation for department-of-veterans-affairs/va.gov-team#14002 we want to standardize the format of the error messages and give reasonable http status codes for file uploads.  We shouldn't be intentionally returning `500` for "something about the file uploaded wasn't what we wanted"

carrierwave already offered file size validation through the `size_range` method, so we can remove our `ValidateFileSize ` custom validator.
## Original issue(s)


spec/request/upload_supporting_evidence_spec.rb
```diff
@@ -9,10 +9,31 @@
     context 'with valid parameters' do
       it 'returns a 200 and an upload guid' do
         post '/v0/upload_supporting_evidence',
-             params: JSON.parse('{"supporting_evidence_attachment": {"file_data": "filename"}}')
+             params: { supporting_evidence_attachment: { file_data: fixture_file_upload('files/sm_file1.jpg') } }
         expect(response).to have_http_status(:ok)
         expect(JSON.parse(response.body)['data']['attributes']['guid']).to eq SupportingEvidenceAttachment.last.guid
       end
+
+      it 'returns a 422  for a malformed pdf' do
```
annaswims, line 17: @Mottie 
example error details are: 
`The uploaded PDF file is invalid and cannot be read`
`The file uploaded was too large`
`You are not allowed to upload "crt" files, allowed types: pdf, png, gif, tiff, tif, jpeg, jpg, bmp, txt`

The last one will be a bigger pain to change because it comes from a gem.

annaswims:
@erluti - This ads more coverage/confidence to the carrierwave upgrade.

LindseySaari:
Nice! I definitely like returning more meaningful status codes and messages, but just wondering if there was a specific reason that this wasn't implemented earlier in favor of a more generic 500 status code and error message? 🤷‍♀️

app/uploaders/profile_photo_attachment_uploader.rb
```diff
@@ -7,7 +7,9 @@ class ProfilePhotoAttachmentUploader < CarrierWave::Uploader::Base
   include AsyncProcessing
   include LogMetrics
 
-  MAX_FILE_SIZE = 10.megabytes
+  def self.max_file_size
```
LindseySaari, line 10: What is your reasoning for changing this to a class method vs a constant? I was thinking that if something is static, a constant is preferred, but since you’re referencing it in the ValidateFileSize class and it's publicly exposed, maybe the class method is the better choice. 

Sorry, probably nitpicky, lots of coffee this am :)

app/uploaders/validate_file_size.rb
```diff
@@ -8,6 +8,11 @@ module ValidateFileSize
   end
 
   def validate_file_size(file)
```
erluti, line 10: this file doesn't have a spec.  I took a stab at making one in my carrierwave update PR https://github.com/department-of-veterans-affairs/vets-api/blob/2b9e53b0cc86d525fffa93c95decfe2ebf82c91d/spec/uploaders/validate_file_size_spec.rb

app/uploaders/profile_photo_attachment_uploader.rb
annaswims: If I'm going to return a http 413 I need a way to trigger the "file too large" error.  I'd rather override the `max_file_size` than add a >50 mb file for a spec (50 is the max elsewhere).

LindseySaari (commented):
Let me know when you're done testing with the frontend and I can re-review after Philip merges in his [PR](https://github.com/department-of-veterans-affairs/vets-api/pull/5022#issuecomment-706248397)

annaswims:
I'm done testing with the FE. 🎉

annaswims:
Gonna come back to this one and update the text of the error messages, based upon @Mottie's suggestions.  
It needs more Canadian-style "We're Sorry"...
https://design.va.gov/patterns/messaging-dictionary#access-messaging

LindseySaari:
> Gonna come back to this one and update the text of the error messages, based upon @Mottie's suggestions.
> It needs more Canadian-style "We're Sorry"...
> https://design.va.gov/patterns/messaging-dictionary#access-messaging

Sounds good! Ping when you're ready 👍

annaswims:
I've changed my mind and will do text updates in a new PR.  Currently users see "internal server error" so our error messages are still an improvement.  @LindseySaari I added a commit to remove our extra validator code for file sizes.  Previously we were only validating max, so I've changed the behavior to validate that the file body is not empty as well (min 1 byte)

app/uploaders/supporting_evidence_attachment_uploader.rb
```diff
@@ -2,11 +2,12 @@
 
 # Files uploaded as part of a form526 submission that will be sent to EVSS upon form submission.
 class SupportingEvidenceAttachmentUploader < CarrierWave::Uploader::Base
-  include ValidateFileSize
   include SetAwsConfig
   include ValidatePdf
 
-  MAX_FILE_SIZE = 50.megabytes
+  def size_range
+    1...50.megabytes
```
edmangimelli, line 9: do we want this as `1..50.megabytes` (two dots) to include 50mb?

maybe even ` 1..50.1.megabytes.round`? --seems like most file managers report to the nearest 10th of a megabyte.

annaswims: this would inclued 50 mb https://github.com/carrierwaveuploader/carrierwave/blob/f15f4ba71dd2c78e8df494bf65938bbaf13960f9/lib/carrierwave/uploader/file_size.rb#L35

LindseySaari:
Can you add to the desc of the PR as to why the validate uploader can be removed?

spec/request/upload_supporting_evidence_spec.rb
LindseySaari: Is there a spec for file size too large?

LindseySaari:
Made a few comments @annaswims

spec/request/upload_supporting_evidence_spec.rb
annaswims: kind of, yes.  There's a spec that covers the size_range  range error https://github.com/department-of-veterans-affairs/vets-api/pull/5059/files#diff-e1189c09092bfde373ce490438c2dabbR38

annaswims: but beyond that I'd say carrierwave specs cover the behavior of their validator.

annaswims:
i think I've addressed all the comments.  Hopefully I didn't miss anything.

LindseySaari (approved):
Thanks for addressing the changes 👍

app/uploaders/supporting_evidence_attachment_uploader.rb
edmangimelli:
```
irb(main):001:0> (1...50).max
=> 49
```
triple dots won't include the top of the range

PR #5063, opened 2020-10-12
Add missing require

annaswims:
## Description of change
I found a missing require when I ran `spec/request/claim_documents_spec.rb` in isolation.

### DIFF
--- a/spec/request/claim_documents_spec.rb
+++ b/spec/request/claim_documents_spec.rb
@@ -3,10 +3,7 @@
 require 'rails_helper'
 RSpec.describe 'Claim Document Attachment', type: :request do
   let(:file) do
-    fixture_file_upload(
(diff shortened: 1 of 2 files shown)


spec/request/claim_documents_spec.rb
```diff
@@ -3,10 +3,7 @@
 require 'rails_helper'
 RSpec.describe 'Claim Document Attachment', type: :request do
   let(:file) do
-    fixture_file_upload(
-      "#{::Rails.root}/spec/fixtures/files/doctors-note.pdf",
-      'application/pdf'
-    )
+    fixture_file_upload('files/doctors-note.pdf')
```
LindseySaari, line 6: Do you need to specify 'application/pdf' type here as a second argument?

annaswims: nope, it's an optional arg.

LindseySaari (approved):
💯

PR #5067, opened 2020-10-12
Mobile letters endpoints

jperk51:
## Description of change
Created mobile letters, download, and beneficiary endpoints for mobile letters and documents.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#13461
department-of-veterans-affairs/va.gov-team#13459

Docs:
https://github.com/department-of-veterans-affairs/vets-api/blob/a8f8252db10397bbb58a5c296e9473be456cc621/modules/mobile/docs/openapi.yaml#L80-L267

Spec to test these endpoints (letters_request_spec.rb)


jjulian:
> Is there more we can do to make the same api endpoints work for both vets-website and mobile? It'd be good to get together and chat about this soon. 😄

Agreed! The team's focus right now is getting the integration between mobile and vets-api working for our MVP, but I'd be happy to represent Team Mobile if you'd like to set something up to chat about this long-term goal. I can explain some of the thoughts that are going into mobile api design.

modules/mobile/app/controllers/mobile/v0/letters_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+        service_response = EVSS::Letters::Service.new(@current_user).get_letter_beneficiary
+        response_template = OpenStruct.new
+        response_template.id = @current_user.id
+        response_template.benefit_information = service_response.benefit_information
+        response_template.military_service = service_response.military_service
+        render json: Mobile::V0::LettersBeneficiarySerializer.new(response_template)
+      end
+
+      def download
+        unless EVSS::Letters::Letter::LETTER_TYPES.include? params[:id]
```
jjulian, line 31: Would it clarify to name the parameter `:type` rather than `:id`?

modules/mobile/config/routes.rb
```diff
@@ -5,8 +5,11 @@
 
   namespace :v0 do
     get '/user', to: 'users#show'
+    get '/letters', to: 'letters#letters'
```
jjulian, line 8: Rails convention would name this action `letters#index`

modules/mobile/config/routes.rb
```diff
@@ -5,8 +5,11 @@
...
+    get '/letters/beneficiary', to: 'letters#beneficiary'
     get '/military-service-history', to: 'military_information#get_service_history'
     put '/user/addresses', to: 'addresses#update'
     put '/user/emails', to: 'emails#update'
+    post '/letters/:id', to: 'letters#download'
```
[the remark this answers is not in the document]
jperk51: Agreed. Added.

modules/mobile/config/routes.rb
jperk51: fixed

modules/mobile/app/controllers/mobile/v0/letters_controller.rb
jperk51: Agreed. Renamed

modules/mobile/app/controllers/mobile/v0/letters_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+        render json: Mobile::V0::LettersBeneficiarySerializer.new(response_template)
+      end
+
+      def download
+        unless EVSS::Letters::Letter::LETTER_TYPES.include? params[:type]
+          Raven.tags_context(team: 'va-mobile-app') # tag sentry logs with team name
+          raise Common::Exceptions::ParameterMissing, 'letter_type', "#{params[:type]} is not a valid letter type"
+        end
+
+        response = EVSS::Letters::DownloadService.new(@current_user).download_letter(params[:type], request.body.string)
```
jjulian, line 36: Looks like the request body is passed through directly to the backend service. Does there need to be any protection on params passed thru?

modules/mobile/docs/openapi.yaml
```diff
@@ -195,16 +195,16 @@ paths:
...
       summary: /v0/letters/{id}
       description: Returns requested letter for download
-      operationId: downloadLetterWithOptions
+      operationId: download
       security:
         - Bearer: [ ]
       parameters:
         - in: path
-          name: id
+          name: type
```
jjulian, line 207: Is the address passed in here?

modules/mobile/app/controllers/mobile/v0/letters_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+require 'evss/letters/service'
+
+module Mobile
+  module V0
+    class LettersController < ApplicationController
+      before_action { authorize :evss, :access? }
+
+      def index
+        service_response = EVSS::Letters::Service.new(@current_user).get_letters
+        response_template = OpenStruct.new
```
kreek, line 14: OpenStruct takes a hash, so you could do
```ruby
response_template = OpenStruct.new(service_response.to_h)
response_template.id = @current_user.id
render json: Mobile::V0::LettersSerializer.new(response_template)
```

modules/mobile/docs/openapi.yaml
jperk51: The address seems to be pulled in by the backend service. Even on .gov the current letters endpoints have no awareness of the address.

modules/mobile/app/controllers/mobile/v0/letters_controller.rb
jperk51: `{
"militaryService": true,
"serviceConnectedDisabilities": true,
"serviceConnectedEvaluation": true,
"nonServiceConnectedPension": true,
"monthlyAward": true,
"unemployable": true,
"specialMonthlyCompensation": true,
"adaptedHousing": true,
"chapter35Eligibility": true,
"deathResultOfDisability": true,
"survivorsAward": true
}`

This is the request body, would you expect any of these to need to be protected? I honestly have no idea if the body should be protected in this case or not.

modules/mobile/app/controllers/mobile/v0/letters_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+        response_template.id = @current_user.id
+        response_template.letters = service_response.letters
+        response_template.full_name = service_response.full_name
+        render json: Mobile::V0::LettersSerializer.new(response_template)
+      end
+
+      def beneficiary
```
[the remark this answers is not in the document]
jperk51: Will do

modules/mobile/app/controllers/mobile/v0/letters_controller.rb
kreek: A regular struct will always be faster though
``` ruby
LettersStruct = Struct.new(:id, :letters) # below the LettersController class end

service_response = service.get_letters
response_template = LettersStruct.new(@current_user.id, service_response.letters)
render json: Mobile::V0::LettersSerializer.new(response_template)
```

modules/mobile/app/controllers/mobile/v0/letters_controller.rb
jjulian: I was just reminded of the rails pattern to filter out unexpected query params before processing, and perhaps we should consider doing the same here. @kreek, is this a concern?

modules/mobile/docs/openapi.yaml
jjulian: Ah ok, I am remembering the discussion about the UX where the user will choose an address to be printed on the letter, but right, that is saved through the profile.

modules/mobile/app/controllers/mobile/v0/letters_controller.rb
kreek: And... since this is all to do with how the data is serialized you could bend the serializer to take an id
```ruby
module Mobile
  module V0
    class LettersSerializer
      include FastJsonapi::ObjectSerializer
      
      set_type :letters
      attributes :letters
      
      def initialize(id, resource, options = {})
        resource = LettersStruct.new(id, resource.letters)
        super(resource, options = {})
      end
    end
    
    LettersStruct = Struct.new(:id, :letters)
  end
end
```

modules/mobile/app/controllers/mobile/v0/letters_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+require 'evss/letters/download_service'
+require 'evss/letters/service'
+
+module Mobile
+  module V0
+    class LettersController < ApplicationController
+      before_action { authorize :evss, :access? }
+
+      def letters
+        service_response = EVSS::Letters::Service.new(@current_user).get_letters
```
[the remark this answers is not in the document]
jperk51: Will do

kreek (approved):
LGTM 🚀 - very clean/easy to read and great test coverage

modules/mobile/spec/request/letters_request_spec.rb
```diff
@@ -0,0 +1,325 @@
...
+                      'letterType' => 'civil_service'
+                    },
+                    {
+                      'name' => 'Benefit Summary Letter',
+                      'letterType' => 'benefit_summary'
+                    },
+                    {
+                      'name' => 'Benefit Verification Letter',
+                      'letterType' => 'benefit_verification'
+                    }
```
thilton-oddball, line 50: Does "certificate_of_eligibility" need to be included here? It may not, I just noticed that it was the only letter_type listed in the schema that was not included here.

thilton-oddball (approved):
One comment to look at but otherwise looks good.

modules/mobile/spec/request/letters_request_spec.rb
jperk51: This is because while certificate_of_eligibility is something the service could return, but the VCR cassette used by the test doesn't include it

PR #5069, opened 2020-10-12
migrations for #5068

annaswims:
## Description of change
migrations for #5068, drop unused table

## Original issue(s)
department-of-veterans-affairs/va.gov-team#12652


LindseySaari (approved):
🥳

PR #5072, opened 2020-10-12
Allow password-protected pdfs (HCA, Preneeds, VIC, form526, CST)

annaswims:
## Description of change
if a veteran supplies a password we can use pdftk to create an unlocked version of the pdf. If no password is provided, the api behavior is unchanged.

## Original issue(s)

I intend to do some e2e testing with a variety of larger files before merging, but didn't want to commit a lot of files to the repo.


app/models/form_attachment.rb
```diff
@@ -34,6 +34,22 @@ def get_file
...
+
+    pdftk = PdfForms.new(Settings.binaries.pdftk)
+    tmpf = Tempfile.new('decrypted')
+    error_messages = pdftk.call_pdftk(file.tempfile.path, 'input_pw', file_password, 'output', tmpf.path)
+    if error_messages.present?
+      log_message_to_sentry(error_messages, 'warn')
+      raise Common::Exceptions::UnprocessableEntity.new(detail: I18n.t('uploads.pdf.incorrect_password'),
+                                                        source: 'FormAttachment.unlock_pdf')
+    end
+    file.tempfile.unlink
```
annaswims, line 48: Unlinking the current tempfile and replacing this with the new one works, but I'd appreciate some extra scrutiny here.

app/models/form_attachment.rb
```diff
@@ -34,6 +34,24 @@ def get_file
 
   private
 
+  def unlock_pdf(file, file_password)
+    return file unless File.extname(file) == '.pdf'
+
+    pdftk = PdfForms.new(Settings.binaries.pdftk)
+    tmpf = Tempfile.new('decrypted')
```
thilton-oddball, line 41: I'm not sure if it is matters but do you need to specify the extension for the temp file here with Tempfile.new(['decrypted', '.pdf'])?

app/models/form_attachment.rb
```diff
@@ -34,6 +34,24 @@ def get_file
...
+    error_messages = pdftk.call_pdftk(file.tempfile.path, 'input_pw', file_password, 'output', tmpf.path)
```
thilton-oddball, line 42: Also not sure if it matters but do you need to call tmpf.close before it replaces the current temp file?

thilton-oddball (approved):
All looks good!

PR #5073, opened 2020-10-13
Add documentation and specs for Claim Status Tool file uploads

annaswims:
## Description of change
Before we start allowing password-protected pdf's (#5072) I want to make sure our file upload code is well-tested and documented.  

This adds swagger docs and rescues an exception that would've raised 500's

in #5070 I'm taking a look at the text of the error messages.

## Original issue(s)

### DIFF
--- a/spec/request/documents_spec.rb
+++ b/spec/request/documents_spec.rb
@@ -41,7 +41,7 @@ RSpec.describe 'Documents management', type: :request do
   end
 
   context 'with unaccepted file_type' do
-    let(:file) { fixture_file_upload('files/invalid_idme_cert.crt', 'application/pdf') }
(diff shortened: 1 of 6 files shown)


spec/request/documents_spec.rb
```diff
@@ -45,13 +40,19 @@
...
   end
 
-  context 'with locked PDF' do
-    let(:locked_file) do
-      fixture_file_upload(
-        "#{::Rails.root}/spec/fixtures/files/locked-pdf.pdf",
-        'application/pdf'
-      )
+  context 'with unaccepted file_type' do
+    let(:file) { fixture_file_upload('files/invalid_idme_cert.crt', 'application/pdf') }
```
LindseySaari, line 44: Is the 'application/pdf' type intentional here? It mismatches the crt file being passed in

LindseySaari (approved):

PR #5075, opened 2020-10-13
Application directory

bradenshipley:
## Description of change

This PR exists to create a new module in vets-api with 2 public endpoints to support an Application Directory that shows the available applications that veterans can connect their VA data to. The related epic is https://vajira.max.gov/browse/API-1125.  This PR has a large amount of commits due to blockers and design changes that popped up during initial development, but the final product is the MVP.
## Original issue(s)
https://vajira.max.gov/browse/API-1125
This PR is large but the size is a necessity due to the large number of files required for a new module.

Tests have been written for models and controller requests. All files have been style checked with rubocop and corrected.


spec/models/directory_application_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'rails_helper'
+
+RSpec.describe DirectoryApplication, type: :model do
+  it 'is invalid without valid attributes' do
+    expect(DirectoryApplication.new).to_not be_valid
```
LindseySaari, line 5: It looks like not_to is preferred over to_not

spec/support/vcr.rb
```diff
@@ -5,7 +5,7 @@
   c.cassette_library_dir = 'spec/support/vcr_cassettes'
   c.hook_into :webmock
   # experiencing VCR-induced frustation? uncomment this:
-  # c.debug_logger = File.open('vcr.log', 'w')
+  c.debug_logger = File.open('vcr.log', 'w')
```
LindseySaari, line 8: debug statement hanging around

LindseySaari (commented):
At first glance, I’d recommend that the migration changes be broken out into a separate PR.

config/sidekiq_scheduler.yml
```diff
@@ -165,3 +165,8 @@ VaForms::FetchLatest:
   cron: "0 0 * * * America/New_York"
   class: VaForms::FormReloader
   description: "Fetches latest VA Forms"
+
+AppsApi::ReloadAppsList:
+  cron: "0 */8 * * * America/New_York"
+  class: AppsApi::DirectoryReloader
```
LindseySaari, line 171: I don't see any changes for the AppsApi::DirectoryReloader class in this PR

db/migrate/20201005134940_fix_directory_application_column_name.rb
```diff
@@ -0,0 +1,5 @@
+class FixDirectoryApplicationColumnName < ActiveRecord::Migration[6.0]
+  def change
+    rename_column :directory_applications, :type, :app_type
```
LindseySaari, line 3: It looks like this name change can just be applied to the original table creation migration

db/migrate/20201005135415_fix_directory_application_service_categories_column_name.rb
```diff
@@ -0,0 +1,5 @@
+class FixDirectoryApplicationServiceCategoriesColumnName < ActiveRecord::Migration[6.0]
+  def change
+    rename_column :directory_applications, :service_cattegories, :service_categories
```
LindseySaari, line 3: Same here and the column rename below as well

lib/okta/service.rb
```diff
@@ -31,6 +32,24 @@ def app(app_id)
       end
     end
 
+    def get_apps(url)
+      with_monitoring do
+        get_url_with_token(url)
```
LindseySaari, line 37: You may be able to combine the get_apps and get_auth_servers methods. They look pretty similar to me and just pass in a url as a param. Just a suggestion.

config/sidekiq_scheduler.yml
bradenshipley: This was a piece of dead code that slipped through. It's been removed

lib/okta/service.rb
bradenshipley: get_apps is a piece of dead code, since the apps are now being stored in records instead of grabbing them externally from Okta. I'll remove the get_apps function

lib/okta/directory_service.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
```
LindseySaari, line 1: Code coverage is a bit low for this service class

LindseySaari:
I think it might make more sense to break out some of the boilerplate module code into a PR and then smaller PRs to introduce the rest of the changes

Gemfile.lock
```diff
@@ -143,6 +143,7 @@ PATH
 
 GEM
```
LindseySaari: the sidekiq pro and ent changes shouldn't be committed

lib/okta/directory_service.rb
bradenshipley: code coverage is at 100% with latest commit #986803c

LindseySaari:
Danger is erroring on /app and /db changes in the same PR. Did you see the comments on breaking up this PR in other ways too? Although this PR mainly code for a new module, conceptually there's multiple changes being introduced here and would be easier to review if broken into smaller PRs.

modules/apps_api/db/migrate/20201012135852_remove_permissions_from_directory_applications.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class RemovePermissionsFromDirectoryApplications < ActiveRecord::Migration[6.0]
```
LindseySaari, line 3: This migration can be removed all together if you change the original migration to create the table without the permissions field

bradenshipley: Done!

thilton-oddball (approved):
With the previously requested changes having been made, going ahead and signing off on this.

PR #5076, opened 2020-10-13
Remove Origami gem

annaswims:
## Description of change
There is no reason to use the `origami` gem.  We're already doing pdf validations with PdfInfo


annaswims:
Thanks! I've resolved the conflicts.

LindseySaari (approved):

PR #5082, opened 2020-10-13
14771 appointment cancellation bug temporary patch

saneshark:
## Description of change
VAMF recently started sending us a 500 error instead of a 409 error for something we had special handling around. We have to patch this temporarily while they work out a solution. More on that in the original issue below.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/14771

VCR cassette added. This patch is temporarily. It will still be noisy in Sentry. We're going to have to address that differently.

### DIFF
--- a/modules/vaos/app/services/vaos/middleware/response/errors.rb
+++ b/modules/vaos/app/services/vaos/middleware/response/errors.rb
@@ -9,7 +9,7 @@ module VAOS
         def on_complete(env)
           return if env.success?
 
-          Raven.extra_context(message: env.body, url: env.url)
+          Raven.extra_context(vamf_status: env.status, vamf_body: env.body, vamf_url: env.url)
           case env.status
           when 400, 409
             error_400(env.body)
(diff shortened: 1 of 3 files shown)


modules/vaos/app/services/vaos/middleware/response/errors.rb
```diff
@@ -18,12 +18,20 @@ def on_complete(env)
...
           when 500..510
-            raise Common::Exceptions::BackendServiceException.new('VAOS_502', source: self.class)
+            error_500(env.body)
           else
             raise Common::Exceptions::BackendServiceException.new('VA900', source: self.class)
           end
         end
 
+        def error_500(body)
+          if /APTCRGT/.match?(body)
```
LindseySaari, line 28: Could you add an inline comment here stating that this solution is temporary

LindseySaari (approved):
One inline comment, but otherwise LGTM 👍

ajmagdub (approved):
LGTM. As discussed changing the labels in line 12 of  the errors.rb file may affect how the errors are grouped in Sentry10 , but not adversely.

grossb1 (approved):

PR #5083, opened 2020-10-13
Remove broken pdf validation for 526 attachments

hinzed1127:
## Description of change

The `PdfInfo ` module is broken and was throwing errors for valid PDFs being submitted to the 526 attachment endpoint. This temporarily removes those checks

## Original issue(s)
https://vajira.max.gov/browse/API-1979

- Manually tested by submitting different PDF attachments locally and uploading to s3

### DIFF
--- a/modules/claims_api/app/controllers/claims_api/base_disability_compensation_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/base_disability_compensation_controller.rb
@@ -84,7 +84,7 @@ module ClaimsApi
       log_message_to_sentry('Upload error in 526', :error, body: e.message)
 
       {
-        errors: [{ status: 422, detail: e.message, source: e.key }]
+        errors: [{ status: 422, detail: e&.message, source: e&.key }]
       }.to_json
     end
   end
--- a/modules/claims_api/app/controllers/claims_api/base_disability_compensation_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/base_disability_compensation_controller.rb
@@ -10,6 +10,10 @@ module ClaimsApi
     STATSD_VALIDATION_FAIL_KEY = 'api.claims_api.526.validation_fail'
     STATSD_VALIDATION_FAIL_TYPE_KEY = 'api.claims_api.526.validation_fail_type'
 
+    # TODO Fix methods in document_validations to work correctly before uncommenting, include broader range of tests
+    # before_action :validate_documents_content_type, only: %i[upload_form_526]
+    # before_action :validate_documents_page_size, only: %i[upload_form_526]
+
     def upload_form_526
       pending_claim = ClaimsApi::AutoEstablishedClaim.pending?(params[:id])
       pending_claim.set_file_data!(documents.first, params[:doc_type])
--- a/modules/claims_api/app/controllers/claims_api/base_disability_compensation_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/base_disability_compensation_controller.rb
@@ -84,7 +84,7 @@ module ClaimsApi
       log_message_to_sentry('Upload error in 526', :error, body: e.message)
 
       {
-        errors: [{ detail: e.message }]
+        errors: [{ status: 422, detail: e.message, source: e.key }]
       }.to_json
     end
   end
--- a/modules/claims_api/app/controllers/claims_api/base_disability_compensation_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/base_disability_compensation_controller.rb
@@ -18,8 +18,6 @@ module ClaimsApi
       ClaimsApi::ClaimUploader.perform_async(pending_claim.id)
 
       render json: pending_claim, serializer: ClaimsApi::AutoEstablishedClaimSerializer
-    rescue
-      render json: unprocessable_response, status: :unprocessable_entity
     end
 
     # rubocop:disable Metrics/MethodLength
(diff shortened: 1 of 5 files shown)


modules/claims_api/app/controllers/claims_api/base_disability_compensation_controller.rb
```diff
@@ -18,8 +18,6 @@ def upload_form_526
       ClaimsApi::ClaimUploader.perform_async(pending_claim.id)
 
       render json: pending_claim, serializer: ClaimsApi::AutoEstablishedClaimSerializer
-    rescue
-      render json: unprocessable_response, status: :unprocessable_entity
```
hinzed1127, line 22: while this was initially removed for debugging purposes, should it stay out or perhaps be updated to be more descriptive?

annaswims:
@hinzed1127 I'm in the middle of troubleshooting a variety of file upload related issues for BAM.  Would you be able to tell me more about what kinds of errors you're seeing?  I don't have access to your Jira.

modules/claims_api/app/controllers/claims_api/base_disability_compensation_controller.rb
bastosmichael: I think the idea of capturing all of the errors was a good one but rushed too quickly in terms of implementation. My recommendation here is instead of removing it do a `rescue => e` and capture the actual error coming in and pass it to the unprocessable_response method to actually give us what the error and message is or might be this way a proper error can be properly processed. Also make sure to send the error using a log to sentry method, we have some examples of that in the controllers.

https://github.com/department-of-veterans-affairs/vets-api/blob/a0df370e2f1bd50367f841c23ff53965be3858fb/modules/claims_api/app/controllers/claims_api/base_disability_compensation_controller.rb#L79-L83

modules/claims_api/app/controllers/claims_api/v0/forms/disability_compensation_controller.rb
```diff
@@ -14,8 +14,8 @@ class DisabilityCompensationController < BaseDisabilityCompensationController
 
         skip_before_action(:authenticate)
         before_action :validate_json_schema, only: %i[submit_form_526 validate_form_526]
-        before_action :validate_documents_content_type, only: %i[upload_supporting_documents]
-        before_action :validate_documents_page_size, only: %i[upload_supporting_documents]
+        # before_action :validate_documents_content_type, only: %i[upload_supporting_documents]
```
bastosmichael, line 17: Include a TODO above it about what needs to happen here, also make sure to include this validation action where the PUT method is also importing a document so that we know to tackle both when the time comes.

modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
```diff
@@ -17,8 +17,8 @@ class DisabilityCompensationController < BaseDisabilityCompensationController
         before_action { permit_scopes %w[claim.write] }
         before_action :validate_json_schema, only: %i[submit_form_526 validate_form_526]
         before_action :validate_initial_claim, only: %i[submit_form_526 validate_form_526]
-        before_action :validate_documents_content_type, only: %i[upload_supporting_documents]
-        before_action :validate_documents_page_size, only: %i[upload_supporting_documents]
+        # before_action :validate_documents_content_type, only: %i[upload_supporting_documents]
```
bastosmichael, line 20: Include a TODO above it about what needs to happen here, also make sure to include this validation action where the PUT method is also importing a document so that we know to tackle both when the time comes.

bastosmichael (changes requested):
Need to include a few changes to improve both the error messages as well as well as track what needs to happen to the validation document methods etc.

hinzed1127:
@annaswims I'll try to do a more thorough breakdown later, but the main issue I was seeing wasn't recognizing valid PDFs as PDF files, so hitting POST `/forms/526/{id}/attachments` was resulting in responses like this for every file I was attempting:

```json
{
  "errors": [
    {
      "status": 422,
      "source": "10-0003K-2.pdf",
      "detail": "10-0003K-2.pdf must be in PDF format"
    }
  ]
}
```
The same uploads were going through fine for PUT `/forms/526/{id}`, which doesn't have the `PdfInfo` validations, so something's going wonky there

hinzed1127:
@bastosmichael Idk if we want to update `modules/claims_api/spec/requests/v0/disability_compensation_request_spec.rb` at all to include things specific to the error messaging (it already has a test for the 422 starting on line 239), otherwise I think this is closer to what we want

modules/claims_api/app/controllers/claims_api/base_disability_compensation_controller.rb
```diff
@@ -76,9 +80,11 @@ def track_526_validation_errors(errors)
       end
     end
 
-    def unprocessable_response
+    def unprocessable_response(e)
+      log_message_to_sentry('Upload error in 526', :error, body: e.message)
+
       {
-        errors: [{ detail: 'An unknown error occurred. Please retry the request' }]
+        errors: [{ detail: e.message }]
```
bastosmichael, line 87: I want to also make sure that we match our API's published error model here as well per what's in the documentation, I should have been more clear about this in my last review.

https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/app/swagger/claims_api/common/error_model_swagger.rb

bastosmichael (changes requested):

bastosmichael (approved):
Looks good now, let's make sure we reference this PR in the changes we make later.

PR #5089, opened 2020-10-15
Include more pact information in webhook 

LindseySaari:
## Description of change
When we want to trace the Pact workflow, it's difficult to determine which pact triggered the verification task in CircleCI. Using dynamic variable substitution via the [pact-broker CLI](https://github.com/pact-foundation/pact_broker-client/#create-webhook),  parameters are added to the request body which contain info about the consumer, tags, etc and sent along to Circle. When the verification task runs, the consumer information is placed into an artifact after the pact_verification workflow runs for tracing purposes.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#14733

- Tested in Circle and triggered the webhook via the HAL browser/inspector in the pact broker 
- Todo: After this PR is merged, the webhook needs to be updated to run against the master branch in Circle instead of this current branch.


thilton-oddball (approved):
LGTM

PR #5090, opened 2020-10-15
added endpoints to doc

jperk51:
## Description of change
Mobile API docs for getting and updating direct deposit info

## Original issue(s)
department-of-veterans-affairs/va.gov-team#14081

Mobile API swagger update
No tests. Docs only change.


modules/mobile/docs/openapi.yaml
```diff
@@ -766,6 +895,67 @@ components:
...
+                  example: "Savings account"
+                  enum:
+                    - Savings account
+                    - Checking account
+                financialInstitutionName:
+                  type: string
+                  example: "PACIFIC PREMIER BANK"
+                financialInstitutionRoutingNumber:
+                  type: string
+                  example: "*****9982"
```
jjulian, line 958: I think ABA routing numbers are public, and do not need to be obscured.

modules/mobile/docs/openapi.yaml
```diff
@@ -766,6 +895,67 @@ components:
...
+    UpdatePaymentInfoRequest:
+      required:
+        - accountNumber
+        - accountType
+        - financialInstitutionName
+        - financialInstitutionRoutingNumber
+      properties:
+        accountNumber:
+          type: string
+          example: "12345678"
```
jjulian, line 907: account numbers are 10-12 digits

modules/mobile/docs/openapi.yaml
```diff
@@ -756,7 +756,136 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Errors"
-
+  /v0/payment_information:
+    get:
+      summary: /v0/payment_information
```
jjulian, line 761: Consider naming of this resource. Suffixes like "information" are often just extra. Are these "payments"? "bank_accounts"? "payment_accounts"?

modules/mobile/docs/openapi.yaml
jperk51: Fixed.

modules/mobile/docs/openapi.yaml
jperk51: Fixed.

modules/mobile/docs/openapi.yaml
jperk51: I'm with you on the suffixes thing. I did have a short convo with @kreek about an additional object that should be part of this response that makes having the suffix in this case make a bit more sense. Should be included in the yaml now.

modules/mobile/docs/openapi.yaml
```diff
@@ -756,7 +756,136 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Errors"
-
+  /v0/payment-information:
```
kreek, line 759: I believe @steve-gov said there's another system that stores direct-deposit in addition to PPIU. So maybe this could be `/v0/payment-information/benefits`?

jperk51: sounds good to me, I'll change it real quick

kreek (approved):
LGTM 🚀

thilton-oddball (approved):
LGTM

jjulian (approved):

PR #5092, opened 2020-10-15
Representative Oauth 500 fix due to user data error

bastosmichael:
## Description of change
The user data I was testing with changed because the application changed and I didn't catch the change.

## Original issue(s)

### DIFF
--- a/modules/veteran/app/models/veteran/user.rb
+++ b/modules/veteran/app/models/veteran/user.rb
@@ -20,7 +20,7 @@ module Veteran
     end
 
     def bgs_service
-      external_key = [@user.first_name, @user.last_name].join(' ')
+      external_key = "#{@user.first_name} #{@user.last_name}"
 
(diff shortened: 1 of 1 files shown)


modules/veteran/app/models/veteran/user.rb
```diff
@@ -20,10 +20,10 @@ def json_data
     end
 
     def bgs_service
-      external_key = @user.common_name || @user.email
+      external_key = [@user.first_name, @user.last_name].join(' ')
```
charleystran, line 23: Why not use string interpolation here "#{@user.first_name} #{@user.last_name}"?

bastosmichael: Changed, want to get it out before the next deploy.

hinzed1127 (approved):

templeman15 (approved):

PR #5100, opened 2020-10-16
14729 use mpi for mvi users controller

erluti:
MVI is an outdataed acronym, and this class is being renamed to reflect the modern acronym of MPI.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#14729


thilton-oddball (approved):
Looks good

PR #5101, opened 2020-10-16
14770 rename MviPolicy

LindseySaari:
# Description of change
app/policies/mvi_policy.rb defines MviPolicy and should become app/policies/mpi_policy.rb and use the inflection to be MPIPolicy. MPIPolicy is derived from the notion of policy classes from the pundit gem. The authorize before_action in the  mvi_users_controller automatically infers that mpi will have a matching MPIPolicy class. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#14705

- [x] Green Test Suite

### DIFF
--- a/app/policies/mpi_policy.rb
+++ b/app/policies/mpi_policy.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-MviPolicy = Struct.new(:user, :mvi) do
+MPIPolicy = Struct.new(:user, :mvi) do
   def access_add_person?
     user.edipi.present? && user.ssn.present? && (user.birls_id.blank? || user.participant_id.blank?)
   end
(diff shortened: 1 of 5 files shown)


app/policies/mpi_policy.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-MviPolicy = Struct.new(:user, :mvi) do
+MPIPolicy = Struct.new(:user, :mvi) do
```
thilton-oddball, line 3: Does the :mvi that is getting passed to the struct need to change to :mpi or should it remain :mvi?

thilton-oddball (approved):
One comment that may or may not require action but otherwise everything looks good.

app/policies/mpi_policy.rb
LindseySaari: Yeah the above mvi shouldn't matter. The :mpi authorization in the before action corresponds to the actual MPIPolicy.

PR #5105, opened 2020-10-19
7782 changes to ppms queries pt4 integration with feature flag

darkbushido:
## Description of change
This PR allows the V1 facilities community care endpoint to use the new `Facilities::PPMS::V1::Client`
it also fixes some of the duplication logic that was magically being handled by Fast Json api

## Original issue(s)
department-of-veterans-affairs/va.gov-team#7782
department-of-veterans-affairs/va.gov-team#14231

* Added `:facility_locator_ppms_use_v1_client` Feature Flag
* The endpoint should return the same data with and without the code


app/controllers/v1/facilities/ccp_controller.rb
```diff
@@ -35,16 +32,41 @@ def specialties
...
+  def index_api_results
+    if Flipper.enabled?(:facility_locator_ppms_use_v1_client)
+      index_api_results_v1
+    else
+      index_api_results_v0
+    end
+  end
+
+  def index_api_results_v0
+    ppms_search.collect do |result|
```
erluti, line 49: it looks like in the search v1 uses pagination, and here it's v0 using pagination?  Do I understand that correctly?  I'm surprised they don't align.

spec/request/v1/facilities/ccp_request_spec.rb
```diff
@@ -10,496 +10,527 @@
...
+              )
+            )
+
+            expect(StatsD).to receive(:measure).with(
+              'facilities.ppms.providers',
+              kind_of(Numeric),
+              hash_including(
+                tags: ['facilities.ppms']
+              )
+            ).exactly(7).times
```
erluti, line 154: Why did this number change?

lib/facilities/ppms/v1/client.rb
```diff
@@ -143,7 +153,7 @@ def provider_locator_params(params)
             hsh["specialtycode#{index + 1}".to_sym] = code
           end
 
-          specialty_codes.merge(base_params(params))
+          base_params(params).merge(specialty_codes)
```
erluti, line 156: was this a bug before?

erluti (commented):
generally looks good, a few questions before I stamp it

app/controllers/v1/facilities/ccp_controller.rb
darkbushido: https://github.com/department-of-veterans-affairs/vets-api/blob/7782-changes-to-ppms-queries-pt4-integration-with-feature-flag/lib/facilities/ppms/v1/response.rb#L51-L56

its one of the improvements for v1

spec/request/v1/facilities/ccp_request_spec.rb
darkbushido: This is a bugfix, we weren't deduplicating results correctly

lib/facilities/ppms/v1/client.rb
darkbushido: they are the same, I'm just trying to get the request hash to have the address information before the other params are added

erluti: the merged hash overrides the original hash's keys, so it just surprised me as a functional change to reverse those
```irb(main):001:0> {thing1: 'hi'}.merge({thing1: 'override'})
=> {:thing1=>"override"}```

app/controllers/v1/facilities/ccp_controller.rb
erluti: ah,  i thought this was adding `.paginate(pagination_params)` to the v0.  I see better what's happening.

erluti (approved):
LGTM

lib/facilities/ppms/v1/client.rb
erluti: I guess to clarify further, I would assume this is the way it should have been originally, but it's not obvious to me if there was a reason base_params should override `params[:specialties]`.  But it looks like they never sharer keys.

erluti:
Merged without dangerbot after review

PR #5107, opened 2020-10-19
Modify how is_cerner is derived

kelsonic:
## Description of change
**Issue:** https://github.com/department-of-veterans-affairs/va.gov-team/issues/14203

This PR changes how `is_cerner` is derived so that it's not only dependent on MVI but also on Flipper feature toggles for exceptions (e.g. Chalmers should be `is_cerner: true`).

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/14203

No.

**Yes, several feature flags have been modified.**

I removed the following feature toggles since they are not used:

```
show_new_get_medical_records_page
show_new_refill_track_prescriptions_page
show_new_secure_messaging_page
show_new_view_test_lab_results_page
```

We currently still use `show_new_schedule_view_appointments_page`, but I'll send up a follow-up PR post-Cerner launch to remove that too.

I added the following feature toggles:

```
cerner_override_668
cerner_override_757
```

These are used to force `is_cerner` to true for these 2 facilities.

No.

No.

No.

No.

### DIFF
--- a/app/services/users/profile.rb
+++ b/app/services/users/profile.rb
@@ -163,10 +163,7 @@ module Users
 
     def facility(facility_id)
       cerner_facility_ids = user.va_profile.cerner_facility_ids || []
-
-      is_cerner = cerner_facility_ids.include?(facility_id)
-      is_cerner = true if facility_id == '668' && Flipper.enabled?(:cerner_override_668)
-      is_cerner = true if facility_id == '757' && Flipper.enabled?(:cerner_override_757)
+      is_cerner = cerner_facility_ids.include?(facility_id) || Flipper.enabled?("cerner_override_#{facility_id}")
 
       {
         facility_id: facility_id,
--- a/app/services/users/profile.rb
+++ b/app/services/users/profile.rb
@@ -163,9 +163,10 @@ module Users
 
     def facility(facility_id)
       cerner_facility_ids = user.va_profile.cerner_facility_ids || []
+      is_cerner = cerner_facility_ids.include?(facility_id) || Settings.cerner.facility_ids.include?(facility_id)
       {
         facility_id: facility_id,
-        is_cerner: cerner_facility_ids.include?(facility_id)
+        is_cerner: is_cerner
       }
     end
 
--- a/app/services/users/profile.rb
+++ b/app/services/users/profile.rb
@@ -163,7 +163,11 @@ module Users
 
     def facility(facility_id)
       cerner_facility_ids = user.va_profile.cerner_facility_ids || []
-      is_cerner = cerner_facility_ids.include?(facility_id) || Settings.cerner.facility_ids.include?(facility_id)
+
+      is_cerner = cerner_facility_ids.include?(facility_id)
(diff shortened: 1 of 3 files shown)


config/settings.yml
```diff
@@ -614,3 +614,9 @@ genisis:
   service_path: /COVID19Service
   user: TestUser
   pass: bogus
+
+# Settings for live Cerner facilities
+cerner:
+  facility_ids:
+    - 757
+    - 668
```
kelsonic, line 622: This is just locally for now so I think it should be fine and will allow us to update `vets-website` to support this new behavior for the `isCerner` flag on each user's facilities

kelsonic:
Closed devops PR in favor of updating this PR to be dependent on Flipper

app/services/users/profile.rb
```diff
@@ -163,9 +163,14 @@ def update_status_and_errors
 
     def facility(facility_id)
       cerner_facility_ids = user.va_profile.cerner_facility_ids || []
+
+      is_cerner = cerner_facility_ids.include?(facility_id)
+      is_cerner = true if facility_id == '668' && Flipper.enabled?(:cerner_override_668)
```
kelsonic, line 168: @patrickvinograd could you let me know if you have any suggestions here? 🙂

config/features.yml
```diff
@@ -244,26 +244,18 @@ features:
...
+  cerner_override_757:
     actor_type: user
     description: >
-      This will show the non-Cerner-user and Cerner-user content for the page /health-care/refill-track-prescriptions/
+      This will show the Cerner facility 757 as `isCerner`.
   show_new_schedule_view_appointments_page:
     actor_type: user
     description: >
       This will show the non-Cerner-user and Cerner-user content for the page /health-care/schedule-view-va-appointments/
-  show_new_secure_messaging_page:
```
kelsonic, line 259: I removed the following feature toggles since they are not used:

```
show_new_get_medical_records_page
show_new_refill_track_prescriptions_page
show_new_secure_messaging_page
show_new_view_test_lab_results_page
```

We currently still use `show_new_schedule_view_appointments_page`, but I'll send up a follow-up PR post-Cerner launch to remove that too.

I added the following feature toggles:

```
cerner_override_668
cerner_override_757
```

These are used to force `isCerner` to true for these 2 facilities.

thilton-oddball (approved):
LGTM

app/services/users/profile.rb
ncksllvn: I really don't know how to write Ruby very well but I am wondering if we can rely on a standard naming convention here to do a dynamic lookup of facility IDs in the Cerner overrides. For example, instead of 

```
is_cerner = true if facility_id == '668' && Flipper.enabled?(:cerner_override_668)
is_cerner = true if facility_id == '757' && Flipper.enabled?(:cerner_override_757)
```

We do

```
is_cerner = cerner_facility_ids.include?(facility_id) || Flipper.enabled?("cerner_override_#{facility_id}")
```

That way we can add new Cerner overrides without needing to update the logic

kelsonic: Updated! 🙂

PR #5108, opened 2020-10-19
Adding the ability to exclude mobile facilities from the results

darkbushido:
## Description of change
We do not want to display mobile vet centers at this time.
We will take a param `exclude_mobile` that will filter out any facility with mobile = true 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#13730

Nothing special other than adding the param


thilton-oddball (approved):
LGTM

PR #5109, opened 2020-10-19
Re adding a feature flag used by the front end

darkbushido:
## Description of change
Re-Adding the feature flag for the front end

## Original issue(s)
department-of-veterans-affairs/va.gov-team#14154

* Feature Flag `facilities_ppms_suppress_pharmacies`

### DIFF
--- a/config/features.yml
+++ b/config/features.yml
@@ -38,12 +38,15 @@ features:
   facility_locator_ppms_legacy_urgent_care_to_pos_locator:
     actor_type: user
     description: force the legacy urgent care path to use the new POS locator
-  facility_locator_ppms_location_query:
+  facility_locator_ppms_use_v1_client:
     actor_type: user
-    description: Use the Lat/Long instead of an address
+    description: Have the V1 API use the PPMS::V1::Client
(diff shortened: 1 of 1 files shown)


config/features.yml
```diff
@@ -38,12 +38,15 @@ features:
   facility_locator_ppms_legacy_urgent_care_to_pos_locator:
     actor_type: user
     description: force the legacy urgent care path to use the new POS locator
-  facility_locator_ppms_location_query:
+  facility_locator_ppms_use_v1_client:
```
thilton-oddball, line 41: Just double checking that you also wanted to change 'facility_locator_ppms_location_query' to 'facility_locator_ppms_use_v1_client'

thilton-oddball (approved):
One comment that should be reviewed but otherwise looks good.

config/features.yml
darkbushido: yep, `: facility_locator_ppms_location_query` was removed with its logic and replaced with `: facility_locator_ppms_use_v1_client`

darkbushido: actually, this change belongs in a different branch. fixing it

PR #5114, opened 2020-10-19
Revert "caregivers google analytics event"

lihanli:
based on discussions with stakeholders we have decided not to send this GA event anymore from the backend
https://github.com/department-of-veterans-affairs/va.gov-team/issues/8081#issuecomment-712333586

Reverts department-of-veterans-affairs/vets-api#4983


thilton-oddball (approved):
Looks good!

PR #5115, opened 2020-10-19
add lihan to flipper admin users

lihanli:
## Description of change
add myself to flipper admin users so i can change ch33 direct deposit flipper settings


thilton-oddball (approved):
Looks good

PR #5118, opened 2020-10-19
14870 mobile API logout

kreek:
## Description of change
Adds the `/mobile/v0/user/logout` endpoint which revokes the user's IAM SSOe OAuth access token and then clears the IAM user, user identity, and session objects from redis.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#14870

* Adds rails logs for logout success or failure
* Mobile swagger docs updated

testing done: local unit and integration tests
testing planned: staging integration tests


modules/mobile/app/controllers/mobile/v0/users_controller.rb
```diff
@@ -9,6 +9,13 @@ def show
         render json: Mobile::V0::UserSerializer.new(@current_user, options)
       end
 
+      def logout
+        session_manager = IAMSSOeOAuth::SessionManager.new(access_token)
+        session_manager.logout
+
+        head(:ok)
```
jjulian, line 16: 👍 for a nice short response!

modules/mobile/spec/support/iam_session_helper.rb
```diff
@@ -36,6 +36,7 @@ def iam_sign_in(iam_user = FactoryBot.build(:iam_user), access_token = nil)
 
   config.before :each, type: :request do
     Flipper.enable('mobile_api')
+    Flipper.enable('va_online_scheduling')
```
jjulian, line 39: Is this supposed to be part of this PR?

lib/iam_ssoe_oauth/configuration.rb
```diff
@@ -35,6 +35,7 @@ def connection
         base_path, headers: base_request_headers, request: request_options, ssl: ssl_options
       ) do |conn|
         conn.use :breakers
+        conn.use Faraday::Response::RaiseError
```
jjulian, line 38: Does this change impact the introspect call?

lib/iam_ssoe_oauth/session_manager.rb
```diff
@@ -15,8 +15,34 @@ def find_or_create_user
...
 
+    def destroy_user_session
+      uuid = @session.uuid
+
+      identity_destroy_count = IAMUserIdentity.find(uuid).destroy
+      user_destroy_count = IAMUser.find(uuid).destroy
+      session_destroy_count = @session.destroy
+
+      # redis returns number of records successfully deleted
+      if [identity_destroy_count, user_destroy_count, session_destroy_count].sum == 3
```
jjulian, line 33: Is it possible one of these counts gets returned as 2? That could throw off the sum and lead to a false positive.

jjulian (changes requested):
Looks good, a couple detailed questions before we move forward

lib/iam_ssoe_oauth/configuration.rb
kreek: It does, although an inactive response from the introspect call is a 200 and doesn't throw an error (we catch it and return unauthorized). I could add the same 500-> 502 catch there.

lib/iam_ssoe_oauth/session_manager.rb
kreek: technically possible 😄 I had it as `.all? { |c| c == 1 } ` before I'll change it to != 0

modules/mobile/spec/support/iam_session_helper.rb
kreek: no, removed

lib/iam_ssoe_oauth/session_manager.rb
```diff
@@ -15,8 +15,34 @@ def find_or_create_user
...
 
+    def destroy_user_session
+      uuid = @session.uuid
+
+      identity_destroy_count = IAMUserIdentity.find(uuid).destroy
+      user_destroy_count = IAMUser.find(uuid).destroy
+      session_destroy_count = @session.destroy
+
+      # redis returns number of records successfully deleted
+      if [identity_destroy_count, user_destroy_count, session_destroy_count].all? { |c| c != 0 }
```
jjulian, line 33:
```suggestion
      if [identity_destroy_count, user_destroy_count, session_destroy_count].all? { |c| c > 0 }
```
Reads "if all of these are greater than zero" which I think is mildly more readable than "if all of these are not equal to zero"

lib/iam_ssoe_oauth/session_manager.rb
```diff
@@ -15,8 +15,34 @@ def find_or_create_user
...
+        Rails.logger.info('IAMUser log out success', uuid: uuid)
+        true
+      else
+        Rails.logger.warn('IAMUser log out failure', uuid: uuid, status: {
```
jjulian, line 37: Should the api consumer make an effort not to call /logout unless they think a session is in-progress? Looks like the only side-effect is this warning log message.

jjulian (approved):

thilton-oddball (approved):
I think I followed everything on this and it looks good.

PR #5121, opened 2020-10-20
Raise on missing tranlsations in test

annaswims:
## Description of change
In rails test mode, raise if we call a missing translation key


thilton-oddball (approved):
LGTM

PR #5122, opened 2020-10-20
added va_eauth_birlsfilenumber to the sanitizer list as it may contai…

joeniquette:
Added va_eauth_birlsfilenumber to the sanitizer list as it may contain SSNs in plain text in some circumstances.

## Original issue(s)


thilton-oddball (approved):
LGTM

PR #5128, opened 2020-10-20
Update EVSS file upload retry

annaswims:
## Description of change
app/workers/evss/disability_compensation_form/submit_uploads.rb
and app/workers/evss/document_upload.rb call the same EVSS endpoint.  We should retry the jobs in case of EVSS/partner downtime.  

The jobs were retrying fo4 4h or 21 days.  I've updated them both to be ~1d which is the likely max downtime.


thilton-oddball (approved):
LGTM

PR #5132, opened 2020-10-21
Circle failing spec

LindseySaari:
## Description of change
A preneeds attachment swagger spec fails in CircleCI if the mime type isn’t specified “application/pdf”.  In [this example](https://app.circleci.com/pipelines/github/department-of-veterans-affairs/vets-api/7730/workflows/52514dc1-6672-422c-ac06-2f0710661239/jobs/16316), if the application/pdf type is not specified, the code attempts a pdf conversion because [file.content_type is nil](https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/uploaders/preneed_attachment_uploader.rb#L44-L46). (If not_pdf? Is true, then the pdf file conversion proceeds and throws an error because content type is nil) By specifying application/pdf, the correct mime type is passed into the uploader. As seen in the PreneedAttachmentsController [spec](https://github.com/department-of-veterans-affairs/vets-api/blob/master/spec/controllers/v0/preneeds/preneed_attachments_controller_spec.rb#L9), the mime type is specified. 

ImageMagick was also not being installed as a dependency.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#13564

## Testing 
- [x] Tested spec test suit run in Circle


thilton-oddball (approved):
Looks good!

PR #5133, opened 2020-10-21
Olivebranch patch to support uninflected va

erluti:
This PR is a patch for olivebranch to return two VA keys when one has been capitalized by the inflection so that it's backwards compatible with endpoints expecting it as `Va`.

Here's the olivebranch middleware I'm overriding the `call` method of https://github.com/vigetlabs/olive_branch/blob/master/lib/olive_branch/middleware.rb

This PR will merge into the main VA inflection PR.


lib/olive_branch_patch.rb
```diff
@@ -0,0 +1,78 @@
...
+  end
+
+  private
+
+  VA_KEY_VALUE_PAIR_REGEX = /("[^"]+VA[^"]*"):("[^"]*"|\d+|true|false|null)/.freeze
+
+  def duplicate_basic_va_keys!(json)
+    json.gsub!(VA_KEY_VALUE_PAIR_REGEX) do |va_key_value|
+      key, value = va_key_value.split(':')
+      "#{key}:#{value}, #{key.gsub('VA', 'Va')}:#{value}"
```
kfrz, line 31: from @erluti This is subbing the "improper" value to be backwards compatible

kfrz (approved):
This looks great! Walked through with you on the call. Nice work testing!

spec/lib/olive_branch_patch_spec.rb
```diff
@@ -0,0 +1,163 @@
+# frozen_string_literal: true
+
```
LindseySaari, line 2: Really nice test coverage here! 💯

PR #5136, opened 2020-10-22
Remove requirement of PIDs for 0966 ITF endpoint

hinzed1127:
This PR makes 2 changes:

1. `participant_vet_id` is not required for `v1` POA endpoints. This removes documentation related to that.
2. For submissions of type `burial`, it initializes the claimant id to  `@current_user.participant_id` (initialized via MVI). IDs submitted in a request are still prioritized over this fallback value

## Original issue(s)
https://vajira.max.gov/browse/API-2627


modules/claims_api/app/controllers/claims_api/base_form_controller.rb
```diff
@@ -76,8 +76,8 @@ def bgs_service
     def intent_to_file_options
       {
         intent_to_file_type_code: ClaimsApi::IntentToFile::ITF_TYPES[form_type],
-        participant_claimant_id: form_attributes['participant_claimant_id'] || target_veteran.participant_id,
-        participant_vet_id: form_attributes['participant_vet_id'] || target_veteran.participant_id,
+        participant_claimant_id: target_veteran.participant_id,
```
bastosmichael, line 79: This needs to be @current_user instead of target user as the claimant is not the veteran but the person making the request on behalf of the the veteran. I would still include form_attributes before calling the || so that if the consumer includes or have a different PID they can do so.

modules/claims_api/app/swagger/claims_api/v1/form_0966_controller_swagger.rb
```diff
@@ -146,6 +146,7 @@ class Form0966ControllerSwagger
                     end
                     property :participant_claimant_id do
                       key :type, :integer
+                      key :required, :false
```
bastosmichael, line 149: should not be required if we are auto loading claimant ID from @current_user above.

bastosmichael (changes requested):
We should be ensuring that we use the participant id of the Oauth user making the request but not sure if participant_id is buried in the object somewhere.

bastosmichael (approved):

PR #5137, opened 2020-10-22
API-1544: Report All Upstream Status Dependencies in our Healthchecks - Benefits Intake API

SlakrHakr:
## Description of change
Brings current vba_documents healthcheck inline with current healthcheck standards and ensures consumers of this API know the status of its various dependencies as well as this specific service health.

## Original issue(s)
https://vajira.max.gov/browse/API-1544

This does change the functionality of the existing healthcheck endpoint. Previously it would present as down if any of this service's dependencies were down. That endpoint will still present up even if one of the downstream dependencies is down. This change includes a new endpoint, "upstream_healthcheck," which will present as down if any upstream dependencies are down and includes additional information about those dependencies.

Interacted with endpoints locally.
Updated specs to resolve existing tests and include new tests for the new endpoint.


thilton-oddball (approved):
LGTM

PR #5142, opened 2020-10-22
Delegate code ownership to the mobile API team

jjulian:
## Description of change
Delegates code ownership of the /mobile module to the [mobile API team](https://github.com/orgs/department-of-veterans-affairs/teams/mobile-api-team), thus no longer requiring the backend-review-group to approve PRs within this dir. **This will allow our team to move faster when working on non-shared code.** This is a continuation of #5010 which added the mobile api team but did not make them the exclusive codeowners of /mobile.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -12,5 +12,5 @@ spec/support/*/vaos/                 @department-of-veterans-affairs/vfs-vaos @d
 lib/bgs @department-of-veterans-affairs/vfs-ebenefits @department-of-veterans-affairs/backend-review-group
 spec/bgs @department-of-veterans-affairs/vfs-ebenefits @department-of-veterans-affairs/backend-review-group
 
-modules/mobile/                      @department-of-veterans-affairs/mobile-api-team @department-of-veterans-affairs/backend-review-group
-spec/support/*/mobile/               @department-of-veterans-affairs/mobile-api-team @department-of-veterans-affairs/backend-review-group
+modules/mobile/                      @department-of-veterans-affairs/mobile-api-team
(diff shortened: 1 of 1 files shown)


kreek (approved):

erluti:
Maybe I misunderstand, but you already shouldn't need the reviews from `backend-review-group` as is.  With `backend-review-group` as secondary owners we can make changes in these files without your approval, which is sometimes necessary when doing platform-wide changes like updating Rails or Ruby.  The current setup should allow you to work on non-shared code unimpeded.

jjulian:
> Maybe I misunderstand, but you already shouldn't need the reviews from `backend-review-group` as is. 

My understanding of the CODEOWNERS feature: all users or teams on the same line as a subdir will be required to approve a PR when all changes are within that subdir. 

See #5149 for an example. All file changes are within `modules/mobile`, but `backend-review-group` is listed as a codeowner in the PR.

I think the desired configuration is to only have the mobile team on the `modules/mobile` line.

erluti:
> > Maybe I misunderstand, but you already shouldn't need the reviews from `backend-review-group` as is.
> 
> My understanding of the CODEOWNERS feature: all users or teams on the same line as a subdir will be required to approve a PR when all changes are within that subdir.
> 
> See #5149 for an example. All file changes are within `modules/mobile`, but `backend-review-group` is listed as a codeowner in the PR.
> 
> I think the desired configuration is to only have the mobile team on the `modules/mobile` line.

Ah ha!  Turns out that `mobile-api-team` doesn't have write access to `vets-api`, so it doesn't mean anything to be a code owner.  I'll see if I can clear that up

jjulian:
Granting the team Write access has created the desired behavior. Thanks!!

PR #5143, opened 2020-10-22
adding key and cert to hqva_mobile

dillo:
## Description of change

- Added cert and key access to config/settings.yml for the `hqva_mobile` section

## Original issue(s)

https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/14991

When querying mock appointment data, the vets-api is returning an error message because the proper key/cert info is not set under the `hqva_mobile` section in the config/settings.yml file. An example of a failing URI would be `localhost:3000/health_quest/v0/appointments/123123` This PR adds a cert/key to the settings, resolving the current problem.

Mock appointment data will be returned from all environments until the backend API for returning live appointment data is configured.

This will be tested/verified by the FE team when the code is merged back into master.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -555,7 +555,8 @@ va_mobile:
 hqva_mobile:
   url: "https://veteran.apps.va.gov"
   mock: true
-  key_path: /fake/client/key/path
+  cert_path: config/certs/vetsgov-localhost.crt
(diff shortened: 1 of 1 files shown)


thilton-oddball (approved):
LGTM

PR #5146, opened 2020-10-22
Update user model references of mvi to mpi

LindseySaari:
##  Description of change
app/models/user.rb has a method defined that references the mpi RedisStore model (recently updated from Mvi to MPIData) that is called "mvi". This PR updates to make use of the term "mpi".

## Original issue(s)
department-of-veterans-affairs/va.gov-team#14727

## Testing


thilton-oddball (approved):
LGTM

PR #5147, opened 2020-10-22
14870 mobile api logout fix

kreek:
## Description of change
Original logout PR included vets-api making a IAM service revoke call but that will be handled by the mobile app. This PR removes that logic but keeps the code to destroy the IAM user, user identity, and session objects from Redis.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#14870

Testing done: local integration tests
Testing planned: staging integration tests


modules/mobile/spec/support/iam_session_helper.rb
```diff
@@ -36,11 +36,13 @@ def iam_sign_in(iam_user = FactoryBot.build(:iam_user), access_token = nil)
 
   config.before :each, type: :request do
     Flipper.enable('mobile_api')
+    Flipper.enable('va_online_scheduling')
```
kreek, line 39: Since the VAOS specs don't disable 'va_online_scheduling' again because our mobile users have a list of available services our specs will break if the VAOS specs run before ours.

jjulian: Prefer to have specs make no assumptions about the current state. Seems correct to do whatever is needed in setup, like this.

jjulian (approved):

lib/iam_ssoe_oauth/session_manager.rb
```diff
@@ -43,6 +36,8 @@ def destroy_user_session
       end
     end
 
+    private
```
thilton-oddball, line 39: Will making these private have any effect on IAM SSOe session metrics in config/initializers/statsd.rb? It might not, I just saw that it is referenced there and want to make sure.

kreek: That's a good question. Looking at the statsd gem source those config calls use meta programming to add in the statsd calls, which aren't blocked by being private. There's (passing 😄) [specs](https://github.com/department-of-veterans-affairs/vets-api/blob/mobile-api-logout-fix/modules/mobile/spec/controllers/application_controller_spec.rb#L108) that check's if the statsd calls are triggered.

thilton-oddball (approved):
Looks good to go

PR #5148, opened 2020-10-22
13456 mobile direct deposit endpoints

jperk51:
## Description of change
Implementation and testing of mobile payment information endpoints (direct deposit)

## Original issue(s)
department-of-veterans-affairs/va.gov-team#13456

API docs are in a PR (https://github.com/department-of-veterans-affairs/vets-api/pull/5090) once that's merged I will link the docs here.
All testing for these endpoints can be found in `modules/mobile/spec/request/payment_information_request_spec.rb`


modules/mobile/app/controllers/mobile/v0/payment_information_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+
+      def index
+        payment_information = service.get_payment_information.responses[0]
+        render json: Mobile::V0::PaymentInformationSerializer.new(@current_user.id,
+                                                                  payment_information.payment_account,
+                                                                  payment_information.control_information)
+      end
+
+      def update
+        payment_information = service.update_payment_information(pay_info).responses[0]
```
jjulian, line 21: Should we check for success here? Or are failures handled by the controller stack?

jperk51: As far as I can tell all execution errors are caught and are ultimately returned with an appropriate HTTP status code. Additionally, based on the service we are using, I don't there would be any case where we would send the confirmation email without successfully updating.

modules/mobile/app/controllers/mobile/v0/payment_information_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+      def service
+        @service ||= EVSS::PPIU::Service.new(@current_user)
+      end
+
+      def ppiu_params
+        params.permit(
+          :account_type,
+          :financial_institution_name,
+          :account_number,
+          :financial_institution_routing_number
```
jjulian, line 39: It looks like the `filter_parameters` lambda is a whitelist, but still may be good to confirm none of these get into logs.

jjulian (approved):

modules/mobile/app/controllers/mobile/v0/payment_information_controller.rb
jperk51: Did some digging and didn't find anything that looked to be logging those values

kreek (approved):
LGTM 🚀

PR #5151, opened 2020-10-23
va.gov-team#15180: Raise proper error for missing keys and values

benniemosher:
## Description of change

We are seeing errors in [Sentry](http://sentry10.vfs.va.gov/organizations/vsp/issues/3223/?environment=production&project=3&query=is%3Aunresolved+assigned%3A%23vsp-identity&sort=freq&statsPeriod=14d) that we are sending an invalid `find profile` response. We were only checking that the keys were present, and not that the values were nil or an empty string. This PR adds those specs as well as the code to validate this.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#15180

### DIFF
--- a/spec/lib/mvi/messages/find_profile_message_spec.rb
+++ b/spec/lib/mvi/messages/find_profile_message_spec.rb
@@ -136,7 +136,7 @@ describe MVI::Messages::FindProfileMessage do
             given_names: %w[John William],
             last_name: '',
             birth_date: Time.new(1980, 1, 1).utc,
-            ssn: Faker::SSN.ssn
+            ssn: rand.to_s[2..11]
           )
         end.to raise_error(ArgumentError, 'required keys are missing')
       end
@@ -147,7 +147,7 @@ describe MVI::Messages::FindProfileMessage do
             given_names: %w[John William],
             last_name: nil,
             birth_date: Time.new(1980, 1, 1).utc,
-            ssn: Faker::SSN.ssn
+            ssn: rand.to_s[2..11]=
           )
         end.to raise_error(ArgumentError, 'required values are missing')
       end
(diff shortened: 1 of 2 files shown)


spec/lib/mvi/messages/find_profile_message_spec.rb
```diff
@@ -120,14 +120,36 @@
...
+
+      it 'throws an argument error for empty value' do
+        expect do
+          MVI::Messages::FindProfileMessage.new(
+            given_names: %w[John William],
+            last_name: '',
+            birth_date: Time.new(1980, 1, 1).utc,
+            ssn: FFaker::SSN.ssn
+          )
+        end.to raise_error(ArgumentError, 'required keys are missing')
```
AGarcia-Clarity, line 141: Can we update the language here to say 'required values are missing' instead of keys?

benniemosher: I decided to have it echo out a different error message for missing a key and a different one for missing a value so that we can have more specific data to troubleshoot on. I am game either way though.

@joeniquette Thoughts?

AGarcia-Clarity: Good point for more insights into troubleshooting.  I'm fine with either.

jholton (approved):
LGTM.  I was curious how this would manifest for the end user, so I tried to see locally and ended up [with this](https://github.com/department-of-veterans-affairs/va.gov-team/issues/15236).
I can't find a "required key missing" error in Sentry, so either it's never happened (unlikely?) or this error is getting `rescue`d somewhere.
In any case, it'll be worth monitoring Sentry for occurrences of the 'required value missing' error after this gets merged/deployed.

thilton-oddball (approved):
LGTM

f1337 (approved):
Excellent! :metal:
Bill & Ted playing air guitar, saying 'Excellent!'

PR #5154, opened 2020-10-26
chomp trailing slash on NOD contestable issues

charleystran:
## Description of change
This change removes trailing slashes from the caseflow NOD contestable issues endpoint. This serves 2 purposes. 
1. aligns the url calls with the Betamocks configuration
2. Creates consistency with our other paths


templeman15 (approved):
Looks good to me. Seems like you might have an un-related spec failing though?

PR #5158, opened 2020-10-26
Adding Claims API Reporting via Email

bastosmichael:
## Description of change
Create a Claims 526 Submissions Email report like Benefits Intake API

## Original issue(s)
https://vajira.max.gov/browse/API-2847


modules/claims_api/app/mailers/claims_api/unsuccessful_report_mailer.rb
```diff
@@ -0,0 +1,40 @@
...
+  class UnsuccessfulReportMailer < ApplicationMailer
+    RECIPIENTS = %w[
+      michael.bastos@oddball.io
+      ryan.link@oddball.io
+      christopher.stone@libertyits.com
+      valerie.hase@va.gov
+      mark.greenburg@adhocteam.us
+      premal.shah@va.gov
+      lee.deboom@oddball.io
+      dan.hinze@adhocteam.us
```
SlakrHakr, line 13: nit: is there any sort of gmail groups we can use/create that would allow us to add/remove people from these emails without a code change?

modules/claims_api/app/workers/claims_api/report_unsuccessful_submissions.rb
```diff
@@ -0,0 +1,47 @@
...
+
+module ClaimsApi
+  class ReportUnsuccessfulSubmissions
+    include Sidekiq::Worker
+
+    def perform
+      if Settings.claims_api.unsuccessful_report_enabled
+        @to = Time.zone.now
+        @from = @to.monday? ? 7.days.ago : 1.day.ago
+        @consumers = ClaimsApi::AutoEstablishedClaim.where(created_at: @from..@to).pluck(:source).uniq
```
SlakrHakr, line 13: I might be missing something but it looks like @consumers is only used within the total method? Is there a reason its defined up in this scope?

modules/claims_api/app/workers/claims_api/report_unsuccessful_submissions.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    def pending
+      ClaimsApi::AutoEstablishedClaim.where(
+        created_at: @from..@to,
+        status: 'pending'
+      ).order(:source, :status)
+    end
+
+    def totals
+      @consumers.map do |name|
```
SlakrHakr, line 34: nit: this logic won't scale all that well. i dont know how many consumers we have atm, and I know this is in the background, but any value in composing a single query to do this work?

SlakrHakr (approved):
Few questions/nitpicks but nothing crazy.

modules/claims_api/app/mailers/claims_api/unsuccessful_report_mailer.rb
bastosmichael: We created this methodology originally just to ensure that the emails were sent to the right people but deserves a ticket to figure out how to do something similar to google groups within the VA for certain teams and organizations instead etc.

modules/claims_api/app/workers/claims_api/report_unsuccessful_submissions.rb
bastosmichael: I agree but our consumer numbers have remained generally low at the moment, also a good candidate for a technical debt refactor in the future.

modules/claims_api/app/workers/claims_api/report_unsuccessful_submissions.rb
bastosmichael: I think this was more of the pattern used in the original email and I just copied it over for use here to be consistent, worth a refactor in the future for sure.

PR #5161, opened 2020-10-26
Update schema files and directory to mpi

LindseySaari:
## Description of change
MVI::Messages::AddPersonMessage uses schema files in config/mvi_schema. Those schema files and their parent directory should use mpi instead.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#14772


thilton-oddball (approved):
LGTM

PR #5162, opened 2020-10-26
API-2936: autoCestPDFGenerationDisabled is not in 526 POST dev portal docs

SlakrHakr:
## Description of change
Fixes auto-generated curl for 526 POST on https://developer.va.gov/explore/benefits/docs/claims?version=current

## Original issue(s)
https://vajira.max.gov/browse/API-2936

This will require deployment of swagger docs to developer.va.gov 

Tested creating a 526 claim locally.


hinzed1127 (approved):
nice fix on the `serviceBranch` value, too

PR #5165, opened 2020-10-26
Refactoring Error Handling and SentryLogging to support composable Sentry behavior based on exception.en.yml

saneshark:
## Description of change
The goal of this PR is to allow for customizing what is reported to Sentry. It updates some of the logic for error handling to support the use of exception.en.yml for any error derived from Common::Exceptions::BaseError in such a way that you could specify how that error class ought to report to Sentry.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/15012

For one, the logic for SentryLogging was a bit confusing. Why do we want to log as info if va_exceptions exist? There are no comments or rationale to describe this so I am changing this behavior. That being said, there are specs for the Pundit::NotAuthorized exception to always be an info, so I've handled that case separately.

Furthermore, I'm not sure why some of the original exceptions we were preventing from getting logged to Sentry were there in the first place. Why wouldn't a RoutingError or Forbidden or BreakersOutage error be something we want to log? I've left these as is, but instead made them composable via exception.en.yml. The behavior to not report to Sentry remains the same according to the specs.


lib/sentry_logging.rb
```diff
@@ -17,8 +17,11 @@ def log_message_to_sentry(message, level, extra_context = {}, tags_context = {})
   # rubocop:disable Metrics/CyclomaticComplexity
   # rubocop:disable Metrics/PerceivedComplexity
   def log_exception_to_sentry(exception, extra_context = {}, tags_context = {}, level = 'error')
+    level = exception.sentry_type if exception.is_a?(Common::Exceptions::BaseError)
+    # Why is this line below required?!?!? it will override whats specified in exception.en.yml; Why info??
     level = 'info' if extra_context.is_a?(Hash) && client_error?(extra_context[:va_exception_errors])
     level = normalize_level(level)
+    return if level == 'none'
```
saneshark, line 24: this is also going to prevent logging to Rails.logger -- I don't know if we want to do that. :/

spec/controllers/application_controller_spec.rb
```diff
@@ -258,6 +313,8 @@ def append_info_to_payload(payload)
             request_uuid: nil
           )
 
+          expect(Raven).not_to receive(:capture_exception)
+
```
saneshark, line 317: you can't see the lines above this, but basically whoever wrote this test probably had the expectation that capture_exception is getting called, why else would they care that extra_context is getting set otherwise? But it's worth noting that Raven is never going to report to Sentry because of the Exclusions in ExceptionHandling Mixin.

spec/controllers/application_controller_spec.rb
```diff
@@ -53,6 +73,41 @@ def append_info_to_payload(payload)
...
+      with_settings(Settings.sentry, dsn: 'T') do
+        example.run
+      end
+    end
+
+    it 'does not log to sentry if Common::Exceptions::Unauthorized' do
+      expect(Raven).not_to receive(:capture_exception)
+      expect(Raven).not_to receive(:capture_message)
+      get :unauthorized
+    end
```
saneshark, line 87: if it were me I would want this logged as "info" or "warning" -- I don't know why we want to exclude it from Sentry altogether.

spec/controllers/application_controller_spec.rb
```diff
@@ -53,6 +73,41 @@ def append_info_to_payload(payload)
...
+
+    it 'does not log to sentry if Common::Exceptions::RoutingError' do
+      expect(Raven).not_to receive(:capture_exception)
+      expect(Raven).not_to receive(:capture_message)
+      get :routing_error
```
saneshark, line 92: if it were me I would want this logged as "info" or "warning" -- I don't know why we want to exclude it from Sentry altogether.

spec/controllers/application_controller_spec.rb
```diff
@@ -53,6 +73,41 @@ def append_info_to_payload(payload)
...
+    end
+
+    it 'does not log to sentry if Common::Exceptions::Forbidden' do
+      expect(Raven).not_to receive(:capture_exception)
+      expect(Raven).not_to receive(:capture_message)
+      get :forbidden
```
saneshark, line 98: if it were me I would want this logged as "info" or "warning" -- I don't know why we want to exclude it from Sentry altogether.

app/controllers/concerns/exception_handling.rb
```diff
@@ -6,17 +6,11 @@
 module ExceptionHandling
   extend ActiveSupport::Concern
 
-  SKIP_SENTRY_EXCEPTION_TYPES = [
-    Common::Exceptions::Unauthorized,
-    Common::Exceptions::RoutingError,
-    Common::Exceptions::Forbidden,
-    Breakers::OutageException
```
saneshark, line 13: I'm not sure if Breakers::OtageException being excluded was getting tested anywhere. Similar to the comments I made below, I think this should be a warn or info, not excluded altogether.

lib/sentry_logging.rb
f1337: I think we want to continue to log to Rails.logger. Eventually, _everything_ sent to Rails.logger will get logged to Loki. That isn't today's reality, but it is in our Q4 2020 plans. It's possible I don't fully understand the context though, so I'm leaving room for me to be wrong. 😸

spec/controllers/application_controller_spec.rb
f1337: > Raven is never going to report to Sentry

I'm confused by this, but perhaps that's my fault. I thought `Raven` was the interface to Sentry?

lib/sentry_logging.rb
saneshark: I can tweak it so that it just skips Sentry but still does rails logging, but if log level is "none" what should the Rails logger "level" be for semantic logger or Loki?

saneshark: alternately, what I could do instead is to have two or three different log_levels, one for each of: Rails, Loki, Sentry.

saneshark: I don't think that's what we're today for the case of :none as evidenced by specs still passing. I could be wrong.

saneshark:
I think this is ready for a second pass. Will update the title to reflect that.

lib/sentry_logging.rb
```diff
@@ -31,13 +29,18 @@ def log_exception_to_sentry(exception, extra_context = {}, tags_context = {}, le
...
   end
-  # rubocop:enable Metrics/CyclomaticComplexity
-  # rubocop:enable Metrics/PerceivedComplexity
 
-  def normalize_level(level)
+  def normalize_level(level, exception = nil)
     # https://docs.sentry.io/clients/ruby/usage/
     # valid raven levels: debug, info, warning, error, fatal
-    level = level.to_s
+    level = if exception.is_a?(Pundit::NotAuthorizedError)
```
erluti, line 36: case statements can match on classes and i feel like they are easier to read and change:
```
level = case exception
            when Pundit::NotAuthorizedError
              'info'
            when Common::Exceptions::BaseError
              exception.sentry_type.to_s
            else
              level
            end
```

lib/sentry_logging.rb
```diff
@@ -31,13 +29,18 @@ def log_exception_to_sentry(exception, extra_context = {}, tags_context = {}, le
     end
     rails_logger(level, exception.backtrace.join("\n")) unless exception.backtrace.nil?
   end
-  # rubocop:enable Metrics/CyclomaticComplexity
-  # rubocop:enable Metrics/PerceivedComplexity
 
-  def normalize_level(level)
+  def normalize_level(level, exception = nil)
```
erluti, line 33: looks like `normalize_level` is only used in two place.  Maybe just explicitly pass in `nil` on line 7?  Or you could even pass in a symbol like `:no_exception` (maybe assigned to a constant?) and it would fall through the same as `nil`.  The default argument doesn't add a lot in my opinion and the scope of the use is pretty narrow.  I would say it would necessitate extra test cases, but it looks like `lib/sentry_logging.rb` is untested.

spec/controllers/application_controller_spec.rb
```diff
@@ -88,6 +93,15 @@ def append_info_to_payload(payload)
       get :not_authorized
     end
 
+    it 'does log exceptions to sentry based on level identified in exception.en.yml' do
```
erluti, line 96: can there be a test for when the level is `'none'`?

erluti (commented):
I like this a lot.  It's touching that stuff that could benefit from more tests, but I like this approach overall

spec/controllers/application_controller_spec.rb
saneshark: there kind of already is -- for `breakers_outage` and others the level is 'none' -- those specs were merged in a previous PR. They are not failing, therefore the behavior is as it was.

saneshark:
@erluti as a second pass to this, one thing I was thinking we could do is to move some of the logic in the rescue_from block into a method that gets called by the rescue_from block, and then convert any errors that are not BaseError into into a BaseError, and the rescue_from would raise and have it's own rescue block that then returns the BaseError. This sounds complicated, but it would basically ensure that any exceptions that are not handled such as Pundit, Breakers, etc always become some variation of BaseError... I'm going to propose we approach that as part of a separate PR though. It could greatly simplify things.

Some other folks appear to be overriding the of the exception_handling mixin methods, I have since reverted to handling the methods they were using to avoid modifying too many files as part of this PR.

saneshark:
I think I've addressed your comments above, and it's ready for another pass if you want.

spec/controllers/application_controller_spec.rb
saneshark: For reference, the following specs were added before these changes because I didn't have confidence that I would breaking something:

https://github.com/department-of-veterans-affairs/vets-api/pull/5200

erluti: It looks like I misunderstood this test a little.  I thought it was defining an exception to raise,  but now I see it'd describing the exception it expects to see.
I saw those other specs and I'm glad they cover this case.  I've just found tests to be very helpful when there's explicitly one for each behavior.
So of course I'd like more tests, lol, but the one I thought I was suggesting isn't so simple and I withdraw my comment

kfrz (approved):
I'm happy with these changes as-is, and will approve the PR. Happy to talk more about this if you'd like and explore future improvements.

lib/sentry_logging.rb
```diff
@@ -14,11 +14,9 @@ def log_message_to_sentry(message, level, extra_context = {}, tags_context = {})
     end
   end
 
-  # rubocop:disable Metrics/CyclomaticComplexity
-  # rubocop:disable Metrics/PerceivedComplexity
   def log_exception_to_sentry(exception, extra_context = {}, tags_context = {}, level = 'error')
-    level = 'info' if extra_context.is_a?(Hash) && client_error?(extra_context[:va_exception_errors])
```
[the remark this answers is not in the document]
saneshark: spoke with @ericbuckley, i will check and remove if its not used elsewhere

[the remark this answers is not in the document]
saneshark: @kfrz Is that a picture of the Hindenberg not exploding? I mean, 2020 has been like living in an alternate reality. I'll grant you that. ;)

PR #5166, opened 2020-10-26
BGS Find Regional Offices

kathleencrawford:
## Description of change

During integration testing with BGS for the Form 686c (Dependency Application), it was discovered that we need to look up the Location ID of the Regional Office closest to the Veteran and submit that location id as part of the form submission.  This PR adds a new BGS call during the 686 form submission to get and set the regional office location id.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#14974

This is a pretty minor PR that is just adding to an already existing form submission workflow, so no new settings or swagger docs have been added.  The method added does log a warning to sentry, but no major changes.

- [x] Local testing
- [x] Specs added and updated


lib/bgs/service.rb
```diff
@@ -196,6 +196,13 @@ def get_regional_office_by_zip_code(zip_code, country, province, lob, ssn)
       '347' # return default location id
     end
 
+    def find_regional_offices
+      find_regional_offices_response = service.data.find_regional_offices
+      find_regional_offices_response[:return]
```
ericboehs, line 201:
```suggestion
      find_regional_offices_response = service.data.find_regional_offices[:return]
```

Would this work?

ericboehs (approved):
It might be good to test the other code paths for location_id (I see one of the three is tested).

Approved nonetheless. 👍

lib/bgs/service.rb
kathleencrawford: Yes!  Thank you - I've updated as suggested.

kathleencrawford:
> It might be good to test the other code paths for location_id (I see one of the three is tested).
> 
> Approved nonetheless. 👍

@ericboehs I've added 2 new tests as suggested.  Thank you for your review and suggestions!

PR #5167, opened 2020-10-27
Add better validations for upload_form_526

hinzed1127:
## Description of change

Right now `upload_form_526` is running into 500s due to `ClaimsApi::AutoEstablishedClaim.pending?` returning false for queries that actually were pending, resulting in method calls on `false` instead of a valid object.

I also added some options to the 526 establishment and upload sidekiq jobs, which seem to be having problems in at least sandbox. This probably needs tweaking but I'm including it here.

## Original issue(s)
https://vajira.max.gov/secure/RapidBoard.jspa?rapidView=469&view=detail&selectedIssue=API-1979


modules/claims_api/app/workers/claims_api/claim_uploader.rb
```diff
@@ -7,6 +7,8 @@ module ClaimsApi
   class ClaimUploader
     include Sidekiq::Worker
 
+    sidekiq_options 'retry': true, unique_until: :success
```
hinzed1127, line 10: I'm not sure if this is what we want here, but do we want some sort of retry logic if these are intermittently failing (at least on sandbox)?

modules/claims_api/app/workers/claims_api/claim_establisher.rb
```diff
@@ -10,6 +10,8 @@ class ClaimEstablisher
     include Sidekiq::Worker
     include SentryLogging
 
+    sidekiq_options 'retry': true, unique_until: :success
```
hinzed1127, line 13: same question here

modules/claims_api/app/workers/claims_api/claim_uploader.rb
```diff
@@ -7,6 +7,8 @@ module ClaimsApi
...
+
     def perform(uuid)
       document = ClaimsApi::SupportingDocument.find_by(id: uuid) || ClaimsApi::AutoEstablishedClaim.pending?(uuid)
       auto_claim = document.try(:auto_established_claim) || document
```
hinzed1127, line 14: - [ ] check to see if `auto_claim` has an `evss_id`

modules/claims_api/app/controllers/claims_api/application_controller.rb
```diff
@@ -29,13 +24,25 @@ def show
 
     private
 
-    def fetch_or_error_local_claim_id
-      claim = ClaimsApi::AutoEstablishedClaim.find_by(id: params[:id])
+    def find_claim
+      claim = ClaimsApi::AutoEstablishedClaim.find_by(id: params[:id]) # , source: source_name)
```
hinzed1127, line 28: @bastosmichael not sure if you still wanted to try to add in the `source` property of if that can be removed now

modules/claims_api/app/workers/claims_api/claim_uploader.rb
hinzed1127: double checking that this is still something we want to add if we have the have the `evss_id` logic added below. I'm not sure how this retry logic differ from any default states there might already be

modules/claims_api/app/controllers/claims_api/application_controller.rb
bastosmichael: Can you remove it from just the find_by methods for now and I'll create a ticket to address this later as technical debt

modules/claims_api/app/models/claims_api/auto_established_claim.rb
```diff
@@ -49,7 +49,7 @@ def to_internal
 
     def self.pending?(id)
```
bastosmichael, line 50: Since we modified this method, did we capture all of the instances where this gets triggered properly?

modules/claims_api/app/controllers/claims_api/base_disability_compensation_controller.rb
```diff
@@ -16,12 +16,17 @@ class BaseDisabilityCompensationController < ClaimsApi::BaseFormController
 
     def upload_form_526
       pending_claim = ClaimsApi::AutoEstablishedClaim.pending?(params[:id])
-      pending_claim.set_file_data!(documents.first, params[:doc_type])
-      pending_claim.save!
 
-      ClaimsApi::ClaimUploader.perform_async(pending_claim.id)
+      if pending_claim
```
bastosmichael, line 20: One last thing I realized that should have been here but is missing is we need to check for if 

form_attributes['autoCestPDFGenerationDisabled'] == true

This is to ensure that no one ever uploads a 526 claim if one is being already generated by EVSS, was in the original spec for this end point but we must have missed it somehow. The error should probably say something like Claim not found or upload disabled due to PDF generation etc.

bastosmichael (approved):
Need to add the form_attributes['autoCestPDFGenerationDisabled'] == true check and then i think you're good to merge.

modules/claims_api/app/models/claims_api/auto_established_claim.rb
hinzed1127: yeah, the base controller is the only spot it presently is called 👍

PR #5170, opened 2020-10-27
Update rakefile to use MPIData

LindseySaari:
##  Description of change
This updates an mvi rakefile task to use the new MPIData model instead of the former.


thilton-oddball (approved):
LGTM

PR #5179, opened 2020-10-27
Removes Sentry tracing from SAML POST binding form

patrickvinograd:
Sentry bundle is causing JS execution issues on IE in compatibility
mode. This removes the Sentry tracing (which is not providing any useful
hits) while we investigate further.


thilton-oddball (approved):
LGTM

CoryTrimmUSDS (approved):
LGTM. 🌮

PR #5182, opened 2020-10-28
Api 3071-- Report All Upstream Status Dependencies in our Healthchecks - Forms API

cshupp1:
## Original issue(s)
https://vajira.max.gov/browse/API-3071

No to the above

Unit tests can be run via:
 rspec ./modules/va_forms/spec/requests/metadata_request_spec.rb


modules/va_forms/config/routes.rb
```diff
@@ -4,6 +4,8 @@
   match '/metadata', to: 'metadata#index', via: [:get]
   match '/v0/*path', to: 'application#cors_preflight', via: [:options]
   match '/v0/healthcheck', to: 'metadata#healthcheck', via: [:get]
+  match '/v0/upstream_healthcheck', to: 'metadata#upstream_healthcheck', via: [:get]
+  match '/v1/upstream_healthcheck', to: 'metadata#upstream_healthcheck', via: [:get]
```
SlakrHakr, line 8: Looks like there isn't a v1 for va_forms yet.

modules/va_forms/spec/requests/metadata_request_spec.rb
```diff
@@ -10,19 +11,95 @@
...
+      context 'v1' do
+        path = '/services/va_forms/v1/upstream_healthcheck'
+        it 'returns correct response and status when healthy' do
+          healthy_checks(path)
+        end
+
+        it 'returns correct status when central_mail is not healthy' do
+          unhealthy_checks(path)
+        end
       end
```
SlakrHakr, line 103: This shouldn't be needed after removing the v1 line from the routes file

modules/va_forms/config/routes.rb
cshupp1: Thank you, I was wondering about that.

SlakrHakr (approved):

PR #5119, opened 2020-10-20
[WIP/DRAFT LARGE NON-MERGABLE HOST PR] [686/674 split]

digitaldrk:
## Original issue(s)
department-of-veterans-affairs/va.gov-team#0000

PR #5186, opened 2020-10-28
Form [674/686 split] job and lib

digitaldrk:
## Description of change
This PR is part of a larger PR I'm splitting up #5119 . Upon more testing with our integration partner BGS, we discovered this form is sort of a 2 in one deal. We need to send different data for the 674 than the 686.

This form is still being tested and is not available in production

## Original issue(s)
department-of-veterans-affairs/va.gov-team#15151

No settings changes
No feature flag was added in this PR. This feature is not publicly available though
No added sentry alerts. This feature already has Sentry alerts
No Prometheus metrics collected
No updates to Swagger
No concerns or questions


app/workers/bgs/submit_form674_job.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    # we do individual service retries in lib/bgs/service.rb
+    sidekiq_options retry: false
+
+    def perform(user_uuid, saved_claim_id, vet_info)
+      user = User.find(user_uuid)
+      claim_data = valid_claim_data(saved_claim_id, vet_info)
+
+      BGS::Form674.new(user).submit(claim_data)
+    rescue
```
ericboehs, line 19: Should we eat all errors here?

app/workers/bgs/submit_form674_job.rb
```diff
@@ -0,0 +1,33 @@
...
+      DependentsApplicationFailureMailer.build(user).deliver_now if user.present?
```
ericboehs, line 20: Would it be better to deliver_later here in case the DependentsApplicationFailureMailer job fails? Or does sidekiq retry in that case?

ericboehs (approved):
A couple questions but looks good.

app/workers/bgs/submit_form674_job.rb
digitaldrk: I think so. We shoot off a lot of requests (like 20) to our integration partner BGS from this job. I can't get a sense of consistency from BGS services in terms of exceptions. We handle retry logic further down stream given some nuances between services. If an error bubbles up to here, failure was catastrophic.

app/workers/bgs/submit_form674_job.rb
digitaldrk: We can do `deliver_later`. Retries in this job are shut off. We have limited opportunities to submit this form in a time frame, we don't want to retry sending with possible bad data.

PR #5187, opened 2020-10-29
[15526] Add filenumber/ssn to 5655 form prefill

camerontesterman:
## Description of change
- Adds last four digits of fileNumber and SSN to the form prefill

## Associated Schema Change PR
https://github.com/department-of-veterans-affairs/vets-json-schema/pull/540

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/15526

Tested locally

### DIFF
--- a/app/models/form_profiles/va5655.rb
+++ b/app/models/form_profiles/va5655.rb
@@ -8,4 +8,13 @@ class FormProfiles::VA5655 < FormProfile
       returnUrl: '/veteran-information'
     }
   end
+
+  private
+
+  def va_file_number_last_four
+    (
+      BGS::PeopleService.new(user).find_person_by_participant_id[:file_nbr].presence ||
+      user.ssn.presence
+    )&.last(4)
+  end
 end
--- a/config/form_profile_mappings/5655.yml
+++ b/config/form_profile_mappings/5655.yml
@@ -1,3 +1,6 @@
+personalIdentification:
+  sSN: [identity_information, ssn_last_four]
+  fileNumber: [va_file_number_last_four]
 personalData:
   fullName: [identity_information, full_name]
   address: [contact_information, address]
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -57,10 +57,10 @@ GIT
 
 GIT
   remote: https://github.com/department-of-veterans-affairs/vets-json-schema
-  revision: 009e0f2249c6400296713ddb0d0339317517f725
+  revision: 1a499b7154f60c7d1acc86d34e395b5db6deed82
   branch: master
   specs:
-    vets_json_schema (17.3.0)
+    vets_json_schema (17.3.1)
       multi_json (~> 1.0)
       script_utils (= 0.0.4)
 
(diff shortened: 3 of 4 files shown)


app/models/form_profiles/va5655.rb
```diff
@@ -8,4 +8,13 @@ def metadata
       returnUrl: '/veteran-information'
     }
   end
+
+  private
+
+  def va_file_number_last_four
+    (
+      BGS::PeopleService.new(user).find_person_by_participant_id[:file_nbr].presence ||
+      user.ssn.presence
```
ericboehs, line 17: What do you think of indenting this line?

config/form_profile_mappings/5655.yml
```diff
@@ -1,3 +1,6 @@
+personalIdentification:
+  sSN: [identity_information, ssn_last_four]
```
ericboehs, line 2: Should this be all caps? or camelCased? This feels weird. 😕

ericboehs (commented):
Should we wait until the vets-json-schema change is merged so we can update the Gemfile with the SHA bump?

If so, it'd be nice to mark this as draft and add a todo to the description.

config/form_profile_mappings/5655.yml
camerontesterman: It is kind of weird, but if you are referring to the keys being camelCased and values within the array represented as snake_case that is an established pattern within other form profile mappings.

app/models/form_profiles/va5655.rb
camerontesterman: Can you send me a snippet of how you think it should be indented?

app/models/form_profiles/va5655.rb
```diff
@@ -8,4 +8,13 @@ def metadata
...
+    )&.last(4)
```
ericboehs, line 18:
```suggestion
    (
      BGS::PeopleService.new(user).find_person_by_participant_id[:file_nbr].presence ||
        user.ssn.presence
    )&.last(4)
```

config/form_profile_mappings/5655.yml
ericboehs: Oh. So it's coming in as `s_s_n`? 😲

app/models/form_profiles/va5655.rb
camerontesterman: I've implemented your suggestion.

config/form_profile_mappings/5655.yml
camerontesterman: so when the frontend hits the in_progress_form endpoint (in this case it would be a GET /v0/in_progress_forms/5655) the keys here will be rendered to the frontend in camelCase as they are intended to.

These could be changed, but I'd rather stick as close as possible to what the DMC is expecting once we get to the POST.

The values within the arrays reference Virtus models and associated methods (i.e. FormIdentityInformation class and the ssn_last_four method) that method gets called to populate the `sSN` key which is what the DMC is expecting.

I could represent this as `ssn: [identity_information, ssn_last_four]` but then on the POST either the frontend would need to alter the key to sSN or the backend would prior to POSTing to the DMC.

I would like to get them to change it, but for right now it should stay as it is in my opinion.

ericboehs (approved):

PR #5190, opened 2020-10-29
14991 mock appt from api

dillo:
## Original issue(s)

- Feature flag is `show_healthcare_experience_questionnaire`.
- Swagger docs are the same.
- Removed appointments index action and associated request specs since we're not required to fetch all appointments for a veteran from Vista.
- Until the MAP team completes work in Staging and Production environments, the health_quest module will depend upon the playground for appointment/sessions/questionnaire data. It will be making all requests to `https://vagovtest.mapsandbox.net` as reflected by the `config/settings.yml` file. 
- When hitting the `show` action in the appointments controller, we're returning mock data since VISTA is not present in the MAP playground environment. Once necessary work is done in MAP staging/production, health_quest will switch its configuration over to those environments.
- `user_service_spec.rb` is now using a separate set of VCR cassettes which point to the playground MAP. As the work in MAP staging and production environments are completed, the VCR cassettes will be pointed back to the veteran.apps.va.gov domain.

- Signed into localhost:3001 and then visited http://localhost:3000/health_quest/v0/appointments/123123 in a browser window. The response body is the appropriate JSON structure for the mocked out appointment.


modules/health_quest/spec/serializers/v0/va_appointments_serializer_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+    end
+  end
+
+  describe 'KEY #data' do
+    it 'has data key' do
+      expect(subject.serializable_hash.key?(:data)).to be(true)
+    end
+
+    it 'data has attributes' do
+      expect(subject.serializable_hash[:data][:attributes].keys).to eq(%i[
```
ericboehs, line 46: Nit: It would look better if you started the `%i[]` on a new line. You wouldn't have to indent all elements 20 times.

dillo: Sure. Rubocop is still breaking it down into two lines but I think it looks better than the original version.

dillo:
@ericboehs closing this since I'm breaking the PR into two or more separate pull requests.

PR #5193, opened 2020-10-30
API-3070: Report All Upstream Status Dependencies in our Healthchecks - Claims API

SlakrHakr:
## Description of change
Brings current claims healthcheck inline with current healthcheck standards and ensures consumers of this API know the status of its various dependencies as well as this specific service health.

## Original issue(s)
https://vajira.max.gov/browse/API-3070

This does change the functionality of the existing healthcheck endpoint. Previously it would present as down if any of this service's dependencies were down. That endpoint will still present up even if one of the downstream dependencies is down. This change includes a new endpoint, "upstream_healthcheck," which will present as down if any upstream dependencies are down and includes additional information about those dependencies.

Interacted with endpoints locally.
Updated specs to resolve existing tests and include new tests for the new endpoint.

### DIFF
--- a/modules/appeals_api/spec/requests/metadata_request_spec.rb
+++ b/modules/appeals_api/spec/requests/metadata_request_spec.rb
@@ -100,54 +100,54 @@ RSpec.describe 'Appeals Metadata Endpoint', type: :request do
           expect(upstream_service['details']['time']).to eq('2020-09-21T00:00:00Z')
         end
       end
+    end
 
-      context 'v1' do
-        it 'returns correct response and status when healthy' do
-          VCR.use_cassette('caseflow/health-check') do
-            get '/services/appeals/v1/upstream_healthcheck'
-            expect(response).to have_http_status(:ok)
+    context 'v1' do
+      it 'returns correct response and status when healthy' do
+        VCR.use_cassette('caseflow/health-check') do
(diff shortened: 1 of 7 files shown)


modules/vba_documents/spec/request/metadata_request_spec.rb
```diff
@@ -92,54 +92,54 @@
         expect(upstream_service['details']['status']).to eq('Unavailable')
         expect(upstream_service['details']['time']).to eq('2020-09-21T00:00:00Z')
       end
+    end
+
+    context 'v1' do
```
SlakrHakr, line 97: this just shifts v1 context out next to v0, instead of inside it for some strange reason

modules/appeals_api/spec/requests/metadata_request_spec.rb
```diff
@@ -100,54 +100,54 @@
           expect(upstream_service['details']['time']).to eq('2020-09-21T00:00:00Z')
         end
       end
+    end
+
+    context 'v1' do
```
SlakrHakr, line 105: this just shifts v1 context out next to v0, instead of inside it for some strange reason

bastosmichael (approved):
LGTM but test the end points before closing the ticket.

PR #5197, opened 2020-10-30
[API 3200] PDF Header Missing Consumer Name

templeman15:
## Description of change
This PR fixes a bug where consumer user name was not interpolated in the stamp on the HLR generated PDF. 

## Original issue(s)
[PDF Header Missing Consumer Name](https://vajira.max.gov/browse/API-3200)

## Discussion 

The headers method inside of `HigherLevelReviewsController` stripped `X-Consumer-Username` from the headers stored in the HLR due to `X-Consumer-Username` being missing from `200996_headers.json`. 

```
 def headers
    HEADERS.reduce({}) do |acc, header_key|
      header_value = request.headers[header_key]

      header_value.nil? ? acc : acc.merge({ header_key => header_value })
    end
  end
```

I added `X-Consumer-Username` to `200996._headers.json` which should allow `X-Consumer-Username` to persist into the HLR model and be retrieved when stamping the PDF.


ElleeB (approved):
LGTM

PR #5198, opened 2020-11-02
Carma betamocks

lihanli:
## Description of change
add carma betamocks

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/9631

tested locally


lib/carma/client/configuration.rb
```diff
@@ -8,6 +8,10 @@ class Configuration < Salesforce::Configuration
       def service_name
         'CARMA'
       end
+
+      def mock_enabled?
```
kevinmirc, line 12: I think since there is a new method being added, we should have a simple unit test for it.

lib/carma/client/client.rb
```diff
@@ -40,7 +40,14 @@ def upload_attachments(payload)
       private
 
       def restforce
-        @client ||= get_client
+        return @client if @client.present?
+
+        @client = get_client
+        if Settings['salesforce-carma'].mock
```
kevinmirc, line 46: There is conditional logic here that I noticed is not being tested.

kevinmirc (changes requested):

spec/lib/carma/client/client_spec.rb
```diff
@@ -26,22 +28,36 @@
...
+      end
+    end
+
+    context 'with betamocks enabled' do
+      test_carma_submission
+    end
+
+    context 'with betamocks disabled' do
+      before do
+        Settings['salesforce-carma'].mock = false
```
kevinmirc, line 57: I think overriding this global variable could potentially result in unexpected side effects or confusion to a developer.

I'd mock this instead of actually overriding it:
```ruby
expect(Settings['salesforce-carma']).to receive(:mock).and_return(false)
```

spec/lib/carma/client/client_spec.rb
```diff
@@ -4,6 +4,19 @@
 require 'carma/client/client'
 
 RSpec.describe CARMA::Client::Client, type: :model do
+  let(:restforce_client) do
+    restforce_client = double
+    expect(subject).to receive(:get_client).and_return(restforce_client)
+
+    if Settings['salesforce-carma'].mock
```
kevinmirc, line 11: I think tucking away this conditional statement in this variable definition, that is used for multiple tests, makes it harder to understand what's happening, and when. It also takes away from the ability to read down the page linearly to understand what mocks cause which expectations.

Since these expectations happen **as a result** of `Settings['salesforce-carma'].mock`, I'd keep them close to where that variable is being mocked. Keeping related context close together makes it much easier to read linearly and understand that there is a causal relationship between that mock and these expectations.

```rb
context 'with betamocks enabled' do
  before do
    expect(Settings['salesforce-carma']).to receive(:mock).and_return(true)

    expect(restforce_client).to receive(:builder).and_return(builder)
    expect(builder).to receive(:insert_before).with(Faraday::Adapter::NetHttp, Betamocks::Middleware)
  end

  it "does the thing"
    # ... (test with it enabled)
  end
end

context 'with betamocks disabled' do
  before do
    expect(Settings['salesforce-carma']).to receive(:mock).and_return(false)
    expect(restforce_client).not_to receive(:builder)
  end

  it "does the thing"
    # ... (test with it disabled)
  end
end
```

spec/lib/carma/client/client_spec.rb
```diff
@@ -15,32 +28,44 @@
...
+      end
+    end
+
+    context 'with betamocks enabled' do
+      test_carma_submission
+    end
+
+    context 'with betamocks disabled' do
+      before do
+        Settings['salesforce-carma'].mock = false
```
kevinmirc, line 57: I think overriding this global variable could potentially result in unexpected side effects or confusion to a developer.

I'd mock this instead of actually overriding it:
```ruby
expect(Settings['salesforce-carma']).to receive(:mock).and_return(false)
```

kevinmirc (approved):
Made some suggestions for improving the readability of the test suite. Other than that, everything else LGTM. Thanks!

kevinmirc (approved):
Looks good. Did we want to test `#upload_attachments` with **betamocks enabled** or save that for another PR? Not sure how many mocks we have at this point.

lihanli:
ok i added a spec for testing `upload_attachments` with betamocks enabled

ericboehs (approved):

PR #5199, opened 2020-11-02
13351 mobile claims docs

jperk51:
## Description of change
API docs for mobile claims and appeals overview endpoint

## Original issue(s)
department-of-veterans-affairs/va.gov-team#13351
(This issue encompasses all the claims and appeals endpoints for mobile while this PR only covers one)

Docs only PR


modules/mobile/docs/openapi.yaml
```diff
@@ -1361,7 +1517,7 @@ components:
               example: VA
             status:
               type: string
-              enum: [BOOKED, CANCELLED]
+              enum: [ BOOKED, CANCELLED ]
```
kreek, line 1520: looks like our IDEs have different settings for array formatting

kreek (approved):
LGTM 🚀

modules/mobile/docs/openapi.yaml
jperk51: should be fixed now

PR #5201, opened 2020-11-02
10-10CG: Certification Checkboxes on PDF Fill

kevinmirc:
Relates to (Prerequisite of): https://github.com/department-of-veterans-affairs/va.gov-team/issues/14702

## Description of change
The UI will begin to send a list of certifications for each form subject (terms that the subject agreed to and attested were true when signing). These need to be read by the PDF Generation in order to check the certification checkbox on the 10-10CG PDF Attachment.

These changes conditionally set a property on the Fill data which, in a future PR, will checkmark the PDF box it if meets the proper conditions.


kevinmirc:
This work has been scrapped since the new version of the PDF will not allow applicants to choose their certifications, they all accept all terms.

Common work was moved to this isolated PR: https://github.com/department-of-veterans-affairs/vets-api/pull/5232

PR #5202, opened 2020-11-02
14991 mock appointment

dillo:
## Description of change

- While the questionnaire feature and the health_quest codebase is in the master branch, please note that this feature is not yet turned "on" in the production environment
- The feature will be turned on when the code is production-ready and we're no longer testing with mocks.
- When hitting the show action in the appointments controller in the health_quest module, we're returning mock data via the appointment_service since VISTA is not available in the MAP playground environment and it's not yet fully functional in MAP staging. **Once necessary work is done in MAP playground/staging environments by the teams working on those products(appt/pgd etc), the health_quest backend team will write code that will get appointment data from those services in their respective environments**.
- Mocking the data directly inside the health_quest module is simple, easy to modify, and consume by FE teams. Mocking it via betamocks/vets-api-mockdata means mocking other service calls(user/session) and is beyond the scope of what we're trying to do.
- The data for retrieving mock appointment will be removed as services for getting appointments in MAP staging becomes available. Until this happens, we need to return mock appointment data so that the FE team is not blocked in their UI endeavors.

## Original issue(s)

Verified by visiting: http://localhost:3000/health_quest/v0/appointments/123123


modules/health_quest/app/services/health_quest/appointment_service.rb
```diff
@@ -17,11 +17,13 @@ def get_appointments(start_date, end_date, pagination_params = {})
       end
     end
 
-    def get_appointment_by_id(id)
+    def get_appointment_by_id(_id)
```
ericboehs, line 20: Why are we no longer looking up appointments by id? It would seem this method by its name should do that.

modules/health_quest/app/services/health_quest/appointment_service.rb
```diff
@@ -17,11 +17,13 @@ def get_appointments(start_date, end_date, pagination_params = {})
...
       with_monitoring do
-        response = perform(:get, "#{get_appointments_base_url}/#{id}", {}, headers, timeout: 55)
+        response =
+          YAML.load_file(Rails.root.join(*appt_file)).with_indifferent_access
```
ericboehs, line 23: Why are we no longer pulling data from an endpoint and instead pulling from a local yml?

I wouldn't think we would want to do this in production, even if the code is behind a feature flag.

modules/health_quest/app/services/health_quest/appointment_service.rb
```diff
@@ -71,5 +73,9 @@ def other_params(use_cache = false)
     def date_format(date)
       date.strftime('%Y-%m-%dT%TZ')
     end
+
+    def appt_file
```
ericboehs, line 77: I think ideally we should move away from abbreviations for single words. Thoughts?

modules/health_quest/config/mock/appointment.yml
```diff
@@ -0,0 +1,57 @@
+---
```
ericboehs, line 1: How did you generate this file? I'm not familiar with the schema.

modules/health_quest/config/rsa/sandbox_rsa
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
ericboehs, line 1: Should we be keeping these keys external to the repo and pulling them in during deploy from credstash or similar sensitive storage system?

modules/health_quest/spec/services/appointment_service_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      expect(appointment[:data].respond_to?(:clinic_friendly_name)).to be(true)
+      expect(appointment[:data].respond_to?(:facility_id)).to be(true)
+      expect(appointment[:data].respond_to?(:sta6aid)).to be(true)
+      expect(appointment[:data].respond_to?(:station_name)).to be(true)
+      expect(appointment[:data].respond_to?(:patient_icn)).to be(true)
+      expect(appointment[:data].respond_to?(:community_care)).to be(true)
+      expect(appointment[:data].respond_to?(:vds_appointments)).to be(true)
+    end
+
+    it 'has pagination information' do
```
ericboehs, line 39: These values seem odd. Shouldn't they all be positive?

modules/health_quest/spec/services/appointment_service_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    it 'is a Hash' do
+      expect(appointment).to be_a(Hash)
+    end
+
+    it 'has a data key of type OpenStruct' do
+      expect(appointment[:data]).to be_a(OpenStruct)
+    end
+
+    it 'data has relevant keys' do
+      expect(appointment[:data].respond_to?(:id)).to be(true)
```
ericboehs, line 27:
```suggestion
      expect(appointment[:data].key?(:id)).to be(true)
```

The test name suggests `appointment[:data]` is a hash rather than an object.

Would `.key?` be more appropriate here? Or should the test be renamed?

modules/health_quest/app/services/health_quest/appointment_service.rb
dillo: I'm open to suggestions. Since we're not using the id argument inside the method body anywhere, rubocop/lint failed. Prefixing it with the underscore was a way to get those checks to pass.

modules/health_quest/app/services/health_quest/appointment_service.rb
dillo: Good question. The endpoint to pull data from local/dev/playground does not exist since VISTA is not available in those environments. The MAS service in staging will support retrieving appointment data in the future but doesn't do so currently. The team responsible for the MAS service is currently working on letting vets-api hit it's appointment endpoints. We're hoping that those endpoints will be available in the next few sprints.

Until this endpoint to communicate with the service is available I'm mocking the data locally in the vets-api. I'm open to suggestions.

modules/health_quest/app/services/health_quest/appointment_service.rb
dillo: Agreed! I will change it to appointment_file.

modules/health_quest/config/mock/appointment.yml
dillo: Great question! The file was composed after speaking with MAP and FE developers. The structure of this appointment was molded after `/vets-api-mockdata/ihub/appointments/default.yml` from the vets-api-mockdata.

modules/health_quest/config/rsa/sandbox_rsa
dillo: Sounds like the safe approach. What is credstash and how do I use it?(I'm new)

modules/health_quest/spec/services/appointment_service_spec.rb
dillo: Good catch. I think I will rename the test. `appointment[:data]` returns an OpenStruct object so sending it `respond_to?` seems a better fit.

modules/health_quest/app/services/health_quest/appointment_service.rb
dillo: Done!

modules/health_quest/config/mock/appointment.yml
dillo: Another source that we used to build the appointment file: `vets-api/spec/support/vcr_cassettes/vaos/appointment_requests/get_requests.yml`

modules/health_quest/spec/services/appointment_service_spec.rb
dillo: In my opinion, I don't think a single appointment should have pagination information. I will confer with the FE team and then either change the values of the pagination section to `1` or just remove that data entirely.

dillo: After speaking with the FE team I've gone ahead and removed the meta section.

modules/health_quest/config/rsa/sandbox_rsa
ericboehs: Reach out to the ops support engineer. He'll be able to guide you.

You'll probably want to rotate that key since it's stored in git now. You can send the key to the engineer over Keybase and they can add it to the devops repo which will drop the file into place upon deploy.

ericboehs:
Everything besides the key looks fine to me. Let me know when you get that resolved and I can +1.

modules/health_quest/config/rsa/sandbox_rsa
annaswims: This is a public repo, so this key is no longer private.

config/settings.yml
```diff
@@ -559,9 +559,9 @@ va_mobile:
 
 hqva_mobile:
   url: "https://veteran.apps.va.gov"
-  mock: true
-  cert_path: config/certs/vetsgov-localhost.crt
-  key_path: config/certs/vetsgov-localhost.key
+  mock: false
+  key_path: /srv/vets-api/secret/health-quest.key
+  local_key_path: modules/health_quest/config/rsa/sandbox_rsa
```
dillo, line 564: @ericboehs All environments other than dev environments(local or anything which returns true for Rails.env.development?) will use the key pair that's in the repository. All other environments including, staging and production will use what Ops provided us: `/srv/vets-api/secret/health-quest.key`

modules/health_quest/app/services/health_quest/configuration.rb
```diff
@@ -14,7 +14,9 @@ def service_name
     end
 
     def rsa_key
-      @key ||= OpenSSL::PKey::RSA.new(File.read(Settings.hqva_mobile.key_path))
+      private_key_path = Rails.env.development? ? Settings.hqva_mobile.local_key_path : Settings.hqva_mobile.key_path
```
dillo, line 17: @ericboehs see: https://github.com/department-of-veterans-affairs/vets-api/pull/5202/files#r517545019

config/settings.yml
ericboehs: Maybe name it `development_key_path`?

modules/health_quest/app/services/health_quest/appointment_service.rb
ericboehs: I think it's fine if we're goin to use it in the future (e.g. rename it back to `id`).

ericboehs (approved):
Small suggestions. 👍

PR #5204, opened 2020-11-02
Updating admin user email for access to flipper feature toggles on staging

anitadewitt-va:
## Description of change

Updating admin user email for access to flipper feature toggles on staging
## Original issue(s)
department-of-veterans-affairs/orchid#139


ericboehs (approved):
Assuming his email doesn't have an "er" on the end, this looks good.

PR #5208, opened 2020-11-03
Add task to populate application directory

bradenshipley:
## Description of change
One worker file and single rake task added to apps_api module to populate directory_applications table if the data is not already present. Scheduled job added to sidekiq-scheduler to run 1m after startup to invoke our worker file. 

## Original issue(s)
https://vajira.max.gov/secure/RapidBoard.jspa?rapidView=777&view=detail&selectedIssue=API-3359

The rubocop error for BlockLength has been disabled for the update_directory_applications.rake file. I can't think of a way to keep avoid the error while keeping the code quality readable. SimpleCov has also been set to ignore this file, since there is no real way to test the `do` blocks of each `find_or_create_by!` call for each attribute set without becoming spaghetti.

Manual testing has been done locally. The task populates the table and development logs show the sidekiq job being invoked. Tests for the worker and the rake file are included.


bradenshipley:
I see that the build is failing due to coverage on the rake file. Give me a bit and I'll push a commit with test for that.

PR #5214, opened 2020-11-03
EDU: Add logging to spool file processing #14967

zurbergram:
## Description of change

- Changing prometheus stat `worker_education_benefits_claim_failed_formatting` to be `worker_education_benefits_claim_failed_formatting_<region_id>_22_<type>`
-  Adding a `rescue` to code that writes data to the spool file for each region, other files will still be created

```
        rescue => error
          exception = SpoolWriteError.new("Error creating #{filename}.\n\n#{error}")
          log_exception_to_sentry(exception)
``` 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#14967

### DIFF
--- a/app/workers/education_form/create_daily_spool_files.rb
+++ b/app/workers/education_form/create_daily_spool_files.rb
@@ -10,6 +10,9 @@ module EducationForm
   class FormattingError < StandardError
   end
 
+  class DailySpoolFileLogging < StandardError
+  end
+
   class DailySpoolFileError < StandardError
   end
 
@@ -141,7 +144,6 @@ module EducationForm
     # Useful for debugging which records were or were not sent over successfully,
     # in case of network failures.
     def log_submissions(records, filename)
-      binding.pry
       ids = records.map { |r| r.record.id }
       log_info("Writing #{records.count} application(s) to #{filename}")
       log_info("IDs: #{ids}")
@@ -165,7 +167,7 @@ module EducationForm
 
     def log_info(message)
       if Flipper.enabled?(:sidekiq_create_daily_spool_file_logging)
-        log_exception_to_sentry(message)
+        log_exception_to_sentry(DailySpoolFileLogging.new(message))
       else
         logger.info(message)
       end
--- a/app/workers/education_form/create_daily_spool_files.rb
+++ b/app/workers/education_form/create_daily_spool_files.rb
@@ -10,6 +10,9 @@ module EducationForm
   class FormattingError < StandardError
   end
 
+  class DailySpoolFileError < StandardError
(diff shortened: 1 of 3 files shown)


shawkey-daniel (approved):
LGTM

delli-gatti-michael (approved):

app/workers/education_form/create_daily_spool_files.rb
```diff
@@ -147,5 +163,13 @@ def track_form_type(type, rpo)
     def stats
       @stats ||= Hash.new(Hash.new(0))
     end
+
+    def log_info(message)
+      if Flipper.enabled?(:sidekiq_create_daily_spool_file_logging)
+        log_exception_to_sentry(DailySpoolFileLogging.new(message))
+      else
+        logger.info(message)
+      end
```
ericboehs, line 172: It looks like you're using our feature flipper as a configuration tool rather than for a temporary feature that we would work to remove.

Putting that aside, I don't see a reason for the conditional at all. Couldn't we just call both?

```suggestion
      log_exception_to_sentry(DailySpoolFileLogging.new(message))
      logger.info(message)
```

What are your thoughts on doing this?

app/workers/education_form/create_daily_spool_files.rb
```diff
@@ -147,5 +163,13 @@ def track_form_type(type, rpo)
     def stats
       @stats ||= Hash.new(Hash.new(0))
     end
+
+    def log_info(message)
+      if Flipper.enabled?(:sidekiq_create_daily_spool_file_logging)
+        log_exception_to_sentry(DailySpoolFileLogging.new(message))
```
annaswims, line 169: The parameters for `log_exception_to_sentry` are `(exception, extra_context = {}, tags_context = {}, level = 'error')`. and it'll log the message to the rails log.

app/workers/education_form/create_daily_spool_files.rb
ericboehs: Also, it looks like `log_exception_to_sentry` [already logs to Rails' logger](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/sentry_logging.rb#L27-L31).

zurbergram: Removed feature flag and set log level to `:info` for this method

app/workers/education_form/create_daily_spool_files.rb
zurbergram: Removed feature flag and set log level to `:info` for this method

ericboehs (approved):

PR #5216, opened 2020-11-03
All emis services are converted to soap 1.2

lukemarkwordtlibertyits:
EMIS uses Soap 1.2. Current integration is set to use 1.1.  Upgraded to use 1.2.


ericboehs (changes requested):
Please revert the changes to the Gemfile.lock.

ericboehs (approved):

PR #5217, opened 2020-11-03
fix - 'Air Force National Guard' -> 'Air National Guard'

annaswims:
## Description of change
We are building a string to get the service branch name from emis, but the proper name is 'Air National Guard' not  'Air Force National Guard'.  

Also, add validation that the service branch is one that the JSON schema agrees to. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#15654

After this PR is merged we'll want to fix the data in prod
````ruby
results = []
in_progress_forms = InProgressForm.where(form_id: '21-526EZ').find_each do |f|
  if f.form_data.include?('"service_branch":"Air Force National Guard"')
    f.form_data = f.gsub('"service_branch":"Air Force National Guard"','"service_branch":"Air National Guard"',)
    f.save!
    results << [f.id, JSON.parse(f.form_data).dig("phone_and_email","email_address"), f.created_at, f.updated_at, f.metadata["return_url"]]
  end
end

puts results
````

### CODE REVIEW

File: app/models/emis_redis/military_information.rb [hunk 1]
```diff
@@ -133,16 +142,7 @@ def tours_of_duty
...
-      branch = case military_service_episode.hca_branch_of_service
-               when 'noaa'
-                 military_service_episode.hca_branch_of_service.upcase
-               when 'usphs'
-                 'Public Health Service'
-               else
-                 military_service_episode.hca_branch_of_service.titleize
-               end
-
+    def service_name_used_in_diability(military_service_episode)
```
ericboehs, line 145: I think this should be "disability" rather than "diability".

app/models/emis_redis/military_information.rb
```diff
@@ -154,20 +154,21 @@ def build_service_branch(military_service_episode)
...
 
     # @return [Array<Hash>] Data about the veteran's service periods
     #  including service branch served under and date range of each
-    #  service period
+    #  service period, used only for Form 526 - Disability form
     def service_periods
       service_episodes_by_date.map do |military_service_episode|
-        # avoid prefilling if service branch is 'other' as this breaks validation
-        return {} if military_service_episode.hca_branch_of_service == 'other'
+        sn = service_name_used_in_diability(military_service_episode)
```
ericboehs, line 167: Could we use `service_name` here instead of `sn`?

spec/models/emis_redis/military_information_spec.rb
```diff
@@ -15,6 +27,14 @@
     end
   end
 
+  describe '#service_name_used_in_diability' do
+    it 'something' do
```
ericboehs, line 31: Could we get a more descriptive test name here?

ericboehs (changes requested):

spec/models/emis_redis/military_information_spec.rb
annaswims: 🤦‍♀️absolutely!

annaswims:
specs will pass once https://github.com/department-of-veterans-affairs/vets-api/pull/5249 is merged

ericboehs (approved):

PR #5225, opened 2020-11-04
massage zeitwerk into working for eb-zeit-de-formz

erluti:
Changes to `eb-zeit-de-formz` so that it will successfully run zeitwerk.


erluti:
Continuing to push through and see what's needed for zeitwerk to turn on, but this will probably be broken into smaller PRs

config/application.rb
```diff
@@ -73,5 +72,9 @@ class Application < Rails::Application
                                    key: 'api_session',
                                    secure: Settings.session_cookie.secure,
                                    http_only: true
+
+    # These files do not contain auto-loaded ruby classes,
+    #   they are loaded through app/workers/education_form/forms/base.rb
+    Rails.autoloaders.main.ignore(Rails.root.join('app', 'workers', 'education_form', 'templates', '1990-disclosure'))
```
ericboehs, line 78: Seems like we'd either have none of these or lots of these. One seems suspicious.

erluti: Interesting point.  These are the only ones that fail tests.  I'll look around for some more

ericboehs: Is there another way to prevent the failures but allow autoloading? Maybe that's what the others are doing.

config/application.rb
```diff
@@ -29,7 +29,6 @@ module VetsAPI
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults 6.0
-    config.autoloader = :zeitwerk
```
ericboehs, line 32: So removing this line enables zeitwerk?

erluti: Yes, zeitwerk is the default for 6.0.  The `config.autoloder = :classic` was added to override defaults.

config/application.rb
erluti: Ah, part of the problem is those files were named with `_CH30.erb` (or other numbers) and there isn't a good way to inflect these to satisfy the loader.  But since they aren't dynamic ruby classes, I think ignoring them here was the best approach.

erluti: Here's the ticket surrounding for this https://app.zenhub.com/workspaces/vsp-5cedc9cce6e3335dc5a49fc4/issues/department-of-veterans-affairs/va.gov-team/15891

erluti: Other places with the `File.read` + `ERB.new` pattern have files that match the autoloading conventions.  But perhaps they should be excepted here.  I think it's largely moot though.

ericboehs (approved):

PR #5228, opened 2020-11-05
Provide a Way to Access All Active BIRLS IDs (MPI lib)

edmangimelli:
MVI::Responses::IdParser's `#parse` method returns the first active BIRLS ID found in the MPI response, and ignores the rest. A veteran could possibly have multiple BIRLS IDs (although probably only one is valid). As a developer, I would like all the BIRLS IDs returned so I can find the right one.

### DIFF
--- a/lib/mvi/responses/id_parser.rb
+++ b/lib/mvi/responses/id_parser.rb
@@ -49,7 +49,7 @@ module MVI
 
       def parse(ids)
         ids = ids.map(&:attributes)
-        {
+        result = {
           icn: select_ids(select_extension(ids, PERMANENT_ICN_REGEX, VA_ROOT_OID))&.first,
           sec_id: select_ids(select_extension(ids, /^\w+\^PN\^200PROV\^USDVA\^A$/, VA_ROOT_OID))&.first,
           mhv_ids: select_ids(select_extension(ids, /^\w+\^PI\^200MH.{0,1}\^\w+\^\w+$/, VA_ROOT_OID)),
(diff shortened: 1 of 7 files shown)


lib/mvi/responses/id_parser.rb
```diff
@@ -59,10 +59,12 @@ def parse(ids)
           vha_facility_ids: select_facilities(select_extension(ids, /^\w+\^PI\^\w+\^USVHA\^\w+$/, VA_ROOT_OID)),
           cerner_facility_ids: select_facilities(select_extension(ids, /^\w+\^PI\^\w+\^USVHA\^C$/, VA_ROOT_OID)),
           cerner_id: select_ids(select_extension(ids, /^\w+\^PI\^200CRNR\^US\w+\^A$/, VA_ROOT_OID))&.first,
-          birls_id: select_ids(select_extension(ids, /^\w+\^PI\^200BRLS\^USVBA\^A$/, VA_ROOT_OID))&.first,
+          birls_ids: birls_ids(ids),
           vet360_id: select_ids(select_extension(ids, /^\w+\^PI\^200VETS\^USDVA\^A$/, VA_ROOT_OID))&.first,
           icn_with_aaid: ICNWithAAIDParser.new(full_icn_with_aaid(ids)).without_id_status
         }
+        result[:birls_id] = result[:birls_ids].first
```
ericboehs, line 66: It sort of feels weird to set a singular value for an array operation.

ericboehs (approved):
Minor nit. Otherwise lgtm.

lib/mvi/responses/id_parser.rb
annaswims: yeah, but that's what keeps it backwards-compatible for all of the places someone is calling `birls_id`

annaswims:
LGTM (github is not letting me put this as an actual review, seems like a UI bug)

PR #5240, opened 2020-11-06
Rollback va inflection

erluti:
slack convo leading up to reversion

1 commit is just a `git revert`
there was a conflict, so the second commit is adding the right pdf


kfrz (approved):
Approving so we can reduce errors in production

ericboehs (approved):
:+1: once tests pass.

PR #5243, opened 2020-11-06
update source app names

annaswims:
## Description of change
our source app list is out of date. it us used in stats like this one http://grafana.vfs.va.gov/d/zGS4XzkMk/backend-to-frontend-app-mapping?orgId=1&from=now-7d&to=now&var-data_source=Prometheus%20(Production)&var-backend_controller=All&var-frontend_app=All

## Original issue(s)


ericboehs (approved):

annaswims:
this list should be empty after deploy http://grafana.vfs.va.gov/d/zGS4XzkMk/backend-to-frontend-app-mapping?viewPanel=6&orgId=1&from=now-7d&to=now&var-data_source=Prometheus%20(Production)&var-backend_controller=All&var-frontend_app=not_in_whitelist

PR #5244, opened 2020-11-06
Mobile push notifications api

jjulian:
## Description of change
Design for Mobile API push notifications. Register, get prefs, set prefs. Will eventually connect with the VEText service. Any suggestions welcome, especially with openapi stuff.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#14083

### POST /v0/push/register
request body: 
```
{
  "data": {
    "attributes": {
      "deviceToken": "740f4707bebcf74f9b7c25d48e3358945f6aa01da5ddb387462c7eaf61bb78ad",
      "osName": "ios",
      "osVersion": "12.1",
      "deviceName": "Jeff's iPhone"
    }
  }
}
```

### GET /v0/push/prefs
response body: 
```
{
  "data": {
    "type": "pushPreferences",
    "attributes": {
      "pushNotificationTypes": [
        {
          "key": "appointment_reminders",
          "displayName": "Appointment Reminders",
          "enabled": true
        }
      ]
    }
  }
}
```

### PUT /v0/push/prefs
request body: 
```
{
  "data": {
    "attributes": {
      "pushNotificationTypes": [
        {
          "key": "appointment_reminders",
          "enabled": true
        }
      ]
    }
  }
}
```


modules/mobile/docs/openapi.yaml
```diff
@@ -1284,6 +1483,31 @@ components:
...
+        osVersion:
+          type: string
+          example: "12.1"
+        deviceName:
+          type: string
+          example: "Jeff's iPhone"
+    PushPrefs:
+      properties:
+        appointment_reminders:
+          type: boolean
```
jjulian, line 1510: We only will support appointment_reminders to start, but what's the best way to handle a growing list of notification types?

modules/mobile/docs/openapi.yaml
```diff
@@ -1122,6 +1122,205 @@ paths:
...
+          application/json:
+            schema:
+              $ref: '#/components/schemas/PushRegistration'
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/PushRegistration'
```
jjulian, line 1145: What is an appropriate response for a basic endpoint like this? The same data that was passed, or just 200 OK? What is the norm with the rest of the mobile api?

kreek: for updates (PUT) we use just a 200 with no body, since this is a POST could it be a 201 created (also no body)?

modules/mobile/docs/openapi.yaml
kreek: Depends how the FE wants to handle it, if the FE 'knows' what preference keys will be available then this works. If it's meant to be truly dynamic then it could be an array of objects with maybe `[{ "name": "appointment_reminders", "enabled": true}]`

modules/mobile/docs/openapi.yaml
```diff
@@ -1284,6 +1483,31 @@ components:
         financialInstitutionRoutingNumber:
           type: string
           example: "021000021"
+    PushRegistration:
```
kreek, line 1486: This core body looks good, because we use JSON:API spec it gets wrapped in a resource e.g.
```json
{
  "data": {
    "type": "preferences",
    "id": "550e8400-e29b-41d4-a716-446655440000",
    "attributes": {

    }
  }
}
```

If it's a list then it's an [array of resources](https://jsonapi.org/examples/). You could debate if preferences is one resource or a list of resources (I lean toward it being one resource).

jjulian: Cool I will wrap these object definitions

modules/mobile/docs/openapi.yaml
jjulian: I think truly dynamic will server the API best in the long run. Changes made.

modules/mobile/docs/openapi.yaml
```diff
@@ -1284,6 +1475,78 @@ components:
...
+        - enabled
+        - displayName
+      properties:
+        key:
+          type: string
+          example: "appointment_reminders"
+        displayName:
+          type: string
+          example: "Appointment Reminders"
+          readOnly: true
```
jjulian, line 1547: `readOnly` keeps this property from being documented in the request body

jperk51 (approved):
LGTM

kreek (approved):
LGTM 🚀

jjulian:
Danger: Ignore "Big PRs are difficult to review"

PR #5245, opened 2020-11-07
When a SubmitForm526 Job Fails, Try a Different BIRLS ID

edmangimelli:
resolves https://github.com/department-of-veterans-affairs/vets-api/issues/5268

(i talk a bit about the implementation in the issue ^^^)


annaswims:
I'm wondering if we might be able to simplify this a lot if we stored all the potential bilrs id's when we create Form526Submission

app/controllers/v0/disability_compensation_forms_controller.rb
```diff
@@ -66,12 +66,17 @@ def create_submission(saved_claim)
       Rails.logger.info(
         'Creating 526 submission', user_uuid: @current_user&.uuid, saved_claim_id: saved_claim&.id
       )
-      Form526Submission.create(
-        user_uuid: @current_user.uuid,
-        saved_claim_id: saved_claim.id,
-        auth_headers_json: auth_headers.to_json,
-        form_json: saved_claim.to_submission_data(@current_user)
-      )
+      ActiveRecord::Base.transaction do
```
annaswims, line 69: not that it's your doing, but the whole `create_submission` method seems like it belongs in the Form526Submission model.

LindseySaari (approved):

LindseySaari:
Looks good.. but looks to be an issue with an existing cassette

PR #5246, opened 2020-11-08
[WIP] Add temp logger to Central Mail to debug issue

digitaldrk:
## Description of change
I was hoping to send this into a review instance but those aren't working currently so I wanted to check in with y'all if this would be ok.

### Problem
We are trying to fire off a form that uses `CentralMail`. We have leveraged existing code but can't seem to successfully fire this off: `response = CentralMail::Service.new.upload(create_request_body)`

_*Locally*_, we get this:
image
image

Since `response.success?` evaluates to `false`, we raise [this](https://github.com/department-of-veterans-affairs/vets-api/blob/169018e9bc884c7c1bdda30d485776f599726ce7/app/workers/central_mail/submit_saved_claim_job.rb#L34)

_*In Staging*_
Obv I can't drop a `binding.pry` so I can't see the response object. I assumed it was the same problem but I submitted a different form in staging that uses this code and it worked.

I'd like to send in this code, test the issue and remove it once it's resolved. WDYT?


app/workers/central_mail/submit_saved_claim_job.rb
```diff
@@ -28,6 +30,7 @@ def perform(saved_claim_id)
       File.delete(@pdf_path)
       @attachment_paths.each { |p| File.delete(p) }
 
+      log_message_to_sentry(response.to_s, :warn, {}, { team: 'vfs-ebenefits' })
```
kfrz, line 33: What does this response object contain? Is there any risk of accidentally capturing PII/PHI in Sentry if this is enabled?

kfrz:
@digitaldrk can you SSM into a staging box and do a `bin/rails c` to manually create this request?

digitaldrk:
> @digitaldrk can you SSM into a staging box and do a `bin/rails c` to manually create this request?

I'll give that a shot

PR #5247, opened 2020-11-09
use VA inflection

erluti:
After a bug was discovered with in progress forms, so VA inflection was rolled back.  This PR brings it back in including a fix for dealing with the in-progress forms.

Here's the ticket for more history
department-of-veterans-affairs/va.gov-team#12659

PR we're bringing back https://github.com/department-of-veterans-affairs/vets-api/pull/4792


spec/lib/olive_branch_patch_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+
+  it 'does not add keys if `VA` is not in the middle of a key' do
+    hash = { hello_there: 'hello there' }
+    get '/some_json', params: hash, headers: { 'X-Key-Inflection' => 'camel' }
+    json = JSON.parse(response.body)
+    expect(json.keys).to eq ['helloThere']
+    expect(json['helloThere']).to eq hash[:hello_there]
+  end
+
+  it 'duplicates va keys containing a colon' do
```
erluti, line 63: this is the test for the use case that was discovered to break this patch

lib/olive_branch_patch.rb
```diff
@@ -0,0 +1,79 @@
...
+    end
+    result
+  end
+
+  private
+
+  VA_KEY_VALUE_PAIR_REGEX = /("[^"]+VA[^"]*"):("[^"]*"|\d+|true|false|null)/.freeze
+
+  def duplicate_basic_va_keys!(json)
+    json.gsub!(VA_KEY_VALUE_PAIR_REGEX) do
```
erluti, line 29: this method changed from the original version of the patch https://github.com/department-of-veterans-affairs/vets-api/pull/5133/files
now using the regex references instead of trying to split on `:`

spec/request/in_progress_forms_request_spec.rb
```diff
@@ -84,19 +84,20 @@
         end
 
         context 'with the x key inflection header set' do
-          let(:form_data) do
-            { 'view:hasVaMedicalRecords' => true }
-          end
-
           it 'converts the json keys' do
+            form_data = { 'view:hasVaMedicalRecords' => true }
```
erluti, line 88: this test was changed to reflect the double VA keys being returned

LindseySaari (approved):
Looks good and nice test coverage!

erluti:
Moved to draft until cross-team collaboration finds an approach.

spec/request/in_progress_forms_request_spec.rb
```diff
@@ -84,19 +84,20 @@
...
             get v0_in_progress_form_url(in_progress_form.form_id),
                 params: nil,
                 headers: { 'HTTP_X_KEY_INFLECTION' => 'camel' }
-            expect(response.body).to eq({
-              form_data: form_data,
-              metadata: in_progress_form.metadata
-            }.deep_transform_keys { |key| key.to_s.underscore.camelize(:lower) }.to_json)
+            body = JSON.parse(response.body)
+            expect(body.keys).to include('formData', 'metadata')
+            # using VA as Va is deprecated, when forms are updated to use VA, this test should reflect that
```
annaswims, line 96: How/why is the deprecation happening?

erluti: it's not.  this comment is old. good catch!

LindseySaari (approved):
Nice spec coverage and the middleware patch approach looks really straightforward. I think I like this solution best (as opposed to previous approaches). I'm liking that we're making Zeitwerk mode WRT inflections happy and FE happy as well with the camel cased keys that they're expecting

lib/olive_branch_patch.rb
```diff
@@ -0,0 +1,41 @@
...
+# have the old-inflection somethingVaSomething.
+
+module OliveBranchMiddlewareExtension
+  def call(env)
+    result = super(env)
+    _status, _headers, response = result
+    if env['HTTP_X_KEY_INFLECTION'] =~ /camel/i
+      response.each do |json|
+        # do not process strings that aren't json (like pdf responses)
+        next unless json.is_a?(String) && json.starts_with?('{')
```
annaswims, line 14: This doesn't prove that it's json and is janky.  you could just look at headers["Content-Type"]

erluti: pushed an update.  uses `send` which is also a bit janky, but it relies on the same method that olive_branch uses to determine if it should transform any values.  What do you think?

spec/lib/olive_branch_patch_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+  end
+
+  after(:all) do
+    Rails.application.reload_routes!
+  end
+
+  it 'does not change response keys when camel inflection is not used' do
+    hash = { hello_to_the_va: 'greetingsVA' }
+    get '/some_json', params: hash
+    json = JSON.parse(response.body)
```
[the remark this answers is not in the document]
erluti: thanks, i'm looking more into this.  seems like it has more to do with my test controller than with the header causing problems.

lib/olive_branch_patch.rb
annaswims: Thanks!  I prefer that!

spec/lib/olive_branch_patch_spec.rb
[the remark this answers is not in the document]
erluti: added the content-type header to the tests

lib/olive_branch_patch.rb
```diff
@@ -0,0 +1,41 @@
...
+# have the old-inflection somethingVaSomething.
+
+module OliveBranchMiddlewareExtension
+  def call(env)
+    result = super(env)
+    _status, headers, response = result
+    # olive branch uses this private method to determine if a tranformation should happen https://github.com/vigetlabs/olive_branch/blob/master/lib/olive_branch/middleware.rb#L92
+    return result if send(:exclude_response?, env, headers)
+
+    if env['HTTP_X_KEY_INFLECTION'] =~ /camel/i
```
annaswims, line 14: you look for `HTTP_X_KEY_INFLECTION` here, but the specs use 'X-Key-Inflection'

erluti: That's how it looks in the ENV in middleware

lib/olive_branch_patch.rb
```diff
@@ -0,0 +1,41 @@
...
+  def un_camel_va_keys!(json)
+    # rubocop:disable Style/PerlBackrefs
+    # gsub with a block explicitly sets backrefs correctly https://ruby-doc.org/core-2.6.6/String.html#method-i-gsub
+    json.gsub!(VA_KEY_REGEX) do
+      key = $1
+      "#{key.gsub('VA', 'Va')}:"
+    end
+    # rubocop:enable Style/PerlBackrefs
+  end
+end
```
edmangimelli, line 35: Is there any way we could accomplish this without regexing unparsed json?

erluti: I can't say "no".  But this seemed to be the most clear and straightforward way to accomplish transforming complex keys.  Is there a use-case it's breaking or a similar situation you need ideas for?

edmangimelli: It should seems a little dangerous when we could parse the json, recurse through it, and operate only on the keys. Agreed, with your regex, it's extremely unlikely that values will be affected, but still possible:
```
require 'json'

VA_KEY_REGEX = /("[^"]+VA[^"]*"):/.freeze

def un_camel_va_keys!(json)
  json.gsub!(VA_KEY_REGEX) do
    key = $1
    puts key
    "#{key.gsub('VA', 'Va')}:"
  end
end

json = <<-HEREDOC
{
  "disabilities": [
    "\\"CVA\\": stemming from traumatic brain injury."
  ]
}
HEREDOC

JSON.parse json

un_camel_va_keys! json

json
# => "{\n  \"disabilities\": [\n    \"\\\"CVa\\\": stemming from traumatic brain injury.\"\n  ]\n}\n"
```

erluti: Is that a real value?  Are there instances of strings that include a quoted value, a colon, and more data?

edmangimelli: 🤷‍♂️ lot's of strange data in the wild, but, like I said, I would guess extremely unlikely.
then again, murphy's law

erluti: I feel that.  But if it's nothing something we'd expect, I like this approach.  Got to remember this code is run for EVERY response we generate; so parsing json, manipulating a hash, and converting it back to a string will take up time and electricity for a "what if?".

PR #5248, opened 2020-11-09
API-3255: Background job for Reading Flashes from schema and posting them to BGS

SlakrHakr:
## Description of change
Implements background job to add provided flashes for a claim to BGS. Job will be referenced in future work. 

## Original issue(s)
https://vajira.max.gov/browse/API-3255

spec tests added

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -34,7 +34,7 @@ GIT
 
 GIT
   remote: https://github.com/department-of-veterans-affairs/bgs-ext.git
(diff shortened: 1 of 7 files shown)


ericboehs (approved):

bastosmichael (approved):
Looks good though I think you're going to have to need to somehow figure out how to pass the Representative user to the sidekiq job if it's not a Veteran Oauth submission.

PR #5252, opened 2020-11-09
Add #create endpoint

ElleeB:
## Description of change
Added the #create endpoint to the notice_of_disagreement controller
- asynchronously initiates NoticeOfDisagreementPdfSubmitJob
- returns serialized json
- added tests to NOD controller spec

## Original issue(s)
https://vajira.max.gov/browse/API-3379

## Testing
Request test added to check for data persistence, headers, and presence of pdf


modules/appeals_api/app/serializers/appeals_api/notice_of_disagreement_serializer.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AppealsApi::NoticeOfDisagreementSerializer
+  include FastJsonapi::ObjectSerializer
+  set_key_transform :camel_lower
+  attributes :updated_at, :created_at, :form_data
```
charleystran, line 6: Are we missing an attribute for status?

ElleeB: Status wasn't included in the migration. I checked with Ed and Steven and they said to add it separately in another ticket/PR. Let me know if I should do this sooner than later.

charleystran (approved):
LGTM :shipit:

PR #5253, opened 2020-11-10
Flipper cleanup - Case insensitive email

annaswims:
## Description of change
we ran into a case today during UAT where the user's email address was added to the whitelist with a capital letter and it didn't work. I'm making email address case insensitive. Doing it in Javascript is weird, but it would've been more of a hassle to open the door to overwriting the flipper API.

While I was in the neighborhood, I deleted code that should never be used (we're past the point where it's feasible to add/remove features in the UI) and correcte the terminology for the FLIPPER_ID cookie (at one point we considered the Google Analytics ID).

it would be a great time to merge this as the only feature that use email address are owned by my team and will be unaffected (we've been adding email address one-by-one for UAT)

## Original issue(s)

updated text
Screen Shot 2020-11-10 at 11 26 53 AM


app/models/user.rb
```diff
@@ -321,7 +321,7 @@ def power_of_attorney
   end
 
   def flipper_id
-    email || account_uuid
+    email.downcase || account_uuid
```
annaswims, line 324: pretty sure that the Identity provider was always sending lower case anyway, adding that just in case it changes.

lib/flipper/views/feature.erb
```diff
@@ -200,28 +200,19 @@
         </div>
       </div>
     </div>
-
-    <% if Flipper::UI.configuration.feature_removal_enabled %>
```
annaswims, line 204: never will be used

lib/flipper/views/feature.erb
```diff
@@ -200,28 +200,19 @@
...
+
+<script type="text/javascript">
+  function lowercase_email() {
+      var srch = document.getElementById("actor-input");
+      var str = srch.value;
+      if (str.indexOf("@") > -1) {
+        srch.value = str.toLowerCase();
+      }
+      var form = document.getElementById("actors_form");
+      form.submit();
```
ericboehs, line 216: Instead of changing the onsubmit behavior, what if we just ran this function `onchange` of the `<input>`?

I was going to suggest just doing `<input onchange: "this.value=this.value.toLowerCase()" ...>` but it seems you only want to do it if it has an `@`. It would simplify it quite a bit if we could remove that requirement.

lib/flipper/views/feature.erb
```diff
@@ -200,28 +200,19 @@
...
-        </div>
-      </div>
-    </div>
-    <% end %>
   </div>
 <% else %>
   <a href ="<%= SAML::URLService::VIRTUAL_HOST_MAPPINGS.dig(request.base_url, :base_redirect) %>">Sign in</a> with an identity-verified admin account to edit feature toggles.
 <% end %>
+
+<script type="text/javascript">
```
ericboehs, line 208: By default all `<script>` tags are assumed to be type `text/javascript`.

lib/flipper/views/feature.erb
```diff
@@ -200,28 +200,19 @@
...
+  function lowercase_email() {
```
ericboehs, line 209: I like Ruby too but JS peeps would name this `lowercaseEmail`.

lib/flipper/views/feature.erb
```diff
@@ -200,28 +200,19 @@
...
+      var srch = document.getElementById("actor-input");
+      var str = srch.value;
```
ericboehs, line 211: I'm not sure if this temp variable adds clarity/value.

lib/flipper/views/feature.erb
```diff
@@ -200,28 +200,19 @@
...
+      if (str.indexOf("@") > -1) {
+        srch.value = str.toLowerCase();
+      }
+      var form = document.getElementById("actors_form");
```
ericboehs, line 215: Same for this temp variable.

lib/flipper/views/feature.erb
```diff
@@ -200,28 +200,19 @@
...
-    </div>
-    <% end %>
   </div>
 <% else %>
   <a href ="<%= SAML::URLService::VIRTUAL_HOST_MAPPINGS.dig(request.base_url, :base_redirect) %>">Sign in</a> with an identity-verified admin account to edit feature toggles.
 <% end %>
+
+<script type="text/javascript">
+  function lowercase_email() {
+      var srch = document.getElementById("actor-input");
```
ericboehs, line 210: What does srch stand for? Any reason not to name it `actorInput`?

lib/flipper/views/feature.erb
```diff
@@ -200,28 +200,19 @@
...
+      var str = srch.value;
+      if (str.indexOf("@") > -1) {
+        srch.value = str.toLowerCase();
+      }
```
ericboehs, line 214:
```suggestion
      if (srch.indexOf("@") > -1) srch.value = srch.value.toLowerCase();
```

ericboehs (approved):
Lot's of small nits/suggestions but nothing blocking. Looks good.

lib/flipper/views/feature.erb
annaswims: Thanks!  It's been a decade or more since I've written JavaScript.

PR #5254, opened 2020-11-10
Add fetch_aud override for claims

hinzed1127:
## Description of change

Overrides the `fetch_aud` method to prepare claims v1 endpoints for the OAuth authorization server isolation effort. Dependent on  department-of-veterans-affairs/devops#7967

## Original issue(s)
https://vajira.max.gov/browse/API-2415

### DIFF
--- a/modules/claims_api/app/controllers/claims_api/v1/claims_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/v1/claims_controller.rb
@@ -25,10 +25,6 @@ module ClaimsApi
       def show
         super
       end
-
-      def fetch_aud
-        Settings.oidc.isolated_audience.claims
-      end
     end
   end
 end
--- a/modules/claims_api/app/controllers/claims_api/v1/claims_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/v1/claims_controller.rb
@@ -25,6 +25,10 @@ module ClaimsApi
       def show
         super
       end
+
+      def fetch_aud
+        Settings.oidc.isolated_audience.claims
+      end
     end
   end
 end
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -410,6 +410,7 @@ oidc:
   isolated_audience:
     default: ~
     veteran_verification: ~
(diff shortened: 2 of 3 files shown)


modules/claims_api/app/controllers/claims_api/v1/claims_controller.rb
```diff
@@ -25,6 +25,10 @@ def index
       def show
         super
       end
+
+      def fetch_aud
+        Settings.oidc.isolated_audience.claims
```
bastosmichael, line 30: So claims controller has nothing to do with the forms controllers so you may want to consider adding this to the application controller and just use the v0? flag to avoid using it in non Oauth scenarios. Also what calls the fetch_aud method?

hinzed1127: - I assumed from the ticket description that this was a change only intended for specifically the `/claims` endpoints. @dbrown-va is that an accurate reading, or should this actually go in the higher-level `claims_api` `application_controller`?
- `fetch_aud` gets called in the [`openid_application_controller`](https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/controllers/openid_application_controller.rb#L47)

hinzed1127: Update: moved it up to the `application_controller`. Non-OAuth endpoints will skip the `:validate` call that utilizes this anyway

dbrown-va: This only applies to the OAuth endpoints for /claims and would otherwise be unused for non OAuth endpoints.

My understanding is there shouldn't be harm for it being in the application controller, but definitely not required at that top level.  You just need to ensure it overrides the base `fetch_aud` method in the `openid_application_controller` for any OAuth protected endpoints.

ericboehs (approved):

bastosmichael (approved):
LGTM per our conversation with Derek...

dbrown-va (approved):

PR #5256, opened 2020-11-10
Updates Contact Us Swagger docs

jgacek:
## Description of change

Changes the example model to be simpler and links to vets-json-schema repo

## Original issue(s)
department-of-veterans-affairs/orchid#162


kfrz (commented):
It looks like this is removing all the required attributes for the request body. Is that the intention? This removes the examples also as far as I can tell. 

If that's the intent this looks fine!

jgacek:
Yep exactly the intention

kfrz (approved):
:+1: :ship:

PR #5258, opened 2020-11-10
Initial commit. Worker, task and scheduler added.

bradenshipley:
### Description of change
One worker file and single service class added to apps_api module to populate directory_applications table if the data is not already present. Scheduled job added to sidekiq-scheduler to run 1m after startup to invoke our worker file.

##Original issue(s)
https://vajira.max.gov/secure/RapidBoard.jspa?rapidView=777&view=detail&selectedIssue=API-3359

##Things to know about this PR
SimpleCov has also been set to ignore the service file, since there is no real way to test the do blocks of each find_or_create_by! call for each attribute set without becoming spaghetti.

Manual testing has been done locally. The task populates the table and development logs show the sidekiq job being invoked. Tests for the worker included.


modules/apps_api/app/workers/apps_api/directory_loader.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+require 'sidekiq'
+require 'rake'
```
SlakrHakr, line 4: Requiring rake shouldn't be required here I don't think?

bradenshipley: ah this is leftover code! whoops

SlakrHakr (approved):

PR #5259, opened 2020-11-10
Ensuring failed user authentication does not return a 500

bastosmichael:
## Description of change
Figured out that the problem with the Non MPI users is that if you attempt to make a veteran request with them it fails and errors out in both cases. Hoping this fixes the issue long term.

## Original issue(s)
https://vajira.max.gov/browse/API-3336


modules/claims_api/app/controllers/claims_api/application_controller.rb
```diff
@@ -133,5 +133,15 @@ def veteran_from_headers(with_gender: false)
...
+
+    def authenticate_token
+      super
+    rescue => e
+      log_message_to_sentry('Authentication Error in claims',
+                            :warning,
+                            body: e.message)
+      render json: { errors: [{ status: 401, detail: 'User not a valid or authorized Veteran for this end point.' }] },
+             status: :unauthorized
+    end
```
SlakrHakr, line 145: why not make this change higher up? this seems to make sense for any endpoint within vets-api no?

SlakrHakr, line 145: nit: all errors within this method aren't guaranteed to mean the person "isn't authorized"
Possible services we use for authentication may be down which would maybe make more sense to respond with a 500 or 503 i would think?
but i could also see them "not being authorized" as technically correct since we can't authorize them because a system we use to do the authorizing isnt available to us :shrug:

bastosmichael: So in this case we are hijacking the authorization method from the Open ID controller and just injecting a raise in the event anything goes wrong with validating the user, All of our controllers inherit from this module controller so made the most sense to put it here so that it'll apply across all of the end points minus the ones in V0 which skip token authentication so this wouldn't apply to them anyhow.

bastosmichael: I've seen too many incidents that could provoke this to fail from Okta not being able to process our token at the time to the user not existing in MPI which we support but va.gov does not so it makes it hard for us to catch all of the instances where this might need to be rescued from potentially. The logging into sentry was a good middle ground for ensuring our users don't get a 500 error because an upstream partner is down or we can't parse the token for whatever reason and so it's still resides in sentry long term and if you think the detail verbiage could be different feel free to recommend something different but our users shouldn't 500 for any instance that we can't authenticate etc.

SlakrHakr: > All of our controllers inherit from this module controller

All of our controllers within the claims module correct? I guess I was more asking why this change doesn't make sense higher up so to affect all modules within the api, but maybe thats simply frowned upon? or maybe each module uses authentication slightly differently rendering a broad change like this almost impossible?

Fair enough though :+1:

SlakrHakr (approved):

modules/claims_api/app/controllers/claims_api/application_controller.rb
hinzed1127: +1 to @SlakrHakr 's question. `authenticate_token` doesn't have any explicit error handling in the `OpenidApplicationController`. I can't see why we wouldn't want this to be higher up.

Put another way: this message feels general enough to live at a higher level of abstraction from claims, but is it too specific to put all the way at the top?

hinzed1127: I think this works back to the the other question for me. Are there other endpoints that would run into similar errors currently that would benefit from getting a better error than a potential 500, or is that outside the realm of our responsibility?

bastosmichael: Each module can have different authentication rules because they belong to different teams and use cases, what we are inheriting is a simpler implementation that we are adding Respresentative Oauth rules that only we need to support for now. Hopefully one day Veteran Verification or even VA.gov support non veterans and we can move this stuff a bit higher upstream but at the moment we are the only ones with this particular use case.

bastosmichael: We shouldn't really be returning a 500 unless our app shit the bed completely, even in cases where upstream access isn't available for whatever reason we should be maybe restricting to a 422 or a 401 with a proper error message to explain what the problem is but I could be wrong here though the primary reason is that too many upstream VA systems can be unreliable and it's our job to mitigate that for our API consumers.

SlakrHakr (approved):

PR #5262, opened 2020-11-10
Api 3058 Veteran Verification Only - Use Mock eMIS

lukemarkwordtlibertyits:
To assist with providing specific test data for Veteran Verification endpoints, team Pivot! has created a mock-emis service. This PR allows the Vets-API to integrate with mock-emis only if specified for the vet-verification service.

[API3058_Tests.docx](https://github.com/department-of-veterans-affairs/vets-api/files/5519757/API3058_Tests.docx)


esubocz-va (changes requested):

modules/veteran_verification/spec/requests/mock_veteran_status_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+  let(:edipi_veteran) { '1068619536' }
+  let(:edipi_non_veteran) { '1140840595' }
+  let(:bad_edipi) { '595' }
+  let(:missing_edipi) { '1111111111' }
+  let(:no_status) { '1005079361' }
+
+  before do
+    Settings.vet_verification.mock_emis = true
+    Settings.vet_verification.mock_emis_host = 'https://vaausvrsapp81.aac.va.gov'
+  end
```
esubocz-va, line 16: so, weird thing with specs and the `before` case, is that it can stay like this for the rest of the tests (no idea why) so for the non-mock tests, you'll need to add a 
```
before do
    Settings.veteran_verification.mock_emis = false
end
```
we found this out the hard way with mock-bgs deployment

esubocz-va:
also, unless I missed it, I think ya still need a mock military information spec

Gemfile.lock
```diff
@@ -1002,4 +991,4 @@ RUBY VERSION
    ruby 2.6.6p146
 
 BUNDLED WITH
-   1.17.3
+   1.17.3
```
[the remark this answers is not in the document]
lukemarkwordtlibertyits: Will revert this.

modules/veteran_verification/app/services/emis/mock_military_information_configuration_v2.rb
```diff
@@ -0,0 +1,11 @@
...
+
+require 'common/client/configuration/soap'
+
+module EMIS
+  class MockMilitaryInfoConfig < MilitaryInformationConfigurationV2
+    def base_path
+      URI.join(Settings.vet_verification.mock_emis_host, Settings.emis.military_information_url.v2).to_s
+    end
+  end
+end
```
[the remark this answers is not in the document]
lukemarkwordtlibertyits: good point

modules/veteran_verification/spec/requests/mock_veteran_status_spec.rb
lukemarkwordtlibertyits: oh good call.

lukemarkwordtlibertyits:
> also, unless I missed it, I think ya still need a mock military information spec

Code coverage is at 100 percent for that module. My MockOpenIDUser tests covers it.

esubocz-va (approved):

modules/veteran_verification/app/controllers/veteran_verification/application_controller.rb
```diff
@@ -11,6 +12,21 @@ class ApplicationController < ::OpenidApplicationController
     # so we set that default.
     before_action { set_default_format_to_json }
 
+    def authenticate_token
+      return false if token.blank?
+
```
dbrown-va, line 17: Depending on which PR goes first, this `authenticate_token` method will also need to bring in the `if token.client_credentials_token?` check to skip the session & open_id user build up.

modules/veteran_verification/app/controllers/veteran_verification/application_controller.rb
```diff
@@ -11,6 +12,21 @@ class ApplicationController < ::OpenidApplicationController
...
+      @session = Session.find(token)
+      establish_session if @session.nil?
+      return false if @session.nil?
+
+      open_id = if Settings.vet_verification.mock_emis
```
dbrown-va, line 22: Just a reminder, but knowing this is still in Draft stages.

Will need config settings to similar to mock_bgs.

https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/deployment/config/vets-api/dev-settings.local.yml.j2#L437

dbrown-va:
I confirmed cached response from the betamocks mocking still work after pointing to the mock-emis service (and the reverse if for some reason a rollback had to occur).

dbrown-va:
This looks good.  I recommend we move this to Read for Review, and add the corresponding devops PR that includes the new settings.

```
Settings.vet_verification.mock_emis
Settings.vet_verification.mock_emis_host
```

modules/veteran_verification/app/controllers/veteran_verification/application_controller.rb
dbrown-va: The Client Credentials validation PR has been merged, this method will need a similar checked mentioned above.

modules/veteran_verification/app/controllers/veteran_verification/application_controller.rb
dbrown-va: https://github.com/department-of-veterans-affairs/devops/pull/7992

dbrown-va (changes requested):
One small change needed in the application_controller since the `authenticate_token` method was brought in to the module.

modules/veteran_verification/app/controllers/veteran_verification/application_controller.rb
dbrown-va: This has been resolved with the addition of the `token.client_credentials_token?` on line 14.

dbrown-va (approved):

PR #5263, opened 2020-11-10
Decision review timeout matches Lighthouse timeout

annaswims:
## Description of change
HLR's decision review service should not time out if Lighthouse's Caseflow service has. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#16030

### DIFF
--- a/spec/lib/decision_review/configuration_spec.rb
+++ b/spec/lib/decision_review/configuration_spec.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require 'caseflow/configuration'
+
+describe DecisionReview::Configuration do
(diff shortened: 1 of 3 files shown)


spec/lib/decision_review/configuration_spec.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require 'caseflow/configuration'
+
+describe DecisionReview::Configuration do
+  describe '.read_timeout' do
+    context 'when Settings.caseflow.timeout is set' do
```
ericboehs, line 8: Should we add a test for when it's not set to verify 20 is there?

ericboehs (approved):
Small question. LGTM.

PR #5265, opened 2020-11-11
15839 ppms skip the round trip

darkbushido:
## Description of change
The CCP endpoint makes a call to the PPMS provider locator endpoint. For each provider, we request additional data from the provider endpoint. These additional calls are slow and since we have to do them serially they add a lot to the overall request. PPMS has recently made a change in their backend to include the provider type (the main thing we were getting from the additional calls)

This PR skips the additional calls for the V1 PPMS client

## Original issue(s)
department-of-veterans-affairs/va.gov-team#15839

* New Feature flag `facility_locator_ppms_skip_additional_round_trips`
* This will skip the additional calls to PPMS
* some of the data fields we currently populate will return `nil`


spec/lib/facilities/ppms/v0/client_spec.rb
```diff
@@ -53,7 +53,7 @@
           Latitude: 33.262403,
           Longitude: -111.808538,
           MainPhone: nil,
-          Miles: 1.679,
+          Miles: 1.842,
```
kfrz, line 56: What's the reason for this value change?

kfrz (approved):
These changes look good to me, just one minor comment.  I can override the Dangerbot for this one, I think, given the changes to a cassette file.

spec/lib/facilities/ppms/v0/client_spec.rb
darkbushido: PPMS returned a different result for the same query.

PR #5270, opened 2020-11-11
SAML cookie logging

bramleyjl:
## Description of change

This change cleans up SAML cookie content creation logic to reduce unnecessary random ID creation by storing the current SAML transaction ID in Redis, as well as including it in the `saml_request_stats` and `saml_response_stats` methods that generate the SAML values logged with Rails logger.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#15179

Manual testing has been done on logins to check for SAML transation IDs, unit tests still need to be updated as mocked SAML request trackers are not generating transaction IDs.

### DIFF
--- a/spec/controllers/v1/sessions_controller_spec.rb
+++ b/spec/controllers/v1/sessions_controller_spec.rb
@@ -108,18 +108,17 @@ RSpec.describe V1::SessionsController, type: :controller do
                 .and trigger_statsd_increment(described_class::STATSD_SSO_SAMLREQUEST_KEY,
                                               tags: ["type:#{type}", "context:#{authn}", 'version:v1'], **once)
 
-              expect(SAMLRequestTracker.keys.length).to eq(1)
-              tracker = SAMLRequestTracker.find(SAMLRequestTracker.keys[0])
-              expect(tracker.payload)
-                .to eq({ type: type, authn_context: authn, transaction_id: tracker.payload[:transaction_id] })
               expect(response).to have_http_status(:ok)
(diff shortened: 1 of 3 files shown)


ericbuckley:
@bramleyjl do you need more direction on how to fix the remaining tests or do you have enough to go on right now?

spec/controllers/v1/sessions_controller_spec.rb
```diff
@@ -547,7 +554,7 @@ def expect_logger_msg(level, msg)
 
         it 'redirects to identity proof URL', :aggregate_failures do
           Timecop.freeze(Time.current)
-          expect_any_instance_of(SAML::PostURLService).to receive(:should_uplevel?).twice.and_return(true)
+          expect_any_instance_of(SAML::PostURLService).to receive(:should_uplevel?).once.and_return(true)
```
ericbuckley, line 557: @bramleyjl I wouldn't have expected this test case to change

bramleyjl: In `saml_cookie_content` we removed the more complicated transaction_id check that included a `should_uplevel` call in favor of the Redis cache, so now `should_uplevel` is only being called once, inside `user_login`

ericbuckley: ah thats right, thanks for pointing that out

ericbuckley (approved):
LGTM

kfrz (approved):
This looks good to me, thanks for adding appropriate spec!

PR #5274, opened 2020-11-11
Cache VA profile get_person calls that return 404s

lihanli:
## Description of change
cache 404 responses from va profile get_person api

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/15971

testing in staging planned


kfrz (approved):
LGTM! :ship:

PR #5275, opened 2020-11-12
Update Appeals API namespacing to include concerns

LindseySaari:
## Description of change
This PR updates the directory structure for the appeals_api json_format_validation concern to ensure proper autoloading. With the updated file structure, the require_dependency's in the code can be removed because if structured properly, concerns are part of the autoload paths. Following best coding practices, concerns don't need to be part of the module namespacing.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#15903

## Testing
- [x] Specs passing locally


erluti (approved):
👍

erluti (approved):
I think this approach is preferable!

PR #5281, opened 2020-11-12
Scope endpoint error handling

bradenshipley:
## Description of change
Changing scope endpoint to send a 204 if the `service_category` passed in the URL does not have an associated `Authentication Server`. Previously this would return an empty array but has been changed to be more clear.

## Original issue(s)
https://vajira.max.gov/secure/RapidBoard.jspa?rapidView=777&view=detail&selectedIssue=API-2189

Request test has been adjusted to account for 204 error code.


modules/apps_api/app/controllers/apps_api/v0/directory_controller.rb
```diff
@@ -9,18 +9,22 @@ class DirectoryController < ApplicationController
...
-          data: parsed_scopes
-        }
+        if parsed_scopes.any? do
+          render json: {
+            data: parsed_scopes
+          }
+           rescue
+             head :no_content
+        end
+        end
```
SlakrHakr, line 27: nit: can't say I've seen an "if do" block like this. might just be my unfamiliarity with it, but it was a bit difficult for me to parse. id maybe recommend having a private method that does the render and handling the rescue in there?
```
def scopes
  directory_service = Okta::DirectoryService.new
  parsed_scopes = directory_service.scopes(params[:category])
  
  render_with_parsed_scopes(parsed_scopes) if parsed_scopes.any?
end

private

def render_with_parsed_scopes(parsed_scopes)
  render json: {
    data: parsed_scopes
  }
rescue StandardError
  head :no_content
end
```

but messing around with the syntax you have here and it seems completely valid. just tossing my 2 cents in :+1:

SlakrHakr (approved):

PR #5282, opened 2020-11-13
CH31 (VA Form 28-1900) PDF Generation

kathleencrawford:
## Description of change

This PR generates the PDF for VA Form 28-1900 (also known as Chapter 31).  It is slightly over the recommended LoC limit - this is due to one file in particular that maps all the fields in the pdf to a hash.  

## Original issue(s)
department-of-veterans-affairs/va.gov-team#12463

There are no new settings or configuration added.  Also, no additional logging or swagger docs.  This is purely to generate the PDF from a spec.  The code to generate the PDF during a form submission will be added in a separate PR.

## Testing

- [x] Local testing done and specs written


kfrz (approved):
This looks fine to me, Dangerbot shouldn't block you from merging!

PR #5284, opened 2020-11-13
API-3251: Updating the claim establisher job to pull and trigger flashes

SlakrHakr:
## Description of change
Publish any flashes submitted with a 526 on successful claim establishment.

## Original issue(s)
https://vajira.max.gov/browse/API-3251

Tested hitting endpoint to submit 526 claim locally.
Updated specs.

### DIFF
--- a/modules/claims_api/app/workers/claims_api/claim_establisher.rb
+++ b/modules/claims_api/app/workers/claims_api/claim_establisher.rb
@@ -20,6 +20,9 @@ module ClaimsApi
       auto_claim.evss_id = response.claim_id
       auto_claim.status = ClaimsApi::AutoEstablishedClaim::ESTABLISHED
       auto_claim.save
+
+      flashes = auto_claim.form_data.dig('veteran', 'flashes')
+      ClaimsApi::FlashUpdater(bgs_user(auth_headers), flashes) if flashes.present?
     rescue ::EVSS::DisabilityCompensationForm::ServiceException => e
       auto_claim.status = ClaimsApi::AutoEstablishedClaim::ERRORED
       auto_claim.evss_response = e.messages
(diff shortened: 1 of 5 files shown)


modules/claims_api/app/workers/claims_api/claim_establisher.rb
```diff
@@ -20,6 +21,9 @@ def perform(auto_claim_id)
       auto_claim.evss_id = response.claim_id
       auto_claim.status = ClaimsApi::AutoEstablishedClaim::ESTABLISHED
       auto_claim.save
+
+      flashes = JSON.parse(form_data).dig('form526', 'veteran', 'flashes')
```
bastosmichael, line 25: We may need to eventually pull this from the model but this is okay for now.

bastosmichael (approved):
LGTM, Keep in mind that these evss_headers are used to send payloads to EVSS, let's make sure that we won't break our EVSS payload submission if we add headers that EVSS isn't expecting meant for BGS. Might want to look at whether or not we are going to break the EVSS stuff by doing this?

SlakrHakr:
> LGTM, Keep in mind that these evss_headers are used to send payloads to EVSS, let's make sure that we won't break our EVSS payload submission if we add headers that EVSS isn't expecting meant for BGS. Might want to look at whether or not we are going to break the EVSS stuff by doing this?

I poked around at where these headers are used for EVSS and every place i tracked down had each header key specifically used, which made me feel good about tacking on headers. The fact that these headers already had a few headers tacked on to the base list (even if they were also EVSS) made me feel pretty solid about tacking my own BGS ones on there as well. I'll doublecheck a few things and comment back here before merging this in though :+1:

SlakrHakr:
Ok, doublechecked and I don't see anything that should be inadvertently affected from the addition to the auth_headers hash. Going to go forward with the merge.

kfrz (approved):
This looks good to me.

PR #5285, opened 2020-11-13
API-3366: Need to filter 526 and 2122 find_by methods using the source label

SlakrHakr:
## Description of change
Protect api from invalid source requesting information solely by id.

## Original issue(s)
https://vajira.max.gov/browse/API-3366

Spec tests.

### DIFF
--- a/modules/claims_api/app/controllers/claims_api/application_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/application_controller.rb
@@ -26,10 +26,21 @@ module ClaimsApi
       Settings.oidc.isolated_audience.claims
     end
 
+    protected
+
+    def source_name
+      if v0?
+        request.headers['X-Consumer-Username']
+      else
+        user = header_request? ? @current_user : target_veteran
+        "#{user.first_name} #{user.last_name}"
+      end
+    end
+
     private
 
     def find_claim
-      claim = ClaimsApi::AutoEstablishedClaim.find_by(id: params[:id])
+      claim = ClaimsApi::AutoEstablishedClaim.find_by(id: params[:id], source: source_name)
 
       if claim && claim.status == 'errored'
         fetch_errored(claim)
@@ -37,6 +48,7 @@ module ClaimsApi
         render json: claim, serializer: ClaimsApi::AutoEstablishedClaimSerializer
       else
         evss_claim = claims_service.update_from_remote(claim.try(:evss_id) || params[:id])
+        # Note: source doesn't seem to be accessible within a remote evss_claim
         render json: evss_claim, serializer: ClaimsApi::ClaimDetailSerializer
       end
     end
--- a/modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
@@ -17,7 +17,8 @@ module ClaimsApi
         FORM_NUMBER = '2122'
 
         def submit_form_2122
-          power_of_attorney = ClaimsApi::PowerOfAttorney.find_by(header_md5: header_md5)
+          power_of_attorney = ClaimsApi::PowerOfAttorney.find_using_identifier_and_source(header_md5: header_md5,
+                                                                                          source_name: source_name)
           unless power_of_attorney&.status&.in?(%w[submitted pending])
             power_of_attorney = ClaimsApi::PowerOfAttorney.create(
               status: ClaimsApi::PowerOfAttorney::PENDING,
@@ -62,7 +63,8 @@ module ClaimsApi
         end
 
         def active
-          power_of_attorney = ClaimsApi::PowerOfAttorney.find_by(header_md5: header_md5)
+          power_of_attorney = ClaimsApi::PowerOfAttorney.find_using_identifier_and_source(header_md5: header_md5,
+                                                                                          source_name: source_name)
           if power_of_attorney
             render json: power_of_attorney, serializer: ClaimsApi::PowerOfAttorneySerializer
           else
@@ -94,11 +96,6 @@ module ClaimsApi
             icn: current_user.icn,
             email: current_user.email
           }
-        end
-
-        def source_name
-          user = header_request? ? @current_user : target_veteran
-          "#{user.first_name} #{user.last_name}"
         end
 
         def find_poa_by_id
--- a/modules/claims_api/app/models/claims_api/power_of_attorney.rb
+++ b/modules/claims_api/app/models/claims_api/power_of_attorney.rb
@@ -18,6 +18,17 @@ module ClaimsApi
     before_validation :set_md5
     validates :md5, uniqueness: true
 
+    def self.find_using_identifier_and_source(source_name:, id: nil, header_md5: nil, md5: nil)
+      primary_identifier = {}
+      primary_identifier[:id] = id if id.present?
+      primary_identifier[:header_md5] = header_md5 if header_md5.present?
+      primary_identifier[:md5] = md5 if md5.present?
+      poa = ClaimsApi::PowerOfAttorney.find_by(primary_identifier)
+      return nil if poa.present? && poa.source_data['name'] != source_name
+
+      poa
+    end
+
     def sign_pdf
       signatures = convert_signatures_to_images
       page_1_path = insert_signatures(1, signatures[:veteran], signatures[:representative])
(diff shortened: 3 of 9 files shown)


modules/claims_api/spec/requests/v0/claims_request_spec.rb
```diff
@@ -52,6 +52,7 @@
       it 'shows a single errored Claim with an error message', run_at: 'Wed, 13 Dec 2017 03:28:23 GMT' do
         create(:auto_established_claim,
                auth_headers: { some: 'data' },
+               source: 'TestConsumer',
```
kfrz, line 55: Should this be consistent throughout the specs, some places have "TestConsumer" and others have "Abe Lincoln" -- is this necessitated by the spec or arbitrary? Want to avoid the mystery guest smell is all

modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
```diff
@@ -16,8 +16,9 @@ class PowerOfAttorneyController < ClaimsApi::BaseFormController
 
         FORM_NUMBER = '2122'
 
-        def submit_form_2122
+        def submit_form_2122 # rubocop:disable Metrics/CyclomaticComplexity
```
kfrz, line 19: I think it's an anti-pattern to increase the complexity in controller here. The controller action should be responsible for routing requests and responding to the requests. `#submit_form_2122` is doing a bit more than that it seems. 

I'd suggest a `FormSubmissionService` which could be used to capture the Poa lookup branching logic? 

Reducing cyclomatic complexity and increasing readability would be the goals of that change.

kfrz: Curious: I'm making the assumption that `Form2122` is a form for modifying/assigning power of attorney for a given user - is this accurate?

modules/claims_api/app/controllers/claims_api/application_controller.rb
```diff
@@ -26,17 +26,29 @@ def fetch_aud
...
+
+    def source_name
+      if v0?
+        request.headers['X-Consumer-Username']
+      else
+        user = header_request? ? @current_user : target_veteran
+        "#{user.first_name} #{user.last_name}"
+      end
+    end
+
```
kfrz, line 39: Is this temporary branching logic? Should we handle versioning concerns in a different namespace - i.e. having parallel implementations rather than branching logic inside the base controller?

kfrz (changes requested):
I have a few comments - would like to see less in the actual controller action and more in a service class, "skinny controller" style, if possible.

kfrz (approved):

modules/claims_api/spec/requests/v0/claims_request_spec.rb
SlakrHakr: its arbitrary. it only matters here because this change requires source of the requester to be the same as the source that created the claim initially.

source was already these different values for different specs, but i can try to sync them up if thats worthwhile. i could see some small benefit to leaving it somewhat disparate so that the "test" of the requester source matching the creator source is a bit less likely to be broken but magically still work. but i can definitely see where you're coming from with the code smell.

happy to do whatever here

modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
SlakrHakr: > I'm making the assumption that Form2122 is a form for modifying/assigning power of attorney for a given user - is this accurate?

Yes, that's correct.

modules/claims_api/app/controllers/claims_api/application_controller.rb
SlakrHakr: source_name is used within this base controller for a simple find that is apparently shared between v0 and v1
https://github.com/department-of-veterans-affairs/vets-api/blob/aab8083f1af918d91359aae8474a893d0e2358c1/modules/claims_api/app/controllers/claims_api/application_controller.rb#L43

modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
SlakrHakr: > I think it's an anti-pattern to increase the complexity in controller here.

Fair point. Ideally I would have been able to simply add source_name to the find_by here and move on, but unfortunately source_name is embedded within source_data which is encrypted in the database, and I couldn't figure out a way to query that value directly :-1: 

I liked the find_by nature of the model so I changed the implementation to this to hopefully help this read a bit better? and reduce cyclomatic complexity at the controller level. Note: rubocop wouldn't let me use the phrasing "find_by_*" because of Rails/DynamicFindBy complaints.
https://github.com/department-of-veterans-affairs/vets-api/commit/fd61ff9f543ad15110d49c05b6cca0a30f88f025

bastosmichael (changes requested):
source will need an index in the DB now that it's being used in a query fashion, we have over 100K submissions in production already, don't want look ups to be slowed down because db lookup was moving too slow, also we will need to ensure that we physically test this against the v0 end point locally at least for the POST 526 end point before merging, I trust your test suite but like to also like to hit the end point physically in case we missed anything in the test suite.

SlakrHakr:
> source will need an index in the DB now that it's being used in a query fashion

Index for source field has been added here: https://github.com/department-of-veterans-affairs/vets-api/pull/5341

SlakrHakr:
> also we will need to ensure that we physically test this against the v0 end point locally at least for the POST 526 end point before merging, I trust your test suite but like to also like to hit the end point physically in case we missed anything in the test suite.

Ok, i was finally able to get the 526 v1 submit to work locally, after fighting auth for a bit. I'm still having a bizarre issue around auth for v0 endpoints but the "find claim by id" logic is shared between v0 and v1 and works fine through v1 auth for me locally.

I can keep trying but the creates didnt really change much at all with this work, was primarily the reads, and the basic "get claim by id" seems to still be working. Let me know if thats not enough validation though and i'll keep poking at my strange v0 auth issues for this.

modules/claims_api/spec/requests/v1/claims_request_spec.rb
```diff
@@ -72,35 +72,59 @@
...
+                 auth_headers: { some: 'data' },
+                 evss_id: 600_118_851,
+                 id: 'd5536c5c-0465-4038-a368-1a9d9daf65c9')
+          VCR.use_cassette('evss/claims/claim') do
+            get(
+              '/services/claims/v1/claims/d5536c5c-0465-4038-a368-1a9d9daf65c9',
+              params: nil, headers: request_headers.merge(auth_header)
+            )
+
+            # TODO: What is the expectation at this point...?
```
hinzed1127, line 126: do we know the answer as to what to assert yet?

hinzed1127 (approved):
looks good pending that question about one of the tests

modules/claims_api/spec/requests/v1/claims_request_spec.rb
SlakrHakr: shoot! i cant believe i missed that! added an expectation here: https://github.com/department-of-veterans-affairs/vets-api/pull/5285/commits/9abce19acf521d19a60eecb97e8e7d75d35857db

modules/claims_api/app/models/claims_api/power_of_attorney.rb
```diff
@@ -18,6 +18,17 @@ class PowerOfAttorney < ApplicationRecord
     before_validation :set_md5
     validates :md5, uniqueness: true
 
+    def self.find_using_identifier_and_source(source_name:, id: nil, header_md5: nil, md5: nil)
+      primary_identifier = {}
+      primary_identifier[:id] = id if id.present?
+      primary_identifier[:header_md5] = header_md5 if header_md5.present?
+      primary_identifier[:md5] = md5 if md5.present?
+      poa = ClaimsApi::PowerOfAttorney.find_by(primary_identifier)
+      return nil if poa.present? && poa.source_data['name'] != source_name
```
bastosmichael, line 27: Why not actually search by source since you added the index specifically for including it in the find_by? Is there a particular strategy here? Including the source rules out duplicate md5's and other potential markers and I'm not saying this is a bad option to doing it this way, just want to understand the logic for not including source_name in the search itself?

SlakrHakr: I added the index to AutoEstablishedClaim which had source available as a column so was easy to just add to the find_by.
PowerOfAttorney stores the source name a bit differently. It has a column named "source_data" that is encrypted within the database. I spent a little time on it but couldn't get any sort of direct query to work but im definitely open to any ideas improving this logic, this is the best i could come up with to mimic the find_by logic youd expect.

bastosmichael (approved):

modules/claims_api/app/models/claims_api/power_of_attorney.rb
bastosmichael: Makes logical sense.

PR #5288, opened 2020-11-13
15580 update pact helper config

LindseySaari:
## Description of change
With the pact broker moving from Heroku to the AWS infrastructure, this updates the pact helper uri to point to the new broker location and updates comments/examples.  

## Things to know
Until coordination efforts can be synced between FE/BE, this PR will remain in a draft state so that the various tasks aren’t out of sync in CI. E.g. Pacts are published to the old broker, but verification and publication are against the pacts in the AWS broker. 

Parallel change to update documentation https://github.com/department-of-veterans-affairs/va.gov-team/pull/16109

## Original issue(s)
department-of-veterans-affairs/va.gov-team#15580

## Testing
- [x] The broker triggers the "contract changed" webhook when a new contract is published or updated
- [x] At least one pact has the consumer tag of master
- [x] The verification workflow in CircleCI successfully triggers when a commit is pushed or webhook is invoked
- [x] Verification results successfully publish back to the new AWS broker


kfrz (approved):
LGTM

PR #5289, opened 2020-11-13
15786 patient pgd service

dillo:
## Description of change

- The code adds the ability for the vets-api/health_quest module to connect to the MAP PGD and retrieve a patient resource by a logged-in veterans icn.
- We're adding the `fhir_client` gem to the project. This gem lets us retrieve the patient as a fhir resource and will be used to retrieve questionnaire and questionnaire response resources as well. Furthermore, this gem will be used to create questionnaire responses as resources on the MAP PGD.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#15786

- The health_quest module is still behind a feature flag. The endpoints in the module, particularly appointment related endpoints currently only return mock data. The questionnaire and questionnaire responses endpoints will also be returning mock data shortly, until all the relevant services for fetching these resources(patient, questionnaire, questionnaire responses, and appointments from MAS) are built out.
- Swagger docs have already been built out previously and there are no changes to them.

- Specs have been written for all the new files introduced.
- Manual verification of the patient service was conducted locally to ensure that the vets-api was getting the patient resource from the MAP PGD in the playground.


Gemfile.lock
```diff
@@ -143,7 +143,6 @@ PATH
 
 GEM
   remote: https://rubygems.org/
-  remote: https://enterprise.contribsys.com/
```
erluti, line 146: don't check in changes to the gemfile that remove `remote: https://enterprise.contribsys.com/` and related gems

modules/health_quest/app/services/health_quest/pgd/fhir_client.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module HealthQuest
+  module Pgd
```
erluti, line 4: Could this be spelled out as `PatientGeneratedData`?

Gemfile.lock
dillo: Fixed! Please re-review. Thanks!

modules/health_quest/app/services/health_quest/pgd/fhir_client.rb
dillo: I thought of doing that in the beginning but it's known as PGD everywhere in the domain and by everyone and acronyms seemed to be the norm for domain level objects/concepts. Is there a push to move away from this convention?

erluti: We're trying to think more about acronyms.  As we move to using the Rails 6 zeitwerk loader, acronyms have impacts on how files are loaded.  As an acronym it should be `PGD`, but if we're taking that approach there should be an entry in `config/initializers/inflections.rb`.  But since that's a universal acronym file, I'm thinking we should be more conservative on what becomes an acronym.  While PGD is known in the domain, it's less obvious from the larger `vets-api` perspective.  We're trying to get zeitwerk across the line before we formalize an approach, so feedback is welcome.  My comment came from a place where we've been running into a lot of these acronyms we're not sure which way to break on, and PGD is not in the `/wtf` slackbot, so I assumed it's not commonly used across the VA.

dillo: I think I'll go with PatientGeneratedData as you suggested. I agree that we should have a larger conversation about acronym usage in the codebase in general, and I think we should potentially involve a lot more people in that conversation. Personally, I think I'd rather avoid using acronyms in any codebase; Just like I'd avoid using a single character variable.

dillo: Done and ready for re-review!

modules/health_quest/app/services/health_quest/patient_generated_data/fhir_client.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module HealthQuest
+  module PatientGeneratedData
+    module FhirClient
```
erluti, line 5: So funnily enough, FHIR does have an inflection in the inflections file, so this should be `FHIRClient`

erluti (commented):
looks good, one more acronym issue

modules/health_quest/app/services/health_quest/patient_generated_data/fhir_client.rb
dillo: Good catch! Changing now!

dillo: Done!

erluti (approved):

PR #5290, opened 2020-11-13
16155 - Update DMC Endpoints

camerontesterman:
## Description of change
Updates DMC's endpoints to the new digital-service endpoints. Adds base config class for DMC that is used for FSR and Debts configs.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#16155

tested locally


lib/dmc/base_configuration.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module DMC
```
kfrz, line 3: What's `DMC` ? I'd prefer to see a full-on name used for the module here to avoid confusion and inflection issues. We're not short on bytes!

lib/dmc/fsr_configuration.rb
```diff
@@ -1,31 +1,11 @@
...
-    def self.base_request_headers
-      super.merge(
-        'client_id' => Settings.dmc.client_id,
-        'client_secret' => Settings.dmc.client_secret
-      )
-    end
+require 'dmc/base_configuration'
 
+module DMC
+  class FSRConfiguration < DMC::BaseConfiguration
```
kfrz, line 6: Would prefer to see `FinancialStatusReportConfiguration` to reduce confusion/inflection issues

kfrz (changes requested):
This is good structurally but I'd like to see less acronyms as we add new services, as a general rule of thumb, in order to increase immediate readability. Thoughts?

kfrz (approved):
Thanks for the changes, this looks good, once CI is green!

PR #5294, opened 2020-11-13
Api 3019 client credentials validation endpoint

dbrown-va:
## Description of change
Adds Lighthouse token validation support for Okta Client Credentials tokens.  These tokens do not contain normal parameters like uuid from Okta, so session & current_user build up has to be skipped.

## Original issue(s)
https://vajira.max.gov/browse/API-3019

* A follow on task, API-3500 will utilize a new OAuth proxy service to retrieve the launch context to populate the ICN in the token response.
* The V0 token validation endpoint is slated for removal (API-3589), but until then it has similar logic to skip the current_user.icn lookup if a client credentials token.

### DIFF
--- a/app/controllers/openid_application_controller.rb
+++ b/app/controllers/openid_application_controller.rb
@@ -33,6 +33,15 @@ class OpenidApplicationController < ApplicationController
   def authenticate_token
     return false if token.blank?
 
+    #issued for a client vs a user
+    if token.is_client_credentials_token?
+      if token.payload["scp"].include?("launch/patient")
+        # TODO fetch launch context
+        # token.payload[:"icn"] = '1234V5678'
+      end
+      return true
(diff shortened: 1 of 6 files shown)


beaugrantham-va (approved):
Confirmed via inspection:

* Token validation (structure, issuer, signing, audience) takes place before checking authorization grant (client credentials)
* Authorization grant (client credentials) confirmed using `cid == sub`
* User identifier is not set (`va_identifiers.icn = nil`)

Understood that tokens issued using the client credentials grant will _not_ populate a `current_user` or `session` or put anything into cache. Modules that expect this to be populated _can not_ use client credentials as is. This is enforced implicitly by audience validation + Okta configuration.

esubocz-va (approved):

PR #5295, opened 2020-11-13
Tweak errors middleware

grossb1:
## Description of change

The main goal of this story was to provide more info on VAOS errors so users could be better equipped when reviewing reports in Sentry.  In addition, the Appointment#Cancel error which is currently mapped to a 500 was mapped to a 400 to remove clutter from both Sentry and Grafana.  This Appointment#cancel error is slated to return a 400 on the VAMF side at which point we will remove this code.

## Original issue(s)
https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/15012

Unit tests were created to test the associated code.


modules/vaos/app/services/vaos/middleware/response/errors.rb
```diff
@@ -9,49 +9,53 @@ class Errors < Faraday::Response::Middleware
         def on_complete(env)
           return if env.success?
 
-          Raven.extra_context(vamf_status: env.status, vamf_body: env.body, vamf_url: env.url)
+          @status = env.status
```
erluti, line 12: since these aren't used as accessors, could these values be passed into the methods that use them instead of being part of the state?

modules/vaos/spec/services/middleware/vaos_errors_spec.rb
```diff
@@ -0,0 +1,156 @@
+# frozen_string_literal: true
```
erluti, line 1: is there a reason this file path doesn't parallel the class's file path?

modules/vaos/app/services/vaos/middleware/response/errors.rb
saneshark: it was a lot of passing around, yes, we could switch to that but I thought this looked a little cleaner. We could also make them accessors but does that violate some rule in ruby? I don't know if it does or not but `on_complete` is not really the same thing as `initialize` so maybe to your point passing them as reference is the way to go.

modules/vaos/spec/services/middleware/vaos_errors_spec.rb
saneshark: i think we just did what worked, no reason. What would you recommend we move it to?

modules/vaos/app/services/vaos/middleware/response/errors.rb
grossb1: I'm a beginner at Ruby, but in our case, I don't think that's going to work.  We have an additional method that uses those accessors (I call them class member variables).
```
def response_values
   {
     detail: detail,
     source: { vamf_url: @url, vamf_body: @body, vamf_status: @status }
   }
end
```

If we were to inject `env` directly into that method, we would have to update the calls `response_values` to pass that as well.  Which we could do, if that's what you're recommending.

grossb1: nm, it's a private method

modules/vaos/app/services/vaos/middleware/response/errors.rb
```diff
@@ -12,38 +12,38 @@ def on_complete(env)
           Raven.extra_context(vamf_status: env.status, vamf_body: env.body, vamf_url: env.url)
           case env.status
           when 400, 409
-            error_400(env.body)
```
saneshark, line 15: @grossb1 now that my PR has been merged, we can probably try to handle the 409 status code using exception.en.yml

The changes we're waiting on from VAOS BE haven't hit production yet, but I think we could still make the change so that it works as expected when those changes go live.

grossb1: I included the condition for 409s

Gemfile.lock
```diff
@@ -151,7 +151,6 @@ PATH
 
 GEM
   remote: https://rubygems.org/
-  remote: https://enterprise.contribsys.com/
```
erluti, line 154: don't remove the enterprise sidekiq-related gems

erluti (commented):
looks like this isn't ready for review yet

Gemfile.lock
saneshark: I am not sure why these changes to Dangerbot and Gemfile were needed as part of this PR anyway. 

`git checkout Gemfile.lock -- master`
`git checkout Jenkinsfile -- master`

modules/vaos/app/exceptions/vaos_backend_service_exception.rb
```diff
@@ -0,0 +1,51 @@
...
+        {
+          detail: detail(@env.body),
+          source: { vamf_url: @env.url, vamf_body: @env.body, vamf_status: @env.status }
+        }
+      end
+
+      private
+
+      def lookup_key(status)
+        VAOS_ERRORS.select { |status_code| status_code === status }.values.first
```
saneshark, line 35: NITPICK: Algorithmically speaking this isn't particularly performant and begs the question why use a Hash at all?

Hash lookup is O(n) if the key is known, which it is in most of your cases. A more performant algorithm would take advantage of this fact, and have an else clause for when everything else. But the way you're using it here, it's not really any more performant than having a Case When Else clause (which would actually be a bit more readable and idiomatic).

saneshark: Also, what happens if a status code of 600 is returned? Shouldn't that be interpreted as a 500 too?

modules/vaos/spec/request/appointments_request_spec.rb
```diff
@@ -339,38 +340,24 @@
...
-            expect(Rails.logger).to receive(:warn).with('VAOS service call failed!', any_args)
+          VCR.use_cassette('vaos/appointments/put_cancel_appointment_409', match_requests_on: %i[method uri]) do
+            expect(Rails.logger).to receive(:warn).with('VAOS service call failed!', any_args).once
             expect(Rails.logger).to receive(:warn).with(
               'Clinic does not support VAOS appointment cancel',
               clinic_id: request_body[:clinic_id],
               site_code: request_body[:facility_id]
-            )
+            ).once
+            # We're checking that the logger is called twice to account for behavior in sentry_logging.rb lines 25-30
```
saneshark, line 350: Someone on VSP should be made aware of this. There is a lot of redundant logging happening there in SentryLogging as a result of that issue. Did someone really introduce an else clause there just to get "punctuation" into the logs? 🤔 

@erluti

saneshark:
overall this looks good to me except for the Nitpick comment that might be worth a closer look.

modules/vaos/app/exceptions/vaos_backend_service_exception.rb
grossb1: I think O(n) would represent a scan of the entire hash, I could be wrong here but on average it should be O(1).  That being said it could be different in Ruby.  In either case, the cardinality of the set we're interested in (i.e. the set of http status codes) is fairly small.  In terms of readability I would have to agree, I think the hash table is a bit more concise though.  Also any unrecognized status coded will get mapped to a VA900 response code.  In those cases, you are correct the complexity of those look-ups is O(n), but in our case n=15, so we're talking picoseconds :).

saneshark: And yes, I meant to say that hash lookup is O(1), and that what you've got is essentially O(n). 

You can make it a little bit faster by doing a #find or #detect instead of #select. This will return the first value match as opposed to #select which will iterate across the entire hash even when a match has been found, returning all of them. With #find or #detect you wouldn't need to do #values.first and could instead do #.last since it returns the key value pair as an array.

grossb1: I got rid of the `select`

saneshark (approved):
LGTM

modules/vaos/spec/request/appointments_request_spec.rb
erluti: Looking at that code it looks like `Common::Exceptions::BackendServiceException` is being treated as more fully featured.  Although line 30 would add the backtrace a second time.

i'm not clear on how this code comment relates to that, though?

@kfrz i see you put 👀 here, do you have any insight?

saneshark: We put this code comment in here because once the changes to SentryLogging are made to make it not log the same thing twice for no good reason, this spec would be broken -- the thought was that whomever makes those fixes would see the comment and feel safe enough to fix the spec to be `.once` instead of `.twice` without further inquiry.

erluti: We're talking about this code?
```
if exception.is_a? Common::Exceptions::BackendServiceException
      rails_logger(level, exception.message, exception.errors, exception.backtrace)
    else
      rails_logger(level, "#{exception.message}.")
    end
    rails_logger(level, exception.backtrace.join("\n")) unless exception.backtrace.nil?
 ```
 Looks like line 30 logs the backtrace a second time if it is a `Common::Exceptions::BackendServiceException` and the exception has errors and a backtrace.  I see your concern now.

erluti: Ticket https://app.zenhub.com/workspaces/vsp-5cedc9cce6e3335dc5a49fc4/issues/department-of-veterans-affairs/va.gov-team/18076

erluti:
Pipeline won't pass until you pull in the latest master

grossb1:
Sorry for the squash and merge noise.

PR #5300, opened 2020-11-13
VBA should be inflected

ericboehs:


ericboehs:
Lindsay beat me to it!

PR #5308, opened 2020-11-16
Reinstate PDF validation for Claims endpoints

hinzed1127:
## Description of change

This adds PDF validation back into claims. Previously it was only getting run for document validation. This also adds it to the `upload_526` method, as well.

## Original issue(s)


modules/claims_api/app/controllers/concerns/claims_api/document_validations.rb
```diff
@@ -32,10 +32,12 @@ def document_content_type_errors
         end
       end
 
+      # rubocop:disable Layout/LineLength
       def pdf?(document)
         extension = document.original_filename.split('.').last
-        ['application/pdf', 'text/plain'].include?(document.content_type) && extension.downcase == 'pdf'
+        ['application/pdf', 'text/plain', 'application/octet-stream'].include?(document.content_type) && extension.downcase == 'pdf'
```
hinzed1127, line 38: reason for this addition is a ruby quirk. basically if `Content-Type` is unspecified, it defaults to `application/octet-stream`. I think if the extension is still `pdf` this is a safe thing to allow

modules/claims_api/app/controllers/concerns/claims_api/document_validations.rb
```diff
@@ -32,10 +32,12 @@ def document_content_type_errors
         end
       end
 
+      # rubocop:disable Layout/LineLength
```
SlakrHakr, line 35: can we not just bump the second clause down to the next line to get around the LineLength cop?

SlakrHakr (approved):

modules/claims_api/app/controllers/concerns/claims_api/document_validations.rb
hinzed1127: good point, I'll make that update before merging

bastosmichael (approved):
Good job checking each other's PR's gents.

PR #5312, opened 2020-11-16
API-3609: Flashes - Assuming Homelessness/Terminally Ill flash from their sections of the 526 payload

SlakrHakr:
## Description of change
Automatically adds "Homeless" and/or "Terminally Ill" flashes to claim if legacy content regarding these items was sent in the payload.

## Original issue(s)
https://vajira.max.gov/browse/API-3609

Will most likely have to be tweaked within this ticket: https://vajira.max.gov/browse/API-3260

Existing spec tests

### DIFF
--- a/modules/claims_api/app/models/claims_api/auto_established_claim.rb
+++ b/modules/claims_api/app/models/claims_api/auto_established_claim.rb
@@ -88,7 +88,7 @@ module ClaimsApi
       initial_flashes.push('Homeless') if homelessness.present?
       initial_flashes.push('Terminally Ill') if is_terminally_ill.present? && is_terminally_ill
 
-      initial_flashes.uniq
+      initial_flashes.present? ? initial_flashes.uniq : []
     end
 
     private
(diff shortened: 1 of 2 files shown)


modules/claims_api/app/models/claims_api/auto_established_claim.rb
```diff
@@ -80,6 +80,17 @@ def uploader
       @uploader ||= ClaimsApi::SupportingDocumentUploader.new(id)
     end
 
+    def flashes
+      initial_flashes = JSON.parse(to_internal).dig('form526', 'veteran', 'flashes')
+      homelessness = JSON.parse(to_internal).dig('form526', 'veteran', 'homelessness')
```
bastosmichael, line 85: If you could just add a spec test for this logic we should be in business, otherwise looks good.

bastosmichael (changes requested):
Just needs a spec test and we should be in business.

bastosmichael (approved):

modules/claims_api/app/models/claims_api/auto_established_claim.rb
bastosmichael: I just now saw your comment about spec tweaking

PR #5317, opened 2020-11-16
Correctly Identify user for Sentry

annaswims:
## Description of change
Sentry expects id, not uuid. https://docs.sentry.io/platforms/ruby/enriching-events/identify-user/

The user's uuid is meaningless on its own but it rather useful in tracking down what the issue is in cases like a validation error where data is relevant.   We've been sending it all along, but not with the proper hash key. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#16216


kfrz (approved):
Looks fine to me - nice catch!

PR #5319, opened 2020-11-17
16056 mobile appointments va adapter

kreek:
## Description of change
Adapter for Mobile VA Appointments. Takes in a list of appointments of various types/schemas and returns a consistent schema.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#16056

*  In the get appointments pipeline this is step 2(a):
1: service -> **2: adapters (a: VA b: CC)** -> 3: models 4: serializer


modules/mobile/app/models/mobile/v0/adapters/va_appointments.rb
```diff
@@ -0,0 +1,209 @@
...
+          location[:url] = details['providers'].first.dig('virtualMeetingRoom', 'url')
+          location[:code] = details['providers'].first.dig('virtualMeetingRoom', 'pin')
+          location
+        end
+
+        def get_time_zone(facility_id)
+          facility = Mobile::VA_FACILITY_TIMEZONES["dfn-#{facility_id}"]
+          return nil unless facility
+
+          facility[:time_zone] if facility
```
jjulian, line 204:
```suggestion
          facility = Mobile::VA_FACILITY_TIMEZONES["dfn-#{facility_id}"]
          facility ? facility[:time_zone] : nil
```

jjulian:
What is `VA_FACILITY_TIMEZONES` all about? Is that data in code because it's not available elsewhere? Is this a "cache" from an upstream system? Will the data ever change? etc. A comment in that file could help explain.

kreek:
> What is `VA_FACILITY_TIMEZONES` all about? Is that data in code because it's not available elsewhere? Is this a "cache" from an upstream system? Will the data ever change? etc. A comment in that file could help explain.

@jjulian cc @jperk51 added a comment in a [separate PR](https://github.com/department-of-veterans-affairs/vets-api/pull/5324/files) (to reduce this PR's size). It's not available elsewhere and the va.gov client has a similar JSON file. Please review that PR first then come back to this one.

modules/mobile/app/models/mobile/v0/adapters/va_appointments.rb
```diff
@@ -0,0 +1,207 @@
...
+        ].freeze
+
+        STATUSES = {
+          booked: 'BOOKED',
+          cancelled: 'CANCELLED'
+        }.freeze
+
+        VIDEO_GFE_FLAG = 'MOBILE_GFE'
+
+        def parse(appointments)
```
kfrz, line 51: Please add class-level and public method docstrings via YARDoc

kfrz (approved):
Nicely done. Only one comment - I'd like to see class-level and public method YARDocs added whenever we can. Won't block approval for this though!

jjulian (approved):
LGTM. There is a ton of domain knowledge in here!

PR #5322, opened 2020-11-17
remove use of awrence gem

erluti:
[This gem](https://github.com/futurechimp/awrence) appears to only be used in two places, is for testing only, and there's an alternative using rails alone.


kfrz (approved):
Cleaning up tech debt. Awwwright. Nice!

PR #5324, opened 2020-11-17
16056  VA facilities names and timezones for mobile appointments

kreek:
## Description of change
Splitting [PR 5319](https://github.com/department-of-veterans-affairs/vets-api/pull/5319) into two to reduce the line size. The timezone data is not available via API and the va.gov client, which does the data transformation on the FE, has a similar [json file](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/applications/vaos/utils/timezones.json). This set of data uses a Hash (Dictionary) by facility id rather than array for faster lookup and we don't need `currentTZ` as Rails can figure that out dynamically - e.g. it's not PST year round it's also PDT so that should not be hard coded - we can also go with just 'Pacific' if switching between PST and PDT is confusing to users.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#16056 

* Should be merged first before [PR 5319](https://github.com/department-of-veterans-affairs/vets-api/pull/5319)

### DIFF
--- a/modules/mobile/config/initializers/va_facilities.rb
+++ b/modules/mobile/config/initializers/va_facilities.rb
@@ -1,0 +1,556 @@
+# frozen_string_literal: true
+
+Mobile::VA_FACILITY_TIMEZONES = {
+  'dfn-358' => {
+    name: 'MANILA VAMC',
+    time_zone: 'Asia/Manila'
+  },
+  'dfn-519' => {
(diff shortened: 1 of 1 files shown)


jjulian (approved):
Cool, mirroring the VA.gov FE implementation. This could be a candidate for sharing via vets-api.

PR #5328, opened 2020-11-17
ignore whitespace in Dangerbot and change LOC maximums by 20%

thilton-oddball:
## Description of change
Altered Dangerfile to ignore whitespace when determining lines of code (LOC) since this behavior is not available with Danger. Also reduced the LOC maximums by 20% since whitespace will no longer be a factor.

## Original issue(s)
department-of-veterans-affairs/va.gov-team/15233


Dangerfile
```diff
@@ -1,7 +1,7 @@
 # Warn if a pull request is too big
 PR_SIZE = {
-  RECOMMENDED_MAXIMUM: 250,
-  ABSOLUTE_MAXIMUM:    1000,
+  RECOMMENDED_MAXIMUM: 200,
+  ABSOLUTE_MAXIMUM:    800,
```
erluti, line 4: I think the RECOMMENDED_MAXIMUM of 200 makes sense, but i think without whitespace counting we should push the ABSOLUTE_MAXIMUM even lower.  I don't think I've seen many PRs approach the 1k that didn't need to be broken up, and without whitespace changes I bet it could be a lower maximum.  But I think I guess I also think we'll get data faster if we set it too low and have to increase it.
I would put it at 500, but I'm going to solicit some more feedback.

Dangerfile
```diff
@@ -10,7 +10,16 @@ changed_files = git.diff.stats[:files]
 
 excluded_changed_files = changed_files.select { |key| EXCLUSIONS.any? { |exclusion| key.include?(exclusion) } }
 filtered_changed_files = changed_files.reject { |key| EXCLUSIONS.any? { |exclusion| key.include?(exclusion) } }
-lines_of_code = filtered_changed_files.sum { |_file, changes| (changes[:insertions] + changes[:deletions]) }
+
+# ignores whitespace for the purpose of determining lines of code changed
+changes = `git diff -w --stat`.split("\n")
```
erluti, line 15: i think the message output by dangerbot below should also mention it's not counting whitespace changes.

Dangerfile
kfrz: I think 500 is good middle ground but without taking a look at historical danger-warnings it's sort of guess-n-check anyhow.

kfrz (approved):
LGTM. Will approve so you can merge but feel free to wait for additional feedback from the team.

erluti (approved):
LGTM
gotta make sure we keep up with PRs after this is merged incase we end up blocking PRs regularly

PR #5329, opened 2020-11-17
Supporting Documents for 686c

kathleencrawford:
## Description of change

When a dependency claim is submitted (VA Form 686c), sometimes attachments are required as "supporting evidence" that are part of the claim submission.  Examples of this include marriage certificate, birth certificate, etc.  This PR takes the documents that have been submitted with the claim, and uploads them to VBMS.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#15129

The form submission process was already generating a PDF and submitting to VBMS, so no additional configuration or logging was required.  This PR just adds to that existing process and uploads any additional documents if they are present.
 

## Testing
- [x] Local testing completed
- [x] Testing on staging planned

### DIFF
--- a/app/models/saved_claim/dependency_claim.rb
+++ b/app/models/saved_claim/dependency_claim.rb
@@ -90,7 +90,7 @@ class SavedClaim::DependencyClaim < SavedClaim
     end
   end
 
-  def upload_to_vbms(path, doc_type = '148')
+  def upload_to_vbms(path, doc_type: '148')
     uploader = ClaimsApi::VBMSUploader.new(
       filepath: path,
       file_number: parsed_form['veteran_information']['ssn'],
--- a/app/models/saved_claim/dependency_claim.rb
+++ b/app/models/saved_claim/dependency_claim.rb
@@ -81,6 +81,25 @@ class SavedClaim::DependencyClaim < SavedClaim
     []
   end
 
+  # Run after a claim is saved, this processes any files that are present
+  def process_attachments!
+    supporting_documents = parsed_form.dig('dependents_application', 'supporting_documents')
+    if supporting_documents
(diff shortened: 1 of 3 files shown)


app/models/saved_claim/dependency_claim.rb
```diff
@@ -81,23 +81,32 @@ def regional_office
...
+  def process_attachments!
+    supporting_documents = parsed_form.dig('dependents_application', 'supporting_documents')
+    if supporting_documents
+      files = PersistentAttachment.where(guid: supporting_documents.map { |doc| doc['confirmation_code'] })
+      files.find_each { |f| f.update(saved_claim_id: id) }
+    end
   end
 
-  def upload_to_vbms(path)
+  def upload_to_vbms(path, doc_type = '148')
```
kfrz, line 93:
```suggestion
  def upload_to_vbms(path, doc_type: '148')
```

Prefer keyword args if possible - this change might require updating any calls to the `#upload_to_vbms` method

kfrz (changes requested):
One minor change, prefer keyword arguments for code clarity and maintainability unless there's a solid argument to maintain terseness offered by positional arguments

app/models/saved_claim/dependency_claim.rb
kathleencrawford: @kfrz I've updated this as suggested, thank you :)  It is a helpful suggestion, I prefer this way better too.  I didn't have to update any calls to this method because currently we are just using the default doc type.  The change to look up and pass in different doc types will be coming in a future PR.  Thank you for pointing that out though!

thilton-oddball (approved):
Looks good

kfrz (approved):

kathleencrawford:
Thanks @thilton-oddball & @kfrz !

PR #5330, opened 2020-11-17
15007 get questionnaire responses from pgd

dillo:
## Description of change

- Adding the `/health_quest/v0/questionnaire_responses` route to the health_quest module to get all QuestionnaireResponses for a logged-in user.
- The QuestionnaireResponse Service communicates with the PGD FHIR service and GETs all of a patients QuestionnaireResponse resources in a Bundle.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#15007

- Tested locally by visiting http://localhost:3000/health_quest/v0/questionnaire_responses
- RSpec tests written

### DIFF
--- a/modules/health_quest/app/controllers/health_quest/v0/questionnaire_responses_controller.rb
+++ b/modules/health_quest/app/controllers/health_quest/v0/questionnaire_responses_controller.rb
@@ -3,6 +3,12 @@
 module HealthQuest
   module V0
     class QuestionnaireResponsesController < HealthQuest::V0::BaseController
+      def index
+        factory = HealthQuest::PatientGeneratedData::QuestionnaireResponse::Factory.new(current_user)
+
+        render json: factory.all.response[:body]
+      end
+
       def show
         head :ok
       end
--- a/modules/health_quest/app/controllers/health_quest/v0/questionnaire_responses_controller.rb
+++ b/modules/health_quest/app/controllers/health_quest/v0/questionnaire_responses_controller.rb
@@ -4,7 +4,7 @@ module HealthQuest
   module V0
     class QuestionnaireResponsesController < HealthQuest::V0::BaseController
       def index
-        factory = HealthQuest::PatientGeneratedData::QuestionnaireResponse::Factory.new(current_user)
+        factory = HealthQuest::PatientGeneratedData::QuestionnaireResponse::Factory.manufacture(current_user)
 
         render json: factory.all.response[:body]
       end
--- a/modules/health_quest/app/controllers/health_quest/v0/questionnaire_responses_controller.rb
+++ b/modules/health_quest/app/controllers/health_quest/v0/questionnaire_responses_controller.rb
@@ -2,7 +2,7 @@
 
 module HealthQuest
   module V0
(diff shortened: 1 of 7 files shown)


modules/health_quest/app/services/health_quest/patient_generated_data/questionnaire_response/factory.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module HealthQuest
+  module PatientGeneratedData
+    module QuestionnaireResponse
+      class Factory
```
kfrz, line 6: I like the Factory pattern, though it should be noted this implementation *does* collide conceptually with the existing use of "factories" in the sense of `spec/factories` -> I'd be interested in your thoughts on this overall?

Does `QuestionnaireResponse` need module-level capabilities or could there be a manufacture method on that class instead of having a dedicated factory class?

kfrz (approved):
Overall this looks good and I will approve so as to not block merging, **_but I am curious_** about your thoughts on the potential for conceptual clashes between `spec/factories` and your use of the Factory pattern here.

modules/health_quest/app/controllers/health_quest/v0/questionnaire_responses_controller.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module HealthQuest
+  module V0
+    class QuestionnaireResponsesController < HealthQuest::V0::BaseController
+      def index
```
erluti, line 6:
```suggestion
      def index
        session_service = HealthQuest::SessionService.new(user)
        map_query = HealthQuest::PatientGeneratedData::QuestionnaireResponse::MapQuery.build(session_service.headers)
        render json: map_query.get(author: user.icn)
      end
```

Any reason this uses a class?  I don't see it reused anywhere.

modules/health_quest/app/services/health_quest/patient_generated_data/questionnaire_response/factory.rb
dillo: Hi, the Factory class is intended to be used as a "service" level object which combines responsibilities that I feel don't belong in the controller from a design perspective. I feel it's different from the ideas behind `spec/factories`. The Factory object is not "set in stone" by any means. Currently, it makes testing smaller and cleaner and it's managed to make it easier to implement single responsibility and open-close principles across the code layer. 

I don't wish to make a call one way or the other just quite yet because I still have a couple more services(talking to PGD) I need to code and implement and I think keeping things loosely coupled as is might be a better call than settling for the wrong abstraction right at this moment. 

In other words, we will be creating a PR refactoring once all PGD endpoints are covered and we hope to get your input and approval on the final product.

modules/health_quest/app/controllers/health_quest/v0/questionnaire_responses_controller.rb
dillo: Hi @erluti, thank you for your much-valued suggestion. Please see https://github.com/department-of-veterans-affairs/vets-api/pull/5330#discussion_r526341365

PR #5333, opened 2020-11-18
Add flashes to the AutoEstablishedClaim model

hinzed1127:
## Description of change

This adds flashes as a property to the the `AutoEstablishedClaim` model. Dependent on merging https://github.com/department-of-veterans-affairs/vets-api/pull/5348 first for migrations to run.

## Original issue(s)
https://vajira.max.gov/browse/API-3260


modules/claims_api/spec/requests/v0/disability_compensation_request_spec.rb
```diff
@@ -51,6 +51,13 @@
       expect(aec.source).to eq('TestConsumer')
     end
 
+    it 'sets the flashes' do
+      post path, params: data, headers: headers
+      token = JSON.parse(response.body)['data']['attributes']['token']
+      aec = ClaimsApi::AutoEstablishedClaim.find(token)
+      expect(aec.flashes).to eq(%w[Hardship POW])
```
SlakrHakr, line 58: Looks like this should equal
```
%w[Hardship Homeless]
```
instead here

modules/claims_api/spec/requests/v1/disability_compensation_request_spec.rb
```diff
@@ -64,6 +64,17 @@
       end
     end
 
+    it 'sets the flashes' do
+      with_okta_user(scopes) do |auth_header|
+        VCR.use_cassette('evss/claims/claims') do
+          post path, params: data, headers: headers.merge(auth_header)
+          token = JSON.parse(response.body)['data']['attributes']['token']
+          aec = ClaimsApi::AutoEstablishedClaim.find(token)
+          expect(aec.flashes).to eq(%w[Hardship POW])
```
SlakrHakr, line 73: Same here it looks like
https://github.com/department-of-veterans-affairs/vets-api/pull/5333/files#r525824863

modules/claims_api/app/models/claims_api/auto_established_claim.rb
```diff
@@ -8,6 +8,7 @@ class AutoEstablishedClaim < ApplicationRecord
     attr_encrypted(:form_data, key: Settings.db_encryption_key, marshal: true, marshaler: JsonMarshal::Marshaller)
     attr_encrypted(:auth_headers, key: Settings.db_encryption_key, marshal: true, marshaler: JsonMarshal::Marshaller)
     attr_encrypted(:evss_response, key: Settings.db_encryption_key, marshal: true, marshaler: JsonMarshal::Marshaller)
+    attribute :flashes
```
SlakrHakr, line 11: Ok, looks like this line is causing your problem regarding this error:
```
*** TypeError Exception: can't cast Array
```
Just need to remove it.

I commented out the flashes setter in the create and tossed a debug point down right after the create in the controller. That's when I noticed `auto_claim.flashes` was returning `"{ }"` instead of an array as expected. I took a look at the "signature_errors" field on the power of attorney model since it was pretty much identical to what we are trying to achieve here and confirmed that `"{ }"` was not what we should be seeing. Doublechecked the power of attorney model and sure enough it didnt have an attribute line for signature_errors.

Hope that helps!

hinzed1127:
https://github.com/department-of-veterans-affairs/vets-api/pull/5348 opened for the migration changes

modules/claims_api/app/controllers/claims_api/v0/forms/disability_compensation_controller.rb
```diff
@@ -23,6 +23,7 @@ def submit_form_526
             status: ClaimsApi::AutoEstablishedClaim::PENDING,
             auth_headers: auth_headers,
             form_data: form_attributes,
+            flashes: form_attributes.dig('veteran', 'flashes') || [],
```
SlakrHakr, line 26: This logic is shared between v0 and v1 so makes sense to have a "flashes" method in a shared parent class of these 2.
Breaking out into a method also makes sense because this logic needs to be a bit more complicated based on what is in master currently:
https://github.com/department-of-veterans-affairs/vets-api/blob/4150d82176fda1dd3f42b3687bf534c2cbf8fd96/modules/claims_api/app/models/claims_api/auto_established_claim.rb#L85

Should hopefully just be a matter of moving that flashes method out of the model and into the controller.

SlakrHakr (changes requested):
Just https://github.com/department-of-veterans-affairs/vets-api/pull/5333/files#r526565211 and only because the logic has changed since this ticket was started. Hopefully not too terrible of a change :+1:

SlakrHakr (approved):

PR #5334, opened 2020-11-18
API-3261: Flashes - Track the successes for Flashes by claim

SlakrHakr:
## Description of change
Persists any responses from bgs if an exception occurs.

## Original issue(s)
https://vajira.max.gov/browse/API-3261

Requires database migration work found here: https://github.com/department-of-veterans-affairs/vets-api/pull/5335

Specs updated and added.

### DIFF
--- a/modules/claims_api/app/models/claims_api/auto_established_claim.rb
+++ b/modules/claims_api/app/models/claims_api/auto_established_claim.rb
@@ -8,6 +8,9 @@ module ClaimsApi
     attr_encrypted(:form_data, key: Settings.db_encryption_key, marshal: true, marshaler: JsonMarshal::Marshaller)
     attr_encrypted(:auth_headers, key: Settings.db_encryption_key, marshal: true, marshaler: JsonMarshal::Marshaller)
     attr_encrypted(:evss_response, key: Settings.db_encryption_key, marshal: true, marshaler: JsonMarshal::Marshaller)
+    attr_encrypted(:bgs_flash_responses, key: Settings.db_encryption_key,
+                                         marshal: true,
+                                         marshaler: JsonMarshal::Marshaller)
 
     after_create :log_flashes
 
--- a/modules/claims_api/app/workers/claims_api/flash_updater.rb
+++ b/modules/claims_api/app/workers/claims_api/flash_updater.rb
@@ -8,13 +8,19 @@ module ClaimsApi
     include Sidekiq::Worker
     include SentryLogging
 
-    def perform(user, flashes)
+    def perform(user, flashes, auto_claim_id: nil)
       service = bgs_service(user).claimant
 
       flashes.each do |flash_name|
         # Note: Assumption that duplicate flashes are ignored when submitted
         service.add_flash(file_number: user.ssn, flash_name: flash_name)
       rescue BGS::ShareError, BGS::PublicError => e
+        if auto_claim_id.present?
+          auto_claim = ClaimsApi::AutoEstablishedClaim.find(auto_claim_id)
+          auto_claim.bgs_flash_responses = [] if auto_claim.bgs_flash_responses.blank?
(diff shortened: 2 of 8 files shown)


modules/claims_api/app/models/claims_api/auto_established_claim.rb
```diff
@@ -8,6 +8,7 @@ class AutoEstablishedClaim < ApplicationRecord
     attr_encrypted(:form_data, key: Settings.db_encryption_key, marshal: true, marshaler: JsonMarshal::Marshaller)
     attr_encrypted(:auth_headers, key: Settings.db_encryption_key, marshal: true, marshaler: JsonMarshal::Marshaller)
     attr_encrypted(:evss_response, key: Settings.db_encryption_key, marshal: true, marshaler: JsonMarshal::Marshaller)
+    attr_encrypted(:bgs_responses, key: Settings.db_encryption_key, marshal: true, marshaler: JsonMarshal::Marshaller)
```
bastosmichael, line 11: Knowing that this is for flashes specifically and that we will be querying this to know when the flashes job passes or fails and that we will soon be working on special issues in the same fashion, does it make sense to store seperate BGS responses for flashes and special issues? Being that it's an encrypted cell we won't really be able to query for when flashes vs special issues errors exist etc.

bastosmichael (changes requested):
Knowing that this is for flashes specifically and that we will be querying this to know when the flashes job passes or fails and that we will soon be working on special issues in the same fashion, does it make sense to store seperate BGS responses for flashes and special issues? Being that it's an encrypted cell we won't really be able to query for when flashes vs special issues errors exist etc.

modules/claims_api/app/models/claims_api/auto_established_claim.rb
SlakrHakr: https://github.com/department-of-veterans-affairs/vets-api/pull/5335#discussion_r526219909

SlakrHakr: > I went ahead with a more specific column name. if we're strongly considering needing a boolean for each flash and special_issues failures, then might as well keep the responses themselves separate.

https://github.com/department-of-veterans-affairs/vets-api/pull/5335#discussion_r526559134

hinzed1127 (approved):
it looks like one of your [flash updater tests](https://github.com/department-of-veterans-affairs/vets-api/pull/5334/files#diff-ba56303bffa2566f97bf25baa6260e5e68e66a2b312ef5ab5e641335502c2562R58) is failing CI, but looks good once that's fixed

modules/claims_api/app/workers/claims_api/flash_updater.rb
```diff
@@ -8,13 +8,19 @@ class FlashUpdater
     include Sidekiq::Worker
     include SentryLogging
 
-    def perform(user, flashes)
+    def perform(user, flashes, auto_claim_id: nil)
```
bastosmichael, line 11: This won't allow our Sidekiq job to be model agnostic so that the va.gov team can use it but the more I think about that the more I think they should just use the gem directly themselves.

bastosmichael (approved):
Good Refactor, only merge and deploy this tomorrow after the database PR is merged and deployed today. Make sure CI/CD passes but otherwise the code looks good.

modules/claims_api/app/workers/claims_api/flash_updater.rb
SlakrHakr: i couldn't think of a better way to do this but im game for any ideas. i made the auto_claim_id optional trying to allow anyone to use the job with just a little less functionality if an error occurs. Letting them use the gem directly sounds good though too :+1:

thilton-oddball (approved):
LGTM

PR #5335, opened 2020-11-18
API-3261: Database migration

SlakrHakr:
## Description of change
Adds columns required for API-3261 work.

## Original issue(s)
https://vajira.max.gov/browse/API-3261

Will require migrations to be run within environment.

Specs within application work utilizing this database migration:
https://github.com/department-of-veterans-affairs/vets-api/pull/5334

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -34,7 +34,7 @@ ActiveRecord::Schema.define(version: 2020_11_17_200716) do
     t.index ["uuid"], name: "index_accounts_on_uuid", unique: true
   end
 
-  create_table "appeals_api_higher_level_reviews", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+  create_table "appeals_api_higher_level_reviews", id: :uuid, default: nil, force: :cascade do |t|
     t.string "status", default: "pending", null: false
     t.string "encrypted_form_data"
     t.string "encrypted_form_data_iv"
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -129,6 +129,8 @@
     t.string "encrypted_file_data_iv"
     t.string "encrypted_evss_response"
     t.string "encrypted_evss_response_iv"
+    t.string "encrypted_bgs_responses"
```
bastosmichael, line 132: Knowing that this is for flashes specifically and that we will be querying this to know when the flashes job passes or fails and that we will soon be working on special issues in the same fashion, does it make sense to store seperate BGS responses for flashes and special issues? Being that it's an encrypted cell we won't really be able to query for when flashes vs special issues errors exist etc.

bastosmichael (changes requested):
Let's figure out if we want to store flashes and special issues in seperate bgs response cells or if we think we can query one encrypted cell for both sets of errors we might get back from BGS?

db/schema.rb
SlakrHakr: i think it depends on what we will ultimately care about. i agree that if we're going to need to query, say, only flash errors over the period of years potentially, then yeah, separate fields is probably a must. if we plan more to check for errors over the period of days or weeks, its probably good enough to query all "bgs errors" and then split out by "type" for the report itself? also might make a difference whether we're primarily going to care about only flashes or only special issues at any given point, or if we're typically caring about "bgs errors" holistically.

happy to make this field more specific if that makes the most sense

SlakrHakr: I went ahead with a more specific column name. if we're strongly considering needing a boolean for each flash and special_issues failures, then might as well keep the responses themselves separate.

hinzed1127 (approved):

bastosmichael (approved):
LGTM, seems repetitive to include multiple error response types in the same model but as we start moving things over to BGS I think it'll be important to track those errors at different levels across different services.

PR #5337, opened 2020-11-18
fix scoping for notification

LindseySaari:
## Description of change
Swagger::Schemas namespace has a Notification class. app/swagger/swagger/schemas/dismissed_status.rb refers to Notification but doesn't mean in the same namespace, so to refer to the Notification model, the scope must be escaped.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#15888


kfrz (approved):
LGTM!

PR #5338, opened 2020-11-18
fall back to identity provider birth date for evss headers

annaswims:
## Description of change
If we can't get the date of birth from MPI get it from the identity provider for EVSS headers.
## Original issue(s)
department-of-veterans-affairs/va.gov-team#6298


kfrz (approved):
Looks good, ship it!

PR #5340, opened 2020-11-18
MVI slo improvements and and additional error logging

bramleyjl:
## Description of change

Updates the v0 authentication logout route (`/sessions/slo/new`) to ONLY load the user and destroy the session before issuing an SLO call to ID.me (similar to what's happening in v1).  Previously the `authenticate` method would also "extend" the session, before issuing SLO.  While in the majority of cases this works, sometimes we need to query MPI to get data in order to "extend" the session.  If the [MPI query fails](http://sentry.vfs.va.gov/organizations/vsp/issues/6593/?environment=production&project=3&query=is%3Aunresolved+%22MVI%3A%3AErrors%3A%3AInvalidRequestError%22&statsPeriod=14d) this can result in an error for the user, but these are unnecessary as the user SHOULD NOT need to do this.

In addition, temporary logging of outgoing XML in the event that InvalidRequestError is returned has been added so that we can further improve the missing key/values validator added in #5151 and #5357   

## Original issue(s)
refs https://github.com/department-of-veterans-affairs/va.gov-team/issues/15180
Sentry error: http://sentry.vfs.va.gov/share/issue/63f9713d288e49dbb743bf0fa5ed24a0/

- [ ] manually test that if an unauthenticated user issues an SLO request to ID.me, that they are redirected back to the va.gov home page (double checking that ID.me doesn't issue an error if the user DOESN'T have a session)
v0 `sessions_controller_spec` updated for changers to pass, manual testing still needed.

### DIFF
--- a/app/controllers/v0/sessions_controller.rb
+++ b/app/controllers/v0/sessions_controller.rb
@@ -33,11 +33,12 @@ module V0
 
       StatsD.increment(STATSD_SSO_NEW_KEY, tags: ["context:#{type}", VERSION_TAG])
       Rails.logger.info("SSO_NEW_KEY, tags: #{["context:#{type}", VERSION_TAG]}")
-      url = url_service.send("#{type}_url")
       if type == 'slo'
+        url = url_service.ssoe_slo_url
         Rails.logger.info('SSO: LOGOUT', sso_logging_info)
         reset_session
       else
+        url = url_service.send("#{type}_url")
         saml_request_stats
       end
       # clientId must be added at the end or the URL will be invalid for users using various "Do not track"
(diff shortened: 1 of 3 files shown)


bramleyjl:
I've changed the MVI/MPI error logging to include the outgoing user_identity XML that we're sending to MVI in the case that it fails, the thought being that if we are able to see what outbound XML is causing the InvalidRequestError responses then we can improve the new FindProfileMessageFields validation to check for whatever specific conditions are continuing to trigger the errors and catch them before the request is made.

thilton-oddball (approved):
LGTM

PR #5344, opened 2020-11-18
only provide separation location if bdd qualified

annaswims:
## Description of change
if a veteran filled out a form 526 when they were bdd-qualified, then some time passed before they submitted and they're no longer bdd-eligible don't send the separation location.  EVSS will reject it. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#16376


kfrz (approved):

annaswims:
I think I may need to change this to if the RAD date is in the future, which is more than if BDD qualified.

PR #5345, opened 2020-11-18
remove acronym and update filenames

LindseySaari:
## Description of change
This PR removes the SCO acronym in favor of the spelled out version. The acronym is very specific to an emailer and not necessarily an official VA adopted acronym. This PR renames EducationForm::SendSCOEmail to EducationForm::SendSchoolCertifyingOfficialsEmail and corresponding filenames to avoid the need of an acronym.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#15890

## Testing


kfrz (approved):
Looks good!!

PR #5346, opened 2020-11-18
Allow larger file for EVSS

annaswims:
## Description of change
EVSS will accept PDF's that are larger than 50mb and the other files must be less than 50mb

## Original issue(s)
department-of-veterans-affairs/va.gov-team#16004

I'll wait until EVSS gives us the go-ahead before merging.


kfrz (approved):
Nice spec! Looks good.

PR #5347, opened 2020-11-18
get banking info from ppiu if submitted info is redacted

lihanli:
## Description of change
if a user submits a form with redacted banking info we should fetch the unredacted info from ppiu

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/11773

tested with unit tests


annaswims (approved):
LGTM!

thilton-oddball (approved):
Looks good!

PR #5351, opened 2020-11-19
Refactored Pact verification workflow to work independently of webhooks

U-DON:
## Description of change
Added a step in the `pact_verification` CircleCI workflow to trigger vets-website's `can-i-deploy` workflow.

This is meant to build on top of #5288.

## Original issue(s)
Related to department-of-veterans-affairs/va.gov-team#16325.

The newly added `notify_pact_consumers` job uses a parameter `consumer_branch`.
That parameter's expected to be passed in from the vets-website job that triggered the `pact_verification` workflow.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,9 +13,6 @@
     consumer_branch:
       type: string
       default: ""
-    consumer_job_id:
-      type: string
-      default: ""
     consumer_name:
       type: string
       default: ""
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -131,7 +145,9 @@
       when: << pipeline.parameters.verify_stable_pacts >>
       jobs:
         - verify_pacts
-        - pact_broker_artifact
+        - notify_pact_consumers:
+            requires:
+              - verify_pacts
```
U-DON, line 150: Not sure how to get this to always run regardless of `verify_pacts` failing or passing.

U-DON: The only way I can think of is making a separate `verify_pacts_and_notify` job that does the notify command with `when: always`.

PR #5352, opened 2020-11-19
Api-3177 Insert email and representative's name into NOD pdf

ElleeB:
## Description of change
Character limits for the NOD Pdf fields: _email_ and _representative_name_, are are too limiting.

In response, we've updated the NOD PDF constructor to insert/draw text onto each of the mentioned fields. This will allow us to control the font size and placement of the text - will be addressed in a subsequent ticket.

## Ticket
https://vajira.max.gov/browse/API-3177

## Testing
Updated PDF files were added to spec > fixtures to satisfy specs


cilestin (approved):
LGTM

charleystran (approved):
LGTM

PR #5353, opened 2020-11-19
[16334] Add VA Awards Gross Amount to 5655 Prefill

camerontesterman:
## Description of change
Using `bgs-ext` the gross award amount entitled to veterans is added to the 5655 form profile prefill. Because of the shape of the payload that will need to be submitted to the Debt Management Center once a veteran submits their form, the form profile model's `convert_mapping` method was also changed in order to allow it to handle hashes that are within arrays in form profile mappings.

Prior to this PR, any array value, regardless of whether it was a string or a hash, within a form profile mapping would try to be interpreted using the `call_methods` method which would result in an error.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/16334

Tested and verified locally.


app/models/form_profiles/va5655.rb
```diff
@@ -1,6 +1,17 @@
 # frozen_string_literal: true
 
+module DebtManagementCenter
+  class VaAwardsComposite
+    include Virtus.model
+    attribute :name, String
+    attribute :amount, String
+    attribute :veteran_or_spouse, String
+  end
+end
```
kfrz, line 10: Any reason not to put this in it's own file? Autoloading will be problematic I think

app/services/bgs/awards_service.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module BGS
+  class AwardsService < BaseService
```
kfrz, line 4: Please add class-level YARD documentation

app/services/bgs/awards_service.rb
```diff
@@ -0,0 +1,15 @@
...
+    def get_awards
+      @service.awards.find_award_by_participant_id(@user.participant_id)
+    rescue => e
```
kfrz, line 7: Is there a specific exception class you can rescue from - this may swallow unrelated errors unexpectedly

app/models/form_profiles/va5655.rb
```diff
@@ -1,6 +1,17 @@
...
+
 class FormProfiles::VA5655 < FormProfile
```
kfrz, line 12: Please add class-level YARD documentation and also for public methods

kfrz (changes requested):
Couple of minor comments, looks good overall. Just looking for some doc-strings

app/models/form_profiles/va5655.rb
camerontesterman: Done.

app/services/bgs/awards_service.rb
camerontesterman: YARD documentation has been added.

app/models/form_profiles/va5655.rb
camerontesterman: Done.

kfrz (approved):
Thanks for the docstrings!  This looks good overall.

PR #5355, opened 2020-11-19
Update Form 10-10CG Logging

kevinmirc:
**Description**
Make updates to the Form 10-10CG logging.


kfrz (approved):
Looks good. I wish we could more elegantly handle logging throughout the app maybe with a better mixin, but that's out of scope here. Nice work!

kevinmirc:
Agreed. I had more ambitious plans for this log filter but it quickly outgrew the scope of this ticket, as you mentioned.

PR #5360, opened 2020-11-19
Log to sentry when 10-10CG PDF generation or attachment submission request fails

kevinmirc:
Fixes: https://github.com/department-of-veterans-affairs/va.gov-team/issues/15146

**Description**
Log the error that occurs when attachment submission, or generation of PDF, for 10-10CG fails.

### DIFF
--- a/app/services/form1010cg/service.rb
+++ b/app/services/form1010cg/service.rb
@@ -12,6 +12,8 @@ module Form1010cg
     class InvalidVeteranStatus < StandardError
     end
 
+    include SentryLogging
+
     attr_accessor :claim, # SavedClaim::CaregiversAssistanceClaim
                   :submission # Form1010cg::Submission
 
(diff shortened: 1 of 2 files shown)


app/services/form1010cg/service.rb
```diff
@@ -85,7 +85,7 @@ def submit_attachment # rubocop:disable Metrics/MethodLength,Metrics/CyclomaticC
 
         carma_attachments.submit!(carma_client)
         submission.attachments = carma_attachments.to_hash
-      rescue
+      rescue => e
```
kfrz, line 88: Is there a specific exception that you would want to rescue from here? Catch-all rescues may swallow up bugs that are unrelated, so specifying `Common::Client::Error` or whatever you'd expect to see error here, you might want to specify.

spec/services/form1010cg/service_spec.rb
```diff
@@ -706,26 +707,42 @@
 
       subject = described_class.new(claim, submission)
 
-      expect(subject.claim).to receive(:to_pdf).and_raise('pdf generation failure')
+      error = StandardError.new('Something went wrong')
```
kfrz, line 710: You might want to be more specific and build an error class for this service, check `lib/` for other examples of custom exceptions

kfrz (changes requested):
Couple of minor comments

app/services/form1010cg/service.rb
kevinmirc: In this case, I want to capture **all** exceptions. This method should only return true and false without raising an error. Mainly to prevent the error from bubbling up to the response and making user's thing their submission did not go through.

But I could also see the argument of the #new, #add, and #attachments= being moved out of the catch block since it doesn't connect to anything externally, it's just source code.

Does `Common::Client::Error` capture all network errors? I'd probably need to catch all 4** and 5** errors.

kevinmirc:
Closing since this method is soon to be replaced with a background job for processing attachments. Will rely on sidekiq to capture these errors: https://github.com/department-of-veterans-affairs/va.gov-team/issues/16952

PR #5362, opened 2020-11-20
update method from mvi to mpi

LindseySaari:
## Description of change
app/controllers/v0/profile/persons_controller.rb has an after_action for `invalidate_mvi_cache` which is now renamed to `invalidate_mpi_cache`. This effort is part of the upgrade path for renaming mvi to mpi related to the Zeitwerk upgrade.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#14783


kfrz (approved):
Ship it!

PR #5365, opened 2020-11-20
14938 mobile claims overview endpoint

jperk51:
## Description of change
Mobile claims and appeals overview endpoint

## Original issue(s)
department-of-veterans-affairs/va.gov-team#14938

All tests are in claims-and-appeals-overview.spec


modules/mobile/app/controllers/mobile/v0/claims_and_appeals_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        get_all_appeals = lambda {
+          begin
+            [appeals_service.get_appeals(@current_user).body['data'], true]
+          rescue => e
+            [Mobile::V0::Adapters::ClaimsOverviewErrors.new.parse(e, 'appeals'), false]
+          end
+        }
+        results = Parallel.map([get_all_claims, get_all_appeals], in_threads: 2, &:call)
```
kreek, line 31: minor but you could use parallel assignment here
```ruby
claims_result, appeals_result = Parallel.map([get_all_claims, get_all_appeals], in_threads: 2, &:call)
```

modules/mobile/app/models/mobile/v0/adapters/claims_overview.rb
```diff
@@ -0,0 +1,31 @@
...
+module Mobile
+  module V0
+    module Adapters
+      class ClaimsOverview
+        def parse(entry)
+          if entry['list_data']
+            {
+              id: entry['evss_id'].to_s,
+              type: 'claim',
+              subtype: entry['list_data']['status_type'],
```
kreek, line 12: when you're traversing down two nodes `.dig('list_data', 'status_type')` will protect you from hitting a nil value along the way

modules/mobile/app/controllers/mobile/v0/claims_and_appeals_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+          full_list.push(appeals[0])
+          status_code
+        else
+          # appeals error
+          error_list.push(appeals[0])
+          status_code == :multi_status ? :bad_gateway : :multi_status
+        end
+      end
+
+      def serialize_list(full_list)
```
kreek, line 64: this looks good and is very clear, but wonder if we can avoid a second iteration through the list by returning the model from the adapter? (I will do same in the appointments adapters) 😄

kreek (approved):
Really well architected, a nice way to handle an endpoint that can partially succeed, a few minor comments but LGTM 🚀

modules/mobile/app/models/mobile/v0/adapters/claims_overview.rb
jperk51: Unfortunately it looks like dig doesn't work with this object.
` undefined method `dig' for #<EVSSClaim:0x00007fe00b0eed48>`

PR #5373, opened 2020-11-23
API-3259: Flashes - Figure out what we want to capture and report on

SlakrHakr:
## Description of change
Adds flash information to existing claims report.

## Original issue(s)
https://vajira.max.gov/browse/API-3259

Updated specs and tested template through ActionMailer preview.


bastosmichael (approved):
I like this a lot, let's just make sure to show no flashes data so that we know when flashes are being used or not used in our payloads daily etc.

PR #5375, opened 2020-11-23
14838 mobile appointments service

kreek:
## Description of change
Creates a mobile version of the appointments service that can make VA and Community Care confirmed appointment calls in parallel.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#14838

* Extends VAOS service but modifies http adapter and sets a unique service name
* Adds CloudWatch logs for failed calls
* StatsD metrics added for get appointments call


kfrz (approved):
Excellent changeset. Especially appreciate the attention to detail in specs and the inline documentation. 

:ship: :+1:

modules/mobile/app/services/mobile/v0/appointments/service.rb
```diff
@@ -0,0 +1,76 @@
...
+
+          config.parallel_connection.in_parallel do
+            responses[:cc], errors[:cc] = get(cc_url, params)
+            responses[:va], errors[:va] = get(va_url, params)
+          end
+
+          if errors.values.any? { |e| !e.nil? }
+            Rails.logger.error('mobile get va appointments call failed') if errors[:va]
+            Rails.logger.error('mobile get community care appointments call failed') if errors[:cc]
+            raise Common::Exceptions::BackendServiceException, 'VAOS_502'
```
jjulian, line 42: During my recent work implementing the VEText service, I discovered that this exception will bury the `VAOS_502` unless it is defined as a key in `config/locales/exceptions.en.yml` file.

modules/mobile/config/initializers/statsd.rb
```diff
@@ -8,3 +8,10 @@
 Mobile::ApplicationController.statsd_count_success :authenticate, 'mobile.authentication'
 StatsD.increment('mobile.authentication.success', 0)
 StatsD.increment('mobile.authentication.failure', 0)
+
+Mobile::V0::Appointments::Service.extend StatsD::Instrument
```
jjulian, line 12: @kreek lets think about what other mobile metrics we may want to capture.

modules/mobile/app/services/mobile/v0/appointments/service.rb
kreek: There's an entry for that key [here](https://github.com/department-of-veterans-affairs/vets-api/blob/master/config/locales/exceptions.en.yml#L561-L566). I thought about adding our own key but it is marking that something happened to the upstream service and didn't want two different errors to map to the same issue.

modules/mobile/config/initializers/statsd.rb
kreek: 👍 my plan is to get the initial set of endpoints in to unblock the FE then circle back have a longer think about which metrics we want to capture.

modules/mobile/app/services/mobile/v0/appointments/service.rb
jjulian: Cool, thanks for double-checking.

Issue #5378, opened 2020-11-23
Using bootsnap to reduce application startup time in development and test environments


bosawt:
It seems that the application loads fairly slowly in development and tests. There's a well known gem called bootsnap that I've used before (and is part of Rails itself now, so it's safe), that generally speeds things up. If we don't want to risk changes in production, we could also implement this for test and development environments only. Here's an example of the speed differences when running our tests over a single file:
Test time without using bootsnap (roughly 15 seconds)
<img width="523" alt="Screen Shot 2020-11-23 at 10 08 08 AM" src="https://user-images.githubusercontent.com/42326799/100007509-09993a00-2d81-11eb-82fa-5579b9361fdd.png">

Test with using bootsnap (roughly 8 seconds)
<img width="499" alt="Screen Shot 2020-11-23 at 10 07 18 AM" src="https://user-images.githubusercontent.com/42326799/100007505-0900a380-2d81-11eb-9239-7dba6ed4d767.png">

PR #5381, opened 2020-11-23
Adding bootsnap gem to improve dev and test environment performance

bosawt:
## Description of change
This PR adds the `bootsnap` gem for `test` and `development` environments to speed up the process of file loading. The speed improvements should be fairly dramatic (I was seeing specs load in roughly half the time). Production deployment should be tested since the `boot.rb` file was changed, but if working properly, production should not be affected.

`bootsnap` simply caches files that Rails loads, and it watches for file changes to update these files. It basically makes `spring` unnecessary, and is a lot safer than `spring` anyway.

## Original issue(s)
https://github.com/department-of-veterans-affairs/vets-api/issues/5378

- This solution assumes `ENV['RAILS_ENV']` is defined by time we initialize Rails in `boot.rb`. I've seen application implementations where this is a problem. If so, we have a workaround where we can attempt to load `bootsnap` regardless of environment, rescue the expected error on production, and continue with initialization. I like the current attempt better so hopefully this works with our implementation. EDIT: I've updated this to default to NOT loading bootsnap if the RAILS_ENV var isn't there. This way it'll only be loaded if the variable is specifically set to either `development` or `test`
- We can actually implement this on production as well, and expect a certain amount of speed benefits. However, it requires the production file system to not be read-only, and this constraint is significant. It's worth looking into this, but might not be possible depending on our production environment.

## How to Test?
-  Run specs or Rails console, try to measure the time it takes to boot to console or start running specs
- `bundle` to load the Gem
- Re-run specs or re-open Rails console a couple times, should see boot times dramatically improve

### DIFF
--- a/config/boot.rb
+++ b/config/boot.rb
@@ -3,3 +3,8 @@
 ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../Gemfile', __dir__)
 
 require 'bundler/setup' # Set up gems listed in the Gemfile.
+
+# Bootsnap is used to speed up application load time. Enabled for development/test
+# environments. If RAILS_ENV isn't set, assumption is that we are in a development/test
+# environment
+require 'bootsnap/setup' if %w[development test].include?(ENV['RAILS_ENV'])
--- a/config/boot.rb
+++ b/config/boot.rb
@@ -5,6 +5,5 @@ ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../Gemfile', __dir__)
 require 'bundler/setup' # Set up gems listed in the Gemfile.
 
 # Bootsnap is used to speed up application load time. Enabled for development/test
-# environments. If RAILS_ENV isn't set, assumption is that we are in a development/test
-# environment
+# environments.
 require 'bootsnap/setup' if %w[development test].include?(ENV['RAILS_ENV'])
--- a/Gemfile
+++ b/Gemfile
@@ -154,6 +154,7 @@ end
 # rubocop:disable Metrics/BlockLength
 group :development, :test do
   gem 'awesome_print', '~> 1.8' # Pretty print your Ruby objects in full color and with proper indentation
+  gem 'bootsnap', require: false
   gem 'brakeman', '~> 4.7'
   gem 'bundler-audit'
   gem 'byebug', platforms: :ruby # Call 'byebug' anywhere in the code to stop execution and get a debugger console
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -209,6 +209,8 @@ GEM
       i18n
     benchmark-ips (2.8.2)
     bindex (0.8.1)
+    bootsnap (1.5.1)
+      msgpack (~> 1.0)
     brakeman (4.9.0)
     breakers (0.4.0)
       faraday (>= 0.7.4, < 0.18)
@@ -501,6 +503,7 @@ GEM
     mini_mime (1.0.2)
     mini_portile2 (2.4.0)
     minitest (5.14.2)
+    msgpack (1.3.3)
     multi_json (1.15.0)
     multi_xml (0.6.0)
     multipart-post (2.1.1)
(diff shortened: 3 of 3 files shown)


config/boot.rb
```diff
@@ -3,3 +3,10 @@
 ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../Gemfile', __dir__)
 
 require 'bundler/setup' # Set up gems listed in the Gemfile.
+
+env = ENV['RAILS_ENV'] || 'development'
+if env == 'development' || env == 'test'
+    require 'bootsnap/setup'
+else
+  STDOUT.puts 'Skipping bootsnap load for non-dev environment'
```
f1337, line 11: I would not output to STDOUT in non-dev environs, as I'm not sure of the downstream impacts in prod, CI, etc.

f1337:
I have no objections, have used `bootsnap` in the past too. I think we should loop @dginther, @jbritt1, et al into the convo, just in case they see any fallout on the ops side.

Thank you Trevor!

config/boot.rb
bosawt: Sounds good, I'll remove this line. We probably don't need it anyway, since we're only building the Gem for test/dev, it will immediately fail if we accidentally require bootsnap here on staging/prod environments

bosawt:
> I have no objections, have used `bootsnap` in the past too. I think we should loop @dginther, @jbritt1, et al into the convo, just in case they see any fallout on the ops side.
> 
> Thank you Trevor!

From Slack, Demian and Jeremy don't see any issues with this: https://dsva.slack.com/archives/CJYRZK2HH/p1607536317268400

erluti:
> This solution assumes ENV['RAILS_ENV'] is defined by time we initialize Rails in boot.rb. I've seen application implementations where this is a problem.

How does this problem manifest?  What were the conditions that made this a problem?

bosawt:
> > This solution assumes ENV['RAILS_ENV'] is defined by time we initialize Rails in boot.rb. I've seen application implementations where this is a problem.
> 
> How does this problem manifest? What were the conditions that made this a problem?

@erluti Ah, you know what, so I looked through my old notes to recall what exactly the case was when this was happening before. It looks like in my previous position, we had a Docker local dev setup where we weren't properly setting up environment variables. So I think this problem was only happening on local dev, and it was an issue with the weird setup at the time.

It actually made me think though, maybe the default should be to 'not' load bootsnap if the env var isn't there for whatever reason. That way we err on the side of safety, and really won't have to worry, since it'll only ever load in case it is set to `development` or `test` specifically

erluti:
> It actually made me think though, maybe the default should be to 'not' load bootsnap if the env var isn't there for whatever reason. That way we err on the side of safety, and really won't have to worry, since it'll only ever load in case it is set to `development` or `test` specifically

nice! I think this is probably the better approach.  I think all our deployed instances run as `production` so this should be good.

erluti (approved):

config/boot.rb
```diff
@@ -3,3 +3,7 @@
 ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../Gemfile', __dir__)
 
 require 'bundler/setup' # Set up gems listed in the Gemfile.
+
+# Bootsnap is used to speed up application load time. Enabled for development/test
+# environments.
+require 'bootsnap/setup' if %w[development test].include?(ENV['RAILS_ENV'])
```
erluti, line 9:
```suggestion
require 'bootsnap/setup' if Rails.env.development? || Rails.env.test?
```
minor suggestion.  I think these helpers make things a little clearer

bosawt: 👍 Good idea

bosawt: Whoops, sorry, I initially implemented your suggestion, but I had to revert it. Turns out in `boot.rb`, we haven't actually instantiated the `Rails` singleton, so we can't use the `Rails.env` helpers yet!

bosawt: @erluti ^

erluti: dang, i was just fooling in the console and not in the boot.rb file.  Thanks for trying it!

PR #5385, opened 2020-11-23
Only skip authentication for some attachment controllers

annaswims:
## Description of change
Rather than excluding all form attachment controllers from authenticating, it should be done on a case-by-case basis. upload_supporting_evidence is called from form 526 which requires all users to be authenticated.  By not skipping authentication, it will make it easier to troubleshoot which users are having problems with uploads, now that the user's uuid is in sentry.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#16216


thilton-oddball (approved):
All looks good!

PR #5386, opened 2020-11-24
API-3727 Use text_box autosizing for some NOD fields

cilestin:
## Description of change

Updated email, rep name, and issues column to use autosized text within a bounding box.

Minimum font size for autosized text is 8pt. Anything that does not fit the bounding box after autosizing will be truncated.

## Original issue(s)
https://vajira.max.gov/browse/API-3727

* New minimum PDF fixture file was generated due to these changes


ElleeB (approved):
This looks great!

PR #5389, opened 2020-11-24
refactor out expirations config file

erluti:
This PR removes the use of a config file that isn't used in part of a deploy or has been modified for other form types ever.  There was an instance discovered where the value was overwritten, but [that class is no longer used](https://github.com/department-of-veterans-affairs/vets-api/pull/5388).
This refactors away that file so that the logic can be seen explicitly in the code.

The original goal of this work was to support leap years instead of using `365.days`, but after some experimenting Rails handles 365.days generally the same as 1.year.


thilton-oddball (approved):
LGTM!

PR #5392, opened 2020-11-24
Pension Questions Default Values

kathleencrawford:
## Description of change

Pension eligibility questions were added to the VA Form 686c (Dependency Claim), but some questions were only added for certain workflows and the questions are OPTIONAL. When the eligibility answers are sent to BGS, they are converted from true/false values to Y/N. This PR is an update so that if the question is not present or was not answered by the user, a nil value is sent instead of a default value of "N".

## Original issue(s)
department-of-veterans-affairs/va.gov-team#16364

This is a minor change so no additional logging or settings were added for this PR.

## Testing
- [x] Local Testing (done)
- [x] Testing in Staging (planned)


lib/bgs/vnp_veteran.rb
```diff
@@ -29,19 +29,13 @@ def create
           va_file_number: @va_file_number,
           claim_type_end_product: claim_type_end_product,
           location_id: location_id,
-          net_worth_over_limit_ind: net_worth_over_limit_ind
+          net_worth_over_limit_ind: veteran.formatted_boolean(@payload['dependents_application']['household_income'])
```
thilton-oddball, line 32: Should this be veteran.formatted_boolean(@payload['dependents_application']['household_income'])
or should it just be formatted_boolean(@payload['dependents_application']['household_income'])?

kathleencrawford: @thilton-oddball yes, It should be `veteran.formatted_boolean` because the `formatted_boolean` method is in the `models/bgs_dependents/base.rb` Base class, and the veteran model inherits from that class

thilton-oddball (approved):
LGTM

kathleencrawford:
thank you @thilton-oddball !

PR #5395, opened 2020-11-25
Update list of enum values for flashes in swagger and 526 schema

hinzed1127:
## Description of change

This adds missing values for the `flashes` enum in 526 schema and swagger

## Original issue(s)
https://vajira.max.gov/browse/API-3254

there's a handful of values with unresolved questions. That's described in the [issue](https://vajira.max.gov/browse/API-3254).

### DIFF
--- a/modules/claims_api/app/swagger/claims_api/forms/form_526_response_swagger.rb
+++ b/modules/claims_api/app/swagger/claims_api/forms/form_526_response_swagger.rb
@@ -229,7 +229,185 @@ module ClaimsApi
                   items do
                     key :type, :string
                     key :example, 'Hardship'
-                    key :enum, ['Hardship', 'Seriously Injured/Very Seriously Injured', 'Terminally Ill', 'Homeless', 'Purple Heart', 'POW', 'Medal of Honor', 'Amyotrophic Lateral Sclerosis', 'Emergency Care']
+                    key :enum, %w[
+                      1151 P&T
+                      38 USC 1151
+                      Active Chapter 31 Case
+                      Agent Orange - Vietnam
+                      Amyotrophic Lateral Sclerosis
+                      Annual IU Eligibility Review-Fraud
+                      Attorney Fee
+                      Authzn to Release Information to Third Party
+                      Automobile Adaptive Equipment Paid
+                      Automobile Allowance Paid
+                      BVA Decision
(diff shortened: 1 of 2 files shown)


modules/claims_api/app/swagger/claims_api/forms/form_526_response_swagger.rb
```diff
@@ -229,7 +229,185 @@ class Form526ResponseSwagger
                   items do
                     key :type, :string
                     key :example, 'Hardship'
-                    key :enum, ['Hardship', 'Seriously Injured/Very Seriously Injured', 'Terminally Ill', 'Homeless', 'Purple Heart', 'POW', 'Medal of Honor', 'Amyotrophic Lateral Sclerosis', 'Emergency Care']
+                    key :enum, %w[
```
SlakrHakr, line 232: `%w` won't work here i believe, since there are spaces within the values.
tested my theory out here and it does seem to be the case
```
2.6.6 :001 > (%w[1151 P&T 38 USC 1151]).count
 => 5 
```
These will need to be converted back to the original array of strings format.

SlakrHakr (changes requested):

modules/claims_api/app/swagger/claims_api/forms/form_526_response_swagger.rb
```diff
@@ -229,7 +229,185 @@ class Form526ResponseSwagger
...
+                      Clear and Unmistakable Error
+                      Closed Chapter 31 Case
+                      Committee on Waiver Case
+                      Compensation Award Line
+                      Compensation Under 1925 Rating Schedule
+                      Congressional
+                      Converted - Automobile Adaptive Equipment Paid
+                      Converted - Automobile Allowance Paid
+                      Converted - Blind
+                      Converted - Comp Under 1925 Rating Schedule
```
bastosmichael, line 264: I'm pretty sure using %w will actually end up converting this into "Converted", "-", "Comp", "Under"... Can you double check and confirm for me in a test that this won't happen?

bastosmichael (changes requested):
Remove %w and type it out unfortunately...

bastosmichael (approved):
Glad to see the fix...

SlakrHakr (approved):

PR #5398, opened 2020-11-25
API-3181 NOD Extra Issues page

cilestin:
## Description of change
Attach an additional page detailing the Veteran's issues when they extend beyond the number allowed by the form.

Stamps all pages with the Veteran's last name & last 4 of SSN to help keep the pages together should they be printed, regardless of the number of pages attached to the form.

## Original issue(s)
https://vajira.max.gov/browse/API-3181

Tested with a number of additional issues to ensure they were in the correct order.

Refactored the "extra" pdf test to do content checking, which ensures the added page includes the issue text. Removed the extra fixture file since it wasn't being used anyway.


ElleeB (approved):
LGTM

PR #5399, opened 2020-11-26
API-3572 Hearing Page for NOD pdf

cilestin:
## Description of change
WIP

Add additional administrative page to pdf where necessary. Currently adds the hearing type requested by the Veteran. Is a separate page since Issues have their own callout for separate pages in the form.

## Original issue(s)
https://vajira.max.gov/browse/API-3338

### DIFF
--- a/modules/appeals_api/lib/appeals_api/notice_of_disagreement_pdf_constructor.rb
+++ b/modules/appeals_api/lib/appeals_api/notice_of_disagreement_pdf_constructor.rb
@@ -30,6 +30,7 @@ module AppealsApi
       return @pdf_options if @pdf_options
 
       options = {
+        additional_pages: [],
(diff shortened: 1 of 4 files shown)


modules/appeals_api/lib/appeals_api/notice_of_disagreement_pdf_constructor.rb
```diff
@@ -66,6 +69,24 @@ def pdf_options
     # rubocop:enable Metrics/PerceivedComplexity
     # rubocop:enable Metrics/AbcSize
 
+    def merge_page(temp_path, output_path)
```
cilestin, line 72: Overrode here as a proof of concept that we will probably want to move to the BasePdfConstructor, but I didn't want to potentially mess with the HLR output without understanding it further.

PR #5402, opened 2020-11-27
Add ch31 vendor endpoint

digitaldrk:
## Description of change
During staging testing, I realized that we didn't have the endpoint url config for an API call. The config itself was superfluous so I added the url to the code base. Also during testing, the dev on the other team with the API we consume asked for a an application_intake id so I added that temporarily while we're testing in staging.

## Original issue(s)
No issue attached to this PR

Just a removal
This feature is not available to the public
Yes, we added a message log to get a reference id
No
No not in this PR
No concerns or questions

### DIFF
--- a/config/settings/test.yml
+++ b/config/settings/test.yml
@@ -89,4 +89,3 @@ veteran_readiness_and_employment:
   base_url: https://fake_url.com
   credentials: fake_auth
   auth_endpoint: https://fake_url.com/auth/oauth/token
-  ch_31_endpoint: /fake/endpoint
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -632,4 +632,3 @@ veteran_readiness_and_employment:
   base_url: ~
   credentials: ~
   auth_endpoint: ~
-  ch_31_endpoint: ~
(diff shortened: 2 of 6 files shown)


config/settings/test.yml
```diff
@@ -89,4 +89,3 @@ veteran_readiness_and_employment:
   base_url: https://fake_url.com
   credentials: fake_auth
   auth_endpoint: https://fake_url.com/auth/oauth/token
-  ch_31_endpoint: /fake/endpoint
```
erluti, line 92: what was the motivation for taking these out of the config?

digitaldrk: Hey @erluti thank you for looking into this.

I suppose the motivation was to reduce complexity.

It was overkill to include that path in the config and I got my wires crossed with ops. I had sent them the `credentials`, `base_url`, and `auth_endpoint` but not the `ch_31_endpoint`.

So in testing, I realized that when we made the request that uses the `ch_31_endpoint`, the string wasn't there.

Instead of opening a request with ops to add the string to credstash, we have this PR. I'm in sort of a time crunch.

I will include some conversations and PR's as reference below.

https://dsva.slack.com/archives/CBU0KDSB1/p1606492837200400

https://dsva.slack.com/archives/CLY6Q69RV/p1605822014289500?thread_ts=1605210607.227700&cid=CLY6Q69RV

https://github.com/department-of-veterans-affairs/devops/pull/8060

erluti (approved):
LGTM

PR #5407, opened 2020-11-30
consistent User#birth_date format

annaswims:
## Description of change
When I call @current_user.birth_date, I'd expect the date to be returned in the same format.  I don't anticipate any issues with this global change because we were already returning the dashed format most of the time. 

Before the dates were both strings in different formats: 

`@current_user.mpi&.profile&.birth_date => "19801201" `
`@current_user.identity.birth_date => "1980-12-01"`

## Original issue(s)
department-of-veterans-affairs/va.gov-team#16649

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -90,7 +90,7 @@ class User < Common::RedisStore
   end
 
   def birth_date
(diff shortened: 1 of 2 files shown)


erluti (approved):
LGTM

PR #5408, opened 2020-11-30
add some sentry logging for activity around okta service calls

erluti:
Fully intend to roll this back, but adding some sentry-info logging in order to see if I can glean some more information about the reason connected apps aren't refreshing immediately.

### DIFF
--- a/app/models/okta_redis/grants.rb
+++ b/app/models/okta_redis/grants.rb
@@ -8,23 +8,36 @@ module OktaRedis
   class Grants < Model
     CLASS_NAME = 'GrantsService'
 
+    after_initialize do |grants|
+      grants.instance_variable_set(:@activity_log, [])
+    end
+
     def all
-      okta_response.body
+      all_grants = okta_response.body
+      log_message_to_sentry('Getting all grants', :info,
+                            grants_service_activity: @activity_log)
+      all_grants
     end
 
     def delete_grants(grant_ids)
       success = grant_ids.reduce(true) { |memo, grant| delete_grant(grant) && memo }
       OktaRedis::Grants.delete(cache_key) if success
+      log_message_to_sentry('Deleting all grants', :info,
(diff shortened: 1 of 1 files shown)


app/models/okta_redis/grants.rb
```diff
@@ -8,34 +8,51 @@ module OktaRedis
   class Grants < Model
     CLASS_NAME = 'GrantsService'
 
+    after_initialize do |grants|
+      grants.instance_variable_set(:@activity_log, [])
```
erluti, line 12: the `Model` class uses `Common::RedisStore` which defines the `after_initialize` hook.  This is mostly to tap into the "constructor" even though, which is in `RedisStore` which has a complex signature and I took this approach to be less invasive.

app/models/okta_redis/grants.rb
```diff
@@ -8,34 +8,51 @@ module OktaRedis
...
+    end
+
     def all
-      okta_response.body
+      all_grants = okta_response.body
+      log_message_to_sentry('Getting all grants', :info,
```
erluti, line 17: `okta_response` will add some data too `@activity_log` and then we'll log all of that to sentry before returning the okta_response.body

app/models/okta_redis/grants.rb
```diff
@@ -8,34 +8,51 @@ module OktaRedis
...
       end
     end
 
     private
 
     def okta_response
-      do_cached_with(key: cache_key) do
+      @activity_log << 'Get okta response including grants'
+      response = do_cached_with(key: cache_key) do
+        @activity_log << 'Fetch new response'
```
erluti, line 50: this is the issue we're ultimately trying to get insight on.  After deleting a grant, we should need to fetch a new response.  BUT if it's something with Okta's service, it wouldn't have changed that much

kfrz (approved):
This looks good to me!

PR #5410, opened 2020-11-30
Parsing birls_id and participant_id on the user model from the User I…

bosawt:
…dentity model before falling back to the MPI Data model

## Description of change
This PR saves a user's birls_id and participant_id on the User Identity model, in order to reduce the necessary calls out to the MPI data object. Basically, we are caching these fields in a redis storage so they can be retrieved over and over with less performance impact.

## Original issue(s)
https://app.zenhub.com/workspaces/vsp-5cedc9cce6e3335dc5a49fc4/issues/department-of-veterans-affairs/va.gov-team/16415

- We need to make sure that the birls_id and participant_id don't contain any information that may need to be additionally encrypted now that it is stored in an additional place. 
- One data point, Joe stated that his birls_id is his SSN, so it's possible that we need to make sure we keep security in mind when storing this.

## Dev Acceptance:
- Set a breakpoint in `app/models/user_session_form.rb`, line 26, after `@user_identity` is set as an instance variable for the `User` model
- Logged in locally with: `http://localhost:3000/v1/sessions/idme/new`
- Skipped the breakpoint the first time (this is called, then user is upleveled with `loa` values, then this line is called again)
- Confirmed that `user.birls_id`, `user.participant_id` exists
- Confirmed that `user.identity.birls_id`, `user.identity.participant_id` exists.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -129,10 +129,8 @@ class User < Common::RedisStore
   delegate :common_name, to: :identity, allow_nil: true
 
   # mpi attributes
-  delegate :birls_id, to: :mpi
   delegate :icn, to: :mpi
   delegate :icn_with_aaid, to: :mpi
-  delegate :participant_id, to: :mpi
   delegate :vet360_id, to: :mpi
   delegate :search_token, to: :mpi
 
@@ -146,6 +144,14 @@ class User < Common::RedisStore
 
   def sec_id
     identity.sec_id || va_profile&.sec_id
+  end
+
+  def birls_id
+    identity&.birls_id || mpi&.birls_id
+  end
+
+  def participant_id
+    identity&.participant_id || mpi&.participant_id
   end
 
   def va_profile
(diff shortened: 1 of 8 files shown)


ericbuckley:
@bosawt not sure if you have access to the proxy yet, but there is a linting error
```
spec/lib/mpi/responses/id_parser_spec.rb:119:7: C: Layout/IndentationWidth: Use 2 (not 0) spaces for indentation.
```

ericbuckley (commented):
@bosawt looks great so far, thanks for adding all the tests.  The one other acceptance test I think we should run locally is authentication with v0.  To do that, start the authentication flow at `http://localhost:3000/sessions/idme/new` (note the missing v1).  I'd expect the identity object to have `nil` values for both `birls_id` and `participant_id` attributes, but we should double check that.

thilton-oddball (approved):
This all looks good!

PR #5411, opened 2020-11-30
added directory#show

bradenshipley:
## Description of change
Added show method to the directory controller to allow for getting a single application in the directory. Previously there was only an index method that returned all records. The `routes.rb` for the apps_api module has been updated.
## Original issue(s)
https://vajira.max.gov/browse/API-3843?workflowName=Lighthouse+Default+Workflow&stepId=3

tests have been updated to account for new route method.

### DIFF
--- a/modules/apps_api/config/routes.rb
+++ b/modules/apps_api/config/routes.rb
@@ -6,9 +6,10 @@ AppsApi::Engine.routes.draw do
   match '/v0/healthcheck', to: 'metadata#healthcheck', via: [:get]
 
   namespace :v0, defaults: { format: 'json' } do
-    get 'directory', to: 'directory#index'
-    get 'directory/scopes/:category', to: 'directory#scopes', as: 'scopes', defaults: { category: 'unknown_category' }
-    get 'directory/:name', to: 'directory#show'
+    scope_default = { category: 'unknown_category' }
+    get 'directory/scopes/:category', to: 'directory#scopes', defaults: scope_default
+    get 'directory/scopes', to: 'directory#scopes', defaults: scope_default
+    resources 'directory', only: %i[index show]
   end
   namespace :docs do
     namespace :v0 do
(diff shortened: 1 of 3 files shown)


modules/apps_api/config/routes.rb
```diff
@@ -6,8 +6,9 @@
   match '/v0/healthcheck', to: 'metadata#healthcheck', via: [:get]
 
   namespace :v0, defaults: { format: 'json' } do
-    get 'directory/scopes/:category', to: 'directory#scopes'
-    resources :directory, only: %i[index]
+    get 'directory', to: 'directory#index'
+    get 'directory/scopes/:category', to: 'directory#scopes', as: 'scopes', defaults: { category: 'unknown_category' }
+    get 'directory/:name', to: 'directory#show'
```
SlakrHakr, line 11: why not just define the scopes situation here? pretty sure something like this should work, and mean you wouldn't need the if check and redirect in the controller.
```
get 'directory/scopes/:category', to: 'directory#scopes', as: 'scopes', defaults: { category: 'unknown_category' }
get 'directory/scopes', to: 'directory#scopes'
get 'directory/:name', to: 'directory#show'
```
Sidenote, im a bit surprised `default` for category wouldn't work that way by default, but im pretty sure adding that line in the middle to explicitly look for "scopes" above ":name" would remove the need for your controller if/redirect bit

SlakrHakr (approved):

modules/apps_api/config/routes.rb
bradenshipley: It works correctly I just end up having to set the default explicitly again. Not sure which is more ugly tbh. The routing is only ugly cause I have the scopes not being handled in their own controller..
`  get 'directory/scopes/:category', to: 'directory#scopes', defaults: { category: 'unknown_category' }
    get 'directory/scopes', to: 'directory#scopes', defaults: { category: 'unknown_category' }
    resources 'directory', only: %i[index show]`

SlakrHakr: i suppose you could set that default category to a variable above and reference it so they stay in sync.
but whatever you think is best. i typically see basic if/then/redirect in a controller and think it feels wrong, seems like something that should be handled solely in the routes file, but i approved the pr if this way makes more sense to you :+1:

PR #5401, opened 2020-11-27
API-3828 Add status to notice of disagreement

ElleeB:
## Description of change
Need to add status to notice of disagreements table
https://vajira.max.gov/browse/API-3828

## Things to know about this PR
Two migration files added

PR #5412, opened 2020-11-30
fix lint error

annaswims:
## Description of change
I don't understand why, but linting passed on the branch of https://github.com/department-of-veterans-affairs/vets-api/pull/5401 but not on master


erluti (approved):
Thanks for the fix!

PR #5413, opened 2020-11-30
Revert "add some sentry logging for activity around okta service call…

erluti:
rolls back this PR https://github.com/department-of-veterans-affairs/vets-api/pull/5408 to remove some investigation logging


erluti:
Used this information to prove we are deleting cache properly on deletes and proved that we do a new fetch afterwards.

ericboehs (approved):

PR #5415, opened 2020-12-01
Increase NOD email & rep fields to 3 lines of text

cilestin:
## Description of change
Allows even more characters to be placed into the Veteran email & representative name fileds of the NOD document. 

Autosizes up to 3 lines of text, increased from 2.

Screen Shot 2020-11-30 at 16 32 01


ElleeB (approved):
solid.

PR #5416, opened 2020-12-01
rubocop ignore migrations for all modules

omgitsbillryan:
## Description of change
Adds config to the root-level rubocop to ignore the db migration files for all sub-modules.

## Why?
In short, PR's were excluding `module/whatever/db/migration/*.rb` when running `rubocop`, whereas `master` CI would still run on these files. This meant that if a linting error was introduced in one of these files in a PR the CI would not catch it. Once merged however, the `master` CI `rubocop` would run on that file and fail. [This ultimately blocked a daily deploy](https://dsva.slack.com/archives/C0MQ281DJ/p1606763358227700?thread_ts=1606762827.226800&cid=C0MQ281DJ).

## Even More Context
- [I logged a bug with rubocop](https://github.com/rubocop-hq/rubocop/issues/9132) that goes into the gory details of how using `--force-exclusion` behaves differently than when not using it.
- PR's only lint the files that were changed in the PR - this was an optimization to speed up PR CI runs ([PR link](https://github.com/department-of-veterans-affairs/vets-api/pull/4036)).

## Original issue(s)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -22,6 +22,7 @@ AllCops:
     - db/migrate/*.rb
     - 'script/**/*'
     - 'vendor/**/*'
+    - modules/**/db/migrate/*.rb
 
 # This allows you to use have writers like self.method_name(name) vs self.method_name=(name)
 Style/TrivialAccessors:


erluti (approved):
thanks for creating the issue with rubocop too

SlakrHakr (approved):

PR #5418, opened 2020-12-01
changed value sorted by

jperk51:
## Description of change
Changed sort of claims overview values from filed date to updated date

## Original issue(s)
N/A

Small change to tests to match sorting change


jjulian:
@lexicalninja what do you think: should the app sort data or should the api do it?

jjulian (approved):

PR #5419, opened 2020-12-01
Accept latitude, longitude, and radius on ccp endpoint

darkbushido:
## Description of change

Due to some rounding errors in Mapbox the center of the search area isn't the center of the map. To fix this the endpoint will take a latitude, longitude, and radius. The V1 PPMS client will pass these directly on. the bbox params should be considered deprecated and it will be removed in the near future. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#16259

The CCP endpoint will now take `latitude`, `longitude`, and `radius` instead of `bbox`


thilton-oddball (approved):
All looks good. One comment, facility_locator_spec.rb is returning 0 examples, 0 failures, however, it was doing that before any changes so I'm not sure if there is an issue there or not.

darkbushido:
Yep, because they are shared examples included inside `spec/request/v1/apidoc/apidoc_spec.rb` hopefully this will keep the apidoc spec files size down and it won't become the giant it is in v0

PR #5420, opened 2020-12-01
13444 refactor okta grant caching

erluti:
This PR removes the use of Redis to cache connected applications (grants) for a user. (See "original issue" below for reason this change is needed.)

## Description of change
* the `all` method just use in-memory caching to prevent additional requests to okta during a single request to the api
* delete no longer needs to delete a cache key since the cache isn't used
* delete also refactored to not track a boolean since the logic breaks with the exception, not a `false`-y value.

## of interest
The `Grant` class should probably not inherit from `Model` which inherits from `Common::RedisStore` but the way this code is accessed is through the `App` service referencing some meta programming in the user class.  Untangling that is a bigger project, so feedback on other changes this PR should include are welcome.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#13444

### DIFF
--- a/app/models/okta_redis/grants.rb
+++ b/app/models/okta_redis/grants.rb
@@ -9,12 +9,11 @@ module OktaRedis
     CLASS_NAME = 'GrantsService'
 
     def all
-      okta_response.body
+      @all_grants ||= service.grants(@user.okta_profile.id).body
     end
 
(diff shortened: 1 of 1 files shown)


thilton-oddball (approved):
Looks good!

PR #5422, opened 2020-12-01
Update Puma to 4.3.7

ericboehs:
There's [a bug in Puma 4.3.5](https://github.com/puma/puma/issues/2304) which prevents some compilers from building the native extensions required for Puma.

I was running into this issue on macOS Big Sur. Updating fixed it.


erluti (approved):
👍

PR #5423, opened 2020-12-01
minor test improvement

erluti:
The `CacheAside` test may have been silently failing because the block contained the expected result.
These changes were made while discovering the reason for recently added apps not being shown in the `connected_applications` endpoint.


spec/request/connected_applications_request_spec.rb
```diff
@@ -27,5 +27,25 @@
         end
       end
     end
+
+    it 'returns no app after deleting an app' do
```
erluti, line 31: This is testing that calling index after a delete will not return a cached result.  But after this change https://github.com/department-of-veterans-affairs/vets-api/pull/5420 it's nothing we intend to do.  So I'm going to remove this test but leaving it in this pr history for posterity.

thilton-oddball (approved):
LGTM

ericboehs (approved):

ericboehs:
I'd squash and then rebase with master to get the tests passing.

PR #5425, opened 2020-12-02
Saml response errors

bramleyjl:
## Description of change

This change improves the `handle_callback_error` function to attempt to tease out more specific `status_detail` information in a failed SAML response to be logged to Sentry instead of logging them all together in a bundle, which does not provide much actionable data to go off of.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#15009

Unit testing was included for the new status_detail method and updated to include a status_detail test case for saml_callback method; manual testing was done with the logged error SAML response via an API client and by attempting to submit a SAML form to ID.ME with an expired timestamp.


ericbuckley (approved):
LGTM; I'll continue to think about how we might be able to test this on staging

thilton-oddball (approved):
Looks good!

PR #5432, opened 2020-12-03
16029 accept multiple institution codes

ajmagdub:
## Description of change

- Added new route to accept multiple facility codes and return the appointment request limits for each code
- Added test for new route and controller additions
- This PR deals with Vets API changes. Vets Website changes will also be needed to complete this feature and will be addressed in a vets-web PR.

## Original issue(s)
Use request limit check when determining which Express Care window to use [#16029](https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/16029)

This work is dependent on the completion of [VAOSR-1401](https://issues.mobilehealth.va.gov/browse/VAOSR-1401) which will include the facilities IDs to the VAMF endpoint response.


modules/vaos/app/services/vaos/systems_service.rb
```diff
@@ -64,6 +64,18 @@ def get_facility_limits(facility_id, type_of_care_id)
       end
     end
 
+    def get_facilities_limits(facility_ids, type_of_care_id)
+      with_monitoring do
+        url = "/var/VeteranAppointmentRequestService/v4/rest/direct-scheduling/patient/ICN/#{@user.icn}/request-limits"
```
LindseySaari, line 69: Can this url be moved into a method or even settings.yml?

grossb1: Good call, will move into a method.

saneshark (approved):
LGTM

PR #5437, opened 2020-12-04
adds handling for internally generated errors in login_stats method

bramleyjl:
## Description of change

Adds a check for an error code in `login_stats` and substitutes a default code if none is found to prevent unexpected exceptions without an error code from generating more errors.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#15883

manual testing with breakpoints of `login_stats` method with no exceptions and exceptions without error codes

### DIFF
--- a/app/controllers/v1/sessions_controller.rb
+++ b/app/controllers/v1/sessions_controller.rb
@@ -261,7 +261,7 @@ module V1
         StatsD.measure(STATSD_LOGIN_LATENCY, url_service.tracker.age, tags: tags)
       when :failure
         code = if error
-                 error.code
+                 error.code || '007'
                else
                  '007'
                end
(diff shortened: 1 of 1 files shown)


thilton-oddball (approved):
LGTM

app/controllers/v1/sessions_controller.rb
```diff
@@ -260,7 +260,12 @@ def login_stats(status, error = nil)
         Rails.logger.info("LOGIN_STATUS_SUCCESS, tags: #{tags}")
         StatsD.measure(STATSD_LOGIN_LATENCY, url_service.tracker.age, tags: tags)
       when :failure
-        tags_and_error_code = tags << "error:#{error.code}"
+        code = if error
```
bosawt, line 263: I can see this is done all over the place, so might not be something we should tackle with this PR, but it would be nice to refer to the error codes with descriptive values. I might recommend having these in `vets-api/lib/saml/responses/base.rb`, something like adding: `UNKNOWN_ERROR_CODE = '007'` at the top of that file, so in places like this we can refer to this string as: `SAML::Responses::Base::UNKNOWN_ERROR_CODE`

PR #5438, opened 2020-12-05
Fix for CH36 add veteranSocialSecurityNumber field

kathleencrawford:
## Description of change

This PR is to fix the CH36 form submission to CentralMail.  CentralMail expects a field called "veteranSocialSecurityNumber" to be at the root level of the hash/payload, so this PR is a fix to ensure that this value is set and in the appropriate location/level.
## Original issue(s)
department-of-veterans-affairs/va.gov-team#16219

This is a minor bug fix PR so no additional settings or logging was necessary.

## Testing 
- [x] Local testing done
- [x] Specs added
- [x] Staging testing planned


thilton-oddball (approved):
Looks good

PR #5439, opened 2020-12-05
initial commit of vetext covid-19 vaccine registry api service

saneshark:
## Description of change
Introduces a new VEText COVID-19 Vaccine Registration API

## Original issue(s)
N/A

This is a new service integration as part of vaccine distribution coordination


modules/vetext/app/models/vetext/v0/vaccine_registry.rb
```diff
@@ -0,0 +1,91 @@
...
+
+      attr_reader :user
+
+      def initialize(attributes, user = nil)
+        super(attributes.except(*UNASSIGNABLE_ATTRIBUTES))
+        @user = user || unauthenticated_user
+      end
+
+      def unauthenticated_user
+        # TODO: figure out how to fetch from MVI when unauthenticated_user
```
annaswims, line 38: you could do someting like this https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/models/health_care_application.rb#L121

annaswims: or, do you have EDIPI or ICN? 

this would do (though MPI might not be cool with it): 
MPI::Service.new.find_profile(OpenStruct.new(mhv_icn: the_icn, dslogon_edipi: the_edipi)).profile

saneshark: We would only have ICN if the user has already signed in, in which case we wouldn't need to do the lookup because we have everything we need already from SSOe or MVI. However we want to try to do an MVI lookup to provide ICN and other details when the user is NOT logged in as well, but we have to do it in a way that doesn't return success/failure, possibly even use a CAPTCHA or something to avoid abuse. The idea would be that we do the MVI lookup and send the appropriate attributes over based on a successful lookup of the attributes the user provides as part of the form... it would be an attribute based lookup in MVI.

modules/covid_vaccine/app/services/covid_vaccine/v0/vetext_configuration.rb
```diff
@@ -0,0 +1,42 @@
...
+        Settings.vetext.url
+      end
+
+      def service_name
+        'Vetext'
+      end
+
+      def connection
+        # TODO: Remove SSL no-verify option!
+        Faraday.new(base_path, headers: base_request_headers, request: request_options, ssl: { verify: false }) do |c|
```
patrickvinograd, line 20: We need to take out this `ssl: { verify: false }` option, waiting for current test run before I do that.

thilton-oddball (approved):
It all looks good!

PR #5440, opened 2020-12-06
Migrates functions to modules/covid_vaccine

patrickvinograd:
Module should be named based on feature, and to avoid possible
future integration with VEText as a notification service.

## Description of change
Duplicates `modules/vetext` into `modules/covid_vaccine`. Adds a bare-bones controller and RegistrationSubmission model/migration. 

I'm thinking there is a place within this module for a vetext_registry service and configuration, since that's reflective of the backend system, but the module itself should be named covid_vaccine.


annaswims (changes requested):
some changes in schema.rb are unrelated to migration. Probably because @johnpaulashenfelter deleted all migrations before he left 😢   so there there were some that got deleted before you got a chance to run them.  Your best bet is to start with a fresh database.

config/features.yml
```diff
@@ -21,6 +21,10 @@ features:
     actor_type: user
     description: >
       Allows user to access chapter 33 benefits direct deposit view/update APIs
+  covid_vaccine_registration:
+    actor_type: user
+    description: >
+      Toggles availability of covid vaccine registration form frontend and API.
```
ncksllvn, line 27: FYI that per a Slack thread I added a separate Flipper feature for the frontend app, https://github.com/department-of-veterans-affairs/vets-api/pull/5448

PR #5442, opened 2020-12-07
API-3036: 2122 (John Doe) Http 500 Error

SlakrHakr:
## Description of change
System is returning a generic 500 response if the endpoint consumer has not provided Veteran information in the auth headers and said consumer is not in MPI. This should resolve that and respond with a 400 response as expected.

## Original issue(s)

Tested locally

### DIFF
--- a/modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
@@ -93,9 +93,19 @@ module ClaimsApi
         def source_data
           {
             name: source_name,
-            icn: current_user.icn,
+            icn: nullable_icn,
             email: current_user.email
           }
+        end
+
+        def nullable_icn
+          current_user.icn
+        rescue => e
+          log_message_to_sentry('Failed to retrieve icn for consumer',
+                                :warning,
+                                body: e.message)
+
+          nil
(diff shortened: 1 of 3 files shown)


modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
```diff
@@ -93,11 +93,21 @@ def header_md5
         def source_data
           {
             name: source_name,
-            icn: current_user.icn,
+            icn: nullable_icn,
```
SlakrHakr, line 96: As far as I can tell, icn is only used within the PoaUpdater job which is not called if headers for a Veteran are passed in
https://github.com/department-of-veterans-affairs/vets-api/blob/49b0e6c5683a9e6e614c7759dc88a273891f3c4f/modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb#L40

And, if icn is needed, it looks like the application defaults to a setting if icn isnt present
https://github.com/department-of-veterans-affairs/vets-api/blob/49b0e6c5683a9e6e614c7759dc88a273891f3c4f/modules/claims_api/app/models/claims_api/power_of_attorney.rb#L130

bastosmichael (approved):
This is awesome, thanks for catching this error, good to have a different pair of eyes on this problem.

PR #5449, opened 2020-12-07
Rename missed PGD service file; fixes intermitent test failure

ericboehs:
Patrick Vinograd reported this issue this morning:
> the filename [needs to] match the class name, so that rails can load the class as needed. Otherwise it’s causing ‘rake spec’ to fail

Running `rake spec` locally would immediately fail with:
```
NameError:
  uninitialized constant HealthQuest::PatientGeneratedDataService
# ./modules/health_quest/spec/services/pgd_service_spec.rb:6
```


erluti (approved):
👍

PR #5450, opened 2020-12-08
need to set last_signed_in for mobile iam users

jperk51:
## Description of change
Needed to set last_signed_in for mobile iam users in order to user EVSS services

## Original issue(s)
No issue created, hot fix

Set last_signed_in and updated tests to check that the field is non-null

### DIFF
--- a/spec/lib/iam_ssoe_oauth/session_manager_spec.rb
+++ b/spec/lib/iam_ssoe_oauth/session_manager_spec.rb
@@ -19,7 +19,6 @@ describe 'IAMSSOeOAuth::SessionManager' do
       before do
         VCR.use_cassette('iam_ssoe_oauth/introspect_active') do
           @user = session_manager.find_or_create_user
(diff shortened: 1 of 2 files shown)


spec/lib/iam_ssoe_oauth/session_manager_spec.rb
```diff
@@ -19,6 +19,7 @@
       before do
         VCR.use_cassette('iam_ssoe_oauth/introspect_active') do
           @user = session_manager.find_or_create_user
+          expect(@user.last_signed_in).not_to be_nil
```
kreek, line 22: Could this go down into an `it` block? Might also want to check that it's a time in addition to not nil

jperk51: will do

kreek (approved):
LGTM 🚀

thilton-oddball (approved):
LGTM

PR #5452, opened 2020-12-08
Migrations for covid_vaccine schema

patrickvinograd:
## Description of change
DB migration and schema update for covid_vaccine module. This migration is tested in dev/review instances, extracted from this PR: https://github.com/department-of-veterans-affairs/vets-api/pull/5439/commits
so that we can apply it separately. 

## Original issue(s)

This feature does store PII in the `encrypted_form_data` table - this contains the raw submission to the covid vaccine registration endpoint. This storage is approved by product owner @batemapf. Vets-api is not the source of truth for this data, but we're retaining it in case we need to replay submissions to the underlying system.


saneshark (approved):
LGTM; working fine with application logic as well.

thilton-oddball (approved):
Looks good!

PR #5453, opened 2020-12-08
17086 mobile address validation

kreek:
## Description of change
Part of a series of PRs around address updates. Adds the address validation endpoint to the mobile API. Although the current PUT address update endpoint does address validation this endpoint also returns suggested addresses and returns a validationKey that needs to be sent along for address validation overrides.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#17086


jperk51 (approved):
LGTM

PR #5454, opened 2020-12-08
16834 mobile synchronous profile update service

kreek:
## Description of change
First of the profile updates to use the new synchronous profile update service. The service moves the polling a FE application would normally do to the backend to reduce the number HTTP requests an app has to make. The polling uses exponential backoff to to try five times over the first five seconds (plus Vet360 service latency) and then five more times a greater intervals or until the 55s timeout window is hit. Note there's a 60s timeout at the proxy level. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#16834

* logs transaction tries, success, and failures by transaction id


modules/mobile/app/services/mobile/v0/profile/sync_update_service.rb
```diff
@@ -0,0 +1,136 @@
...
+            retry
+          end
+        end
+
+        def get_elapsed(start)
+          seconds_since_epoch - start
+        end
+
+        def tries_or_time_exhausted?(sleep_duration, elapsed, try)
+          try == 9 || elapsed + sleep_duration > TIMEOUT_SECONDS
```
kreek, line 88: there isn't a running clock so the OR checks to see if the next sleep duration would exceed the timeout rather than go to sleep unnecessarily (and hit the proxy timeout).

jperk51 (approved):
LGTM

PR #5458, opened 2020-12-09
Fix pagerduty staging

patrickvinograd:
## Description of change
Allows pagerduty maintenance window configuration to work per-environment, by making the service query prefix configurable. In conjunction with a devops PR to set the new configuration to `Staging: External: ` this will make the maintenance window and backend statuses endpoints work correctly in staging. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#9308

New settings.yml value will vary per environment but the default value is appropriate for production. Dev and staging are currently broken but will be fixed by introduction of this config setting. 

Tested locally by overriding local settings. Re-recorded the main VCR cassette that verifies a successful query of external services.


LindseySaari (commented):
Does the new config prefix apply to both dev/staging? Looks like it does, but just want to double check

patrickvinograd:
Yes, this devops PR introduces the new prefix for dev/staging: https://github.com/department-of-veterans-affairs/devops/pull/8177

I know that frequently we make baseline settings.yml appropriate for dev/local development, but in this case I left the baseline to be prod-like so that these two PRs don't need to be coordinated. I don't think a lot of people are using maintenance windows in local env but there is frequently a desire to test them in staging.

LindseySaari (approved):

PR #5444, opened 2020-12-07
Create migrations for form1010cg_submissions table

kevinmirc:
Relates to: https://github.com/department-of-veterans-affairs/va.gov-team/issues/10280
Relates to: https://github.com/department-of-veterans-affairs/vets-api/pull/5447

**Description**:
Create the table and indices for Form 10-10CG submissions that will soon be persisted for background processing of attachments.

---
Next: https://github.com/department-of-veterans-affairs/vets-api/pull/5447

PR #5459, opened 2020-12-09
update claim_guid column on form1010cg_submissions as uuid type

kevinmirc:
**Description**
Made a migration yesterday https://github.com/department-of-veterans-affairs/vets-api/pull/5444 to create the `form1010cg_submissions` table but mistyped the data type for `claim_guid`. This removes the column, creates it again (as uuid), and adds the index.

**Original issue(s)**
https://github.com/department-of-veterans-affairs/va.gov-team/issues/10280


LindseySaari (approved):

kevinmirc:
Just realized I left the `disable_ddl_transaction!` call in the migration. This could be a problem since there are multiple statements in the single migration.

In the event that the migration fails on one or more statements, I can do a follow-up migration to drop the entire table and re-created it (since it's not yet in use).

PR #5461, opened 2020-12-09
Use the correct schema name and add items for type array

nwalker-va:
## Description of change

This is a PR for [API-142](https://vajira.max.gov/browse/API-142). The 'items' field was missing for type array on 'scp' and the schema name was incorrect (VeteranStatusConfirmation instead of ValidatedToken).

## Original issue(s)
https://github.com/department-of-veterans-affairs/vets-contrib/issues/3771

The acceptance criteria of using the correct path was already fixed by someone else in a separate PR.


LindseySaari (approved):
👍

PR #5462, opened 2020-12-09
Api 3998 turn off sending uploads to gdit

gpbowman-git:
## Description of change

Commenting out jobs that push uploads to GDIT starting on 12/10/2020 during a transition from GDIT to GovCIO for central mail.

## Original issue(s)
https://vajira.max.gov/browse/API-3998

### DIFF
--- a/modules/vba_documents/app/controllers/vba_documents/v1/uploads_controller.rb
+++ b/modules/vba_documents/app/controllers/vba_documents/v1/uploads_controller.rb
@@ -34,7 +34,7 @@ module VBADocuments
           submission.status = request.headers['Status-Override']
           submission.save
         else
-          puts 'remove me! The linter wants a line here'
+          Rails.logger.info 'remove me! The linter wants a line here'
           # submission.refresh_status! unless submission.status == 'expired'
         end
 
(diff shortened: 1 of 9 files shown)


modules/vba_documents/app/controllers/vba_documents/v1/uploads_controller.rb
```diff
@@ -34,7 +34,8 @@ def show
           submission.status = request.headers['Status-Override']
           submission.save
         else
-          submission.refresh_status! unless submission.status == 'expired'
+          Rails.logger.info 'remove me! The linter wants a line here'
```
LindseySaari, line 37: Is this PR ready for review or are you still debugging? Just noticed this line

LindseySaari:
Do you plan on reverting these changes at some point after the transition? I would recommend even deleting the code instead of commenting it out. If a revert is in the plans, it might be good to link to the ticket for that work in the PR description also

modules/vba_documents/app/controllers/vba_documents/v1/uploads_controller.rb
gpbowman-git: This is ready for review. These changes will be rolled back once the transition to GovCIO is complete.

bastosmichael (approved):
Yes our goal is to revert this completely once the GDIT to GCIO transition is complete, I do like the idea of removing instead of commenting out but for time purposes I'll approve this as is for now.

PR #5464, opened 2020-12-09
[API-3631] Fail gracefully on ITF submit calls for invalid VSRs

hinzed1127:
## Description of change

VSRs that are absent in MPI cannot submit ITFs for type `burial`. This rescues that error and provides a more accurate description of that error in the response.

## Original issue(s)
https://vajira.max.gov/browse/API-3631

### DIFF
--- a/modules/claims_api/app/controllers/claims_api/v1/forms/intent_to_file_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/v1/forms/intent_to_file_controller.rb
@@ -20,6 +20,9 @@ module ClaimsApi
           bgs_response = bgs_service.intent_to_file.insert_intent_to_file(intent_to_file_options)
           render json: bgs_response,
                  serializer: ClaimsApi::IntentToFileSerializer
+        rescue Common::Exceptions::Forbidden => e
+          render json: { errors: [{ status: 403, detail: e.errors[0]&.detail }] },
+                 status: :forbidden
         rescue Savon::SOAPFault => e
           error = {
             errors: [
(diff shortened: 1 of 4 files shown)


LindseySaari (approved):

modules/claims_api/app/controllers/claims_api/v1/forms/intent_to_file_controller.rb
```diff
@@ -20,6 +20,9 @@ def submit_form_0966
           bgs_response = bgs_service.intent_to_file.insert_intent_to_file(intent_to_file_options)
           render json: bgs_response,
                  serializer: ClaimsApi::IntentToFileSerializer
+        rescue Common::Exceptions::Forbidden => e
+          render json: { errors: [{ status: 403, detail: e.errors[0]&.detail }] },
+                 status: :forbidden
```
SlakrHakr, line 25: Is this even needed actually? Assuming we keep with Common::Exceptions::Forbidden as the exception, I would assume we're catching and returning a 403 at some point in the stack?

SlakrHakr: Ok, I just tried it out locally and yes, this specific exception is already handled at some point in the stack. Received a 403 status code along with the following response body:
```
{
    "errors": [
        {
            "title": "Forbidden",
            "detail": "Representative cannot file for type 'burial'",
            "code": "403",
            "status": "403"
        }
    ]
}
```
so i dont think we need this extra rescue here, unless we want some other than forbidden of course

hinzed1127: ahh, ok cool. that makes sense. I'm just adding the custom `detail` attribute and the rest of it should already get caught. will remove 👍

bastosmichael (approved):
Dealing with edge case business restrictions is never easy so not a perfect solution in my opinion but it's function over form and we can always refactor it later if need be.

SlakrHakr (approved):

PR #5469, opened 2020-12-09
Bundler: use path block for modules

ericboehs:
Use the path directive in a ruby block https://bundler.io/man/gemfile.5.html#PATH


erluti:
I'm a huge fan
👍

erluti (approved):

Gemfile.lock
```diff
@@ -73,82 +73,34 @@ GIT
...
-  remote: modules/vba_documents
-  specs:
     vba_documents (0.0.1)
       aws-sdk-s3 (~> 1)
       sidekiq
-
-PATH
-  remote: modules/veteran_confirmation
-  specs:
+    veteran (0.0.1)
```
LindseySaari, line 100: I'm thinking this was moved for alphabetizing purposes? Correct?

LindseySaari (approved):
💯

Gemfile.lock
ericboehs: Yeah. The diff looked a little weird here but this was all bundlers doing /shrug

PR #5471, opened 2020-12-10
upgrade pact gem and deps

LindseySaari:
## Description of change
This PR upgrades the pact gem and its dependencies to the latest v1.55.7. In [v1.55.2](https://github.com/pact-foundation/pact-ruby/releases/tag/v1.55.2), pact released a bug fix for correctly calculating the exit code when a mix of pending and non-pending pacts are verified. We're currently experiencing some odd behavior in our CircleCI verification job related to pending pacts.

## Testing


erluti (approved):
👍

PR #5474, opened 2020-12-10
BDD: Cache servicelocation

lihanli:
## Description of change
cache the response for the `/v0/disability_compensation_form/separation_locations` API for 24 hours. the data is the same for every user

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/10275

tested with unit tests


annaswims (approved):
LGTM

lib/evss/reference_data/response_strategy.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module EVSS
+  module ReferenceData
+    class ResponseStrategy < Common::RedisStore
+      include Common::CacheAside
+      redis_config_key :reference_data_response
+
+      def cache_or_service(user_uuid, type)
```
erluti, line 9: I think this name is a little confusing, since this method has nothing to do with the service.  I get that's the intended purpose, but used in context I was surprised the block initialized the service because I hadn't seen this method's implementation yet.
I'd propose something like `cache_by_user_and_type` because it makes the purpose clear and explains the arguments.  But there might be a better name.

erluti (approved):
Looks good, just some nitpicking of the method name

PR #5476, opened 2020-12-10
Pv covid vax facility lookup

patrickvinograd:
## Description of change
Adds facility lookup, based on provided zipcode, using the lighthouse facilities APIs. Attempt to find the nearest VAMC, taking into account some business logic around facility ID semantics. 

Along the way, adapted to some changes in the form data contract with the frontend as well
as adjusting what gets sent to the vetext API. 

## Original issue(s)


LindseySaari:
Could you add a description to this PR please?

config/initializers/zcta.rb
```diff
@@ -2,4 +2,7 @@
 
 require 'csv'
 
-ZCTA = CSV.read(Rails.root.join('lib', 'facilities', '2018_Gaz_zcta_national.txt'), col_sep: "\t")
+ZCTA = CSV.read(Rails.root.join('lib', 'facilities', '2019_Gaz_zcta_national.tsv'),
```
patrickvinograd, line 5: I changed ZCTA to a map, keyed by zip, just for convenience

app/models/facilities_query/zip_query.rb
```diff
@@ -5,10 +5,10 @@ class ZipQuery < Base
     def run
       # TODO: allow user to set distance from zip
       zip_plus0 = params[:zip][0...5]
-      requested_zip = ZCTA.select { |area| area[0] == zip_plus0 }
+      requested_zip = ZCTA[zip_plus0]
```
patrickvinograd, line 8: I'm not even sure this class is being used anymore, but changed it to reflect that ZCTA is now a map.

patrickvinograd:
BTW the LOC on this PR is due to swapping out the newer zipcode census data file `2018_Gaz_zcta_national.txt` which is 33k lines

saneshark (approved):
LGTM

modules/covid_vaccine/app/services/covid_vaccine/v0/facility_lookup_service.rb
```diff
@@ -0,0 +1,75 @@
...
+        sta3n = result.last if result.last.length == 3
+        sta6a = result.first if result.first.length > 3
+        sta6a = result.last if result.last.length > 3
+        {
+          sta3n: sta3n,
+          sta6a: sta6a
+        }
+      rescue
+        # For now just bail on any exception while getting facilities
+        {}
```
LindseySaari, line 60: Is there a plan to incorporate exceptions here eventually?

modules/covid_vaccine/app/services/covid_vaccine/v0/facility_lookup_service.rb
```diff
@@ -0,0 +1,75 @@
...
+        if result.blank?
+          # Does not seem feasible that a location would be closer to
+          # 30 clinics than any VAMCs
+          response = client.get_facilities(lat: lat, long: lng,
+                                           per_page: 30, type: 'health')
+          result = nearest_vamc(response.map { |x| x.id.delete_prefix('vha_') })
+        end
+        sta3n = result.last if result.last.length == 3
+        sta6a = result.first if result.first.length > 3
+        sta6a = result.last if result.last.length > 3
```
LindseySaari, line 53: Just making sure, if result.first.length > 3 and result.last.length > 3 are both true, sta6a should be set to result.last correct?

LindseySaari (approved):
Just a few questions [here](https://github.com/department-of-veterans-affairs/vets-api/pull/5476/files#diff-7de5e7a870fc85222e4ed035b49c0df6767297ce578d0de033391873c7abb39c), but going to approve to get this moving along for staging. Also, great comments... it made for an easier understanding of the code

modules/covid_vaccine/app/services/covid_vaccine/v0/facility_lookup_service.rb
patrickvinograd: The facility information is optional, in the sense that if we can't derive it, just passing through the user-supplied zipcode is a decent fallback. But yes we can refine this somewhat.

modules/covid_vaccine/app/services/covid_vaccine/v0/facility_lookup_service.rb
patrickvinograd: Yes...tried to describe the logic but that means that the nearest VAMC is one of the "consolidated facilities" with a > 3 length station ID, so it will be both first and last in the list, and need to go into the sta6a field.

PR #5465, opened 2020-12-09
Create an Sidekiq job for 10-10CG Attachments

kevinmirc:
Fixes: https://github.com/department-of-veterans-affairs/va.gov-team/issues/17204

**Migrations**:
- https://github.com/department-of-veterans-affairs/vets-api/pull/5444
- https://github.com/department-of-veterans-affairs/vets-api/pull/5459

**Description**
Create a Sidekiq job for sending the PDF of 10-10CG submissions to CARMA.

**Original issue(s)**
https://github.com/department-of-veterans-affairs/va.gov-team/issues/16952

**Summary** 
- [x] Create the Job
  - [x] Error handling in job
    - [x] Raise error if submission is not found
    - [x] Raise error if claim not found
    - [x] Raise error if PDF Generation fails
  - [x] Destroy models after job is complete
  - [x] Destroy PDF after job is complete
  - [x] Ensure idempotence
- [x] Update the `Form1010cg::Service`
  - [x] Create a method (`::submit_attachment!`) to process attachment and raise on failure

**Considerations**
- [x] How to use the `Form1010cg::Service` from attachment job?
  - Creating a stateless `Form1010cg::Service::submit_attachment!` method that just builds and sends the request (no validation on claim or submission. Required file path to be passed in.).
- [ ] Document and configure how many reties and when they happen

**Followup**
- [x] Add specs for `CARMA::Models::Attachments#submit` encountering a `400` response. https://github.com/department-of-veterans-affairs/va.gov-team/issues/17334
  - Update the specs in `Form1010cg::Submission` that expect `Faraday::ClientError`
- [x] Add specs to `Form1010cg::Submission` on network errors https://github.com/department-of-veterans-affairs/va.gov-team/issues/17334
  - Expected to raise error
- [x] Remove `#to_pdf` stubs in integration test to catch errors regarding the generation of the PDF. https://github.com/department-of-veterans-affairs/va.gov-team/issues/17335

---
Prev: https://github.com/department-of-veterans-affairs/vets-api/pull/5447 | Next: https://github.com/department-of-veterans-affairs/vets-api/pull/5477

PR #5477, opened 2020-12-10
Process Form 10-10CG attachments in background job

kevinmirc:
Epic: https://github.com/department-of-veterans-affairs/va.gov-team/issues/16952
Fixes: https://github.com/department-of-veterans-affairs/va.gov-team/issues/17205

**Description** 
This enabled the ability to process 10-10CG attachments using sidekiq through a new feature toggle `:async_10_10_cg_attachments`.

**Summary**
- [x] Based on the feature toggle, do PDF generation and attachment work asynchronously.
- [x] Log the attachment job id.

**Considerations**
- [x] Updates to logs: Since the attachments key in the "successful submission" log will be empty, we need a way to signify that it was queued.
  - Added "attachment_job_id" to "successful submission" logs
  - This will still allow us to identify dropped attachments via logs where`(attachments is {} or null) and (attachment_job_id is null)`
- [ ] Should feature toggle be cookie based?
  - Would like to do a percentage-based rollout in prod, but not sure how it gets the cookie_id to store the state in Flipper for any specific user

---
Prev: https://github.com/department-of-veterans-affairs/vets-api/pull/5465


LindseySaari (approved):
👍

kevinmirc:
Thanks for the review!

PR #5478, opened 2020-12-10
adds VANotify mailer for covid registration

saneshark:
## Description of change
Adds a mailer for covid vaccine registration using VANotify.

## Original issue(s)
none

Settings.yml updated with the new VAnotify, separate PR with template id already created for devops repo per env.
Should there be a feature flag for the mailer by itself?

Only Rspec testing so far, will want to verify on staging it work s as expected.


modules/covid_vaccine/app/services/covid_vaccine/v0/registration_service.rb
```diff
@@ -27,9 +27,17 @@ def submit(attributes, account_id)
         # TODO: error handling
         response = CovidVaccine::V0::VetextService.new.put_vaccine_registry(attributes)
         Rails.logger.info("Vetext Response: #{response}")
-        CovidVaccine::V0::RegistrationSubmission.create({ sid: response[:sid],
-                                                          account_id: account_id,
-                                                          form_data: attributes })
+        record = CovidVaccine::V0::RegistrationSubmission.create({ sid: response[:sid],
```
patrickvinograd, line 30: this method is now no longer returning the `record` object which the controller wants to use to serialize a response to the user.

patrickvinograd (changes requested):
See comment, need to return registration object to controller

modules/covid_vaccine/app/services/covid_vaccine/v0/registration_service.rb
saneshark: doh... how come no specs are failing though Mr. Vinograd. 😂

saneshark: my bad, it was my specs for create that were not verifying anything.

patrickvinograd:
I created https://github.com/department-of-veterans-affairs/vets-api/pull/5486 in the hopes that we could get that merged easily and then the rest of this work is contained within the covid_vaccine module

modules/covid_vaccine/app/services/covid_vaccine/v0/registration_service.rb
patrickvinograd: I updated some specs in a since-merged PR so maybe you can upmege and re-test and see how it does

patrickvinograd (approved):
LGTM!

PR #5479, opened 2020-12-10
Clean up appeals_api base pdf constructor

nathanbwright:
This PR addresses [API-3830](https://vajira.max.gov/browse/API-3830)

## Description of change
This updates the `BasePdfConstructor#merge_page` method so both the HLR and NOD subclasses can use it without having to override it.

### DIFF
--- a/modules/appeals_api/lib/appeals_api/higher_level_review_pdf_constructor.rb
+++ b/modules/appeals_api/lib/appeals_api/higher_level_review_pdf_constructor.rb
@@ -87,7 +87,7 @@ module AppealsApi
           end
         else
           text = "Issue: #{issue['attributes']['issue']} - Decision Date: #{issue['attributes']['decisionDate']}"
-          options[:additional_page] = "#{text}\n#{options[:additional_page]}"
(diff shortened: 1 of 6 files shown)


modules/appeals_api/lib/appeals_api/higher_level_review_pdf_constructor.rb
```diff
@@ -87,7 +87,7 @@ def pdf_options
           end
         else
           text = "Issue: #{issue['attributes']['issue']} - Decision Date: #{issue['attributes']['decisionDate']}"
-          options[:additional_page] = "#{text}\n#{options[:additional_page]}"
+          options[:additional_pages] = "#{text}\n#{options[:additional_page]}"
```
cilestin, line 90: What is `options[:additional_page]` in the assignment string doing in this case? Feels like it's just blank all the time now.

If I'm reading this right, in the case where we have 10 contestable issues, only the 10th issue will show up on the additional page and we'll lose the 6th to 9th issues since it's a straight assignment and not pushing the issue text onto the array element.

cilestin (changes requested):

cilestin (approved):
LGTM! 🎉

PR #5480, opened 2020-12-10
Fixing whitespace code for Dangerfile

thilton-oddball:
## Description of change
The code that was added to not count whitespace when determining lines of code changed was not working properly and thus dangerbot was not alerting when PRs exceeded the LOC limit. This change should fix that issue.


LindseySaari (approved):
💯

PR #5481, opened 2020-12-10
Fix vetext params and request method

patrickvinograd:
## Description of change
Fixes the method used for submissions to VEText, and reconciles the expected form data
parameters with what the frontend is sending.

This is a subset of https://github.com/department-of-veterans-affairs/vets-api/pull/5476/files
without the facility API changes to facilitate a quicker fix for current staging env. 

## Original issue(s)


saneshark (approved):
This works for me

PR #5482, opened 2020-12-11
Exclude .tsv (tab-separated-value) from Dangerbot

patrickvinograd:
## Description of change
Added a .tsv in place of a .txt file in https://github.com/department-of-veterans-affairs/vets-api/pull/5476/files
to better represent the format. This excludes that datafile from Dangerbot checks.

## Original issue(s)


thilton-oddball (approved):
LGTM

PR #5484, opened 2020-12-11
Adds serialized submission attributes to post/get

patrickvinograd:
## Description of change
Serializes the submitted/persisted covid_vaccine form submission attributes back as
a response to the `GET` operation. 

- The GET operation does now return PII but will only return a result to an LOA3 user. Anonymous or LOA1 users will not receive any of those attributes. 
- The POST operation only returns a summary of the submission - created at time, vaccine interest, and zip code


LindseySaari (approved):
👍

PR #5485, opened 2020-12-11
Instrument registration flow

patrickvinograd:
## Description of change
Adds Statsd instrumentation around the potentially expensive parts of the registration flow -
the MPI query and the vetext call. 
(Facilities query instrumentation added in that PR). 

Tested locally and verified stats measure and increment calls appearing in logs.

### DIFF
--- a/modules/covid_vaccine/config/initializers/statsd.rb
+++ b/modules/covid_vaccine/config/initializers/statsd.rb
@@ -13,9 +13,7 @@ CovidVaccine::V0::RegistrationService.statsd_measure :attributes_from_mpi,
 CovidVaccine::V0::RegistrationService.statsd_measure :submit,
                                                      'covid_vaccine.vetext_submit.measure'
 CovidVaccine::V0::RegistrationService.statsd_count_success :attributes_from_mpi,
-                                                           'covid_vaccine.mpi_query' do |result|
-  result.present?
-end
+                                                           'covid_vaccine.mpi_query', &:present?
 CovidVaccine::V0::RegistrationService.statsd_count_success :submit,
                                                            'covid_vaccine.vetext_submit'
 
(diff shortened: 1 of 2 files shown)


modules/covid_vaccine/config/initializers/statsd.rb
```diff
@@ -6,3 +6,18 @@
...
+                                                     'covid_vaccine.mpi_query.measure'
+CovidVaccine::V0::RegistrationService.statsd_measure :submit,
+                                                     'covid_vaccine.vetext_submit.measure'
+CovidVaccine::V0::RegistrationService.statsd_count_success :attributes_from_mpi,
+                                                           'covid_vaccine.mpi_query', &:present?
+CovidVaccine::V0::RegistrationService.statsd_count_success :submit,
+                                                           'covid_vaccine.vetext_submit'
+
+StatsD.increment('covid_vaccine.mpi_query.success', 0)
+StatsD.increment('covid_vaccine.mpi_query.failure', 0)
```
LindseySaari, line 21: I've seen in other places that we've been defining STATSD_KEY_PREFIX on some of the service classes for statsd keys. In this case, you could create a constant for covid_vaccine.mpi_query and covid_vaccine.vetext_submit

LindseySaari (approved):
Left a comment [here](https://github.com/department-of-veterans-affairs/vets-api/pull/5485/files#r540947036) but overall, looks good!

modules/covid_vaccine/config/initializers/statsd.rb
patrickvinograd: Since it's all restricted to one place in this statsd file right now, I did not see much value in defining a constant just to use it twice immediately below, but if this winds up spanning out into other files much further I'll definitely refactor that.

PR #5486, opened 2020-12-11
Adds setting stub for covid vaccine vanotify use

patrickvinograd:
## Description of change
Adds the baseline settings.yml value for covid vaccine module's use of VANotify. The env-specific values are already set in the devops repo, and the implementation work is restricted to a codeowners module so can happen in a separate PR:

https://github.com/department-of-veterans-affairs/vets-api/pull/5478/files


LindseySaari (approved):
💯

PR #5487, opened 2020-12-11
fix example and update additional doc around burial

SlakrHakr:
## Description of change
Additional endpoint schema documentation for burial needed to be updated. Also removed participant_id from v1 0966 post endpoint. 

## Original issue(s)
https://vajira.max.gov/browse/API-3631

Viewed in developer-portal locally


hinzed1127 (approved):
lgtm

PR #5491, opened 2020-12-11
Adds some audit logging of covid vaccine submission

patrickvinograd:
## Description of change
Adds audit logging of submissions; sensitive fields are not logged, just the boolean presence/absence of them is logged.


modules/covid_vaccine/app/services/covid_vaccine/v0/registration_service.rb
```diff
@@ -45,6 +47,21 @@ def submit(attributes)
         CovidVaccine::V0::VetextService.new.put_vaccine_registry(attributes)
       end
 
+      def audit_log(attributes, user_type)
```
patrickvinograd, line 50: Maybe there's a more elegant way of doing this but I wanted to be explicit about what attrs were being logged and booleanized.

saneshark: you could use Hash#slice but it looks like you're casting some of them boolean so I think what you're doing makes more sense and is more readable.

saneshark (approved):
LGTM

PR #5492, opened 2020-12-11
add csv files to dangerbot

erluti:
csv files are for data and shouldn't block PRs (but they should still be reviewed)

also made a few style changes while I was in there


LindseySaari (approved):
Nice cleanup! Better readability!

PR #5494, opened 2020-12-11
ignores missing id.me uuid for v0 and inbound v1 logins [#17306]

bramleyjl:
## Description of change
In an effort to reduce the amount of non-necessary errors logged to sentry, we are only raising and logging an error for missing idme_uuids when the login request is:
* outbound
* using v1 sessions_controller

This change prevents an error from being raised if the idme_uuid is missing and either the login request is inbound or the sessions_controller for the request is v0.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#17306

* Unit tests updated to include passing/failing cases for `should_raise_idme_uuid_error`
* Unit tests for other sessions_controller methods updated where new inclusion of `saml_response.settings` is needed
* Manual testing with hard-coded missing_uuid values for inbound and outbound logins

### DIFF
--- a/lib/saml/user_attributes/ssoe.rb
+++ b/lib/saml/user_attributes/ssoe.rb
@@ -16,7 +16,7 @@ module SAML
 
       attr_reader :attributes, :authn_context, :warnings
 
-      def initialize(saml_attributes, authn_context, saml_settings)
+      def initialize(saml_attributes, authn_context, saml_settings = nil)
         @attributes = saml_attributes # never default this to {}
         @authn_context = authn_context
         @saml_settings = saml_settings
@@ -199,9 +199,11 @@ module SAML
       private
 
       def should_raise_idme_uuid_error
-        is_v1 = @saml_settings.assertion_consumer_service_url =~ %r{/v1/}
-        if !idme_uuid && is_v1 && (@authn_context != INBOUND_AUTHN_CONTEXT)
-          true
+        if !idme_uuid
+          if @saml_settings.nil? ||
+             @saml_settings.assertion_consumer_service_url =~ %r{/v1/} && (@authn_context != INBOUND_AUTHN_CONTEXT)
+            true
+          end
         else
           false
         end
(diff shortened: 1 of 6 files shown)


joeniquette (approved):

lib/saml/user_attributes/ssoe.rb
```diff
@@ -197,6 +198,17 @@ def validate!
 
       private
 
+      def should_raise_idme_uuid_error
```
bosawt, line 201: I suggest we clean up the logic here a bit, make it look more 'rubyish':

> def should_raise_idme_uuid_error
      return false unless idme_uuid
    @saml_settings.nil? ||
    @saml_settings.assertion_consumer_service_url =~ %r{/v1/} && (@authn_context != INBOUND_AUTHN_CONTEXT)
end

lib/saml/user_attributes/ssoe.rb
```diff
@@ -16,9 +16,10 @@ class SSOe
 
       attr_reader :attributes, :authn_context, :warnings
 
-      def initialize(saml_attributes, authn_context)
+      def initialize(saml_attributes, authn_context, saml_settings = nil)
```
bosawt, line 19: Ah, I was gonna say we need to make this change to all of the user attributes classes, but it looks like doing it here and in base takes care of that!

lib/saml/user_attributes/ssoe.rb
```diff
@@ -197,6 +198,17 @@ def validate!
 
       private
 
+      def should_raise_idme_uuid_error
+        if !idme_uuid
+          if @saml_settings.nil? ||
+             @saml_settings.assertion_consumer_service_url =~ %r{/v1/} && (@authn_context != INBOUND_AUTHN_CONTEXT)
```
bosawt, line 204: This is purely a matter of taste, but I like taking complex conditionals and creating private functions with descriptive names in this case, it'd look something like this:
`if @saml_settings.nil? || auth_context_is_v1_and_not_inbound`
Then the function

> def auth_context_is_v1_and_not_inbound
@saml_settings.assertion_consumer_service_url =~ %r{/v1/} && (@authn_context != INBOUND_AUTHN_CONTEXT)
end

Just an idea to think about

bosawt (commented):
Good job! Just a recommendation for clearing up the logic a bit but looks good!

LindseySaari (approved):

PR #5496, opened 2020-12-11
POST questionnaire responses for an appointment

dillo:
## Description of change

- The ability for the BE to receive questionnaire response data from the FE for a given appointment and then create a QuestionnaireResponse resource using this data in the PGD by POSTing the data to the downstream service.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#15006

- Feature flag: show_healthcare_experience_questionnaire

- [x] rspec tests written


modules/health_quest/app/services/health_quest/patient_generated_data/common/identity_meta_info.rb
```diff
@@ -0,0 +1,109 @@
...
+        #
+        CODING_SYSTEM = 'https://pki.dmdc.osd.mil/milconnect'
+        ##
+        # Operation Resource meta tag for the FHIR::DSTU2::Meta object
+        #   VA URI
+        META_SYSTEM = 'https://wiki.mobilehealth.va.gov/display/PGDMS/Client+Provenance+Mapping'
+        ##
+        # Operation Resource meta tag for the FHIR::DSTU2::Meta object
+        #   VA identifier
+        META_CODE = 'vagov-a0e116eb-faa1-4703-aafe-1a270128607a'
```
LindseySaari, line 22: Just checking if the values stores in the above constants are something that would be better suited for credstash?

modules/health_quest/app/services/health_quest/patient_generated_data/patient/resource.rb
```diff
@@ -66,59 +54,34 @@ def prepare
...
+        # @return [String]
         #
-        def name
-          [{
-            use: NAME_USE,
-            family: [user.last_name],
-            given: [user.first_name]
-          }]
+        def identifier_code
+          'ICN'
```
LindseySaari, line 84: Will this always be ICN?

dillo: @LindseySaari on a Patient resource I believe that the identifier is unique. However, I'm not sure going forward into the future and on version 4 of the FHIR spec, if that would still be the same.

modules/health_quest/app/services/health_quest/patient_generated_data/common/identity_meta_info.rb
dillo: That's a really good idea! I'd like to wait for a little bit until I've implemented the upcoming Questionnaire resource story before I proceed with such a refactor, using its own PR.

LindseySaari: Ok great, definitely a good refactor for a future PR

LindseySaari (approved):
Nice comments throughout the files, made for an easier review 👍

PR #5499, opened 2020-12-12
COVID module asynchronous submission processing

patrickvinograd:
## Description of change
Updates covid_vaccine submission to be asynchronous. 
Saves raw submission to db, then spawns worker to perform the MPI and facilities queries and vetext submision. This means submission from frontend is much faster and decoupled from
failures in any backend systems. 

Return value of POST is unchanged, except that the `id` attribute will always be blank. 

Return value of GET changes slightly in that the fields will be closer to what was submitted rather than reflecting what was posted to vetext. There is still possibility that for LOA3 users, if first/last/dob/etc in current_user are different than what was submitted, that's what will get returned. 

Note DB migration is being submitted separately as:
https://github.com/department-of-veterans-affairs/vets-api/pull/5500
will rebase once that's merged to clear dangerbot.

Tested locally and confirmed POST endpoint and sidekiq worker processing.

### DIFF
--- a/modules/covid_vaccine/app/services/covid_vaccine/v0/facility_lookup_service.rb
+++ b/modules/covid_vaccine/app/services/covid_vaccine/v0/facility_lookup_service.rb
@@ -57,6 +57,7 @@ module CovidVaccine
         }
       rescue
         # For now just bail on any exception while getting facilities
+        # TODO Add Sentry logging
         {}
       end
 
--- a/modules/covid_vaccine/spec/factories/registration_submissions.rb
+++ b/modules/covid_vaccine/spec/factories/registration_submissions.rb
@@ -21,18 +21,18 @@ FactoryBot.define do
         sta6a: '648GI'
       }
     }
-     
+
     raw_form_data {
       {
-        'vaccine_interest' => 'INTERESTED', 
+        'vaccine_interest' => 'INTERESTED',
         'zip_code' => '97212',
         'zip_code_details' => 'YES',
         'phone' => '808-555-1212',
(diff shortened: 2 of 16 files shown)


ncksllvn (approved):

patrickvinograd:
One TODO I thought of is to make the job more idempotent by ensuring that if sid is already set, it does not re-process.

modules/covid_vaccine/app/services/covid_vaccine/v0/facility_lookup_service.rb
```diff
@@ -57,6 +57,7 @@ def nearest_facilities(lat, lng)
         }
       rescue
         # For now just bail on any exception while getting facilities
+        # TODO Add Sentry logging
```
LindseySaari, line 60: Just checking if you want to add Sentry logging or if it's ok to move forward without the logging for the time being?

modules/covid_vaccine/spec/factories/registration_submissions.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 FactoryBot.define do
-  factory :covid_vaccine_registration_submission, class: 'CovidVaccine::V0::RegistrationSubmission' do
+  factory :covid_vax_registration, class: 'CovidVaccine::V0::RegistrationSubmission' do
```
LindseySaari, line 4: Is there a specific reason for renaming this from covid_vaccine_registration_submission?

LindseySaari (approved):
A few small questions/comments, but nothing blocking 👍

PR #5500, opened 2020-12-13
Adds migrations to allow async covid registration

patrickvinograd:
## Description of change
Updates covid_vaccine schema in preparation for the changes in this PR to support asynchronous form processing:
https://github.com/department-of-veterans-affairs/vets-api/pull/5499

- Adds an encrypted raw_form_data field for the submission received from the user
- Makes sid nullable, so that we can create a record before submitting to the vetext API

## Original issue(s)

* raw_form_data column may contain PII but is encrypted

* Tested locally with async submission process


ncksllvn (approved):

bastosmichael (approved):
LGTM

PR #5503, opened 2020-12-14
16098 update the v1 mashup endpoint to use the v1 ppms client

darkbushido:
## Description of change

When I added the v1 client I did not implement it for the mashup controller. This adds support for the V1 client.
There is an issue with the PPMS data in AZ. searching by street addresses instead of just street returns no results. searching by lat/long with a higher precision also finds no results. To handle this issue I removed the AZ results and updated all of the tests to use the NJ location.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#16098

The mashup controller will also use `facility_locator_ppms_use_v1_client` to decide which version of the client to use.


LindseySaari:
Does PPMS know about the issue with the AZ data?

darkbushido:
Not yet, it's something I'm going to bring up in our call tomorrow.

LindseySaari:
Are the failing specs related to your new changes? I can re-review once the specs are passing

darkbushido:
the two specs that are failing are `spec/workers/form1010cg/deliver_pdf_to_carma_job_spec.rb:191` and `modules/appeals_api/spec/workers/higher_level_review_pdf_submit_job_spec.rb:119`
They are unrelated to my changes.

LindseySaari (approved):
👍

PR #5504, opened 2020-12-14
API-3264: Special Issues - 526 JSON Schema update (include schema inside of disability objects)

SlakrHakr:
## Description of change
Ensure all valid special issues are allowed through the API

## Original issue(s)
https://vajira.max.gov/browse/API-3264

Specs for valid/invalid special issues through the API seem to already exist:
https://github.com/department-of-veterans-affairs/vets-api/blob/42b7769ffba861692eb5bdabfceb0b148c2bb52c/modules/claims_api/spec/requests/v0/disability_compensation_request_spec.rb#L127
https://github.com/department-of-veterans-affairs/vets-api/blob/42b7769ffba861692eb5bdabfceb0b148c2bb52c/modules/claims_api/spec/requests/v1/disability_compensation_request_spec.rb#L156


bastosmichael (changes requested):
Do we have a list of what these all mean that we can attach either in PDF form with a link in the descriptions or something so that it's not at all confusing to the end consumer what these all mean?

SlakrHakr:
Human readable items have been swapped in and mapper between those and internal codes was added as well.

app/swagger/swagger/schemas/form526/disability.rb
```diff
@@ -82,15 +82,166 @@ class Disability
...
-              MST
+            [
+              'ALS',
+              'HEPC',
+              'POW',
+              'PTSD/1',
+              'PTSD/2',
+              'PTSD/3',
+              'PTSD/4',
+              'MST',
```
SlakrHakr, line 93: these are the special issues that are already published out to production

modules/claims_api/lib/claims_api/special_issue_mapper.rb
```diff
@@ -0,0 +1,190 @@
...
+  class SpecialIssueMapper
+    SPECIAL_ISSUES = [
+      { name: 'ALS', code: 'ALS' },
+      { name: 'HEPC', code: 'HEPC' },
+      { name: 'POW', code: 'POW' },
+      { name: 'PTSD/1', code: 'PTSD/1' },
+      { name: 'PTSD/2', code: 'PTSD/2' },
+      { name: 'PTSD/3', code: 'PTSD/3' },
+      { name: 'PTSD/4', code: 'PTSD/4' },
+      { name: 'MST', code: 'MST' },
```
SlakrHakr, line 13: these are the special issues that are already published out to production

bastosmichael (approved):
LGTM

PR #5507, opened 2020-12-14
[17323] Improving the SSN Mismatch logs on the session controllers so…

bosawt:
… they now differentiate between the controller versions

## Description of change
This PR updates the SSN Mismatch log in the Sessions Controller v0/v1 routes, so it is more descriptive and differentiates between the controller version. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team/issues/17323

- This may change Sentry logs around the SSN mismatch, so we may want to watch for this and merge the Sentry log before/after the change

- I just ran specs, to confirm. I didn't attempt to recreate the SSN mismatch error, but shouldn't be a big deal since this is only a change in logs. We'll be able to easily see the change in Sentry Staging environment


joeniquette (approved):

erluti (approved):
👍

PR #5513, opened 2020-12-14
add reference as one of the attributes we send to VANotify

saneshark:
## Description of change
Note: I've only verified that VaNotify service receives this reference attribute and that it sends it to Notification gem. I have not done an end to end verification that the reference attribute can actually get serialized and gets sent to VaNotify. 

## Original issue(s)
On going work for Covid-Vaccine... we want to be able to pass the confirmation id to VANotify so that we can later go to them with that confirmation_id if any reconciliation is needed to verify why or why not an email was sent.

Tagging @philherbert since he is most familiar with the notification gem and VaNotify as an integration on vets-api.

### DIFF
--- a/modules/covid_vaccine/spec/jobs/covid_vaccine/registration_email_job_spec.rb
+++ b/modules/covid_vaccine/spec/jobs/covid_vaccine/registration_email_job_spec.rb
@@ -20,7 +20,7 @@ RSpec.describe CovidVaccine::RegistrationEmailJob, type: :worker do
     end
 
     it 'executes perform' do
-      instance = instance_double(VaNotify::Service, send_email: nil)
+      instance = instance_double(VaNotify::Service, send_email: { id: '123456789' })
       allow(VaNotify::Service).to receive(:new).and_return(instance)
 
       expect(instance)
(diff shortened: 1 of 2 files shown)


philherbert (approved):
Looks good to me! Thanks

stephenBarrs (approved):
Makes sense and looks good to me.

modules/covid_vaccine/spec/jobs/covid_vaccine/registration_email_job_spec.rb
```diff
@@ -20,7 +20,7 @@
     end
 
     it 'executes perform' do
-      instance = instance_double(VaNotify::Service, send_email: nil)
+      instance = instance_double(VaNotify::Service, send_email: { id: '123456789' })
```
erluti, line 23: should this id be a variable or a `let`?

erluti (approved):
LGTM

modules/covid_vaccine/spec/jobs/covid_vaccine/registration_email_job_spec.rb
saneshark: it should probably be more of a: `expect_any_instance_of(VaNotify::Service).to_receive(:send_mail).with(...).and_return({ id: '123456789' }).once`

PR #5516, opened 2020-12-15
Log when Form 10-10CG Attachment Job is successful

kevinmirc:
Fixes: https://github.com/department-of-veterans-affairs/va.gov-team/issues/17468

**Summary**
- [x] Update `Form1010cg::Auditor` to accept a logger
- [x] Update reference of `Form1010cg::Auditor`
- [x] Add a Logging event for **Attachments Delivered**
- [x] Log when `Form1010cg::DeliverPdfToCARMAJob` succeeds

### DIFF
--- a/app/services/form1010cg/auditor.rb
+++ b/app/services/form1010cg/auditor.rb
@@ -2,7 +2,7 @@
 
 module Form1010cg
   class Auditor
-    include Singleton
+    attr_reader :logger
 
     STATSD_KEY_PREFIX   = 'api.form1010cg'
     LOGGER_PREFIX       = 'Form 10-10CG'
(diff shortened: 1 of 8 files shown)


app/services/form1010cg/auditor.rb
```diff
@@ -25,6 +25,10 @@ def self.metrics
       )
     end
 
+    def initialize(logger)
```
erluti, line 28:
```suggestion
    def initialize(logger = Rails.logger)
```
Just a suggestion, but this way the auditor can accept other loggers, but in the common uses it doesn't need too explicitly add it.

app/services/form1010cg/auditor.rb
```diff
@@ -2,7 +2,7 @@
 
 module Form1010cg
   class Auditor
-    include Singleton
```
erluti, line 5: 💯

erluti (approved):
Looks good!

PR #5517, opened 2020-12-15
Adding basic validations to the CovidVaccine form data

saneshark:
## Description of change
This introduces validations on `%w[email zip_code vaccine_interest]`.

email validation is performed using REGEX from `URI::MailTo::EMAIL_REGEXP`
zip_code validation is performed using REGEX = /\A^\d{5}(-\d{4})?$\z/
vaccine_interest is just a non blank, presence check.

### DIFF
--- a/modules/covid_vaccine/spec/request/covid_vaccine/v0/registration_request_spec.rb
+++ b/modules/covid_vaccine/spec/request/covid_vaccine/v0/registration_request_spec.rb
@@ -52,7 +52,14 @@ RSpec.describe 'Covid Vaccine Registration', type: :request do
     end
 
     context 'when encountering an Internal Server Error' do
-      let(:registration_attributes) { { date_vaccine_reeceived: '' } }
+      let(:registration_attributes) do
+        {
+          vaccine_interest: 'yes',
+          email: 'jane.doe@email.com',
+          zip_code: '94402',
+          date_vaccine_reeceived: ''
+        }
+      end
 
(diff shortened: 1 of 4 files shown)


modules/covid_vaccine/spec/request/covid_vaccine/v0/registration_request_spec.rb
```diff
@@ -52,7 +52,14 @@
     end
 
     context 'when encountering an Internal Server Error' do
-      let(:registration_attributes) { { date_vaccine_reeceived: '' } }
+      let(:registration_attributes) do
+        {
+          vaccine_interest: 'yes',
+          email: 'jane.doe@email.com',
+          zip_code: '94402',
+          date_vaccine_reeceived: ''
```
stephenBarrs, line 60: Not sure if it matters here but FE is no longer collecting the date a vaccine was received.

stephenBarrs (approved):
This all looks good to me and makes sense, with caveat that I am not a seasoned BE dev!

erluti (approved):
LGTM

modules/covid_vaccine/spec/request/covid_vaccine/v0/registration_request_spec.rb
saneshark: yeah, I think that's just for a dumb spec to test what happens when a 500 error returns -- could be any misspelled field that we send to VEText.

PR #5520, opened 2020-12-15
Create Module Generator

LindseySaari:
## Description of change
This PR creates a new module generator that stands as a first step for creating a new module in vets-api. 

The generator creates the basic rails directory structure and leverages templates to create the base files required when getting started with a new module.

This PR also removes the current module generator. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#17215

## Acceptance Criteria
- [x] Empty directories are created with new generator
- [x] Engine file is created
- [x] Engine is loaded in Gemfile
- [x] Rakefile is generated 
- [x] Readme file is generated
- [x] bin/rails file is generated
- [x] spec helper add group
- [x] simplecov add group
- [x] ./spec/spec_helper
- [x] routes file is generated
- [x] gemspec file is generated
- [x] Gemfile is generated
- [x] run bundle
- [x] specs written & passing 

### Existing Generator
- [x] Remove the existing module generator


LindseySaari:
The current method of inserting the module into the gemfile conflicts with [This PR](https://github.com/department-of-veterans-affairs/vets-api/pull/5469/files). Will have to update the after option similar to the example [here](https://rubydoc.info/github/wycats/thor/master/Thor/Actions)

lib/generators/module/templates/gemspec.erb
```diff
@@ -3,11 +3,11 @@
 $LOAD_PATH.push File.expand_path('lib', __dir__)
 
 # Maintain your gem's version:
-require '<%= class_name.downcase %>/version'
+require '<%= class_name.underscore %>/version'
 
 # Describe your gem and declare its dependencies:
 Gem::Specification.new do |spec|
-  spec.name        = '<%= class_name.downcase %>'
+  spec.name        = '<%= class_name.underscore %>'
```
erluti, line 10: should this be downcase and underscore?

spec/generator/modules_generator_spec.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Modules', type: :generator do
```
erluti, line 5: should this file be removed since the tests are in `spec/lib/generators/module/module_generator_spec.rb`?

erluti (commented):
a couple questions

lib/generators/module/templates/gemspec.erb
LindseySaari: "Makes an underscored, lowercase form from the expression in the string." https://apidock.com/rails/String/underscore

spec/generator/modules_generator_spec.rb
LindseySaari: Yes! Thanks for pointing this out

lib/generators/module/templates/gemspec.erb
erluti: I should have checked first!

lib/generators/module/USAGE
```diff
@@ -1,23 +1,23 @@
 Description:
-    Creates a Rails engine starter app that wires up a service backed endpoint.
+    This generator will create an initial file structure and engine file for VFS developers.  The purpose of this being so they can quickly start developing code in an isolated file structure exclusive to the new business logic thay are creating.
+
+When a generator is invoked, each public method in the generator is executed sequentially in the order that it is defined.
 
 Example:
-    rails va new foo
+    rails generate module Thing
```
erluti, line 7: i don't love the name "Thing" for the example, but I never liked "foo" or "HelloWorld" either.  Personally I'd pick something like "CoolNewVAFeature" but I'm really just bikeshedding here.  Feel free to ignore this, lol

lib/generators/module/module_generator.rb
```diff
@@ -5,49 +5,66 @@
...
+    %w[controllers models serializers service].each do |dir|
+      FileUtils.mkdir_p "modules/#{file_name}/app/#{dir}" unless Dir.exist?("modules/#{file_name}/app/#{dir}")
+    end
   end
 
-  def create_lib
+  def create_engine
+    # create engine file
     path = "modules/#{file_name}/lib"
     template 'lib/namespace/engine.rb.erb', File.join(path, file_name, 'engine.rb')
```
erluti, line 18: Does this support someone typing `rails generate module potato` and then the engine class properly name it `Potato`?  The template doesn't seem to change the casing.  I think we should consider lazy users.  Now that I'm typing, that should be a new ticket for iterating on this.

erluti (approved):
nice!

lib/generators/module/module_generator.rb
LindseySaari: Correct, the engine template will use the class_name and name it Potato https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/generators/module/templates/lib/namespace/engine.rb.erb#L3

The above file was unchanged, so that's why it's not in the changed files

PR #5523, opened 2020-12-15
Tightens up guard around covid vaccine MPI query

patrickvinograd:
## Description of change
Adds check that date is valid enough to be turned into an MPI query,
since it is possible for a partial date like '2000-12-XX" to be submitted.

Solves this Sentry issue: http://sentry.vfs.va.gov/organizations/vsp/issues/11421/events/523bfe0e64124ad989fbed943b7ee779/?query=is%3Aunresolved+CovidVaccine&statsPeriod=14d

We can also tighten up the frontend submission, but we want to add this PR so that current submissions with this issue will be retried succesfully by sidekiq.

## Original issue(s)

Added spec based on observed production behavior.

### DIFF
--- a/modules/covid_vaccine/app/services/covid_vaccine/v0/registration_service.rb
+++ b/modules/covid_vaccine/app/services/covid_vaccine/v0/registration_service.rb
@@ -99,10 +99,22 @@ module CovidVaccine
         end
       end
 
-      # if user_type == loa3, then we already had the information from their
-      # authenticated session and added it to the raw form data
+      ## Guard around MPI query
+      # 1. If user_type == loa3, we already have their information from MPI in the
+      # authenticated session
+      # 2. If not all of the required MPI query keys are present, we can't query MPI
+      # 3. If a partial, unparseable date of birth was submitted, we can't query MPI
+      #
       def should_query_mpi?(form_data, user_type)
-        user_type != 'loa3' && (REQUIRED_QUERY_TRAITS & form_data.keys).size == REQUIRED_QUERY_TRAITS.size
+        return false if user_type == 'loa3'
+        return false unless (REQUIRED_QUERY_TRAITS & form_data.keys).size == REQUIRED_QUERY_TRAITS.size
+
+        begin
+          Date.parse(form_data['birth_date'])
+        rescue ArgumentError
+          return false
+        end
+        true
       end
     end
   end
--- a/modules/covid_vaccine/spec/models/covid_vaccine/v0/raw_form_data_spec.rb
+++ b/modules/covid_vaccine/spec/models/covid_vaccine/v0/raw_form_data_spec.rb
@@ -68,7 +68,7 @@ RSpec.describe CovidVaccine::V0::RawFormData, type: :model do
       let(:attributes) { { email: 'jane.doe@email.com', zip_code: '12345-1234',
                            vaccine_interest: 'yes', birth_date: '' } }
 
-      it 'is not valid without the presence of vaccine_interest' do
+      it 'is valid' do
         expect(subject).to be_valid
       end
     end
@@ -79,7 +79,9 @@ RSpec.describe CovidVaccine::V0::RawFormData, type: :model do
 
       it 'is not valid' do
         expect(subject).not_to be_valid
-      end
+        expect(subject.errors.full_messages)
+          .to eq(["Birth date should be in the form yyyy-mm-dd"])
+       end
     end
 
   end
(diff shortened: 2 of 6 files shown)


stephenBarrs (approved):
LGTM!

modules/covid_vaccine/app/services/covid_vaccine/v0/registration_service.rb
```diff
@@ -99,10 +99,22 @@ def attributes_from_mpi(form_data)
...
+      # 3. If a partial, unparseable date of birth was submitted, we can't query MPI
+      #
       def should_query_mpi?(form_data, user_type)
-        user_type != 'loa3' && (REQUIRED_QUERY_TRAITS & form_data.keys).size == REQUIRED_QUERY_TRAITS.size
+        return false if user_type == 'loa3'
+        return false unless (REQUIRED_QUERY_TRAITS & form_data.keys).size == REQUIRED_QUERY_TRAITS.size
+
+        begin
+          Date.parse(form_data['birth_date'])
+        rescue ArgumentError
```
saneshark, line 114: are we likely to encounter anything other than ArgumentError? Wonder if we should just rescue StandardError of any type, probably fine as is.

modules/covid_vaccine/app/services/covid_vaccine/v0/registration_service.rb
```diff
@@ -99,10 +99,22 @@ def attributes_from_mpi(form_data)
...
-      # authenticated session and added it to the raw form data
+      ## Guard around MPI query
+      # 1. If user_type == loa3, we already have their information from MPI in the
+      # authenticated session
+      # 2. If not all of the required MPI query keys are present, we can't query MPI
+      # 3. If a partial, unparseable date of birth was submitted, we can't query MPI
+      #
       def should_query_mpi?(form_data, user_type)
-        user_type != 'loa3' && (REQUIRED_QUERY_TRAITS & form_data.keys).size == REQUIRED_QUERY_TRAITS.size
+        return false if user_type == 'loa3'
```
saneshark, line 109: NITPICK: i suppose you could have just based this on authenticated flag

modules/covid_vaccine/app/services/covid_vaccine/v0/registration_service.rb
```diff
@@ -99,10 +99,22 @@ def attributes_from_mpi(form_data)
...
+        return false unless (REQUIRED_QUERY_TRAITS & form_data.keys).size == REQUIRED_QUERY_TRAITS.size
```
saneshark, line 110: hmm, i don't understand why we need this. Wouldn't it be better do: 

`form_data.select { |h| h.key.in?(REQUIRED_QUERY_TRAITS) && h.value.present? }` 

or something instead?

saneshark (approved):
Mostly nitpick comments, but please look them over before merging.

modules/covid_vaccine/app/services/covid_vaccine/v0/registration_service.rb
saneshark: you're comparing the intersection of of those keys to make sure they're all present so i guess that's a little bit different than what I'm proposing. Nevermind, but I would think a presence check is still needed, because the key could be present but value would be nil.

patrickvinograd: your expression would find all the supplied attributes that were required, but it wouldn't ensure that *all* required elements were present. I'm doing an intersection and comparing size to check that all required keys are present. You're right that I don't check for a non-empty values though...

saneshark: `(REQUIRED_QUERY_TRAITS & form_data.select { |h| h.value.present? }.keys).size == REQUIRED_QUERY_TRAITS.size`

maybe something like ☝️

saneshark: resolving per comment on slack:

> i think so. Only downside is not checking for blank parameter values but all that will do is result in a mpi query with no results

modules/covid_vaccine/spec/models/covid_vaccine/v0/raw_form_data_spec.rb
```diff
@@ -63,5 +63,24 @@
           .to eq(["Vaccine interest can't be blank"])
       end
     end
+
+    context 'without presence of birth_date' do
+      let(:attributes) { { email: 'jane.doe@email.com', zip_code: '12345-1234',
+                           vaccine_interest: 'yes', birth_date: '' } }
+
+      it 'is not valid without the presence of vaccine_interest' do
```
saneshark, line 71: the label on this spec is wrong (copy and paste), but LGTM

saneshark (approved):
one of the specs is labeled wrong, but LGTM

modules/covid_vaccine/spec/models/covid_vaccine/v0/raw_form_data_spec.rb
patrickvinograd: Fixed

PR #5529, opened 2020-12-16
fix non-performant query

annaswims:
## Description of change
our least performant query is 

`SELECT "accounts".* FROM "accounts" WHERE ("accounts"."idme_uuid" = $1 AND "accounts"."idme_uuid" IS NOT NULL OR "accounts"."sec_id" IS NULL AND "accounts"."sec_id" IS NOT NULL)`

{"total"=>44972.5460984851, "avg"=>1267.0309644372}

It's meant to be “match whichever id we have, but make sure we don’t match a nil id”

but `OR "accounts"."sec_id" IS NULL AND "accounts"."sec_id" IS NOT NULL` is bad news when `OR 1=0` would do. 
## Original issue(s)
department-of-veterans-affairs/va.gov-team#0000

PR #5530, opened 2020-12-16
account scopes use none

erluti:
[This PR](https://github.com/department-of-veterans-affairs/vets-api/pull/5529) was rolled out in an emergency and this adds some tests

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -32,13 +32,7 @@ class Account < ApplicationRecord
   redis_ttl REDIS_CONFIG[:user_account_details][:each_ttl]
 
   scope :idme_uuid_match, lambda { |v| where(idme_uuid: v) if v.present? }
-  scope :sec_id_match, lambda { |v|
-                         if v.present?
-                           where(sec_id: v)
-                         else
-                           where('1 = 0')
-                         end
-                       }
+  scope :sec_id_match, lambda { |v| where(sec_id: v) if v.present? }
 
   # Returns the one Account record for the passed in user.
   #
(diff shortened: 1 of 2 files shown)


LindseySaari (approved):
I think this is cleaner, but there could be a specific reason for the where('1 = 0') that i'm not aware of. Approving, but maybe we should wait to merge until we hear back from @annaswims

annaswims (changes requested):
It's possible/likely for a user to have either a nil `idme_uuid` or `sec_id`.  We want to find all Accounts with ONLY matching id's.  We don't want to run the risk of someone calling `idme_uuid_match(nil)`, thinking they're getting only matching accounts when in fact they're getting all accounts with a nil `idme_uuid`

annaswims:
`"accounts"."sec_id" IS NULL AND "accounts"."sec_id" IS NOT NULL` is an expensive table scan for something that will never be true, so we do the cheap version with `1 = 0`

edmangimelli:
Nice catch, Anna! I did not know that about scopes:
> Adds a class method for retrieving and querying objects. The method is intended to return an ActiveRecord::Relation object, which is composable with other scopes. If it returns nil or false, an all scope is returned instead.

https://api.rubyonrails.org/classes/ActiveRecord/Scoping/Named/ClassMethods.html#method-i-scope

edmangimelli:
+1 to adding a comment to the `1 = 0` line. It makes sense now (why we're returning a relation that by definition will always be empty) that I know how I footgun-y returning nil in a _scope's lambda_ is

erluti:
added some tests and used `none`

annaswims (approved):
Ah, yes! That's more clear.

bosawt (approved):
Thanks for the tests! Agree that `none` here makes it more clear what's going on when the id is nil as well.

PR #5536, opened 2020-12-16
17086 mobile address validation fix

kreek:
## Description of change
Fixes a bug in the mobile implementation and works around quirk in how the website address validation works. Bug fix is adding the missing `validation_key` to the meta data. Work around is that although `addressPou` is required in the model it's not really required for validation (in the upstream service), but it is required for submission. Change here is to take it from the request and put it back in the response so the mobile app can just send the entire suggested address along rather than re-add `addressPou` in the client.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#17086


jperk51 (approved):
LGTM

PR #5539, opened 2020-12-17
[7424] Adding icn to the user identity redis cache object to lessen c…

bosawt:
…alls to MPI database

## Description of change
This PR adds the `icn` id as a field on the UserIdentity redis cache, this should help prevent calls to the MPI Lookup when we serialize the `icn` id. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team/issues/7424

- We'll want to make sure this works for SSOe as well as non-SSOe.
- We should be able to use similar testing for dev acceptance as the birls_id and participant_id PR, since this is very similar: https://github.com/department-of-veterans-affairs/vets-api/pull/5410

## Dev Acceptance:

- Set a breakpoint in app/models/user_session_form.rb, line 26, after @user_identity is set as an instance variable for the User model
- Logged in locally with: http://localhost:3000/v1/sessions/idme/new
- Skipped the breakpoint the first time (this is called, then user is upleveled with loa values, then this line is called again)
- Confirmed that user.icn exists
- Confirmed that user.identity.icn exists.
- Confirmed that user.mpi.icn exists and is the same as icn off the identity


lib/mpi/responses/parser_base.rb
```diff
@@ -40,25 +40,12 @@ def sanitize_edipi(edipi)
...
-      def sanitize_birls_id(birls_id)
-        return if birls_id.nil?
-
-        # Get rid of non-digit characters like 'UNK'/'ASKU'
-        sanitized_result = birls_id.match(/\d+/)&.to_s
-        if sanitized_result != birls_id
-          Rails.logger.info "Birls id sanitized, was: '#{birls_id}' now: '#{sanitized_result}'."
-        end
+        sanitized_result = id.match(/\d+/)&.to_s
+        Rails.logger.info "Id sanitized, was: '#{id}' now: '#{id}'." if sanitized_result != id
```
joeniquette, line 48: I know I missed this in the last review, but this shouldn't be a common rails log. Its not a good practice to send sensitive data to generic system logs. We should make this a debug level log and only show this log if we turn on debug mode for this service. I will go research now but Johnny (previous engineer on the team) showed me a while back one of the debug logs that could be turned on through a code change. We certainly would want to do this through a flipper, but we may want at least settle for not logging the real id to rails logger. This applies to all rails logger statements in this PR.

lib/mpi/responses/parser_base.rb
```diff
@@ -40,25 +40,12 @@ def sanitize_edipi(edipi)
...
 
-      def sanitize_participant_id(participant_id)
-        return if participant_id.nil?
+      def sanitize_id(id)
+        return if id.nil?
 
         # Get rid of non-digit characters like 'UNK'/'ASKU'
-        sanitized_result = participant_id.match(/\d+/)&.to_s
-        if sanitized_result != participant_id
-          Rails.logger.info "Participant id sanitized, was: '#{participant_id}' now: '#{sanitized_result}'."
```
joeniquette, line 49: this should be modified as per the statement below about not logging sensitive data.

lib/mpi/responses/parser_base.rb
joeniquette: [Here](https://github.com/department-of-veterans-affairs/va.gov-team/issues/14921#issuecomment-715360767) is the example from Johnny

joeniquette (changes requested):

bosawt:
@joeniquette I removed the logs from the ParserBase class and refactored it a bit. Also, there was no spec file for that class specifically, so I added `parser_base_spec.rb` to completely test the functions in the class. If you look at the `describe #sanitize_id` section of the specs you can see the tests we are now running to confirm the functionality, instead of potentially relying on a log

joeniquette (approved):
Looks good, and a repeatable construct that should be shared.

LindseySaari (approved):

PR #5543, opened 2020-12-17
Mobile API doc refactor

kreek:
## Description of change
The openapi.yaml file is getting very large and causing editors/IDEs to become unresponsive. This PR looks spicier 🌶️ than it is 😄. Main change is extracting each schema from "#/components/schemas/$schema" to its own file: "./schemas/$schema.yml" e.g. 
```
$ref: "#/components/schemas/Appointments"
```
now points to the file in the schemas directory
```
$ref: "./schemas/Appointments.yml"
```
and refs within referenced schemas are relative
```
# Appointments.yml
type: object
required:
  - data
properties:
  data:
    type: array
    items:
      $ref: "./Appointment.yml"
```

## Original issue(s)
N/A

* Doc source YAML files have moved, rendered HTML remains in the same location
* Some multiple example request and responses have also been extracted to the new `./docs/examples` dir


jjulian (approved):
:wow: how did we live without this? Great!

PR #5546, opened 2020-12-17
Api 3866 remove unecessary swss headers

lukemarkwordtlibertyits:
## Description of change

Calls to the okta metadata endpoint and keys endpoint do not require a ssws key. Vets-API calls were using this key. Since it is sensitive information, the key should not be used when unnecessary.

## Original issue(s)
https://vajira.max.gov/browse/API-3866

[API3866_Tests.docx](https://github.com/department-of-veterans-affairs/vets-api/files/5712439/API3866_Tests.docx)

### DIFF
--- a/lib/okta/service.rb
+++ b/lib/okta/service.rb
@@ -26,6 +26,14 @@ module Okta
       end
     end
 
+    def call(action, url)
+      connection.send(action) do |req|
+        req.url url
+        req.headers['Content-Type'] = 'application/json'
+        req.headers['Accept'] = 'application/json'
+      end
+    end
+
     def app(app_id)
       with_monitoring do
         get_url_with_token("#{APP_API_BASE_PATH}/#{app_id}")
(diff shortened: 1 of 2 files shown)


lib/okta/service.rb
```diff
@@ -65,14 +65,16 @@ def delete_grant(user_id, grant_id)
     def metadata(iss)
       proxied_iss = iss.gsub(Settings.oidc.issuer_prefix, Settings.oidc.base_api_url + 'oauth2')
       with_monitoring do
-        get_url_with_token(proxied_iss + '/.well-known/openid-configuration')
+        get(proxied_iss + '/.well-known/openid-configuration', {},
```
beaugrantham-va, line 68: Not a ruby expert.

It looks like the previous behavior returned a `Okta::Response` and this may not?

dbrown-va: Yeah, that response would not get wrapped in the `Okta::Response.new`.

It looks like the `Okta::Response` adds a `success?` helper method vs just the body and status code from a base `get`.  But for metadata and oidc_jwks_keys it isn't used.

We could consider updating `call_with_token` to instead include a boolean arg like `include_token`, and then add one more `define_method` at the bottom?

lukemarkwordtlibertyits: I believe okta responses can wrap any response, so just adding a `Okta:Response:new` to the calls should suffice. I would not have an issue with creating a new method however.

dbrown-va: I think a new method will be preferable here.  Then we aren't subject to unexpected changes to the Common::Client::Base `get` method definition.  Plus we aren't building up the request headers in 2 different ways.

dbrown-va (approved):

beaugrantham-va (approved):

PR #5548, opened 2020-12-18
form inflections

annaswims:
just a starting point for discussion....

Clearly do this better


spec/request/in_progress_forms_request_spec.rb
```diff
@@ -99,6 +99,27 @@
             }.deep_transform_keys { |key| key.to_s.underscore.camelize(:lower) }.to_json)
```
annaswims, line 99: lookee here!  convert as olivebranch does???

app/controllers/v0/in_progress_forms_controller.rb
```diff
@@ -15,13 +15,14 @@ def show
       if form
         render json: form.data_and_metadata
       else
+        # todo
         render json: FormProfile.for(form_id: form_id, user: @current_user).prefill
```
annaswims, line 19: also do the thing that olivebranch does....

PR #5549, opened 2020-12-18
Add sidekiq job adding special issues to an auto-established claim

hinzed1127:
## Description of change
Adds a background job implementation to append new special issues to a particular BGS contention.

## Original issue(s)
https://vajira.max.gov/browse/API-3272

Will retry the default number of times, sidekiq default defined for the project.
May require retrying if the contention is not yet on the claim when this job first tries to run.


modules/claims_api/app/workers/claims_api/special_issue_updater.rb
```diff
@@ -0,0 +1,61 @@
...
+    include SentryLogging
+
+    def perform(user, special_issues, auto_established_claim: nil)
+      service = bgs_service(user).contention
+
+      contentions = service.find_contentions_by_ptcpnt_id(user.participant_id)
+
+      contention = contentions[:benefit_claims].select do |claim|
+                     claim[:clm_id] == auto_established_claim.evss_id.to_s
+                   end [0]
```
hinzed1127, line 18: I still need to handle the cause there are _no_ matches, but is it valid to assume that any matches will be to a _single_ contention?

SlakrHakr: i dont think so? i thought one of our last discussions around all that was that each "disability" in the list of disabilities in the json payload could/would map to a different contention. i was planning on sending something like this in if i took the ticket that modifies the other background job to call this job:
```
[
  { contention_id: ###, special_issues: [{special_issue_stuff_here}] }
]
```

modules/claims_api/app/workers/claims_api/special_issue_updater.rb
```diff
@@ -0,0 +1,62 @@
...
+                     claim[:clm_id] == auto_established_claim&.evss_id.to_s
+                   end [0]
+
+      options = contention_options(contention)
+
+      options[:contentions] =
+        [{
+          clm_id: contention[:clm_id],
+          special_issues: special_issues.map { |special_issue| { spis_tc: special_issue } }
+        }]
```
SlakrHakr, line 26: im not sure this is what we want here, it seems it would blow away all the other contentions currently on the Veteran

modules/claims_api/app/workers/claims_api/special_issue_updater.rb
```diff
@@ -0,0 +1,62 @@
...
+
+      service.manage_contentions(options)
```
SlakrHakr, line 28: theres a lot more data that is required for this method
https://github.com/department-of-veterans-affairs/bgs-ext/blob/ca867c91cf61afe775decd70d0e534c08a564480/spec/bgs/services/contention_spec.rb#L23
i was hoping most/all of it could be grabbed from the contentions call above? but line 22 doesnt contain any of that information. id assume this would raise an exception around required fields?

SlakrHakr (changes requested):

modules/claims_api/app/workers/claims_api/special_issue_updater.rb
hinzed1127: I think that's what's getting added below in `contention_options`? Linting was complaining about method length and line 25 was the only thing that wasn't a straight 1:1 mapping, so I pulled all the other properties into that

modules/claims_api/app/workers/claims_api/special_issue_updater.rb
hinzed1127: you're right, it would. I don't think that would be an issue for updating special issues on a new claim, but it would for an _existing_ contention that had special issues. given some edge cases I outlined [here](https://vajira.max.gov/browse/API-3272?focusedCommentId=770924&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-770924), it might be more sense to have a separate method to update special issues for existing contentions.

modules/claims_api/app/workers/claims_api/special_issue_updater.rb
hinzed1127: ahh ok, I assumed that a call could have multiple special issues, for for a single contention. I updated this logic now to handle multiple matches if that were the case

modules/claims_api/app/workers/claims_api/special_issue_updater.rb
SlakrHakr: youre absolute correct, i somehow completely missed that method below :(

modules/claims_api/app/workers/claims_api/special_issue_updater.rb
```diff
@@ -0,0 +1,98 @@
...
+
+      BGS::Services.new(
+        external_uid: user.icn || user.uuid,
+        external_key: external_key
+      )
+    end
+
+    def append_provided_special_issues(options, special_issues)
+      options[:contentions].each do |contention|
+        next unless contention[:cntntn_id] == contention_id
```
SlakrHakr, line 39: This is a bit confusing to me. I know there is some sort of matcher between the `disability` the api consumer sends in and the contention associated with the claim, but i'm concerned `cntntn_id` on the existing contention is incorrect.

Gemfile
```diff
@@ -43,7 +43,7 @@ gem 'attr_encrypted', '3.1.0'
 gem 'aws-sdk-s3', '~> 1'
 gem 'aws-sdk-sns', '~> 1'
 gem 'betamocks', git: 'https://github.com/department-of-veterans-affairs/betamocks', branch: 'master'
-gem 'bgs_ext', git: 'https://github.com/department-of-veterans-affairs/bgs-ext.git', require: 'bgs'
+gem 'bgs_ext', git: 'https://github.com/department-of-veterans-affairs/bgs-ext.git', require: 'bgs', branch: 'slakrhakr/API-3272'
```
SlakrHakr, line 46: Requires this pr for bgs-ext before this can go through
https://github.com/department-of-veterans-affairs/bgs-ext/pull/51

modules/claims_api/app/workers/claims_api/special_issue_updater.rb
SlakrHakr: ok, through various discussions, this has changed to a hash of values to match on:
```
{ claim_id:, code:, name: }
```
`claim_id` is required and at least one of either `code` or `name` is required.

modules/claims_api/app/workers/claims_api/special_issue_updater.rb
```diff
@@ -0,0 +1,141 @@
...
+  class SpecialIssueUpdater
+    include Sidekiq::Worker
+    include SentryLogging
+
+    # Update special issues for a single contention/disability
+    #
+    # @param user [OpenStruct] Veteran to attach special issues to
+    # @param contention_id [Hash(claim_id:, code:, name:)] Identifier to match existing contention
+    # @param special_issues [Array(String)] List of special issues to append
+    def perform(user, contention_id, special_issues)
```
SlakrHakr, line 16: concept here is that this logic will be called for each disability provided in the json payload, so:
```
payload[:disabilities].each do |disability|
   ClaimsApi::SpecialIssueUpdater.perform_later("special_issues from both disability and secondary disabilities for that disability")
end
```

SlakrHakr (approved):

Gemfile
SlakrHakr: done

modules/claims_api/app/workers/claims_api/special_issue_updater.rb
```diff
@@ -0,0 +1,146 @@
...
+        ptcpnt_id_vet: claim[:ptcpnt_id_vet],
+        ptcpnt_suspns_id: claim[:ptcpnt_suspns_id],
+        soj_lctn_id: claim[:soj_lctn_id],
+        contentions: existing_contentions(claim, contention_id, special_issues)
+      }
+    end
+
+    # Generate expected contentions payload.
+    #
+    # @param claim [Hash] BGS claim object
```
bastosmichael, line 91: I'm okay with this style of annotation as long as we are consistent in implementing it in all code moving forward, probably something for us to bring up in our team agreement meeting next week.

bastosmichael (approved):
Good work here, on the code comments and annotations, I want to make sure we bring up that style in order to maintain consistency across code implementation on claims module.

PR #5553, opened 2020-12-21
Remove/disable facility locator jobs

LindseySaari:
## Description of change
V0 facilities is no longer being maintained in favor of moving to an external app that we call via the api. The FacilityLocationDownloadJob is polluting the oncall channel and the error we’re seeing is just basically old sidekiq jobs that are still running to update facility data. The benefits and memorials team form 526 still relies on this endoint, but the job to sync the data should be depreciated.

It was [requested](https://dsva.slack.com/archives/C30LCU8S3/p1608553618219900) to silence/remove the alert from oncall and the work to disable the facility sync jobs was already in the backlog.

## Original issue(s)
**See this ticket for more background information:**
department-of-veterans-affairs/va.gov-team#16522


erluti (approved):
👍

PR #5554, opened 2020-12-21
[CH31 Form] Updates from e2e testing

digitaldrk:
## Description of change
During testing with our integration partner, we've found small data changes that they require to get all data in from the CH31 form submitted from the Frontend. These small changes were found during end to end testing where we would submit forms through our dev env and they would verify that they did or did not see that data.

The result of the above is this PR. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#17571

    * No additions
    * This feature is FF'd and it is not available in Prod to anyone
    * No Sentry logging added in this PR
    * Negative
    * Nope
    * No concerns
-->


app/models/saved_claim/veteran_readiness_employment_claim.rb
```diff
@@ -46,7 +46,7 @@ def regional_office
   def prepare_form_data
     form_copy = parsed_form
     appointment_time_preferences = form_copy['appointmentTimePreferences'].map do |key_value|
-      key_value[0] if key_value[1] == true
+      key_value[0].downcase if key_value[1] == true
```
erluti, line 49: I don't see a test to show that these values should be downcased

lib/vre/ch31_form.rb
```diff
@@ -63,49 +63,71 @@ def format_payload_for_vre
...
+        isForeign: client_hash['country'] != 'USA',
+        isMilitary: client_hash['isMilitary'] || false,
+        countryName: client_hash['country'],
+        addressLine1: client_hash['street'],
+        addressLine2: client_hash['street2'],
+        addressLine3: client_hash['street3'],
+        city: client_hash['city'],
+        stateCode: client_hash['state'],
+        zipCode: client_hash['postalCode']
+      }.with_indifferent_access
```
erluti, line 130: what does indiffernt access get you?  looks like you're using symbol keys everywhere

lib/vre/ch31_form.rb
```diff
@@ -63,49 +63,71 @@ def format_payload_for_vre
...
-          "province": new_address['state'],
-          "internationalPostalCode": new_address['postalCode']
-        }
+      new_address = mapped_address_hash(claim_form_hash['newAddress'])
+
+      adjusted_new_address = {
+        newAddress: new_address
       }
+
+      return adjusted_new_address if adjusted_new_address.dig(:newAddress, :isForeign) == false
```
erluti, line 109:
```suggestion
      return adjusted_new_address unless new_address[:isForeign]
```

lib/vre/ch31_form.rb
```diff
@@ -63,49 +63,71 @@ def format_payload_for_vre
...
+
+      new_vet_international_address = adjusted_new_address[:newAddress]
```
erluti, line 111: I would just use `newAddress` here instead of declaring a new variable.  If you're worried about the context, this would be a good place for a comment like `# change zipCode and stateCode to the international names`

spec/controllers/v0/veteran_readiness_employment_claims_controller_spec.rb
```diff
@@ -30,7 +30,7 @@
     context 'visitor' do
       it 'validates successfully' do
         VCR.use_cassette 'veteran_readiness_employment/send_to_vre' do
-          form_params = { veteran_readiness_employment_claim: { form: test_form_no_vet_info.form } }
+          form_params = { veteran_readiness_employment_claim: { form: test_form.form } }
```
erluti, line 33: why did this test change?

spec/lib/vre/ch31_form_spec.rb
```diff
@@ -23,6 +23,24 @@
...
+  let(:foreign_vet_address_claim) do
+    claim = create(:veteran_readiness_employment_claim)
+    form_copy = claim.parsed_form
+    form_copy['veteranAddress']['country'] = 'DEU'
+    claim.form = form_copy.to_json
+
+    claim
+  end
+
+  let(:foreign_new_address_claim) do
```
erluti, line 35: These two lets should be moved into the context where they are used

erluti (commented):
got a few questions and suggestions

lib/vre/ch31_form.rb
digitaldrk: Nice👌

lib/vre/ch31_form.rb
digitaldrk: Good callout. Just being overly cautious

spec/controllers/v0/veteran_readiness_employment_claims_controller_spec.rb
digitaldrk: Good question, after we ran the e2e tests and added more code to this PR, this test threw the good ol' undefined method for nil error. 

Turns out that we were allowing `nil` values through for the`veteranInformation` hash attribute in the form. The form we were passing here did not have `veteranInformation`.

Your question though made me think we should add validation for that so I added it in the model too.

erluti (commented):
Looking good.  Although the new validation broke some tests

spec/lib/vre/ch31_form_spec.rb
```diff
@@ -69,5 +69,87 @@
         end
       end
     end
+
+    context 'user\'s current (veteran) address is foreign' do
```
ericboehs, line 73:
```suggestion
    context "user's current (veteran) address is foreign" do
```

spec/lib/vre/ch31_form_spec.rb
```diff
@@ -69,5 +69,87 @@
...
+
+        expect(foreign_vet_address_claim_service).to receive(:send_to_vre).with(
+          payload: a_string_including('"province":"FL"')
+        ) { response_double }
+
+        foreign_vet_address_claim_service.submit
+      end
+    end
+
+    context 'user\'s new address is foreign' do
```
ericboehs, line 114:
```suggestion
    context "user's new address is foreign" do
```

ericboehs (approved):

digitaldrk:
@ericboehs and @erluti Thank you for the reviews!

PR #5555, opened 2020-12-21
STEM: Notify applicant when contacting SCO #17363

zurbergram:
## Description of change

Send an email to applicant when SCOs are notified because CCing is not available.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#17363


db/schema.rb
```diff
@@ -322,6 +322,15 @@
     t.index ["job_id"], name: "index_form526_job_statuses_on_job_id", unique: true
   end
 
+  create_table "form526_opt_ins", id: :serial, force: :cascade do |t|
```
zurbergram, line 325: This change appears every time a `make rebuild` is done

lib/config_helper.rb
```diff
@@ -7,6 +7,7 @@ module ConfigHelper
 
   def setup_action_mailer(config)
     config.action_mailer.preview_path = Rails.root.join('spec', 'mailers', 'previews')
+    config.action_mailer.show_previews = Rails.env.development? || FeatureFlipper.staging_email?
```
zurbergram, line 10: this should make both http://localhost:3000/rails/mailers and https://staging-api.va.gov/rails/mailers/ available to use.  This should not enable this in production

shawkey-daniel (approved):
LGTM

lib/config_helper.rb
LindseySaari: I would propose this and related changes for the mailer preview functionality in staging be broken out into a separate PR please

zurbergram: https://github.com/department-of-veterans-affairs/vets-api/pull/5557

LindseySaari:
Looks like some associated specs are failing

db/schema.rb
LindseySaari: It appears that your branch doesn't have this change: https://github.com/department-of-veterans-affairs/vets-api/commit/42b7769ffba861692eb5bdabfceb0b148c2bb52c#diff-cbf8b25d6d853acf58fa9057841f407d29ffe90649c75cf9e3f51b2d6e3aa1d3

zurbergram: i haven't done a `make migrate` in a while which is why was continuing to show up when doing a `make rebuild`

zurbergram: I removed this change but even after a `make migrate` > `make rebuild` the change still shows up

zurbergram:
> Looks like some associated specs are failing

@LindseySaari I fixed related specs that were failing but other specs are failing

delli-gatti-michael (approved):

LindseySaari (approved):

PR #5564, opened 2020-12-22
BGS 686 Set Child Married Indicator

kathleencrawford:
## Description of change

During testing of the 686 (Dependency Claims), we discovered that when reporting the marriage of a child under 18, the child previously married indicator must be set to Y and the relationship type must be set to Biological.  This PR is a fix to update those flags accordingly.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#17810
department-of-veterans-affairs/va.gov-team#12425

No additional settings or configuration have been added.  The dependency application process is already in place, this PR is just updating a couple flags for a specific workflow.

## Testing
- [x] Local testing
- [x] Specs updated
- [x] Staging testing planned

### DIFF
--- a/app/models/bgs_dependents/child_marriage.rb
+++ b/app/models/bgs_dependents/child_marriage.rb
@@ -10,7 +10,8 @@ module BGSDependents
       {
         'event_date': @child_marriage['date_married'],
         'ssn': @child_marriage['ssn'],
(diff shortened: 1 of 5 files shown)


digitaldrk (approved):
:octocat: 👌

kfrz (approved):

PR #5565, opened 2020-12-22
Update lib/hca/enrollment_system.rb to support multiple addresses pro…

kevinmirc:
Fixes: https://github.com/department-of-veterans-affairs/va.gov-team/issues/17809
Schema: https://github.com/department-of-veterans-affairs/vets-json-schema/pull/557

**Description**
Update the 10-10EZ (HCA) to accept multiple addresses provided for the veteran. This feature will be toggled in the front-end, so these changes should still support the previous version where only one address is provided.

**Summary**
- [x] Accept a mailing (`P` - permanent) and home (`R` - residential) address for veteran.
- [x] When only one address provided, set as `P` (permanent).
- [ ] Link to latest Schema changes (💡 will link in following PR, not needed for data to flow through)
  - [ ] test src code changes before adding new schema into api

**Testing (e2e)**
- [ ] Send with one address
  - [ ] Ensure data is received in ES (enrollment system) as it does today in prod
- [ ] Send with two address
  - [ ] Ensure data flows through successfully
  - [ ] Get confirmation from ES team that data is received as expected (format, type, etc.)


lib/hca/enrollment_system.rb
```diff
@@ -85,7 +85,7 @@ def financial_flag?(veteran)
       veteran['understandsFinancialDisclosure'] || veteran['discloseFinancialInformation']
     end
 
-    def format_address(address)
+    def format_address(address, type_code = nil)
```
kfrz, line 88: Prefer keyword arguments to reduce ambiguity in method parameter orders, e.g.,

`def format_address(address, type_code: nil)`

kfrz (approved):

PR #61, opened 2016-09-01
Convert case middleware

lihanli:
closes #60

PR #5566, opened 2020-12-22
S-I-P Bug - Approach 1: Manually "olivebranch" the Metadata Field for Save-In-Progress Data

edmangimelli:
The FE is going to switch to only sometimes sending the inflection header for save-in-progress routes --that is, the FE will sometimes opt-out of [OliveBranch](https://github.com/vigetlabs/olive_branch) when using save-in-progress routes.

## Why?
form_data sometimes has fancy keys that do not survive olivebranch's
`camelCase` ➡️ `snake_case` ➡️ back to `camelCase`
(or `snake_case` ➡️ `camelCase` ➡️ back to `snake_case`)

example:
```
irb(main):001:0> "Hello there, Sam-I-Am!".underscore.camelize(:lower).underscore
=> "hello there, sam_i_am!"
```

(Sometimes sentences with spaces and punctuation end up being keys in the json (checkboxes).)

Because the BE does not operate on form_data --it acts as a passive store-- it's (hopefully) safe to cut out olivebranch for SIP.

From the FE, it won't look like a change:

we're going from:
```
FE sends camelCase
     \
      \ tranformation!
       \
        `--> Database stores snake_case
            /
           / transformation!
          /
camel <--'
```
to:
```
FE sends camelCase
     \
      \ no tranformation
       \
        `--> Database stores camelCase
            /
           / no transformation
          /
camel <--'
```
## The Catch!

Both the FE and BE operate on the metadata field, and both are expecting it to be in their own case.

Therefore, even when the inflection header is not sent, the metadata field must be transformed as if [OliveBranch](https://github.com/department-of-veterans-affairs/vets-api/pull/61) operated on it.

### DIFF
--- a/app/controllers/v0/in_progress_forms_controller.rb
+++ b/app/controllers/v0/in_progress_forms_controller.rb
@@ -13,15 +13,15 @@ module V0
       form    = InProgressForm.form_for_user(form_id, @current_user)
 
       if form
-        render json: form.data_and_metadata
+        render json: form.data_and_camelized_metadata
       else
-        render json: FormProfile.for(form_id: form_id, user: @current_user).prefill
+        render json: FormProfile.for(form_id: form_id, user: @current_user).camelized_prefill
       end
     end
 
     def update
       form = InProgressForm.where(form_id: params[:id], user_uuid: @current_user.uuid).first_or_initialize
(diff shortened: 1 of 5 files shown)


edmangimelli:
Being replaced by:
https://github.com/department-of-veterans-affairs/vets-api/pull/5582

PR #5571, opened 2020-12-29
API-3265: [database-schema] Special Issues - Add special issues persistence for established claim

SlakrHakr:
## Description of change
Adds a "special_issues" column to the existing ClaimsApi::AutoEstablishedClaim database table. Will be used to store special issues for a claim long-term.

## Original issue(s)
https://vajira.max.gov/browse/API-3265

Migrated locally and implemented functionality around storing data in this new column. Have a spec to test the storing of that data in the next pr for this.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -47,13 +47,13 @@ ActiveRecord::Schema.define(version: 2020_12_17_223026) do
   end
 
   create_table "appeals_api_notice_of_disagreements", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
-    t.string "status", default: "pending", null: false
     t.string "encrypted_form_data"
     t.string "encrypted_form_data_iv"
     t.string "encrypted_auth_headers"
     t.string "encrypted_auth_headers_iv"
     t.datetime "created_at", precision: 6, null: false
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -47,13 +47,13 @@
...
 
   create_table "appeals_api_notice_of_disagreements", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
-    t.string "status", default: "pending", null: false
     t.string "encrypted_form_data"
     t.string "encrypted_form_data_iv"
     t.string "encrypted_auth_headers"
     t.string "encrypted_auth_headers_iv"
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
+    t.string "status", default: "pending", null: false
```
bastosmichael, line 56: Was this from a different PR?

bastosmichael (approved):
Let's make sure that the status thing was actually us otherwise it looks good.

db/schema.rb
SlakrHakr: it just moved it from line 50 to the end of this create table statement for whatever reason

PR #5573, opened 2020-12-29
15330 removing ppms v1 clients and round trip feature flags

darkbushido:
## Description of change
We are happy with how the new feature flags are working and the significant performance increase in response times.
removing the ppms v1 client flag and the skip round trip flags and associated code

## Original issue(s)
department-of-veterans-affairs/va.gov-team#15330

### Removing the Following feature flags
* `facility_locator_ppms_use_v1_client`
* `facility_locator_ppms_skip_additional_round_trips`


kfrz (approved):
Love the ratio on this PR!

PR #5575, opened 2020-12-29
Refactor deprecated code

ElleeB:
## Description of change
The appeals_api test suite was reporting deprecation errors when run.

- update `where.not` when querying multiple attributes 

## Original issue(s)
https://vajira.max.gov/browse/API-4168

### DIFF
--- a/modules/appeals_api/app/models/appeals_api/higher_level_review.rb
+++ b/modules/appeals_api/app/models/appeals_api/higher_level_review.rb
@@ -92,7 +92,7 @@ module AppealsApi
 
     scope :received_or_processing, -> { where status: RECEIVED_OR_PROCESSING }
     scope :completed, -> { where status: COMPLETE_STATUSES }
-    scope :has_pii, -> { where.not encrypted_form_data: nil, encrypted_auth_headers: nil }
+    scope :has_pii, -> { where.not(encrypted_form_data: nil).or(where.not(encrypted_auth_headers: nil)) }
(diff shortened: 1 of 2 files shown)


yakschuss:
What are the errors that this removes?

ElleeB:
> What are the errors that this removes?

I believe they are in the ticket

modules/appeals_api/app/models/appeals_api/higher_level_review.rb
```diff
@@ -92,7 +92,7 @@ def log_bad_central_mail_response(resp)
 
     scope :received_or_processing, -> { where status: RECEIVED_OR_PROCESSING }
     scope :completed, -> { where status: COMPLETE_STATUSES }
-    scope :has_pii, -> { where.not encrypted_form_data: nil, encrypted_auth_headers: nil }
+    scope :has_pii, -> { where(" encrypted_form_data != 'nil' OR encrypted_auth_headers != 'nil'") }
```
cilestin, line 95: Double check that the old code and the new are creating the same SQL statement. Remember that an empty value in ruby is `nil` but in SQL it's `null`.

cilestin (changes requested):

modules/appeals_api/app/models/appeals_api/higher_level_review.rb
ElleeB: Ohhhh! Yes, I see! Multiple ways to solve. I'll push a new one up in a sec

cilestin (approved):

PR #5576, opened 2020-12-29
API-3267: Special Issues - Update the claim establisher job to check for special issues

SlakrHakr:
## Description of change
Kicks off special issue updater job if special issues are found on the claim submission.
Also ensures the equality checks within the special issue updater job doesn't have issues with some items being strings and others being integers.

## Original issue(s)
https://vajira.max.gov/browse/API-3267

Updated a fixture to ensure this new code section gets executed within existing specs.


bastosmichael (approved):
LGTM like the refactor for sure.

PR #5577, opened 2020-12-29
API-3274: Special Issues - Set up logging to capture what we need

SlakrHakr:
## Description of change
Writes special issues to log when an ClaimsApi::AutoEstablishedClaim record is created, for tracking purposes.

## Original issue(s)
https://vajira.max.gov/browse/API-3274

spec


bastosmichael (approved):
LGTM

PR #5578, opened 2020-12-29
API-4221 Remove PDF Checksumming from specs

cilestin:
## Description of change
The specs were using finicky MD5 checksumming to verify the generated pdfs were as expected against their gold master. This change continues to use the gold master files, but instead reads the pdf for much more reliable text data.

## Original issue(s)
https://vajira.max.gov/browse/API-4221


yakschuss (approved):
LGTM! I like the matcher.

PR #5579, opened 2020-12-30
Add char limit to BGS addresses

digitaldrk:
## Description of change
During testing with our integration partner BGS, we learned that they will not accept more than 20 characters per line for addresses. This PR combines all lines in an address and then splits them into 3 lines w/o going beyond 20 characters and w/o splitting up words. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#15775

    * No settings added
    * This form is not available to the general public currently. No FF was added in this PR though
    * No sentry logging added
    * No prometheus metrics added
    * None updated
    * No concerns


kfrz (approved):
:+1:

PR #5582, opened 2020-12-30
Save-In-Progress Bug Fix: Use CamelCase in the Back End

edmangimelli:
The Front End is no longer going to _always_ send [the inflection header](https://github.com/vigetlabs/olive_branch#use) for save-in-progress routes (for now, [only Form526 will stop sending the inflection header](https://github.com/department-of-veterans-affairs/vets-website/pull/15517)).

## Why?
OliveBranch is a middleware that ensures that the Front End only receives camelCase, and the Back End only receives snake_case. That requires switching the case of keys on both the request and the response. For example, let's say some form data originates on the Front End.

1. The Front End sends JSON with camelCase keys
2. Before it hits a back-end controller, it's transformed to snake_case
3. The Back End stores it in the DB (as snake_case)

Later, the Front End retrieves that form data:

4. The Back End sends snake_case
5. _After the controller renders_, the OliveBranch middleware transforms it to camelCase before it is finally sent

This is fine for most keys:
```
> "addressOnFile".underscore.camelize(:lower)
=> "addressOnFile"
```
 but some keys in Form526 (keys for checkboxes, as an example) are corrupted during this process:
```
> "Hello there, Sam-I -Am!".underscore.camelize(:lower)
=> "hello there, samI Am!"
```
To make matters worse, repeating the process might produce new wrong results:
```
> "Hello there, Sam-I -Am!".underscore.camelize(:lower).underscore.camelize(:lower)
=> "hello there, samI am!"  # no capital A in this one
```

## The Change
Currently, the save-in-progress flow that happens every time:
```
  FE sends camelCase and inflection header
       \
        \ Rails server receives camelCase
         \ transformation! (OliveBranch middleware)
          \ Controller receives snake_case
           \
            `--> Database stores snake_case
                              /
                             / FE retrieves data
                            / transformation! (OliveBranch middleware)
                           /
  FE receives camelCase <-'
```
The new alternative flow that **never** deep transforms keys (neither with OliveBranch nor a serializer):
```
  FE sends camelCase WITHOUT the inflection header
       \
        \ Rails server receives camelCase
         \ 
          \ Controller receives camelCase
           \
            `--> Database stores camelCase
                              /
                             / FE retrieves data
                           /
  FE receives camelCase <-'
```

## How to Stop Using OliveBranch for a Specific Form

To switch a form to using the new flow (new flow == the Back End **never** deep transforming keys (neither with OliveBranch nor a serializer)),

1. Fix (camelize) the `InProgressForms` for that specific form type in the DB (a rake task that camelizes nested keys) ([example](https://github.com/department-of-veterans-affairs/vets-api/pull/5601) --note: this PR is doing more than just camelizing).
2. Stop sending the inflection header ([example](https://github.com/department-of-veterans-affairs/vets-website/pull/15517)).

#### A Note on Step 1

Step 1 is for Veterans who started the form _before_ you deploy step 2, and will resume _after_ you deploy step 2. That is, someone who started the form with the inflection header, didn't complete the form (snake_case save-in-progress data in the DB), and resumes the form after you've switched to not using the inflection header. If the data in the DB isn't fixed first, the Back End will send snake_case keys to the Front End (which the Front End won't know how to use). **Practically, you might have to run your rake task right before the deploy of step 2, _and_ also right after --just to be certain you've caught any stragglers.** [The rake task we've cooked up for Form526](https://github.com/department-of-veterans-affairs/vets-api/pull/5601) camelizes keys but also fixes keys that were corrupted by all of the automatic key transformation (see the "Why?" section above).

## Implementation Summary
1. The controller now expects top-level keys to be in either case
2. the Back End expects _nested keys_ to be camelCase _when operating on InProgressForm data_ (which isn't often --the Back End mostly acts as a passive store for the Front End) (the Front End "owns" the data)
3. the controller exclusively sends out camelCase top-level keys and unaltered nested keys (though, of course, OliveBranch will still deep transform keys if the inflection header is being used)

The end result is that, if the inflection header is sent, keys get deep transformed to camelCase, if the inflection header isn't sent, camelCase _top-level keys_ are returned _while nested keys are untransformed_ (that is, the Back End sends exactly what's in the DB).

## I found this to be a tricky problem --your eyeballs are greatly appreciated. 🙌


app/controllers/v0/in_progress_forms_controller.rb
```diff
@@ -5,7 +5,9 @@ class InProgressFormsController < ApplicationController
     include IgnoreNotFound
 
     def index
-      render json: InProgressForm.where(user_uuid: @current_user.uuid)
+      render(
+        json: InProgressFormSerializer.new(InProgressForm.where(user_uuid: @current_user.uuid)).serializable_hash
+      )
```
edmangimelli, line 10: switching to fast_jsonapi (which vets-api already has) made it a little easier to prevent a deep transformation of the keys

app/models/in_progress_form.rb
```diff
@@ -9,11 +9,11 @@ def cast(value)
     alias serialize cast
   end
 
-  RETURN_URL_SQL = "CAST(metadata -> 'return_url' AS text)"
-  scope :has_attempted_submit, -> { where("(metadata -> 'submission' ->> 'has_attempted_submit')::boolean") }
+  RETURN_URL_SQL = "CAST(metadata -> 'returnUrl' AS text)"
```
annaswims, line 12: these scopes are only used in a rarely used rake task so I'm fine with not making them work with both cases.  They'll just work with new data.

annaswims:
I like everything you've got so far.  Just one thing... I don't think this provides us a path to transition for users who start a form today.  

If a user starts a form526 today their data will be snake case in the database. At Some later date we merge this and stop sending the header from the FE.  Then the FE will receive snake case form data and break.  Can we add an extra attribute to the metadata to let us know if it needs to be converted?

app/models/in_progress_form.rb
edmangimelli: oh whoa! big oversight on my part. are you sure? we could something fancy:
```
where (return_url: x, updated_at: before_pr) OR where (returnUrl: x, updated_at: after_pr)
```
?

annaswims: I honestly don't think it's worth the effort.  Most of the time people are looking for recent issues with these scopes.

app/models/in_progress_form.rb
```diff
@@ -31,7 +31,7 @@ def self.form_for_user(form_id, user)
 
   def data_and_metadata
     {
-      form_data: JSON.parse(form_data),
```
annaswims, line 34: https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/workers/delete_attachment_job.rb#L22 needs to be changed

spec/factories/in_progress_forms/in_progress_forms.rb
```diff
@@ -7,7 +7,7 @@
     metadata do
       {
         version: 1,
-        return_url: 'foo.com'
```
annaswims, line 10: Should this be changed?  most forms will still be sending the header, so more often this will be snake case

spec/request/in_progress_forms_request_spec.rb
```diff
@@ -55,7 +55,7 @@
         subject
         items = JSON.parse(response.body)['data']
         expect(items.size).to eq(2)
-        expect(items.dig(0, 'attributes', 'form_id')).to be_a(String)
```
annaswims, line 58: what do you think about adding a few more specs here that use the olive branch header?

edmangimelli: added (see also #5630)

spec/factories/in_progress_forms/in_progress_forms.rb
edmangimelli: yeah, i was conflicted for that reason. i'm fine with swapping it back 👍

edmangimelli:
> I like everything you've got so far. Just one thing... I don't think this provides us a path to transition for users who start a form today.
> 
> If a user starts a form526 today their data will be snake case in the database. At Some later date we merge this and stop sending the header from the FE. Then the FE will receive snake case form data and break. Can we add an extra attribute to the metadata to let us know if it needs to be converted?

Anna and I have since hashed this problem out, and are going to use a [rake task](https://github.com/department-of-veterans-affairs/vets-api/pull/5601) to fix the data before (and after) deploy.

erluti:
Does this swagger doc need an update since you can't get snake_keys back anymore? https://github.com/department-of-veterans-affairs/vets-api/blob/b6f3da99b3107c718dc4eba503b866aa3dd61a08/app/swagger/swagger/requests/in_progress_forms.rb#L103

edmangimelli:
> Does this swagger doc need an update since you can't get snake_keys back anymore?
> 
> https://github.com/department-of-veterans-affairs/vets-api/blob/b6f3da99b3107c718dc4eba503b866aa3dd61a08/app/swagger/swagger/requests/in_progress_forms.rb#L103

You can technically still send snake_case by using the inflection header.

Looks like this documentation never listed the inflection header --I'll update that. Also, the controller now expects either camelCase _or_ snake_case parameters --I'll make that update too. Thanks!

annaswims:
I merged #5651 so you're good to merge too.  I think you'll also need to change form_data here https://github.com/department-of-veterans-affairs/vets-api/pull/5651/files#diff-44c762898b8060f6797122264433efd34b508fab93767acf56ffdafc4baf90b5R71

spec/factories/in_progress_forms/hca_forms.rb
```diff
@@ -8,7 +8,7 @@
       {
         savedAt: 1_595_954_803_670,
         version: 6,
-        return_url: '/form-url/review-and-submit'
+        returnUrl: '/form-url/review-and-submit'
```
annaswims, line 11: I think that this should remain unchanged.  HCA will still use inflection headers for now, so it would be `return_url` in the database.

annaswims (approved):
LGTM

edmangimelli:
documentation updates https://github.com/department-of-veterans-affairs/vets-api/pull/5662 (wip)

ericboehs (approved):

PR #5584, opened 2020-12-30
API-4222/3725 character limits and copy changes

yakschuss:
## Description of change
Copy updates to NOD swagger documentation

## Original issue(s)
[4222](https://vajira.max.gov/browse/API-4222)
[3725](https://vajira.max.gov/browse/API-3725)

### DIFF
--- a/modules/appeals_api/config/schemas/10182.json
+++ b/modules/appeals_api/config/schemas/10182.json
@@ -46,7 +46,7 @@
       "additionalProperties": false,
       "properties": {
         "homeless":            { "type": "boolean" },
-        "representativesName": { "type": "string" }
+        "representativesName": { "type": "string", "maxLength": "120" }
       },
       "required": [ "homeless" ]
     },
@@ -75,8 +75,8 @@
       "type": "object",
       "additionalProperties": false,
       "properties": {
-        "issue":                     { "$ref": "#/definitions/nodCreateNonBlankString" },
-        "decisionDate":              { "$ref": "#/definitions/nodCreateDate" },
+        "issue":                     { "$ref": "#/definitions/nodCreateNonBlankString", "maxLength": "368" },
+        "decisionDate":              { "$ref": "#/definitions/nodCreateDate", "maxLength": "10" },
         "decisionIssueId":           { "type": "integer" },
         "ratingIssueReferenceId":    { "type": "string" },
         "ratingDecisionReferenceId": { "type": "string" }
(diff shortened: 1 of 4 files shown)


modules/appeals_api/config/schemas/10182.json
```diff
@@ -75,8 +75,8 @@
       "type": "object",
       "additionalProperties": false,
       "properties": {
-        "issue":                     { "$ref": "#/definitions/nodCreateNonBlankString" },
-        "decisionDate":              { "$ref": "#/definitions/nodCreateDate" },
+        "issue":                     { "$ref": "#/definitions/nodCreateNonBlankString", "maxLength": 368 },
+        "decisionDate":              { "$ref": "#/definitions/nodCreateDate", "maxLength": 10 },
```
cilestin, line 79: I feel like the `maxLength` should be moved down to the `nodCreateDate` definition, since its pattern is looking for an exact number of characters anyway.

Since the pattern expects an exact number of characters, let's enforce `minLength` too.

modules/appeals_api/config/schemas/10182_headers.json
```diff
@@ -29,7 +29,7 @@
     "nodCreateHeadersSsn": {
       "allOf": [
         { "description": "social security number" },
-        { "type": "string", "pattern": "^[0-9]{9}$" }
+        { "type": "string", "maxLength": 9, "pattern": "^[0-9]{9}$" }
```
cilestin, line 32: Since the pattern expects an exact number of characters, let's enforce `minLength` too.

modules/appeals_api/config/schemas/10182_headers.json
```diff
@@ -64,6 +66,7 @@
     "nodCreateHeadersBirthDate": {
       "allOf": [
         { "description": "birth date" },
+        { "maxLength": 10 },
```
cilestin, line 69: Same here. I feel like the date definition itself should have the `minLength` & `maxLength` associated with it since it knows the number of characters it requires for a valid date.

cilestin (changes requested):
A few quick things but otherwise lookin good!

modules/appeals_api/config/schemas/10182.json
yakschuss: I'm not sure what you mean by "moved down to the nodCreateDate definition" - isn't this a specific length requirement on the `issues` key?

modules/appeals_api/config/schemas/10182_headers.json
yakschuss: Wouldn't the minLength be the same as maxLength in the case of date?

cilestin (approved):
Looks good to me!

PR #5585, opened 2020-12-30
Fix DetailedSchemaErrors

cilestin:
## Description of change
Fixes errors of DetailedSchemaErrors where translations were not receiving the proper interpolated values to display as expected.

Before: `:detail => "'%{value}' did not fit within the defined length limits"`
After: `:detail => "'Ed' did not fit within the defined length limits"`


yakschuss (approved):
Apologies, thought I had approved this earlier! Looks good to me!

PR #5589, opened 2020-12-31
API-3270: Special Issues - Portal documentation for any new / changed features from consumer standpoint

SlakrHakr:
## Description of change
Add some basic documentation regarding special issues.
Also adds example special issues to the example curl request within our documentation.

## Original issue(s)
https://vajira.max.gov/browse/API-3270

Viewed in developer portal locally.


bastosmichael (approved):
My guess is we probably want Mark and Darcy's review on here as well but from a technical point of view it looks good to me.

PR #5590, opened 2020-12-31
Switch to LOA3 user check

digitaldrk:
## Description of change
During testing for the CH36 form, we realized that some logged-in users could be missing basic information based on their LOA status. We were simply checking if the user was logged in and the consequence was that we were updating the user's form with nil values and thus failing vets JSON schema validation.

Rather than simply checking that a user is logged in, this PR checks the users LOA is == 3 and if so, adds the in-memory information in `current_user`. If the LOA != 3, we assume the FE is sending necessary data to us.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#17970

    *  No changes
    *  No feature flag added to this PR
    *  No sentry logging added
    *  No prometheus metrics added to this PR
    *  No updates
    *  No concerns


micahchiang (approved):
🙌

app/controllers/v0/education_career_counseling_claims_controller.rb
```diff
@@ -5,7 +5,7 @@ class EducationCareerCounselingClaimsController < ClaimsBaseController
     def create
       load_user
       claim = SavedClaim::EducationCareerCounselingClaim.new(form: filtered_params[:form])
-      claim.add_claimant_info(current_user) if current_user
+      claim.add_claimant_info(current_user) if current_user&.loa3?
```
erluti, line 8: can we get some tests for creating a claim that's not a loa3 user? (or one that does expect loa3 user?  interesting the existing tests didn't break)

kfrz (changes requested):
Looks good - after addressing Philip's testing comment.

app/controllers/v0/education_career_counseling_claims_controller.rb
digitaldrk: For sure. Tests added no problem.

> interesting the existing tests didn't break
Might be statin' the obvious but I was loading an LOA3 user in the tests. So essentially, the code was doing the same thing it was before we added these changes.

erluti (approved):
Thanks for the added tests!

kfrz (approved):

PR #5591, opened 2020-12-31
Adds changes to pdf keys

digitaldrk:
## Description of change
While working on [this PR](https://github.com/department-of-veterans-affairs/vets-api/pull/5590), we came across an issue where not all data was getting to the proper PDF fields. Turns out the keys we were looking for in the PDF code were inaccurate so the fields `stateCode` and `addressLine1` were returning `nil`.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#17975

  * No settings added
  * No sentry added in this PR
  * No Prometheus for this feature
  * No new swagger docs
  * No concerns


kfrz (approved):
:ship:

PR #5595, opened 2021-01-04
17772: STEM AD: TIMS Work

elder-joseph:
## Description of change
As a developer, I need to create a new TIMS doc type and corresponding spool file to record a permanent record of the STEM automated decision.

## Original issue(s)
https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/17772

## Testing
Additional test cases added for 10203DNY form.

### DIFF
--- a/app/models/form_profiles/va10203dny.rb
+++ b/app/models/form_profiles/va10203dny.rb
@@ -23,8 +23,6 @@ module VA10203dny
 end
 
 class FormProfiles::VA10203dny < FormProfiles::VA10203
-  VA10203dny.table_name = 'va10203dny'
-
   attribute :remaining_entitlement, VA10203dny::FormEntitlementInformation
   attribute :school_information, VA10203dny::FormInstitutionInfo
 end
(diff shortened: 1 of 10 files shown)


app/models/form_profiles/va10203dny.rb
```diff
@@ -0,0 +1,28 @@
...
+    attribute :state, String
+    attribute :country, String
+  end
+
+  class FormEntitlementInformation
+    include Virtus.model
+
+    attribute :months, Integer
+    attribute :days, Integer
+  end
```
shawkey-daniel, line 22: This code is duplicate of VA10203. Can we make it common to both?

shawkey-daniel (changes requested):

app/models/form_profiles/va10203dny.rb
dmccurdybah: I don't think we need this because the form_profile is used for the form prefill, but we won't have anyone actually filling out a va10203dny from the website.

shawkey-daniel: Good point. Makes sense.

PR #5598, opened 2021-01-04
[17915] BGS Awards Method Fix

camerontesterman:
## Description of change
The `find_award_by_participant_id` method now requires a second argument, the user's `ssn`. This is used as the identifier which allows mock data to be used in place of retrieving award amounts from BGS when developing locally.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#17915

Tested locally


erluti (approved):
LGTM

PR #5602, opened 2021-01-05
Refactor pii expunge

yakschuss:
## Description of change
This PR is a functionally complete refactor of the `remove_pii` proc defined both on NoD and  HLR. 
In preparation for combining the types into a single abstraction "Appeal", we're preparing to consolidate shared behavior into modules, and shared actions into services. This serves as an initial example of code that can be abstracted away from the model layer.

## Original issue(s)
[4229](https://vajira.max.gov/browse/API-4229)


modules/appeals_api/spec/workers/notice_of_disagreement_clean_up_week_old_pii_spec.rb
```diff
@@ -4,27 +4,14 @@
...
-      worker.perform
-    end
+      described_class.new.perform
 
-    it 'logs to sentry if no pii was removed /but/ some should have been' do
-      allow_any_instance_of(described_class).to(
-        receive(:no_notice_of_disagreements_are_ready_to_have_pii_expunged?).and_return(false)
-      )
-      worker = described_class.new
-      expect(worker).to receive(:log_message_to_sentry)
```
ElleeB, line 26: Are we concerned with testing our logging? 
I just ran into a spot where we weren't testing our logging, and it turns out that the logging wasn't actually working, which could have been caught with a test. Thoughts?

ElleeB (commented):
This looks awesome! See one comment about logging.

modules/appeals_api/spec/workers/notice_of_disagreement_clean_up_week_old_pii_spec.rb
yakschuss: I don't disagree with testing the logging in principle - although in this case, we're testing that the call to the Logging "class" or wrapper exists, which is done. Testing the logging functionality here would be out of scope.

ElleeB: K. I need to understand this more!

ElleeB (approved):
this looks great.

modules/appeals_api/app/services/appeals_api/remove_pii.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    def remove_pii!
+      records_to_be_expunged.update(form_data: nil, auth_headers: nil)
+    end
+
+    def validate_form_type!
+      raise ArgumentError, 'Invalid Form Type' unless valid_form_type
+    end
+
+    def valid_form_type
```
cilestin, line 38: Not sure this method is even necessary since ruby already gives us `in?`, and it's only used in one spot. That or it should be a predicate method since it's a question being asked of the data.

cilestin (approved):
LGTM! Had a comment but it's not a blocker for merging.

modules/appeals_api/app/services/appeals_api/remove_pii.rb
yakschuss: I agree it should be a predicate method! I think the `valid_form_type?` wrapper of the `in?` method is for readability. Is that what you're referring to?

PR #5603, opened 2021-01-05
15964 grafana updates

cferris32:
## Description of change
This story is preliminary work to create a new Grafana dashboard where we will group logs by endpoints and http status codes. For this new dashboard, we're going to track the status codes from our VAMF responses. We are now recording the HTTP status code as a tag in the statsd plugin.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#15964

Unit tests were added to test the updated statsd plugin logic.


grossb1:
There was an additional request in the original ticket to add url filtering to help streamline resulting urls in the VAOS Latency report.  This will be done in an associated ticket:
https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/18128

kfrz:
http://grafana.vfs.va.gov/d/zGS4XzkMk/backend-to-frontend-app-mapping?orgId=1

erluti:
Pull in the latest master to get past the security check

lib/breakers/statsd_plugin.rb
```diff
@@ -2,12 +2,15 @@
 
 module Breakers
   class StatsdPlugin
-    def get_tags(request)
+    def get_tags(request, response = nil)
       tags = []
       if request
```
saneshark, line 7: can request ever really be nil or false seeing as it is always passed in? 🤔 

I'm not sure if this condition is necessary.

lib/breakers/statsd_plugin.rb
```diff
@@ -2,12 +2,15 @@
...
         tags.append("endpoint:#{filtered_endpoint_tag(request.url.path)}") if request.url&.path
         tags.append("method:#{request.method}") if request.method
       end
+      if response
```
saneshark, line 11: I guess response could be nil if there was an exception. I'm curious if this http status would be captured before any other middleware would have possibly updated the http status code or raises an error. When would response be nil? In the event of a 404?

saneshark (approved):
I'm going to approve this because I know y'all just added to existing code and followed precedence, but I can't help to think that this code could be cleaned up a bit in future refactor.

ericboehs (approved):

lib/breakers/statsd_plugin.rb
grossb1: Good point, it may just there to prevent `request.method` from throwing if for whatever reason `request` is *nil*.  I don't think it's a bad gate though.

lib/breakers/statsd_plugin.rb
grossb1: Those are good questions.  I'd need to step through code to really answer that.

PR #5605, opened 2021-01-05
Add source field to NOD and HLR

nathanbwright:
This PR is part of [API-4180](https://vajira.max.gov/browse/API-4180).

## Description of change
To allow for metrics to be better reported we need to add a `source` field to both the NoticeOfDisagreement and HigherLevelReview models.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_12_31_033748) do
+ActiveRecord::Schema.define(version: 2021_01_05_174800) do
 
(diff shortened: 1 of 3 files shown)


yakschuss:
Is the fact that we're not adding this to NoticeOfDisagreement because NOD isn't live yet? Or do we anticipate different needs regarding the source key for that data model?

If the former, is there a reason we can't add it here?

nathanbwright:
> Is the fact that we're not adding this to NoticeOfDisagreement because NOD isn't live yet? Or do we anticipate different needs regarding the source key for that data model?
> 
> If the former, is there a reason we can't add it here?

Good point. I was just focused on HLR because it's the one that will be used first...

I'll add push up another migration file for NOD.

ElleeB (approved):

PR #5615, opened 2021-01-06
Refactor Login URL to not be a long method

jimiray:
## Description of change
Convert the case statement to use meta programming to send the url method

### DIFF
--- a/app/controllers/v1/sessions_controller.rb
+++ b/app/controllers/v1/sessions_controller.rb
@@ -181,24 +181,11 @@ module V1
     # rubocop:disable Metrics/CyclomaticComplexity
     def login_params(type)
       raise Common::Exceptions::RoutingError, type unless REDIRECT_URLS.include?(type)
-
-      case type
-      when 'signup'
-        url_service.signup_url
-      when 'mhv'
-        url_service.mhv_url
-      when 'dslogon'
-        url_service.dslogon_url
-      when 'idme'
-        url_service.idme_url
-      when 'mfa'
-        url_service.mfa_url
-      when 'verify'
-        url_service.verify_url
-      when 'custom'
+      if type == 'custom'
         raise Common::Exceptions::ParameterMissing, 'authn' if params[:authn].blank?
-
         url_service(false).custom_url params[:authn]
+      else
+        url_service.send("#{type}_url")
       end
     end
     # rubocop:enable Metrics/CyclomaticComplexity


jimiray:
This is something I did during the break, this method was just bugging me.

app/controllers/v1/sessions_controller.rb
```diff
@@ -181,24 +181,11 @@ def saml_cookie_content
...
-        url_service.mfa_url
-      when 'verify'
-        url_service.verify_url
-      when 'custom'
+      if type == 'custom'
         raise Common::Exceptions::ParameterMissing, 'authn' if params[:authn].blank?
-
         url_service(false).custom_url params[:authn]
+      else
+        url_service.send("#{type}_url")
```
bosawt, line 188:
```suggestion
        url_service.send("#{type}_url") if url_service.respond_to?("#{type}_url")
```

bosawt (commented):
That's funny, check `app/controllers/v0/sessions_controller.rb`, line 41. Looks like @bramleyjl  had the same instinct.

I think this is a better approach than the existing implementation, however, my concern here is the interpolation of `type`.  It looks like we set `type` on line 32 of this file, with `type = params[:type]`, then passing it down a couple levels and then using it here. I'm not sure if it's a super bit security issue in this case, but I could see some strange no_method errors coming from this.

Overall, my opinion of the right approach would be to refactor the `url_service` class (`PostURLService`) to deal with all this funkiness itself (so something like `url_service.url(url_type)` to deal with all of these functions), but given that we're not necessarily overhauling the entire url_service at this point, I wonder if my concern can be alleviated with just a `respond_to?` call?

app/controllers/v1/sessions_controller.rb
jimiray: Funny, I was thinking the same thing after digging deeper, this just doesn't belong here. Maybe I'll take a look at the PostURLService. I think I was just looking in this one file. Something like: ```url_service.url_for_type(type)```

bramleyjl (commented):
LGTM

app/controllers/v1/sessions_controller.rb
bosawt: Yeah, I feel like that approach is super straightforward. I just now did a quick check, and I think you'd only have to refactor the functions in both sessions controllers (`v0` and `v1`), so maybe wouldn't be the worst refactor ever

patrickvinograd:
One reason this may become a rabbit hole is that you're now calling `send` based on the `type` URL param which is considered unsafe by rubocop (since it could let somebody craft a URL to invoke an unexpected method). Rubocop is not quite smart enough to know that there's also an allow-list of possible type params.

PR #5618, opened 2021-01-06
Combine Hearing Preference and Additional Issues

yakschuss:
This PR combines the additional pages at the end of the generated Notice of Disagreement PDF. Initially we had a large amount of blank space underneath the `Hearing type requested` (copy changed), and this combines additional issues that are being contested to utilize that blank space.

## Original issue(s)
[4233](https://vajira.max.gov/browse/API-4233)


cilestin (changes requested):
Looks like the minimum pdf has an extra blank page at the end :(

cilestin (approved):
👍

PR #5620, opened 2021-01-06
retry form 526 rake task

annaswims:
## Original issue(s)

### DIFF
--- a/rakelib/form526.rake
+++ b/rakelib/form526.rake
@@ -230,4 +230,42 @@ namespace :form526 do
     end
     puts "reuploaded files for #{form_submissions.count} submissions"
   end
+
+  # Resubmit to EVSS
+  desc 'Resubmit entire claim'
+  task retry_526: :environment do |_, args|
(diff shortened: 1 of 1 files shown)


rakelib/form526.rake
```diff
@@ -230,4 +230,44 @@ namespace :form526 do
...
+  # Resubmit to EVSS
+  desc 'Resubmit entire claim'
+  task retry_submission: :environment do |_, args|
+    raise 'No Form526Submission ids provided' unless args.extras.count.positive?
+
+    def get_open_claims_from_evss(auth_headers)
+      open_claim_count = 0
+      raw_claims = EVSS::ClaimsService.new(auth_headers).all_claims.body
+      raw_claims.dig('open_claims').each do |open_claim|
+        open_claim_count += 1 if open_claim
```
annaswims, line 243: might need to do more here to verify that it's a claim for disability compensation.  And maybe we don't want to see if it's open, but only later than the failed submission.

PR #5621, opened 2021-01-06
Update appointment serializer to get id from VAMF

cferris32:
## Description of change
This work is to update the Appointments endpoint to return an appointment id from VAMF. We're replacing the current logic which generates an artificial id based on appointment start date.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#18205


saneshark (approved):
LGTM

PR #5623, opened 2021-01-07
Consolidate hlr/nod central mail status logic into concern

ElleeB:
## Description of change
Both HLR and NOD models contain identical code, controlling the updating of instance statuses to match the status returned from central mail. We want to consolidate this code into a shared concern. 

## Original issue(s)
https://vajira.max.gov/browse/API-4219

Refactoring validations (see raises) into tests

### DIFF
--- a/modules/appeals_api/app/models/concerns/appeals_api/central_mail_status.rb
+++ b/modules/appeals_api/app/models/concerns/appeals_api/central_mail_status.rb
@@ -1,0 +1,108 @@
+# frozen_string_literal: true
+
+module AppealsApi
+  module CentralMailStatus
+    extend ActiveSupport::Concern
+
+    include SentryLogging
+
+    included do
+      scope :received_or_processing, -> { where status: RECEIVED_OR_PROCESSING }
+
+      STATUSES = %w[pending submitting submitted processing error uploaded received success expired].freeze
+
+      validates :status, inclusion: { 'in': STATUSES }
+
+      CENTRAL_MAIL_STATUS_TO_APPEAL_ATTRIBUTES = lambda do
+        hash = Hash.new { |_, _| raise ArgumentError, 'Unknown Central Mail status' }
+        hash['Received'] = { status: 'received' }
+        hash['In Process'] = { status: 'processing' }
+        hash['Processing Success'] = hash['In Process']
+        hash['Success'] = { status: 'success' }
+        hash['Error'] = { status: 'error', code: 'DOC202' }
+        hash['Processing Error'] = hash['Error']
+        hash
+      end.call.freeze
+
+      raise "One or more CENTRAL_MAIL_STATUS_TO_APPEAL_ATTRIBUTES values is invalid" unless status_attributes_valid?
+
+      CENTRAL_MAIL_ERROR_STATUSES = ['Error', 'Processing Error'].freeze
+      raise "One or more CENTRAL_MAIL_ERROR_STATUSES is invalid" unless error_statuses_valid?
+
+      RECEIVED_OR_PROCESSING = %w[received processing].freeze
+      raise "One or more RECEIVED_OR_PROCESSING stats invalid" unless statuses_valid?(RECEIVED_OR_PROCESSING)
+
+      COMPLETE_STATUSES = %w[success error].freeze
+      raise "One or more COMPLETE_STATUSES is invalid" unless statuses_valid?(COMPLETE_STATUSES)
+
+      def update_status_using_central_mail_status!(status, error_message = nil)
+        begin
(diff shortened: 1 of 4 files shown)


modules/appeals_api/app/models/concerns/appeals_api/central_mail_status.rb
```diff
@@ -0,0 +1,83 @@
...
+        central_mail_status_objects = parse_central_mail_response(response).select { |struct| struct.id.present? }
+        ActiveRecord::Base.transaction do
+          central_mail_status_objects.each do |obj|
+            appeals.find { |h| h.id == obj.id }
+                   .update_status_using_central_mail_status!(obj.status, obj.error_message)
+          end
+        end
+      end
+
+      def date_from_string(string)
```
yakschuss, line 30: This method shouldn't be pulled out. It's not the responsibility of the CentralMail interaction, it's more likely something that would go along with PDF construction or header parsing.

modules/appeals_api/spec/models/concerns/appeals_api/central_mail_status_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+require 'rails_helper'
+# require Rails.root.join('spec', 'lib', 'sentry_logging_spec_helper.rb')
+
+describe AppealsApi::CentralMailStatus, type: :concern do
+  # include_examples 'a sentry logger'
+
+  let(:client_stub) { instance_double('CentralMail::Service') }
+  let(:faraday_response) { instance_double('Faraday::Response') }
+  let!(:upload) { create(:notice_of_disagreement) }
+  let(:in_process_element) do
```
yakschuss, line 12: What does "in_process_element" mean?

ElleeB: It's the expected body of a faraday response reporting an appeal with a current status of "In Process", but the appeal has not yet gone through the `update_status_using_central_mail_status!`, so still represents the Central Mail status that will then be used to update the appeal object's status attribute.

modules/appeals_api/app/models/concerns/appeals_api/central_mail_status.rb
ElleeB: Yeah, I think you're right. This is unrelated to Central Mail.

nathanbwright:
Looks great overall! One thing I thought of while I was looking over this was this - do you think it would be useful to use an RSpec shared example for testing the concern?

modules/appeals_api/spec/models/concerns/appeals_api/central_mail_status_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+            expect(upload.detail).to eq('Downstream status: You did a bad')
+          end
+        end
+      end
+    end
+  end
+
+  context 'when verifying model status structures' do
+    it 'fails if one or more CENTRAL_MAIL_STATUS_TO_APPEAL_ATTRIBUTES keys or values is mismatched' do
+      expect(upload.status_attributes_valid?).to be true
```
yakschuss, line 90: This pattern unfortunately still has some indirection, and I think we can make it a little clearer what's going on here!

We can remove the validity methods entirely by refactoring them into plain RSPec tests. For example, with the error status test, we can do something like this:

```
 statuses = subject::CENTRAL_MAIL_STATUS_TO_APPEAL_ATTRIBUTES.keys
 error_statuses = subject::CENTRAL_MAIL_ERROR_STATUSES
 expect(statuses).to include(*error_statuses)
 ```
 And we'd have the same functional test, without the meta programming or reaching into the model.

yakschuss: Another thing to note, is neither the model status method nor this approach validates whether a status is _missing_, it only checks to see if any added statuses aren't reflected properly against the main list of statuses. Not sure if that's what we want, just something to be aware of.

yakschuss (changes requested):
I think that we still have some room to remove the validity methods from the concern directly, and move them into a test. I provided an example in the comments

modules/appeals_api/spec/models/concerns/appeals_api/central_mail_status_spec.rb
ElleeB: on it

ElleeB: Reworked, but haven't address your second question. I'm going to dig into it a bit more and see if i have any recommendation one way or the other.

yakschuss (approved):
Looks good to me!

PR #5624, opened 2021-01-07
API-4263: 404 "Claim not found" error on PUT to /526 after successful POST

SlakrHakr:
## Description of change
No longer runs the ClaimEstablisher background job on POST submission if autoCestPDFGenerationDisabled is set to true. In this case, a 526 pdf form is required to be sent in through the PUT endpoint before the claim can be established.

Note: This only applies to v1. 

## Original issue(s)
https://vajira.max.gov/browse/API-4263

Updated existing broken specs and added new to handle this new workflow.
Added a bit of clarification to the developer portal documentation around both these changes and the fact that 526 POST submission is an asynchronous process.


bastosmichael (approved):
It makes sense if the logic has changed to move it up from the base class.

PR #5625, opened 2021-01-07
Fix quote mismatch in OneRecord description

LeKeve:
## Description of change
This PR updates the OneRecord app description and fixes mismatched quotation marks.

## Original issue(s)
https://vajira.max.gov/browse/API-4086

- updated app description, locally tested


bradenshipley (approved):
LGTM. The CLI to update the app directory is going to be rolled out soon and then we can stop doing this.

PR #5627, opened 2021-01-07
API-4074 Fix NOD header data retrieval

cilestin:
## Description of change
Model header data retrieval methods were using the wrong keys and returning blank strings, causing downstream calls to CentralMail to fail despite it being a valid API request.

## Original issue(s)


ElleeB (approved):
👍

PR #5628, opened 2021-01-07
[17781] FSR PDF Download

camerontesterman:
## Description of change

Adds an endpoint that allows veterans to download a copy of the their filled VA-5655 form.

### Additions and Changes
- `download_pdf` controller method
- Financial Status Report config added to redis.yml
- `download_pdf` route added
- FSR Downloader class added; contains one method to download the PDF and a private method which returns a hash of the headers used to make the call to the Debt Management Center
- Wrapper method added to FSR Service - `get_pdf`

## Original issue(s)
department-of-veterans-affairs/va.gov-team#17781


LindseySaari (commented):
Are there any swagger doc changes for the new download_pdf endpoint?

lib/debt_management_center/financial_status_report_downloader.rb
```diff
@@ -0,0 +1,35 @@
...
+    def initialize(financial_status_report)
+      @financial_status_report = financial_status_report
+    end
+
+    def download_pdf
+      uri = URI.parse(
+        "#{Settings.dmc.url}financial-status-report/documentstream?objectId=#{@financial_status_report.filenet_id}"
+      )
+      http = Net::HTTP.new(uri.host, uri.port)
+      request = Net::HTTP::Get.new(uri.request_uri)
```
LindseySaari, line 17: Instead of using Net::Http could the Common::Client service pattern be used here along with the existing perform method?

camerontesterman: Yes, I initially tried to go that route but it seemed to be more trouble than it's worth, as it's designed around accepting JSON responses by default, and there wasn't a good class between the SOAP and REST for handling this particular use case, where the response is coming back as binary data which in turn gets presented to the user as a PDF.

LindseySaari: Ah, ok. Your approach makes sense then!

LindseySaari:
Looks good, but just the one question about swagger docs

LindseySaari (approved):

PR #5630, opened 2021-01-07
Add More Tests for the InProgressForms Serializer

edmangimelli:
**No functionality changed or added** --just testing existing behavior.

Putting these tests in place for [another PR](https://github.com/department-of-veterans-affairs/vets-api/pull/5582) (so I know whether or not I'm breaking them).


spec/request/in_progress_forms_request_spec.rb
```diff
@@ -30,16 +30,89 @@
...
       let(:user) { loa3_user }
-      let!(:in_progress_form_edu) { FactoryBot.create(:in_progress_form, form_id: '22-1990', user_uuid: user.uuid) }
+      let!(:in_progress_form_edu) do
+        FactoryBot.create(:in_progress_form, :with_nested_metadata, form_id: '22-1990', user_uuid: user.uuid)
+      end
       let!(:in_progress_form_hca) { FactoryBot.create(:in_progress_form, form_id: '1010ez', user_uuid: user.uuid) }
 
       context 'when the user is not loa3' do
         let(:user) { loa1_user }
+        let(:json) { JSON.parse response.body }
```
erluti, line 40: I find this to be a bit of a confusing use of `let`.
[Rspec docs](https://www.rubydoc.info/github/rspec/rspec-core/RSpec%2FCore%2FMemoizedHelpers%2FClassMethods%3Alet) say `let` should be used to create state and tests can get confusing if you use more than a few.
These are effectively helper methods, so they could just be written as a method to communicate intent.
```
def json
  JSON.parse response.body
end
```
I'd also argue it'd need a better name, like `response_body_json`.
In my personal coding I think the tests are clearer when I just put the parse in-line.
I guess to sum up my concern, this is introducing a convention to these tests that doesn't exist elsewhere an ultimately makes it more cryptic.

edmangimelli: personally, I feel like `json` is a pretty good name for the parsed response body, in the context of request tests (i've used it for a while).

I leverage `let` a lot --dynamic scope fan-- but I can get the same result with `def` in this scenario.

I'm down for tweaking this if it makes it clearer 👍 

>this is introducing a convention to these tests that doesn't exist elsewhere
i've been doing this for a while, so, for better or worse, there is _some_ precedent (which maybe is tech debt now? 😬)

erluti: > i've been doing this for a while, so, for better or worse, there is some precedent (which maybe is tech debt now? 😬
LOL
I'll just keep my comments to this PR

ericboehs (approved):

PR #5632, opened 2021-01-07
API-4074 Fix NOD header data retrieval

cilestin:
## Description of change
Temporarily use ZIP `00000` to ensure we only need the zip (and not more of the address) to pass properly into centralmail.

## Original issue(s)


cilestin:
Bad push. Closing.

PR #5633, opened 2021-01-07
API-4074 Temporarily use 0s for centralmail zip

cilestin:
## Description of change
CentralMail requires a 5-digit zipcode in its metadata or else it will error. Supply it with 0's for now.

## Original issue(s)


nathanbwright (approved):
👍

PR #5634, opened 2021-01-07
Api 4180 - appeals_api report email

nathanbwright:
This PR addresses [API-4180](https://vajira.max.gov/browse/API-4180)

## Description of change
Initial implementation of email report for HLR and NODs.

### Some questions
- How often should this report go out? (will need to update sidekiq scheduler)
- Who all should be included?
- Should the report be turned on (via the `settings.yml`) by default?

### DIFF
--- a/modules/appeals_api/app/mailers/appeals_api/decision_review_mailer.rb
+++ b/modules/appeals_api/app/mailers/appeals_api/decision_review_mailer.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+require 'appeals_api/decision_review_report'
+
+module AppealsApi
+  class DecisionReviewMailer < ApplicationMailer
+    RECIPIENTS = %w[
+      example@va.gov
+    ].freeze
+
+    def build(date_from:, date_to:)
+      @report = DecisionReviewReport.new(from: date_from, to: date_to)
+
+      template = File.read(path)
+      body = ERB.new(template).result(binding)
+
+      mail(
+        to: RECIPIENTS,
+        subject: 'Decision Review API report',
+        content_type: 'text/html',
+        body: body
+      )
+    end
+
+    private
+
+    def path
+      AppealsApi::Engine.root.join(
+        'app',
+        'views',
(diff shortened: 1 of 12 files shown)


modules/appeals_api/app/mailers/appeals_api/decision_review_mailer.rb
```diff
@@ -0,0 +1,37 @@
...
+        to: RECIPIENTS,
+        subject: 'Decision Review API report',
+        content_type: 'text/html',
+        body: body
+      )
+    end
+
+    private
+
+    def path
```
yakschuss, line 27: It's odd this isn't done by default. Is this a result of the engine file structure?

yakschuss (approved):
Nice work!

modules/appeals_api/app/mailers/appeals_api/decision_review_mailer.rb
nathanbwright: That's what I'm assuming... Just copy-pasting what I found with the other mailers. lol

modules/appeals_api/app/workers/appeals_api/decision_review_report_job.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+require 'sidekiq'
+
+module AppealsApi
+  class DecisionReviewReportJob
+    include Sidekiq::Worker
+
+    def perform
```
cilestin, line 9: 1. Let's take the `to` and `from` dates as default keyword arguments so that if we need to run this manually due to a delivery failure, we can do it by a `AppealsApi::DecisionReviewReportJob.perform(from: 3.days.ago.utc)` or similar.
1. If the schedule is only set to run MON-FRI, how are we going to consolidate all the submissions that may have happened over the weekend? `1.day.ago` isn't going to cover it.

cilestin (changes requested):

cilestin (approved):
LGTM!

PR #5640, opened 2021-01-08
API-4074 Add zipcode into NOD schema

cilestin:
## Description of change
We require a zip code to talk to CentralMail, so this change makes `data.attributes.veteran.address.zipCode5` a required part of the NOD create form data.

Note: HLR is already requiring this, so the precedent has already been set.

## Original issue(s)

### DIFF
--- a/modules/appeals_api/app/swagger/appeals_api/v1/schemas/notice_of_disagreements.rb
+++ b/modules/appeals_api/app/swagger/appeals_api/v1/schemas/notice_of_disagreements.rb
@@ -43,6 +43,8 @@ module AppealsApi::V1
                     key :type, :string
                     key :description, '5-digit zipcode. Use "00000" if Veteran is outside the United States'
                     key :example, '20001'
+                    key :maxLength, 5
+                    key :minLength, 5
                   end
                 end
               end
--- a/modules/appeals_api/app/swagger/appeals_api/v1/schemas/notice_of_disagreements.rb
+++ b/modules/appeals_api/app/swagger/appeals_api/v1/schemas/notice_of_disagreements.rb
@@ -26,12 +26,24 @@ module AppealsApi::V1
               property :veteran do
                 key :type, :object
                 key :description, 'Veteran Object being submitted in appeal'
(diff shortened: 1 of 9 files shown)


modules/appeals_api/app/swagger/appeals_api/v1/schemas/notice_of_disagreements.rb
```diff
@@ -26,13 +26,27 @@ class NoticeOfDisagreements
...
                 property :homeless do
                   key :type, :boolean
                   key :example, false
                   key :description, 'Flag if Veteran is homeless'
                 end
+
+                property :address do
+                  key :type, :object
+                  key :description, 'Address of the Veteran'
+                  key :required, %i[zipCode5]
```
yakschuss, line 40: I think in this case you can add the `key :required, true` directly to zipCode5, but that's purely stylistic

modules/appeals_api/spec/support/pdf_matcher.rb
```diff
@@ -4,9 +4,16 @@
...
-    actual_text == expected_text
+    actual_reader = PDF::Reader.new(actual)
+    actual_pages = actual_reader.pages.size
+    actual_text = actual_reader.pages.map(&:text)
+
+    expected_reader = PDF::Reader.new(expected)
+    expected_pages = expected_reader.pages.size
+    expected_text = expected_reader.pages.map(&:text)
+
+    # Check both page count and text content in case there are extraneous blank pages
```
yakschuss, line 15: 👍

yakschuss (approved):

modules/appeals_api/app/swagger/appeals_api/v1/schemas/notice_of_disagreements.rb
cilestin: @yakschuss FYI - I tried doing that but it doesn't actually do anything other than add the `required` key to the zipCode5 entry. Doesn't even add the red asterisk to show it's required :( So had to keep this `required zipCode5` around regardless.

PR #5649, opened 2021-01-11
API-4354: Discover and document response when POA active endpoint is used by someone that is not POA

SlakrHakr:
## Description of change
Add note around eased authentication around active POA endpoint.

## Original issue(s)
https://vajira.max.gov/browse/API-4354

Viewed through local developer portal.


bastosmichael (approved):
I like the approach here.

PR #5650, opened 2021-01-11
Dangerfile whitespace

thilton-oddball:
## Description of change
This change is an attempt to remedy and ongoing issue with Danger counting whitespace when it calculates the number of lines of code changed. Can be run locally with `danger local `or `bundle exec danger local`.

## Original issue(s)
https://app.zenhub.com/workspaces/vsp-5cedc9cce6e3335dc5a49fc4/issues/department-of-veterans-affairs/va.gov-team/17249


Dangerfile
```diff
@@ -22,13 +22,25 @@ changed_files = git.diff.stats[:files]
...
 # ignores whitespace for the purpose of determining lines of code changed
-changes = `git diff master... -w --stat`.split("\n")
-lines_of_code = changes.sum(0) do |change|
-  if change == changes.last || EXCLUSIONS.any? { |exclusion| change.match(exclusion) }
-    0
-  else
-    change.match(/\|\s+(\d+)/)[1].to_i
+diff_changes = `git diff master..#{current_branch} -w --stat`.split("\n")
+
+if diff_changes.empty?
```
erluti, line 35: what's this case where the diff is empty?

thilton-oddball: I was coming up with some situations locally where it would sometimes be blank. I'm really uncertain whether or not that will happen in production or if it is just one of the issues with testing locally. In any case, if it comes back empty I want it to fail back to the way it was originally calculating the LOC changed. Better to have it count whitespace in LOC changed than to have it pass on something that has too many lines.

erluti: I think it's worth trying to discover what causes that case.  If this is inconsistent I don't know how helpful it will be if it sometimes flags PRs and sometimes doesn't because it did/didn't count whitespace.
I think catching fail cases is a good idea though, it'll help the dangerbot to run successfully instead of being blank.

ericboehs: It could be that you need to use a [three dot diff](https://stackoverflow.com/questions/462974/what-are-the-differences-between-double-dot-and-triple-dot-in-git-com) instead of two.

Dangerfile
```diff
@@ -22,13 +22,25 @@ changed_files = git.diff.stats[:files]
...
+  lines_of_code = filtered_changed_files.sum { |_file, changes| (changes[:insertions] + changes[:deletions]) }
+else
+  lines_of_code = diff_changes.sum(0) do |change|
+    if change == diff_changes.last || EXCLUSIONS.any? { |exclusion| change.match(exclusion) } || change.match(/\|\s+Bin/)
+      0
+    else
+      change.match(/\|\s+(\d+)/)[1].to_i
```
ericboehs, line 42: What's the regex doing here?

Dangerfile
```diff
@@ -22,13 +22,25 @@ changed_files = git.diff.stats[:files]
 excluded_changed_files = changed_files.select { |key| EXCLUSIONS.any? { |exclusion| key.include?(exclusion) } }
 filtered_changed_files = changed_files.reject { |key| EXCLUSIONS.any? { |exclusion| key.include?(exclusion) } }
 
+# fetch master for diff comparison
+`git fetch origin master`
+
+# get branch name
+branches = `git branch`
+current_branch = branches.match(/\*\s+([^\s]+)/m)[1]
```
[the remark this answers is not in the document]
thilton-oddball: Already made that changed but haven't pushed it yet.

ericboehs:
I've confirmed this works via:

```
danger --head=aff36d --base=aff36d~1 local | grep -A10 Errors:
```

This is the commit from https://github.com/department-of-veterans-affairs/vets-api/pull/5439 which should have failed Dangerbot but didn't.

Running the above command on master does not produce any errors. Running it on this branch does.

One thing to note, the script isn't exiting non-zero (i.e. it's exiting zero) so we may need to `exit 1` in order to fail CI.

Dangerfile
```diff
@@ -22,13 +22,25 @@ changed_files = git.diff.stats[:files]
...
-    0
-  else
-    change.match(/\|\s+(\d+)/)[1].to_i
+diff_changes = `git diff master..#{current_branch} -w --stat`.split("\n")
+
+if diff_changes.empty?
+  lines_of_code = filtered_changed_files.sum { |_file, changes| (changes[:insertions] + changes[:deletions]) }
+else
+  lines_of_code = diff_changes.sum(0) do |change|
+    if change == diff_changes.last || EXCLUSIONS.any? { |exclusion| change.match(exclusion) } || change.match(/\|\s+Bin/)
```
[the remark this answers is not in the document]
thilton-oddball: Yes, the notation for binaries that comes back from `git diff` is different than it is for any other file changes and the addition of a binary file isn't something that we want to necessarily count as LOC changes.

Dangerfile
```diff
@@ -22,13 +22,25 @@ changed_files = git.diff.stats[:files]
...
+current_branch = branches.match(/\*\s+([^\s]+)/m)[1]
+
 # ignores whitespace for the purpose of determining lines of code changed
-changes = `git diff master... -w --stat`.split("\n")
-lines_of_code = changes.sum(0) do |change|
-  if change == changes.last || EXCLUSIONS.any? { |exclusion| change.match(exclusion) }
-    0
-  else
-    change.match(/\|\s+(\d+)/)[1].to_i
+diff_changes = `git diff master..#{current_branch} -w --stat`.split("\n")
```
ericboehs, line 33:
```suggestion
diff_changes = `git diff master...#{current_branch} -w --stat`.split("\n")
```

This should be 3 dots here.

Dangerfile
thilton-oddball: As it iterates through the diff list of files changed, it is checking to see if it is the last line of the diff (which isn't an actual part of the list of files changed) or an excluded file and if so, do not count it in the sum. Otherwise, grab the lines of code changed for that file, make it an integer, and add it to the sum. An example of what it spits out is below. So we would want it to grab the 3 and the 1 and add them to the sum but ignore the last line that is "2 files changed, 4 insertions(+)".

 Dangerfile       | 3 +++
 config/routes.rb | 1 +
 2 files changed, 4 insertions(+)

Dangerfile
```diff
@@ -22,13 +22,25 @@ changed_files = git.diff.stats[:files]
 excluded_changed_files = changed_files.select { |key| EXCLUSIONS.any? { |exclusion| key.include?(exclusion) } }
 filtered_changed_files = changed_files.reject { |key| EXCLUSIONS.any? { |exclusion| key.include?(exclusion) } }
```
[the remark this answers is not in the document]
ericboehs: On second thought, I don't think I've named them correctly. Maybe just included_files vs excluded_files?

Dangerfile
thilton-oddball: I'm kind of fuzzy on this. I had initially had it as 3 dots but changed it back to two after reading some explanations about the two. My understanding from what I read was that using 3 dots does a diff comparison between your branch and where it branched off from master as opposed to where master currently is and that 2 dots does the diff comparison between your current branch and where master is currently.

Dangerfile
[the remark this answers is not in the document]
thilton-oddball: Yes, that might be more accurate.

Dangerfile
ericboehs: Three dot diff is what GitHub shows in the "Files Changed" tab.

I think we should be counting what's visible in that tab instead of something hidden.

Also, I believe if master changed significantly it's going to be counting the diffs from master. Say master deleted 1000 lines of code. Well now the current branch has a 1000 line lead in the line count race.

Take this branch for example, run this command:

```
❯ git diff origin/master..tjh-dangerfile --stat
...
128 files changed, 1208 insertions(+), 215727 deletions(-)
```
Now run it with three dots:
```
❯ git diff origin/master...tjh-dangerfile --stat
Dangerfile | 31 +++++++++++++++++++++----------
1 file changed, 21 insertions(+), 10 deletions(-)
```
That's the same as what GitHub shows.

Dangerfile
ericboehs: Maybe add a description to the regex then. Like so
```suggestion
      changed_lines_count_regex = /\|\s+(\d+)/ # Captures the number in: | 123 ++++
      change.match(changed_lines_count_regex)[1].to_i
```

Same for the binary regex above.

ericboehs (approved):
Looks good. Small change recommendation.

PR #5653, opened 2021-01-12
added require for Parallel and fixed docs

jperk51:
## Description of change
Required Parallel library because we believe it not being required is causing issues in staging. 

## Original issue(s)
No issue

Also fixed a typo in the api docs. No new tests.


modules/mobile/app/controllers/mobile/v0/claims_and_appeals_controller.rb
```diff
@@ -4,6 +4,7 @@
 require_relative '../../../models/mobile/v0/adapters/claims_overview'
 require_relative '../../../models/mobile/v0/adapters/claims_overview_errors'
 require_relative '../../../models/mobile/v0/claim_overview'
+require 'parallel'
```
erluti, line 7: parallel is in the gemfile for the test and development groups online.  To make it available here you should move it in the gemfile out of that group.

modules/mobile/docs/openapi.yaml
```diff
@@ -1467,7 +1467,7 @@ paths:
             application/json:
               schema:
                 $ref: "./schemas/Errors.yml"
-  /v0/claims-and-appeals/overview:
+  /v0/claims-and-appeals-overview:
```
erluti, line 1470: "fixing docs" seems like a different concept than fixing the parallel gem.  Those should probably be two separate PRs.  Maybe the change is small enough the reviewer will allow it, but in the future please try to have one concept per PR

jperk51: Gotcha, will do in the future

modules/mobile/app/controllers/mobile/v0/claims_and_appeals_controller.rb
jperk51: I see that there is a gem called "parallel_tests" would that be the appropriate gem to move?

erluti: You're right!  I overlooked it.  rubocop and parallel_tests both depend on parallel.  But both of those gems are in the test/development groups.  I'd just add `parallel` to the gemfile outside those groups.  So it'd be a new entry and not a "move".  I think the gems are generally in alphabetical order, so probably on line 91

erluti: I went ahead and pushed up what I was talking about

jperk51: Awesome thanks

ericboehs (approved):

PR #5654, opened 2021-01-12
18427 BGS Awards Lookup Fix

camerontesterman:
## Description of change
Lookup fails sometimes when looking up award amounts via participant id. Secondary look up within the `get_awards` method added to attempt the look up via SSN if first lookup fails.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#18427

Tested and verified locally using both BGS calls and mock data.

### DIFF
--- a/app/services/bgs/awards_service.rb
+++ b/app/services/bgs/awards_service.rb
@@ -11,9 +11,8 @@ module BGS
     # @return [Hash]
     #
     def get_awards
-      response = @service.awards.find_award_by_participant_id(@user.participant_id, @user.ssn)
-      response = @service.awards.find_award_by_ssn(@user.ssn) if response.nil?
-      response
+      @service.awards.find_award_by_participant_id(@user.participant_id, @user.ssn) ||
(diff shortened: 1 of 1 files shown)


app/services/bgs/awards_service.rb
```diff
@@ -11,7 +11,9 @@ class AwardsService < BaseService
     # @return [Hash]
     #
     def get_awards
-      @service.awards.find_award_by_participant_id(@user.participant_id, @user.ssn)
+      response = @service.awards.find_award_by_participant_id(@user.participant_id, @user.ssn)
+      response = @service.awards.find_award_by_ssn(@user.ssn) if response.nil?
+      response
```
ericboehs, line 16:
```suggestion
      @service.awards.find_award_by_participant_id(@user.participant_id, @user.ssn) ||
        @service.awards.find_award_by_ssn(@user.ssn)
```

What do you think of this?

ericboehs (approved):

PR #5655, opened 2021-01-12
Api-4361 pii and status flags and specs

ElleeB:
Adding flags for expunging PII and batch updating status for NOD and HLR.

Ticket: https://vajira.max.gov/browse/API-4361


cilestin (changes requested):
Tiny request but otherwise LGTM!

cilestin (approved):

PR #5656, opened 2021-01-12
Get lighthouse access_token

dillo:
## Description of change

- Provide the ability for the Health Quest module to request an access_token from Lighthouse so that the vets-api can use it to procure FHIR resources from that service for a signed-in user
- First, we check to see if a lighthouse session exists in Redis. If it does we return it back so that the user can access FHIR resources from the lighthouse.
- If the lighthouse session does not exist, then we make an HTTP POST call to the lighthouse service to obtain an access_token. This is then stored in Redis as a SessionStore and then returned back to the user so that they can use it during subsequent calls to the lighthouse.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#17904

- There are additions to `settings.yml` and the values will vary by environment.
- The entire feature is on a Flipper: in_progress_form_custom_expiration.

- RSpec tests written and passing.
- Manually obtained an access_token from Lighthouse by running the code under a break point.

### DIFF
--- a/modules/health_quest/app/services/health_quest/lighthouse/claims_token.rb
+++ b/modules/health_quest/app/services/health_quest/lighthouse/claims_token.rb
@@ -1,0 +1,85 @@
+# frozen_string_literal: true
+
+module HealthQuest
+  module Lighthouse
+    ##
+    # An object responsible for handling JWT claims used for Lighthouse authentication
+    #
+    class ClaimsToken
+      EXPIRATION_DURATION = 15
+      SIGNING_ALGORITHM = 'RS512'
+      TOKEN_PATH = '/v1/token'
+
+      ##
+      # Builds a Lighthouse::ClaimsToken instance
+      #
+      # @return [Lighthouse::ClaimsToken] an instance of this class
+      #
+      def self.build
+        new
+      end
+
+      ##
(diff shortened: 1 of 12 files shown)


modules/health_quest/app/services/health_quest/lighthouse/claims_token.rb
```diff
@@ -0,0 +1,85 @@
...
+      def expires_at_time
+        @expires_at_time ||= EXPIRATION_DURATION.minutes.from_now.to_i
+      end
+
+      def signing_algorithm
+        SIGNING_ALGORITHM
+      end
+
+      def private_key_path
+        Rails.env.development? || Rails.env.test? ? lighthouse.local_key_path : lighthouse.key_path
```
ericboehs, line 77: Why not just name this key_path in the settings.yml so you don't have to do the conditional?

ericboehs (approved):

modules/health_quest/app/services/health_quest/lighthouse/claims_token.rb
dillo: Good point. I Will address that in a separate PR in the nearby future.

PR #5657, opened 2021-01-13
Srsuddath 16367 add click tracking endpoint

srsuddath:
## Description of change
This change adds an endpoint to access search.gov's click tracking api

## Original issue(s)
department-of-veterans-affairs/va.gov-team#16367

This POST request is unique in that it sends all information as query params rather than as part of the body

Tests are still in progress

### DIFF
--- a/lib/search_click_tracking/service.rb
+++ b/lib/search_click_tracking/service.rb
@@ -38,7 +38,7 @@ module SearchClickTracking
     #
     def track_click
       with_monitoring do
-        response = perform(:post, track_click_url, query_params, {'Content-Type' => 'application/x-www-form-urlencoded'})
+        perform(:post, track_click_url, query_params)
       end
     rescue => e
       handle_error(e)
--- a/lib/search_click_tracking/service.rb
+++ b/lib/search_click_tracking/service.rb
@@ -1,0 +1,108 @@
+# frozen_string_literal: true
+
+require 'uri'
+require 'common/client/base'
+require 'common/client/concerns/monitoring'
+require 'common/client/errors'
+require 'search_click_tracking/response'
+require 'search_click_tracking/configuration'
+
+module SearchClickTracking
+  # This class builds a wrapper around Search.gov web results API. Creating a new instance of class
+  # will and calling #results will return a ResultsResponse upon success or an exception upon failure.
+  #
+  # @see https://search.usa.gov/sites/7378/api_instructions
+  #
+  class Service < Common::Client::Base
+    include Common::Client::Concerns::Monitoring
+
+    STATSD_KEY_PREFIX = 'api.search_click_tracking'
+
+    configuration SearchClickTracking::Configuration
+
+    attr_reader :url
+    attr_reader :query
+    attr_reader :position
+    attr_reader :client_ip
+    attr_reader :user_agent
+
+    def initialize(url, query, position, client_ip, user_agent)
+      @url = url
+      @query = query
+      @position = position
+      @client_ip = client_ip
+      @user_agent = user_agent
+    end
+
+    # POSTs click tracking query param data to search.gov 
+    #
+    def track_click
+      with_monitoring do
+        response = perform(:post, track_click_url, query_params, {'Content-Type' => 'application/x-www-form-urlencoded'})
+      end
+    rescue => e
+      handle_error(e)
+    end
+
+    private
+
+    def track_click_url
+      config.base_path
+    end
+
+    # Required params [affiliate, access_key, module_code, url, query, position, client_ip, user_agent]
+    #
+    # @see https://search.usa.gov/sites/7378/api_instructions
+    #
+    def query_params
+      URI.encode_www_form(
+        {
+          affiliate: affiliate,
+          access_key: access_key,
+          module_code: module_code,
+          url: url,
+          query: query,
+          position: position,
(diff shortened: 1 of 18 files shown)


Gemfile.lock
```diff
@@ -105,7 +105,6 @@ PATH
 
 GEM
   remote: https://rubygems.org/
-  remote: https://enterprise.contribsys.com/
```
annaswims, line 108: The changes to Gemfile.lock should be reverted. To avoid accindetally making these changes in the future see https://github.com/department-of-veterans-affairs/va.gov-team-sensitive/blob/master/platform/engineering/sidekiq-enterprise-setup.md

db/schema.rb
```diff
@@ -327,6 +327,15 @@
     t.index ["job_id"], name: "index_form526_job_statuses_on_job_id", unique: true
   end
 
+  create_table "form526_opt_ins", id: :serial, force: :cascade do |t|
```
annaswims, line 330: Why is this committed?

node_modules/.yarn-integrity
```diff
@@ -0,0 +1,10 @@
+{
```
annaswims, line 1: this file shouldn't be here.

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
annaswims, line 1: this file shouldn't be here.

db/schema.rb
srsuddath: removed!

node_modules/.yarn-integrity
srsuddath: removed!

yarn.lock
srsuddath: removed!

ncksllvn:
Hey @srsuddath I would use a different strategy here. I would build onto the existing search controller and service. I put together this branch to help get started, https://github.com/department-of-veterans-affairs/vets-api/compare/search-click-tracking.

That code is 99% untested but it sets up a POST endpoint at `http://localhost:3000/v0/search/track_search_result_click`.

ncksllvn:
cc @annaswims if you have a moment to skim for any quick major issues w/https://github.com/department-of-veterans-affairs/vets-api/compare/search-click-tracking

annaswims:
looks good so far @ncksllvn.   You can get the ip from`request.remote_ip` but you might want to check with VSP before passing it on.

srsuddath:
Hey all, I am creating this PR despite the tests failing. I have made some comments in the spec file, and if any backend folks can give it a look and point me in the right direction that would be super appreciated.

lib/search_click_tracking/service.rb
```diff
@@ -0,0 +1,108 @@
...
+      @position = position
+      @client_ip = client_ip
+      @user_agent = user_agent
+    end
+
+    # POSTs click tracking query param data to search.gov 
+    #
+    def track_click
+      with_monitoring do
+        response = perform(:post, track_click_url, query_params, {'Content-Type' => 'application/x-www-form-urlencoded'})
```
erluti, line 41: `query_params` is `"affiliate=va&access_key=TESTKEY&module_code=I14Y&url=&query=&position=&client_ip=&user_agent="` here, and the error seems to be the response isn't a json hash but `"Query can't be blank"`

spec/lib/search_click_tracking/service_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      VCR.use_cassette('search_click_tracking/success', VCR::MATCH_EVERYTHING) do
+        response = subject.track_click
+        expect(response.status).to eq 200
+      end
+    end
+  end
+
+  #need to mess up params here, which I think this does
+  let(:url) { '' }
+  let(:query) { '' }
```
erluti, line 34: this is the last `let` in the file, so this is what all the tests see.  If you want to override `let`s, put them in a `context` block with the test.
Try removing these and changing the `describe 'error handling'` block to something like this
```
describe 'with empty params' do
    let(:url) { '' }
    let(:query) { '' }
    let(:position) { '' }
    let(:client_ip) { '' }
    let(:user_agent) { '' }
    it 'raises a 400 exception' do
      VCR.use_cassette('search_click_tracking/failure', VCR::MATCH_EVERYTHING) do
        expect { subject.track_click }.to raise_error do |e|
          expect(e).to be_a(Common::Exceptions::BackendServiceException)
          expect(e.status_code).to eq(400)
          expect(e.errors.first.code).to eq('SEARCH_CLICK_TRACKING_400')
        end
      end
    end
  end
```

erluti (commented):
I think I found your issue.  Also be sure to run the linter

spec/lib/search_click_tracking/service_spec.rb
srsuddath: Thanks so much! I will make these changes and give them a test

srsuddath:
Closing this pull request to open a clean one with the new history

erluti:
> Closing this pull request to open a clean one with the new history

I don't think this is necessary.  The history is good and typically we squash and merge, so the final commit message can leave out all the "lint" commit messages (or other things that don't describe the change).

PR #5659, opened 2021-01-13
cleanup for generator specs

erluti:
Rspec was complaining about using "should" instead of "expect."  Also added some code to hide generator output during specs.


LindseySaari (approved):

spec/lib/generators/module/module_generator_spec.rb
```diff
@@ -4,6 +4,14 @@
 require 'generators/module/module_generator'
 
 describe ModuleGenerator do
+  before(:all) d
```
LindseySaari, line 7: Typo here

erluti: woops!  another reminder to run tests before commit/push, lol

PR #5662, opened 2021-01-13
Documentation Updates for Save-In-Progress Routes

edmangimelli:
For PR #5582

***NOTE:*** Some specs added to test the `InProgressFormsController#update` schema


ericboehs:
Could merge/rebase master in so we can see the actual diff?

edmangimelli:
> Could you merge/rebase master in so we can see the actual diff?

@ericboehs 

it looks right to me 🤷 

let me know if I need to take further action

spec/request/in_progress_forms_request_spec.rb
```diff
@@ -307,6 +306,32 @@ def response_body
...
+              headers: { 'CONTENT_TYPE' => 'application/json' }
+          expect(response).to have_http_status(:ok)
+        end
+
+        it 'can\'t have nil formData' do
+          put v0_in_progress_form_url(new_form.form_id)
+          expect(response).to have_http_status(:error)
+        end
+
+        it 'can\'t have non-hash formData' do
```
ericboehs, line 321: Instead of escaping, could you use `"` instead please?

ericboehs (approved):
Small request.

spec/request/in_progress_forms_request_spec.rb
edmangimelli: No prob! For some reason i though our rubocop settings were enforcing single quotes when not interpolating

PR #5667, opened 2021-01-13
added index on name column for directory_applications

bradenshipley:
## Description of change
Added unique index to `name` column so that uniqueness can be verified when hitting `#create` method of `DirectoryController` in PR to come shortly. This PR was made to seperate `db` and `app` changes.


charleystran (approved):
LGTM :shipit:

PR #5668, opened 2021-01-13
Improve claim submission error response if issues in record creation

SlakrHakr:
## Description of change
Added logic to return a 422 with the relevant error message if a claim submission creation fails for whatever reason.

## Original issue(s)
semi-related to https://vajira.max.gov/browse/API-3273

Tested locally and ensured all existing specs still pass.


hinzed1127 (approved):
good add 👍

PR #5672, opened 2021-01-14
18206 add show action

grossb1:
## Description of change

Currently, the Appointments list (#index) endpoint returns a years worth of appointments; this includes any relevant appointment details. The front-end would like to access specific appointment details without having to request the entire list. To provide this service the vets-api will expose a #show endpoint that will take an appointment id as specified in #18205 and return appointment details related to that appointment. This will allow the front-end to access an appointment's details without having to query the entire list each time.

## Original issue(s)
https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/18206

Integration and unit tests were added to cover these changes


modules/vaos/app/controllers/vaos/v0/appointments_controller.rb
```diff
@@ -40,6 +44,11 @@ def appointment_service
         VAOS::AppointmentService.new(current_user)
       end
 
+      def appointment
+        @appointment ||=
+          appointment_service.get_appointment(id)
```
ericboehs, line 49:
```suggestion
        @appointment ||= appointment_service.get_appointment(id)
```

modules/vaos/app/services/vaos/appointment_service.rb
```diff
@@ -92,6 +101,10 @@ def get_appointments_base_url(type)
       end
     end
 
+    def show_appointment_base_url(id)
```
ericboehs, line 104: I'd recommend `show_appointment_url` or perhaps `show_appointment_path`. `base_url` generally refers to the beginning part of a URL, not the end.

modules/vaos/app/services/vaos/appointment_service.rb
```diff
@@ -17,6 +17,15 @@ def get_appointments(type, start_date, end_date, pagination_params = {})
       end
     end
 
+    def get_appointment(id)
+      params = {}
+
+      with_monitoring do
+        response = perform(:get, show_appointment_url(id), params, headers, timeout: 55)
```
saneshark, line 24: Get appointment with an ID should return much faster than get appointments, so lets not pass a timeout at all and use default. 

Lets also open a ticket to investigate latency issues with get_appointments, because the present value of 55 appears extreme.

modules/vaos/app/services/vaos/appointment_service.rb
```diff
@@ -69,7 +78,7 @@ def deserialized_appointments(json_hash, type)
                          end
       return [] unless appointment_list
 
-      appointment_list.map { |appointments| OpenStruct.new(appointments) }
+      appointment_list.map { |appointment| OpenStruct.new(appointment) }
```
saneshark, line 81: good call!

modules/vaos/config/routes.rb
```diff
@@ -5,6 +5,7 @@
     resources :appointments, only: %i[index create] do
       put 'cancel', on: :collection
     end
+    get '/appointments/:type/:id', to: 'appointments#show', type: /va/
```
saneshark, line 8: we can limit the REXEX with `/\Ava\z/` so that it doesn't match on vaos for example.

saneshark:
Overall this looks good to me other than the minor changes I've called out.

modules/vaos/config/routes.rb
grossb1: We tried adding the `\A\z` anchors to the regex and Ruby didn't like that.

modules/vaos/app/services/vaos/appointment_service.rb
```diff
@@ -17,6 +17,15 @@ def get_appointments(type, start_date, end_date, pagination_params = {})
       end
     end
 
+    def get_appointment(id)
+      params = {}
+
+      with_monitoring do
+        response = perform(:get, show_appointment_url(id), params, headers)
```
saneshark, line 24: maybe just pass empty hash instead of assigning to params?

saneshark (approved):
approving, but i'd take a look at that params empty hash thing -- if its there because rubocop complained thats just a rule I don't agree with.

PR #5674, opened 2021-01-14
Address 500's for when payments are empty

digitaldrk:
## Description of change
We recently launched view payments functionality 🥳 which allows vets to see their payment and return payment history.

We were alerted to rare 500 errors occurrences where we were calling `[]` on nil.
image

This PR checks for the `payment` hash so we don not call it on `nil`. It also adjusts the empty default payload that we had set in the service.

What changed in the empty default payload/Why was it not right in the beginning? The short answer is I'm not sure. Throughout the development of this feature, we had to stumble through a few services to find the one that was _just_ right; the service that had everything we wanted. I think we had set the empty payload according to one of those services and then once we got to the one we're using now, never updated the empty default to reflect what the service actually returns.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#18440

  * Negative
  * No FF
  * Not added in this PR
  * no
  * no
  * no concerns

### DIFF
--- a/app/services/bgs/payment_service.rb
+++ b/app/services/bgs/payment_service.rb
@@ -21,7 +21,7 @@ module BGS
     private
 
     def empty_response
-      { payments: [], return_payments: [] }
+      { payments: {payment: []}}
     end
   end
 end
--- a/app/services/bgs/payment_service.rb
+++ b/app/services/bgs/payment_service.rb
@@ -9,7 +9,6 @@ module BGS
         '00', # payee code
         person[:ssn_nbr]
       )
-
       return empty_response if response[:payments].nil?
 
       response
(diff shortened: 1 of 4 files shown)


app/services/bgs/payment_service.rb
```diff
@@ -21,7 +20,7 @@ def payment_history(person)
     private
 
     def empty_response
-      { payments: [], return_payments: [] }
```
ericboehs, line 24: Was `return_payments` not used here?

ericboehs (approved):

app/services/bgs/payment_service.rb
digitaldrk: Turns out the return payments are actually `payments` with special attributes. We check for truthiness in the serializer and build them out.

https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/serializers/vet_payment_history_serializer.rb#L35

PR #5677, opened 2021-01-14
Api 4306/pdf constructor refactor

yakschuss:
This is a refactor of the PDF construction being used by AppealsApi, in order to incorporate flexibility, ease of change, and readability for future form types and additional pdf changes.

## Description of change
This abstracts away the PDF construction to a single API, `AppealsApi::PdfConstruction::Generator.new(appeal).generate`.

It also separates form specific structure using composition, and enables versioning of specific form types by separating the form field locations that are used by pdftk for template form filling and form data fills.

## Original issue(s)
[4306](vajira.max.gov/browse/API-4306)

This PR doesn't incorporate the upcoming HLR refactor, so the changes here don't remove old constructor code just yet.


modules/appeals_api/app/services/appeals_api/pdf_construction/notice_of_disagreement/v1/form_data.rb
```diff
@@ -0,0 +1,132 @@
...
+          board_review_option == 'hearing' ? 1 : 'Off'
+        end
+
+        def extra_contestable_issues?
+          contestable_issues.size > 5 ? 1 : 'Off'
+        end
+
+        def soc_opt_in?
+          notice_of_disagreement.form_data&.dig('data', 'attributes', 'socOptIn') ? 1 : 'Off'
+        end
```
nathanbwright, line 61: Should these be predicate methods?

modules/appeals_api/app/services/appeals_api/pdf_construction/notice_of_disagreement/v1/structure.rb
```diff
@@ -0,0 +1,142 @@
...
+
+        # rubocop:disable Metrics/AbcSize
+        def form_fill
+          options = {
+            form_fields.veteran_name => form_data.veteran_name,
+            form_fields.veteran_ssn => form_data.veteran_ssn,
+            form_fields.veteran_file_number => form_data.veteran_file_number,
+            form_fields.veteran_dob => form_data.veteran_dob,
+            form_fields.mailing_address_number_and_street => form_data.mailing_address_number_and_street,
+            form_fields.homeless? => form_data.homeless?,
```
nathanbwright, line 21: Do the `form_fields` predicate methods indicate a checkbox type field?

nathanbwright (commented):
Really liking this overall! Nice work!

I left a couple of comments/questions, I may just be missing context...

modules/appeals_api/app/services/appeals_api/pdf_construction/generator.rb
```diff
@@ -0,0 +1,94 @@
...
+
+        # This line is due to HLR being live when the updated stamp was added.
+        # Once HLR bumps a version, we should refactor NoD's stamp method to be
+        # generic to HLR/NOD/SC. For now, the HLR#Structure.stamp method will
+        # just return the stamped path.
+        structure.stamp(stamped_pdf_path)
+      end
+
+      def combine_form_fill_and_additional_pages(additional_pages_added_path)
+        path = "/tmp/#{appeal.id}-completed-unstamped-tmp.pdf"
```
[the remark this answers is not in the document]
yakschuss: Do you mean the root dir in terms of using something other than `tmp/` to house these temp files? I'm open to suggestions, don't think it occurred to me.

modules/appeals_api/app/services/appeals_api/pdf_construction/notice_of_disagreement/v1/form_data.rb
yakschuss: Are you saying that since they're not actually returning a true Boolean value ('Off' instead of false), they shouldn't have the `?`? I figured the form field is basically a true false value in the form of a checkbox, but I'm not sure now.

modules/appeals_api/app/services/appeals_api/pdf_construction/notice_of_disagreement/pages/hearing_type_and_additional_issues.rb
```diff
@@ -0,0 +1,63 @@
...
+          def hearing_type_text
+            return if no_hearing_type?
+
+            "\nHearing Type Preference: #{form_data.hearing_type_preference.humanize}\n"
+          end
+
+          def extra_issues_table_data
+            header = ['A. Specific Issue(s)', 'B. Date of Decision']
+
+            data = form_data.contestable_issues.drop(5).map do |issue|
```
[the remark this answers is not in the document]
yakschuss: Good point!

modules/appeals_api/app/services/appeals_api/pdf_construction/notice_of_disagreement/v1/structure.rb
yakschuss: Essentially, yes.

modules/appeals_api/app/services/appeals_api/pdf_construction/generator.rb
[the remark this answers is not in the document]
cilestin: Good call out @nathanbwright but I think we'll be ok to leave it as is, at least until we find out there's a concrete need to make the extraction. The filename being set to the appeal's id means we can find its artifacts in the directory fairly easily even if we're using a different version later down the line.

modules/appeals_api/app/services/appeals_api/pdf_construction/notice_of_disagreement/v1/form_data.rb
cilestin: I'm having the same question. Can't decide whether the FormData or the Structure should take care of knowing exactly how to tie the model's data into the pdf fields :(

nathanbwright: I'm thinking the predicate style makes sense in the context of the form... _But_ I'm not sure if it makes sense overall, for example if someone in the future was writing some logic around the form data and they attempted to do something like:


```
if direct_review?
  # ...
end
```

I'm torn about this... I like it in this immediate context, but I think it could be somewhat confusing too...

yakschuss: I think it's worth pointing out that the `FormData` class is specific to data _for the form_. The value methods contained in that class are specifically to answer questions about the form in question, which, in this case is `NoticeOfDisagreement::V1`

ElleeB (commented):
I'm so happy about these changes! This looks awesome. I am interested in the ongoing 'predicate method' thread.

ElleeB (approved):
:)

cilestin (approved):

PR #5683, opened 2021-01-15
Api 4523 fix already defined constants warnings

ElleeB:
This PR fixes the warnings introduced here: https://github.com/department-of-veterans-affairs/vets-api/pull/5623
Ticket: https://vajira.max.gov/browse/API-4523

---

- The central mail concern is included in both our HLR and NOD classes
- Via the concern, constants were defined within the `included` block
- The `included` block was running for each HLR and NOD, thus the code was attempting to redefine the constants
- Removing the constants from the `included` block allows the definitions to occur one time only


yakschuss (approved):
Looks good! Thanks for adding the explanation.

PR #5684, opened 2021-01-15
14853 mobile GET appointments fixes

kreek:
## Description of change
Fixes a couple issues found during appointments staging integration testing. 

- Fixes missing or invalid startDate and endDate params handling
- Fixes the params format in the cal to the facilities API that retrives VA location data 

Also updates the endpoint to still return some appointments if one of the VA or Community Care appointment services is down.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#14853


jperk51 (approved):
LGTM

PR #5685, opened 2021-01-15
18330 - Create show action for appointment requests

ajmagdub:
## Description of change

The front-end would like to access specific request details without having to request the entire list. To provide this service the vets-api will expose a #show endpoint that will take a request id as specified in #18205 and return request details related to that request. This will allow the front-end to access an request's details without having to query the entire list each time.

## Original issue(s)
https://app.zenhub.com/workspaces/vft-59c95ae5fda7577a9b3184f8/issues/department-of-veterans-affairs/va.gov-team/18330

System and request tests were added to cover these updates


LindseySaari (commented):
Looks good to me, but there's a danger CI issue to address

LindseySaari (approved):

modules/vaos/spec/request/appointment_requests_request_spec.rb
```diff
@@ -10,6 +10,8 @@
     allow_any_instance_of(VAOS::UserService).to receive(:session).and_return('stubbed_token')
   end
 
+  let(:inflection_header) { { 'X-Key-Inflection' => 'camel' } }
```
saneshark, line 13: Looks like on line 95 you could use this rather variable instead of writing it out, possibly other places.

modules/vaos/app/controllers/vaos/v0/appointment_requests_controller.rb
```diff
@@ -46,6 +46,12 @@ def update
         render json: VAOS::V0::AppointmentRequestsSerializer.new(response[:data])
       end
 
+      def show
+        response = appointment_requests_service.get_request(id)
+        log_appointment_request(response)
```
saneshark, line 51: no need to log for idempotent operation

saneshark (approved):
Mostly nitpicks but please address the unnecessary logging one at least.

grossb1:
Decided to omit `codeNumeric` during the serialization process seeing that it was never intended to be returned in the response.

PR #5687, opened 2021-01-15
M1 Mac Docker Support

ericboehs:
Running `make up` previously failed on M1 Macs running Docker. This update to pg_query fixes it.


erluti (approved):
looks like only upside in that minor version update

PR #5693, opened 2021-01-18
Fix deadlock issue with pdf_info on some bad pdfs.

cshupp1:
## Description of change
Current State:
When a customer uploads a bad pdf pdf_info.rb invoked as follows:

```
PdfInfo::Metadata.read('bad_pdf')
```
Goes into deadlock causing sidekiq to timeout the job.

Expected State:
The code should not deadlock and 
```
PdfInfo::Metadata.read('bad_pdf')
```
should return.

Steps to reproduce:

In a rails c:
```
require_dependency 'vba_documents/multipart_parser'
require_dependency 'vba_documents/payload_manager'
require_dependency 'vba_documents/pdf_inspector'
parts = VBADocuments::MultipartParser.parse('bob')
file = parts['content']Open3.popen2e('pdfinfo', file.path) do |_stdin, stdout, wait|
PdfInfo::Metadata.read(file.path) #blocks forever
```
The cause is:
 ```
Open3.popen2e('pdfinfo', file.path) do |_stdin, stdout, wait|
#exit_status = wait.value
stdout.each_line do |line|
p line
end
exit_status = wait.value
end
```

## Original issue(s)
https://vajira.max.gov/browse/API-4566


gpbowman-git (approved):
looks good

PR #5695, opened 2021-01-18
14853 mobile appointments hidden status

kreek:
## Description of change
Some appointments statuses should be hidden. i.e. the status itself is hidden not that the appointment is hidden. This was unaccounted for in the original PR and was causing a bug in staging. Specs updated to use fixture data from staging.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#14853


jperk51 (approved):
LGTM

PR #5699, opened 2021-01-18
Removing GWOT from Flashes after production testing

bastosmichael:
## Description of change
Production testing showed that anything GWOT related is unchangeable in Flashes so removing from schema list.

## Original issue(s)
https://vajira.max.gov/browse/API-4275


bradenshipley (approved):
LGTM

nathanbwright (approved):
👍

PR #5701, opened 2021-01-18
Add Rails cache to rated disabilities

annaswims:
## Description of change
In preparation for calling the rated_disabilities endpoint more often (https://github.com/department-of-veterans-affairs/va.gov-team/issues/18681), we're going to start caching it. 

I have no idea why we write a bunch of code to cache when `Rails.cache.fetch` will do. So, I updated the rails config to `config.cache_store = :redis_cache_store` for dev/prod. 

It's [been](https://dsva.slack.com/archives/CBU0KDSB1/p1585674077228800?thread_ts=1585673379.225700&cid=CBU0KDSB1) [discussed](https://dsva.slack.com/archives/C0MQ281DJ/p1592339101258300?thread_ts=1592313140.256100&cid=C0MQ281DJ) a bit before and nobody had a reason not to use it.  

## Original issue(s)

When I ran the tests locally I had `config.cache_store = :redis_cache_store` but I didn't think we'd want to change the setting for all tests.   I will validate in stating after merge as well


config/environments/production.rb
```diff
@@ -84,7 +84,7 @@
   # config.logger = ActiveSupport::TaggedLogging.new(SyslogLogger.new)
 
   # Use a different cache store in production.
-  # config.cache_store = :mem_cache_store
+  config.cache_store = :redis_cache_store
```
ericboehs, line 87: What is the key eviction policy for our Redis? Typically for Rails cache you'll want an LRU or similar strategy. But for Sidekiq you most definitely do not want an LRU. You can set different policies per db index but you'll need to specify the index to connect to somewhere.

annaswims: I think @omgitsbillryan might be able to tell you more about it.   All of the models that inherit from `Common::RedisStore` are using `Redis.current` as a cache... just with a lot more code to get there. 

In prod there are two different redis servers, one for app_data and one for sidekiq. 

https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/deployment/config/vets-api/prod-settings.local.yml.j2#L152

ericboehs: @omgitsbillryan Any information on how our redis is setup?

omgitsbillryan: Models that inherit from `Common::RedisStore` also have an entry in [`redis.yml`](https://github.com/department-of-veterans-affairs/vets-api/blob/master/config/redis.yml) that configures the `ttl` in seconds (aka how long they persist before redis clears it out). This [ttl is refreshed each time the model is re-saved](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/common/models/redis_store.rb#L91). Not sure if that's the info you seek.

omgitsbillryan: Side note, you might be interested [`CacheAside` module](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/common/models/concerns/cache_aside.rb), which was intended to cache responses from backend services to lighten the load on them.

omgitsbillryan: Also, don't you need to provide config to point to the proper redis? i.e.

```
config.cache_store = :redis_cache_store, {url: "redis://xxx.xxx.xxx.xxx:6379/0"}
```
By default it assumes `localhost:6379` AFAIK.

omgitsbillryan: You'll probably need to stand up a separate redis instance for this:

> Deployment note: Take care to use a *dedicated Redis cache* rather than pointing this at your existing Redis server. It won't cope well with mixed usage patterns and it won't expire cache entries by default.

([source](https://api.rubyonrails.org/classes/ActiveSupport/Cache/RedisCacheStore.html))

Apologies if all of this was already discussed.

ericboehs: I believe it will read from `ENV['REDIS_URL']`. As long as it's writing/reading from the right Redis DB and not exhausting our redis storage I think I'm good with it.

@annaswims Do you know the default TTL here?

ericboehs: Oh, looks like there is not TTL as per Ryan's comment that I overlooked.

So I was under the impression the key eviction policy on our Redis was set to an LRU-like setting? But from your last comment I feel that may not be the case?

lib/evss/disability_compensation_form/service.rb
```diff
@@ -36,12 +36,15 @@ def initialize(headers)
...
       #
       def get_rated_disabilities
+        if @headers['va_eauth_birlsfilenumber'].blank?
+          Rails.logger.info('Missing `birls_id`', edipi: @headers['va_eauth_dodedipnid'])
+        end
         with_monitoring_and_error_handling do
-          if @headers['va_eauth_birlsfilenumber'].blank?
-            Rails.logger.info('Missing `birls_id`', edipi: @headers['va_eauth_dodedipnid'])
+          Rails.cache.fetch("evss_rated_disabilities/#{@headers['va_eauth_dodedipnid']}-#{@headers['va_eauth_pnid']}",
+                            expires_in: 30.minutes) do
```
annaswims, line 44: the `expires_in: 30.minutes` sets the ttl.

config/environments/production.rb
annaswims: The CacheAside is just so much code for what could be a one-liner. I know it's there, but I don't see why I'd want to use it.  It'll take just a bit of work to verify the configuration for `Rails.cache` in this pr. 

@omgitsbillryan - I don't follow why you say we'd need  third redis server to use this.  We already have two redis servers in prod.  `app_data` is essentially used for cache data already (using the CacheAside module)  and `sidekiq` for sidekiq.  

I'll take some time to verify that we don't need to pass the redis url, but I think it's using the `Redis.currrent` url that's already configured.

annaswims: Good catch, I thought  `Rails.cache.redis.connection` was defaulting to `Redis.current.connection` but I was mistaken. They just happened to have the same value.

annaswims:
I've update the config to explicitly pass in the connection url and have a default ttl/expiry of 30 min so I think this should be good to go.

config/environments/production.rb
omgitsbillryan: +1 for `Rails.cache` is the future, just wanted to make you aware of `CacheAside` in case you didn't want to setup the plumbing that will be required for `Rails.cache` 👍 .

> Take care to use a dedicated Redis cache rather than pointing this at your existing Redis server.

`app_data` is our "existing Redis server" that the rails doc specifically advises against using. I can only really guess as to why - I bet `Rails.cache` makes some assumptions that it's the only thing CRUD'ing entries in and out of Redis. Such an assumption would be false if we pointed `Rails.cache` at `app_data` since all our RedisStore models are constantly manipulating data in that Redis instance.

annaswims: @dginther would you or someone else in Ops be able to weigh in on this?  
[rails docs](https://api.rubyonrails.org/v5.2.0/classes/ActiveSupport/Cache/RedisCacheStore.html) say 
>Deployment note: Take care to use a *dedicated Redis cache* rather than pointing this at your existing Redis server. It won't cope well with mixed usage patterns and it won't expire cache entries by default.

But we're already using the [app_data redis](https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/deployment/config/vets-api/prod-settings.local.yml.j2#L152) as a cache server.  I'm assuming it's set up as a [lru cache](https://redis.io/topics/lru-cache). I'm also assuming that the [sidekiq](https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/deployment/config/vets-api/prod-settings.local.yml.j2#L155) is configured as a persistent store.  But i don't have a way of verifying that.  

I'm pretty sure that what the rails docs are trying to say is "don't use the same redis for cache and persistent store", which we're not.

annaswims: @cohnjesse 👋 This is the thread I was referring to in Ops office hours. Thanks!

config/environments/development.rb
```diff
@@ -31,6 +31,8 @@
   # Raise an error on page load if there are pending migrations.
   # config.active_record.migration_error = :page_load
 
+  config.cache_store = :redis_cache_store, { url: Settings.redis.app_data.url, expires_in: 30.minutes }
```
annaswims, line 34: I totally picked 30.minutes arbitrarily, but it seemed like a good start since it's the default session length.  I'm assuming that most will have an explicit override for each kind of cache.

annaswims:
note to self https://github.com/department-of-veterans-affairs/devops/pull/8425 was merged. so change  Settings.redis.app_data.url to to Settings.redis.rails_cache.url

omgitsbillryan (approved):
LGTM. 

WRT testing, I think the 24-hour cycle in dev & staging should suffice, but you could always add a flipper conditional if you need more time.

omgitsbillryan (approved):

LindseySaari (approved):
👍 This looks good. Feel free to use a feature toggle if you see any issues in dev/staging when testing

PR #5703, opened 2021-01-19
API-4490 Move boilerplate pages to end of NOD form

cilestin:
## Description of change
Moves the NOD boilerplate pages to the end of the document when necessary.

## Original issue(s)
https://vajira.max.gov/browse/API-4490

See also
NOD PDF refactor: https://github.com/department-of-veterans-affairs/vets-api/pull/5677


modules/appeals_api/app/services/appeals_api/pdf_construction/notice_of_disagreement/pages/hearing_type_and_additional_issues.rb
```diff
@@ -27,7 +27,7 @@ def build!
 
           private
 
-          attr_accessor :pdf, :form_data, :notice_of_disagreement
+          attr_accessor :pdf, :form_data
```
yakschuss, line 30: Good catch

modules/appeals_api/app/services/appeals_api/pdf_construction/generator.rb
```diff
@@ -52,6 +55,15 @@ def insert_additional_pages
         combine_form_fill_and_additional_pages(additional_pages_added_path)
       end
 
+      def rebuild_pages
```
yakschuss, line 58: Is this rebuilding the completed PDF, or reordering it? Is the reordering specific to NOD, or is this going to be a final pass through step for every PDF?

modules/appeals_api/app/services/appeals_api/pdf_construction/generator.rb
```diff
@@ -14,9 +14,12 @@ def generate
         @form_fill_path = fill_unique_form_fields
         #=> '{appeal.id}-tmp.pdf' OR '/tmp/#{appeal.id}-overlaid-form-fill-tmp.pdf'
 
-        @unstamped_path = insert_additional_pages
+        @all_pages_path = insert_additional_pages
```
yakschuss, line 17: My only concern is the name of this variable, not sure if it's indicative of what the process is past this point.

modules/appeals_api/app/services/appeals_api/pdf_construction/generator.rb
cilestin: Was having trouble naming this because I wanted to try to get across that rebuilding could be a rearranging of the pages and/or a removal of pages, since that's come up a few times. I was thinking this would be a step for all PDFs, even if (like with add_additional_pages) the step is skipped by returning the original path for forms where it's not needed.

modules/appeals_api/app/services/appeals_api/pdf_construction/generator.rb
cilestin: Yeah I wasn't sure what to call this either. It's not the final pdf, since the rebuild step might take place, but it does have all the pages for the final pdf...happy to rename to something else if you've got suggestions :)

modules/appeals_api/app/services/appeals_api/pdf_construction/generator.rb
ElleeB: format_pages?

modules/appeals_api/app/services/appeals_api/pdf_construction/generator.rb
ElleeB: I like _all_pages_, since it's exactly what it is. However, I could get behind using a word like _prepared_ to indicate this collection of pages may be destined for more!

modules/appeals_api/app/services/appeals_api/pdf_construction/generator.rb
```diff
@@ -52,6 +55,15 @@ def insert_additional_pages
         combine_form_fill_and_additional_pages(additional_pages_added_path)
       end
 
+      def finalize_pages
```
yakschuss, line 58: Maybe I'm just not seeing it, but where is this called?

nathanbwright (approved):
Looks good to me!

PR #5704, opened 2021-01-19
13020 mobile discovery endpoint

jperk51:
## Description of change
Discovery endpoint docs for mobile api

## Original issue(s)
department-of-veterans-affairs/va.gov-team#13020

Docs only PR.


modules/mobile/docs/schemas/DiscoveryResponse.yml
```diff
@@ -0,0 +1,55 @@
...
+        type: string
+        example: "1.0"
+        description: API Version number
+      type:
+        type: string
+        example: "discovery"
+      attributes:
+        type: object
+        required:
+          - endpoints
```
lexicalninja, line 22: can we add apiRoot here as well, since we are returning different values for dev/staging and prod?

jperk51: Added

lexicalninja (approved):

modules/mobile/docs/schemas/DiscoveryResponse.yml
```diff
@@ -0,0 +1,67 @@
...
+          - displayMessage
+        properties:
+          endpoints:
+            type: object
+            additionalProperties:
+              type: object
+            example:
+              appointments: {
+                url: "/v0/appointments",
+                verb: "GET"
```
kreek, line 36: could we change `verb` to `method` as that's the terminology used in other parts of the API

kreek (approved):
One minor comment but otherwise LGTM 🚀

modules/mobile/docs/schemas/DiscoveryResponse.yml
jperk51: sure thing

PR #5705, opened 2021-01-19
13020 mobile discovery endpoint implementation

jperk51:
## Description of change
Implementation of the mobile discovery endpoint

## Original issue(s)
department-of-veterans-affairs/va.gov-team#13020

Tests are in discovery_spec


modules/mobile/app/controllers/mobile/discovery_controller.rb
```diff
@@ -6,10 +6,164 @@ module Mobile
...
+        },
+        api_version: '1.0',
+        webviews: {
+          corona_FAQ: 'https://www.va.gov/coronavirus-veteran-frequently-asked-questions',
+          facility_locator: 'https://www.va.gov/find-locations/'
+        },
+        endpoints: {
+          appeal_details: {
+            url: '/appeal/:id',
+            verb: 'GET'
```
jjulian, line 44: Is it useful to respond with the verb, or will a mobile app already be coded to be a GET/POST/PUT?

jperk51: Not sure yet, waiting to see what Patrick thinks

jperk51: Patrick said the verbs would be useful for the frontend to have. Let me know if we should all get together to discuss that more.

modules/mobile/app/controllers/mobile/discovery_controller.rb
```diff
@@ -6,10 +6,171 @@ module Mobile
...
+      staging: 'https://staging-api.va.gov/mobile',
+      prod: 'https://api.va.gov/mobile'
+    }.freeze
+
+    def welcome
       render json: { data: { attributes: { message: 'Welcome to the mobile API' } } }
     end
+
+    def index
+      response = OpenStruct.new
```
kreek, line 145: Right now the discovery info is static but eventually will be dynamic what do you think about making this OpenStruct a model and moving all the business logic (the maps above) there?

jperk51: Sounds good to me, I'll get on that asap

modules/mobile/app/models/mobile/v0/adapters/discovery.rb
```diff
@@ -0,0 +1,176 @@
+# frozen_string_literal: true
+
+module Mobile
+  module V0
+    module Adapters
```
kreek, line 5: I think the implementation is good just not sure it's an 'adapter' 😄 appointments and claims meet that description because they are taking objects with one shape and mapping their data to another. In this case we're taking a couple args and returning new data. Also wonder if the below logic could move to a model then the model's constructor would assign the correct values to its attributes when it's instantiated. e.g.

```ruby
discovery = Discovery.new(build_number, os)
render json: DiscoverySerializer.new(discovery)
```

jperk51: Gotcha, that makes sense. I'll move that stuff

kreek (approved):
LGTM 🚀

PR #5714, opened 2021-01-19
Don't require backend review for mobile only changes

ericboehs:
According to Github's [CODEOWNERS Syntax](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax) docs, backend-review-group will be required unless only the `modules/mobile` and `spec/support/*/mobile/` have been changed:

```
# Order is important; the last matching pattern takes the most
# precedence. When someone opens a pull request that only
# modifies JS files, only @js-owner and not the global
# owner(s) will be requested for a review.
*.js    @js-owner
```


ericboehs:
Phillip pointed out:

> The problem that CODEOWNERS line is addressing is the opposite.  As the BE Tools team we need to do things like update that code for Rails 6.1 (or whatever) and not be blocked by them for review

PR #5715, opened 2021-01-19
cleaner dangerbot text

erluti:
Excluded files test is only necessary when something excluded in dangerbot warning


LindseySaari (approved):
💯

PR #5717, opened 2021-01-19
update rated disabilities for in-progress forms.

annaswims:
## Description of change
Originally, rated disabilities were only prefilled when a veteran started a form, this caused problems when substantial time passed (days/weeks/months) between when a form was initiated and when it was submitted.  EVSS will reject forms that have outdated rated_disabilites lists. 

We considered using the rated_disabilites endpoint and updating the data on the front end, but settled on updating when someone resumes a form. 
The `disability_compensation_in_progress_forms` endpoints are identical to the `in_progress_forms`, except for 
`/v0/disability_compensation_in_progress_forms/{id}` which updates the `ratedDisabilities` portion of the form, if the data has changed and adds a `ratedDisabilitiesUpdated` boolean to the metadata.

## Original issue(s)
Because we're getting rated diabilities more often we'll cache them. https://github.com/department-of-veterans-affairs/vets-api/pull/5701


app/models/form_profiles/va526ez.rb
```diff
@@ -87,6 +87,20 @@ def metadata
     }
   end
 
+  def initialize_rated_disabilities_information
```
annaswims, line 90: same method, no longer private

LindseySaari (approved):

PR #5718, opened 2021-01-19
API-4594: Improve existing upstream healthchecks and ignore async dependencies

SlakrHakr:
## Description of change
We have a need to potentially update statuspage with specific statuses based on which upstream service may be experiencing issues (Things like "degraded performance" instead of "down" if the dependency is something we interact with only in a background job). I wanted to take a crack at adding in something ive used in the past that i think works well for these types of healthchecks: okcomputer.

## Original issue(s)
https://vajira.max.gov/browse/API-4594

okcomputer gives you several endpoints for free:
* http://localhost:3000/services/claims/v1/healthcheck (same as our current healthcheck, just whether our service is up or not)
* http://localhost:3000/services/claims/v1/healthcheck/all (same as our current upstream_healthcheck, both our service health and our dependencies)
* http://localhost:3000/services/claims/v1/healthcheck/evss (any specific check, just checks that specific dependency)

Note: I left http://localhost:3000/services/claims/v1/upstream_healthcheck in and it acts exactly like http://localhost:3000/services/claims/v1/healthcheck/all, so pingdom shouldnt act any differently yet.

Note2: All urls above also have a v0 version that routes to the exact same responses.


SlakrHakr:
An example of an okcomputer response:
```
{
"default": {
"message": "Application is running",
"success": true,
"time": 0.0000021089981601107866
},
"database": {
"message": "Schema version: 20201231033748",
"success": true,
"time": 0.002682882000954123
},
"evss": {
"message": "EVSS is running",
"success": true,
"time": 0.000004264002200216055
},
"mpi": {
"message": "MPI is running",
"success": true,
"time": 0.000005855999916093424
},
"bgs": {
"message": "BGS is unavailable",
"success": false,
"time": 0.0019261490015196614
},
"vbms": {
"message": "VBMS is unavailable",
"success": false,
"time": 20.00102783800321
}
}
```

modules/claims_api/app/controllers/claims_api/upstream_healthcheck_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+    skip_before_action(:authenticate)
+
+    def index
+      checks = OkComputer::Registry.all
+      checks.run
+
+      respond checks, status_code(checks)
+    end
+  end
+end
```
SlakrHakr, line 15: this is purely for backwards compatibility with our current endpoint.

modules/claims_api/config/routes.rb
```diff
@@ -2,21 +2,21 @@
 
 ClaimsApi::Engine.routes.draw do
   match '/metadata', to: 'metadata#index', via: [:get]
-  match '/v0/healthcheck', to: 'metadata#healthcheck', via: [:get]
-  match '/v1/healthcheck', to: 'metadata#healthcheck', via: [:get]
-  match '/v0/upstream_healthcheck', to: 'metadata#upstream_healthcheck', via: [:get]
-  match '/v1/upstream_healthcheck', to: 'metadata#upstream_healthcheck', via: [:get]
+  match '/v0/upstream_healthcheck', to: 'upstream_healthcheck#index', via: [:get], defaults: { format: 'json' }
+  match '/v1/upstream_healthcheck', to: 'upstream_healthcheck#index', via: [:get], defaults: { format: 'json' }
```
SlakrHakr, line 6: https://github.com/department-of-veterans-affairs/vets-api/pull/5718/files#r560555774

modules/claims_api/app/controllers/claims_api/upstream_healthcheck_controller.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+module ClaimsApi
+  class UpstreamHealthcheckController < ::OkComputer::OkComputerController
+  end
+end
```
SlakrHakr, line 6: this is purely for backwards compatibility with our current endpoint.

modules/claims_api/config/routes.rb
```diff
@@ -2,14 +2,14 @@
...
```
SlakrHakr, line 6: https://github.com/department-of-veterans-affairs/vets-api/pull/5718/files#r560637712

hinzed1127 (approved):
Tiny question/comment about backwards compatibility, but otherwise looks good 👍

modules/claims_api/app/controllers/claims_api/upstream_healthcheck_controller.rb
SlakrHakr: the backwards compatibility is less for our outside consumers and more just because pingdom is already set up to hit this endpoint.

PR #5719, opened 2021-01-20
Api-4520 Add NOD show to docs

nathanbwright:
## Description of change
This PR adds documentation for the NOD show route.

## Original issue(s)
https://vajira.max.gov/browse/API-4520

### DIFF
--- a/modules/appeals_api/app/swagger/appeals_api/v1/notice_of_disagreements_controller_swagger.rb
+++ b/modules/appeals_api/app/swagger/appeals_api/v1/notice_of_disagreements_controller_swagger.rb
@@ -8,7 +8,6 @@ class AppealsApi::V1::NoticeOfDisagreementsControllerSwagger
   PATH_ENABLED_FOR_ENV = Settings.modules_appeals_api.documentation.notice_of_disagreements_v1
 
   NOD_TAG = ['Notice of Disagreements'].freeze
-  OBJ = :object
 
   read_file = ->(path) { File.read(AppealsApi::Engine.root.join(*path)) }
   read_json = ->(path) { JSON.parse(read_file.call(path)) }
--- a/modules/appeals_api/app/swagger/appeals_api/v1/notice_of_disagreements_controller_swagger.rb
+++ b/modules/appeals_api/app/swagger/appeals_api/v1/notice_of_disagreements_controller_swagger.rb
@@ -14,35 +14,6 @@ class AppealsApi::V1::NoticeOfDisagreementsControllerSwagger
   read_json = ->(path) { JSON.parse(read_file.call(path)) }
   read_json_from_same_dir = ->(filename) { read_json.call(['app', 'swagger', 'appeals_api', 'v1', filename]) }
 
-  response_nod_show_not_found = read_json_from_same_dir['response_nod_show_not_found.json']
-  example_all_fields_used = read_json[['spec', 'fixtures', 'valid_10182.json']]
-
-  response_nod_show_success = lambda do
-    properties = {
-      status: { '$ref': '#/components/schemas/nodStatus' },
-      updatedAt: { '$ref': '#/components/schemas/timeStamp' },
-      createdAt: { '$ref': '#/components/schemas/timeStamp' },
-      formData: { '$ref': '#/components/schemas/nodCreate' }
-    }
-    type = :noticeOfDisagreement
-    schema = {
-      type: OBJ,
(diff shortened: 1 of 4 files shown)


modules/appeals_api/app/swagger/appeals_api/v1/notice_of_disagreements_controller_swagger.rb
```diff
@@ -8,11 +8,41 @@ class AppealsApi::V1::NoticeOfDisagreementsControllerSwagger
...
   read_file = ->(path) { File.read(AppealsApi::Engine.root.join(*path)) }
   read_json = ->(path) { JSON.parse(read_file.call(path)) }
   read_json_from_same_dir = ->(filename) { read_json.call(['app', 'swagger', 'appeals_api', 'v1', filename]) }
 
+  response_nod_show_not_found = read_json_from_same_dir['response_nod_show_not_found.json']
+  example_all_fields_used = read_json[['spec', 'fixtures', 'valid_10182.json']]
+
+  response_nod_show_success = lambda do
+    properties = {
+      status: { '$ref': '#/components/schemas/nodStatus' },
```
yakschuss, line 22: These look like Open API refs to the JSON schema. Is that reference being populated using the swagger-blocks DSL?

modules/appeals_api/app/swagger/appeals_api/v1/notice_of_disagreements_controller_swagger.rb
```diff
@@ -8,11 +8,41 @@ class AppealsApi::V1::NoticeOfDisagreementsControllerSwagger
...
+  OBJ = :object
 
   read_file = ->(path) { File.read(AppealsApi::Engine.root.join(*path)) }
   read_json = ->(path) { JSON.parse(read_file.call(path)) }
   read_json_from_same_dir = ->(filename) { read_json.call(['app', 'swagger', 'appeals_api', 'v1', filename]) }
 
+  response_nod_show_not_found = read_json_from_same_dir['response_nod_show_not_found.json']
+  example_all_fields_used = read_json[['spec', 'fixtures', 'valid_10182.json']]
+
+  response_nod_show_success = lambda do
```
yakschuss, line 20: Curious about this pattern. Do we think it's better to use the lambda .call pattern rather than the Swagger::Blocks DSL for this?

nathanbwright: No good reason for this, just copying what already existed in HLR... I'll migrate it to use the DSL.

Thanks!

modules/appeals_api/app/swagger/appeals_api/v1/notice_of_disagreements_controller_swagger.rb
```diff
@@ -189,6 +190,86 @@ class AppealsApi::V1::NoticeOfDisagreementsControllerSwagger
...
+      parameter name: 'uuid', 'in': 'path', required: true, description: 'Notice of Disagreement UUID' do
+        schema { key :'$ref', :uuid }
+      end
+
+      response 200 do
+        key :description, 'Info about a single Notice of Disagreement.'
+
+        content 'application/json' do
+          schema do
+            key :type, :object
```
cilestin, line 209: Since you added the `OBJ` constant, shouldn't you be using it here?

modules/appeals_api/app/swagger/appeals_api/v1/notice_of_disagreements_controller_swagger.rb
```diff
@@ -8,6 +8,7 @@ class AppealsApi::V1::NoticeOfDisagreementsControllerSwagger
   PATH_ENABLED_FOR_ENV = Settings.modules_appeals_api.documentation.notice_of_disagreements_v1
 
   NOD_TAG = ['Notice of Disagreements'].freeze
+  OBJ = :object
```
cilestin, line 11: Actually I guess the better question is: why was this constant added? I'm not seeing it used in the file or any other part of the changes. Is there something I'm missing about the changes that requires this constant be present?

nathanbwright: Whoops, I pulled this in originally because it was in the HLR one and didn't double check if I actually needed it. Good catch!

cilestin (approved):
LGTM!

PR #5721, opened 2021-01-20
log when users have a mvi_response.status of OK then return a va profile MVI201 error

lihanli:
## Description of change
log when users have a mvi_response.status of OK then return a va profile MVI201 error in order to report the user details to va profile

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/18349

tested with unit tests


lib/vet360/contact_information/transaction_response.rb
```diff
@@ -66,8 +66,42 @@ def self.log_error
...
+
+      def self.log_mpi_error
+        if error?
+          @response_body['tx_messages'].each do |tx_message|
+            if tx_message['code'] == NOT_FOUND_IN_MPI_CODE
+              return log_message_to_sentry(
+                'va profile mpi not found',
+                :error,
+                {
+                  icn: @user.icn,
```
LindseySaari, line 90: I think the icn and edipi both PII that shouldn't be logged to Sentry

lihanli: someone looked it up before and it seems like they are not considered PII
https://github.com/department-of-veterans-affairs/vets-api/pull/2888#issuecomment-474173431

LindseySaari: Ok, thanks for the link. That was helpful 💯

LindseySaari (approved):

PR #5722, opened 2021-01-20
Add search_gov click tracking endpoint

srsuddath:
## Description of change
This change adds an endpoint to access search.gov's click tracking api

## Original issue(s)
department-of-veterans-affairs/va.gov-team#16367

This POST request is unique in that it sends all information as query params rather than as part of the body

### DIFF
--- a/app/controllers/v0/search_click_tracking_controller.rb
+++ b/app/controllers/v0/search_click_tracking_controller.rb
@@ -1,0 +1,75 @@
+# frozen_string_literal: true
+
+require 'search_click_tracking/service'
+
+module V0
+  class SearchClickTrackingController < ApplicationController
+    include ActionView::Helpers::SanitizeHelper
+
+    skip_before_action :authenticate
+    skip_before_action :verify_authenticity_token
+
+    # Sends click tracking data to search.gov analytics, based on the passed url,
+    # query, position, client_ip, and user_agent.
+    #
+    def create
+      response = SearchClickTracking::Service.new(url, query, position, user_agent, client_ip).track_click
+      if response.success?
+        render nothing: true, status: 204
+      else
+        render json: response.body, status: 400
+      end
+    end
+
+    private
+
+    def click_params
+      params.permit(:url, :query, :position, :client_ip, :user_agent)
+    end
+
+    # Returns a sanitized, permitted version of the passed url params.
+    #
+    # @return [String]
+    # @see https://api.rubyonrails.org/v4.2/classes/ActionView/Helpers/SanitizeHelper.html#method-i-sanitize
+    #
+    def url
+      sanitize click_params['url']
+    end
+
+    # Returns a sanitized, permitted version of the passed query params.
+    #
+    # @return [String]
+    # @see https://api.rubyonrails.org/v4.2/classes/ActionView/Helpers/SanitizeHelper.html#method-i-sanitize
+    #
+    def query
+      sanitize click_params['query']
+    end
+
+    # Returns a sanitized, permitted version of the passed position params.
+    #
+    # @return [String]
+    # @see https://api.rubyonrails.org/v4.2/classes/ActionView/Helpers/SanitizeHelper.html#method-i-sanitize
+    #
+    def position
+      sanitize click_params['position']
+    end
+
+    # Returns a sanitized, permitted version of the passed client_ip params.
+    #
+    # @return [String]
+    # @see https://api.rubyonrails.org/v4.2/classes/ActionView/Helpers/SanitizeHelper.html#method-i-sanitize
+    #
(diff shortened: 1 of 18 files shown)


LindseySaari:
Looks like there's some failing specs

app/controllers/v0/search_click_tracking_controller.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
+
+require 'search_click_tracking/service'
```
LindseySaari, line 3: It looks like this may still be a WIP PR. Do you have changes that include request specs?

config/settings/test.yml
```diff
@@ -26,18 +26,15 @@ saml_ssoe:
   cert_path: spec/support/certificates/ruby-saml.crt
   key_path: spec/support/certificates/ruby-saml.key
 
-
 edu:
   prefill: true
   sftp:
-    relative_path: 'spool_files'
-
+    relative_path: "spool_files"
```
LindseySaari, line 32: Looks like this may have been added by mistake. rubocop prefers single quotes

config/settings/test.yml
```diff
@@ -26,18 +26,15 @@ saml_ssoe:
...
 
 pension_burial:
   prefill: true
   sftp:
-    relative_path: 'VETSGOV_PENSION'
-
+    relative_path: "VETSGOV_PENSION"
```
LindseySaari, line 37: Here as well ☝️

lib/search_click_tracking/service.rb
```diff
@@ -0,0 +1,86 @@
...
+      @query = query
+      @position = position
+      @client_ip = client_ip
+      @user_agent = user_agent
+    end
+
+    # POSTs click tracking query param data to search.gov
+    #
+    def track_click
+      Faraday.post(url_with_params, '')
```
LindseySaari, line 40: It doesn't look like you're utilizing the monitoring concern/with_monitoring here to increment the statsd metric

LindseySaari, line 40: Can you use the post method defined in the common/client/base class?

app/controllers/v0/search_click_tracking_controller.rb
srsuddath: Thanks for pointing this out, I am new to working with the BE and missed these. I am creating these specs now. Thanks!

config/settings/test.yml
srsuddath: Oh thanks, I will fix this now!

lib/search_click_tracking/service.rb
srsuddath: This is a weird 3rd party request that returns an empty body or a string. Using the perform / post method defined in the base class I was getting some unwanted errors as it tried to parse the body. @dillo helped me come up with this solution and may be able to explain it better.

LindseySaari:
It looks like there's a security issue that's been brought up associated with the access key being publicly exposed [here].(https://github.com/department-of-veterans-affairs/vets-api/blob/b07bf30efdb1a82fac4099f5440d0e80223b8794/spec/support/vcr_cassettes/search_click_tracking/success.yml#L5) The key will need to be rebuilt.

srsuddath:
The access key is a public key, but I have removed all references to it and have squashed and re-written the commit history so it will not appear. This should fix the issue.

lib/search_click_tracking/service.rb
LindseySaari: Ok strange, so passing in the query params [here](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/common/client/base.rb#L94) as the third parameter to the method wasn't returning the expected results?

srsuddath: Correct. Passing in the params here allowed us to send the request properly, but somewhere in the way that was handling the response was causing it to fail when we got an empty response back. I can explore this further.

kelsonic (approved):
LGTM! 🎉

lib/search_click_tracking/response.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require 'common/client/concerns/service_status'
+require 'common/models/base'
+
+module SearchClickTracking
```
LindseySaari, line 6: Can this response class be removed? Doesn't look like it's being used currently.

srsuddath: Possibly! I will get it a test.

LindseySaari (approved):

lib/search_click_tracking/response.rb
srsuddath: Removed it!

PR #5723, opened 2021-01-20
Api 4527/hlr constructor refactor

yakschuss:
## Description of change

This PR remakes the HLR PDF constructor in the same style as the previous refactor of NOD [4306](https://vajira.max.gov/browse/API-4306)

## Original issue(s)
[4527](https://vajira.max.gov/browse/API-4527)


modules/appeals_api/app/workers/appeals_api/higher_level_review_pdf_submit_job.rb
```diff
@@ -12,23 +11,15 @@ class HigherLevelReviewPdfSubmitJob
     include Sidekiq::Worker
     include CentralMail::Utilities
 
-    def perform(higher_level_review_id, retries = 0)
+    def perform(higher_level_review, retries = 0)
```
cilestin, line 14: Nice! So much cleaner!

cilestin (changes requested):
Lookin' good! 😎  Just need to retract the formatting changes to the output PDF when an extra issues page is present.

cilestin (approved):
LGTM!

modules/appeals_api/app/workers/appeals_api/higher_level_review_pdf_submit_job.rb
ElleeB: Did we change this to take in an object instead of the id?

yakschuss: Yep, removed the ID in favor of the reference. We were doing multiple finds in the job which didn't seem necessary.

PR #5727, opened 2021-01-20
Mark.viterna/api 4177 benefits api documentation

mviterna:
## Description of change
API 4177 - There were some discrepancies and missing items in the intake API. There is still an issue with the pdf attributes showing up in the initial location step (should be  "uploaded_pdf": null) but cannot get swagger blocks to override it).
## Original issue(s)
https://vajira.max.gov/browse/API-4177

Locally verified that changes are showing in portal. No linter errors seen


gpbowman-git:
Looks good. As we discussed, let's look to change the location response to reflect the 'pending' status.

gpbowman-git (approved):
looks good

PR #5734, opened 2021-01-21
API-4589: Record not saving in sandbox after claim establishment through EVSS

SlakrHakr:
## Description of change
After working through this, im not 100% sure why this is working correctly in production now, but it definitely is somehow.
When i run through the internals of the ClaimEstablisher job in sandbox it fails to save because of the `uniqueness` validation on the record's md5. That md5 will always be the same at that point in the logic, because it is generated from the record's `auth_headers` and `form_data` which are unchanged at that point.

Even weirder is that calling `save` once more, after the initial save fails, succeeds (all in sandbox i should reiterate). I'm not sure why this is the case...

Hopefully this change makes overall sense for this logic (and also seems to resolve sandbox as a nice side-effect). It essentially only validates the uniqueness of the md5 on create, not update. 

## Original issue(s)
https://vajira.max.gov/browse/API-4589

Tested manually in sandbox environment.
Existing spec tests still pass.

### DIFF
--- a/modules/claims_api/app/workers/claims_api/claim_establisher.rb
+++ b/modules/claims_api/app/workers/claims_api/claim_establisher.rb
@@ -20,7 +20,7 @@ module ClaimsApi
       response = service(auth_headers).submit_form526(form_data)
       auto_claim.evss_id = response.claim_id
       auto_claim.status = ClaimsApi::AutoEstablishedClaim::ESTABLISHED
-      auto_claim.save
+      auto_claim.save!
 
       queue_flash_updater(auth_headers, auto_claim.flashes, auto_claim_id)
       queue_special_issues_updater(auth_headers, auto_claim.special_issues, auto_claim)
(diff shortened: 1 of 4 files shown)


modules/claims_api/app/workers/claims_api/claim_establisher.rb
```diff
@@ -20,7 +20,7 @@ def perform(auto_claim_id)
       response = service(auth_headers).submit_form526(form_data)
       auto_claim.evss_id = response.claim_id
       auto_claim.status = ClaimsApi::AutoEstablishedClaim::ESTABLISHED
-      auto_claim.save
+      auto_claim.save!
```
SlakrHakr, line 23: I changed this so at least the background job would fail out at this point. Currently, in sandbox, this just returns false and moves on.

bastosmichael (changes requested):
Can you include a test to test for and ensure failure?

SlakrHakr:
> Can you include a test to test for and ensure failure?

Done here: https://github.com/department-of-veterans-affairs/vets-api/pull/5734/commits/d35db39ab108ccfec75b0b67be2c612538d74104

hinzed1127 (approved):

bastosmichael (approved):

PR #5742, opened 2021-01-21
moves simplecov helper add_filter and add_group lines to separate methods

bramleyjl:
## Description of change

Moves add_filter and add_group lines to their own methods due to their combined size reaching the assignment branch condition limit of 40 lines. This is being done in preparation for the addition of a new module which would push the ABC over 40 lines.

All unit tests continue to pass with this change.

### DIFF
--- a/spec/simplecov_helper.rb
+++ b/spec/simplecov_helper.rb
@@ -13,39 +13,8 @@ class SimpleCovHelper
 
       track_files '**/{app,lib}/**/*.rb'
 
-      add_filter 'app/controllers/concerns/accountable.rb'
-      add_filter 'config/initializers/clamscan.rb'
-      add_filter 'lib/config_helper.rb'
-      add_filter 'lib/feature_flipper.rb'
-      add_filter 'lib/gibft/configuration.rb'
-      add_filter 'lib/ihub/appointments/response.rb'
-      add_filter 'lib/salesforce/configuration.rb'
-      add_filter 'lib/vet360/address_validation/configuration.rb'
-      add_filter 'lib/search/response.rb'
-      add_filter 'lib/vet360/exceptions/builder.rb'
-      add_filter 'lib/vet360/response.rb'
-      add_filter 'modules/claims_api/app/controllers/claims_api/v0/forms/disability_compensation_controller.rb'
-      add_filter 'modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb'
-      add_filter 'modules/claims_api/app/swagger/*'
-      add_filter 'modules/claims_api/lib/claims_api/health_checker.rb'
-      add_filter 'lib/bip_claims/configuration.rb'
-      add_filter 'version.rb'
-
-      # Modules
-      add_group 'Policies', 'app/policies'
-      add_group 'Serializers', 'app/serializers'
-      add_group 'Services', 'app/services'
-      add_group 'Swagger', 'app/swagger'
-      add_group 'Uploaders', 'app/uploaders'
-      add_group 'AppealsApi', 'modules/appeals_api/'
-      add_group 'ClaimsApi', 'modules/claims_api/'
-      add_group 'CovidVaccine', 'modules/covid_vaccine/'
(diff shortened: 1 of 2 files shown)


spec/simplecov_helper.rb
```diff
@@ -70,6 +39,43 @@ def merge_results
...
+    add_filter 'modules/claims_api/app/controllers/claims_api/v0/forms/disability_compensation_controller.rb'
+    add_filter 'modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb'
+    add_filter 'modules/claims_api/app/swagger/*'
+    add_filter 'modules/claims_api/lib/claims_api/health_checker.rb'
+    add_filter 'lib/bip_claims/configuration.rb'
+    add_filter 'version.rb'
+  end
+
+  def add_modules
+    add_group 'Policies', 'app/policies'
```
LindseySaari, line 64: This change will break the way that the module generator inserts entries into the simplecov helper when a new module is created. https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/generators/module/module_generator.rb#L65-L66

LindseySaari: It uses the # Modules comment to insert a new entry

LindseySaari: If you could kindly keep the comment OR update to after: "def add_modules\n"

jimiray: Another approach would be to use a data structure, Array in the case of filters and just iterate over them.  This eliminates the complexity of the method while also stating that the list of "filters" are a data object. Something like:

```ruby

# Not All Filters Added For Brevity.
FILTERS = [
'config/initializers/clamscan.rb'
  'lib/config_helper.rb'
  'lib/feature_flipper.rb'
  'lib/gibft/configuration.rb'
  'lib/ihub/appointments/response.rb'
  'lib/salesforce/configuration.rb'
  'lib/vet360/address_validation/configuration.rb'
  'version.rb'
]

def add_filters 
  FILTERS.each { |f| add_filter f }
end
```

Something similar could be done with groups, I'd probably do use some magic though.

```ruby
GROUPS = {
  serializers: 'app/serializers',
  services:    'app/services'
}

def add_groups 
  GROUPS.each do { |k, v| add_group(k.classify, v) }
end
```

jimiray: I would defer to @LindseySaari on what all it would break though.

bosawt:
Can you try parallel specs with this and just make sure it doesn't break them (it really shouldn't, just making sure):
`bundle exec rails parallel:setup`
`RAILS_ENV=test bundle exec rails parallel:spec`
If you don't add the `NOCOVERAGE` flag, it will run with coverage. It should still technically fail since we don't hit the coverage threshold, but it should run and spit out a report.

spec/simplecov_helper.rb
LindseySaari: Ooo that's clean, I like it @jimiray! Unfortunately, it would break the generator logic and the cleanup may be better suited for a separate ticket/PR associated with the generator.

LindseySaari (commented):
This looks good, but i'll wait to approve until the parallel specs question is addressed

bramleyjl:
Parallel specs works with one exception:

`rspec ./spec/lib/common/convert_to_pdf_spec.rb:25 # Common::ConvertToPdf#run with an image converts an image to pdf format`

I pulled the latest down on master branch and it's failing the same test there so I'm pretty confident it's unrelated to this change.

bosawt:
> Parallel specs works with one exception:
> 
> `rspec ./spec/lib/common/convert_to_pdf_spec.rb:25 # Common::ConvertToPdf#run with an image converts an image to pdf format`
> 
> I pulled the latest down on master branch and it's failing the same test there so I'm pretty confident it's unrelated to this change.

Sounds good to me, I'll ask you about that failure separately (wanna see if it's a flaky spec in parallel), but agree it's prob not due to these changes

LindseySaari (approved):

PR #5744, opened 2021-01-21
API-4615 Add environment to report email

cilestin:
## Description of change
Adds environment name to the reporting email in both the email subject & body.

## Original issue(s)
https://vajira.max.gov/browse/API-4615


nathanbwright (approved):
👍

PR #5748, opened 2021-01-22
18740 consolidate va profile references inside user class

bosawt:
## Description of change
This PR is a first step towards refactoring the va_profile object in the User class. It probably works fine, but we probably want to split this PR up.

## Original issue(s)
department-of-veterans-affairs/va.gov-team/issues/18740

There should be no change on prod. But we will want to verify that

- Specs run
- Will need more involved dev acceptance


app/controllers/v0/post911_gi_bill_statuses_controller.rb
```diff
@@ -78,14 +78,14 @@ def user_json(user)
...
         vet360_id: user.vet360_id,
         ssn: user.ssn,
-        birth_date: iso8601_birth_date(user)
+        birth_date: iso8601_birth_date(user.birth_date)
       }.to_json
     end
 
-    def iso8601_birth_date(user)
-      return nil unless user&.va_profile&.birth_date
+    def iso8601_birth_date(birth_date)
```
jimiray, line 85: Should we not just assume that the birth_date should be in ISO8601 format? Seems like to me that formatting is a presentation thing and we should send and expect to get ISO8601.

app/controllers/v0/sessions_controller.rb
```diff
@@ -256,7 +256,7 @@ def log_persisted_session_and_warnings
       # We want to log when SSNs do not match between MVI and SAML Identity. And might take future
       # action if this appears to be happening frequently.
       if current_user.ssn_mismatch?
-        additional_context = StringHelpers.heuristics(current_user.identity.ssn, current_user.va_profile.ssn)
+        additional_context = StringHelpers.heuristics(current_user.identity.ssn, current_user.mpi.profile.ssn)
```
jimiray, line 259: I feel like there are too many steps to get to SSN, could we do something like current_user.mpi_profile.ssn?

app/models/form_profile.rb
```diff
@@ -286,15 +286,15 @@ def vet360_mailing_address
     vet360_contact_info&.mailing_address
   end
 
-  def va_profile_address_hash
-    user.va_profile&.address &&
+  def user_address_hash
```
jimiray, line 289: I'm also not sure about calling out that a method is a "hash", should it just be address and return the Hash? A case could probably be made that there should be an Address PRO that takes the hash and returns an object that can also have it's own serializer? I might be taking a purist POV here.

app/models/gi_bill_feedback.rb
```diff
@@ -28,11 +28,10 @@ def get_user_details
     profile_data = {}
 
     if user.present?
```
jimiray, line 30: The assignment to profile_data seems unecessary when looking at this deeper. And I'm a big fan of the pattern of just returning before the if condition. Something like this:

```
def get_user_details 
  return unless user.present?
  { profile_data: { 'active_ICN'     => user.icn,
                            'historical_ICN' => user.historical_icns,
                           'sec_ID'             => user.sec_id } }
end
```

app/models/id_card_attributes.rb
```diff
@@ -16,12 +16,12 @@ def traits
       'edipi' => @user.edipi,
```
jimiray, line 16: do we know why these are using strings v. symbols?

lib/evss/disability_compensation_auth_headers.rb
```diff
@@ -30,14 +30,14 @@ def eauth_json
     end
 
     def gender
-      case @user.gender || @user.mpi&.profile&.gender
+      case @user.gender || @user.gender
```
jimiray, line 33: These seems like an opportunity to standardize what "gender" is to us? Is it the word or just the letter? I don't think we should be using gender here this way though.

modules/health_quest/app/services/health_quest/jwt_wrapper.rb
```diff
@@ -66,15 +66,15 @@ def gender
     end
 
     def dob
-      user.mpi&.profile&.birth_date
+      user.birth_date
```
jimiray, line 69: If needed we should alias user.dob to user.birth_date and try to document what a "user" is so we can update everyone to be using the same attributes.

modules/vaos/app/models/vaos/appointment_request_form.rb
```diff
@@ -133,25 +133,25 @@ def patient_identifier
     end
 
     def first_name
-      @user.mpi&.profile&.given_names&.first
+      @user.given_names&.first
```
jimiray, line 136: Can we not just delegate these to user? I'm not sure how that works across modules / components.

app/controllers/v0/post911_gi_bill_statuses_controller.rb
bosawt: This is a good idea. I'm gonna make a new Zenhub issue focused only on birth_date, and split this PR off to that. I'll put this recommendation in the issue description. I think we have a couple places that want birth_date in another format, but it should be up to them to massage it out of ISO8601 into whatever form they want

app/controllers/v0/sessions_controller.rb
bosawt: Yeah, I actually don't like this that much anyway. I might remove `additional_context` altogether, and just show this log with the `user_uuid` is there's an SSN mismatch

app/models/form_profile.rb
bosawt: Ah yeah, that's a good recommendation. I prob will keep the `hash` designation here for now, but mostly just because that's how a lot of other functions that do similar things in this file are named. This whole file is messy and could use a refactor in general though

app/models/gi_bill_feedback.rb
bosawt: Ah yeah, that could work, although it would be just a bit different, since it's not actually returning nil is the user isn't present:
`{ 'profile_data' => {} } unless user.present?`

bosawt: I guess the original thinking was probably that the last line would show that the function returns a hash with `profile_data` key regardless of the presence of `User`

app/models/id_card_attributes.rb
bosawt: Dunno, looks to me like it's because these are real close to the controller, so it's basically a serializer at this point. It looks like `id_card_attributes_controller` asks for these, barely massages them, and then renders it.

We use strings in SO many places where we could use symbols. Worth an investigation in our general code base for sure

app/models/mhv_account.rb
```diff
@@ -200,15 +200,15 @@ def multiple_active_mhv_ids?
     if previously_upgraded? || previously_registered?
```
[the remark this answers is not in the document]
bosawt: 👍  Sounds good, I'll change it

app/models/saved_claim/education_career_counseling_claim.rb
```diff
@@ -53,10 +53,6 @@ def send_to_central_mail!
   private
 
   def claimant_birth_date(current_user)
```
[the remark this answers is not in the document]
bosawt: 👍

app/models/saved_claim/veteran_readiness_employment_claim.rb
```diff
@@ -25,7 +25,7 @@ def add_claimant_info(user)
       'pid' => user.participant_id,
       'edipi' => user.edipi,
       'vet360ID' => user.vet360_id,
-      'dob' => parsed_date(user.birth_date)
+      'dob' => parsed_birth_date(user)
```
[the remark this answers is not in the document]
bosawt: 👍

app/models/user.rb
```diff
@@ -78,27 +106,47 @@ def ssn_normalized
...
+  def family_name
+    mpi.profile&.family_name
   end
 
   def gender
-    identity.gender || (mhv_icn.present? ? mpi&.profile&.gender : nil)
+    identity.gender || (mhv_icn.present? ? va_profile&.gender : nil)
   end
 
   def birth_date
```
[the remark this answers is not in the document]
bosawt: Yeah, it is just nil for the else. This way is cleaner regardless

app/models/user.rb
```diff
@@ -61,7 +61,35 @@ def pciu_alternate_phone
...
   def first_name
-    identity.first_name || (mhv_icn.present? ? mpi&.profile&.given_names&.first : nil)
+    identity.first_name || (mhv_icn.present? ? va_profile&.given_names&.first : nil)
+  end
+
+  def given_names
+    mpi.profile&.given_names
+  end
+
+  def suffix
```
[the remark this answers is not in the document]
bosawt: This is a huge bugbear of mine, I agree with removing the safe operators in User. In fact, I put a story in the backlog to address this issue. I don't think this specific PR should cover it, but our data consistency would definitely improve if we dealt with this data inconsistency in a sensible manner. I think we should have a more general conversation about it anyway, because we should probably figure out standards of logging when there is missing or inconsistent data, instead of the equivalent of `rescue nil` all over the place

lib/evss/disability_compensation_auth_headers.rb
bosawt: Haha, whoops!

modules/vaos/app/models/vaos/appointment_request_form.rb
bosawt: I only kept `given_names.first` to attempt to keep behavior consistent (if we request `first_name` from the user, it will first try to grab the name off of the `UserIdentity` object, and if that doesn't exist, grab it off the MPI object in a similar way to how it does it here). However, since I've decided to split this PR, I think I have more stomach for the small changes in behavior with letting `User` figure out how to retrieve the field. I'll change this to `user.first_name` and perhaps in a specific PR figure out if this changes anything

PR #5749, opened 2021-01-22
expectations for generator test output

erluti:
## Description of change
In an effort to remove some stdout output during test runs, I came across this way to test output to stdout and applied it to these generator tests.

### DIFF
--- a/spec/lib/generators/provider_state/provider_state_generator_spec.rb
+++ b/spec/lib/generators/provider_state/provider_state_generator_spec.rb
@@ -5,14 +5,10 @@ require 'generators/provider_state/provider_state_generator'
 
 describe ProviderStateGenerator do
   describe 'create_provider_state_file' do
-    context 'once generated' do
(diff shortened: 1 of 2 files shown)


spec/lib/generators/provider_state/provider_state_generator_spec.rb
```diff
@@ -5,14 +5,16 @@
 
 describe ProviderStateGenerator do
   describe 'create_provider_state_file' do
-    context 'once generated' do
-      before(:all) { ProviderStateGenerator.new(['test']).create_provider_state_file }
-
-      after(:all) { FileUtils.rm_rf('spec/service_consumers/provider_states_for/test.rb') }
+    after do
```
erluti, line 8: this is an after block to be sure file is cleaned up when an expectation fails

LindseySaari (approved):
Ooo nice! I like this!

PR #5757, opened 2021-01-22
API-4615 Use Settings.vsp_environment for appeals email report

cilestin:
## Description of change
Use `vsp_environment` as the environment name when sending appeal report email.

## Original issue(s)
https://vajira.max.gov/browse/API-4615


nathanbwright (approved):
🚀

PR #5758, opened 2021-01-22
API-4400 Remove participant ID from responses for Claims Endpoints

hinzed1127:
Removes Participant ID from any public facing responses. We don't currently require it, and all the lookups for it are internal, so there's no need to expose it at this time

## Original issue(s)
https://vajira.max.gov/browse/API-4400

### DIFF
--- a/modules/claims_api/app/serializers/claims_api/power_of_attorney_serializer.rb
+++ b/modules/claims_api/app/serializers/claims_api/power_of_attorney_serializer.rb
@@ -2,6 +2,6 @@
 
 module ClaimsApi
   class PowerOfAttorneySerializer < ActiveModel::Serializer
-    attributes :status, :date_request_accepted, :representative, :veteran, :previous_poa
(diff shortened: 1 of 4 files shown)


SlakrHakr:
> we might need to split out the POA stuff into a v0/v1 division where v0 still keeps participant ID if we want that.

all the acceptance criteria on this issue ticket references v1 specifically, so im guessing we want to leave v0 alone for now?

modules/claims_api/app/serializers/claims_api/power_of_attorney_serializer.rb
```diff
@@ -2,6 +2,6 @@
 
 module ClaimsApi
   class PowerOfAttorneySerializer < ActiveModel::Serializer
-    attributes :status, :date_request_accepted, :representative, :veteran, :previous_poa
```
hinzed1127, line 5: removing `veteran since the only thing being provided appears to be the participant ID

SlakrHakr (approved):

PR #5754, opened 2021-01-22
hook up HLR constructor to new generator

yakschuss:
## Description of change
This takes the refactor from [5723](https://github.com/department-of-veterans-affairs/vets-api/pull/5723) and hooks it up to the existing functionality. It was split up because of dangerbot LOC issues.

## Original issue(s)
[4527](https://vajira.max.gov/browse/API-4527)

❗ this PR is blocked by #5723 ❗  DO NOT MERGE until #5723 is merged.

PR #5730, opened 2021-01-20
Refactor Dangerfile

ericboehs:

PR #5759, opened 2021-01-23
Add missing method - fixes failing specs

cilestin:
## Description of change
Adds missing `final_page_adjustments` method to the HLR PDF Structure

## Analysis
PRs #5703 and #5754 were created separately and each had passed CI and review.

#5730 added the `final_page_adjustment` method but #5754 was not aware of the need to include it in its own code. Had tests been re-run between merging either PR, the issue would have been caught. Since both were already green, however, they were merged in and caused the issue after the fact.

Luckily the timing worked in our favor and #5754 was merged after the final deploy time on Fri, Jan 21 so won't cause issues on production over the weekend.


yakschuss (approved):
Good catch!

PR #5760, opened 2021-01-23
API-4568: Don't replace GUID with EVSS Claim ID

SlakrHakr:
## Description of change
Always return our database uuid for the id value in every claim response.
The only time the evss_id will be used is if the user provides an evss_id (in the case where the particular claim may not have been submitted through our api).

## Original issue(s)
https://vajira.max.gov/browse/API-4568

Tested locally.
Updated and add specs.

### DIFF
--- a/modules/claims_api/app/controllers/claims_api/application_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/application_controller.rb
@@ -46,10 +46,13 @@ module ClaimsApi
 
       if claim && claim.status == 'errored'
         fetch_errored(claim)
-      elsif claim && claim.evss_id.nil?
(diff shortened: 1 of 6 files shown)


hinzed1127:
Small observation: `AutoEstablishedClaim` model has a `get_by_id_or_evss_id` method that appears to only be used internally. If I'm reading it properly, has similar kind of slippery logic where someone provides an id, and then we infer whether it's an EVSS id or not by checking for a hyphen in the id, which isn't super explicit at first glance without look at method.

Idk if being similarly specific in that code is useful or worthwhile, but just something that came to mind. Otherwise looks good.

hinzed1127 (approved):

PR #5762, opened 2021-01-25
Add 'lob' to CentralMail upload

yakschuss:
[API-3817](https://vajira.max.gov/browse/API-3817)

This ticket sends the NOD line of business with parameter 'lob' to CentralMail uploading, using the NOD value of 'BVA', in order to route the 10182 correctly.


nathanbwright:
Maybe we should add a spec to verify that the `lob` is actually being sent to CentralMail?

nathanbwright (approved):
Looks good to me!

PR #5765, opened 2021-01-25
Output CSV for Form 526 Rake Task

edmangimelli:
Tweaks the `form526:submissions` rake task to include an easy way to get BDD stats.
Instead of supplying a date range, simply make the first arg `bdd`, and
1) submissions will be restricted to BDD forms
2) output is printed in CSV form
3) participant IDs are partially redacted
4) daily submission counts are printed (for easy graphing)

Larger-than-I-would-have-liked refactor. Also tested in console.


rakelib/form526.rake
```diff
@@ -4,40 +4,159 @@ require 'pp'
...
-  task :submissions, %i[start_date end_date] => [:environment] do |_, args|
+  desc <<~HEREDOC
+    Get all submissions within a date period:
+      rake form526:submissions[2021-01-23,2021-01-24]
+    Last 30 days:
+      rake form526:submissions[]
+    BDD stats mode:
+      rake form526:submissions[bdd]
+  HEREDOC
+  task :submissions, %i[first second] => [:environment] do |_, args|
```
LindseySaari, line 15: Would it make more sense to name the arguments something like first_date, second_date just to make it clear that the arguments should be dates?

LindseySaari (approved):
Looks good, just one comment about the args

rakelib/form526.rake
edmangimelli: I could! I was kinda torn though. There is one scenario where the first arg isn't a date (it's just the string "BDD").

annaswims (approved):
LGTM

PR #5776, opened 2021-01-26
API-4076 - Change max length of nod issue

nathanbwright:
## Description of change
This PR updates our NOD contestable issues `issue` field to be 255 chars (in both the docs and the schema). This is based on Caseflow's Contention class that truncates at 255 bytes (https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/models/contention.rb).

## Original issue(s)
https://vajira.max.gov/browse/API-4076


yakschuss (approved):
👍

cilestin (approved):
:shipit:

PR #5777, opened 2021-01-26
restores tud_account creation service and tests #17214

bramleyjl:
## Description of change

Restores tud_account creation service and unit tests from larger initial commit.

## Original issue(s)
https://app.zenhub.com/workspaces/vsp-identity-5f5bab705a94c9001ba33734/issues/department-of-veterans-affairs/va.gov-team/17214

Basic unit tests added to cover new methods, manual testing was performed as well


modules/test_user_dashboard/db/migrate/20210121161138_add_tud_accounts.rb
```diff
@@ -10,4 +10,4 @@ def change
       t.timestamps
     end
   end
-end
+end
```
LindseySaari, line 13: Looks like your editor may have added this

modules/test_user_dashboard/app/services/test_user_dashboard/create_test_user_account.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module TestUserDashboard
+  class CreateTestUserAccount
+    attr_accessor :row, :test_user_account
+
+    def initialize(row = {})
+      account_hash = row.to_h
```
jimiray, line 8: I think the row is a Hash in this and doesn't need to be cast to it using to_h. I've also seen recently that it seems to be more common in our codebase to call ```to_hash``` rather than ```to_h``` which is how I've always used it.

I may have suggested this earlier on but I don't think you need to declare the TudAccount as being from the module TestUserDashboard since we have already put the class in it.

```
@test_user_account = TudAccount.new(row)
```

modules/test_user_dashboard/spec/requests/tud_accounts_request_spec.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Test User Dashboard', type: :request do
+  context 'index' do
+    it 'returns a json formatted respone' do
```
jimiray, line 7: Again this might be a me thing:

```ruby
context 'with json' do 
  before { get '/test_user_dashboard/tud_accounts') }
  
  subject :response

  it {  expect(response.status).to eq 200 }
  it { expect(response.content_type).to eq 'application/json; charset=utf-8' }
end
```

I'm pretty particular about only 1 expectation per it block. 

I tend to follow any recommendation from Better Specs:

https://www.betterspecs.org/#single

modules/test_user_dashboard/spec/services/create_test_user_account_spec.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe TestUserDashboard::CreateTestUserAccount do
+  subject { described_class.new }
+
+  file_path = Rails.root.join('modules', 'test_user_dashboard', 'spec', 'support', 'spec_users.csv')
```
jimiray, line 8: @bramleyjl We should convert line 8 and 9 to be:

```ruby
let(:file_path) {  File.join(Rails.root, 'modules/test_user_dashboard/spec/support/spec_users.csv') }
let(:users).     { CSV.read(file_path, headers: true) } 

subject { described_class.new }
```

I typically don't test initializers but that's mostly because I don't believe anything should really happen in them but assignment.

modules/test_user_dashboard/spec/services/create_test_user_account_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  spec_users = CSV.read(file_path, headers: true)
+
+  describe '#initialize' do
+    it 'has a test_user_account attribute' do
```
jimiray, line 12: I'm not sure the text is doing much for us here, the expectation is self explanatory. How about:

```ruby
describe 'initialize' do 
  it { expect(subject.test_user_account).to be_a(TestUserDashboard::TudAccount) } 
end
```

jimiray: Also, if it is a TudAccount then the checking to make sure it's assigned is a duplicate test. If the test at like 17 is true the test at like 13 will also be true always.

modules/test_user_dashboard/app/controllers/test_user_dashboard/tud_accounts_controller.rb
```diff
@@ -4,5 +4,11 @@
 
 module TestUserDashboard
   class TudAccountsController < ApplicationController
+    def index
+      tud_accounts = TudAccount.all
+      render(
```
jimiray: For consistency for our code:

```
render json: tud_accounts
```

modules/test_user_dashboard/spec/requests/tud_accounts_request_spec.rb
LindseySaari: Agreed Jim! I think it's a personal preference, but one per block definitely reads better

modules/test_user_dashboard/app/services/test_user_dashboard/create_test_user_account.rb
bramleyjl: On my local build the row seems to always be an Array and causes an error when trying to use it to create a new TudAccount without casting it to a hash first.

LindseySaari: If the row is an array, you may want to call this out in a comment in the code or adjust the argument (just a thought).

LindseySaari (approved):

PR #5778, opened 2021-01-26
18830 dangerbot checks for enterprise gemfile change

erluti:
## Description of change
Prevent developers from accidentally removing sidekiq enterprise.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#18830

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -8,7 +8,18 @@ module VSPDanger
       [
         ChangeLimiter.new.run,
         MigrationIsolator.new.run
-      ]
+      ].sort(&:severity_sort)
+    end
+
+    # expects severities of :error > :warning > :info
+    def severity_sort(result_a, result_b)
+      severity_a = result_a[:severity]
+      severity_b = result_b[:severity]
+      return 0 if severity_a == severity_b
+      return -1 if severity_a == :error
+      return 1 if severity_b == :error
+      return -1 if severity_a == :warning
+      return 1 if severity_b == :warning
     end
 
     def self.prepare_git
--- a/Dangerfile
+++ b/Dangerfile
@@ -210,7 +210,7 @@ module VSPDanger
     end
 
     def gemfile_diff
-      `git diff #{BASE_SHA}:Gemfile.lock #{HEAD_SHA}:Gemfile.lock`
+      `git diff #{BASE_SHA}...#{HEAD_SHA} -- Gemfile.lock`
     end
   end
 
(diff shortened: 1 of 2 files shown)


Dangerfile
```diff
@@ -192,6 +204,48 @@ module VSPDanger
...
+      `git diff #{base_sha}:Gemfile.lock #{head_sha}:Gemfile.lock`
+    end
+
+    def head_sha
+      `git rev-parse --abbrev-ref HEAD`.chomp
+    end
+
+    def base_sha
+      'origin/master'
+    end
```
ericboehs, line 246: We're reusing these a lot (head_sha and base_sha). Could we extract them somewhere? Maybe module constants?

Dangerfile
```diff
@@ -6,9 +6,21 @@ module VSPDanger
...
 
       [
+        GemfileProtector.new.run,
         ChangeLimiter.new.run,
         MigrationIsolator.new.run
-      ]
+      ].sort(&method(:severity_sort))
+    end
+
+    # expects severities of :error > :warning > :info
```
ericboehs, line 15: Why do we have to warn in this order?

Dangerfile
```diff
@@ -192,6 +204,48 @@ module VSPDanger
     end
   end
 
+  class GemfileProtector
```
ericboehs, line 207: What about something less scoped for the name? `GemfileEnterprisinator`? Also a comment before the class describing it (e.g. # Prevents Sidekiq Enterprise from being removed from Gemfile).

Dangerfile
erluti: I was wondering about that too.  That sounds like a good place for them!

Dangerfile
erluti: This PR solves it https://github.com/department-of-veterans-affairs/vets-api/pull/5779

So I'll take it out of here.  But we were doing a warn/fail based on the order of the array.  Which would mean they would have gotten a warning for 200+ lines of code and never saw a fail for changing migrations.

Dangerfile
```diff
@@ -180,15 +176,41 @@ module VSPDanger
...
-    def head_sha
-      `git rev-parse --abbrev-ref HEAD`.chomp
+    def gemfile_ok_message
+      { severity: :info, message: 'Gemfile changes acceptable.' }
     end
 
-    def base_sha
-      'origin/master'
+    def gemfile_diff
+      `git diff #{BASE_SHA}:Gemfile.lock #{HEAD_SHA}:Gemfile.lock`
```
ericboehs, line 213: It might be good to use the same diffing strat that Github uses:

```suggestion
      `git diff #{BASE_SHA}...#{HEAD_SHA} -- Gemfile.lock`
```

Dangerfile
erluti: Also, I misunderstood `fail`.  Apparently Dangerfile will return all warns and fails.

thilton-oddball (approved):
Looks good!

PR #5779, opened 2021-01-26
Dangerfile refactor messaging system to support future changes

erluti:
Add a `Result` class to clarify and normalize code in Dangerfile.  This makes it a bit obvious what's being returned and how it's being messaged.  Hopefully make it easier to manipulate how results are handled in the future too.


Dangerfile
```diff
@@ -209,12 +225,12 @@ module VSPDanger
 end
 
 if $PROGRAM_NAME != __FILE__
-  VSPDanger::Runner.run.each do |output|
-    case output[:severity]
-    when :error
-      fail output[:message] # rubocop:disable Style/SignalException
-    when :warning
-      warn output[:message]
+  VSPDanger::Runner.run.sort.each do |result|
```
erluti, line 228: Implemented `<=>` so that the Results are sortable, making sure we fail an error instead of suppressing it with a warning.

Dangerfile
```diff
@@ -16,25 +16,46 @@ module VSPDanger
     end
   end
 
+  class Result
+    ERROR = :error
+    WARNING = :warning
+    SUCCESS = :success
```
ericboehs, line 22: Why are these extracted to constants?

erluti: to prevent errors and troubleshooting issues with "magic strings".

Dangerfile
```diff
@@ -209,12 +217,12 @@ module VSPDanger
...
-    when :error
-      fail output[:message] # rubocop:disable Style/SignalException
-    when :warning
-      warn output[:message]
+  VSPDanger::Runner.run.each do |result|
+    case result.severity
+    when VSPDanger::Result::ERROR
+      failure result.message
+    when VSPDanger::Result::WARNING
+      warn result.message
```
ericboehs, line 225: The logic doesn't seem to be different than before.

Is the real fix here using `failure` over `fail`?

(I'm fine with the Result refactor.)

erluti: Yeah... It was, but then I backtracked the more I read about how Danger handles failure and warnings

erluti: At one point it was this
```
results = VSPDanger::Runner.run

  errors = results.select { |result| result.severity == VSPDanger::Result::ERROR }
  if errors.any?
    fail(errors.collect(&:message)) # rubocop:disable Style/SignalException
  end

  warnings = results.select { |result| result.severity == VSPDanger::Result::WARNING }
  if warnings.any? # rubocop:disable Style/IfUnlessModifier
    warn(warnings.collect(&:message))
  end
  ```
  But I assumed fail and warn were terminal, but they are not

ericboehs (approved):

ericboehs (approved):

PR #5780, opened 2021-01-27
Rename Vet360Redis to VAProfileRedis

lihanli:
## Description of change
rename `Vet360Redis` to `VAProfileRedis`
vet360 name is no longer in use and they want us to rename all references of vet360 to va_profile
all existing Vet360Redis caches will be invalidated, but it should just fetch data again and there shouldn't be any problems

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/19201

tested with unit tests

### DIFF
--- a/config/redis.yml
+++ b/config/redis.yml
@@ -69,12 +69,9 @@ development: &defaults
   pciu_address_dependencies:
     namespace: pciu-address-dependencies
     each_ttl: 604800
-  vet360_contact_info_response:
-    namespace: vet360-contact-info-response
+  va_profile_contact_info_response:
+    namespace: va-profile-contact-info-response
     each_ttl: 3600 # 1 hour
-  vet360_reference_data_response:
-    namespace: vet360-reference-data-response
-    each_ttl: 86400
   intent_to_file_response:
     namespace: intent-to-file-response
     each_ttl: 86400
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -310,7 +310,7 @@ RSpec/ScatteredLet:
     - 'spec/lib/sentry/processor/log_as_warning_processor_spec.rb'
     - 'spec/mailers/spool_submissions_report_mailer_spec.rb'
     - 'spec/models/form_profile_spec.rb'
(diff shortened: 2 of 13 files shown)


LindseySaari:
We may want this change to go out on an off cycle deploy since all data will need to be fetched again. Thoughts?

lihanli:
> We may want this change to go out on an off cycle deploy since all data will need to be fetched again. Thoughts?

that works for me

LindseySaari (approved):
Looks good & straightforward. I'll leave it up to you to schedule an off cycle deploy for this. From previous experience, I'd possibly wait until Monday so that any caching issues don't arise over the weekend.

bosawt (approved):
Looks good! Appeared to retrieve cached Profile data properly on deployed branch

config/redis.yml
```diff
@@ -69,12 +69,9 @@ development: &defaults
...
     namespace: pciu-address-dependencies
     each_ttl: 604800
-  vet360_contact_info_response:
-    namespace: vet360-contact-info-response
+  va_profile_contact_info_response:
+    namespace: va-profile-contact-info-response
     each_ttl: 3600 # 1 hour
-  vet360_reference_data_response:
-    namespace: vet360-reference-data-response
-    each_ttl: 86400
```
omgitsbillryan, line 77: No rename on this one? What about the class that uses this key?

omgitsbillryan: Just grep'd and it looks like no classes use `vet360_reference_data_response`, 🤷  👍 .

omgitsbillryan (approved):
If you don't care about the old redis data, and the new redis data will build itself up over time, I don't see the motivation for an off-cycle deploy. What's your thinking @LindseySaari ?

LindseySaari:
> If you don't care about the old redis data, and the new redis data will build itself up over time, I don't see the motivation for an off-cycle deploy. What's your thinking @LindseySaari ?

I didn't get to dig into the code, but my thinking from the past was that when I renamed a cache key, every request had to then make a call to MPI instead of pulling from the cache and didn't want to overload some external system. This may not be the case here

PR #5781, opened 2021-01-27
18753 mobile request claim decision

jperk51:
## Description of change
Implementation of claims decision request for mobile 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#18753

Tests added and docs updated


modules/mobile/app/controllers/mobile/v0/claims_and_appeals_controller.rb
```diff
@@ -76,7 +83,7 @@ def upload_documents
         )
         raise Common::Exceptions::ValidationErrors, document_data unless document_data.valid?
 
-        jid = document_upload_service.upload_document(document_data)
+        jid = evss_claim_service.upload_document(document_data)
         render json: { data: { job_id: jid } }, status: :accepted
```
kreek, line 87: do we want to add 'happy logging' here for the job id? Then we could look up the job status in sidekiq if it ever needed debugging.

kreek (approved):
One comment but otherwise LGTM 🚀

modules/mobile/app/controllers/mobile/v0/claims_and_appeals_controller.rb
jperk51: I added a sentry log to both doc upload and request decision reporting the claim id and the jid for that request.

PR #5782, opened 2021-01-27
CH31 Form bug fix

digitaldrk:
## Description of change
We made a change [here](https://github.com/department-of-veterans-affairs/vets-api/pull/5675). It was originally intended for the CH36 form (The focus of this PR is the CH31 form).

Both forms work similarly so I made a change for the 31. Unfortunately, instead of adding data a user submitted we were overwriting data and thus not sending required information to our partner VRE/CMS.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#9180


digitaldrk:
https://github.com/department-of-veterans-affairs/vets-api/pull/5782

LindseySaari (approved):

PR #5783, opened 2021-01-27
Log Failed HLR-related Requests to the PII Log

edmangimelli:
from [the ticket](https://github.com/department-of-veterans-affairs/va.gov-team/issues/19143)
>Currently we aren't capturing the complete request data anywhere when an HLR request fails. Now that we've soft launched, let's capture the complete request data in the PersonalInformationLog (which gets cleared every 2 weeks) for failures so that we can better debug failed requests.

### DIFF
--- a/app/controllers/appeals_base_controller.rb
+++ b/app/controllers/appeals_base_controller.rb
@@ -5,14 +5,8 @@ require 'decision_review/service'
 
 class AppealsBaseController < ApplicationController
   include ActionController::Serialization
+  include FailedRequestLoggable
   before_action { authorize :appeals, :access? }
-
-  class << self
-    def exception_hash(exception)
-      %i[message backtrace key response_values original_status original_body]
-        .reduce({}) { |hash, key| hash.merge({ key => exception.try(key) }) }
-    end
-  end
 
   private
 
@@ -49,32 +43,5 @@ class AppealsBaseController < ApplicationController
       request_body_class_name: request.try(:body).class.name,
       request_body_string: request.try(:body).try(:string)
     }
-  end
-
-  def current_user_hash
-    hash = {}
-
-    %i[first_name last_name birls_id icn edipi mhv_correlation_id participant_id vet360_id ssn]
-      .each { |key| hash[key] = @current_user.try(key) }
-
-    hash[:assurance_level] = begin
-                               @current_user.loa[:current]
-                             rescue
-                               nil
-                             end
-
-    hash[:birth_date] = begin
-                          @current_user.va_profile.birth_date.to_date.iso8601
-                        rescue
-                          nil
-                        end
-    hash
-  end
-
-  def log_exception_to_personal_information_log(exception, error_class:, data: {})
-    PersonalInformationLog.create!(
-      error_class: error_class,
-      data: { user: current_user_hash, error: self.class.exception_hash(exception) }.merge(data)
-    )
   end
 end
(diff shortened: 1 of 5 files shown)


app/controllers/v0/higher_level_reviews_controller.rb
```diff
@@ -4,12 +4,32 @@ module V0
...
     def create
       render json: decision_review_service
         .create_higher_level_review(request_body: request_body_hash, user: current_user)
         .body
+    rescue => e
+      request_data = begin
+                       { body: request_body_hash }
+                     rescue
+                       request_body_debug_data
+                     end
```
edmangimelli, line 25: the intent of the rescue is to capture any exception thrown by `create_higher_level_review`, but there's an off-chance that an exception is thrown before `create_higher_level_review` is called via `request_body_hash`

app/controllers/appeals_base_controller.rb
```diff
@@ -7,6 +7,13 @@ class AppealsBaseController < ApplicationController
   include ActionController::Serialization
   before_action { authorize :appeals, :access? }
 
+  class << self
+    def exception_hash(exception)
+      %i[message backtrace key response_values original_status original_body]
```
LindseySaari, line 12: Will the 'key' value present in this array clash with the key arg below?

app/controllers/appeals_base_controller.rb
```diff
@@ -36,4 +43,38 @@ def get_hash_from_request_body
   def request_body_is_not_a_hash_error
     DecisionReview::ServiceException.new key: 'DR_REQUEST_BODY_IS_NOT_A_HASH'
   end
+
+  def request_body_debug_data
```
LindseySaari, line 47: Can these methods that you introduced here be moved out to a helper or concern to avoid controller bloating?

app/controllers/appeals_base_controller.rb
edmangimelli: shouldn't! unless i'm misunderstanding you.

trying to capture all of these (if they're present)
https://github.com/department-of-veterans-affairs/vets-api/blob/8546477896cc1247aa5d2d54866007df13eb5bec/lib/common/exceptions/backend_service_exception.rb#L12

edmangimelli: but agreed --it's a funny situation that the variable `key` will get set to the symbol `:key` while reducing that array. I don't think it's funny enough to change my variable name, though. I guess i could go to `k` if our linter allows

app/controllers/appeals_base_controller.rb
edmangimelli: @LindseySaari  I'm down for that! Do you have any suggestions on a good name? I'm not so great at naming Concerns.

LindseySaari: Trying to think of how to categorize these new methods? Maybe something like Appeals Data? Or even AppealsBaseConcern might work?

edmangimelli: most of these methods are for helping log a failed request to the PII table.
`FailedRequestLoggable`
`PersonalInformationLoggable`
`PiiLoggable`
?

LindseySaari: Ooo I like all of those. Maybe FailedRequestLoggable? You're name suggestions were WAY better than mine 😺

LindseySaari (approved):

edmangimelli:
thanks for the review @LindseySaari ! 🍻

PR #5784, opened 2021-01-27
Update mobile api docs to fix phone extension

kreek:
## Description of change
Extension attribute was missing from the payload for phone creation/updates. Also fixes a yml ref from when I moved them into individual files.

## Original issue(s)
N/A


jperk51 (approved):
LGTM

PR #5792, opened 2021-01-28
remove git commit / success tracking

LindseySaari:
## Description of change
In favor of using DataDog metrics for vtk usage/success metric tracking, this PR removes the previous success tracking approach to create a git commit with a standard commit message that could be searchable via GH.


erluti (approved):
LGTM

PR #5798, opened 2021-01-28
Setting Appropriate Doc Type for Form 674

kathleencrawford:
## Description of change

When submitting a 674 (VA Form 21-674b), we generate a PDF and send it to VBMS but currently we are using the incorrect document type. We need to set the document type that corresponds to the 674 form, instead of using the default doc type.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#19089

This is a minor fix so no new settings or logging have been added.

## Testing
- [x] Local testing
- [x] Updated specs
- [x] Staging testing planned

### DIFF
--- a/app/workers/vbms/submit_dependents_pdf_job.rb
+++ b/app/workers/vbms/submit_dependents_pdf_job.rb
@@ -36,7 +36,7 @@ module VBMS
 
     def generate_pdf(claim, submittable_686, submittable_674)
       claim.upload_pdf('686C-674') if submittable_686
-      claim.upload_pdf('21-674') if submittable_674
+      claim.upload_pdf('21-674', doc_type: '143') if submittable_674
     end
   end
 end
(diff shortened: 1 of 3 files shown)


app/workers/vbms/submit_dependents_pdf_job.rb
```diff
@@ -36,7 +36,7 @@ def send_error_to_sentry(error, saved_claim_id)
 
     def generate_pdf(claim, submittable_686, submittable_674)
       claim.upload_pdf('686C-674') if submittable_686
-      claim.upload_pdf('21-674') if submittable_674
+      claim.upload_pdf('21-674', doc_type: '143') if submittable_674
```
LindseySaari, line 39: Just want to make sure that the 143 doc type above isn't a typo. I noticed that the other changes have a doc type of 148

kathleencrawford: @LindseySaari thank you for double checking.  It is not a typo.  148 is the doc type that should be used as the "default".  In this one particular line, we are generating a different doc type, so we do want 143 here.  But again, thank you for checking as that could easily have been a typo!

LindseySaari: Ok great!

LindseySaari (approved):

PR #5799, opened 2021-01-29
Document `error_key` Behavior in EVSS::ServiceException

edmangimelli:
When debugging lib code, I found [this comment](https://github.com/department-of-veterans-affairs/vets-api/blob/27e8132dfd1e125bcfe80661446229baa4743a8f/lib/evss/service_exception.rb#L18) slightly confusing.
Clarifying it and adding some tests for further clarification of existing behavior.


thilton-oddball (approved):
Looks good!

PR #5802, opened 2021-01-29
API 4702 Add Veteran Address, Email, Phone

ElleeB:
## Description of change

- We refactored NOD form data to print 'USE DATA ON FILE', hoping to increase accuracy during intake
- The Board has asked that we return to accepting and printing the Veteran's address, email, and ph number

- Please pay close attention to the Swagger file. I am still unfamiliar with all the ins and outs :)

## Original ticket
https://vajira.max.gov/browse/API-4702


modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
```diff
@@ -20,7 +20,8 @@ def self.date_from_string(string)
     attr_encrypted(:form_data, key: Settings.db_encryption_key, marshal: true, marshaler: JsonMarshal::Marshaller)
     attr_encrypted(:auth_headers, key: Settings.db_encryption_key, marshal: true, marshaler: JsonMarshal::Marshaller)
 
-    validate :validate_hearing_type_selection
+    validate :validate_address_unless_homeless, on: :create, if: :form_data_present?
```
ElleeB, line 23: Please take a look here and make sure I'm not missing any cases. I changed this to `on: create` as opposed to the previous return statement in the custom validation method. After removing PII, the NOD is updated, thus the validation fails unless skipped, since the address will have been removed.

modules/appeals_api/app/services/appeals_api/pdf_construction/notice_of_disagreement/v1/form_fields.rb
```diff
@@ -20,7 +20,7 @@ def veteran_dob
           'F[0].Page_1[0].DateSigned[0]'
         end
 
-        def mailing_address_number_and_street
+        def mailing_address
```
ElleeB, line 23: I changed the name because the address object contains much more than just the number and street (quite different than HLR).

modules/appeals_api/config/schemas/10182.json
```diff
@@ -46,18 +46,50 @@
...
       "additionalProperties": false,
-      "required": [ "zipCode5" ]
+      "properties": {
+        "addressLine1":            { "type": "string" },
+        "addressLine2":            { "type": "string" },
+        "addressLine3":            { "type": "string" },
+        "city":                    { "type": "string" },
+        "stateCode":               { "$ref": "#/definitions/nodCreateStateCode" },
+        "countryName":             { "type": "string" },
+        "zipCode5":                { "type": "string", "pattern": "^[0-9]{5}$" },
```
cilestin, line 68: zipCode5 should still have `minLength` and `maxLength` set.

modules/appeals_api/config/schemas/10182.json
```diff
@@ -46,18 +46,50 @@
...
+    },
+
+
+    "nodCreatePhone": {
+      "$comment": "the phone fields must not exceed 25 chars, when concatenated",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "countryCode":     { "type": "string", "pattern": "^[0-9]+$" },
+        "areaCode":        { "type": "string", "pattern": "^[2-9][0-9]{2}$" },
```
cilestin, line 81: Add `minLength` & `maxLength` limits to any of these address parts where the regex defines a length.

modules/appeals_api/config/schemas/10182.json
```diff
@@ -46,18 +46,50 @@
       "additionalProperties": false,
       "properties": {
         "homeless":            { "type": "boolean" },
-        "representativesName": { "type": "string", "maxLength": 120 },
-        "address": { "$ref": "#/definitions/nodCreateAddress" }
+        "address":             { "$ref": "#/definitions/nodCreateAddress" },
+        "phone":               { "$ref": "#/definitions/nodCreatePhone" },
+        "emailAddressText":    { "$ref": "#/definitions/nodCreateEmail" },
```
cilestin, line 51: add a `minLength: 3` and `maxLength` that will fit within the autosize box without truncation.

modules/appeals_api/app/services/appeals_api/pdf_construction/notice_of_disagreement/v1/structure.rb
```diff
@@ -17,10 +17,10 @@ def form_fill
             form_fields.veteran_ssn => form_data.veteran_ssn,
             form_fields.veteran_file_number => form_data.veteran_file_number,
             form_fields.veteran_dob => form_data.veteran_dob,
-            form_fields.mailing_address_number_and_street => form_data.mailing_address_number_and_street,
+            form_fields.mailing_address => form_data.mailing_address,
```
cilestin, line 20: How large can this get before the pdf field fails? Do you think it'd make sense to turn this into an autosize box (as a separate ticket)?

cilestin (changes requested):

modules/appeals_api/app/services/appeals_api/pdf_construction/notice_of_disagreement/v1/structure.rb
```diff
@@ -17,10 +17,10 @@ def form_fill
...
             form_fields.homeless => form_data.homeless,
             form_fields.preferred_phone => form_data.preferred_phone,
-            form_fields.preferred_email => form_data.preferred_email,
+            # form_fields.preferred_email => form_data.preferred_email,
```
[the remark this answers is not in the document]
ElleeB: missed that. ooop!

modules/appeals_api/app/services/appeals_api/pdf_construction/notice_of_disagreement/v1/structure.rb
ElleeB: I was going to, but then tested it out and it fits quite a bit. I'll test again and let you know exactly how much.

modules/appeals_api/config/schemas/10182.json
ElleeB: In looking into this, I noticed that Puerto Rico' country code has a hyphen in it `1-787`, but our regex accepts . Is this something we need to consider? But! It appears that we accept up to 14 characters for a phone number, so that would account for someone entering `17873136612345`, for example. Thoughts?

modules/appeals_api/config/schemas/10182.json
ElleeB: I'll figure out truncation cut-off and update! What's the rationale for `minLength: 3` ?

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
```diff
@@ -114,5 +149,9 @@ def birth_date(who)
     def header_field_as_string(key)
       auth_headers&.dig(key).to_s.strip
     end
+
+    def form_data_present?
+      proc { |a| a.form_data.present? }
```
yakschuss, line 154: What would be a situation where form data would not be present? It seems like we're "not validating in the case of form data not being present." Is that an acceptable edge case?

modules/appeals_api/app/swagger/appeals_api/v1/schemas/notice_of_disagreements.rb
```diff
@@ -143,6 +154,22 @@ class NoticeOfDisagreements
                       key :description, 'value of submitted homeless key'
                     end
```
yakschuss, line 156: I think these should have character lengths as well, right?

modules/appeals_api/config/schemas/10182.json
```diff
@@ -46,18 +46,50 @@
       "additionalProperties": false,
       "properties": {
         "homeless":            { "type": "boolean" },
```
yakschuss, line 48: Can/does the schema validate homeless OR address with `oneOf`?

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
ElleeB: I was wondering the same thing. It is leftover from the original code. HLR has the same (I moved it, but it's the same). I don't know the answer to this question. I will see what I can find out!

modules/appeals_api/app/swagger/appeals_api/v1/schemas/notice_of_disagreements.rb
ElleeB: Will do!

modules/appeals_api/config/schemas/10182.json
ElleeB: This was nixed way back when, during the early days of the NOD project, due to the complexity and supposedly unreadable errors. Maybe we should revisit?

modules/appeals_api/config/schemas/10182.json
cilestin: Hyphens shown in phone numbers are for readability - they aren't a requirement for input. Additionally, for Puerto Rico in particular, its country code is `1` while its area codes are `787` or `939`, so a full phone number would be something like `17875551234`.

Similarly, if you were trying to call Bermuda, which still resides in the `1` country code, a phone number for it would be `14415551234`.

Germany, however, has `49` as its country code along with a number of different area & specialty codes to where some numbers going into the Kelley Barracks, for instance, could have number similar to:
```countryCode: 49
areaCode: 711
phoneNumber: 5551234```

File: modules/appeals_api/config/schemas/10182.json (hunk 5 above)
Comment (cilestin) [reply modules/appeals_api/config/schemas/10182.json:51]: This was for email address. The pattern is looking for 3 total characters to pass, so that's what I based the `minLength` off of.

File: modules/appeals_api/config/schemas/10182.json (hunk 10 above)
Comment (cilestin) [reply modules/appeals_api/config/schemas/10182.json:48]: If we are going to revisit, let's do it as a separate ticket.

File: modules/appeals_api/config/schemas/10182.json [hunk 11]
```diff
@@ -46,18 +46,50 @@
...
+      "required": [ "addressLine1", "city", "countryName", "zipCode5" ]
+    },
+
+
+    "nodCreatePhone": {
+      "$comment": "the phone fields must not exceed 25 chars, when concatenated",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "countryCode":     { "type": "string", "pattern": "^[0-9]+$", "minLength": 1, "maxLength": 1 },
```
cilestin, line 80: This maxLength is incorrect according to the pattern. I'm not sure how much of an edge case it would be, but if someone's main contact info was still in Germany (because they were stationed there recently?) then they wouldn't be able to enter `49` as the country code given the `maxLength: 1`.

modules/appeals_api/config/schemas/10182.json
ElleeB: Ok, I thought maybe they couldn't be separated given the resource I was looking at. I wasn't worried about the hyphen though, but number of characters. As long as we're certain country codes will always only be one digit, we're good.

modules/appeals_api/config/schemas/10182.json
ElleeB: Oh! Email, not address. Gotcha.

modules/appeals_api/config/schemas/10182.json
```diff
@@ -46,18 +46,50 @@
...
+        "areaCode":        { "type": "string", "pattern": "^[2-9][0-9]{2}$", "minLength": 3, "maxLength": 3 },
+        "phoneNumber":     { "type": "string", "pattern": "^[0-9]{1,14}$", "minLength": 7, "maxLength": 14 },
```
cilestin, line 82: This minLength doesn't follow the pattern and may not be correct for phone numbers outside the US.

cilestin (changes requested):

modules/appeals_api/config/schemas/10182.json
yakschuss: Potentially! With Drew's error class, this might be a good way to validate the incoming schema.

modules/appeals_api/app/services/appeals_api/pdf_construction/notice_of_disagreement/v1/structure.rb
ElleeB: Mailing address can take in 165 characters the way it is now. Thoughts?

modules/appeals_api/config/schemas/10182.json
ElleeB: Yeah, that was a derp on my part. So:

- Isle of Man country code: `44-1624`

Safe to say 6 digits max?

modules/appeals_api/config/schemas/10182.json
ElleeB: Creating a ticket now

modules/appeals_api/config/schemas/10182.json
ElleeB: `Thanks to the international phone numbering plan (ITU-T E. 164), phone numbers cannot contain more than 15 digits. The shortest international phone numbers in use contain seven digits.`

modules/appeals_api/config/schemas/10182.json
cilestin: We are not certain, thus my rant regarding the German country code ;) From what I can tell the max for a country code is 3 digits, but I'm not incredibly well versed in non-US phone number validation, so let's take off the maxLength limit and [add a note to the phone number description](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/appeals_api/config/schemas/200996.json#L128) regarding max length of the combined phone number.

modules/appeals_api/app/services/appeals_api/pdf_construction/notice_of_disagreement/v1/structure.rb
cilestin: The combining of the...5+ fields that make up the address likely mean we'll need to create a custom validation so that the combined mailing list doesn't exceed 165 characters. We should also add a `description` to the address schema noting that the maximum combined character limit is 165.

modules/appeals_api/config/schemas/10182.json
cilestin: Just to be clear, the Isle of Man's country code is `44` (for the country of United Kingdom) and its area code is `1624` (for the area of Isle of Man). The note about ITU-T E. 164 is useful though :) I believe we limit the combined phone number (including extension) to 20 characters for HLR - we might be able to get away with the same here, too.

modules/appeals_api/config/schemas/10182.json
ElleeB: It appears there is a standard for maximum phone number length: 15 digits. I provide more info in a comment below in the schema file

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
ElleeB: I checked with Charley and he can't recall exactly why this check was put into place, other than protecting against unwanted stores to the database. Given that the check is simply blocking the validations themselves, I am going to remove it, since the validations will stop the process if data isn't there. Please confirm I'm considering this correctly. The remaining concern: is there a reason we'd want to catch an empty form before validation? Performance is the only thing I can think of, and it doesn't seem like a valid issue.

modules/appeals_api/app/swagger/appeals_api/v1/schemas/notice_of_disagreements.rb
```diff
@@ -143,6 +216,22 @@ class NoticeOfDisagreements
...
 
+                    property :address do
+                      key :type, :object
+                      key :example, false
```
cilestin, line 221: No need for a false example. Remove it.

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
```diff
@@ -21,6 +21,7 @@ def self.date_from_string(string)
     attr_encrypted(:auth_headers, key: Settings.db_encryption_key, marshal: true, marshaler: JsonMarshal::Marshaller)
 
     validate :validate_hearing_type_selection
+    validate :validate_address_unless_homeless, on: :create
```
cilestin, line 24: Reminder to move the validations to a `validate` method with an `if:` proc like HLR does.

cilestin (changes requested):

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
ElleeB: Duh. It's the check to be sure the validations don't run after the removal of PII. So, I don't need both checks, having added `on create`, and this, so this will stay.

modules/appeals_api/app/swagger/appeals_api/v1/schemas/notice_of_disagreements.rb
```diff
@@ -42,8 +42,35 @@ class NoticeOfDisagreements
 
                 property :address do
                   key :type, :object
-                  key :description, 'Address of the Veteran'
-                  key :required, %i[zipCode5]
+                  key :description, 'Address of the Veteran - not required if Veteran is homeless. Cannot exceed 165
```
yakschuss, line 45: I assume the `Cannot exceed 165` is because it's a total length over 5 separate fields?

ElleeB: yes sir!

cilestin (approved):

yakschuss (approved):
This looks good to me!

PR #5806, opened 2021-01-29
Cris.shupp/3461 active storage

cshupp1:
## Original issue(s)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -105,7 +105,6 @@ PATH
 
 GEM
   remote: https://rubygems.org/
-  remote: https://enterprise.contribsys.com/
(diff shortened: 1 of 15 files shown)


Gemfile.lock
```diff
@@ -105,7 +105,6 @@ PATH
 
 GEM
   remote: https://rubygems.org/
-  remote: https://enterprise.contribsys.com/
```
annaswims, line 108: adding the  [sidekiq licence key](https://github.com/department-of-veterans-affairs/va.gov-team-sensitive/blob/master/platform/engineering/sidekiq-enterprise-setup.md)  locally will prevent this change

PR #5809, opened 2021-01-29
[5763] Refactoring user birth_date so that MPI or va_profile referenc…

bosawt:
…es stay within the User class

## Description of change
This PR refactors the User `birth_date` attribute so that other classes will no longer have to decide whether to choose from the `MPI` birth_date, or the `UserIdentity` birth_date from the authentication payload. This keeps that logic within the User class

## Original issue(s)
department-of-veterans-affairs/vets-api/issues/5763

There technically shouldn't be any Production changes with this refactor, although, since many places are no longer specifically deciding whether to pull the birth_date from UserIdentity or MPI, it's possible this may lead to an issue if a service was specifically expecting it to come from one or the other. I don't think this should be the case, but it's something we can watch out for

To test this:
- Ran specs
- Logged in with SSOe user and confirmed that `birth_date` can be inspected on the user. To do this, set a breakpoint in `app/controllers/v1/sessions_controller.rb`, line 58, after `user_login`, and try `@current_user.birth_date`. Can also inspect `@current_user.identity.birth_date` and `@current_user.mpi.profile.birth_date` to make sure the values match

### DIFF
--- a/app/controllers/v0/post911_gi_bill_statuses_controller.rb
+++ b/app/controllers/v0/post911_gi_bill_statuses_controller.rb
@@ -4,6 +4,7 @@ require 'evss/gi_bill_status/external_service_unavailable'
 require 'evss/gi_bill_status/gi_bill_status_response'
 require 'evss/gi_bill_status/outside_working_hours'
 require 'evss/gi_bill_status/service'
+require 'formatters/date_formatter'
 
 module V0
   class Post911GIBillStatusesController < ApplicationController
@@ -78,14 +79,8 @@ module V0
         participant_id: user.participant_id,
         vet360_id: user.vet360_id,
         ssn: user.ssn,
-        birth_date: birth_date_formatted(user.birth_date)
+        birth_date: Formatters::DateFormatter.format_date(user.birth_date, :datetime_iso8601)
       }.to_json
-    end
-
-    def birth_date_formatted(birth_date)
-      return unless birth_date
-
-      DateTime.parse(birth_date).iso8601
     end
 
     def skip_sentry_exception_types
--- a/app/controllers/v0/post911_gi_bill_statuses_controller.rb
+++ b/app/controllers/v0/post911_gi_bill_statuses_controller.rb
@@ -78,14 +78,14 @@ module V0
         participant_id: user.participant_id,
         vet360_id: user.vet360_id,
         ssn: user.ssn,
-        birth_date: iso8601_birth_date(user)
+        birth_date: birth_date_formatted(user.birth_date)
       }.to_json
     end
 
-    def iso8601_birth_date(user)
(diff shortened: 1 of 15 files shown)


bosawt:
Sorry for the reviewer groups on here (besides Identity), accidentally created real PR before converting to draft. This is intended to be reviewed by Identity only at first!

jimiray (approved):

LindseySaari:
It looks like there's some failing specs. I can re-review after they're passing

app/controllers/v0/post911_gi_bill_statuses_controller.rb
```diff
@@ -78,14 +78,14 @@ def user_json(user)
...
         vet360_id: user.vet360_id,
         ssn: user.ssn,
-        birth_date: iso8601_birth_date(user)
+        birth_date: birth_date_formatted(user.birth_date)
       }.to_json
     end
 
-    def iso8601_birth_date(user)
-      return nil unless user&.va_profile&.birth_date
+    def birth_date_formatted(birth_date)
```
LindseySaari, line 85: Does this method need to live in the controller?

LindseySaari: It also looks like the same method is repeated in lib/evss/auth_headers.rb and . lib/evss/disability_compensation_auth_headers.rb These repeat methods could probably be pulled out to a shared helper

bosawt: Oh yeah, hmmm, that might not be a bad idea. Basically you can call `user.birth_date` directly if you want a birth_date but you don't care much about the format (where the format is some default sensible option). But if you want a specific format, you use the `DateFormatterHelper` or whatever we'd call it

app/models/user.rb
```diff
@@ -336,6 +349,31 @@ def flipper_id
 
   private
 
+  def mpi_profile
+    if mpi.nil?
+      Rails.logger.info "[User] Cannot find MPI record for User with uuid: #{uuid}"
```
annaswims, line 354: this is going to be super noisy.  MPI will be nil for all users that are not LOA3 and this will be called a LOT.

bosawt: Ah yeah I'll remove this log. I was wondering in what cases MPI was nil

bosawt:
@annaswims @LindseySaari @jimiray I changed the solution a bit by adding a `DateFormatter` helper class. Would be good to know if this seems like it makes it more clear how different forms want to parse dates. If this works well, we could potentially try to leverage all presented dates to follow this pattern

jimiray (approved):
Much better, I like the formatter. Now if we can get everyone to use it.

LindseySaari (approved):
Nice job on the Formatter!

PR #5819, opened 2021-02-01
Update Appeal status if error throws during PDF submit job

yakschuss:
[API-4763](https://vajira.max.gov/browse/API-4763)

We're not currently changing the status of an appeal unless the error thrown is an Upload error from Central Mail.

This changes the error handling pattern to capture _any_ error during the PDF submission, and update it's status accordingly.

### DIFF
--- a/modules/appeals_api/app/workers/appeals_api/higher_level_review_pdf_submit_job.rb
+++ b/modules/appeals_api/app/workers/appeals_api/higher_level_review_pdf_submit_job.rb
@@ -20,8 +20,8 @@ module AppealsApi
         higher_level_review.update!(status: 'submitting')
         upload_to_central_mail(higher_level_review, stamped_pdf)
         File.delete(stamped_pdf) if File.exist?(stamped_pdf)
-      rescue
-        higher_level_review.update!(status: 'error')
+      rescue => e
+        higher_level_review.update!(status: 'error', code: e.class.to_s, detail: e.message)
         raise
       end
     end
--- a/modules/appeals_api/app/workers/appeals_api/higher_level_review_pdf_submit_job.rb
+++ b/modules/appeals_api/app/workers/appeals_api/higher_level_review_pdf_submit_job.rb
@@ -13,11 +13,17 @@ module AppealsApi
 
     def perform(higher_level_review_id, retries = 0)
       higher_level_review = AppealsApi::HigherLevelReview.find(higher_level_review_id)
-      @retries = retries
-      stamped_pdf = AppealsApi::PdfConstruction::Generator.new(higher_level_review).generate
-      higher_level_review.update!(status: 'submitting')
-      upload_to_central_mail(higher_level_review, stamped_pdf)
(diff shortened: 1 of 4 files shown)


cilestin (approved):
🚢

modules/appeals_api/app/workers/appeals_api/higher_level_review_pdf_submit_job.rb
```diff
@@ -13,11 +13,17 @@ class HigherLevelReviewPdfSubmitJob
...
-      File.delete(stamped_pdf) if File.exist?(stamped_pdf)
+
+      begin
+        @retries = retries
+        stamped_pdf = AppealsApi::PdfConstruction::Generator.new(higher_level_review).generate
+        higher_level_review.update!(status: 'submitting')
+        upload_to_central_mail(higher_level_review, stamped_pdf)
+        File.delete(stamped_pdf) if File.exist?(stamped_pdf)
+      rescue
+        higher_level_review.update!(status: 'error')
```
cilestin, line 24: Looking over this again, I feel like we should explore including a code, description, and source to this. Is there anything in `map_response` that might aid us?

cilestin (changes requested):

modules/appeals_api/app/workers/appeals_api/higher_level_review_pdf_submit_job.rb
yakschuss: I'm not familiar with `map_response`. What are you referring to specifically?

cilestin: When we send the data to CentralMail we process its response and use `map_response` to figure out what `code`, `status`, and `description` to update the model with. I thought it might be a good jumping off point to supplying our own data into those same fields when non-CentralMail errors happen.

Even something simple like the `code` being "Exception" and the `description` having the error class & truncated message would be beneficial, I think.

yakschuss: To be clear, the map_response looks for the error returned from `CentralMail`, and then populates the `AppealsApi::UploadError` which inherits from `CentralMail::UploadError`. It then populates the code, and status - necessary for initializing the UploadError. It also increments the StatsD error count for that specific error. 

I think we can entertain the idea of defining a custom catchall error for job failure, but at this point - it's not exactly clear what we're going to be rescuing... such as a syntax error, or serialization error as seen with the `perform_async(hlr)` issue we had earlier.

cilestin: Yeah, a catchall error is really all I was looking for, with the `description` populated with _something_ from the exception that was thrown so that we have at least some idea of what went on to narrow down our search within the logs.

cilestin (approved):
Thanks for the updates! LGTM 🚀

PR #5823, opened 2021-02-01
Breakers should consider 500 range errors outages

annaswims:
## Description of change
When we get a 500 range response Faraday raises `Faraday::ClientError`. [ It seems like breakers intends](https://github.com/department-of-veterans-affairs/breakers/blob/master/lib/breakers/uptime_middleware.rb#L58) to count any response in the 500-range, but [rescue block](https://github.com/department-of-veterans-affairs/breakers/blob/master/lib/breakers/uptime_middleware.rb#L85) doesn't include  `Faraday::ClientError`

we found this when [EVSS 500 errors in sentry ](http://sentry.vfs.va.gov/organizations/vsp/issues/466/?project=3&query=is%3Aunresolved+500&statsPeriod=14d) weren't being counted as errors in the [Backend service report](http://grafana.vfs.va.gov/d/000000050/backend-service-report?orgId=1)

## Original issue(s)
department-of-veterans-affairs/va.gov-team#19394

There will likely be more external service alerts as things that have been erroring all along get counted as errors.

This might make a big difference in VSP's SLO dashboard.

### DIFF
--- a/lib/common/client/configuration/base.rb
+++ b/lib/common/client/configuration/base.rb
@@ -103,6 +103,8 @@ module Common
               (500..599).cover?(exception.response_values[:status])
             elsif exception.is_a?(Common::Client::Errors::HTTPError)
               (500..599).cover?(exception.status)
+            elsif exception.is_a?(Faraday::ClientError)
+              (500..599).cover?(exception.response[:status])
             else
               false
             end
--- a/lib/common/client/configuration/base.rb
+++ b/lib/common/client/configuration/base.rb
@@ -109,9 +109,8 @@ module Common
               (500..599).cover?(exception.response_values[:status])
             elsif exception.is_a?(Common::Client::Errors::HTTPError)
               (500..599).cover?(exception.status)
-            elsif exception.is_a?(Faraday::ClientError) || exception.is_a?(Faraday::ServerError)
-              # we're not yet using Faraday > 1.0, but when we do, 500 errors will be a ServerError
-              # before then they will be Faraday::ClientError
+            elsif exception.is_a?(Faraday::ClientError)
+              # we're not yet using Faraday > 1.0, but when we do, 500 errors will be Faraday::ServerError
               (500..599).cover?(exception.response[:status])
             else
               false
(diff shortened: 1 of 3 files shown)


lib/common/client/configuration/base.rb
```diff
@@ -92,23 +92,30 @@ def request_options
         def breakers_service
           return @service if defined?(@service)
 
+          @service = create_new_breakers_service(breakers_matcher, breakers_exception_handler)
```
annaswims, line 95: Rubocop was yelling at me about complexity so I split out exception_handler and matcher into methods.  the intended change is `elsif exception.is_a?(Faraday::ClientError)`

LindseySaari: Cool, I think this reads better anyways when broken up 👍

lib/common/client/configuration/base.rb
```diff
@@ -92,23 +92,30 @@ def request_options
...
+        end
 
-          exception_handler = proc do |exception|
+        def breakers_exception_handler
+          proc do |exception|
             if exception.is_a?(Common::Exceptions::BackendServiceException)
               (500..599).cover?(exception.response_values[:status])
             elsif exception.is_a?(Common::Client::Errors::HTTPError)
               (500..599).cover?(exception.status)
+            elsif exception.is_a?(Faraday::ClientError)
```
LindseySaari, line 112: Are we going to make a note to update this once we upgrade to Faraday 1.0?

LindseySaari: It also looks like we're covering all 500 errors here, are we only concerned with 502's and 503's?

annaswims: my bad.  I was seeing examples the problem with 502/503 but I did mean to include the whole 500-range.

LindseySaari (approved):

lib/common/client/configuration/base.rb
annaswims: Faraday has an [upgrade guide for 1.0](https://github.com/lostisland/faraday/blob/master/UPGRADING.md#faraday-10) that makes it clear, but I figured a comment wouldn't hurt either. I got lead astray because the [faraday documentation for ClientError](https://www.rubydoc.info/gems/faraday/Faraday/ClientError) changed as of 1.0

PR #5826, opened 2021-02-01
Update Proc State Type Code for 686

kathleencrawford:
## Description of change

During 686 testing, BGS asked us to update the proc state type from `Manual` to `MANUAL_VAGOV` so they can see that VA.gov is the application that is setting the claim to manual. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#19087

This is a very tiny PR that just changes one value, so no additional logging or configurations have been added.

- [x] Local testing
- [x] Staging testing planned


thilton-oddball (approved):
LGTM

PR #5827, opened 2021-02-02
Vaos grafana related changes

saneshark:
## Description of change
Moved the filtering logic in statsd_plugin.rb into lib/string_helpers.rb because it's now being used elsewhere as well. 
Will revert other changes made to statsd_plugin in previous PR since those did not produce expected result. 

Key findings: 

1. `status` is a reserved word in promQL and it will never work as a tag, it's being used elsewhere too. Should probably change that to `http_status` as we've done here.
2. statsd_plugin `on_error` is only concerned with 500 level exceptions. This should have been more obvious, but perhaps a helpful comment in there would help others grief of thinking they can somehow leverage that plugin for 400 level error instrumentation as well. 

Here we have resorted to introducing a new set of instrumentation for http_status responses that get returned from va_mobile based. You might ask why we didn't use `with_monitoring.` Even though there is a way to query in promQL for all of the different __name__ types prefixed with VAOS, tags are based on method name via `caller` instead of METHOD and PATH like we would like. Furthermore, it returns a pretty generic exception instead of an HTTP status code which would be much more helpful. What we've got now is as part of this PR is just more helpful and provides a more rich set of context to build dashboards with. 

## Original issue(s)


kreek (approved):
LGTM

grossb1 (approved):

thilton-oddball (approved):
All looks good

PR #5828, opened 2021-02-02
Add a date range filter to mobile Community Care appointments

kreek:
## Description of change
This is actually a bug in the upstream VAOS service but there's no timeline for them to fix it. This patch in our middleware will filter out any appointments not within the date range passed to out API endpoint.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#19361


jperk51 (approved):
LGTM

PR #5831, opened 2021-02-02
Add NODSerializer spec to have parity with HLR

yakschuss:
This PR adds a spec for NODSerializer, to maintain parity between HLR/NOD. Currently HLR serializer is being tested, NOD is not.

[4704](https://vajira.max.gov/browse/API-4704)


nathanbwright (approved):
I realize that you're just mirroring what currently exists for the HLR serializer and I think this is probably fine for now. I think at some point it would be nice to tighten up our serializer specs instead of having so many `expect` statements in one spec...

yakschuss:
> I realize that you're just mirroring what currently exists for the HLR serializer and I think this is probably fine for now. I think at some point it would be nice to tighten up our serializer specs instead of having so many `expect` statements in one spec...

I  definitely can make it cleaner! Will do.

PR #5833, opened 2021-02-02
Appeals api - weekly appeals report

nathanbwright:
## Description of change
This PR adds a weekly report (same format/data points as the current daily email with the exception of adding `Daily`/`Weekly` to the subject and title) for reporting stats and errors on HLR and NOD. It renames the current daily email job to better differentiate between the daily and weekly email.

## Original issue(s)
https://vajira.max.gov/browse/API-4807


cilestin (approved):
🚢

PR #5834, opened 2021-02-03
Remove double value of 'POW' in 526 specialIssues enum array

hinzed1127:
Linter caught this double value, removing one of them.

### DIFF
--- a/modules/claims_api/config/schemas/526.json
+++ b/modules/claims_api/config/schemas/526.json
@@ -28,7 +28,6 @@
         "enum": [
           "ALS",
           "HEPC",
-          "POW",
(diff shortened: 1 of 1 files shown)


LindseySaari (approved):

annaswims:
@hinzed1127  EVSS let us know that their documentation is incorrect and that  "PTSD/1" should be   "PTSD_1". I can forward the email if it's helpful.

hinzed1127:
@annaswims please do. We intend to do an audit and see if we really need to display the full list of flashes and special issues in the documentation, but definitely don't want to have anything explicitly wrong

PR #5844, opened 2021-02-03
14839 mobile cancel appointment

kreek:
## Description of change
Adds the mobile version of the cancel appointment endpoint. VA facilities keep different lists for why an appointment is cancelled. A call must be first be made to check if the facility has a cancel reason that is or is adjacent to 'unable to keep appointment'.

Related PR with just VCR cassettes to get this PR's line count under 500 https://github.com/department-of-veterans-affairs/vets-api/pull/5869

## Original issue(s)
department-of-veterans-affairs/va.gov-team#14839

* The mobile app is trying not to use Virtus gem based models or forms as the gem is no longer supported and the author has replaced it with the dry-rb gems. This PR adds dry-validation to the mobile gemspec.


jperk51 (approved):
LGTM

thilton-oddball (approved):
LGTM

PR #5849, opened 2021-02-03
remove two of VAOS flat facility feature flags

jenniemc:
Remove the flat facility feature flag because the VAOS application no longer calls on the flat facility flag names.


thilton-oddball (approved):
LGTM

PR #5853, opened 2021-02-04
Minor Tweaks to Form526 Reports Rake Task

edmangimelli:
Cleans up `print_total` function for BDD mode (removes colon and space). Switch to using ISO8601 styles dates for CSV output (to ensure that dates are lexicographically sortable and don't contain commas).


thilton-oddball (approved):
Looks good

PR #5856, opened 2021-02-04
grape poc

SlakrHakr:
Example payloads utilizing HATEOAS.

POST  http://localhost:3000/services/claims/v2/veterans/{token}/forms/526
```
{
    "id": "ece87700-c0bc-4e92-a5b0-f317d1e17a40",
    "status": "pending",
    "@links": [
        {
            "rel": "Check status of claim",
            "type": "GET",
            "url": "http://localhost:3000/services/claims/v2/claims/ece87700-c0bc-4e92-a5b0-f317d1e17a40"
        },
        {
            "rel": "Upload supporting documents",
            "type": "POST",
            "url": "http://localhost:3000/services/claims/v2/forms/526/ece87700-c0bc-4e92-a5b0-f317d1e17a40/attachments"
        }
    ]
}
```

GET  http://localhost:3000/services/claims/v2/veterans/{token}/claims/ece87700-c0bc-4e92-a5b0-f317d1e17a40
```
{
    "id": "ece87700-c0bc-4e92-a5b0-f317d1e17a40",
    "status": "pending",
    "attributes": {
        "vbmsClaimId": null,
        "claimType": null

        ...
    },
    "@links": [
        {
            "rel": "self",
            "type": "GET",
            "url": "http://localhost:3000/services/claims/v2/claims/ece87700-c0bc-4e92-a5b0-f317d1e17a40"
        }
    ]
}
}
```

This work also strove to generate documentation from the actual implementation, so hopefully doc never gets out of sync with the implementation itself.
For this, grape has a concept of "entities" which are essentially just serializers.
This then takes those entities, among other things, and generates openapi doc from those definitions.
Should mean no more doc separated from the code.

Documentation generated from grape entities:
screencapture-localhost-3001-explore-benefits-docs-claims-2021-02-04-11_40_31

Edit: Updated the above endpoint response examples after updating the implementation of HATEOAS in this poc to allow for multiple actions in a response.

Edit 2: Added the concept of a "Veteran token." The concept is essentially to break auth from the actual Veteran being acted upon. This token would then be required on all paths that relate to the Veteran, whether the consumer is a Veteran representative or the Veteran themselves. There is a new endpoint added for this concept to work that essentially takes in the 4 fields we're currently having representatives pass in endpoint headers and returns a token representation of that Veteran to the consumer.

Pros of this concept:
1. More obvious these endpoints are each for a particular Veteran. should resolve questions like: "if im a representative and i dont pass Veteran headers, can i get all claims for all the Veterans I service?"
2. Should remove the need for headers (outside of auth) entirely
3. Enforces the same workflow whether the consumer is a Veteran or a Veteran representative. Consistent workflow, with less branches, should allow for a better experience due to strange situations cropping up less often

Cons of this concept:
1. This uses JWT to generate the tokens from the provided data, which means the tokens can be pretty large
2. This requires consumers to hit an additional endpoint to generate this new token

How the documentation looks with these path changes:
screencapture-localhost-3001-explore-benefits-docs-claims-2021-02-11-21_53_42


modules/claims_api/app/controllers/claims_api/entities/v2/claim_entity.rb
```diff
@@ -0,0 +1,50 @@
+module ClaimsApi
```
SlakrHakr, line 1: good example of a grape-entity. serializer that doc is then generated from

modules/claims_api/app/controllers/claims_api/v2/claims.rb
```diff
@@ -0,0 +1,72 @@
...
+          detail <<~X
+            Returns pending claims submitted through this API as well as any established claims submitted
+            from other sources.
+          X
+          success ClaimsApi::Entities::V2::ClaimEntity
+          failure [[401, 'Unauthorized', 'ClaimsApi::Entities::V2::ErrorsEntity']]
+          tags ['Claims']
+          security [{ bearer_token: [] }]
+        end
+        get '/' do
```
SlakrHakr, line 28: good example of a typical endpoint with grape. this is the "all claims" endpoint

modules/claims_api/app/controllers/claims_api/v2/claims.rb
```diff
@@ -0,0 +1,72 @@
...
+      resource :claims do
+        desc 'Return all claims associated with Veteran.' do
+          detail <<~X
+            Returns pending claims submitted through this API as well as any established claims submitted
+            from other sources.
+          X
+          success ClaimsApi::Entities::V2::ClaimEntity
+          failure [[401, 'Unauthorized', 'ClaimsApi::Entities::V2::ErrorsEntity']]
+          tags ['Claims']
+          security [{ bearer_token: [] }]
```
SlakrHakr, line 26: this is all just additional documentation for the openapi json doc
notice it referencing the entities. if the entities ever change, both the serialized response would change as well as this documentation

app/controllers/openid_application_controller.rb
```diff
@@ -12,82 +12,25 @@
...
 class OpenidApplicationController < ApplicationController
+  include OAuthConcerns
   skip_before_action :verify_authenticity_token
   skip_after_action :set_csrf_header
   before_action :authenticate
   TOKEN_REGEX = /Bearer /.freeze
 
   private
 
-  def permit_scopes(scopes, actions: [])
```
SlakrHakr, line 22: this was the most difficult part of all this, because a lot of this implementation is in the controller module controllers just extend from. unfortunately, this wouldnt work because grape api definitions have to extend a grape specific class. 

what this does is just move this methods out into a module (OAuthConcerns) that is then included in this file above and can then be included as "helpers" in grape endpoints below.

markgreenburg:
> Example payloads utilizing HATEOS.
> 
> POST http://localhost:3000/services/claims/v2/forms/526
> 
> ```
> {
>   "id": "####-####-####-####",
>   "self": "http://localhost:3000/services/claims/v2/claims/##-####-####-####",
>   "status": "pending"
> }
> ```
> 
> GET http://localhost:3000/services/claims/v2/claims/##-####-####-####
> 
> ```
> {
>   "id": "####-####-####-####",
>   "self": "http://localhost:3000/services/claims/v2/claims/##-####-####-####",
>   "status": "pending",
>   "attributes": {
>     "vbmsClaimId": "########",
>     "claimType": "Compensation"
> 
>     ...
>   }
> }
> ```
Out of curiosity, what's the driver of using the `self` property instead of something like an array of possible links? I haven't actually used hateoas concept in real life but the examples I've seen around look like they consolidate possible next actions into a single property, which I would think would make it more straightforward for a consumer to always be able to check something like `links` for consistency.

SlakrHakr:
> Out of curiosity, what's the driver of using the `self` property instead of something like an array of possible links? I haven't actually used hateoas concept in real life but the examples I've seen around look like they consolidate possible next actions into a single property, which I would think would make it more straightforward for a consumer to always be able to check something like `links` for consistency.

I've seen it both ways in various apis: self to denote the claim itself or a list of links.
I went with self just because it feels a bit simpler when youre basically just saying: "this is where you can get the claim itself." This link might not be as obvious what it is if it was just in a list of other links. And I don't believe there are any other GET links for these type of resources?

I could look into if there is an official standard and we could go with that, the more important thing is to be consistent of course.

modules/claims_api/app/controllers/claims_api/v2/forms/disability_compensation.rb
```diff
@@ -0,0 +1,112 @@
...
+              requires :attributes, type: Hash do
+                requires :veteran, type: Hash do
+                  requires :currentlyVAEmployee, type: Boolean
+                  # TODO: define necessary schema here
+                end
+              end
+            end
+          end
+          post '/' do
+            status 202
```
SlakrHakr, line 85: https://oddball.slack.com/archives/C019XKGS1CN/p1612893592002600

SlakrHakr:
https://github.com/department-of-veterans-affairs/vets-api/pull/5856#issuecomment-775273303
> I could look into if there is an official standard and we could go with that, the more important thing is to be consistent of course.

I saw various ways to implement HATEOAS, for better or for worse.
I went ahead and changed this poc's implementation to a list of links (example responses have been updated in the main description of this pr) to allow for multiple actions in a given response.

jw81:
Let me preface by saying that these are just some things that popped into my mind that I don't know the answers to, but they're things I wonder about.

- I know that web browsers implement url character lengths limits.  So if an endpoint url got pretty long due to JWT bloat, maybe that could be a prob?  (maybe not also because maybe browsers won't be hitting those urls directly).  I also don't know if servers enforce their own url character lengths limits.
- Do we have to be concerned with somehow encrypting the JWTs?  There's admittedly a lot I don't know about JWTs yet.  I just wonder how difficult it would be for a malicious person to parse some logs, find endpoint urls with JWTs, decode them, and extract the PII out of the JWTs.
- Do we have to be concerned with making a JWT expire very quickly and/or be like a single-use thing?
- I'm still new to the `vets-api` codebase as a whole but I wonder if the use of JWTs in URLs and HATEOAS would take us on an entirely different path from "what everyone else is doing"?  Which isn't necessarily a "bad" thing.  But it might be something we'd want to consider.

Just some initial thoughts.

SlakrHakr:
> * I know that web browsers implement url character length limits.  So if an endpoint url got pretty long due to JWT bloat, maybe that could be a prob?

JWT tokens are always the same size, regardless of what is packaged in them.
There is a url limit but i believe its something like 2000 characters, so while i hate the length of those jwt tokens right in the middle of the path, it shouldnt cause a url length issue.

> * Do we have to be concerned with somehow encrypting the JWTs?  There's admittedly a lot I don't know about JWTs yet.  I just wonder how difficult it would be for a malicious person to parse some logs, find endpoint urls with JWTs, decode them, and extract the PII out of the JWTs.

As far as I know, JWT tokens can only be decoded with the same seed that was used to encode them. So the only way for someone to access the information package inside the token would be if they somehow had access to the seed used to create the token initially.

> * Do we have to be concerned with making a JWT expire very quickly and/or be like a single-use thing?

I'm not really sure on this one. youd probably not want them to be single-use, force them to hit the Veterans endpoint for each and every call, but no expiration seems bad as well. Maybe its a 24 hour expiration thing?

I'm not 100% sold on this JWT token thing honestly. I like the general idea of scoping the paths to a specific Veteran with some sort of id and JWT tokens were the only thing i could think of that could work in this type of scenario.

> * I'm still new to the `vets-api` codebase as a whole but I wonder if the use of JWTs in URLs and HATEOAS would take us on an entirely different path from "what everyone else is doing"?  Which isn't necessarily a "bad" thing.  But it might be something we'd want to consider.

I'm not super sure on this one. I know we need to have a similar vibe between the various services, but something like HATEOAS could be something we pioneer and convince others to add to the other services. I'm honestly not sure how much we can "go outside the box" so to speak.

SlakrHakr:
This didn't really get a warm reception.
Closing this for now. We will most likely go straight rails if kept in vets-api or reconsider grape if claims api is broken into its own service.

PR #5857, opened 2021-02-04
Prevent Country Code Lookup When There's No Address Info (VA 21P530 Form Profile)

edmangimelli:
Fixes `undefined method 'country' for nil:NilClass`
([sentry](http://sentry.vfs.va.gov/organizations/vsp/issues/17590/?query=is%3Aunresolved+InProgressFormsController&statsPeriod=14d))
(not sure which team this form belongs to)


thilton-oddball (approved):
LGTM

PR #5860, opened 2021-02-04
19489 mobile appointments proxy

kreek:
## Description of change
This PR includes some appointment service updates and refactors that are in https://github.com/department-of-veterans-affairs/vets-api/pull/5860 but that PR was getting a little spicy 🌶️  so the non cancel appointment changes are going in here.

- moves the fetch facilities logic/combining of VA and CC appointments into a proxy class that sits between the controller and the various services needed for all the appointments list call.
- recently the upstream VAOS service added appointment ids to their response so the 'real' ids are used rather than a generated ones.
- clinicId is required for cancelling appointments so that has been added to the appointments list response.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#19489

* Adds a mobile module relative exceptions.yml local file for custom exceptions


jperk51 (approved):
LGTM

PR #5862, opened 2021-02-04
Apps_api refactor for DirectoryController

bradenshipley:
## Description of change
Due to a change in the kong_gateway settings for the `apps_api` module, there is no longer a need to have the `verify_auth` `before_action` in the `DirectoryController`. Now those routes are protected by an `apikey` requirement in the kong gateway config, so there's no need to add additional header-based authorization. Rspec tests referencing the `verify_auth` method have been adjusted.

Now that the gateway changes have been merged, the `DirectoryController` has full crud available to it, so the previous way of adding records to the application directory using a worker and a string of `find_or_create_by!` methods is no longer needed and has been deleted. This method wasn't scalable and was just a stop-gap.

It was also found while going over documentation for the read-only methods of the `DirectoryController` that the swagger documentation was missing the `#show` endpoint, so this has been added as well. The `apps_api` documentation is not shown on the developer portal yet, but this documentation is here just in case that becomes necessary in the near future.

## Original issue(s)
Kong gateway PR:
https://github.com/department-of-veterans-affairs/lighthouse-api-gateway-configs/pull/26

**Are there Swagger docs that were updated?**
Yes, a missing endpoint for the  `/directory/{app_name}` path has been added to the swagger docs.


thilton-oddball (approved):
All looks good!

PR #5800, opened 2021-01-29
add gems needed for feature to Gemfile

holdenhinkle:
# Part 1 (PR 1 of 8) in This Series

**Note about dependencies in this series:**  
Each PR in this series is dependent on the previous PR being merged.

The original branch is here (before it was broken up into these 8 branches) - https://github.com/department-of-veterans-affairs/vets-api/tree/16994_update_vets_website_config/cypress.json_file

**In this PR:**

- Update `Gemfile`

3 new gems have been added to `Gemfile`:

`google-api-client`
https://rubygems.org/gems/google-api-client
https://github.com/googleapis/google-api-ruby-client

`googleauth`
https://rubygems.org/gems/googleauth
https://github.com/googleapis/google-auth-library-ruby

`octokit`
https://rubygems.org/gems/octokit
https://github.com/octokit/octokit.rb

# General (mentioned on all prs in this series)

This 8-part PR resolves ticket https://app.zenhub.com/workspaces/vsp-5cedc9cce6e3335dc5a49fc4/issues/department-of-veterans-affairs/va.gov-team/16994

This feature updates the following files in `vets-website` automatically via a Sidekiq job each month with recent data from Google Analytics:

- `config/cypress.json`
- `src/platform/testing/e2e/cypress/support/commands/viewportPreset.js`

It:

- Requests two Google Analytics reports from the ga.gov organization:
  - Total number of users in the previous month
  - Total number of users in the previous month using each screen resolution (viewport width/height), sorted by number of users descending
- Parses the reports and creates viewport objects, categorized by mobile, tablet, desktop
- Grabs the the `cypress.json` and `viewportPreset.js` files from Github
- Updates the files
- Creates a new feature branch
- Commits the updated files
- Submits a pull request

## Google Analytics and GitHub Settings:

Google Analytics (`google_analytics_cvu`) and GitHub (`github_cvu`) settings have been added to credstash and the following files in `devops` (see [the PR](https://github.com/department-of-veterans-affairs/devops/pull/8457)):

- `ansible/deployment/config/vets-api/dev-settings.local.yml.j2`
- `ansible/deployment/config/vets-api/prod-settings.local.yml.j2`
- `ansible/deployment/config/vets-api/sandbox-settings.local.yml.j2`
- `ansible/deployment/config/vets-api/staging-settings.local.yml.j2`

`vets-api` `config/settings.yml` is updated in PR part 1.5 in this series (see below).

## Test Coverage

Simplecov reports test coverage at 100% for all files.

[...]

PR #5863, opened 2021-02-04
add Gemfile.lock for #5800

annaswims:
#5800 added gems to Gemfile but didn't commit changes to Gemfile.lock.


thilton-oddball (approved):
LGTM

PR #5865, opened 2021-02-05
Update NOD schema to check email against JSON API email spec

nathanbwright:
## Description of change
This PR updates our NOD schema to check supplied emails against the JSON schema draft 7 `email` format.

## Original issue(s)
https://vajira.max.gov/browse/API-4822


config/locales/exceptions.en.yml
```diff
@@ -244,6 +244,12 @@ en:
           detail: "The %{size} items provided did not match the definition"
           code: 147
           status: 422
+        format:
+          <<: *defaults
+          title: Invalid format
+          detail: "'%{value}' did not match the defined format"
+          code: 148
```
yakschuss, line 251: 😬  sorry! I just merged code 148 as const. You might want to change this to 149.

spec/lib/common/exceptions/detailed_schema_errors_spec.rb
```diff
@@ -130,11 +131,20 @@
   end
 
   context 'patterns' do
+    it 'has title, detail, and meta' do
+      data['pattern'] = Faker::Lorem.sentence
+      expect(subject[:title]).to eq 'Invalid format'
+      expect(subject[:detail]).to eq "'#{data['pattern']}' did not match the defined format"
+      expect(subject[:meta][:regex]).to eq '.@.'
```
yakschuss, line 138: Is this regex from the draft-7? This is what we had earlier, isn't it?

yakschuss (changes requested):
[EDIT] - whoops, didn't mean to request changes.

yakschuss (approved):
Looks good to me! I'd replace the code on `format` before merging, though.

spec/lib/common/exceptions/detailed_schema_errors_spec.rb
nathanbwright: Yeah, the diff is being strange. It's what we had before.

config/locales/exceptions.en.yml
```diff
@@ -250,6 +250,12 @@ en:
           detail: "The %{size} items provided did not match the definition"
           code: 147
           status: 422
+        format:
```
cilestin, line 253: So I think keeping the same `title` and `detail` as `pattern` could lead to some confusion, since the only difference in the response is the `code` and `meta`.  I think `format` should mention that it's a built-in part of the spec and not a custom regex like `pattern` is.

That or we just combine them both into the `pattern` error. I'm fine with either.

cilestin (changes requested):

cilestin (approved):
Thanks for the updates. LGTM!

PR #5866, opened 2021-02-05
Add sentry tag for each app

annaswims:
## Description of change
Add a sentry tag with the front end app name for all reqests that have the  `Source-App-Name` header.  The Source-App-Name header should be present for all apps because of our [frontend to backend mapping stats](http://grafana.vfs.va.gov/d/zGS4XzkMk/backend-to-frontend-app-mapping?orgId=1)

I did an end to (almost) end test by making a call from vets-website, to vets-api and verifying that the tag was present in the sidekiq job to create a sentry issue. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#7249


thilton-oddball (approved):
Looks good

PR #5870, opened 2021-02-05
Move model validation for address

ElleeB:
## Description of change
A Veteran's address is dependent on whether they are experiencing homelessness
We were validating this on the model
This PR moves the validation from the model to json schema.

## Ticket
https://vajira.max.gov/browse/API-4816

## Testing
Test removed from model spec and added to detailed_schema_errors_spec

### DIFF
--- a/modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
+++ b/modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
@@ -20,11 +20,7 @@ module AppealsApi
     attr_encrypted(:form_data, key: Settings.db_encryption_key, marshal: true, marshaler: JsonMarshal::Marshaller)
     attr_encrypted(:auth_headers, key: Settings.db_encryption_key, marshal: true, marshaler: JsonMarshal::Marshaller)
 
-    validate(
-      :validate_hearing_type_selection,
-      :validate_address_unless_homeless,
-      if: proc { |a| a.form_data.present? }
-    )
+    validate :validate_hearing_type_selection, if: :pii_present?
 
     def pdf_structure(version)
       Object.const_get(
@@ -110,12 +106,6 @@ module AppealsApi
 
     private
 
-    def validate_address_unless_homeless
-      return if veteran_homeless?
-
-      errors.add :form_data, I18n.t('appeals_api.errors.not_homeless_address_missing') if mailing_address.blank?
-    end
-
     def validate_hearing_type_selection
       return if board_review_hearing_selected? && includes_hearing_type_preference?
 
@@ -151,6 +141,11 @@ module AppealsApi
 
     def header_field_as_string(key)
       auth_headers&.dig(key).to_s.strip
+    end
+
+    # After expunging pii, form_data is nil, update will fail unless validation skipped
+    def pii_present?
+      proc { |a| a.form_data.present? }
     end
   end
 end
--- a/modules/appeals_api/config/schemas/10182.json
+++ b/modules/appeals_api/config/schemas/10182.json
@@ -40,18 +40,19 @@
       "required": [ "boardReviewOption", "socOptIn" ]
     },
 
-
     "nodCreateVeteran": {
       "type": "object",
       "additionalProperties": false,
       "properties": {
         "homeless":            { "type": "boolean" },
-        "address":             { "$ref": "#/definitions/nodCreateAddress", "minLength": 5, "maxLength": 165 },
+        "address":             { "$ref": "#/definitions/nodCreateAddress" },
         "phone":               { "$ref": "#/definitions/nodCreatePhone" },
         "emailAddressText":    { "$ref": "#/definitions/nodCreateEmail", "minLength": 5, "maxLength": 120 },
         "representativesName": { "type": "string", "maxLength": 120 }
       },
-      "required": [ "homeless", "phone", "emailAddressText" ]
+      "required": ["homeless",  "phone", "emailAddressText"],
+      "if": { "properties": { "homeless": { "const":  false } } },
+      "then": { "required":  ["address"] }
     },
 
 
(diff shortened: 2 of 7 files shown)


modules/appeals_api/spec/requests/v1/notice_of_disagreements_controller_spec.rb
```diff
@@ -69,9 +69,9 @@ def base_path(path)
       end
 
       it 'returns error objects in JSON API 1.0 ErrorObject format' do
-        expected_keys = %w[code detail links meta sentry_type source status title]
```
ElleeB, line 72: links and sentry_type are optional keys and were causing failures as they were passing nil.

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
```diff
@@ -20,11 +20,7 @@ def self.date_from_string(string)
     attr_encrypted(:form_data, key: Settings.db_encryption_key, marshal: true, marshaler: JsonMarshal::Marshaller)
     attr_encrypted(:auth_headers, key: Settings.db_encryption_key, marshal: true, marshaler: JsonMarshal::Marshaller)
 
-    validate(
-      :validate_hearing_type_selection,
-      :validate_address_unless_homeless,
-      if: proc { |a| a.form_data.present? }
-    )
+    validate(:validate_hearing_type_selection, if: proc { |a| a.form_data.present? })
```
ElleeB, line 23: Determined the complexity of this validation is best handled via the model. The logic required the use of the "not" keyword in the schema and could not be written in a way that would return errors tailored for the dependent field.

yakschuss: Now that we're using a single validation, perhaps extracting this proc into a method for readability makes sense?

`validate :hearing_type_selection, if: pii_still_present?`

yakschuss (approved):
Looks good to me!

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
ElleeB: I did that in my original and it was suggested I return to this, since I suppose the proc itself is pretty readable, but I'm with you in the "even more readable" camp. I'm going to change it. It will be more readable, especially if we refer to the pii in the method name. Done. Doing it.

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
```diff
@@ -152,5 +142,10 @@ def birth_date(who)
     def header_field_as_string(key)
       auth_headers&.dig(key).to_s.strip
     end
+
+    # After expunging pii, form_data is nil, update will fail unless validation skipped
```
ElleeB, line 146: I can remove this, but I really feel like leaving this is the nice thing to do for anyone in the future coming to the code.

cilestin: Have you verified this works as expected? My understanding with the `validate(...if:...)` was that it checks the type passed into the `if` and knows to call `self` as the passed object into the proc. By changing it to a symbol & method call, I'm not confident the `proc` inside `pii_present?` will do what you expect - it has no context for what `a` is, in this case.

I could totally be wrong about that, of course, it just looks odd to me (e.g. instead of doing a `self.form_data.present?`) and I want to make sure :)

modules/appeals_api/config/schemas/10182.json
```diff
@@ -40,18 +40,19 @@
...
     },
 
-
     "nodCreateVeteran": {
       "type": "object",
       "additionalProperties": false,
       "properties": {
         "homeless":            { "type": "boolean" },
-        "address":             { "$ref": "#/definitions/nodCreateAddress", "minLength": 5, "maxLength": 165 },
+        "address":             { "$ref": "#/definitions/nodCreateAddress" },
```
cilestin, line 48: I know I asked you to take this out but it _might_ be showing up in the documentation as a way of indicating to the consumer that the concatenated length has limits. I think we also added a `description` to the address somewhere that indicated these limits as well, so as long as there's _some_ indication in the documentation of the concatenated limit, it's all good.

cilestin (approved):
Overall looking good - I just had a couple comments to double check, but shouldn't be blockers if you find it's good as-is!

Approving to unblock if none of my comments need addressing.

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
ElleeB: I'll go back and check the context inside the new method. I do really like naming it what it actually is. I found that proc to be so confusing for so long until I finally dug into the removal of pii. As mentioned before, I just think it needs clarity, so I have no problem putting the proc back with the validation, but I'm gonna keep pushing for a comment in the code.

modules/appeals_api/config/schemas/10182.json
ElleeB: K I'll check!

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
ElleeB: It works :) Variable `a` is an instance of NOD and accurately returns true or false depending on the presence of form_data.

yakschuss: I don't really see a reason to use a `proc` here, though. Why don't we just call `form_data` on self?

`form_data.present?`

modules/appeals_api/config/schemas/10182.json
ElleeB: Description exists: `Address of the Veteran - not required if Veteran is homeless. Cannot exceed 165 characters when all fields are concatenated.`

PR #5872, opened 2021-02-05
Handle VA appointment errors that include a partial response

kreek:
## Description of change
The underlying VAOS VistA Scheduling Service can throw an error but have VAOS still return an empty result. Adds a VCR cassette and a spec for this case.

## Original issue(s)
N/A


jperk51 (approved):
LGTM

PR #5873, opened 2021-02-05
Add mhv correlation ID to IAM user/identity

patrickvinograd:
## Description of change
Adds  mhv correlation ID parsed  from IAM OAuth token into IAM user/useridentity models, to  enable API access to MHV upstream data. 

A couple of TODOs but this should enable  testing of secure messaging API access for mobile. 

## Original issue(s)


app/models/iam_user_identity.rb
```diff
@@ -29,6 +29,7 @@ def self.build_from_iam_profile(iam_profile)
       iam_icn: iam_profile[:fediam_mviicn],
       iam_edipi: iam_profile[:fediam_do_dedipn_id],
       iam_sec_id: iam_profile[:fediamsecid],
+      iam_mhv_id: valid_mhv_id(iam_profile[:fediam_mhvien]),
```
jjulian, line 32:
```suggestion
      iam_mhv_id: safe_mhv_id(iam_profile[:fediam_mhvien]),
```

jjulian:
fyi rubocop not happy with your trailing whitespace

app/models/iam_user_identity.rb
```diff
@@ -52,4 +53,14 @@ def set_expire
   def uuid
     Digest::UUID.uuid_v5(@iam_sec_id, @iam_icn)
   end
+
+  # Return a single mhv id from a possible comma-separated list value attribute
+  def self.valid_mhv_id(id_from_profile)
+    # TODO: site login fails if multiple _different_ MHV IDs are present
+    # Should this code be doing the same? Or at least logging a warning?
```
kreek, line 60: Some questions about this.

-  What kind of issues does this cause, would their messages potentially diverge between accounts? 
- How does a veteran resolve having multiple MHV ids?

If the issues doesn't cause data issues, I think logging would be sufficient.

kreek (approved):
One question, but overall LGTM

app/models/iam_user_identity.rb
patrickvinograd: Yes, they'd potentially have different message contents between the two accounts.

MHV website  and VA.gov (as of the SSOe integration) prevent a user in such state from signing  in,  and presents a  message instructing them to call the help desk to resolve.

kreek: 👍 ok, sounds like we should do the same thing as .gov and raise an error (and map that error to a helpful in-app message)

patrickvinograd:
This still needs specs, and has gotten a little unwieldy. My plan is to  split it out into 3  PRs:
1) Adding the MHV ID to IAMUserIdentity
2) Adding the module-specific SM client  and configuration
3) All the controllers/serializers.

patrickvinograd:
Obsoleted by subsequent PRs

PR #5875, opened 2021-02-08
Ensure array gets returned when user has only one dependent

digitaldrk:
## Description of change
When a user views their dependents (/view-change-dependents/view/) and has multiple dependents, the FE gets an array of 'person' javascript objects (hashes).

HOWEVER

When a user views their dependents and has one dependent, the FE gets a single javascript object (hash).

In an effort to maintain a consistent interface for the FE, we are checking the type for `persons`. If it's a hash, we wrap it in an array

If we don't do this, the FE will think that there are no dependents and display that which is misleading to the user.

## Before
Screen Shot 2021-02-08 at 9 52 50 AM

## After
Screen Shot 2021-02-08 at 9 53 13 AM

## Original issue(s)
department-of-veterans-affairs/va.gov-team#19090


thilton-oddball (approved):
Nice change!

PR #5878, opened 2021-02-08
18777 mobile claims multiple img upload

jperk51:
## Description of change
Add functionality to mobile doc upload endpoint to accept multiple images and consolidate them into one pdf for upload

## Original issue(s)
department-of-veterans-affairs/va.gov-team#18777

Docs updated and new tests added


modules/mobile/app/controllers/mobile/v0/claims_and_appeals_controller.rb
```diff
@@ -98,6 +108,19 @@ def upload_documents
 
       private
 
+      def generate_multi_image_pdf(image_list)
```
kreek, line 111: My initial thought with PDF generation was that it would also be in a background job. But if it's not too intensive it makes sense for generation to succeed then return a 200 to the user and upload in the background (or fail and let the user know). 

What is collective size limit for images? Have you tried generating a PDF at that limit and if so any metrics on how long it takes?

modules/mobile/app/controllers/mobile/v0/claims_and_appeals_controller.rb
```diff
@@ -98,6 +108,19 @@ def upload_documents
...
+        Prawn::Document.generate(TMP_PDF_PATH) do |pdf|
+          image_list.each do |img|
+            File.open(TMP_IMG_PATH, 'wb') { |f| f.write Base64.decode64(img) }
+            pdf.image TMP_IMG_PATH, fit: [pdf.bounds.right, pdf.bounds.top]
```
kreek, line 115: does `fit` only scale if it doesn't? does it do so proportionally?

modules/mobile/app/controllers/mobile/v0/claims_and_appeals_controller.rb
jperk51: I'll try that, the limit is 50 MB I believe

modules/mobile/app/controllers/mobile/v0/claims_and_appeals_controller.rb
jperk51: Unsure about the first question, I'll test that. As for the second question, from the Prawn docs, `:fit	
scale the dimensions of the image proportionally to fit inside [width,height]`

modules/mobile/app/controllers/mobile/v0/claims_and_appeals_controller.rb
kreek: Could be good to have some statsd metrics to see how long things take on staging/prod.

modules/mobile/app/controllers/mobile/v0/claims_and_appeals_controller.rb
```diff
@@ -5,12 +5,21 @@
...
 require_relative '../../../models/mobile/v0/claim_overview'
 require 'sentry_logging'
+require 'prawn'
 
 module Mobile
   module V0
     class ClaimsAndAppealsController < ApplicationController
       include IgnoreNotFound
+      TMP_BASE_PATH = Rails.root.join 'tmp', 'uploads', 'cache'
+      TMP_IMG_PATH = "#{TMP_BASE_PATH}tempFile.jpg"
```
kreek, line 15: Will there be a collision if two user requests are writing at the same time?

jperk51: Good call, I would guess that would cause an issue. Would a single user ever be able to write to it twice at the same time? I could append the user id if they can't. Any other suggestions to get around this issue?

kreek: Don't think a user would be able to write a the same time. But to be extra sure you could use a uuid (`SecureRandom.uuid`) for the file. Might want to create a directory for the user and just blow that away at the end, or on error, rather than have to clean up each file.

jperk51: Nice, that's a great idea. Thanks

modules/mobile/app/controllers/mobile/v0/claims_and_appeals_controller.rb
jperk51: Fit does scale up images that are smaller than the pdf bounds along with scaling down images that are larger.

modules/mobile/app/controllers/mobile/v0/claims_and_appeals_controller.rb
jperk51: I tested a 50MB image which would take 6-8 seconds to finish. I tried smaller individual images (2 25MBs, 5 10MBs, etc), but the smaller each individual image was the faster the request finished. Also added statsd to record the time it takes so we can check it in upper envs.

kreek: I think because of the upper bounds and phone cameras outputting larger and larger images that it could make send to move the PDF processing to a background job as well. If the images are valid and are saved successfully we could return a response to the user. This can be another ticket as staging testing will give us more data about processing times in a more production like env.

modules/mobile/app/controllers/mobile/v0/claims_and_appeals_controller.rb
```diff
@@ -5,12 +5,24 @@
...
+require 'prawn'
+require 'fileutils'
+require 'mini_magick'
 
 module Mobile
   module V0
     class ClaimsAndAppealsController < ApplicationController
       include IgnoreNotFound
+      UUID = SecureRandom.uuid
+      TMP_BASE_PATH = Rails.root.join 'tmp', 'uploads', 'cache', UUID
```
kreek, line 17: Although the random UUID is added this is still constant for the duration of the the app run on the instance of the server. Could that part of the directory path creation move into the `upload_documents` method? You can set it with a class instance variable `@variable_name` that way the `after_action` hook will have it in scope.

jperk51: Gotcha, thanks

kreek (approved):
LGTM 🚀

thilton-oddball (approved):
LGTM

PR #5879, opened 2021-02-08
add person_types to SSOe parsing

jimiray:
## Description of change

Add parsing person_type from SSOe for use in User roles.

## Original issue(s)

https://app.zenhub.com/workspaces/vsp-identity-5f5bab705a94c9001ba33734/issues/department-of-veterans-affairs/va.gov-team/19600

### DIFF
--- a/spec/lib/saml/ssoe_user_spec.rb
+++ b/spec/lib/saml/ssoe_user_spec.rb
@@ -118,7 +118,8 @@ RSpec.describe SAML::User do
           participant_id: nil,
           birls_id: nil,
           icn: nil,
-          common_name: nil
+          common_name: nil,
+          person_types: nil
         )
       end
 
@@ -157,7 +158,8 @@ RSpec.describe SAML::User do
           participant_id: nil,
           birls_id: nil,
           icn: nil,
-          common_name: nil
+          common_name: nil,
+          person_types: nil
         )
       end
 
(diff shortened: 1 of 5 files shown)


spec/lib/saml/ssoe_user_spec.rb
```diff
@@ -758,7 +768,8 @@
           birls_id: '796123607',
           icn: '1012740600V714187',
           multifactor: multifactor,
-          common_name: 'dslogon10923109@gmail.com'
+          common_name: 'dslogon10923109@gmail.com',
+          person_types: ['PAT']
```
joeniquette, line 772: Do we want a spec for multiple person types?

bosawt (commented):
Would it make sense to also parse from the MPI response and then have a function in `user.rb` like: 
`def person_types
identity&.person_types || mpi&.person_types
end`
If you feel like it doesn't fit in this PR, then I'm fine with approving what's currently here

spec/lib/saml/ssoe_user_spec.rb
bosawt: Oh yeah, might be worth adding a spec for multiple person types to test the parsing logic

jimiray: added multiple person types to the same test and checked to ensure that it comes back as an array of the names.

jimiray:
> Would it make sense to also parse from the MPI response and then have a function in `user.rb` like:
> `def person_types identity&.person_types || mpi&.person_types end`
> If you feel like it doesn't fit in this PR, then I'm fine with approving what's currently here

I thought this one would be a good launching point and we can separate the changes.

thilton-oddball (approved):
Looks good!

bramleyjl (approved):
LGTM

PR #5880, opened 2021-02-08
15091 sunset v0 facilities va controller

darkbushido:
## Description of change
This removes the `/v0/facilities/va` route and the following related code:

- v0 VaController
- related swagger docs
- related specs

## Original issue(s)
department-of-veterans-affairs/va.gov-team#15091

 * Just removing old code that shouldn't be used anymore.


spec/models/facilities_query/ids_query_spec.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe FacilitiesQuery::IdsQuery do
+  # NOTE: this spec exisits ONLY to raise simple_cov coverage
+  # I'm not even sure that I'm adding the ids params correctly...
+  it 'can increase simple_cov coverage' do
```
ericboehs, line 8: Was this required to get CI to pass?

ericboehs (approved):
Let me know when you're ready to merge and I can let the Dangerbot check through.

Also, I don't like the test that is there just to increase test coverage. Wish there was a better way. 🤔

darkbushido:
That code is only used by the stuff that's being removed in this PR. I'm not happy with it either but the other class has the exact same test. In another PR all this should be getting cleaned up as well.

PR #5881, opened 2021-02-08
Fix DetailedSchemaErrors when running individually

cilestin:
## Description of change
Spec would fail when run on its own, but pass when in a collection. Using `rails_helper` instead of `spec_helper` included the missing pieces needed to run the spec file on its own.


ElleeB (approved):
Solid work

PR #5887, opened 2021-02-08
Bug fix provider info params

ajmagdub:
## Description of change
Found Sentry NoMethodError, http://sentry.vfs.va.gov/organizations/vsp/issues/14776/?project=3&query=is%3Aunresolved+ppms&statsPeriod=14d . 

## Original issue(s)
No ticket created


lib/facilities/ppms/v1/client.rb
```diff
@@ -159,7 +159,8 @@ def pos_locator_params(params, pos_code)
         end
 
         def provider_locator_params(params)
-          specialty_codes = params[:specialties].first(4).map.with_index.with_object({}) do |(code, index), hsh|
+          specialties = Array.wrap(params[:specialties])
+          specialty_codes = specialties.first(4).map.with_index.with_object({}) do |(code, index), hsh|
```
thilton-oddball, line 163: I can confirm that your change fixes the specific issue of `undefined method `first' for nil:NilClass`. However, I am unable to determine if there might end up being additional errors further down stream (i.e. there may be a deeper root to this problem) so it may be worth keeping an eye on.

thilton-oddball (approved):
Left a comment but looks good.

lib/facilities/ppms/v1/client.rb
ajmagdub: Thanks, will keep an eye on it

PR #5889, opened 2021-02-08
Add ch31 upload to VBMS

digitaldrk:
## Description of change
We need to add PDF generation and VBMS submission to the CH31 form, that's what this PR does 🥳 .

The CH31 form is in the final stages of development with testing happening in staging. In other words, this form is not accessible in production yet.

Once we had the uploader wired up, we realized that we needed to update the PDF fill code. When we originally coded the PDF fill, we were getting back snake_case form params. We had our FE team change from form params to the standard JSON string w/ camel case in between when the PDF fill code was merged and now. So the PDF fill code had to be updated to get in line with cameCase params.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#19088

  * No changes
  * No FF added here but the form is not accessible to production users
  * No sentry logging added
  * No metrics added
  * No swagger docs updated
  * concerns


thilton-oddball (approved):
All looks good

PR #5890, opened 2021-02-08
Apps API faraday changes

bradenshipley:
## Description of change
Since the AppsApi now requires an `apikey` for all routes due to kong gateway changes, we must allow for certain routes to be available without an apikey. Specifically the read only methods inside the `DirectoryController` : `#index`, `#show`, and `#scopes`. This PR adds a proxy to the GET methods of the `DirectoryController` in the same way that this problem was solved by the `forms` module previously.

The `AppsController` added here uses JSON validation for all routes listed above, with only one exception being that the `#scopes` method returns 204 when a valid `service_category` is not passed and skips the json validation.

## Original issue(s)
https://vajira.max.gov/secure/RapidBoard.jspa?rapidView=777&view=detail&selectedIssue=API-4863

There is dummy data added to `settings.yml` and `test.yml`.


spec/spec_helper.rb
```diff
@@ -36,7 +38,6 @@
     add_filter 'modules/appeals_api/app/swagger'
     add_filter 'modules/apps_api/app/controllers/apps_api/docs/v0/api_controller.rb'
     add_filter 'modules/apps_api/app/swagger'
-    add_filter 'modules/apps_api/lib/apps_api/directory_application_creator.rb'
```
bradenshipley, line 39: This was removed as the file no longer exists.

config/routes.rb
```diff
@@ -35,6 +35,9 @@
 
     resources :preferred_facilities, only: %i[index create destroy]
 
+    resources :apps, only: %i[index show]
+    get 'apps/scopes/:category', to: 'apps#scopes', defaults: { category: 'unknown_category' }
+
```
bradenshipley, line 40: this explicit routing is necessary so that `apps/scopes` isnt interpreted as a call to `#show`

modules/apps_api/app/controllers/apps_api/v0/directory_controller.rb
```diff
@@ -55,13 +55,12 @@ def destroy
...
+        if parsed_scopes.any?
           render json: {
             data: parsed_scopes
           }
-           rescue
-             head :no_content
-        end
+        else
+          head :no_content
         end
```
bradenshipley, line 64: This rescue statement was confusing to read, so it was changed

lib/apps/client.rb
```diff
@@ -0,0 +1,83 @@
...
+        Apps::Responses::Response.new(raw_response.status, raw_response.body, 'apps')
+      end
+    rescue => e
+      handle_error(e)
+    end
+
+    # Get an individual app
+    #
+    def get_app
+      with_monitoring do
```
bradenshipley, line 39: this is put here for `app_name`s that have whitespace in them, like `Apple Health`.

lib/apps/responses/response.rb
```diff
@@ -0,0 +1,38 @@
...
+      attribute :body, String
+      attribute :status, Integer
+
+      def initialize(status, body, schema_name)
+        self.body = if status == 204
+                      nil
+                    else
+                      json_format_is_valid?(body, schema_name) ? body : {}
+                    end
+        self.status = status
```
bradenshipley, line 27: I am open to recommendations on how to refactor this if statement here. The only time it would not need to validate the json is when the head is a `:no_content`. Using a ternary statement versus a nested if seemed like the better of two bad options on line 25.

modules/apps_api/app/controllers/apps_api/v0/directory_controller.rb
```diff
@@ -2,7 +2,7 @@
 
 # dependencies
 require_dependency 'apps_api/application_controller'
-require 'okta/directory_service.rb'
+require 'okta/directory_service'
 module AppsApi
```
bradenshipley, line 6: rubocop changes

modules/apps_api/spec/requests/v0/application_directory_request_spec.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 require 'rails_helper'
-require_relative '../../../app/controllers/apps_api/v0/directory_controller.rb'
+require_relative '../../../app/controllers/apps_api/v0/directory_controller'
```
bradenshipley, line 5: rubocop changes

thilton-oddball (approved):
All looks good

PR #5893, opened 2021-02-09
add code for engine factory initializer

LindseySaari:
## Description of change
It was brought up in the #vsp-tools-be channel that some initialization code needs to be added to the module generator engine template to load factories automatically when running specs. This PR adds the code and the changes are confirmed as working when invoking the generator. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#19647


thilton-oddball (approved):
Looks good! Nice work!

PR #5894, opened 2021-02-09
test_user_dashboard checkout & checkin accounts

bramleyjl:
## Description of change

This change adds checkout/checking functionality for TUD accounts as well as an `AfterLogoutJob` worker modeled on the existing `AfterLoginJob`. In non-production environments the login/logout workers will attempt to match the user account's uuid to a test account and if a match is found mark the account as available or unavailable to prevent collisions between two users attempting to use the same test user account.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/17221
https://github.com/department-of-veterans-affairs/va.gov-team/issues/17223

Unit tests have been added for the new test_user_dashboard services that manage checkout/checkin functionality as well as the availability checking method.


jimiray (approved):
LGTM

app/controllers/concerns/authentication_and_sso_concerns.rb
```diff
@@ -59,6 +59,7 @@ def clear_session
 
   # Destroys the users session in 1) Redis, 2) the MHV SSO Cookie, 3) and the Session Cookie
   def reset_session
+    AfterLogoutJob.perform_async(account_uuid: @current_user&.account_uuid)
```
bosawt, line 62: Should we put the `unless Rails.env.production?` on this line? Shouldn't be a big deal since they'll just cancel out, but we'll queue up a lot of Sidekiq jobs on prod that won't do anything

modules/test_user_dashboard/spec/serializers/tud_account_serializer_spec.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
```
bosawt, line 1: Always love it when I see such established specs

app/controllers/concerns/authentication_and_sso_concerns.rb
bramleyjl: The thinking was that we should make the new `AfterLogoutJob` worker mirror the existing `AfterLoginJob` as closely as possible and put the environment check within it to allow for future logic to be placed within the worker; I suppose since it just has the TUD function right now we should move it and leave it to whoever ends up adding more functionality to the `AfterLogoutJob` in the future to put it back where it is now.

thilton-oddball (approved):
Looks good

PR #5896, opened 2021-02-09
Update process10203_submissions.rb

zurbergram:
## Description of change

- Update to pull most recently stored `auth_headers`.
- Move check that prevents previously processed submissions down to prevent state being changed after initial processing if EVSS is unaccessible with store `auth_headers`
- Remove retry for sidekiq job as this job runs once an hour from 6-18

## Original issue(s)
department-of-veterans-affairs/va.gov-team#19658


shawkey-daniel (approved):
LGTM

app/workers/education_form/process10203_submissions.rb
```diff
@@ -29,7 +27,7 @@ def perform(
         saved_claims: {
           form_id: '22-10203'
         }
-      )
+      ).order('created_at')
```
thilton-oddball, line 30: Since you are joining multiple tables, this 'created_at' that you are wanting to order by is ambiguous and is causing the CI check to fail. I think if you specify the specific table like` .order('TABLE_NAME.created_at')` it should work.

thilton-oddball (changes requested):

app/workers/education_form/process10203_submissions.rb
zurbergram: updated

thilton-oddball (approved):
LGTM

PR #5898, opened 2021-02-09
18732 - removed message text length validation, and removed unnecessary tests

grossb1:
## Description of change

We removed validation for `message_text` in `appointment_request` model and instead made it a required field.

## Original issue(s)

Tests were updated to reflect this change


thilton-oddball (approved):
LGTM

PR #5899, opened 2021-02-09
[19411] Add compensation & pension, and education benefit amounts to 5655 form prefill

camerontesterman:
## Description of change
Adds Compensation & Pension and Education benefit amounts to the Financial Status Report (5655) form prefill.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#19411

Tested and verified locally.


lib/debt_management_center/payments_service.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module DebtManagementCenter
+  class Payments
```
thilton-oddball, line 4: Does the class need to be PaymentsService instead of Payments?

camerontesterman: Yes, good catch. Fixed

thilton-oddball (approved):
Looks good!

PR #5901, opened 2021-02-10
Update save in progress

digitaldrk:
## Description of change
I'm particularly interested in y'all's feedback on this one. The caveat we have with this form is it  _can_ be 2 forms from the VA's perspective, the "686" and the "674".

Before this PR, we were clearing save in progress data (SIP) upon successful submission _to vets-api_. This was suboptimal because given the 686/674 form makes _SO MANY_ requests to complete its mission, an error might occur and the user would have to start all over again. 

This PR makes a good faith effort to maintain SIP data in the event of a failure to submit a 686/674

## Original issue(s)
department-of-veterans-affairs/va.gov-team#14538


ericboehs (approved):
This seems like a fine approach. The only thing I might have done different is instead of inheriting from a class to inherit the methods, I'd use a module and include it.

PR #5904, opened 2021-02-10
[API-4737] - Add missing statuses to "ClaimsShow" Swagger schema

jw81:
## Description of change

Adds the following statuses to the `ClaimsShow` Swagger schema ::
- `pending`
- `errored`

## Original issue(s)
[API-4737](https://vajira.max.gov/browse/API-4737)

The JSON output from http://localhost:3000/services/claims/docs/v1/api will now include the two new statuses.

### DIFF
--- a/modules/claims_api/app/swagger/claims_api/claims/claims_response_swagger.rb
+++ b/modules/claims_api/app/swagger/claims_api/claims/claims_response_swagger.rb
@@ -299,10 +299,12 @@ module ClaimsApi
               key :description, 'Current status of the claim'
               key :enum, [
                 '"Claim received"',
+                '"Complete"',
                 '"Initial review"',
+                '"Errored"',
                 '"Evidence gathering, review, and decision"',
-                '"Preparation for notification"',
-                '"Complete"'
+                '"Pending"',
+                '"Preparation for notification"'
               ]
             end
           end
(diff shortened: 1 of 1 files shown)


modules/claims_api/app/swagger/claims_api/claims/claims_response_swagger.rb
```diff
@@ -299,10 +299,12 @@ class ClaimsResponseSwagger
               key :description, 'Current status of the claim'
               key :enum, [
                 '"Claim received"',
+                '"Complete"',
                 '"Initial review"',
+                '"Errored"',
                 '"Evidence gathering, review, and decision"',
-                '"Preparation for notification"',
-                '"Complete"'
+                '"Pending"',
```
SlakrHakr, line 306: these new values should *not* be capitalized. This collection of values is a combination of statuses we have in addition to values we get back from EVSS. Unfortunately we don't capitalize our statuses but EVSS does :(

jw81: Ahhhhh, ok.
Updated [here](https://github.com/department-of-veterans-affairs/vets-api/pull/5904/commits/1bbdc8a7a65bed50126677fa3c13b27660964296).

SlakrHakr (approved):

PR #5907, opened 2021-02-10
Update 686c Manual Status for Unsupported Workflows

kathleencrawford:
## Description of change

During 686c (Dependents Application) partner integration testing, BGS asked us to manually set the state type code for certain unsupported workflows.  These include death of a child, death of a parent, and reporting that a child has stopped attending school.  This PR checks if one of these workflows are being submitted, and if so, sets the state type code to `MANUAL_VAGOV`.

It also comments out one line that was skipping a function for a particular workflow.  I've left the line commented out for now, in case we need it back after more integration testing with BGS.  

## Original issue(s)
department-of-veterans-affairs/va.gov-team#19823
department-of-veterans-affairs/va.gov-team#19801

This is an update to an existing feature, so no new logging or settings have been added.

## Testing
- [x] Local testing
- [x] Specs updated
- [x] Staging testing planned


thilton-oddball (approved):
LGTM

PR #5908, opened 2021-02-10
14839 mobile cancel appointment fixes

kreek:
## Description of change

- Appointments have ids but they are not used for cancellation. Instead the (local) time, clinic id, facility id, and service are used to find the id to cancel. To avoid having this complexity bubble up to the mobile app we base 64 encode those values into a cancel id.
- Removes facility id and clinic id from the appointment response as were only needed for cancelling which is now handled by the cancel id.
- Staging facilities are mocked and their ids do not match the real facilities that the staging facilities api uses. They're toggled when showing the appointment list but also need to be toggled back when cancelling appointments in staging. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#14839


modules/mobile/app/services/mobile/v0/appointments/proxy.rb
```diff
@@ -37,18 +37,20 @@ def get_appointments(start_date, end_date)
...
+          facility_id = Mobile::V0::Appointment.toggle_non_prod_id!(params[:facilityId])
+
+          cancel_reasons = get_facility_cancel_reasons(facility_id)
           cancel_reason = extract_valid_reason(cancel_reasons.map(&:number))
 
           raise Common::Exceptions::BackendServiceException, 'MOBL_404_cancel_reason_not_found' if cancel_reason.nil?
 
           put_params = {
-            appointment_time: DateTime.parse(params[:appointmentTime]).strftime('%m/%d/%Y %H:%M:%S'),
+            appointment_time: params[:appointmentTime].strftime('%m/%d/%Y %H:%M:%S'),
```
thilton-oddball, line 48: Will params[:appointmentTime] just be a string? I'm guessing with the removal of DateTime.parse it won't be but just wanted to double check.

modules/mobile/app/models/mobile/v0/adapters/va_appointments.rb
```diff
@@ -77,23 +77,39 @@ def parse(appointments)
...
             comment: comment(details, type),
-            clinic_id: appointment_hash[:clinic_id],
             facility_id: facility_id,
             healthcare_service: healthcare_service(details, type),
             location: location(details, type, facility_id),
             minutes_duration: minutes_duration(details, type),
-            start_date_local: start_date_utc.in_time_zone(time_zone),
+            start_date_local: start_date_local,
             start_date_utc: start_date_utc,
             status: status(details, type, start_date_utc),
```
thilton-oddball, line 114: Should this just be `status: status,` since status is set above on (the new) line 92?

thilton-oddball (approved):
A couple of comments to review but otherwise looks good.

modules/mobile/app/services/mobile/v0/appointments/proxy.rb
kreek: ya it's `DateTime` now, it get's parsed further up the chain

modules/mobile/app/models/mobile/v0/adapters/va_appointments.rb
kreek: 👍 good catch, I'll use the status that was set

jperk51 (approved):
LGTM

PR #5909, opened 2021-02-11
Update NOD statuses

nathanbwright:
## Description of change
This PR does several things:
- Updates the swagger docs to only show statuses that we actually use.
- It moves the logic for updating NOD status to the model itself.
- Updates the HLR and NOD pdf submit job specs to be a little more robust.

We still need to update the docs to describe each of the statuses and how they relate to each other.

## Original issue(s)


nathanbwright:
Realized there is still some more research that I need to do with this before it's ready for review.

PR #5911, opened 2021-02-11
 longer timeout for evss docs

annaswims:
## Description of change
Increase the timeout for EVSS docs.  We are seeing successful EVSS requests take > 200s for a 50 mb PDF (stats from fwdproxy) but because rails times out at 180 we're considering the job failed and re-trying, which creates duplicate uploads. We're just about to triple the file upload size limit to 150 mb so I'm allowing even more time before a timeout.  

These uploads do not need to be done while the user waits, so they are low-priority. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#19680


thilton-oddball (approved):
LGTM

PR #5912, opened 2021-02-11
Save Complete Lighthouse Response When It Fails to Validate Against Schema (DecisionReview::Service)

edmangimelli:
[A contestable_issues GET request failed](http://sentry.vfs.va.gov/organizations/vsp/issues/19829/?environment=production&project=3&query=is%3Aunresolved+ContestableIssues&statsPeriod=14d) (threw an exception) because the [response from Lighthouse](https://github.com/department-of-veterans-affairs/vets-api/blob/8cab530812d1af224fc8330d1d06755be1f0d668/lib/decision_review/service.rb#L69) didn't match the [schema we were expecting](https://github.com/department-of-veterans-affairs/vets-json-schema/blob/b6f33f49e6d476d8eb348c68efda7d9d2c654eb9/src/schemas/HLR-get-contestable-issues-response-200/schema.json). We logged enough about the error to sentry and the PersonalInformationLog to get an idea of why the response didn't match the schema (i can see what part of the schema failed), **but we need to log the complete response to pin down exactly what went wrong.**

## Acceptance Criteria
- [x] when a Lighthouse response doesn't validate against our schema (DecisionReview::Service#validate_against_schema), log the entire response to the PersonalInformationLog so we can figure out exactly why the response was invalid (according to our schema).


app/controllers/concerns/failed_request_loggable.rb
```diff
@@ -5,7 +5,7 @@ module FailedRequestLoggable
 
   class_methods do
     def exception_hash(exception)
-      %i[message backtrace key response_values original_status original_body]
+      %i[message backtrace key response_values original_status original_body errors status_code sentry_type]
```
edmangimelli, line 8: adding methods that `Common::Exceptions::SchemaValidationErrors` have

thilton-oddball (approved):
LGTM

PR #5916, opened 2021-02-11
Refactor zipCode5 for zeros if homeless and no address

ElleeB:
## Description of change
If a veteran is experiencing homelessness, including an address is optional. However, CentralMail requires a 5-digit zip code from our request. We've refactored so that `zipCode5 => 00000` is sent via the request in cases of homelessness with no address. 

## Original issue(s)
https://vajira.max.gov/browse/API-4832

## Testing
`zipCode5 formatting` tested in model spec
`address presence if not experiencing homelessness` tested in controller spec

### DIFF
--- a/lib/common/exceptions/detailed_schema_errors.rb
+++ b/lib/common/exceptions/detailed_schema_errors.rb
@@ -92,8 +92,7 @@ module Common
       end
 
       def schema(_error)
-        data = i18n_interpolated :schema
-        data
+        i18n_interpolated :schema
       end
 
       def array_items(error)
--- a/modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
+++ b/modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
@@ -97,7 +97,8 @@ module AppealsApi
     end
 
     def zip_code_5
-      form_data&.dig('data', 'attributes', 'veteran', 'address', 'zipCode5')
+      # schema already validated address presence if not homeless
+      veteran_contact_info&.dig('address', 'zipCode5') || '00000'
     end
 
     def lob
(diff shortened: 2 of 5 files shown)


lib/common/exceptions/detailed_schema_errors.rb
```diff
@@ -92,8 +92,7 @@ def required(error)
       end
 
       def schema(_error)
-        data = i18n_interpolated :schema
-        data
+        i18n_interpolated :schema
```
ElleeB, line 95: saw it had to do it

modules/appeals_api/config/schemas/10182.json
```diff
@@ -40,6 +40,7 @@
       "required": [ "boardReviewOption", "socOptIn" ]
     },
 
+
```
ElleeB, line 43: because all of the others have a double space between them

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
```diff
@@ -97,7 +97,9 @@ def hearing_type_preference
     end
 
     def zip_code_5
-      form_data&.dig('data', 'attributes', 'veteran', 'address', 'zipCode5')
+      # schema already validated address presence if not homeless
+      address = form_data&.dig('data', 'attributes', 'veteran', 'address')
+      address ? address['zipCode5'] : '00000'
```
cilestin, line 102:
```
veteran_contact_info&.dig('address', 'zipCode5') || '00000'
```
seems cleaner to me

cilestin (changes requested):

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
ElleeB: yes

cilestin (approved):

PR #5918, opened 2021-02-11
adds email uniqueness validation to tud_accounts

bramleyjl:
makes email a unique field for `tud_accounts` to prevent accidental duplicate account creation


thilton-oddball (approved):
LGTM

PR #5919, opened 2021-02-11
Exclude Irrelevant Methods From `exception_hash` (FailedRequestLoggable)

edmangimelli:
If a `nil` is included in `exception_hash`, I want to know that that's because `error.method` returned `nil`, not that `try` failed.

### Acceptance Criteria
- [x] For the list of methods in `exception_hash`, filter out those that the current exception doesn't know how to respond to.


thilton-oddball (approved):
Looks good

PR #5920, opened 2021-02-11
Add "HLR" to `error_class` When Logging to PersonalInformationLog

edmangimelli:
just so it's a little easier to query for these


thilton-oddball (approved):
LGTM

PR #5922, opened 2021-02-12
Allow Specifying a Date Range For BDD Stats Mode (form526 rake task)

edmangimelli:
Running BDD stats is very slow as it must decrypt many submissions.

### Acceptance Criteria

- [x] Allow specifying a date range when pulling BDD stats, allowing someone to pull stats only for the dates they need. *(I end up running this once or twice a week, and I keep the stats for all the previous months. This way I can pull stats just for the days I don't have.)*


thilton-oddball (approved):
Looks good!

PR #5926, opened 2021-02-12
Questionnaire Manager must return aggregated data

dillo:
## Description of change

- Questionnaire Manager now returns an aggregated set of data when the `/health_quest/v0/questionnaire_manager` route is invoked.
- The `transformer.rb` class is a bit heavy on the data structure and algorithms implementation and is intentionally imperative in style and design. We can think about abstracting some of the code away into newer classes if necessary but I feel that having the computation heavy pieces in one area will make the code easier to read and understand.
## Original issue(s)
department-of-veterans-affairs/va.gov-team#19015

- [x] RSpec tests have been written and are passing.


thilton-oddball (approved):
All looks good. Temporarily disabling LOC check based on long specs.

PR #5929, opened 2021-02-12
Updates IAM user model with MHV id, LOA fix

patrickvinograd:
Adds MHV identifier to IAM user identity payload, for use in invoking MHV APIs. 

Also fixes LOA interpretation to match VA.gov's, so that DSLogon or MHV assurance level 2 will be interpreted as LOA3 for authorization purposes.

## Original issue(s)

### DIFF
--- a/app/models/iam_user_identity.rb
+++ b/app/models/iam_user_identity.rb
@@ -66,7 +66,7 @@ class IAMUserIdentity < ::UserIdentity
 
   # Return a single mhv id from a possible comma-separated list value attribute
   def self.valid_mhv_id(id_from_profile)
-    # TODO: For now, log instances of duplicate MHV ID. 
+    # TODO: For now, log instances of duplicate MHV ID.
     # See issue #19971 for consideration of whether to reject access
     # to features using this identifier if this happens.
     mhv_ids = (id_from_profile == 'NOT_FOUND' ? nil : id_from_profile)
--- a/app/models/iam_user_identity.rb
+++ b/app/models/iam_user_identity.rb
@@ -1,10 +1,14 @@
 # frozen_string_literal: true
 
+require 'sentry_logging'
+
 # Subclasses the `UserIdentity` model. Adds a unique redis namespace for IAM user identities.
 # Like the it's base model it acts as an adapter for the attributes from the IAMSSOeOAuth::Service's
 # introspect endpoint.Adds IAM sourced versions of ICN, EDIPI, and SEC ID to pass to the IAMUser model.
 #
 class IAMUserIdentity < ::UserIdentity
+  extend SentryLogging
+
   PREMIUM_LOAS = [2, 3].freeze
   UPGRADE_AUTH_TYPES = %w[DSL MHV].freeze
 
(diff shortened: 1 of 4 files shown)


patrickvinograd (commented):
Alastair's additions LGTM

kreek (approved):
LGTM 🚀

app/models/iam_user_identity.rb
```diff
@@ -5,6 +5,9 @@
 # introspect endpoint.Adds IAM sourced versions of ICN, EDIPI, and SEC ID to pass to the IAMUser model.
 #
 class IAMUserIdentity < ::UserIdentity
+  PREMIUM_LOAS = [2, 3].freeze
+  UPGRADE_AUTH_TYPES = %w[DSL MHV].freeze
```
jjulian, line 9: Can these values ever come to us in lower or mixed case?

app/models/iam_user_identity.rb
```diff
@@ -52,4 +59,14 @@ def set_expire
   def uuid
     Digest::UUID.uuid_v5(@iam_sec_id, @iam_icn)
   end
+
+  # Return a single mhv id from a possible comma-separated list value attribute
+  def self.valid_mhv_id(id_from_profile)
+    # TODO: site login fails if multiple _different_ MHV IDs are present
+    # Should this code be doing the same? Or at least logging a warning?
+    # see lib/saml/user_attributes/ssoe.rb for example validation
```
jjulian, line 67: I feel like we should resolve these comments before merging.

jjulian (approved):
Looks good, just two notes

app/models/iam_user_identity.rb
patrickvinograd: It's not clear  what the right thing to do is. On VA.gov, when this  occurs we prevent sign-in, and we have the opportunity to present a clear  error  message during the sign-in flow. 

Here, IAM is going to issue the access token successfully. We can prevent this API request, but we could be instantiating this user identity object for any of the OAuth-based API endpoints - it's not like there's a first API  request  that always happens after token issuance. That means any part of the app that makes an API request needs to be prepared to handle getting an error back and presenting something meaningful to the user. And we need to define what that error is.

Would you be okay if we filed a launch-blocker ticket and revisited it, in the interest of  getting the MHV integration moving  along?

jjulian: That sounds good, thanks Patrick. For the sake of this PR, let's log a warning and at least we have some visibility into how often this occurs.

lexicalninja: Do we want to just log a warning or create a Crashlytics event for this?

lexicalninja: I guess probably not crashlytics here, but is there a way to make it easy to pull a report on this?

patrickvinograd: Two other mitigation options:

1) Get the IAM team to alter their logic such that they don't issue an access token if they detect multiple MHV IDs. But what if the  second  MHV ID is added _after_ the refresh token is issued? Complex.

2) Just gate access to MHV features such as SM - i.e. remove the MHV ID from the session, and  thus secure messaging will disappear from the authorizedServices list. There's no reason to gate access to claims becaause of a duplicate  MHV ID, for example.  Might need a "why can't I  access?" UX  consideration though.

patrickvinograd: I filed this issue for follow-up: https://github.com/department-of-veterans-affairs/va.gov-team/issues/19971

patrickvinograd: We can also log a sentry event for it for visibility.

thilton-oddball (approved):
Looks good!

app/models/iam_user_identity.rb
kreek: I'd need to double check with the IAM team, but I don't think so

patrickvinograd: Anything's possible but I really doubt it.

PR #5932, opened 2021-02-13
Issuer Non-Dependent JWKs endpoint

lukemarkwordtlibertyits:
## Description of change

Okta JWKs endpoint should be issuer agnostic.

## Original issue(s)
https://vajira.max.gov/browse/API-4992

Currently, the okta_service keys method reaches out to the issuer's metadata endpoint to find the jwks endpoint. The issuer's metadata endpoint is assumed to be at `/.well-known/openid-configuration`. IAM integration will make this assumption false. The metadata endpoint needs to be agnostic to the issuer.

Access Tokens have the issuer, but not the metadata endpoint. There is no way to know just by looking at the Access Token where the metadata endpoint is. For this reason, Vets-API will have to be aware of the endpoints. I purpose adding the following config values to Vets-API.

```yml
oidc:
  issuers:
    - prefix: "https://example.com/"
      metadata: ".well-known/openid-configuration"
      proxy: "internal-dsva-vagov whatever"
```
Vets-API can use this information to determine the metadata endpoint location based on the issuer-prefix provided by the Access Token.

## TODO

- [x] Created an Automated test script to assist in development
- [x] Create Config PR
  - https://github.com/department-of-veterans-affairs/devops/pull/8522
- [x] Any Changes that need to be made to the Config Folder???
- [x] Any Test adjustments
- [x] Implement
- [x] Revert Gemfile
- [x] Draft to Ready

# Testing

```sh
curl -X POST \
    -s \
    -w "%{http_code}" \
    -H 'Accept: application/json' \
    -H 'Content-Type: application/json' \
    -H "Authorization: Bearer ${token}" \
    --data-raw "{ 'aud': "$AUD" }" \
    -o "$curl_body" \
    "$HOST/internal/auth/v1/validation" > "$curl_status"
cat $curl_status
cat $curl_body
```
## Happy Path
- Make call to validation endpoint on a whitelisted issuer.

### expected results
```
# curl_status
200
```

```json
# Validation object
{
  "data": {
    "id": "...",
    "type": "validated_token",
    "attributes": {
      "ver": 1234,
      "jti": "...",
      "iss": "http://example.com",
      "aud": "...",
      "iat": 1234,
      "exp": 2345,
      "cid": "...",
      "uid": "...",
      "scp": [
        "...",
        "...",
        "..."
      ],
      "sub": "...",
      "va_identifiers": {
        "icn": "..."
      }
    }
  }
}
```
- [x] Passed

## Non whitelisted issuer prefix
- Make call to validation endpoint on a non whitelisted issuer prefix.

### Expected Results
- [x] status 401
- [x] following body
```json
{
  "errors": [
    {
      "title": "Open Id Service Error",
      "detail": "Invalid issuer",
      "code": "401",
      "status": "401"
    }
  ]
}
```
- [x] Passed

## Bad issuer
- Make validation call for a whitelisted issuer prefix, with a bad issuer reference

### Expected results
- [x] status 404
- [x] following body
```json
{
  "errors": [
    {
      "title": "Open Id Service Error",
      "detail": "Issuer not found",
      "code": "404",
      "status": "404"
    }
  ]
}
```

- [x] Passed


beaugrantham-va (changes requested):
One of the items discussed around this was to move out of the "okta/service.rb" class, as it's no longer Okta specific. It looks like this never made it into the A/C for the story but I think it should be there. I'll update the story.

- [x] Refactor this functionality out of `okta/service.rb`

This change bypasses the forward-proxy and goes right to the issuer. We should continue to go through the proxy as this is the pattern that vets-api generally follows globally. This may mean refactoring the config to store multiple attributes about an issuer.

- [ ] Traffic should continue to go through the forward-proxy (defined by issuer prefix)

lib/oidc/service.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    def oidc_jwks_keys(iss)
+      url = metadata(iss).body['jwks_uri']
+      with_monitoring do
+        call_no_token('get', url)
+      end
+    end
+
+    def get_metadata_endpoint(iss)
+      metadata_endpoint = Settings.oidc.issuers.find { |s| iss.downcase.include? s['prefix'].downcase }
```
beaugrantham-va, line 36: Two observations...

I think this should be `starts_with` instead of `include`, and I think it needs some error handling if it doesn't know how to proxy the request.

Given:
```
irb(main):042:0> issuers = [ { 'prefix' => 'https://www.example.com/known-issuer', 'proxy' => 'https://www.example.com/proxy' } ]
=> [{"prefix"=>"https://www.example.com/known-issuer", "proxy"=>"https://www.example.com/proxy"}]
```

Happy path - request is modified and sent through the proxy
```
irb(main):049:0> iss = 'https://www.example.com/known-issuer'
=> "https://www.example.com/known-issuer"
irb(main):050:0> metadata_endpoint = issuers.find { |s| iss.downcase.include? s['prefix'].downcase }
=> {"prefix"=>"https://www.example.com/known-issuer", "proxy"=>"https://www.example.com/proxy"}
irb(main):051:0> proxied_iss = iss.gsub(metadata_endpoint['prefix'], metadata_endpoint['proxy'])
=> "https://www.example.com/proxy"
```

Known issuer, but in random location - request is modified but does not go through the proxy
```
irb(main):056:0> iss = 'https://www.example.com/unknown-issuer?q=https://www.example.com/known-issuer'
=> "https://www.example.com/unknown-issuer?q=https://www.example.com/known-issuer"
irb(main):057:0> metadata_endpoint = issuers.find { |s| iss.downcase.include? s['prefix'].downcase }
=> {"prefix"=>"https://www.example.com/known-issuer", "proxy"=>"https://www.example.com/proxy"}
irb(main):058:0> proxied_iss = iss.gsub(metadata_endpoint['prefix'], metadata_endpoint['proxy'])
=> "https://www.example.com/unknown-issuer?q=https://www.example.com/proxy"
```

Unknown issuer - exception
```
irb(main):062:0> iss = 'https://www.example.com/unknown-issuer'
=> "https://www.example.com/unknown-issuer"
irb(main):063:0> metadata_endpoint = issuers.find { |s| iss.downcase.include? s['prefix'].downcase }
=> nil
irb(main):064:0> proxied_iss = iss.gsub(metadata_endpoint['prefix'], metadata_endpoint['proxy'])
Traceback (most recent call last):
        4: from /usr/bin/irb:23:in `<main>'
        3: from /usr/bin/irb:23:in `load'
        2: from /Library/Ruby/Gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
        1: from (irb):64
NoMethodError (undefined method `[]' for nil:NilClass)
```

beaugrantham-va: > and I think it needs some error handling if it doesn't know how to proxy the request.

Or maybe these do just fall through and skip the proxy? I wonder if Derek has an opinion.

dbrown-va: I've thought the same about requests that don't appear to go through the proxy.  There are a number of url's listed in the config yaml's that skip the fwd proxy, so I think the request would fall through and skip the proxy.

If a prefix is found, then the `iss.gsub(metadata_endpoint['prefix'], metadata_endpoint['proxy'])` should succeed.

Maybe `issuers.find` needs to return an error immediately if a match isn't found rather than relying on the NoMethodError for cases without matches?

dbrown-va: The `app/models/token.rb` `valid_issuer?` method will need to be updated to check the different possible issuers as well.
https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/models/token.rb#L50

With that check in place, the `get_metadata_endpoint` should be able to focus on just translating the issuer to the fwd proxied url.

lukemarkwordtlibertyits: I see your points. As Derek mentioned if when we update the `valid_issuer?` method then the invalid prefix error would return
```json
{
  "errors": [
    {
      "title": "Not authorized",
      "detail": "Invalid issuer",
      "code": "401",
      "status": "401"
    }
  ]
}
```
Which will be an acceptable error, but the unknown issuer on a valid prefix is a bit more complicated. 

If we let the error bubble up we would receive

```json
{
  "errors": [
    {
      "title": "Internal server error",
      "detail": "Internal server error",
      "code": "500",
      "status": "500",
      "meta": {
        "exception": "undefined method `index' for nil:NilClass",
        "back_trace": [..]
      }
    }
  ]
}  
```
I'm unsure if this is ok? We could throw an invalid issuer error?

TODO
- [ ] issuer check should be `starts_with?` 
- [ ] add error handling for invalid prefix
- [ ] add error handling for unkown_issuer

beaugrantham-va: Re-grouping my thoughts on this...

This metadata-fetching-functionality is generic (supports any issuer) but that issuer must be whitelisted in the config, otherwise it has no idea which endpoint (`openid-configuration`, `oauth-authorization-server`, etc) to build/query.

Based on that, I think the question about fall-through is not applicable.

If a matching config entry can not be found, an error should be gracefully raised. The caller is responsible for deciding how they want to handle it.

If a matching config entry is found, it should be expected to have a proxy route and a metadata endpoint.

app/models/token.rb
```diff
@@ -50,7 +50,8 @@ def validate_token
   def valid_issuer?
     decoded_token = JWT.decode(@token_string, nil, false, algorithm: 'RS256')
     iss = decoded_token[0]['iss']
-    !iss.nil? && iss.match?(%r{^#{Regexp.escape(Settings.oidc.issuer_prefix)}/\w+$})
```
beaugrantham-va, line 53: This is going to change some behavior and permit unexpected issuers. Let's have a call to walk through when Derek is available.

dbrown-va (approved):

dbrown-va:
Devops configuration addition has been merged and built.  Merging now.

PR #5933, opened 2021-02-13
Add mobile-specific SM backend integration profile

patrickvinograd:
## Description of change
Adds a  mobile-app-specific MHV profile - i.e. a unique appToken configuration, and a separate session store to avoid userid collisions. 

~Note this is based on top of https://github.com/department-of-veterans-affairs/vets-api/pull/5929 which I expect to get merged first.~ Merged

## Original issue(s)

- Adds a new settings.yml element for the  new mobile-specific MHV appToken.  

Planned test: Invoke SM API with configured staging appToken and verify difference  in MHV account activity logs.

### DIFF
--- a/config/redis.yml
+++ b/config/redis.yml
@@ -35,6 +35,9 @@ development: &defaults
   sm_store:
     namespace: sm-service
     each_ttl: 1200
+  sm_store_mobile:
+    namespace: sm-service-mobile
+    each_ttl: 1200
   mdot:
     namespace: mdot
     each_ttl: 1800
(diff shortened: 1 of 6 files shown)


config/redis.yml
```diff
@@ -102,6 +102,9 @@ development: &defaults
   va_mobile_session_refresh_lock:
     namespace: va-mobile-session-refresh-lock
     each_ttl: 60
+  va_mobile_sm_store:
```
kreek, line 105: This is admittedly 100% confusing, and partly my fault as I worked on VAOS before the 'flagship' mobile app existed, but the other `va_mobile_...` things here don't refer to the mobile app. They refer to the VA Mobile Framework (VAMF, of which we only use VAOS related services). IAM sessions for the that the Mobile App uses for user sessions are under `iam_...`. Although obviously since this is for SM client sessions you don't really want that either 😅

patrickvinograd: hmm...yeah, maybe I'll flip it  to "sm_store_mobile" or something.

jjulian:
There may be no concern, but changing the redis namespace will create a number of MISSes immediately after deployment.

patrickvinograd:
Devops config here: https://github.com/department-of-veterans-affairs/devops/pull/8524

thilton-oddball (approved):
Looks good!

PR #5935, opened 2021-02-15
API-5203 Add NOD schema specs and rename HLR specs

ElleeB:
I noticed we had json schema validation specs for HLR, but not for NOD. Added the checks for NOD. I ran both HLR and NOD tests and the `valid headers test` for each failed.

1. The files were named incorrectly and were not running in the test suite - fixed the file names
2. Both header schemas were missing the `Consumer ID` keys - added
3. Tests running and passing locally

Ticket https://vajira.max.gov/browse/API-5203

*** Please let me know if the schema description could be more specific, and be sure to check that it is ok to simply check for a non-blank string for the `Consumer ID` ***


modules/appeals_api/config/schemas/10182_headers.json
```diff
@@ -15,7 +15,8 @@
         "X-VA-Veteran-SSN":             { "$ref": "#/definitions/nodCreateHeadersSsn" },
         "X-VA-Veteran-File-Number":     { "$ref": "#/definitions/nodCreateHeadersFileNumber" },
         "X-VA-Veteran-Birth-Date":      { "$ref": "#/definitions/nodCreateHeadersBirthDate" },
-        "X-Consumer-Username":          { "$ref": "#/definitions/nodCreateHeadersConsumerUsername"}
+        "X-Consumer-Username":          { "$ref": "#/definitions/nodCreateHeadersConsumerUsername"},
+        "X-Consumer-ID":                { "$ref": "#/definitions/nodCreateHeadersConsumerId"}
```
yakschuss, line 19: Is this definitely supposed to be required?

yakschuss (approved):
Assuming we're actually supposed to be requiring Consumer ID, this looks good to me. Good find!

modules/appeals_api/config/schemas/10182_headers.json
ElleeB: It's not required (take note it is not included in the required collection), but supplied by Kong for the stamp on our pdf along with `Consumer-Username`. :)

ElleeB: ^ Which now makes me wonder if is _needs_ to be required hmmmmm...

ElleeB: Update: not required

PR #5944, opened 2021-02-16
remove view payments feature toggle

micahchiang:
## Description of change

The application that relies on this feature toggle is available in production to 100% of users, so it is no longer needed.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#19553


thilton-oddball (approved):
LGTM

PR #5945, opened 2021-02-16
Greg.bowman/api 3392 daily email update

gpbowman-git:
## Description of change

Updating the daily email report to include table styling along with the success rate and a summary total row for the reporting period.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#3392

image


modules/vba_documents/app/workers/vba_documents/report_unsuccessful_submissions.rb
```diff
@@ -30,20 +29,66 @@ def stuck
       ).order(:consumer_name, :status)
     end
 
+    # rubocop:disable Metrics/MethodLength
+    # rubocop:disable Metrics/PerceivedComplexity
+    # rubocop:disable Metrics/CyclomaticComplexity
+    # rubocop:disable Metrics/AbcSize
     def totals
```
ericboehs, line 36: This method is definitely due for refactoring. I think disabling the cops here should be done in [the todo file](https://github.com/department-of-veterans-affairs/vets-api/blob/master/.rubocop_todo.yml) rather than inline. Disabling inline/in file is equivalent to saying "Rubocop was wrong here and this code is how we want it."

modules/vba_documents/spec/mailers/unsuccessful_report_mailer_spec.rb
```diff
@@ -83,7 +95,8 @@ class << self
         let(:env_url) { [k, v] }
 
         it "State the environment #{k.to_s.upcase} in the body" do
```
ericboehs, line 97:
```suggestion
        it "states the environment #{k.to_s.upcase} in the body" do
```

ericboehs (approved):
A few suggestions, but I'm not going to block on them.

modules/vba_documents/spec/mailers/unsuccessful_report_mailer_spec.rb
```diff
@@ -83,7 +95,8 @@ class << self
...
-          expect(@email.body.to_s.split("\n").first).to match(/.*#{k.to_s.upcase}.*/)
+          @email.body.to_s =~ %r{<h1>(.*)?</h1>}i
+          expect(Regexp.last_match(1)).to match(/.*#{k.to_s.upcase}.*/)
```
ericboehs, line 99: This is a bit difficult to read.

PR #5947, opened 2021-02-16
Api 4960/nod status

nathanbwright:
## Description of change
This PR extracts the responsibility of talking to central mail from the appeal concern into its own object. 

I still need to rename the concern to be something more suitable but that would push me over the 500 line limit, so that will have to come in another PR... :(

## Original issue(s)


cilestin (approved):
:shipit:

PR #5948, opened 2021-02-16
19897 add mobile appointments cache flag

kreek:
## Description of change
Adds a flag for whether or not to use the upstream VAOS service's cache of appointments. Changes default to be false.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#19897


jperk51 (approved):
LGTM

ericboehs (approved):

PR #5951, opened 2021-02-16
store invalid addresses in PII logs table

lihanli:
## Description of change
log to pii logs table when a va profile record fails validation (frontend validations are supposed to match backend validations so this should not happen normally)

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/20025

tested with unit tests

### DIFF
--- a/app/controllers/concerns/vet360/writeable.rb
+++ b/app/controllers/concerns/vet360/writeable.rb
@@ -39,7 +39,13 @@ module Vet360
     end
 
     def validate!(record)
-      raise Common::Exceptions::ValidationErrors, record unless record.valid?
+      if !record.valid?
+        PersonalInformationLog.create!(
+          data: record.to_h,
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/vet360/writeable.rb
```diff
@@ -39,7 +39,13 @@ def build_record(type, params)
     end
 
     def validate!(record)
-      raise Common::Exceptions::ValidationErrors, record unless record.valid?
+      unless record.valid?
```
ericboehs, line 42: I'm surprised rubocop didn't say anything here. This should fail the [guard clause cop](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/GuardClause).

Should be:
```
return if record.valid?

# code
```

ericboehs (approved):

PR #5952, opened 2021-02-16
Adds Mobile messaging infra and triage teams API

patrickvinograd:
## Description of change
Mobile secure messaging APIs part 1 of n

### DIFF
--- a/modules/mobile/app/serializers/mobile/v0/triage_team_serializer.rb
+++ b/modules/mobile/app/serializers/mobile/v0/triage_team_serializer.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module Mobile
+  module V0
+    class TriageTeamSerializer < ActiveModel::Serializer
+      def id
+        object.triage_team_id
+      end
+
+      attribute :triage_team_id
+      attribute :name
+      attribute :relation_type
+    end
+  end
+end
--- a/modules/mobile/spec/request/triage_teams_request_spec.rb
+++ b/modules/mobile/spec/request/triage_teams_request_spec.rb
@@ -1,0 +1,59 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require_relative '../support/iam_session_helper'
+require_relative '../support/mobile_sm_client_helper'
+
+RSpec.describe 'Mobile Triage Teams Integration', type: :request do
+  include Mobile::MessagingClientHelper
+  include SchemaMatchers
+
+  let(:va_patient) { true }
+  # let(:current_user) { build(:user, :mhv, va_patient: va_patient, mhv_account_type: mhv_account_type) }
+
+  before do
+    allow_any_instance_of(MHVAccountTypeService).to receive(:mhv_account_type).and_return(mhv_account_type)
+    allow(Mobile::V0::Messaging::Client).to receive(:new).and_return(authenticated_client)
+    iam_sign_in(build(:iam_user, iam_mhv_id: '123'))
+  end
+
+  context 'Premium User' do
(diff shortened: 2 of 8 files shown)


modules/mobile/app/serializers/mobile/v0/triage_team_serializer.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Mobile
+  module V0
+    class TriageTeamSerializer < ActiveModel::Serializer
```
kreek, line 5: Could this use the fast fast_jsonapi/jsonapi-serializer? Our serializers use the former but we're moving the latter soon. Fast JSON API was developed at Netflix and is faster than ActiveModel serializer (especially at large resource counts), it was recently forked to [jsonapi-serializer](https://github.com/jsonapi-serializer/jsonapi-serializer).

modules/mobile/spec/request/triage_teams_request_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+require 'rails_helper'
+require_relative '../support/iam_session_helper'
+require_relative '../support/mobile_sm_client_helper'
+
+RSpec.describe 'Mobile Triage Teams Integration', type: :request do
+  include Mobile::MessagingClientHelper
+  include SchemaMatchers
+
+  let(:va_patient) { true }
+  # let(:current_user) { build(:user, :mhv, va_patient: va_patient, mhv_account_type: mhv_account_type) }
```
kreek, line 12: stray commented out code?

modules/mobile/spec/support/mobile_sm_client_helper.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require 'mobile/v0/messaging/client'
+
+module Mobile
+  module MessagingClientHelper
+    TOKEN = 'GkuX2OZ4dCE=48xrH6ObGXZ45ZAg70LBahi7CjswZe8SZGKMUVFIU88='
```
kreek, line 7: Should this be a setting? Also confused about `used_id` below 😄, is that a placeholder?

modules/mobile/app/serializers/mobile/v0/triage_team_serializer.rb
kreek: I guess it depends how well it plays with the OG secure messaging client and its metadata req, it would be nice to be consistent, but we could also refactor later if we want to get this in for testing.

kreek (approved):
Couple comments/questions but LGTM 🚀

modules/mobile/app/serializers/mobile/v0/triage_team_serializer.rb
patrickvinograd: I ran across that when debugging what was going on with the metadata links...I'm in favor of refactoring later.

modules/mobile/spec/request/triage_teams_request_spec.rb
patrickvinograd: yep, will remove

modules/mobile/spec/support/mobile_sm_client_helper.rb
patrickvinograd: This is basically stubbing out the client step where it gets a session token; the value of token and client_id don't  matter because VCR  is just going  to match on the route. I'll make it a more obviously fake value.

patrickvinograd:
Closing this in favor of https://github.com/department-of-veterans-affairs/vets-api/pull/5953 since some of the route helper code isn't passing coverage in this PR.

PR #5953, opened 2021-02-16
Adds mobile messaging folders API

patrickvinograd:
## Description of change
Adds initial revision of mobile-specific secure messaging APIs - folders and triage teams and supporting code.
Mobile secure messaging APIs parts 1 and 2 of 3

### DIFF
--- a/modules/mobile/app/controllers/mobile/v0/folders_controller.rb
+++ b/modules/mobile/app/controllers/mobile/v0/folders_controller.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+require_dependency 'mobile/messaging_controller'
+require_dependency 'mobile/v0/folder_serializer'
+
+module Mobile
+  module V0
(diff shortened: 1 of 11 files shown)


modules/mobile/app/controllers/mobile/v0/folders_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+
+require_dependency 'mobile/messaging_controller'
+require_dependency 'mobile/v0/folder_serializer'
+
+module Mobile
+  module V0
+    class FoldersController < MessagingController
+      def index
+        resource = client.get_folders
+        resource = resource.paginate(pagination_params)
```
kreek, line 11: not sure we'll ever hit this endpoint with pagination_params (as they can be blank) but invalid pagination_params seem to throw a 500 rather than a 404 or 422

kreek (approved):
One minor issue but otherwise LGTM 🚀

PR #5954, opened 2021-02-16
Adds mobile SM messages API

patrickvinograd:
Mobile secure messaging APIs part 3 of 3

### DIFF
--- a/modules/mobile/spec/request/messages_request_spec.rb
+++ b/modules/mobile/spec/request/messages_request_spec.rb
@@ -1,0 +1,185 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require_relative '../support/iam_session_helper'
+require_relative '../support/mobile_sm_client_helper'
+
+RSpec.describe 'Mobile Messages Integration', type: :request do
+  include Mobile::MessagingClientHelper
+  include SchemaMatchers
+
+  let(:user_id) { '10616687' }
+  let(:inbox_id) { 0 }
+  let(:message_id) { 573_059 }
+  let(:va_patient) { true }
+
+  before do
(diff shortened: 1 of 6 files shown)


modules/mobile/spec/request/messages_request_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+        it 'with attachments' do
+          VCR.use_cassette('sm_client/messages/creates/a_new_message_with_4_attachments') do
+            post '/mobile/v0/messaging/health/messages', headers: iam_headers, params: params_with_attachments
+          end
+
+          expect(response).to be_successful
+          expect(response.body).to be_a(String)
+          expect(JSON.parse(response.body)['data']['attributes']['subject']).to eq('CI Run')
+          expect(JSON.parse(response.body)['data']['attributes']['body']).to eq('Continuous Integration')
+          expect(response).to match_camelized_response_schema('message_with_attachment')
```
kreek, line 110: A little bit painful but for the JSON schema checks we have our own set of _almost_ duplicate schemas dir in support as a way to protect us from the core lib being change from out from under us. At the risk of the PR line count 🤖 coming after you, you could add `match_camelized_response_schema` to our support/matches/json_schema_matcher.rb and copy the core lib ones over. (also helps us avoid needing VSP approval if there's a future mobile only change)

kreek (approved):
One note about mobile schemas but LGTM 🚀 

Is there another PR on the way with api docs? In the modules docs folder there's a openapi.yml file and a script to generate the static HTML docs the mobile front-end devs use for reference.

modules/mobile/spec/request/messages_request_spec.rb
patrickvinograd: That's a good idea; I meant to sync with you about  where you'd landed on duplication vs. re-use, generally.

PR #5967, opened 2021-02-18
change which status we key off of for claims

jperk51:
## Description of change
Changed which attribute to key off of for making a claim open/closed

## Original issue(s)
N/A

Noticed a mismatch in staging. This should fix that and on a second look the status attribute seems to be the better attribute to use.


kreek (approved):
LGTM 🚀

modules/mobile/app/models/mobile/v0/adapters/claims_overview.rb
```diff
@@ -10,7 +10,7 @@ def parse(entry)
                              id: entry['evss_id'].to_s,
                              type: 'claim',
                              subtype: entry['list_data']['status_type'],
-                             completed: entry['list_data']['claim_status'] != 'PEND',
+                             completed: entry['list_data']['status'] != 'COMPLETE',
```
ericboehs, line 13:
```suggestion
                             completed: entry['list_data']['status'] == 'COMPLETE',
```

Should the completed key here but based on if the status is NOT complete?

jperk51: Whoops, thanks. Also I'll add a few more checks to the test cases so something like that can't slip though.

ericboehs (approved):

PR #5969, opened 2021-02-18
Add attachments endpoint to mobile SM API

patrickvinograd:
## Description of change
Adds the attachments controller/serializer allowing secure message attachments to be downloaded, and to be properly rendered as sub-resources in a message. Swagger docs added for the newly-available endpoint.

Tested locally.


modules/mobile/docs/openapi.yaml
```diff
@@ -1985,6 +1985,40 @@ paths:
             application/json:
               schema:
                 $ref: "./schemas/Errors.yml"
+  /v0/messaging/health/messages{id}/attachments/{attachment_id}:
```
kreek, line 1988:
```suggestion
  /v0/messaging/health/messages/{id}/attachments/{attachment_id}:
```

kreek: added this suggestion but you'll need to render again 😄

kreek (approved):
LGTM 🚀

modules/mobile/docs/openapi.yaml
patrickvinograd: Good catch, thank you!

PR #5973, opened 2021-02-18
add new va profile async transaction models

lihanli:
## Description of change
add new `AsyncTransaction::VAProfile` models
old `AsyncTransaction::Vet360` models are kept for backwards compatability, but new transactions will not be created using those models
when fetching a transaction by id `V0::Profile::TransactionsController` will return both `AsyncTransaction::VAProfile` and `AsyncTransaction::Vet360` models so that transactions created before the update won't cause errors. this behavior can be removed in [one month](https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/models/async_transaction/base.rb#L9) since the old transactions will be deleted by then.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/19858

tested on staging


ericboehs (approved):

kreek (approved):
LGTM 🚀

PR #5974, opened 2021-02-19
19897 mobile default get appointments time range

kreek:
## Description of change
Updates `GET /mobile/v0/appointments` to return a default date range of appointments from 3 months prior to and 6 months after the beginning of today's UTC date.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#19897


jperk51 (approved):
LGTM

PR #5982, opened 2021-02-22
Mark.viterna/api 4655 faraday timeout

mviterna:
## Description of change
Updating the Faraday timeout on FormReloader to correct timeout issues on Sandbox

## Original issue(s)
https://vajira.max.gov/browse/API-4655

Tested locally and confirmed timeout is working as intended


cshupp1 (approved):

modules/va_forms/app/workers/va_forms/form_reloader.rb
```diff
@@ -13,7 +14,11 @@ def perform
       Rails.logger.info('VAForms::FormReloader is being called.')
       query = File.read(Rails.root.join('modules', 'va_forms', 'config', 'graphql_query.txt'))
       body = { query: query }
-      response = connection.post('graphql', body.to_json)
+      response = connection.post do |req|
+        req.path = 'graphql'
+        req.body = body.to_json
+        req.options.timeout = 300
```
ericboehs, line 20: 5 minute timeout for a http request? Seems a bit high.

modules/va_forms/app/workers/va_forms/form_reloader.rb
```diff
@@ -3,6 +3,7 @@
 require 'sidekiq'
 
 module VAForms
+  # rubocop:disable Metrics/MethodLength
```
ericboehs, line 6: I'd prefer we disable cops via .rubocop_todo.yml instead of in file unless we think shortening the method length here would be wrong (which I don't know when that would ever be the case). Even better, shorten the method via extraction.

ericboehs (approved):
Two requests; non-blocking.

modules/va_forms/app/workers/va_forms/form_reloader.rb
mviterna: @ericboehs So, my understanding on Faraday is that this would be the READ timeout, rather than OPEN (for opening the connection). Based on my CURL testing, the payload can take up to 3 minutes to return the data. The problem is that we are retrieving all the form data at once in a large JSON payload. It probably needs some optimization, but this is to fix a current issue where we are seeing timeouts and the forms are not getting updated.

modules/va_forms/app/workers/va_forms/form_reloader.rb
mviterna: @ericboehs Sure, I shortened it. Will check it in

PR #5984, opened 2021-02-22
Hotfix for lib/apps/client.rb#get_app

bradenshipley:
## Description of change

Originally the `get_app` method used `URI.escape` to handle a `search_term` with whitespace in it (IE: 'Apple Health'). Rubocop throws a deprecation error for `URI.escape` and suggests using `CGI.escape`. The test written for `get_app` used `iBlueButton` as the `search_term`, and therefor didn't catch the issue when replacing `URI.escape` with `CGI.escape`. When passing a `search_term` with whitespace in it , IE: `"Apple Health"`, `CGI.escape` parses that as "Apple+Health" instead of "Apple%20Health" as `URI.escape` did.

The method now uses `ERB::Util.url_encode`, which works as expected, and a test has been included to ensure it parses whitespace correctly.

## Original issue(s)

### DIFF
--- a/lib/apps/client.rb
+++ b/lib/apps/client.rb
@@ -4,7 +4,7 @@ require 'common/client/base'
 require 'common/client/concerns/monitoring'
 require_relative 'configuration'
 require_relative 'responses/response'
-require 'cgi'
+require 'erb'
 
 module Apps
   # Proxy Service for Apps API.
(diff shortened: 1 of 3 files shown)


nathanbwright (approved):
Looks good to me.

PR #5986, opened 2021-02-22
API 5222 evidence document validation

ElleeB:
## Description of change
Validation service for appeals uploaded evidence files.

## Original issue(s)
https://vajira.max.gov/browse/API-5222


modules/appeals_api/app/services/appeals_api/uploads/document_validation.rb
```diff
@@ -0,0 +1,48 @@
...
+      end
+
+      def validate
+        # TODO: log anything here if errors raised or log in the controller if fails validation?
+        raise UploadValidationError, non_pdf_error unless document_is_pdf?
+        raise UploadValidationError, max_dimension_error unless valid_page_dimensions?
+        raise UploadValidationError, file_size_error unless valid_file_size?
+      end
+
+      def document_is_pdf?
```
ElleeB, line 19: This was interesting. The solution I finally came up with seems to account for both non-pdf files with non-pdf extensions, _and_ non-pdf files with false .pdf extensions.

modules/appeals_api/spec/services/appeals_api/uploads/document_validation_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    context 'when supported file uploaded (100MB 11in x 11in PDF)' do
+      it 'does not raise an error' do
+        expect { described_class.new(valid_pdf).validate }.not_to raise_error
+      end
+    end
+
+    context 'when unsupported file type uploaded' do
+      it 'raises an error when non-pdf file type' do
+        expect { described_class.new(non_pdf).validate }.to raise_error AppealsApi::Uploads::DocumentValidation::UploadValidationError
```
ElleeB, line 21: Going to add checks for the actual text content of the errors found in the locale files.

modules/appeals_api/app/services/appeals_api/uploads/document_validation.rb
```diff
@@ -0,0 +1,48 @@
...
+        contents = File.read(@document)
+        contents.include?('%PDF-')
```
cilestin, line 21: How much of the file needs to be read to determine this? Right now it appears to read the entire file into memory to check for any mention of `%PDF-` anywhere in the file, which could be a problem for large files.

Assuming this is something that can be found in the header of the file, we should try to only read a minimum amount of data before breaking away and failing.

modules/appeals_api/app/services/appeals_api/uploads/document_validation.rb
```diff
@@ -0,0 +1,48 @@
...
+    class DocumentValidation
+      def initialize(upload)
+        @document = upload[:document]
+      end
+
+      def validate
+        # TODO: log anything here if errors raised or log in the controller if fails validation?
+        raise UploadValidationError, non_pdf_error unless document_is_pdf?
+        raise UploadValidationError, max_dimension_error unless valid_page_dimensions?
+        raise UploadValidationError, file_size_error unless valid_file_size?
```
cilestin, line 16: Let's try to make sure that each of these checks only reads the contents into memory as few times as possible - either by caching the contents in an instance variable or only reading the minimum of the file necessary to get the job done for each validation. Most of these checks are likely doing this already, but we should verify that.

If some of these require the full load of the file into memory, let's rearrange these validations in order of "expense", where the leanest come first and the most memory intensive are last.

modules/appeals_api/app/services/appeals_api/uploads/document_validation.rb
ElleeB: OK. Cool. I need to research a bit more, because it's possible it's always captured within the first 8 characters, but I haven't been able to find anything absolutely definitive. I'll keep looking. I'll let you know what I find.

modules/appeals_api/app/services/appeals_api/uploads/document_validation.rb
ElleeB: OK I'll dig into this.

modules/appeals_api/app/services/appeals_api/uploads/document_validation.rb
```diff
@@ -0,0 +1,45 @@
...
+module AppealsApi
+  module Uploads
+    class DocumentValidation
+      def initialize(upload)
+        @document = upload[:document]
+      end
+
+      def validate
+        pdf_metadata_present?
+        raise UploadValidationError, file_size_error unless valid_file_size?
```
yakschuss, line 14: Should we be raising here? The idea is the return meaningful data about the data validation to the client, so they can change/improve on what they're submitting.

An example from the matsumoto RFC documenting the pattern. I don't know if the schema should look like this, but something to think about.

```
   "attachments": [
                        {
                            "page_count": 1,
                            "dimensions": {
                                "height": 8.5,
                                "width": 11.0,
                                "oversized_pdf": false
                            }
                        },
                        {
                            "page_count": 1,
                            "dimensions": {
                                "height": 8.5,
                                "width": 11.0,
                                "oversized_pdf": false
                            }
                        }
                    ]
```

ElleeB: This is great info. Thank you for sharing.

modules/appeals_api/app/services/appeals_api/evidence_submission/upload_validator.rb
```diff
@@ -0,0 +1,111 @@
...
+    class UploadValidator
+      include SentryLogging
+
+      def initialize(upload)
+        @document = upload[:document]
+      end
+
+      def call
+        return file_type_error unless pdf_metadata_present?
+
```
yakschuss, line 16: I think it might be possible to use guard clauses here for the entire `call` method. 

With the multi-tiered if statement, any future validations or changes might be more difficult to implement. What do you think?

ElleeB: This is what I had at some point:
`return file_type_error unless pdf_metadata_present?
  return file_size_error unless valid_file_size?
  return max_dimension_error unless valid_page_dimensions?

  successful_validation_message
`
Then I just kept hearing "one return only" replaying in my head so I changed it. I prefer the return statements to the conditional. So as long as you're cool with it, i'm cool with it

yakschuss: Yeah. I think that reads much more cleanly, and will be easier to adjust if necessary, personally.

ElleeB: Perfect. Rewind!

modules/appeals_api/spec/services/appeals_api/evidence_submission/file_validator_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+  describe '#validate' do
+    let(:fixtures_path) { '/modules/appeals_api/spec/fixtures' }
+
+    let(:valid_pdf) { File.new("#{::Rails.root}#{fixtures_path}/expected_10182_extra.pdf") }
+    let(:non_pdf) { File.new("#{::Rails.root}#{fixtures_path}/test_incorrect_type.docx") }
+    let(:pretend_pdf) { File.new("#{::Rails.root}#{fixtures_path}/test_fake_pdf.pdf") }
+    let(:oversize_pdf) { File.new("#{::Rails.root}#{fixtures_path}/oversize_11x17.pdf") }
+
+    context 'when supported file uploaded (100MB 11in x 11in PDF)' do
+      it 'does not raise an error' do
```
cilestin, line 15: To me, the word "raise" is something specific around exceptions - e.g `expect(x).to raise_error XYZ`. This is returning structured data about an error, not raising.

Something like `returns error data when...` is clearer to me for these types of tests.

modules/appeals_api/app/services/appeals_api/evidence_submission/file_validator.rb
```diff
@@ -0,0 +1,103 @@
...
+        @pdf_metadata = PdfInfo::Metadata.read(@file)
+      rescue PdfInfo::MetadataReadError => e
+        @pdf_error = I18n.t('appeals_api.evidence_submission.pdf_read_error')
+        log_exception_to_sentry(e, {}, {}, :warn)
+      ensure
+        @pdf_metadata.present?
+      end
+
+      def valid_file_size?
+        current_file_size <= 100.megabytes
```
cilestin, line 34: Feels like a magic number. Let's move the 100.megabytes into a constant

modules/appeals_api/app/services/appeals_api/evidence_submission/file_validator.rb
```diff
@@ -0,0 +1,103 @@
...
+      end
+
+      def valid_page_dimensions?
+        @dimensions = @pdf_metadata.page_size_inches
+        @dimensions[:height] <= 11 && @dimensions[:width] <= 11
```
cilestin, line 39: More magic numbers. Maybe a constant like `MAX_PAGE_SIZE = {width: 11, height: 11}`

cilestin (changes requested):

modules/appeals_api/spec/services/appeals_api/evidence_submission/file_validator_spec.rb
ElleeB: Nice. Thank you!

modules/appeals_api/spec/services/appeals_api/evidence_submission/file_validator_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+              pages: 1,
+              detail: 'File exceeds the maximum page dimensions of 11 inches x 11 inches',
+              file_dimensions: { height: 11.0, width: 17.0 }
+            }
+        )
+      end
+    end
+
+    context 'when file size is too large' do
+      before { allow_any_instance_of(described_class).to receive(:valid_file_size?).and_return(false) }
```
cilestin, line 71: We should mock out the size of the file rather than mocking parts of the very class we're trying to test.

cilestin (changes requested):
Definitely getting there 👍 

As we talked about, I'm still not entirely sure this class should be trying to build the result that's ultimately the controller's job, but we an revisit it as we start working on the endpoint.

modules/appeals_api/spec/services/appeals_api/evidence_submission/file_validator_spec.rb
ElleeB: I don't understand why this doesn't work, but I can mock out the file size instead insert 🤪

modules/appeals_api/app/services/appeals_api/evidence_submission/file_validator.rb
```diff
@@ -0,0 +1,90 @@
...
+      def valid_page_dimensions?
+        @dimensions = @pdf_metadata.page_size_inches
+        @dimensions[:height] <= MAX_PAGE_SIZE[:height] && @dimensions[:width] <= MAX_PAGE_SIZE[:width]
+      end
+
+      def file_type_error
+        msg = I18n.t('appeals_api.evidence_submission.pdf_read_error')
+        log_exception_to_sentry(UploadValidationError.new(msg), {}, {}, :warn)
+        {
+          title: 'invalid file type',
```
cilestin, line 59: `exceptions.en.yml`appears to have the first letter of the title capitalized - let's follow the same format.

cilestin (changes requested):
⚙️

modules/appeals_api/app/services/appeals_api/evidence_submission/file_validator.rb
```diff
@@ -0,0 +1,90 @@
...
+        {
+          title: 'invalid dimensions',
+          detail: msg,
+          meta: { filename: @filename, dimensions: { width: @dimensions[:width], height: @dimensions[:height] } }
+        }
+      end
+
+      def file_size_error
+        msg = I18n.t('appeals_api.evidence_submission.max_file_size_error')
+        log_exception_to_sentry(UploadValidationError.new(msg), {}, {}, :warn)
```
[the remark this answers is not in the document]
ElleeB: ^ I'm still not sure If I should keep them here at all. I keep going back and forth. Should they live here or in the controller...do we want to log every specific failure, or just that there was an upload failure?

modules/appeals_api/app/services/appeals_api/evidence_submission/file_validator.rb
```diff
@@ -0,0 +1,90 @@
...
+        {
+          title: 'invalid file size',
+          detail: msg,
+          meta: { filename: @filename, file_size: "#{file_size_mb} MB" }
```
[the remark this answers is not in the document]
ElleeB: Derp thought i had it in the message

modules/appeals_api/app/services/appeals_api/evidence_submission/file_validator.rb
```diff
@@ -0,0 +1,90 @@
...
+        }
+      end
+
+      def max_dimension_error
+        msg = I18n.t('appeals_api.evidence_submission.max_dimensions_error')
+        log_exception_to_sentry(UploadValidationError.new(msg), {}, {}, :warn)
+        {
+          title: 'invalid dimensions',
+          detail: msg,
+          meta: { filename: @filename, dimensions: { width: @dimensions[:width], height: @dimensions[:height] } }
```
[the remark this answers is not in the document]
ElleeB: Derp thought i had it in the message

modules/appeals_api/config/locales/en.yml
```diff
@@ -6,3 +6,7 @@ en:
         hearing_type_preference_missing: "If '/data/attributes/boardReviewOption' 'hearing' is selected, '/data/attributes/hearingTypePreference' must also be present"
         nod_not_found: "NoticeOfDisagreement with uuid %{id} not found."
         not_homeless_address_missing: "If not homeless, address must be provided: '/data/attributes/veteran/address'"
+      evidence_submission:
+        max_dimensions_error: "File exceeds the maximum page dimensions."
+        max_file_size_error: "File exceeds the max-file-size."
```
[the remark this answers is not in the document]
ElleeB: ya ya

modules/appeals_api/app/services/appeals_api/evidence_submission/file_validator.rb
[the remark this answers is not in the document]
ElleeB: Ok, rewired according to your recommendation and put the log_error method right in the detail. It's clean and makes enough sense to keep it here and not worry about it in the controller.

modules/appeals_api/app/services/appeals_api/file_validator.rb
```diff
@@ -0,0 +1,91 @@
...
+      @file = file
+      @filename = File.basename(@file)
+    end
+
+    def call
+      return file_type_error unless pdf_metadata_present?
+      return file_size_error unless valid_file_size?
+      return max_dimension_error unless valid_page_dimensions?
+
+      validated
```
cilestin, line 25: Why extract out a lone `true` call with a method? Just return `true` if that's what you want to do.

What do you think about returning a two-element array of status & data instead of this split brain problem where the caller needs to worry about data types since errors return a hash but success returns a boolean `true`?

```suggestion
return [:error, file_type_error] unless pdf_metadata_present?
return [:error, file_size_error] unless valid_file_size?
return [:error, max_dimension_error] unless valid_page_dimensions?

[:ok, {}]
```

Then the caller can do a `status, data = FileValidator.new(myfile).call` and know via `status` whether the result was ok or an error, and then work with the `data` if it needs to.

cilestin (changes requested):
One last thing and then I think we're good!

modules/appeals_api/app/services/appeals_api/file_validator.rb
yakschuss: This makes me wish Ruby had better support for tuples. I agree with the premise, however - The interface should probably be consistent so that there's not a cycle of `if` statements to parse out the response.

ElleeB: The lone true call: me playing around with the return of a successful validation. I'm having trouble putting the bigger picture together here. What is he ultimate responsibility of this validator, and how we want to handle its output in the controller?

yakschuss: Said simply, I think the responsibility of the validator is to accept a document, and determine whether it passes the criteria or not.

As for handling the output of the controller, we shouldn't be assuming that the Validator's lone client will be a controller. In theory, we could call this interface anywhere.

cilestin (approved):
\o/

PR #5993, opened 2021-02-23
[Bugfix] Send CentralMail receiveDt in central timezone

cilestin:
## Description of change
CentralMail expects its `receiveDt` to be in Central timezone. This fixes the bug where we were sending the date in UTC instead of Central timezone.

## Original issue(s)
https://vajira.max.gov/browse/API-5366


nathanbwright (approved):
Looks good to me! 👍

PR #5996, opened 2021-02-23
replaces 'user.va_profile.address' with 'user.address' for User model

bramleyjl:
## Description of change

This PR refactors the User model to remove references to `va_profile.address` as part of the larger move to simplify accessing User properties and move references to external data providers into the User class. The property `user.address` is now used in place of `user.va_profile.address`.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#18740
department-of-veterans-affairs/va.gov-team#20327

As `user.address` is being used simply to reference `user.va_profile.address` there should be no functional/noticeable changes to end users as a result of this PR.

* Updated related specs and ran them
* Manual testing done by mocking an `mpi.profile.address` and confirming the `user.address` accurately reflected the mocked changes.


app/models/id_card_attributes.rb
```diff
@@ -16,10 +16,10 @@ def traits
       'edipi' => @user.edipi,
       'firstname' => @user.first_name,
       'lastname' => @user.last_name,
-      'address' => @user.va_profile&.address&.street || '',
-      'city' => @user.va_profile&.address&.city || '',
-      'state' => @user.va_profile&.address&.state || '',
-      'zip' => @user.va_profile&.address&.postal_code || '',
+      'address' => @user&.address&.street || '',
```
bosawt, line 19: I think we don't need the first safety operator for these (so `@user.address&.street`), since we know that `@user` won't be nil

app/models/user.rb
```diff
@@ -107,6 +107,10 @@ def birth_date
     Formatters::DateFormatter.format_date(birth_date)
   end
 
+  def address
+    mpi.profile&.address
```
bosawt, line 111: Can you call `mpi_profile` here? It's a little silly right now, but the goal is to have the `mpi profile` calls going through a single function so we can check for nil (and possibly do other things like log the call in the future, etc)

app/models/user.rb
```diff
@@ -107,6 +107,10 @@ def birth_date
...
+  end
```
bosawt, line 112: I wonder if this function should be expanded to always return a hash, like this:

> def address
>   address = mpi_profile&.address
>  {
>    street: address&.street,
>    city: address&.city,
>    state: address&.state,
>    country: address&.country,
>    zip: address&.zip
>  }
> end

That could make it so the other anything calling `user.address` could always assume that the fields on `address` will be there, even if it still is `nil`, and could let you do `user.address.state` without the safety operators.

app/services/mhv_accounts_service.rb
```diff
@@ -94,13 +94,13 @@ def handle_upgrade!
...
   def address_params
-    if user.va_profile&.address.present?
+    if user.address.present?
       {
-        address1: user.va_profile.address.street,
-        city: user.va_profile.address.city,
-        state: user.va_profile.address.state,
-        zip: user.va_profile.address.postal_code,
-        country: user.va_profile.address.country
+        address1: user.address.street,
```
bosawt, line 99: So annoying that this one address hash has `address1`, could have had the same address hash the same everywhere if it weren't for that, haha

bosawt (commented):
Nice work! I Just had a single suggestion, really, but looks good! Lemme know what you think about the potential change, and I'll approve

spec/models/form_profile_spec.rb
```diff
@@ -52,10 +52,10 @@
...
       'street' => street_check[:street],
       'street2' => street_check[:street2],
-      'city' => user.va_profile[:address][:city],
-      'state' => user.va_profile[:address][:state],
-      'country' => user.va_profile[:address][:country],
-      'postal_code' => user.va_profile[:address][:postal_code][0..4]
+      'city' => user.address.city,
+      'state' => user.address.state,
+      'country' => user.address.country,
+      'postal_code' => user.address.postal_code[0..4]
```
ericboehs, line 58:
```suggestion
      'postal_code' => user.address.postal_code[0...5]
```

I'm not sure which is more intuitive. One has to be expecting zero-index here. Maybe my programmer brain is on wrong but I was looking for suspiciously at the `4`.

You could also do:

```suggestion
      'postal_code' => user.address.postal_code.slice(0, 5)
```

ericboehs (approved):

app/models/form_profile.rb
```diff
@@ -286,16 +286,17 @@ def vet360_mailing_address
...
-      {
-        street: user.va_profile.address.street,
-        street2: nil,
-        city: user.va_profile.address.city,
-        state: user.va_profile.address.state,
-        country: user.va_profile.address.country,
-        postal_code: user.va_profile.address.postal_code
-      }
+  def user_address_hash
+    return unless user.address
```
bosawt, line 290: Now you don't even need this line! Because the `user.address` object always returns a hash, it'll always have the `street`, `city`, etc fields on it, even if it's nil

app/services/mhv_accounts_service.rb
```diff
@@ -94,13 +94,13 @@ def handle_upgrade!
   end
 
   def address_params
-    if user.va_profile&.address.present?
+    if user.address.present?
```
bosawt, line 97: Don't need this line either anymore

spec/request/in_progress_forms_request_spec.rb
```diff
@@ -209,10 +209,10 @@ def response_body
...
               'street' => street_check[:street],
               'street2' => street_check[:street2],
-              'city' => user.va_profile.address.city,
-              'state' => user.va_profile.address.state,
-              'country' => user.va_profile.address.country,
-              'postalCode' => user.va_profile.address.postal_code[0..4]
+              'city' => user.address.city,
+              'state' => user.address.state,
+              'country' => user.address.country,
+              'postalCode' => user.address.postal_code[0..4]
```
bosawt, line 215: I suppose if you're gonna `slice` the one above, may as well do the same here

bosawt (approved):
Thanks, new changes look good!

PR #5998, opened 2021-02-23
[department-of-veterans-affairs/notification-api#346] add missing api_key

saman-moshafi-va:
api_key for light house was missing in settings.yml file

## Description of change
PR 5962 enhanced VANotify::Service so each business line can have a dedicated api_key.  The api_key for lighthouse was missing 

## Original issue(s)
department-of-veterans-affairs/notification-api#346

tested on the review instance


config/settings.yml
```diff
@@ -797,8 +797,10 @@ vanotify:
         form526_confirmation_email: fake_template_id
         covid_vaccine_registration: fake_template_id
     lighthouse:
-      disconnection_template: fake_template_id
-      connection_template: fake_template_id
+      api_key: fake_secret
+      template_id:
```
ffafara-tw, line 801: We discussed changing it to templates.
Is it in separate PR?

saman-moshafi-va: No, We have to modify devops as well. I think just modifying the devops template_id to templates will break vets-api aocde unless we use a new feature flag.  I can do it, but it might mean two PRs on vets-api and one PR in devops.

ericboehs (approved):

PR #5999, opened 2021-02-23
Strip extra quotes from SM attachment filename

narin:
## Description of change

This PR strips out extra quotes that may have been included in the filename of a Secure Messaging attachment, and updates the mock response we are using for our specs.

MHV appears to be providing us the value of the filename with extra quotes, and then we also add quotes, so it's winding up as:
`Content-Disposition: attachment; filename="%22test.pdf%22"`

## Original issue(s)
department-of-veterans-affairs/va.gov-team#20188


spec/support/vcr_cassettes/sm_client/messages/nested_resources/gets_a_single_attachment_by_id.yml
```diff
@@ -24,7 +24,7 @@ http_interactions:
       Server:
       - Apache-Coyote/1.1
       Content-Disposition:
-      - attachment; filename=noise300x200.png
+      - attachment; filename=%22noise300x200.png%22
```
narin, line 27: Wasn't sure whether to set this as `filename=%22noise300x200.png%22` or `filename="noise300x200.png"`, but the replace code I added covers both cases.

ericboehs (approved):

PR #6002, opened 2021-02-24
19463 pre-cache mobile appointments

kreek:
## Description of change
The appointments list is currently being pre-cached in the mobile app on load even if the user never goes to that screen. This PR pre-caches appointments (for the duration of a user session) in the background after the user profile is fetched. Then when the user navigates to the appointments screen and makes an API call we fetch appointments from Redis rather than the VAOS service. Note pre-cached appointments are only returned once, and can always be skipped with the `useCache` flag set to false.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#19463


jperk51 (approved):
LGTM

ericboehs (approved):

PR #6004, opened 2021-02-24
API-4997 - remove old settings

nathanbwright:
## Description of change
Removes old setting that were needed for backwards compatibility while devops was being updated.

## Original issue(s)
https://vajira.max.gov/browse/API-4997


yakschuss (approved):
LGTM!

PR #6005, opened 2021-02-24
API-5338 - Provide better error when v0 form2122 endpoints can't find POA

jw81:
## Description of change

In `/claims/v0/forms/2122/active` and `/claims/v0/forms/2122/{id}/status`, it's possible that a Power of Attorney (POA) cannot be found.  When that happens, we need to return a 404 rather than allowing the code to blow up and return a 500.

## Original issue(s)
[API-5338](https://vajira.max.gov/browse/API-5338)

### DIFF
--- a/modules/claims_api/app/swagger/claims_api/v0/form_2122_controller_swagger.rb
+++ b/modules/claims_api/app/swagger/claims_api/v0/form_2122_controller_swagger.rb
@@ -351,6 +351,7 @@ module ClaimsApi
             key :name, 'X-VA-SSN'
             key :in, :header
             key :description, 'SSN of Veteran being represented'
+            key :example, '123121234'
             key :required, true
             key :type, :string
           end
@@ -359,6 +360,7 @@ module ClaimsApi
             key :name, 'X-VA-First-Name'
             key :in, :header
             key :description, 'First Name of Veteran being represented'
+            key :example, 'John'
             key :required, true
             key :type, :string
           end
@@ -367,6 +369,7 @@ module ClaimsApi
             key :name, 'X-VA-Last-Name'
             key :in, :header
             key :description, 'Last Name of Veteran being represented'
+            key :example, 'Doe'
             key :required, true
             key :type, :string
           end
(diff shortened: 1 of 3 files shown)


modules/claims_api/app/swagger/claims_api/v0/form_2122_controller_swagger.rb
```diff
@@ -367,6 +369,7 @@ class Form2122ControllerSwagger
             key :name, 'X-VA-Last-Name'
             key :in, :header
             key :description, 'Last Name of Veteran being represented'
+            key :example, 'Doe'
```
SlakrHakr, line 372: These example values are incorrect I believe. These values should be of a Veteran being represented. John Doe is not a Veteran but is representing several Veterans. This could be any of the following instead: 001, 002, 003, 004, 005, and 006 (Tamara Ellis, Janet Moore, Ralph Lee, Jess Gray, Pauline Foster, and Russell Freeman)

source for above example values: https://github.com/department-of-veterans-affairs/vets-api-clients/blob/master/test_accounts.md#representative

modules/claims_api/app/swagger/claims_api/v0/form_2122_controller_swagger.rb
```diff
@@ -468,47 +465,53 @@ class Form2122ControllerSwagger
...
 
           parameter do
-            key :name, 'X-VA-User'
+            key :name, 'X-VA-LOA'
             key :in, :header
-            key :description, 'VA username of the person making the request'
-            key :required, false
+            key :description, 'The level of assurance of the user making the request'
+            key :example, '3'
+            key :required, true
```
SlakrHakr, line 514: nice making all these required for v0 💯

SlakrHakr (changes requested):
Logic looks great to me, its really just the example header values that need to be changed I think.

modules/claims_api/app/swagger/claims_api/v0/form_2122_controller_swagger.rb
jw81: Updated [here](https://github.com/department-of-veterans-affairs/vets-api/pull/6005/commits/b070d39b2ab82cbb2ed5c2d39f9c9c63704d6b31).

SlakrHakr (approved):

PR #6006, opened 2021-02-24
fixed date filed null for nonlegacy claims

jperk51:
## Description of change
Need to key off of different event for non legacy appeals. Also added one HLR and Sup claim to appeals yml since it was only legacyAppeals. 

## Original issue(s)
N/A

Just the above


kreek (approved):
LGTM 🚀

PR #6011, opened 2021-02-25
Adds and documents "hybrid mode" vets-api setup.

patrickvinograd:
## Description of change
Adds instructions and a docker-compose file for running Postgres
and Redis in docker but running vets-api natively.

This is a setup that works well for me and I thought others might benefit. Happy to have either the documentation or the setup itself altered if people see an opportunity for streamlining. 

* It's possible to map Redis and Postgres straight through on their default ports, but I figured if people already had either of those things running natively it would conflict and generate support questions, so I went with `default port * 10`
* If redis is mapped to its default port, then no settings.local.yml stanza is needed. A local configuration *is* still needed for Postgres because the default settings tries to connects on a unix domain socket which doesn't work with Docker without some extra doing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,10 +48,11 @@ For frontend, see [vets-website](https://github.com/department-of-veterans-affai
 
    **DO NOT commit local Gemfile modifications that remove the `sidekiq-ent` and `sidekiq-pro` gems.**
 
-1. Developers who work with vets-api daily tend to prefer the native setup because they don't have to deal with the abstraction of docker-compose while those who would to spend less time on getting started prefer the docker setup. Docker is also useful when it's necessary to have a setup as close to production as possible.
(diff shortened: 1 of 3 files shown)


omgitsbillryan (approved):
Seems like a win to me. 

Docker can be a headache when trying to rapidly test ruby code / rails server - which is why I think a lot of BE devs avoid it. For the tooling around rails however, it's more "set it and forget it", so docker makes more sense.

I could even see this being extended to common development tools. I have `redis-commander` installed locally for example, but it's kind of a pain to manage.

LindseySaari (dismissed):
Awesome, thanks for this! We may wrap this into the toolkit for the future setup command. 👍

patrickvinograd:
@LindseySaari @omgitsbillryan this fell off my radar, I force-pushed to get vets-api latest master rebased so would need another review to merge this in.

LindseySaari (approved):

PR #6014, opened 2021-02-25
API-5015: Audit Special Issues/Flashes Enum values for EVSS and BGS

SlakrHakr:
## Description of change
Maps special issues sent to us to values EVSS expects. This change also removes any special issues EVSS does not accept from our payload to EVSS. (Those additional values will be added through BGS in an existing subsequent background job)

## Original issue(s)
https://vajira.max.gov/browse/API-5015

Tested locally and added specs


SlakrHakr:
Once this pr: https://github.com/department-of-veterans-affairs/vets-api/pull/6013 gets into master, this pr here should be able to just be rebased from master and merged in.

jw81 (approved):

PR #6019, opened 2021-02-25
Fetch appointments from lighthouse

dillo:
## Description of change

- This PR provides the `health-quest` module the ability to fetch a list of `appointments` for a `patient` from the `Lighthouse Health API`.
## Original issue(s)
department-of-veterans-affairs/va.gov-team#19656

- [x] Feature Flag: show_healthcare_experience_questionnaire

- [x] RSpec test suite passing

### DIFF
--- a/modules/health_quest/app/controllers/health_quest/v0/lighthouse_appointments_controller.rb
+++ b/modules/health_quest/app/controllers/health_quest/v0/lighthouse_appointments_controller.rb
@@ -4,11 +4,22 @@ module HealthQuest
   module V0
     class LighthouseAppointmentsController < HealthQuest::V0::BaseController
       def index
-        render json: []
+        render json: factory.search(request.query_parameters).response[:body]
       end
 
       def show
         render json: {}
+      end
+
+      private
+
+      def appointments_params
+        params.permit!
+      end
+
+      def factory
+        @factory =
+          HealthQuest::HealthApi::Appointment::Factory.manufacture(current_user)
       end
     end
   end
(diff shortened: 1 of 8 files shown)


modules/health_quest/app/controllers/health_quest/v0/lighthouse_appointments_controller.rb
```diff
@@ -4,12 +4,23 @@ module HealthQuest
...
       end
 
       def show
         render json: {}
       end
+
+      private
+
+      def appointments_params
+        params.permit!
```
ericboehs, line 17: Are we sure we want to permit all params here?

dillo: > Are we sure we want to permit all params here?

Good catch! I was going to redo them separately in PR and handle the `show` action as well but I think I'll make the change here and now. Thank you!

dillo: @ericboehs I've modified the allowed strong params. This will be used for the upcoming PR which implements the full functionality for the `show` action. The index action can accept anything from the `request.query_parameters` call; We'll be cherry-picking the parameters which will get used in the downstream modules and classes.

On a side note, I will be creating a PR or two down the line, when this feature is nearing code complete, which will perform security sweeps on the health-quest module and tie up any loose ends like the example above.

ericboehs (approved):

PR #6023, opened 2021-02-25
fixes for claims overview updateAt dates

jperk51:
## Description of change
Changed which values our updateAt field is based on for claims and appeals. This change puts us in line with .gov

## Original issue(s)
N/A

Fixing some bad assumptions from the initial implementation.


kreek (approved):
LGTM 🚀

PR #6027, opened 2021-02-25
Handle partial results returned from MAS; Fixes #20483

saneshark:
## Description of change
Add errors partial to meta. Updates schema definitions for both va_appointments and cc_appointments schemas. Will always include an errors array even when no elements in the array. When errors are present http status will become 207 Multi Status, essentially successful with partial errors. 

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/20483

### DIFF
--- a/modules/vaos/app/controllers/vaos/v0/appointments_controller.rb
+++ b/modules/vaos/app/controllers/vaos/v0/appointments_controller.rb
@@ -8,7 +8,7 @@ module VAOS
       before_action :validate_params, only: :index
 
       def index
-        if appointments[:meta][:errors].any?
+        if appointments[:meta][:errors]&.any?
           render json: each_serializer.new(appointments[:data], meta: appointments[:meta]), status: 207
         else
           render json: each_serializer.new(appointments[:data], meta: appointments[:meta]), status: 200
--- a/modules/vaos/app/controllers/vaos/v0/appointments_controller.rb
+++ b/modules/vaos/app/controllers/vaos/v0/appointments_controller.rb
@@ -8,7 +8,11 @@ module VAOS
       before_action :validate_params, only: :index
 
       def index
-        render json: each_serializer.new(appointments[:data], meta: appointments[:meta])
+        if appointments[:meta][:errors].any?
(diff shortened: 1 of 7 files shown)


modules/vaos/app/services/vaos/appointment_service.rb
```diff
@@ -102,6 +94,20 @@ def pagination(pagination_params)
       }
     end
 
+    def partial_errors(response)
+      if response.status == 200 && response.body[:errors]
```
ericboehs, line 98: Should this be `response.body[:errors]&.any?`?

modules/vaos/app/controllers/vaos/v0/appointments_controller.rb
```diff
@@ -8,7 +8,11 @@ class AppointmentsController < VAOS::V0::BaseController
       before_action :validate_params, only: :index
 
       def index
-        render json: each_serializer.new(appointments[:data], meta: appointments[:meta])
+        if appointments[:meta][:errors].any?
```
ericboehs, line 11: Do we need to nil check this?

modules/vaos/app/services/vaos/appointment_service.rb
saneshark: no, because VAMF actually drops null valued keys. However at the controller level when check to make see if there are errors, we're always passing an empty array if VAMF provided us with null. There we check for .any

but to your point, &.any? would work just as well.

modules/vaos/app/controllers/vaos/v0/appointments_controller.rb
saneshark: No, because in the service we're ensuring that it's always an array if VAMF provided us with nil. ` || [] `

grossb1 (approved):
LGTM

modules/vaos/app/services/vaos/appointment_service.rb
saneshark: made this change anyway, because why not?

PR #6028, opened 2021-02-25
20164 mobile add canUpdatePayment flag to payment information response

kreek:
## Description of change
Adds a canUpdatePayment flag to the mobile payment information response. I was originally going to add this in meta data but there's similar flags in the response so I added it there instead. Current spec user had permission to update so I added a cassette for a user without permission.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#20164


jperk51 (approved):
LGTM

PR #6031, opened 2021-02-26
19656 get lighthouse appointment

dillo:
## Description of change

- This PR gives the health-quest module the ability to fetch a single appointment resource by its ID from the Lighthouse Health API.
## Original issue(s)
department-of-veterans-affairs/va.gov-team#19656

- [x] Feature Flag: show_healthcare_experience_questionnaire

- [x] RSpec tests are written and the entire suite is passing

### DIFF
--- a/modules/health_quest/app/controllers/health_quest/v0/lighthouse_appointments_controller.rb
+++ b/modules/health_quest/app/controllers/health_quest/v0/lighthouse_appointments_controller.rb
@@ -8,7 +8,7 @@ module HealthQuest
       end
 
       def show
-        render json: {}
+        render json: factory.get(params[:id]).response[:body]
       end
 
       private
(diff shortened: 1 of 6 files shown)


modules/health_quest/app/controllers/health_quest/v0/lighthouse_appointments_controller.rb
```diff
@@ -8,15 +8,11 @@ def index
...
       def show
-        render json: {}
+        render json: factory.get(params[:id]).response[:body]
       end
 
       private
 
-      def appointment_params
-        params.permit(:id)
-      end
```
dillo, line 18: Removing this method since the controller is not assigning attributes to a model and the `show` action accesses the only `:id` param directly.

ericboehs (approved):

PR #6033, opened 2021-02-26
adds idme_uuid column to tud_accounts table

bramleyjl:
## Description of change

updates `tud_accounts` table to add a column for holding an account's `idme_uuid`

## Original issue(s)
department-of-veterans-affairs/va.gov-team#18967


modules/test_user_dashboard/db/migrate/20210226193901_add_idme_uuid_to_tud_accounts.rb
```diff
@@ -0,0 +1,5 @@
+class AddIdmeUuidToTudAccounts < ActiveRecord::Migration[6.0]
+  def change
+    add_column :test_user_dashboard_tud_accounts, :idme_uuid, :string
```
ericboehs, line 3: Should this be `:uuid` type instead of `:string`?

bramleyjl: Good point, changed the type

ericboehs (approved):

PR #6034, opened 2021-02-27
Notif changes for readability and performance

bradenshipley:
## Description of change

This PR addresses a couple of typos in the `apps_api` module in the README, and adds some event validation to the notifications_service. It seems that calling the Systems Log api for Okta has multiple duplicate events when a user connects or disconnects an application, so a simple hash check has been added to get around that. Also, the `published` date has been formatted into a readable date so that the email template doesn't show just an ISO date for the event time. The `NotificationService` unit tests have been updated where needed to reflect these changes.

## Original issue(s)


modules/apps_api/lib/apps_api/notification_service.rb
```diff
@@ -46,17 +48,18 @@ def parse_event(event)
...
+    def create_hash(app_record:, user:, published:, uuid:)
       {
+        'uuid' => uuid,
         'app_record' => app_record,
         'user_email' => user.body['profile']['email'],
         'options' => {
           'first_name' => user.body['profile']['firstName'],
           'application' => app_record ? app_record['name'] : nil,
-          'time' => published,
+          'time' => Time.zone.parse(published).strftime('%m/%d/%Y at %T:%M%p'),
```
yakschuss, line 62: This could probably be extracted to a private method for readability. What does this time represent?

modules/apps_api/lib/apps_api/notification_service.rb
```diff
@@ -46,17 +48,18 @@ def parse_event(event)
         app_record = DirectoryApplication.find_by(name: event['actor']['displayName'])
       end
       user = @okta_service.user(user_id)
-      create_hash(app_record: app_record, user: user, published: event['published'])
+      create_hash(app_record: app_record, user: user, published: event['published'], uuid: event['uuid'])
     end
 
-    def create_hash(app_record:, user:, published:)
+    def create_hash(app_record:, user:, published:, uuid:)
```
yakschuss, line 54: what kind of hash is this? `create_event_record`?

modules/apps_api/lib/apps_api/notification_service.rb
```diff
@@ -65,16 +68,20 @@ def create_hash(app_record:, user:, published:)
     end
 
     def event_is_invalid(event)
-      event['outcome']['result'] != 'SUCCESS' ||
+      # checking if the event is unable to be processed,
+      # or has already been processed.
+      @handled_events.include?(event['uuid']) ||
```
yakschuss, line 73: my gut says that extracting compound conditionals here would make this logic easier to parse. What do you think?

LindseySaari: ➕

modules/apps_api/lib/apps_api/notification_service.rb
LindseySaari: ➕

modules/apps_api/lib/apps_api/notification_service.rb
```diff
@@ -46,35 +48,53 @@ def parse_event(event)
...
       }
     end
 
+    def format_published_time(published)
+      # Formats iso8601 time stamp into readable language
+      # 2020-11-29T00:23:39.508Z -> 11/29/2020 at 00:23:39:23AM
+      Time.zone.parse(published).strftime('%m/%d/%Y at %T:%M%p')
+    end
+
     def event_is_invalid(event)
```
SlakrHakr, line 76: nit: event_is_invalid?
since this is asking a question and the return is always either true or false

secondary nit: this might be just personal preference, but i almost always think the positive = true concept works better. so, "event_is_valid?" instead of "event_is_invalid?"

SlakrHakr: sort of the same thing with several of these methods below as well. ?'s at the end and positive "event_successful?" instead of "event_unsuccessful?"

bradenshipley: i tend to always write these helper method as a way to filter out failures, rather than to confirm successes, if that makes sense. I think it has to do with preferring the `unless foo_is_bad, continue` method over the `if foo_is_good, continue` way, if that makes sense. I'm with you on changing the method name to include ? though all the way

modules/apps_api/lib/apps_api/notification_service.rb
bradenshipley: extracted in latest commit

modules/apps_api/lib/apps_api/notification_service.rb
bradenshipley: extracted to separate methods in latest commit

yakschuss (approved):
Looks good to me

modules/apps_api/lib/apps_api/notification_service.rb
SlakrHakr: fair enough, the positive `if foo_is_good, continue` makes a ton more sense to my brain but definitely a nitpick.

PR #6040, opened 2021-03-01
Api 5225/evidence upload status endpoint

yakschuss:
[API-5225](https://vajira.max.gov/browse/API-5225)

This PR adds support for `EvidenceSubmission` model, as well as retrieving the status of the EvidenceSubmission.

NOTE: This PR will need a migration run for EvidenceSubmission to work correctly.

[Migration PR](https://github.com/department-of-veterans-affairs/vets-api/pull/6018) ( #6018 )


modules/appeals_api/app/serializers/appeals_api/v1/evidence_submission_serializer.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require 'fast_jsonapi'
+
+module AppealsApi
+  module V1
```
cilestin, line 6: Where is this being used? I see that `modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb` is calling `AppealsApi::EvidenceSubmissionSerializer.new(submissions)`. Should it be calling the V1 version instead?

But if that's the case then why is the non-V1 around? One of these serializers seems to be unused 🤔

cilestin (approved):
Looking good except for what I suspect is an accidental additional serializer file. If it just needs removed, then the approval stands once it's gone. If not, let me know and I'll give it another look.

modules/appeals_api/app/serializers/appeals_api/v1/evidence_submission_serializer.rb
yakschuss: Ah, I think that was supposed to be removed. Must've missed it! Good catch.

PR #5997, opened 2021-02-23
Add endpoint for dependency verification

digitaldrk:
## Description of change
This is an endpoint that delivers 'Diaries' which are vet's dependent award statuses. Below is an example of a record this returns within a colleciton:

```
{
    "data": {
        "id": "",
        "type": "dependency_decs",
        "attributes": {
            "dependency_verifications": [
                {
                    "award_effective_date": "2016-06-01T00:00:00.000-05:00",
                    "award_event_id": "185878",
                    "award_type": "CPL",
                    "begin_award_event_id": "171629",
                    "beneficiary_id": "13367440",
                    "birthday_date": "1995-05-01T00:00:00.000-05:00",
                    "decision_date": "2015-02-24T10:50:42.000-06:00",
                    "decision_id": "125932",
                    "dependency_decision_id": "53153",
                    "dependency_decision_type": "SCHATTT",
                    "dependency_decision_type_description": "School Attendance Terminates",
                    "dependency_status_type": "NAWDDEP",
                    "dependency_status_type_description": "Not an Award Dependent",
                    "event_date": "2016-05-01T00:00:00.000-05:00",
                    "first_name": "JAMES",
                    "full_name": "JAMES E. SMITH",
                    "last_name": "SMITH",
                    "middle_name": "E",
                    "modified_action": "I",
                    "modified_by": "VHAISWWHITWT",
                    "modified_date": "2015-02-24T10:50:41.000-06:00",
                    "modified_location": "317",
                    "modified_process": "Awds/RBA-cp_dependdec_pkg.do_cre",
                    "person_id": "600086386",
                    "sort_date": "2015-02-24T10:50:42.000-06:00",
                    "sort_order_number": "0",
                    "veteran_id": "13367440",
                    "veteran_indicator": "N"
                }
            ]
        }
    }
}
```

## Things to know about this PR
* Are there additions to a `settings.yml` file? Do they vary by environment?
  * No changes
* Is there a feature flag? What is it?
  * No FF but this will not be used in prod for a while
* Is there some Sentry logging that was added? What alerts are relevant?
  * none added
* Are there any Prometheus metrics being collected? What Grafana dashboard were they added do?
  * no Grafana stuff added
* Are there Swagger docs that were updated?
  * none added
* Is there any PII concerns or questions?
  * I am removing the the SSN of the dependent in the serializer. Is that a concern?

PR #6042, opened 2021-03-01
Permanent fix viewport spec

digitaldrk:
## Description of change
While working on [this PR](https://github.com/department-of-veterans-affairs/vets-api/pull/5997), I hit [a failing spec](https://github.com/department-of-veterans-affairs/vets-api/pull/5997/files#diff-538bb209a64e00cdefa341e0341fb6d8a97e0c1f83c4d5c6bbf617557b2c93ebL69).

Being in a hurry and considering it was a static string already, I changed the dates with the intent to make a more permanent solution in a separate PR. @bosawt HMU on slack asking about it and this is the PR from that conversation.

Trevor suggested `Timecop.freeze(Time.zone.parse('2021-03-01')` but I couldn't get that to work.

I'm not _wild_ about my solution. Seems to just do what the code is doing that it's supposed to test. I added another expectation as another option.

## Original issue(s)
No issue for this

  * No tests
  * no FF
  * Nope
  * no metrics
  * no updates
  * no concerns

### DIFF
--- a/spec/jobs/cypress_viewport_updater/viewport_spec.rb
+++ b/spec/jobs/cypress_viewport_updater/viewport_spec.rb
@@ -66,7 +66,10 @@ RSpec.describe CypressViewportUpdater::Viewport do
 
   describe '#percentTrafficPeriod' do
     it 'returns the correct value' do
-      expect(@viewport.percentTrafficPeriod).to eq('From: 02/01/2021, To: 02/28/2021')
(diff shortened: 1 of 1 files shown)


bosawt (approved):
Ah, I was wondering why `Timecop` wasn't working for you. It looks like I was wrong about the `it` block, if you put the `Timecop.freeze` in the top level of the file in a `before` block, then it works.

On the other hand, I think your solution is overall better, makes it more clear exactly what we're testing for here. Looks good!

ericboehs (approved):

PR #6043, opened 2021-03-01
API-5382 Return service history as array type, add 401 status code

bdgrider-va:
## Description of change
Specifying in the documentation that service history endpoint returns an array of ServiceHistoryEpisode objects. Adding 401 unauthorizes status code response to all veteran verification endpoints. 

## Original issue(s)
https://vajira.max.gov/browse/API-5382

Veteran Verification swagger docs were updated.

Verified that the documentation was parsing and displaying correctly in the editor.swagger.io webpage.

### DIFF
--- a/modules/veteran_verification/VETERAN_VERIFICATION.yml
+++ b/modules/veteran_verification/VETERAN_VERIFICATION.yml
@@ -89,9 +89,7 @@ paths:
                   data:
                     $ref: '#/components/schemas/VeteranStatusConfirmation'
         '401':
-          description: Unauthorized.
-        '403':
-          description: Provided Access Token Does not have the required scopes to access.
+          description: Not authorized
         '502':
           description: eMIS failed to respond or responded in a way we cannot handle.
   /disability_rating:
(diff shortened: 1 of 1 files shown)


modules/veteran_verification/VETERAN_VERIFICATION.yml
```diff
@@ -88,6 +88,8 @@ paths:
                 properties:
                   data:
                     $ref: '#/components/schemas/VeteranStatusConfirmation'
+        '401':
+          description: Unauthorized.
```
dbrown-va, line 92: For consistency with the returned error, recommend updating to `Not authorized` for the 401 descriptions.

modules/veteran_verification/VETERAN_VERIFICATION.yml
```diff
@@ -88,6 +88,8 @@ paths:
...
         '403':
```
dbrown-va, line 93: I just tested out this scenario and it actually is returning a 401 Not Authorized as well if the required scope isn't provided.

Recommend removing the 403 block.

dbrown-va (approved):

PR #6044, opened 2021-03-01
Api 4960 - update statuses

nathanbwright:
## Description of change
Makes HLR and NOD statuses separate, along with how the central mail statuses map to them. Updates to the docs to provide more documentation around what each of the statuses mean.

## Original issue(s)


modules/appeals_api/app/models/concerns/appeals_api/nod_status.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+require 'appeals_api/central_mail_updater'
+
+module AppealsApi
+  module NodStatus
```
yakschuss, line 6: if we wanted `NODStatus`, we'd have to name the file something like `n_o_d_status`, right? Is there a way to inflect that?

yakschuss: I guess it's probably not worth flouting "convention over configuration"

modules/appeals_api/lib/appeals_api/central_mail_updater.rb
```diff
@@ -64,7 +79,7 @@ def parse_central_mail_response(raw_response)
     end
 
     def update_appeal_status!(appeal, status)
```
yakschuss, line 81: with the method chain `status.status`, it's not immediately obvious what this object being passed in. Is there another name for it?

modules/appeals_api/app/models/concerns/appeals_api/nod_status.rb
nathanbwright: Hmm, I'll look into this...

modules/appeals_api/lib/appeals_api/central_mail_updater.rb
nathanbwright: Yeah, that is a good point. Thanks!

modules/appeals_api/app/models/concerns/appeals_api/nod_status.rb
nathanbwright: If we would make an entry for `NOD` here: https://github.com/department-of-veterans-affairs/vets-api/blob/master/config/initializers/inflections.rb we could use `NODStatus` and still have the file name `nod_status`... Do you think it would be worth it?

cilestin (approved):

cilestin (approved):
Nice! 🚀

PR #6047, opened 2021-03-02
Fix Secure Messaging POSTs

narin:
## Description of change
Secure Messaging `POST` requests that include a body such as sending a message and replying to a message were not being sent correctly because args were being passed as `ActionController:Parameters` instead of hashes.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#20189

May need additional changes to support file uploads

### DIFF
--- a/lib/sm/client.rb
+++ b/lib/sm/client.rb
@@ -210,7 +210,7 @@ module SM
     def post_create_message(args = {})
       validate_create_context(args)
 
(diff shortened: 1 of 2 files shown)


patrickvinograd:
I think the fact that we are setting the content-type header to multipart/form-data means that file attachments will work as expected, per: https://lostisland.github.io/faraday/middleware/multipart

patrickvinograd (approved):
LGTM

PR #6056, opened 2021-03-02
Add basic upload endpoint

ElleeB:
Adding basic NOD evidence submission upload endpoint
It should:
- take in a document and uuid
- validate document
- return meaningful response if document is invalid

Ticket
https://vajira.max.gov/browse/API-5221

### DIFF
--- a/modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
+++ b/modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
@@ -23,7 +23,7 @@ module AppealsApi::V1
           if status == :ok
             render json: { document: params[:document], uuid: params[:uuid] }
           else
-            render json: error, status: 422
+            render json: { errors: [error] }, status: 422
           end
         end
       end
--- a/modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
+++ b/modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
@@ -5,6 +5,7 @@ module AppealsApi::V1
     module NoticeOfDisagreements
       class EvidenceSubmissionsController < AppealsApi::ApplicationController
         skip_before_action :authenticate
+
         def show
           submissions = AppealsApi::EvidenceSubmission.where(
             supportable_id: params[:id],
@@ -14,6 +15,10 @@ module AppealsApi::V1
           serialized = AppealsApi::EvidenceSubmissionSerializer.new(submissions)
 
           render json: serialized.serializable_hash
+        end
+
+        def upload
+          render json: { message: 'Good work.', document: params[:document], uuid: params[:uuid] }
         end
       end
     end
--- a/modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
+++ b/modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
@@ -17,8 +17,14 @@ module AppealsApi::V1
           render json: serialized.serializable_hash
         end
 
-        def upload
-          render json: { message: 'Good work.', document: params[:document], uuid: params[:uuid] }
+        def create
+          status, error = AppealsApi::FileValidator.new(params[:document]).call
+
+          if status == :ok
+            render json: { document: params[:document], uuid: params[:uuid] }
+          else
+            render json: error, status: 422
+          end
         end
       end
     end
--- a/modules/appeals_api/config/routes.rb
+++ b/modules/appeals_api/config/routes.rb
@@ -22,11 +22,7 @@ AppealsApi::Engine.routes.draw do
       end
       namespace :notice_of_disagreements do
         get 'contestable_issues', to: 'contestable_issues#index'
-        resources :evidence_submissions, only: %i[show] do
-          collection do
-            post 'upload'
-          end
-        end
+        resources :evidence_submissions, only: %i[create show]
       end
       resources :notice_of_disagreements, only: %i[create show] do
         collection do
--- a/modules/appeals_api/config/routes.rb
+++ b/modules/appeals_api/config/routes.rb
@@ -22,12 +22,16 @@ AppealsApi::Engine.routes.draw do
       end
       namespace :notice_of_disagreements do
         get 'contestable_issues', to: 'contestable_issues#index'
-        resources :evidence_submissions, only: %i[show]
+        resources :evidence_submissions, only: %i[show] do
+          collection do
+            post 'upload'
+          end
+        end
       end
       resources :notice_of_disagreements, only: %i[create show] do
         collection do
-          get 'schema', to: 'notice_of_disagreements#schema'
-          post 'validate', to: 'notice_of_disagreements#validate'
+          get 'schema'
+          post 'validate'
         end
       end
     end
--- a/modules/appeals_api/spec/requests/v1/notice_of_disagreements/evidence_submissions_controller_spec.rb
+++ b/modules/appeals_api/spec/requests/v1/notice_of_disagreements/evidence_submissions_controller_spec.rb
@@ -4,19 +4,15 @@ require 'rails_helper'
 require AppealsApi::Engine.root.join('spec', 'spec_helper.rb')
 
 describe AppealsApi::V1::DecisionReviews::NoticeOfDisagreements::EvidenceSubmissionsController, type: :request do
+  include FixtureHelpers
+
   let(:notice_of_disagreement) { create(:notice_of_disagreement) }
   let(:evidence_submissions) { create_list(:evidence_submission, 3, supportable: notice_of_disagreement) }
-
-  def base_path(path)
-    "/services/appeals/v1/decision_reviews/notice_of_disagreements/evidence_submissions/#{path}"
-  end
-  # let(:path) { '/services/appeals/v1/decision_reviews/notice_of_disagreements/evidence_submissions/' }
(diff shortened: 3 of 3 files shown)


modules/appeals_api/config/routes.rb
```diff
@@ -22,11 +22,16 @@
...
         get 'contestable_issues', to: 'contestable_issues#index'
+        resources :evidence_submissions do
+          collection do
+            post 'upload'
+          end
+        end
       end
       resources :notice_of_disagreements, only: %i[create show] do
         collection do
-          get 'schema', to: 'notice_of_disagreements#schema'
```
ElleeB, line 28: This shouldn't be necessary so bye. (Check me on this)

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
```diff
@@ -15,6 +16,19 @@ def show
 
           render json: serialized.serializable_hash
         end
+
+        def upload
+          status, error = AppealsApi::FileValidator.new(params[:document]).call
+
+          if status == :ok
+            @document = params[:document]
```
yakschuss, line 24: What is the purpose of having these variables be ivars?

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
```diff
@@ -15,6 +16,19 @@ def show
 
           render json: serialized.serializable_hash
         end
+
+        def upload
```
yakschuss, line 20: instead of upload, could this be "#create"? If we really want the endpoint to be 'upload', we can always do a `as: 'upload`. Just to keep it RESTful. What do you think?

modules/appeals_api/config/routes.rb
```diff
@@ -22,12 +22,16 @@
       end
       namespace :notice_of_disagreements do
         get 'contestable_issues', to: 'contestable_issues#index'
-        resources :evidence_submissions, only: %i[show]
+        resources :evidence_submissions, only: %i[show] do
+          collection do
```
yakschuss, line 26: Is this a collection? Are we set up to handle multiple documents at a time? Perhaps we should require the client to submit a single request per document?

modules/appeals_api/spec/requests/v1/notice_of_disagreements/evidence_submissions_controller_spec.rb
```diff
@@ -6,24 +6,51 @@
...
   let(:notice_of_disagreement) { create(:notice_of_disagreement) }
   let(:evidence_submissions) { create_list(:evidence_submission, 3, supportable: notice_of_disagreement) }
-  let(:path) { '/services/appeals/v1/decision_reviews/notice_of_disagreements/evidence_submissions/' }
+
+  def base_path(path)
+    "/services/appeals/v1/decision_reviews/notice_of_disagreements/evidence_submissions/#{path}"
+  end
 
   describe '#show' do
+    let(:path) { base_path notice_of_disagreement.id }
```
yakschuss, line 15: I like this, nice!

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
ElleeB: draft-mode nonsense...unnecessary.

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
ElleeB: Yes! I didn't context switch when i moved from the nod controller to the submission controller. I'll change it now.

modules/appeals_api/config/routes.rb
ElleeB: ^ this was me misunderstanding collection it seems. I'll refactor

modules/appeals_api/spec/requests/v1/notice_of_disagreements/evidence_submissions_controller_spec.rb
ElleeB: I can't take credit lol. It's all over the codebase ;)

ElleeB: ...and I took it out after changing upload to create lol

modules/appeals_api/spec/requests/v1/notice_of_disagreements/evidence_submissions_controller_spec.rb
```diff
@@ -26,4 +26,25 @@
       expect(body).to match_array(serialized)
     end
   end
+
+  describe '#create' do
+    let(:fixtures_path) { '/modules/appeals_api/spec/fixtures' }
+    let(:uploaded_file) { Rack::Test::UploadedFile }
+    let(:valid_upload_params) do
+      { document: uploaded_file.new("#{::Rails.root}#{fixtures_path}/expected_10182_extra.pdf"), uuid: '1234' }
```
yakschuss, line 34: Only thing I'd say is to pull this root/fixtures_path out, but looks good otherwise!

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
```diff
@@ -15,6 +16,16 @@ def show
 
           render json: serialized.serializable_hash
         end
+
+        def create
+          status, error = AppealsApi::FileValidator.new(params[:document]).call
+
+          if status == :ok
+            render json: { document: params[:document], uuid: params[:uuid] }
```
yakschuss, line 24: Have we confirmed that params[:document] is enough for the document to hit the server/be validated?

modules/appeals_api/spec/requests/v1/notice_of_disagreements/evidence_submissions_controller_spec.rb
ElleeB: Ugh. I was doing this to avoid getting pulled over by rubocop. I can't stand forcing things on a million different lines if I don't have to, so I went with this. I still prefer this to multiple lines, but I could be persuaded...

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
ElleeB: It runs through the validation. Given the basic nature of this endpoint, the shape of the eventual and final payload will be hashed out in more detail later I'm assuming. Either way, whatever we take in, must have a document that we can pull out and send to the validator, so that's how I wrote the spec. Am I understanding your question?

ElleeB: In relation, maybe the uuid will be in the headers, but since I don't think we're entirely sure exactly what we want that to look like, this was the solution. If you have a more concrete idea of what we will ask the consumer to provide, please share!

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
```diff
@@ -15,6 +16,16 @@ def show
...
+          else
+            raise Common::Exceptions::UnprocessableEntity.new(detail: error)
```
cilestin, line 26: The error we're getting back from the validator is roughly equivalent to a JSON API ErrorObject already. The `detail` of an ErrorObject (which the Common::Exceptions generally are, IIRC) is supposed to be a human-readable string, so putting the `error` object into it breaks ErrorObject norms. The `error` should already have the `title`, `detail`, and `meta` that we can then mix into an error array. See how DetailedSchemaErrors builds up a hash to turn into a final response - this controller should do the same.

modules/appeals_api/spec/requests/v1/notice_of_disagreements/evidence_submissions_controller_spec.rb
cilestin: Feel free to add some extra helper functions into our spec/support files that make getting the fixture path easier, if they don't already exist.

cilestin (changes requested):

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
ElleeB: I feel like we can do something much more simple, actually. Check out my solution, yo!

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
```diff
@@ -15,6 +16,16 @@ def show
...
           render json: serialized.serializable_hash
         end
+
+        def create
+          status, error = AppealsApi::FileValidator.new(params[:document]).call
+
+          if status == :ok
+            render json: { document: params[:document], uuid: params[:uuid] }
+          else
+            render json: error, status: 422
```
cilestin, line 26: Check out [the error docs on JSON:API](https://jsonapi.org/format/1.0/#errors). As noted:
> ... Error objects *MUST* be returned as an array keyed by errors in the top level of a JSON:API document. ...

```suggestion
            render json: { errors: [error] }, status: 422
```

cilestin (changes requested):
One final bit and then I think we're good!

cilestin (approved):

PR #6057, opened 2021-03-03
Fix NameError exception when running jobs

cilestin:
## Description of change

* Fixes [NameError: uninitialized constant ... CentralMailUpdater](http://sentry.vfs.va.gov/organizations/vsp/issues/26094/?project=3&query=is%3Aunresolved+appealsApi&statsPeriod=14d)

* Sets sane retries for the jobs

## Original issue(s)

https://vajira.max.gov/browse/API-5507

## Notes

Really weird that this wasn't caught by our spec runs. Running the spec file manually with no changes to the code threw the same error right away - not sure why that wasn't picked up by or CI runs. The file appears to be in the correct spot & named appropriately to get run automatically but it either isn't or somehow gets itself into a different autoloading context to have the class loaded properly. Gonna have to investigate that separately.

### DIFF
--- a/modules/appeals_api/app/workers/appeals_api/higher_level_review_upload_status_batch.rb
+++ b/modules/appeals_api/app/workers/appeals_api/higher_level_review_upload_status_batch.rb
@@ -7,7 +7,7 @@ module AppealsApi
     include Sidekiq::Worker
 
     # No need to retry since the schedule will run this every hour
-    sidekiq_options retry: false, unique_until: :success
+    sidekiq_options retry: false
 
     def perform
       return unless enabled? && higher_level_review_ids.present?
--- a/modules/appeals_api/app/workers/appeals_api/higher_level_review_upload_status_batch.rb
+++ b/modules/appeals_api/app/workers/appeals_api/higher_level_review_upload_status_batch.rb
@@ -6,7 +6,8 @@ module AppealsApi
   class HigherLevelReviewUploadStatusBatch
     include Sidekiq::Worker
 
-    sidekiq_options 'retry': true, unique_until: :success
+    # No need to retry since the schedule will run this every hour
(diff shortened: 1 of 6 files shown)


modules/appeals_api/app/workers/appeals_api/higher_level_review_upload_status_batch.rb
```diff
@@ -6,7 +6,8 @@ module AppealsApi
   class HigherLevelReviewUploadStatusBatch
     include Sidekiq::Worker
 
-    sidekiq_options 'retry': true, unique_until: :success
+    # No need to retry since the schedule will run this every hour
+    sidekiq_options retry: false, unique_until: :success
```
yakschuss, line 10: is the `unique_until` necessary if it's `false`?

modules/appeals_api/app/workers/appeals_api/notice_of_disagreement_upload_status_batch.rb
```diff
@@ -6,7 +6,8 @@ module AppealsApi
   class NoticeOfDisagreementUploadStatusBatch
     include Sidekiq::Worker
 
-    sidekiq_options 'retry': true, unique_until: :success
+    # No need to retry since the schedule will run this every hour
+    sidekiq_options retry: false, unique_until: :success
```
yakschuss, line 10: same question

yakschuss (approved):
One question, but LGTM!

modules/appeals_api/app/workers/appeals_api/higher_level_review_upload_status_batch.rb
cilestin: Fix'd :)

nathanbwright (approved):

PR #6061, opened 2021-03-03
20246 locations

dillo:
## Description of change

- This PR gives the health-quest API to fetch Location resources from the lighthouse Health API
- This PR refactors and extracts the abstraction from the services which facilitate the interaction with the lighthouse PGD and Health APIs.
- This PR implements most of the newly refactored code in the `modules/health_quest/app/services/health_quest/fhir` folder
- Unused code and tests have been deleted.
- Test coverage is 90% or greater for all newly written code.
## Original issue(s)
department-of-veterans-affairs/va.gov-team#20246

- [x] RSpec tests written and are passing
- [x] RSpec tests for the entire health-quest module is green
- [x] Manual testing of all involved endpoints are passing and working as intended


modules/health_quest/app/controllers/health_quest/v0/patients_controller.rb
```diff
@@ -4,17 +4,21 @@ module HealthQuest
   module V0
     class PatientsController < HealthQuest::V0::BaseController
       def signed_in_patient
-        render json: factory.get.response[:body]
+        render json: factory.get(current_user.icn).response[:body]
       end
 
       def create
-        render json: factory.create.response[:body]
+        render json: factory.create({}).response[:body]
```
dillo, line 11: Passing in blank params because we create the Patient resource from the logged-in current_user. Also, we don't want anyone passing any parameters in here during a POST call.

LindseySaari:
I think this can be broken up a bit. What about doing a first pass on the new /services (factory classes) that you're introducing. I could see the next pass introducing the new controllers (in separate PRs) and deleting the factories that are no longer used.

PR #6063, opened 2021-03-03
Removed problem tests

lukemarkwordtlibertyits:
Addressing intermittent unit test failures in validation_request_spec

## Original issue(s)


dbrown-va (approved):
Confirmed move of tests from validation_request_spec to the key_service_spec to avoid intermittent test failures with singleton `current_keys` array.

LindseySaari (approved):

PR #6071, opened 2021-03-04
[16582] Refactor MPI GCIds parsing logic to live in a separate lib directory

bosawt:
## Description of change
This PR refactors the GC Ids specific parsing logic that used to live in MPI::Parsers so that it now is located in Identity::Parsers. This was done because we need this parsing in constructing the MPI profile, as well as parsing the SAML payload, so this separates the concerns hopefully with a more well defined line

## Original issue(s)
department-of-veterans-affairs/va.gov-team/issues/16582

- The changes here should not actually have an impact on prod
- Code quality review should be prioritized, this should not have an impact on performance or functionality, purely refactor to hopefully improve code quality technical debt

Testing done:
- Logged in as a user
- Logged in and set a breakpoint at `mpi_profile` in `user.rb
- Confirmed set of MPI data matches expectations
- Confirmed SAML response data (filled in `user.identity` object) matches expectations
- Confirmed SAML response data and MPI data matches each other


lib/identity/parsers/gc_ids_constants.rb
```diff
@@ -0,0 +1,46 @@
...
+  module Parsers
+    module GCIdsConstants
+      # Originating IDs, defines what underlying system the ID has been sourced from
+      VA_ROOT_OID = '2.16.840.1.113883.4.349'
+      DOD_ROOT_OID = '2.16.840.1.113883.3.42.10001.100001.12'
+
+      # The follow set of regex match each full ID format and extract the specific ID value from each
+      # They tend to follow the format: <icn>^<id_type>^<assigning_authority>^<assigning_facility>^<id_state>
+
+      # ICN_REGEX, ex. 16701377^NI^200M^USVHA^A
```
jimiray, line 13: Can we add an empty line in between these to make it a little easier to read?

lib/identity/parsers/gc_ids.rb
```diff
@@ -0,0 +1,96 @@
...
+      # @param ids [Array] An array of XML objects representing ids to parse
+      # @return [Hash] An hash representing the parsed ids
+      def parse_xml_gcids(ids)
+        return unless ids.is_a? Array
+
+        ids = ids.map(&:attributes)
+
+        birls_ids = select_ids(select_extension(ids, BIRLS_IDS_REGEX, VA_ROOT_OID)) || []
+
+        {
```
jimiray, line 21: Is it possible to build a data structure that we can iterate over and reduce the number of lines here?

```
parsed_ids = {}
REGEX_MAPPINGS.each do | k, v|
  ids = select_ids(select_extension(ids, v[:pattern], v[:root])
  if v[:first] == true 
    parsed_ids[k] = ids&.first
  else
    parsed_ids[k] = ids
  end
end
```

lib/identity/parsers/gc_ids_helper.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Identity
+  module Parsers
+    module GCIdsHelper
+      # @param edipi[String] An string representing the edipi id
+      # @return [String] A string of 10 numerical digits representing the sanitized id
+      def sanitize_edipi(edipi)
+        return unless edipi.present? && edipi.is_a?(String)
```
jimiray, line 9: If we want to ensure that edipi is present and a String should we raise an error rather than just return nil? That way it never actually will return without the correct argument?

lib/identity/parsers/gc_ids_constants.rb
bosawt: 👍  Sounds good

lib/identity/parsers/gc_ids_helper.rb
bosawt: I think that's not a bad point, I think for now I'm going to leave it, and the reason is that the only place that invokes this function is `lib/mpi/responses/profile_parser.rb`, and it very explicitly wants `nil` rather than an error when it calls `sanitize_edipi` and `sanitize_id`.
Overall we might want to consider just rolling these sanitize functions into into the parsed GCIds hash itself, and I think that could be a good way to address this feedback.

bosawt: Although a question I do have:
- `sanitize_edipi` is used in the profile_parser only
- `sanitize_id` is used in add_parser, profile_parser, and ssoe
Does it seem like these aren't actually parser functions, and would be better served as a formatter or something? They're pretty small functions so might not make sense to make a mountain out of a molehill here, but both of these functions are in a strange spot currently...

lib/identity/parsers/gc_ids.rb
bosawt: Cool, check it out now, I took this suggestion. Had to move a couple things around to make this more clear but should still be good

lib/identity/parsers/gc_ids.rb
```diff
@@ -14,26 +14,23 @@ module GCIds
...
+              select_ids(extension)&.first
+            when :multiple_ids
+              select_ids(extension)
+            when :facility
+              select_facilities(extension)
+            when :icn_with_aaid
+              select_icn_with_aaid(extension)
+            end
+          { id_to_parse => parsed_ids }
+        end.reduce(:merge)
```
jimiray, line 33: I like this.

jimiray (approved):

lib/identity/parsers/gc_ids.rb
```diff
@@ -0,0 +1,96 @@
...
+
+        mapped_ids = ids.split(IDS_SPLIT_TOKEN).map do |id|
+          OpenStruct.new(attributes: { extension: id, root: root_oid })
+        end
+        parse_xml_gcids(mapped_ids)
+      end
+
+      private
+
+      def select_ids(extensions)
```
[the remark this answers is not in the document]
bosawt: 👍

LindseySaari:
I think this may be able to be broken up into a few separate PRs. At a quick glance, possibly the two new lib/identity/parsers/… and specs could be broken into separate PRs and the deletions could be its own as well.

bosawt:
@LindseySaari I rebased the changes on master from the other split PRs into this branch, so this should encompass the remaining changes for this work. Should be good 2 go, assuming I didn't break a spec in the meantime or otherwise introduce some strange merge-related issue

bosawt:
@LindseySaari Ouch, dang, I swear I counted the lines. Looks like I'll be asking for a slight `danger` exception here too. Hopefully sometime tomorrow works.

LindseySaari (approved):

PR #6072, opened 2021-03-04
refactor - adding factory and helpers

dillo:
## Description of change

- Refactor step two: Adding the `factory` class and the `factory_types` module.
## Original issue(s)
department-of-veterans-affairs/va.gov-team#20246

- [x] RSpec tests written and are passing


modules/health_quest/app/services/health_quest/resource/factory.rb
```diff
@@ -0,0 +1,42 @@
...
+        @user = opts[:user]
+        @resource_identifier = opts[:resource_identifier]
+        @session_service = HealthQuest::Lighthouse::Session.build(user: user, api: opts[:api])
+        @query = Query.build(session_store: session_service.retrieve,
+                             api: opts[:api],
+                             resource_identifier: resource_identifier)
+        @options_builder = Shared::OptionsBuilder
+      end
+
+      def get(id) # rubocop:disable Rails/Delegate
```
LindseySaari, line 22: Do you need the rubocop comment here?

dillo: Yes, I'm afraid so. My local rubocop fails if I don't have it in place. I wanted that method to be explicit since it's pretty crucial.

LindseySaari (approved):
Nice work on breaking this all up! Thanks 💯

PR #6079, opened 2021-03-04
API-5521 - Move CentralMailUpdater into services

nathanbwright:
## Description of change
It makes more sense to put the `CentralMailUpdater` object in services instead of lib based on its functionality.

## Original issue(s)
https://vajira.max.gov/browse/API-5521


yakschuss (approved):
LGTM! :shipit:

cilestin (approved):

PR #6084, opened 2021-03-04
Updated scope method and extracted helper methods

bradenshipley:

## Description of change

The scopes endpoint used to query for all authorization servers, then
select the server whose name matched the `category` param passed to the
method. This worked okay as an MVP to get the scope endpoint up and
running, however changes made to the Okta Authorization Servers have
made this refactor necessary. We now handle the `scopes` invocation when
the category is `health` by retrieving the information on a specific
server_id, and handle all other calls by checking for an auth_server that includes the category name. For readability,
helper methods were extracted from the `scopes` method.

Dummy data has been added for `directory.health_server_id`.
This `health_server_id` has been added in this devops PR that has already been merged. https://github.com/department-of-veterans-affairs/devops/pull/8619/files


lib/okta/directory_service.rb
```diff
@@ -6,20 +6,45 @@ module Okta
...
+
+    def handle_health_server
+      server = @okta_service.get_auth_server(Settings.directory.health_server_id)
+      scopes = @okta_service.get_server_scopes(server.body['id'])
+      remove_scope_keys(scopes)
+    end
+
+    def handle_nonhealth_server(category)
+      servers = @okta_service.get_auth_servers
       server = servers.body.select { |auth_server| auth_server['name'].include?(category.downcase) }
```
SlakrHakr, line 29: maybe handle the use-case if category is nil?
not sure if thats ever possible but even just raising a more specific exception would be better than the null reference exception that would happen on this line

spec/lib/okta/directory_service_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+
+require 'rails_helper'
+require 'okta/directory_service.rb'
+require 'okta/service'
+require 'vcr'
+
+RSpec.describe Okta::DirectoryService do
+  let(:subject) { described_class.new }
+  let(:server_scopes) do
+    {
```
SlakrHakr, line 11: nit: maybe store this in a separate json file? or in some sort of factory?

SlakrHakr (approved):

PR #6086, opened 2021-03-04
removes standard and available columns

bramleyjl:
## Description of change

Cleanup of TUD table, removes unused `standard` and `available` columns

## Original issue(s)

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_02_26_193901) do
+ActiveRecord::Schema.define(version: 2021_03_05_155617) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "btree_gin"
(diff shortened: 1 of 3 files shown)


db/schema.rb
```diff
@@ -612,8 +612,6 @@
     t.string "phone"
     t.string "email"
     t.string "password"
-    t.boolean "standard"
```
LindseySaari, line 615: Wondering if we should introduce a new migration for the removal of these columns from the schema? Or was the migration below not run as of yet? https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/engineering/backend/database-migrations.md#schema-updates

bramleyjl: No, that makes sense. https://github.com/department-of-veterans-affairs/vets-api/pull/6096 makes both columns ignored; after that I'll update this PR to introduce a migration to remove them from the schema.

LindseySaari: 💯

LindseySaari (approved):

PR #6089, opened 2021-03-05
19464 mobile pre cache claims and appeals

kreek:
## Description of change
Pre-caches a mobile user's claims and appeals data after their profile is retrieved. As the background worked needs the same logic that was in the claims controller it has been moved to a service proxy class which fronts the core lib claims and appeals services.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#19464


modules/mobile/app/controllers/mobile/v0/appointments_controller.rb
```diff
@@ -35,13 +35,13 @@ def cancel
...
       def fetch_cached_or_service(validated_params)
-        json = Mobile::V0::Appointment.get_cached_appointments(@current_user) if validated_params[:use_cache]
+        json = nil
+        json = Mobile::V0::Appointment.get_cached(@current_user) if validated_params[:use_cache]
 
         # if JSON has been retrieved from redis, delete the cached version and return recovered appointments
         # otherwise fetch appointments from the upstream service
         if json
           Rails.logger.info('mobile appointments cache fetch', user_uuid: @current_user.uuid)
-          Mobile::V0::Appointment.delete_cached_appointments(@current_user)
```
kreek, line 44: This is intended. Originally this was added so the cache was only available the first time the appointments screen was hit but 1) it's confusing it works that way and 2) API users have control of if they want to hit the cache or not via the `useCache` query param.

modules/mobile/app/models/mobile/v0/appointment.rb
```diff
@@ -54,22 +59,6 @@ def self.toggle_non_prod_id!(id)
         return id.sub(match[0], (%w[442 983] - [id]).first) if %w[442 983].include? match[0]
         return id.sub(match[0], (%w[552 984] - [id]).first) if %w[552 984].include? match[0]
       end
-
```
kreek, line 57: Caching methods have been moved to a concern so they can be mixed in as needed.

LindseySaari (approved):
I think backend-review-group is only needed for the redis.yml, LGTM

jperk51 (approved):
LGTM

PR #6093, opened 2021-03-05
Add office check

digitaldrk:
## Description of change
During a meeting with our integration partner (VRE/CMS), it was revealed that we (vets-api) need to only send them applications if the veteran's closest office's number is in a list. This PR adds this number check. No matter what we want to send the PDF that's generated to VBMS

## Original issue(s)
department-of-veterans-affairs/va.gov-team#20847

  * No new settings
  * No FF but it's not live in prod either
  * I'd like to keep track of failed location look ups so I added a sentry call to a rescue
  * no metrics
  * No updates to swagger
  * No concerns


spec/models/saved_claim/veteran_readiness_employment_claim_spec.rb
```diff
@@ -17,6 +17,7 @@
         city: 'Arroyo Grande',
         province: 'CA',
         internationalPostalCode: '93420'
+
```
LindseySaari, line 20: Are the newline commits necessary?

app/models/saved_claim/veteran_readiness_employment_claim.rb
```diff
@@ -54,6 +61,29 @@ def regional_office
...
+    )
+
+    regional_office_response[:regional_office][:number]
+  rescue => e
+    log_message_to_sentry(e.message, :warn, {}, { team: 'vfs-ebenefits' })
+    '000'
+  end
+
+  def bgs_client
+    external_key = parsed_form.dig('veteranInformation', 'fullName', 'first') || parsed_form['email']
```
LindseySaari, line 79: It seems like this could be moved to its own method

spec/models/saved_claim/veteran_readiness_employment_claim_spec.rb
digitaldrk: Nope! Nice catch

LindseySaari (approved):

PR #6098, opened 2021-03-05
updated rated disablities

annaswims:
## Description of change
After a disuccion with @Mottie we settled on adding a `updatedRatedDisabilities` element to the form rather than replacing `ratedDisabilities` and adding something to the metadata.  This controller is not yet used in prod. 

## Original issue(s)

### DIFF
--- a/app/controllers/v0/disability_compensation_in_progress_forms_controller.rb
+++ b/app/controllers/v0/disability_compensation_in_progress_forms_controller.rb
@@ -28,10 +28,8 @@ module V0
           metadata['returnUrl'] = '/disabilities/rated-disabilities'
         end
         evss_rated_disabilites = JSON.parse(rated_disabilities_evss.rated_disabilities.to_json)
-        prased_form_data['ratedDisabilities'] = camelize_with_olivebranch(evss_rated_disabilites)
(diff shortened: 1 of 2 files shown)


app/controllers/v0/disability_compensation_in_progress_forms_controller.rb
```diff
@@ -28,10 +28,8 @@ def data_and_metadata_with_updated_rated_disabilites
           metadata['returnUrl'] = '/disabilities/rated-disabilities'
         end
         evss_rated_disabilites = JSON.parse(rated_disabilities_evss.rated_disabilities.to_json)
-        prased_form_data['ratedDisabilities'] = camelize_with_olivebranch(evss_rated_disabilites)
-        metadata['ratedDisabilitiesUpdated'] = true
+        prased_form_data['updatedRatedDisabilities'] = camelize_with_olivebranch(evss_rated_disabilites)
```
LindseySaari, line 31: I see that prased_form_data is a typo but it appears to be the same copy/paste error elsewhere

annaswims: 🤦 Thanks!

LindseySaari (approved):

PR #6099, opened 2021-03-05
veteran must have an address in order to make an evss letters call

annaswims:
## Description of change
A veteran must have an address available to EVSS before we can make a call to the letters api or [we see an error](http://sentry.vfs.va.gov/organizations/vsp/issues/6135/?query=source%3Aletters&statsPeriod=14d).  

In a roundabout way, if we see there is an address in Va Profile then we know that EVSS can get the address. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#7816

### DIFF
--- a/app/controllers/v0/letters_controller.rb
+++ b/app/controllers/v0/letters_controller.rb
@@ -6,7 +6,7 @@ require 'evss/letters/service'
 
 module V0
   class LettersController < ApplicationController
-    before_action { authorize :evss, :access? }
+    before_action { authorize :evss, :access_letters? }
 
     def index
       response = service.get_letters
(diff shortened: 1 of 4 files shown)


annaswims:
@Mottie I'd like your approval on this before I merge.  
vets-api has been sending a 403 response all along if the user is not authorized to call EVSS (doesn't have edipi, ssn or participant id). 

I'm adding the condition that they must have a mailing address with address_line1 present to make a letters api call to EVSS.  Will you need/want to make any changes before this is merged?

LindseySaari (approved):

Mottie:
This sounds like a good idea to me. I think we should show an error message to users if the FE gets that response though.

ttm1995:
Good afternoon, I was having issues with claim and I was wondering if this was maybe the solution?
2973498 is my application number. If you could help that would be great I keep resubmitting and have not had a solution for a month now.

PR #6102, opened 2021-03-08
Adding Pension Eligibility Questions to VA Form 686/674

kathleencrawford:
## Description of change

We have been asked to add pension eligibility questions to VA Forms 686c and 674 (Dependents Application).  This PR takes the value of the responses to the pension questions from the front end payload and passes those values to BGS during the form submission.  The values from the front end come in as true/false, but must be converted to Y/N values when passed to BGS.  Nil values are also allowed as the questions are optional.  

## Original issue(s)
department-of-veterans-affairs/va.gov-team#20618

This PR is an update to existing functionality (submitting VA Form 686c or 674) so no new logging or settings have been added.  We are just adding a few more boolean values to the form submission.

## Testing
- [x] Local testing
- [x] Specs updated
- [x] Staging testing planned


app/models/bgs_dependents/adult_child_attending_school.rb
```diff
@@ -64,7 +66,8 @@ def described_class_attribute_hash
       {
         ssn: @ssn,
         birth_date: @birth_date,
-        ever_married_ind: @was_married == true ? 'Y' : 'N'
+        ever_married_ind: @was_married == true ? 'Y' : 'N',
```
LindseySaari, line 69: Can you use the formatted_boolean method here for the ever_married_ind value logic?

spec/factories/686c/form_686c_674.rb
```diff
@@ -132,7 +132,8 @@
               'suffix' => 'II'
             },
             'ssn' => '370947141',
-            'birth_date' => '2001-03-03'
+            'birth_date' => '2001-03-03',
+            'dependent_income' => true
```
LindseySaari, line 136: Does this need to be a Y/N value?

kathleencrawford: This value comes in as true/false from the front end, and then we have to convert it to Y/N before sending it to our back end partner.  So for the spec payload data I have it coming in as true to mock how it would be coming in from the FE.  In the spec test, it will convert the value to Y.

LindseySaari: Ah, I see!

app/models/bgs_dependents/adult_child_attending_school.rb
kathleencrawford: Great observation Lindsey!  The ever_married_ind was there prior to the formatted_boolean method being created.  I've updated this and ran a few spec tests just to make sure it is working properly. Thank you for the suggestion.

LindseySaari (approved):

PR #6108, opened 2021-03-09
Adjust AppealsAPI Job Schedules

cilestin:
## Description of change
Reports were coming in the previous day that were supposed to be sent the following - e.g. On Sunday we would receive Monday's email, since midnight+1second UTC is technically Monday even though after timezone adjustments the email would get sent out Sunday night for anywhere with a negative UTC offset.

This adjust the time the job is scheduled to be at the end of the desired day rather than beginning by technicality.


bradenshipley (approved):
LGTM

PR #6109, opened 2021-03-09
Add Diary update service

digitaldrk:
## Description of change
This PR:
* Updates the gemfile to the latest version of the [bgs-ext gem](https://github.com/department-of-veterans-affairs/bgs-ext)
* Adds a service that calls the new code added from above

Some interesting information is that we have to retrieve all diaries associated with the user, update the `status_date`, and then send that data back. If we don't send everything back, we will overwrite the data with what we sent. Is it an anti-pattern to call a public service method in the same service? Wondering what you think about that.
 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#19777

  * no additions
  * No FF but this will be FF'd. Also, this code is not publicly accessible
  * None added
  * none added
  * none added
  * no concerns at this time


LindseySaari (approved):
👍

PR #6113, opened 2021-03-09
[API-5538] - Update flash updater worker to use simple arguments

jw81:
## Description of change

Switches the ClaimsAPI FlashUpdater worker to require simple arguments.  This should prevent the problems that occur when marshalling/unmarshalling complex JSON objects into and out of REDIS.

## Original issue(s)
[API-5538](https://vajira.max.gov/browse/API-5538)


lib/bgs/auth_headers.rb
```diff
@@ -14,8 +14,8 @@ def to_h
 
     def auth_json
       {
-        external_uid: @user.uuid,
-        external_key: @user.email
+        external_uid: @user.ssn,
+        external_key: @user.ssn
```
jw81, line 18: These `external_uid` and `external_key` values eventually get used [here](https://github.com/department-of-veterans-affairs/vets-api/pull/6113/files#diff-53edf2d59af06de8bd09bb8ac579f023e6c344fea20d08959cdd9fa55e3ce1f1L44) and I guess they really only exist in order to dump some unique identifiers into BGS' logging.

If we can simply just supply the `ssn` as those logging values, then I wonder if we even need this logic and this class anymore?  🤷   Perhaps we wouldn't even need [this line](https://github.com/department-of-veterans-affairs/vets-api/pull/6113/files#diff-975d964cd5f478b1004b6c518030143618a6cc108760b936d41ca37d1d54b3a5R39) anymore as well?  🤷 🤷‍♂️ 🤷‍♀️ 

I could easily be misunderstanding something though.

SlakrHakr: we need those values in ClaimEstablisher
https://github.com/department-of-veterans-affairs/vets-api/blob/96bc7369a5187c618e0497d2a8f9835ec59f9321/modules/claims_api/app/workers/claims_api/claim_establisher.rb#L79

which, we should have the Veteran's ssn if this is guaranteed to be a representative submission, but because it could be the Veteran themself, those values need to get sent in there still i believe

SlakrHakr: dude... i think youre right... i dont think this class is needed at all anymore :(

jw81: Updated [here](https://github.com/department-of-veterans-affairs/vets-api/pull/6113/commits/a77c2cf001186ea1420f16eadaabc36c2dcddaea).

SlakrHakr (approved):

modules/claims_api/app/workers/claims_api/claim_establisher.rb
```diff
@@ -68,18 +67,13 @@ def service(auth_headers)
...
-      user.uuid = bgs_auth_headers['external_uid']
-      user.email = bgs_auth_headers['external_key']
-
-      user
+      {
+        'ssn' => auth_headers['va_eauth_pnid'],
+        'uuid' => nil,
+        'email' => nil,
+        'icn' => nil,
+        'common_name' => nil
```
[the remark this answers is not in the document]
jw81: Updated [here](https://github.com/department-of-veterans-affairs/vets-api/pull/6113/commits/fff0c6b3e8d0e1994bd587555e9938d7ab4da8fe).

SlakrHakr (approved):

LindseySaari (approved):

PR #6116, opened 2021-03-09
16026 vetext api integration v2

jperk51:
## Description of change
VEText push notification service

## Original issue(s)
department-of-veterans-affairs/va.gov-team#16026

Service added to vets-api to interface with VEText push notification service


lib/vetext/response_error.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module VEText
+  # An error representing a 200 response but with error details.
```
kreek, line 4: My understanding was VEText was a fairly new service. I'm surprised they're doing the 'error in a 200' thing. @jjulian do you think we could get them to change this?

lib/vetext/service.rb
```diff
@@ -0,0 +1,140 @@
...
+        raise e
+      end
+    end
+
+    def get_app_sid(app_name)
+      settings = Settings.vetext_push
+      if settings.key?("#{app_name}_sid".to_sym)
+        settings["#{app_name}_sid".to_sym]
+      else
+        raise Common::Exceptions::BackendServiceException, 'VETEXT_PUSH_404'
```
jjulian, line 136: I wouldn't consider this a VETEXT error, since we're generating it ourselves. I think we can raise a standard error, and return a 400 from the controller.

config/settings.yml
```diff
@@ -835,6 +835,13 @@ genisis:
   user: TestUser
   pass: bogus
 
+# Settings for the VEText integration (mobile push notifications)
+vetext_push:
+  va_mobile_app_sid: EC15A7E1767C4018AC50B442AAC50D1E # This is a placeholder. Should be replaced with real SID
+  base_url: https://vetext1.r01.med.va.gov
+  user: vets-api-username
+  pass: secret
```
jjulian, line 843: These all all placeholders, is there value in setting these to "real-looking" data?

lib/vetext/service.rb
```diff
@@ -0,0 +1,140 @@
...
+    private
+
+    # Raise an error if the service returned an error in the
+    # body of a 200 response
+    #
+    def raise_if_response_error(body)
+      raise VEText::ResponseError, body if body[:success] == false
+    end
+
+    def remap_error(e)
```
kreek, line 118: Could the remapping happen in middleware? See the [SOAP middleware](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/common/client/middleware/response/soap_parser.rb) for an example.

lib/vetext/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Set a user's pref
+PUT /api/vetext/pub/mobile/push/preference/client
+
+Set a Preference to true or false, whether the user wants to receive them. Call with one Preference at a time.
+
+### Send a push notification
+POST /api/vetext/pub/mobile/push/send
+
+Send a push notification. Pass Endpoint SID or ICN as well a Preference SID.
```
jjulian, line 61: Consider merging this info into the api docs

lib/vetext/service.rb
```diff
@@ -0,0 +1,140 @@
...
+    #
+    # @return Hash            response object
+    #
+    def get_preferences(endpoint_sid)
+      response = perform(
+        :get,
+        "#{PREFERENCES_PATH}/#{endpoint_sid}",
+        nil
+      )
+      response.body
```
jjulian, line 58: `raise_if_response_error` needed here?

lib/vetext/service.rb
```diff
@@ -0,0 +1,140 @@
...
+    configuration VEText::Configuration
+
+    BASE_PATH = '/api/vetext/pub/mobile/push'
+    REGISTER_PATH = BASE_PATH + '/endpoint'
+    PREFERENCES_PATH = BASE_PATH + '/preferences/client'
+    SEND_PATH = BASE_PATH + '/send'
+
+    # Register a user's mobile device with the push notification service.
+    #
+    # @app_name String       name from
```
jjulian, line 19:
```suggestion
    # @app_name String       unique key for the mobile app
```

lib/vetext/service.rb
```diff
@@ -0,0 +1,140 @@
...
+                                                              { detail: e.body[:error] }, e.status,
+                                                              e.body)
+      when 500..599
+        raise Common::Exceptions::BackendServiceException.new('VETEXT_PUSH_502', {}, e.status)
+      else
+        raise e
+      end
+    end
+
+    def get_app_sid(app_name)
```
jjulian, line 131: ruby style often omits `get_`

kreek (commented):
A couple comments but looks good so far. Main thing would be moving the error mapping to middleware.

spec/lib/vetext/service_spec.rb
```diff
@@ -0,0 +1,170 @@
...
+              }
+            )
+          end.to raise_error(Common::Exceptions::BackendServiceException, /VETEXT_PUSH_400/)
+        end
+      end
+    end
+
+    context 'bad app_name provided' do
+      it 'raises a an error' do
+        VCR.use_cassette('vetext/register_success') do
```
jjulian, line 51: This case may not need a cassette

lib/vetext/service.rb
```diff
@@ -0,0 +1,140 @@
...
+
+    # Send a push notification to a single device
+    #
+    # @endpoint_sid String    the registration id as returned from `register`
+    # @template_id String     id of the push notification content template
+    # @personalisation Hash   data map provided by sender to fill in specified template
+    #
+    # @return Hash            response object
+    #
+    def send_notification(endpoint_id, template_id, personalization = {})
```
jjulian, line 94: @lexicalninja does this call (which the api will expose) provide enough detail for testing push?

jjulian (changes requested):
Looking good, I have a few suggestions and questions

lib/vetext/service.rb
lexicalninja: It should.

 I can't think of anything other than sid, template, and data that needs to be in here. It may need to change after Vetext actually has the templates, but this looks like everything that is in the preliminaries

lib/vetext/service.rb
jperk51: The `success` key isn't part of the response for the `get_preferences` endpoint. I assumed that means there's no way to have a 200 response that's an error.

lib/vetext/README.md
jperk51: All of this info is more or less part of the mobile API docs, or it will be when I add the send endpoint. I don't see too much reason for this file anymore. I'm happy to get rid of it if that's what we'd prefer.

jjulian: I like the top half of the file, which will allow us to "translate" some of the abstract concepts from the VEText API

lib/vetext/service.rb
jjulian: True, good find.

lib/vetext/response_error.rb
jperk51: Is this something we can bring up in the meeting tomorrow morning?

lib/vetext/README.md
jperk51: sounds good, I'll leave the dictionary portion and remove the endpoints since I think that should be covered by API documentation anyway.

lib/vetext/response_error.rb
jperk51: Based on the below chat I think we can just send the response straight through to the controllers. 
https://dsva.slack.com/archives/C01CSM3EZGT/p1615409557008500

lib/vetext/service.rb
jperk51: Based on the same chat I don't think we actually need this. I'll be rewriting the implementation/tests to reflect what the VEText API is actually doing. https://dsva.slack.com/archives/C01CSM3EZGT/p1615409557008500

lib/vetext/response_error.rb
jperk51: The very bottom of the document in that thread is where all the errors are listed out

lib/vetext/service.rb
jperk51: Is it necessary to remap a `Common::Client::Errors::ClientError` to a custom `Common::Exceptions::BackendServiceException`? Just not sure what the precedent is for this.

jjulian: When I started this PR, I did some poking around other service integrations in this repo, and this seemed like best practice. Happy to be wrong about that though.

jperk51: Sounds good to me. I'll get with Alastair to figure out this middleware stuff.

lib/vetext/service.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    BASE_PATH = '/api/vetext/pub/mobile/push'
+    REGISTER_PATH = BASE_PATH + '/endpoint'
+    PREFERENCES_PATH = BASE_PATH + '/preferences/client'
+    SEND_PATH = BASE_PATH + '/send'
+
+    def register(app_name, device_token, icn, os_info, device_name = nil)
+      app_sid = app_sid(app_name)
+      perform(:put, REGISTER_PATH, {
+                appSid: app_sid,
```
jjulian, line 20:
```suggestion
      perform(:put, REGISTER_PATH, {
                appSid: app_sid(app_name),
```

config/locales/exceptions.en.yml
```diff
@@ -1051,12 +1039,6 @@ en:
         code: VET360_ADDR300
         detail: Cannot modify an existing inactive record.
         status: 400
-      VET360_ADDR300:
-        <<: *external_defaults
-        title: Address Inactive
-        code: VET360_ADDR300
-        detail: Cannot modify an existing inactive record.
-        status: 400
```
jjulian, line 1059: accidental line deletes?

lib/vetext/response_error.rb
```diff
@@ -0,0 +1,13 @@
...
+  #
+  class ResponseError < StandardError
```
jjulian, line 6: no longer used

jjulian (commented):
Looks really good, great work!

config/locales/exceptions.en.yml
jperk51: Nope, I believe these were all duplicates. Had to tell in the github minified change view.

lib/vetext/response_error.rb
jperk51: Good call

lib/vetext/service.rb
lexicalninja: Now that @patrickvinograd says something, this should probably be icn

lib/common/client/middleware/response/vetext_errors.rb
```diff
@@ -0,0 +1,27 @@
...
+            return if env.success?
+
+            case env.status
+            when 400..499
+              raise Common::Exceptions::BackendServiceException.new('VETEXT_PUSH_400', {}, env.status,
+                                                                    env.body)
+            when 500..599
+              raise Common::Exceptions::BackendServiceException.new('VETEXT_PUSH_502', {}, env.status)
+            else
+              raise env
```
kreek, line 18: Is this case covered by specs? I'm not sure what happens if you raise on even but I think it would be a StandardError. That is if it's even possible to get here 😄 Maybe have it be a generic `Common::Exceptions::BackendServiceException`

kreek (approved):
One question, but if @jjulian's change requests are all resolved this LGTM 🚀

lib/common/client/middleware/response/vetext_errors.rb
jperk51: It's covered in service_spec.rb. If you think it's more appropriate to have it just be a generic `Common::Exceptions::BackendServiceException` that's easy enough to change. I'm not exactly sure what the advantages/disadvantages of having it be either or would be.

kreek: Looks like if you raise on `env` which is a `Faraday::Env` struct you'll get the following:
```
pry(#<Common::Client::Middleware::Response::VETextError>)> raise env
TypeError: exception class/object expected
```

jperk51: After refreshing myself on what the http status codes are supposed to mean, I do think 502 the most accurate choice for status code.

jjulian (approved):

jperk51:
@department-of-veterans-affairs/backend-review-group This PR is ready for review from VSP

thilton-oddball (approved):
This all looks good!

PR #6119, opened 2021-03-09
Api 5537/va forms add example for required parameter

mviterna:
## Description of change

We added an example for a required field on the form detail swagger doc and set the example form to 10-10EZ

## Original issue(s)
https://vajira.max.gov/browse/API-5537

Testing was performed locally on Ubuntu running native and against developer-portal. Confirmed docs were updated and curl command is generated and pulls back 10-10EZ


alexandervalencia (approved):
Confirmed tests passed and example working as intended. Thanks @mviterna 🚀

mviterna:
@ericboehs Can I get a PR on this change for @alexandervalencia ? Alex, are you sure we want "form_name" in the example description? I can put it to "VA Form Name" or something similar. Just checking. I'll put whatever your team wants there.

alexandervalencia:
As far as I know, adding the example was the only thing we needed changed. Thanks for asking!

cshupp1 (approved):

PR #6120, opened 2021-03-09
Add lighthouse appointments

dillo:
## Description of change

- Adding the ability for the Questionnaire Response Factory to fetch appointments from Lighthouse. This is the first in a series of PRs which will get created in order to complete the requirements of the story.
- Note that we're still fetching appointments from the MAP/MAS service. This ability will be removed once the rest of the code has been added to the Questionnaire Manager and other classes. The reason for keeping the MAP appointments around, for the time being, is so that the tests don't break if they get removed and we don't end up going on a large refactoring spree!
## Original issue(s)
department-of-veterans-affairs/va.gov-team#20248

- [x] RSpec test suites are passing.


LindseySaari (approved):
👍 but it looks like an unrelated spec is failing

PR #6122, opened 2021-03-10
VAOS - Better Error Instrumentation and Logging

saneshark:
## Description of change
We were previously only doing instrumentation based on a response, but in the case of GatewayTimeout and other socket related errors that could arise, we would not have appropriate instrumentation and logging. This improves upon that.

## Original issue(s)
No ticket, just noticed a gap in our Grafana dashboard for a recent outtage.


modules/vaos/spec/services/middleware/vaos_logging_integration_spec.rb
```diff
@@ -45,12 +49,116 @@
...
     end
 
+    context 'with a timeout' do
+      around do |example|
+        VCR.turn_off!
+        WebMock.allow_net_connect!
+        example.run
+        VCR.turn_on!
+        WebMock.disable_net_connect!
+      end
```
grossb1, line 59: I'm putting my Kam hat on =).
Nitpick, this can probably be broken out into a method.

grossb1 (approved):

saneshark:
@kreek tagging you just so you have a heads up on this change and because you touched this code a bit in the past.

Gemfile.lock
```diff
@@ -1058,6 +1058,7 @@ DEPENDENCIES
   sidekiq-pro!
   sidekiq-scheduler (~> 3.0)
   simplecov (< 0.18)
+  slack-notify
```
LindseySaari, line 1061: The latest master includes this change

modules/vaos/app/services/vaos/middleware/vaos_logging.rb
```diff
@@ -23,27 +25,49 @@ def initialize(app)
...
+      end
+
+      def statsd_increment(key, env, error = nil)
+        StatsDMetric.new(key: key).save
+        tags = [
+          "method:#{env.method.upcase}",
+          "url:#{StringHelpers.filtered_endpoint_tag(env.url.path)}",
+          "http_status:#{error.present? ? error.class : env.status}"
+        ]
+        StatsD.increment(key, tags: tags)
```
LindseySaari, line 68: Does this metric needs to be initialized to zero (see the statsd.rb initializer)

saneshark: It already is. We have it in our `modules/vaos/config/statsd.rb` :)

PR #6125, opened 2021-03-10
Increase Breakers Error Threshold (DecisionReview Service)

edmangimelli:
Resolves https://github.com/department-of-veterans-affairs/va.gov-team/issues/20695

### DIFF
--- a/lib/decision_review/configuration.rb
+++ b/lib/decision_review/configuration.rb
@@ -57,7 +57,7 @@ module DecisionReview
     end
 
     def breakers_error_threshold
-      80
+      80 # breakers will be tripped if error rate reaches 80% over a two minute period.
     end
   end
 end
--- a/lib/decision_review/configuration.rb
+++ b/lib/decision_review/configuration.rb
@@ -55,5 +55,9 @@ module DecisionReview
     def mock_enabled?
       Settings.decision_review.mock || false
     end
+
+    def breakers_error_threshold
+      80
+    end
   end
 end


lib/decision_review/configuration.rb
```diff
@@ -55,5 +55,9 @@ def connection
     def mock_enabled?
       Settings.decision_review.mock || false
     end
+
+    def breakers_error_threshold
+      80
```
LindseySaari, line 60: I think 80 is a percentage, correct? Would you be able to place a comment here? It's not super clear that the integer is actually the percentage of errors over which an outage will be reported

LindseySaari: Do you actually want to increase the read timeout instead in the above config file? Current configuration: `self.read_timeout = Settings.caseflow.timeout || 20`

edmangimelli: it is a percentage! yes. i'll note that.

edmangimelli: After talking this over with @annaswims, I think we wanted to try this route first. Our upstream dependency caches some data, so a second request is often pretty quick, and I think we were nervous about allowing a lot of lengthy connections all at once (Anna correct me if I'm wrong)

annaswims: yeah, the current timeout is set to 40s and we probably don't want to go much higher unless we make it async. Because we're calling lighthouse, there's not much of a chance that lighthouse will be down when we're not.  So breakers is for stats more than for being a circuit breaker.

edmangimelli:
tested in console to see if the DecisionReview breakers updated:
```
pp DecisionReview::Configuration.instance.breakers_service.instance_variable_get :@configuration
```
(it did)

LindseySaari (approved):

edmangimelli:
thanks as always! @LindseySaari

PR #6126, opened 2021-03-10
18526 copy ppms models to engine

darkbushido:
## Description of change

This creates a new rails engine FacilitiesApi
I've modified the generated files so they are able to work in the scope of the engine
Copying over the Provider and Specialties models as well as inverting the version naming convention (v1/ppms instead of ppms/v1)

## Original issue(s)
department-of-veterans-affairs/va.gov-team#11608
department-of-veterans-affairs/va.gov-team#18526

This is a small building block to puting facilities in an engine.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -361,6 +361,7 @@ Rails.application.routes.draw do
   # Modules
   mount CovidResearch::Engine, at: '/covid-research'
   mount CovidVaccine::Engine, at: '/covid_vaccine'
+  # mount FacilitiesApi::Engine, at: '/facilities_api'
   mount HealthQuest::Engine, at: '/health_quest'
   mount Mobile::Engine, at: '/mobile'
   mount VAOS::Engine, at: '/vaos'
--- a/modules/facilities_api/README.rdoc
+++ b/modules/facilities_api/README.rdoc
@@ -1,0 +1,10 @@
+= FacilitiesApi
+TODO: Short description and motivation.
+
+== Installation
+Ensure the following line is in the root project's Gemfile:
+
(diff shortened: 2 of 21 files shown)


modules/facilities_api/spec/factories/facilities_api/v1/ppms/specialties.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :v1_ppms_specialty, class: 'V1::Ppms::Specialty' do
```
LindseySaari, line 4: Should Ppms be PPMS?

config/routes.rb
```diff
@@ -361,6 +361,7 @@
   # Modules
   mount CovidResearch::Engine, at: '/covid-research'
   mount CovidVaccine::Engine, at: '/covid_vaccine'
+  # mount FacilitiesApi::Engine, at: '/facilities_api'
```
LindseySaari, line 364: Are you commenting this out until it's ready to go?

modules/facilities_api/README.rdoc
```diff
@@ -0,0 +1,10 @@
+= FacilitiesApi
+TODO: Short description and motivation.
+
+== Installation
+Ensure the following line is in the root project's Gemfile:
+
+  gem 'facilitiesapi', path: 'modules/facilitiesapi'
```
LindseySaari, line 7: Was this auto generated? Seems like this should be facilities_api (underscore)

darkbushido: It's auto-generated. I had to fix this in several other places but the generator is dropping the `_`

LindseySaari (approved):

PR #6132, opened 2021-03-10
Api 5224/evidence submission temporary upload

yakschuss:
[API-5224](https://vajira.max.gov/browse/API-5224)

This PR adds support for Temporary Upload to S3 using carrier wave (after validation). 

This approach assumes that we will be able to handle a single file upload of a max 100MB size without timeouts, which is being tested currently.

NOTE: This PR is blocked/dependent on [#6117](https://github.com/department-of-veterans-affairs/vets-api/pull/6117)

### DIFF
--- a/modules/appeals_api/app/services/appeals_api/supporting_evidence/evidence_uploader.rb
+++ b/modules/appeals_api/app/services/appeals_api/supporting_evidence/evidence_uploader.rb
@@ -7,7 +7,7 @@ module AppealsApi
         @appeal = appeal
         @document = document
         @type = type
-        raise ArgumentError, 'invalid type' unless valid_type!
+        raise ArgumentError, 'invalid type' unless valid_type?
       end
 
       def process!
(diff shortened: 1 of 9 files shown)


modules/appeals_api/app/services/appeals_api/supporting_evidence/evidence_uploader.rb
```diff
@@ -0,0 +1,45 @@
...
+          status: 'submitted',
+          file_data: {
+            filename: uploader.filename
+          }
+        )
+      end
+
+      def valid_type!
+        type.in?(%i[notice_of_disagreement])
+      end
```
nathanbwright, line 42: Should this be a `valid_type?` instead?

yakschuss: Yes!

thilton-oddball (approved):
Looks good

nathanbwright (approved):
Nice work! 🚀

cilestin (approved):

PR #6135, opened 2021-03-10
Pv sm attachment fix

patrickvinograd:
## Description of change
Updates the mobile messaging controller to handle posting messages with/without attachments. 

A message with attachments gets sent as a `multipart/form-data`, and current implementation of ActionController::Parameters treats the JSON payload in a multipart payload as a String instead of parsing it into a Hash. So we coerce that into an object that we can do strong parameter validation on. 

Then also turn Parameters objects into Hashes before passing to the SM client. 

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/20750

Tested locally to make sure eventual payload to SM client was correct in all cases.

### DIFF
--- a/modules/mobile/app/controllers/mobile/v0/messages_controller.rb
+++ b/modules/mobile/app/controllers/mobile/v0/messages_controller.rb
@@ -36,17 +36,16 @@ module Mobile
       end
 
       def create
-        binding.pry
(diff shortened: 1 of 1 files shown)


narin (approved):
LGTM!

kreek (approved):
LGTM 🚀

PR #6137, opened 2021-03-10
Convert dd_type to symbol in direct deposit email jobs

lihanli:
## Description of change
when symbols are passed as arguments to sidekiq background jobs, they get converted to strings. they should be converted back into symbols in the direct deposit email jobs.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/21391

testing in staging planned

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1058,6 +1058,7 @@ DEPENDENCIES
   sidekiq-pro!
   sidekiq-scheduler (~> 3.0)
   simplecov (< 0.18)
+  slack-notify
   spring
   spring-commands-rspec
   staccato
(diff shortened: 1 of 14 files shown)


Gemfile.lock
```diff
@@ -1058,6 +1058,7 @@ DEPENDENCIES
   sidekiq-pro!
   sidekiq-scheduler (~> 3.0)
   simplecov (< 0.18)
+  slack-notify
```
LindseySaari, line 1061: I _think_ this isn't mean to be here

lihanli: ok thanks, i removed it, it gets readded every time i run bundler though so maybe it should be in master?

LindseySaari (approved):
👍

PR #6138, opened 2021-03-10
API-5651 Adjust retry attempts for email report jobs

cilestin:
## Description of change

Having a daily or weekly report retry for 21 days doesn't really make a whole lot of sense. This PR adjusts the retry count to be more sensible for each report type.

It also makes sure that the start time of the reports is less dependent on the runtime of the job, setting itself to the beginning of the day the job is ran to avoid missing something as the retries get longer and longer.

## Original issue(s)
https://vajira.max.gov/browse/API-5651


nathanbwright (approved):
Looks good to me!

PR #6151, opened 2021-03-11
20248 Fix questionnaire manager QR post issue

dillo:
## Description of change

- Fix an issue with POSTing Questionnaire Responses params to the Questionnaire Manager.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#20248

- [x] RSpec test suites are passing.

### DIFF
--- a/modules/health_quest/app/services/health_quest/resource/client_model/questionnaire_response.rb
+++ b/modules/health_quest/app/services/health_quest/resource/client_model/questionnaire_response.rb
@@ -65,8 +65,7 @@ module HealthQuest
         end
 
         def set_questionnaire
-          questionnaire_id = data.dig(:questionnaire, :id) || DEFAULT_QUESTIONNAIRE_ID
-
+          questionnaire_id = data.dig(:questionnaire, :id)
           "Questionnaire/#{questionnaire_id}"
         end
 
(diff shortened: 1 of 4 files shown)


modules/health_quest/app/services/health_quest/resource/client_model/questionnaire_response.rb
```diff
@@ -65,8 +65,7 @@ def set_source
         end
 
         def set_questionnaire
-          questionnaire_id = data.dig(:questionnaire, :id) || DEFAULT_QUESTIONNAIRE_ID
-
+          questionnaire_id = data.dig(:questionnaire, :id)
```
LindseySaari, line 68: Do you want an || statement here still?

dillo: @LindseySaari we don't want the default to be set for the title. If it's missing we actually want to see that it's missing as opposed to giving the user and the system the wrong idea of what the `questionnaire` actually is.

LindseySaari: Ok, great! 👍

LindseySaari (approved):

PR #6152, opened 2021-03-12
Add missing Secure Messaging exceptions

narin:
## Description of change
Adds additional exception mappings based on version 2.5 of the MHV API docs: https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/products/health-care/secure-messaging/mhv-api/Secure_Messaging_API_Veteran_2.5.doc

Also removes some duplicates I found for non-SM errors.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#20770

### DIFF
--- a/config/locales/exceptions.en.yml
+++ b/config/locales/exceptions.en.yml
@@ -275,16 +275,60 @@ en:
     # Below this line just define the minor code as key to be used by client error
       VA900:
         <<: *external_defaults
-      SM105: # What casues this? Perhaps then we can come up with more customized detail below
+      SM100:
+        <<: *validation_errors
+        code: 'SM100'
+        detail: 'The message body cannot be blank'
+      SM101:
+        <<: *external_defaults
+        code: 'SM101'
+        detail: 'Application authentication failed'
+        status: 422
+      SM102:
+        <<: *external_defaults
+        code: 'SM102'
+        detail: 'Application authorization failed'
+        status: 422
+      SM103:
+        <<: *external_defaults
+        code: 'SM103'
+        detail: 'Invalid User Credentials'
+        status: 422
+      SM104:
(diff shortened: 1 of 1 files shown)


config/locales/exceptions.en.yml
```diff
@@ -751,12 +863,6 @@ en:
         code: VET360_ADDR110
         detail: Size must be between 0 and 100.
         status: 400
-      VET360_ADDR110:
-        <<: *external_defaults
-        title: Address Line 2 Size
-        code: VET360_ADDR110
-        detail: Size must be between 0 and 100.
-        status: 400
```
narin, line 759: This exception and the following 2 I removed are duplicates.

config/locales/exceptions.en.yml
```diff
@@ -301,11 +345,55 @@ en:
...
+        <<: *external_defaults
+        code: 'SM118'
+        detail: 'It is not possible to move a message from the DRAFTS folder'
+        status: 422
       SM119: # 'Triage team does not exist' when trying to create new message
         <<: *external_defaults
         code: 'SM119'
         detail: 'Triage team does not exist'
         status: 422
+      SM120:
```
LindseySaari, line 358: I noticed that some of the exception types below don't have a status attribute. Is that intentional?

patrickvinograd: If no status is defined it will default to a 400:
https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/common/exceptions/backend_service_exception.rb#L75

patrickvinograd:
I think that some of these MHV errors are probably not possible/probable from our client, but no harm in defining them in case they crop up.

patrickvinograd (approved):
LGTM

PR #6155, opened 2021-03-12
Added catch for opaque tokens.

dbrown-va:
## Description of change
In preparation for SSOi and SSOe access tokens usage with the Lighthouse API's, this adds a small foundation for detecting those non-JWT tokens prior to the Token.rb instantiation which relies on JWT access token structure.

## Original issue(s)
https://vajira.max.gov/browse/API-5576

* This is a first step PR only to gracefully handle SSOi & SSOe opaque tokens rather than stack trace errors on JWT decode failures.
* Future additions will add support for introspect calls, scope validation, etc.

### DIFF
--- a/app/controllers/openid_application_controller.rb
+++ b/app/controllers/openid_application_controller.rb
@@ -50,9 +50,10 @@ class OpenidApplicationController < ApplicationController
   def token_from_request
     auth_request = request.authorization.to_s
     return unless auth_request[TOKEN_REGEX]
+
     token_string = auth_request.sub(TOKEN_REGEX, '').gsub(/^"|"$/, '')
 
-    if is_jwt?(token_string)
+    if jwt?(token_string)
       Token.new(token_string, fetch_aud)
     else
       # Future block for opaque tokens
@@ -60,11 +61,11 @@ class OpenidApplicationController < ApplicationController
     end
   end
 
-  def is_jwt?(token_string)
+  def jwt?(token_string)
     JWT.decode(token_string, nil, false, algorithm: 'RS256')
-    return true
-  rescue JWT::DecodeError => e
-    return false
+    true
+  rescue JWT::DecodeError
+    false
   end
 
   def error_klass(error_detail_string)
--- a/app/controllers/openid_application_controller.rb
+++ b/app/controllers/openid_application_controller.rb
@@ -50,8 +50,28 @@ class OpenidApplicationController < ApplicationController
   def token_from_request
     auth_request = request.authorization.to_s
     return unless auth_request[TOKEN_REGEX]
+    token_string = auth_request.sub(TOKEN_REGEX, '').gsub(/^"|"$/, '')
 
-    Token.new(auth_request.sub(TOKEN_REGEX, '').gsub(/^"|"$/, ''), fetch_aud)
+    if is_jwt?(token_string)
+      Token.new(token_string, fetch_aud)
+    else
+      # Future block for opaque tokens
+      raise error_klass('Opaque tokens not supported.')
+    end
+  end
+
+  def is_jwt?(token_string)
+    JWT.decode(token_string, nil, false, algorithm: 'RS256')
+    return true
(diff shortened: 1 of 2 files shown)


lukemarkwordtlibertyits:
# Tests

## JWT Functionality Remains Consistant

**Test Steps**

- Hit validate endpoint with JWT

**Expected Results**

- [ ] 200 Response
- [ ] Json
  ```json
  {
    "data": {
      "id": "id",
      "type": "validated_token",
      "attributes": {
        "ver": 1,
        "jti": "jti",
        "iss": "iss",
        "aud": "aud",
        "iat": 1234,
        "exp": 1234,
        "cid": "cid",
        "uid": "uid",
        "scp": [
          "..."
        ],
        "sub": "sub",
        "va_identifiers": {
          "icn": "icn"
        }
      }
    }
  }
  ```
## Checks for opaque token

**Test Steps**

- Hit Validate endpoint with opaque token (random string)

**Expected Results**

- [ ] 401 Status
- [ ] Json
  ```json
  {
    "errors": [
      {
        "title": "Not authorized",
        "detail": "Opaque tokens not supported.",
        "code": "401",
        "status": "401"
      }
    ]
  }
  ```

dbrown-va:
> # Tests
> ## JWT Functionality Remains Consistant
> ...
> 
> Confirmed functionality. I am not sure why the build is failing.

Linting issues that popped up in the PR build.  Those have now been corrected.

app/controllers/openid_application_controller.rb
```diff
@@ -51,7 +51,28 @@ def token_from_request
...
     return unless auth_request[TOKEN_REGEX]
 
-    Token.new(auth_request.sub(TOKEN_REGEX, '').gsub(/^"|"$/, ''), fetch_aud)
+    token_string = auth_request.sub(TOKEN_REGEX, '').gsub(/^"|"$/, '')
+
+    if jwt?(token_string)
+      Token.new(token_string, fetch_aud)
+    else
+      # Future block for opaque tokens
+      raise error_klass('Opaque tokens not supported.')
```
beaugrantham-va, line 60: I recommend we change this error message to one of two things.

- A copy of what gets returned today for invalid JWTs 
- A generic error (e.g., `Invalid token`)

My only concern calling out opaque tokens specifically is that some APIs return this to the client and it's unnecessarily detailed. They need to be aware of the types of tokens that are valid, just that what they have used is invalid.

spec/controllers/openid_application_controller_spec.rb
```diff
@@ -221,6 +221,18 @@ def index
       end
     end
 
+    context 'with an opaque token supplied and no session' do
+      it 'returns 200 and add the user to the session' do
```
beaugrantham-va, line 225: I don't think this test expectation is correct.

spec/controllers/openid_application_controller_spec.rb
```diff
@@ -221,6 +221,18 @@ def index
...
+        with_okta_configured do
```
beaugrantham-va, line 226: Is Okta configuration necessary here since we don't actually get that far?

dbrown-va: Not necessary since this is for the non-Okta case.

I have removed the `with_okta_configured` wrapper and confirmed the test still succeeds.

spec/controllers/openid_application_controller_spec.rb
dbrown-va: Yeah, missed updating that description and only did the context.

I've updated it to describe the 401 returned.

app/controllers/openid_application_controller.rb
dbrown-va: Understood & agreed.  

Invalid JWTs to date return segment encoding errors along with the 401.  Since that wouldn't apply, I went with the `Invalid token` suggestion.

beaugrantham-va (approved):

PR #6156, opened 2021-03-12
21434 add VEText id to mobile appointments response

kreek:
## Description of change
Adds `vetextId' to the appointments response which is the Appointment clinic id/time id that VEText passes along in appointment push notifications. Also updates our hashed cancelId. Moves the cancelId encoding and decoding methods to the Appointment model.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#21434


Gemfile.lock
```diff
@@ -1058,6 +1058,7 @@ DEPENDENCIES
   sidekiq-pro!
   sidekiq-scheduler (~> 3.0)
   simplecov (< 0.18)
+  slack-notify
```
LindseySaari, line 1061: If you merge in the latest master, this is taken care of

jperk51 (approved):
LGTM

PR #6157, opened 2021-03-12
19464 create mobile claims proxy (pre-req for caching in worker)

kreek:
## Description of change
Created to get the line count down for https://github.com/department-of-veterans-affairs/vets-api/pull/6089

Moves claims and appeals logic in the controller to a proxy that sits between it and the service so that a background worker can also use the proxy to pre-cache claims and appeals data for a user.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#19464


jperk51 (approved):
LGTM

PR #6158, opened 2021-03-12
Cris.shupp/api 5293 status and size

cshupp1:
## Description of change
Records size of benefits intake document multipart file and records length of time spent in a given state.  So if we are in the 'pending' state, and state changes to updated we record the length of time spent in pending.  This is done for all states.

## Original issue(s)
https://vajira.max.gov/browse/API-5293
https://vajira.max.gov/browse/API-5694

Several RSPEC tests have been written.


mviterna (approved):

gpbowman-git (approved):
Looks good

PR #6160, opened 2021-03-12
Remove most direct user.mpi calls [#5411]

bramleyjl:
## Description of change

Further refactoring of the User model to internally handle requests for identity information, this change takes out most `user.mpi` calls that occur outside of the model. This should not result in no noticeable change with the exception of instances when the cached SAML payload's information differs from MPI's, since the User model defaults to using the SAML payload's information over MPI.

## Original issue(s)

Existing unit tests continue to pass and cover User model methods replacing User.mpi calls

### DIFF
--- a/lib/evss/disability_compensation_auth_headers.rb
+++ b/lib/evss/disability_compensation_auth_headers.rb
@@ -31,7 +31,7 @@ module EVSS
     end
 
     def gender
-      case @user.gender || @user.mpi&.profile&.gender
+      case @user.gender
       when 'F'
         'FEMALE'
       when 'M'
--- a/modules/vaos/app/services/vaos/jwt_wrapper.rb
+++ b/modules/vaos/app/services/vaos/jwt_wrapper.rb
@@ -24,11 +24,11 @@ module VAOS
     def payload
       {
         authenticated: true,
-        sub: @user.icn,
+        sub: user.icn,
         idType: ID_TYPE,
         iss: ISS,
-        firstName: @user.first_name,
-        lastName: @user.last_name,
+        firstName: user.first_name,
+        lastName: user.last_name,
         authenticationAuthority: AUTHORITY,
         jti: SecureRandom.uuid,
         nbf: 1.minute.ago.to_i,
@@ -38,13 +38,13 @@ module VAOS
         gender: gender,
         dob: parsed_date,
         dateOfBirth: parsed_date,
-        edipid: @user.edipi,
-        ssn: @user.ssn
+        edipid: user.edipi,
+        ssn: user.ssn
       }
     end
 
     def gender
-      type = @user.gender
+      type = user.gender
       return '' unless type.is_a?(String)
 
       case type.upcase[0, 1]
--- a/modules/vaos/app/services/vaos/jwt_wrapper.rb
+++ b/modules/vaos/app/services/vaos/jwt_wrapper.rb
@@ -24,11 +24,11 @@ module VAOS
     def payload
       {
         authenticated: true,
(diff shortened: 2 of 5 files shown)


modules/vaos/app/services/vaos/jwt_wrapper.rb
```diff
@@ -24,11 +24,11 @@ def token
     def payload
       {
         authenticated: true,
-        sub: icn,
+        sub: @user.icn,
```
bosawt, line 27: Same here, since there's an `attr_reader` in this class, these can all refer to `user` without the symbol

bosawt (approved):
Looks good! Just some observations about minor things regarding code convention and such

thilton-oddball (approved):
LGTM

lib/evss/disability_compensation_auth_headers.rb
```diff
@@ -31,7 +31,7 @@ def eauth_json
     end
 
     def gender
-      case @user.gender || @user.mpi&.profile&.gender
```
annaswims, line 34: @bramleyjl this change will negatively impact users.  It was intentional to fall back to MPI's gender when gender was nil from the identity provider.  Can you please revert it?

modules/vaos/app/services/vaos/jwt_wrapper.rb
```diff
@@ -38,25 +38,13 @@ def payload
...
+        ssn: user.ssn
       }
     end
 
-    def icn
-      user.icn
-    end
-
-    def first_name
-      user.mpi&.profile&.given_names&.first
```
annaswims, line 51: @saneshark Did you see these changes?  I assume you made an active choice to get the data from MPI over the user object.

lib/evss/disability_compensation_auth_headers.rb
bosawt: The `@user.gender` function that this leverages currently falls back to the MPI gender. Although there is a check for `mhv_icn` to use the MPI Profile gender, so there is a slight difference. Would that be applicable in this case?

annaswims: no, the mhv_icn is only present when MHV is the identity provider. if the gender is nil in ID.me or DS logon this change will make the request fail.

modules/vaos/app/services/vaos/jwt_wrapper.rb
saneshark: We made an active choice to derive from MPI because VAMF uses MPI. It's necessary that we provide the MPI values here vs whatever might be in user since not all paths to LOA3 are necessarily derived from IAM / SSOE.

saneshark: The JWT that we send to VAMF to authenticate against their APIs will only work if it is consistent with what they can lookup via ICN on their side. 

For example on staging there is a user Judy Morrison, when you sign in as this user first_name is "Judy", but MPI shows "JUDY." These types of differences would cause issues with our ability to authenticate against the VAMF services we are integrated with.

saneshark: I would have liked to comment on these changes prior to them getting merged since it affects VAOS services. Can we make sure to include VAOS backend team when making changes to modules/vaos/* in the future?

bramleyjl: Absolutely, I'll make sure to include you and @annaswims when I submit the refactored pull request and anything else that touches the VAOS module.

PR #6164, opened 2021-03-12
Adding migration and schema

cshupp1:
## Description of change
These migrations enable tracking of file size and status changes in benefits intake API

## Original issue(s)
https://vajira.max.gov/browse/API-5293
https://vajira.max.gov/browse/API-5694

migration only


gpbowman-git (approved):
looks good

PR #6166, opened 2021-03-12
Use correct case for va profile address types

lihanli:
## Description of change
in the contact information api the `address_type` should be titleized. right now all caps are accepted but this may change in the future.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/21031

tested in staging


thilton-oddball (approved):
Looks good

PR #6171, opened 2021-03-15
[API-5536] - Update SpecialIssueUpdater worker to use simple arguments

jw81:
## Description of change

Switches the ClaimsAPI SpecialIssuesUpdater worker to require simple arguments. This should prevent the problems that occur when marshalling/unmarshalling complex JSON objects into and out of REDIS.

## Original issue(s)
[API-5536](https://vajira.max.gov/browse/API-5536)

### DIFF
--- a/modules/claims_api/spec/workers/special_issue_updater_spec.rb
+++ b/modules/claims_api/spec/workers/special_issue_updater_spec.rb
@@ -9,14 +9,18 @@ RSpec.describe ClaimsApi::SpecialIssueUpdater, type: :job do
     Sidekiq::Worker.clear_all
   end
 
-  let(:user) { FactoryBot.create(:evss_user, :loa3) }
+  let(:user) do
+    {
+      'ssn' => '796043735'
+    }
+  end
   let(:contention_id) { { claim_id: '123', code: '200', name: 'contention-name-here' } }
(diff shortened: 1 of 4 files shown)


modules/claims_api/spec/workers/special_issue_updater_spec.rb
```diff
@@ -9,14 +9,18 @@
     Sidekiq::Worker.clear_all
   end
 
-  let(:user) { FactoryBot.create(:evss_user, :loa3) }
+  let(:user) do
+    {
+      'ssn' => '796043735'
+    }
+  end
```
SlakrHakr, line 16: might be worth keeping the user factory generated here and then just build the hash to pass in off of that factory record. that way this ssn lives only in the factory, in case it ever needs to change or something.

SlakrHakr (approved):

modules/claims_api/spec/workers/special_issue_updater_spec.rb
jw81: Good idea.
I had made a similar change over in the FlashUpdater spec the other day.  Went ahead and made this change there too.
[See](https://github.com/department-of-veterans-affairs/vets-api/pull/6171/commits/c0e9c1b08abbb50b770b9722c05760d89e9dba5b)

PR #6173, opened 2021-03-15
Update EducationStemAutomatedDecision to have email sent timestamps

zurbergram:
## Description of change

Add new columns to `education_stem_automated_decision` table and usages of columns.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#21522

### DIFF
--- a/config/sidekiq_scheduler.yml
+++ b/config/sidekiq_scheduler.yml
@@ -24,7 +24,7 @@ EducationForm::Create10203SpoolSubmissionsReport:
   description: "Send the daily 10203 report to the call center about spool file submissions"
 
 EducationForm::Create10203ApplicantDecisionLetters:
-  cron: "45 4 * * MON-FRI America/New_York"
+  every: '2m'
   description: "Send 10203 applicant decision letters to applicants"
 
 Preneeds::DeleteOldUploads:
--- a/db/migrate/20210315162119_add_email_sent_at_to_education_stem_automated_decision.rb
+++ b/db/migrate/20210315162119_add_email_sent_at_to_education_stem_automated_decision.rb
@@ -1,0 +1,6 @@
+class AddEmailSentAtToEducationStemAutomatedDecision < ActiveRecord::Migration[6.0]
+  def change
+    add_column :education_stem_automated_decisions, :denial_email_sent_at, :timestamp
+    add_column :education_stem_automated_decisions, :confirmation_email_sent_at, :timestamp
+  end
+end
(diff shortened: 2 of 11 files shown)


dmccurdybah (approved):
LGTM

sonntag-adam (approved):

config/sidekiq_scheduler.yml
```diff
@@ -24,7 +24,7 @@ EducationForm::Create10203SpoolSubmissionsReport:
   description: "Send the daily 10203 report to the call center about spool file submissions"
 
 EducationForm::Create10203ApplicantDecisionLetters:
-  cron: "45 4 * * MON-FRI America/New_York"
+  every: '2m'
```
thilton-oddball, line 27: This is going from running M-F once a day to running every 2 minutes? That seems like quite a jump. Just want to verify that this is intended.

db/migrate/20210315162119_add_email_sent_at_to_education_stem_automated_decision.rb
```diff
@@ -0,0 +1,6 @@
+class AddEmailSentAtToEducationStemAutomatedDecision < ActiveRecord::Migration[6.0]
```
thilton-oddball, line 1: This migration and the schema change will need to be removed to their own PR. Dangerbot won't allow any other changes to be included with migrations.

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_03_12_211214) do
+ActiveRecord::Schema.define(version: 2021_03_15_162119) do
```
thilton-oddball, line 13: As mentioned above, this will need to be in a separate PR with the migration.

thilton-oddball (changes requested):
One question on the scheduler and then the migration and schema change need to be removed from this PR and placed in a separate PR.

config/sidekiq_scheduler.yml
zurbergram: Updating frequency for UAT testing. This change will be reverted after UAT testing

db/migrate/20210315162119_add_email_sent_at_to_education_stem_automated_decision.rb
zurbergram: https://github.com/department-of-veterans-affairs/vets-api/pull/6176

db/schema.rb
zurbergram: https://github.com/department-of-veterans-affairs/vets-api/pull/6176

thilton-oddball (approved):
All looks good

PR #6179, opened 2021-03-16
20352 aws script reporting

ajmagdub:
Created scripts to query Vets API Rails servers logs. The results of the CloudWatch queries are cached in local Redis datastore for quick analysis.

## Description of change

Created the Ruby scripts which are to be run from the command line and drive the CloudWatch queries and Redis caching.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#20352

The scripts require a local Redis instance running on the computer where the scripts are to be run. Access to AWS CloudWatch is required. Also a command line utility awslogs should also be installed on the local computer along with the devops script issue_mfa.sh

The scripts were tested by running locally and verification that the appropriate CloudWatch records were retrieved and stored in Redis.


saneshark (approved):
LGTM

Would like for Redis to just be a dependency in future iteration, but not a core competency for utilizing the tool. Ideally you would have a ruby interface or something for doing that instead.

PR #6180, opened 2021-03-16
Audit quick fixes

SlakrHakr:
## Description of change
Just a couple quick fixes i found when doing my audit.
Misspelling in the docs and one of BGS's services not being optional as it is currently defined.

## Original issue(s)
https://vajira.max.gov/browse/API-5769


modules/claims_api/config/initializers/okcomputer.rb
```diff
@@ -97,4 +97,4 @@ def name
 OkComputer::Registry.register 'bgs-contention', BgsCheck.new('contention')
 OkComputer::Registry.register 'vbms', VbmsCheck.new
 
-OkComputer.make_optional %w[vbms bgs-vet_record bgs-claimant bgs-contention]
+OkComputer.make_optional %w[vbms bgs-vet_record bgs-contention]
```
SlakrHakr, line 100: bgs-claimant is actually used for detecting poa if the api consumer is a Veteran representative

jw81 (approved):

PR #6182, opened 2021-03-16
Updating LOB to remove LOG MED BUR DROC. We need to come back and rem…

mviterna:
…ove it from the actual endpoint. Doing swagger first for an upcoming demo

## Original issue(s)
https://vajira.max.gov/browse/API-5554

Tested local


gpbowman-git (approved):
looking good

PR #6188, opened 2021-03-17
Last fix and some additional line breaks

mviterna:
## Original issue(s)
https://vajira.max.gov/browse/API-5594

Tested locally


gpbowman-git (approved):
looks good

PR #6190, opened 2021-03-17
Update awesome print to latest minor version

LindseySaari:
## Description of change
As the next step in the gem update series, this PR updates the awesome_print gem to the latest minor version. It's best to update each gem individually, in order to avoid issues when rolling back.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#20261

## Testing Completed
- [x] ci make target passing locally


thilton-oddball (approved):
LGTM

PR #6194, opened 2021-03-17
OpenAPI Docs - Added file attachment notes to create message and reply

narin:
## Description of change
Added some additional notes about file attachment restrictions to new message and reply endpoints.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#20472


thilton-oddball (approved):
Looks good

PR #6195, opened 2021-03-18
21584 delete /mobile/v0/user/emails

kreek:
## Description of change
Adds `DELETE /mobile/v0/user/emails` which performs an async Vet360 email PUT to update the email's `effective_end_date` to the current UTC time.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#21584


jperk51 (approved):
LGTM

PR #6203, opened 2021-03-18
Mark.viterna/api 4108 row id migration

mviterna:
## Description of change

Adding a rowid column to va_forms_forms

## Original issue(s)
https://vajira.max.gov/browse/API-4108

tested locally


mviterna:
Migration file out of date

PR #6206, opened 2021-03-18
removes redundant User getter methods [#5411]

bramleyjl:
This is an updated pull request from https://github.com/department-of-veterans-affairs/vets-api/pull/6160, moving direct-to-MPI calls to within the User model getter methods.

## Description of change

This change is part of the larger change to consolidate User & Identity functionality in preparation for the planned Identity Module / Component.

## Original issue(s)

Updated unit tests for new User model getter methods


app/models/user.rb
```diff
@@ -61,8 +61,12 @@ def pciu_alternate_phone
     pciu&.get_alternate_phone&.to_s
   end
 
-  def first_name
-    identity.first_name || (mhv_icn.present? ? mpi&.profile&.given_names&.first : nil)
+  def first_name(args = { mpi: false })
```
saneshark, line 64: consider replacing this with the more idiomatic: [ruby keyword arguments](https://thoughtbot.com/upcase/videos/ruby-keyword-arguments)

Or if you don't want named arguments consider using splat syntax (though that is less clear / readable).

Additionally, since you're running into some complexity issues, I wonder if it would make sense to use the decorator patten (while admittedly not a fan, it might be a good way to abstract the source / origin of some these first names.

app/models/user.rb
```diff
@@ -82,12 +86,20 @@ def middle_name
     identity.middle_name || (mhv_icn.present? ? mpi&.profile&.given_names.to_a[1..-1]&.join(' ').presence : nil)
   end
 
-  def last_name
-    identity.last_name || (mhv_icn.present? ? mpi&.profile&.family_name : nil)
+  def last_name(args = { mpi: false })
```
saneshark, line 89: Same comment as above. This isn't idiomatic ruby.

app/models/user.rb
```diff
@@ -82,12 +86,20 @@ def middle_name
...
+    if args[:mpi] == true
+      mpi&.profile&.family_name
+    else
+      identity.last_name || (mhv_icn.present? ? mpi&.profile&.family_name : nil)
+    end
   end
 
-  def gender
-    identity.gender || (mhv_icn.present? ? mpi&.profile&.gender : nil)
+  def gender(args = { mpi: false })
```
saneshark, line 97: Same comment as above. This isn't idiomatic ruby.

app/models/user.rb
```diff
@@ -122,8 +134,12 @@ def zip
     identity.zip || (mhv_icn.present? ? mpi&.profile&.address&.postal_code : nil)
   end
 
-  def ssn
-    identity.ssn || (mhv_icn.present? ? mpi&.profile&.ssn : nil)
+  def ssn(args = { mpi: false })
```
saneshark, line 137: Same comment as above. This isn't idiomatic ruby.

app/models/user.rb
```diff
@@ -163,8 +179,12 @@ def loa
   delegate :military_person?, to: :veteran_status
   delegate :veteran?, to: :veteran_status
 
-  def edipi
-    loa3? && dslogon_edipi.present? ? dslogon_edipi : mpi&.edipi
+  def edipi(args = { mpi: false })
```
saneshark, line 182: Same comment as above. This isn't idiomatic ruby.

app/models/user.rb
```diff
@@ -61,8 +61,12 @@ def pciu_alternate_phone
     pciu&.get_alternate_phone&.to_s
   end
 
-  def first_name
-    identity.first_name || (mhv_icn.present? ? mpi&.profile&.given_names&.first : nil)
+  def first_name(args = { mpi: false })
+    if args[:mpi] == true
```
bosawt, line 65: I might even suggest that we don't combine these into singular functions. I would consider having two functions, `first_name`, and `first_name_mpi`. Once we're 100% SSOe we can just find/replace all instances of `first_name_mpi` to `first_name`.

bosawt (approved):
I think this looks good now, lets us pull the complexity of the larger `MPI::Profile` structure into the `User` class while preserving functionality by allowing us to directly reference specific MPI fields.

thilton-oddball (approved):
Looks good!

saneshark:
I think this broke some vaos specs. I am not sure how this was not caught by CI.

rspec ./modules/vaos/spec/request/appointment_requests_request_spec.rb:107 # vaos appointment requests POST /vaos/v0/appointment_requests creates a new appointment request
rspec ./modules/vaos/spec/request/appointment_requests_request_spec.rb:119 # vaos appointment requests POST /vaos/v0/appointment_requests Community Cares creates a new appointment request
rspec ./modules/vaos/spec/request/appointment_requests_request_spec.rb:161 # vaos appointment requests PUT /vaos/v0/appointment_requests/:id cancels an appointment request
rspec ./modules/vaos/spec/models/cc_appointment_request_form_spec.rb:12 # VAOS::CCAppointmentRequestForm responds to the correct attributes
rspec ./modules/vaos/spec/models/cc_appointment_request_form_spec.rb:74 # VAOS::CCAppointmentRequestForm has preferred providers as an array
rspec ./modules/vaos/spec/models/cc_appointment_request_form_spec.rb:68 # VAOS::CCAppointmentRequestForm has facilities with an array of children
rspec ./modules/vaos/spec/services/appointment_requests_service_spec.rb:28 # VAOS::AppointmentRequestsService#post_request creates a new community care appointment request creates a new CC appointment request
rspec ./modules/vaos/spec/services/appointment_requests_service_spec.rb:16 # VAOS::AppointmentRequestsService#post_request with valid creation attributes from factory creates a new appointment request
rspec ./modules/vaos/spec/services/appointment_requests_service_spec.rb:58 # VAOS::AppointmentRequestsService put_request with valid cancellation attributes from factory cancels a pending appointment request

PR #6210, opened 2021-03-19
NOD - Create Skeleton Controller

edmangimelli:
putting this skeleton in place so that the following PRs (adding methods to the controller) can be reviewed in parallel

### DIFF
--- a/app/controllers/v0/notice_of_disagreements_controller.rb
+++ b/app/controllers/v0/notice_of_disagreements_controller.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module V0
+  class NoticeOfDisagreementsController < AppealsBaseController
+    private
+
+    def error_class(method:, exception_class:)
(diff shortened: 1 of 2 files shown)


app/controllers/v0/notice_of_disagreements_controller.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module V0
+  class NoticeOfDisagreementsController < AppealsBaseController
+    private
+
+    def error_class(method:, exception_class:)
+      "#{self.class.name}##{method} exception #{exception_class} (NOD)"
```
thilton-oddball, line 8: Just want to verify that the double # is intended.

thilton-oddball (approved):
One comment, otherwise looks good.

app/controllers/v0/notice_of_disagreements_controller.rb
edmangimelli: yep! it is. good 👁️

PR #6216, opened 2021-03-19
[21004] Add relationship object to MPI Profile

bosawt:
## Description of change
This PR adds the `relationships` field to the `MVIProfile` class, which is a recursive `MVIProfile` that defines an entity with a relationship to the MPI User. This definition is parsed from the `personalRelationship` tag in an MPI Response

## Original issue(s)
department-of-veterans-affairs/va.gov-team/issues/21004

- There shouldn't be any changes on the frontend at this point, and we currently do not have access to relationship data with MPI, so basically testing should just show that this currently doesn't break anything.
- On local, if you check out the branch `https://github.com/department-of-veterans-affairs/vets-api-mockdata/tree/5987_development_branch_mpi_relationship_mock`, and sign in with `vets.gov.user+dependent.1@gmail.com`, you should get a mocked response from MPI with relationship information

### DIFF
--- a/lib/mpi/responses/profile_parser.rb
+++ b/lib/mpi/responses/profile_parser.rb
@@ -20,14 +20,10 @@ module MPI
       SUBJECT_XPATH = 'controlActProcess/subject'
       PATIENT_XPATH = 'registrationEvent/subject1/patient'
       STATUS_XPATH = 'statusCode/@code'
-      GENDER_XPATH = 'patientPerson/administrativeGenderCode/@code'
-      DOB_XPATH = 'patientPerson/birthTime/@value'
-      SSN_XPATH = 'patientPerson/asOtherIDs'
-      NAME_XPATH = 'patientPerson/name'
-      ADDRESS_XPATH = 'patientPerson/addr'
-      PHONE = 'patientPerson/telecom'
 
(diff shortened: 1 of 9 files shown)


lib/mpi/responses/profile_parser.rb
```diff
@@ -75,17 +81,17 @@ def parse
       # rubocop:disable Metrics/MethodLength
       # rubocop:disable Metrics/AbcSize
       def build_mvi_profile(patient)
```
jimiray, line 83: is it possible that the bulid_mvi_profile and build_relationship_mvi_profile could be refactored to take a "root" as an argument? Instead of having 2 different methods?

bosawt: Yeah that's not a bad point, there is a lot of repeated stuff here. Given that we don't fully know a lot about what to expect from the relationships array until we can actually grab it from MPI, I'm also considering severely scaling back the fields we parse from it, maybe even create an `MPIRelationshipProfile` model to differentiate

bosawt: Alright, lemme know how it looks now

jimiray (approved):

ericboehs (approved):

PR #6228, opened 2021-03-19
Updates mobile SM API docs with additional details

patrickvinograd:
## Description of change
Adds more details to several of the SM API docs. 

No testing or other code concerns, exclusively a documentation update


modules/mobile/docs/schemas/SecureMessageReplyRequest.yml
```diff
@@ -1,6 +1,22 @@
 type: object
 required:
+  - recipient_id 
+  - category
+  - subject
   - body
 properties:
+  recipient_id:
```
patrickvinograd, line 8: Actually I was wondering if recipient_id is required for replies, or if there's validation that the recipient matches the sender of the replied-to message. That would be good to clarify.

patrickvinograd:
Happy to have additional doc updates incorporated into this PR.

modules/mobile/docs/schemas/SecureMessageReplyRequest.yml
narin: In my testing, I've found that aside from file uploads, only `body` is required/allowed for replies.  `recipient_id` is ignored on the MHV side.  I'll make an update to our API's validation.

narin (approved):
Fixed a couple merge conflicts. LGTM!

PR #6235, opened 2021-03-22
NOD - Add "Create NOD" Method to DecisionReview::Service

edmangimelli:
(for the [new NOD product](https://github.com/department-of-veterans-affairs/va.gov-team/issues/2140))


app/models/user.rb
```diff
@@ -82,6 +82,11 @@ def middle_name
     identity.middle_name || (mhv_icn.present? ? mpi&.profile&.given_names.to_a[1..-1]&.join(' ').presence : nil)
   end
 
+  # returns a string or nil (never returns a ".blank?" string)
+  def middle_initial(append = '')
```
bosawt, line 86: Curious what the need of `append` functionality is here? It doesn't look like it's used

bosawt: Overall we're planning on moving the presentation layer outside of the User model (for example, providing just the `middle_name` and leaving it up to classes outside of `User` to extract the Middle Name from it), so my preference would be to just have the line in `lib/decision_review/service.rb `: `'X-VA-Middle-Initial' => middle_name.to_s.strip.first.upcase`.

However, especially if you foresee other places needing `middle_initial` in the future, I won't push back too hard if you prefer it here

bosawt (commented):
Just a quick couple of comments about the change in the `User` model!

app/models/user.rb
edmangimelli: at one point I was thinking I'd need append, but, you're right, I didn't end up using it. I do think it's a nice feature, though: (only append if there is a middle initial)

```
user.middle_name # Jacob
"#{user.first_name}--#{user.middle_initial('--'}#{user.last_name}" # "John--J--Jingleheimer"

user.middle_name # nil
"#{user.first_name}--#{user.middle_initial('--'}#{user.last_name}" # "John--Jingleheimer"
```

but, not a hill i want to die on 👍

bosawt: Oh I see, ya that totally makes sense. If we do feel like we need that functionality and/or start needing Middle Initial in multiple places, I think we can revisit whether it belongs in the User model for sure.

bosawt (approved):
Considering there aren't changes left in Identity-related code anymore, that's an easy approval on my end!

Also while I was reviewing I just looked through the usage of the different `User` fields, and it looks like the function chains won't raise no-method errors regardless of presence of these fields, so passes my quick sanity check

thilton-oddball (approved):
Looks good!

PR #6238, opened 2021-03-22
Change vet360 reference in settings.yml

lihanli:
## Description of change

Change vet360 reference in settings.yml to preferentially use `va_profile` key

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/18259

tested with unit tests

### DIFF
--- a/app/models/form_profile.rb
+++ b/app/models/form_profile.rb
@@ -2,6 +2,7 @@
 
 require 'string_helpers'
 require 'sentry_logging'
+require 'va_profile/configuration'
 
 # TODO(AJD): Virtus POROs for now, will become ActiveRecord when the profile is persisted
 class FormFullName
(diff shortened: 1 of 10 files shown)


app/models/form_profile.rb
```diff
@@ -276,7 +277,7 @@ def vet360_contact_info
     return @vet360_contact_info if @vet360_contact_info_retrieved
 
     @vet360_contact_info_retrieved = true
-    if Settings.vet360.prefill && user.vet360_id.present?
+    if VAProfile::Configuration::SETTINGS.prefill && user.vet360_id.present?
```
thilton-oddball, line 280: It looks like prefill is an option in some of the other entries in settings.yml but it doesn't look like it is part of vet360/va_profile. Granted, this isn't a change that was made in this PR but it might be worth looking at.

thilton-oddball (approved):
One comment worth looking into but otherwise looks good.

app/models/form_profile.rb
lihanli: looks like it's set in the devops settings.yml files

bosawt (approved):
Looks good, just a simple comment but non-blocking

lib/va_profile/configuration.rb
```diff
@@ -5,6 +5,8 @@
 
 module VAProfile
   class Configuration < Common::Client::Configuration::REST
+    SETTINGS = Settings.va_profile || Settings.vet360
```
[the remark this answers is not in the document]
lihanli: i'm not sure when i'll be able to merge the devops change to settings.yml so i wanted to make sure vets-api was compatible with both versions

PR #6243, opened 2021-03-23
API-5227 - evidence upload error handling

nathanbwright:
## Description of change
This PR adds basic error handling for if something goes wrong during the upload to s3 process.

Two questions:
- What should the status verbiage be if an error occurs?
- What should we be pushing to sentry if an error occurs?

## Original issue(s)
https://vajira.max.gov/browse/API-5227


ElleeB (approved):
I'll stay tuned for the discussion around your questions, but as far as basic requirement, LGTM!

modules/appeals_api/app/services/appeals_api/supporting_evidence/evidence_uploader.rb
```diff
@@ -15,7 +17,11 @@ def initialize(appeal, document, type:)
       def process!
         generate_evidence_submission!
         uploader.store!(document)
-        update_submission!
+        update_submission!('submitted')
+      rescue => e
+        log_message_to_sentry('Error saving to S3', :warning, error: e.to_s)
+        update_submission!('s3_error')
```
cilestin, line 23: I've actually found (while dealing with HLR errors) that having the error code & details stored with the model has helped greatly in triaging. What do you think about adding `code` and `details` to the EvidenceSubmissions model that we can stuff with these exception details as well as sending to sentry?

Even if we don't go that direction, I think the status should just be `error` since we'll use sentry to track down the problem with that GUID

modules/appeals_api/app/services/appeals_api/supporting_evidence/evidence_uploader.rb
```diff
@@ -15,7 +17,11 @@ def initialize(appeal, document, type:)
       def process!
         generate_evidence_submission!
         uploader.store!(document)
-        update_submission!
+        update_submission!('submitted')
+      rescue => e
+        log_message_to_sentry('Error saving to S3', :warning, error: e.to_s)
```
cilestin, line 22: Let's make sure we're also sending the GUID of the submission and the exception message to sentry. Maybe the supportable_id and type too?

cilestin (changes requested):

modules/appeals_api/app/services/appeals_api/supporting_evidence/evidence_uploader.rb
ElleeB: 👀

cilestin (approved):

cilestin (approved):

PR #6250, opened 2021-03-23
18533 copy ppms controller to engine

darkbushido:
## Description of change

Copying the CCP controller over to the rails engine. This also needed the PPMS serializers.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#18533

There is an issue with the request specs. if I don't use the full URL it tries to make requests to example.com


modules/facilities_api/spec/spec_helper.rb
```diff
@@ -2,14 +2,16 @@
 
 # Configure Rails Envinronment
 ENV['RAILS_ENV'] = 'test'
-# require File.expand_path('dummy/config/environment.rb', __dir__)
+require File.expand_path('dummy/config/environment.rb', __dir__)
```
ericboehs, line 5: Does this file exist anywhere? Trying to figure out how where `dummy/config/environment.rb` lives.

darkbushido: Not really. I was trying to get specs to run from within the engine and forgot to recommend it.
you can see that the [generator](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/generators/module/templates/spec/spec_helper.rb.erb#L5) we are using adds that in as well

ericboehs (approved):

PR #6259, opened 2021-03-24
Pv covid vax facilities endpoint

patrickvinograd:
## Description of change
Adds an endpoint `covid_vaccine/v0/facilities/:zip`  for finding  the nearest facilities to a specified zip code. Allows specifying a `count` query parameter to adjust the number of results, defaults to 5. 

For now this is using the same logic as the vetext submission to determine which facilities are VAMCs. We may refine this to filter from an allow list either in this PR or a follow-on if we can generate such a list. 

Returns a 422 on an invalid zip. Returns a 502 on any error from the facilities API. 

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/21968

* There may eventually be a settings update to incorporate the allow list, but for now, no settings changes. 

Tested locally against the lighthouse facilities API sandbox.


patrickvinograd:
Typical response:

```
$ curl 'http://localhost:3000/covid_vaccine/v0/facilities/60607' | jq '.'

{
  "data": [
    {
      "id": "vha_537",
      "type": "vaccination_facility",
      "attributes": {
        "name": "Jesse Brown Department of Veterans Affairs Medical Center",
        "distance": 1.41,
        "city": "Chicago",
        "state": "IL"
      }
    },
    {
      "id": "vha_578",
      "type": "vaccination_facility",
      "attributes": {
        "name": "Edward Hines Junior Hospital",
        "distance": 9.6,
        "city": "Hines",
        "state": "IL"
      }
    },
    {
      "id": "vha_556",
      "type": "vaccination_facility",
      "attributes": {
        "name": "Captain James A. Lovell Federal Health Care Center",
        "distance": 31.74,
        "city": "North Chicago",
        "state": "IL"
      }
    },
    {
      "id": "vha_695",
      "type": "vaccination_facility",
      "attributes": {
        "name": "Clement J. Zablocki Veterans' Administration Medical Center",
        "distance": 81.03,
        "city": "Milwaukee",
        "state": "WI"
      }
    },
    {
      "id": "vha_550",
      "type": "vaccination_facility",
      "attributes": {
        "name": "Danville VA Medical Center",
        "distance": 120.48,
        "city": "Danville",
        "state": "IL"
      }
    }
  ]
}
```

batemapf:
Will this handle addresses in Mexico, Canada, and/or Philippines?

batemapf:
Is it possible to run this with `count` set to a large number (i.e. >200) to return all possibilities? If so, please run and send over a dump of all of the results so I can run that list by VHA and confirm that all possibilities will be vaccinating / ask them to identify who we need to black list.

modules/covid_vaccine/app/services/covid_vaccine/v0/facility_suggestion_service.rb
```diff
@@ -0,0 +1,71 @@
...
+            state: f.address.dig('physical', 'state') }
+        end
+
+        # Get 50 nearest health facilities inclusive of clinics, etc, and
+        # then filter down to VAMCs only.
+        # There may not be the number of requested VAMCs in this list
+        # if there are many clinics closer by, but return as many as are
+        # available up to requested count.
+        response = client.get_facilities(lat: lat, long: lng,
+                                         per_page: 50, type: 'health')
```
batemapf, line 59: There is a `classification` attribute in the `FacilitiesResponse` schema that we might be able to use here to return only `VA Medical Center (VAMC)`'s. Should return ~171 facilities; one of them might not be active. If that is an accurate list according to VHA, we should be able to just use that.

modules/covid_vaccine/app/services/covid_vaccine/v0/facility_suggestion_service.rb
```diff
@@ -0,0 +1,71 @@
...
+      # longer than 3 digit station ID
+      CONSOLIDATED_FACILITIES = %w[528 528A4 528A5 528A6 528A7 528A8
+                                   549 589 589A4 589A5 589A6 589A7
+                                   620 620A4 626 626A4 636 636A6
+                                   636A8 657 657A4 657A5].freeze
+
+      # return a list of nearby facilities based on provided zipcode
+      def facilities_for(zipcode, count = 3)
+        zcta_row = ZCTA[zipcode&.[](0...5)]
+        raise Common::Exceptions::UnprocessableEntity.new(detail: 'Invalid ZIP Code') if zcta_row.blank?
```
batemapf, line 24: @stephenBarrs @cohnjesse Will need a FE error state to handle unknown zip codes. The user should be told something along the lines of, "We were unable to recognize your zip code to offer you a facility preference. You will be registered with the medical center nearest to your given address and will be contacted by that facility for scheduling."

On the backend, these will likely go into some kind of dead letter queue for enrollment folks to handle. But TBD on that :-)

batemapf: https://github.com/department-of-veterans-affairs/va.gov-team/issues/22001

patrickvinograd:
This will not support Mexico/Canada/Phillipines. 

There is an address validation service as part  of VA Profile that will geocode a mailing addess for you. I don't know what countries it supports either.

modules/covid_vaccine/app/services/covid_vaccine/v0/facility_suggestion_service.rb
patrickvinograd: I don't  think the lighthouse facilities API lets you filter by that at runtime, but I'll generate that list offline and see if it looks right. I think there is also a field for an active flag.

modules/covid_vaccine/app/services/covid_vaccine/v0/facility_suggestion_service.rb
```diff
@@ -0,0 +1,71 @@
...
+        response = client.get_facilities(lat: lat, long: lng,
+                                         per_page: 50, type: 'health')
+
+        result = response.select(&allowed_vamcs).map(&elements)
+        result = result.first(count)
+        result
+      rescue => e
+        # For now just log any exception while getting facilities and return an empty result
+        log_exception_to_sentry(e)
+        raise Common::Exceptions::BadGateway
```
saneshark, line 67: The assumption that the error resulted from downstream rather than an exception we caused makes me chuckle a bit, but this is probably fine.

saneshark (approved):
Left a couple comments. Looks good to me.

modules/covid_vaccine/app/controllers/covid_vaccine/v0/facilities_controller.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+require_relative '../../../serializers/covid_vaccine/v0/registration_submission_serializer'
+require_relative '../../../services/covid_vaccine/v0/facility_suggestion_service'
+
+module CovidVaccine
+  module V0
+    class FacilitiesController < CovidVaccine::ApplicationController
+      include IgnoreNotFound
```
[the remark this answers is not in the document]
patrickvinograd: We shouldn't even generate any 404s so removing this line.

PR #6260, opened 2021-03-24
21005 add relationship hash to user model

bosawt:
## Description of change
This PR adds a function on the `User` model, `relationships`, that describes the parsed relationships from the `MPI` response. Right now these `relationships` have just a subset of the MPI data for a relationship. We'll expand the data presented here as necessary

## Original issue(s)
department-of-veterans-affairs/va.gov-team/issues/21005

- This shouldn't have negative impacts on a user without a relationship defined. 
- For testing, you'll have to check out the `5987_development_branch_mpi_relationship_mock` branch in `vets-api-mockdata`, then log in with `vets.gov.user+dependent.1@gmail.com`
- No data will be serialized to the frontend yet, so you'll have to set a `binding.pry` in the codebase somewhere after the user has been authenticated and `MPI` has been called, one place is in `sessions_controller`, `log_persisted_session_and_warnings` (make sure to skip the first pass and stop the second time this function is called). Then check `current_user.relationships` to confirm it has been filled out

- For testing, I authenticated with a user that has no relationship data and a user that has relationship data, to confirm both work fine.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -359,7 +359,22 @@ class User < Common::RedisStore
     email&.downcase || account_uuid
   end
 
+  def relationships
+    @relationships ||= mpi_profile_relationships.map { |relationship| relationship_hash(relationship) }
+  end
+
   private
+
+  def relationship_hash(mpi_relationship)
+    return unless mpi_relationship
+
+    {
+      first_name: mpi_relationship.given_names&.first,
+      last_name: mpi_relationship.family_name,
+      birth_date: mpi_relationship.birth_date,
+      person_type_code: mpi_relationship.person_type_code
+    }
+  end
 
   def mpi_profile
     return nil unless mpi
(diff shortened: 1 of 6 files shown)


app/models/user.rb
```diff
@@ -378,6 +393,12 @@ def mpi_profile_birth_date
     mpi_profile.birth_date
   end
 
+  def mpi_profile_relationships
```
jimiray, line 396: Do we need this? Or can we just use mpi_profile.relationships?

jimiray (approved):

app/models/user.rb
bosawt: My thinking generally with these functions has been to eventually move direct access of `mpi_profile` itself either outside of User completely or with very restricted cases. Also I think in this case I wanted to make sure we returned an empty array on nil because I call `.map` on this function

ericboehs (approved):

PR #6271, opened 2021-03-24
update simplecov to latest version

rileyanderson:
## Description of change
As part of the gem update series, this PR updates the simplecov gem to the latest version. Need to insure that CodeClimate still works after updating this.
Updates:
`simplecov -> 0.21.2`
`docile -> 1.3.5`
`json- > 2.5.1`
`simplecov-html -> 0.12.3`
`simplecov_json_formatter -> 0.1.2`

## Original issue(s)
department-of-veterans-affairs/va.gov-team#20261


LindseySaari:
Note: I would also check to make sure that this doesn't break anything in the CircleCI run as well

rileyanderson:
@LindseySaari confirmed this works on circle still. This is ready for review

spec/simplecov_helper.rb
```diff
@@ -32,10 +32,10 @@ def all_results
   end
 
   def merge_results
-    results = all_results.map { |file| SimpleCov::Result.from_hash(JSON.parse(File.read(file))) }
-    SimpleCov::ResultMerger.merge_results(*results).tap do |result|
-      SimpleCov::ResultMerger.store_result(result)
-    end
+    SimpleCov.collate all_results
+  rescue RuntimeError => e
+    puts e
```
LindseySaari, line 37: Looks like a debug statement 😸

Gemfile.lock
```diff
@@ -295,7 +295,7 @@ GEM
     descendants_tracker (0.0.4)
       thread_safe (~> 0.3, >= 0.3.1)
     diff-lcs (1.4.4)
-    docile (1.3.2)
+    docile (1.3.5)
```
LindseySaari, line 298: Could you also note the gem dependencies of simplecov that had to be updated in the PR description please. Looks like docile, simplecov-html and simplecov_json_formatter

Makefile
```diff
@@ -149,7 +149,6 @@ ifeq ($(ENV_ARG), dev)
 	@$(BASH_DEV) "RAILS_ENV=test DISABLE_BOOTSNAP=true NOCOVERAGE=true bundle exec parallel_rspec ${SPEC_PATH}"
 else
 	@$(COMPOSE_TEST) $(BASH) -c "CIRCLE_JOB=true RAILS_ENV=test DISABLE_BOOTSNAP=true bundle exec parallel_rspec ${SPEC_PATH}"
-	@$(BASH_TEST) "DISABLE_BOOTSNAP=true RUN_COVERAGE=true bundle exec rake simplecov:report_coverage"
```
LindseySaari, line 152: Should this be removed?

LindseySaari: Oh, so it looks like if we're not using this anymore, can we just completely remove simplecov_parallel.rake and the simplecov_helper

rileyanderson: From what I could tell it wasn't doing anything. That rake task is specific to circleci file path for coverage and cirlceci doesn't use the this makefile only Jenkins does. 

CircleCI reports coverage here: https://github.com/department-of-veterans-affairs/vets-api/blob/master/.circleci/config.yml#L111  

Jenkins does it here: https://github.com/department-of-veterans-affairs/vets-api/blob/master/Jenkinsfile#L81-L84

rileyanderson: CircleCI still uses `simplecov_parallel.rake` and the `simplecov_helper`

LindseySaari: Ah, ok! Makes sense!

LindseySaari (approved):

PR #6277, opened 2021-03-24
[API-5922] Remove maxLength from NOD Veteran Name fields

cilestin:
## Description of change
Removes the `maxLength` property on the veteran name fields to allow for more flexible names where real world issues have already cropped up. Instead, truncates the concatenated full name to a max of 180 characters.

Adjusts all places where the name was used in the generated pdf - the name field, the signature, and the name & SSN header stamp (now truncated to 35 characters of the last name to fit in the space it has available).

## Original issue(s)
https://vajira.max.gov/browse/API-5922

## Testing done
Tested that the max width for each line would fit both the name field and signature field on the pdf, which ended up being 90 characters per line, or 180 characters total. Will take in more than that, but will truncate the full name after 180 characters in both spots consistently.


modules/appeals_api/app/services/appeals_api/pdf_construction/notice_of_disagreement/v1/form_data.rb
```diff
@@ -59,11 +59,11 @@ def soc_opt_in
         end
 
         def signature
-          first_last = [first_name('Veteran'), last_name('Veteran')]
-          formatted_name = first_last.map(&:presence).compact.map(&:strip).join(' ')
-          auth_statement = 'signed by digital authentication to api.va.gov'
+          first_last = [first_name('Veteran')&.[](0...90), last_name('Veteran')&.[](0...90)]
```
nathanbwright, line 62: Do you think it would be useful to have a couple specs around this truncation logic?

nathanbwright (approved):
Looks good to me!

PR #6285, opened 2021-03-25
adding the new VANotify job for covid vaccine confirmation emails

saneshark:
## Description of change
Introduces a new VANotify Email job. 

## Original issue(s)
none

Just a sidekiq worker to invoke VANotify Service


modules/covid_vaccine/app/jobs/covid_vaccine/confirmation_email_job.rb
```diff
@@ -0,0 +1,56 @@
...
+module CovidVaccine
+  class ConfirmationEmailJob
+    include Sidekiq::Worker
+    include SentryLogging
+    sidekiq_options expires_in: 1.day, retry: 2
+
+    STATSD_ERROR_NAME = 'worker.covid_vaccine_confirmation_email.error'
+    STATSD_SUCCESS_NAME = 'worker.covid_vaccine_confirmation_email.success'
+
+    def perform(email, date, sid)
```
patrickvinograd, line 15: Based on the proposed flow: https://github.com/department-of-veterans-affairs/va.gov-team/issues/22118

there will never be a `sid` at this point in the process because we will not have submitted to vetext at this point in the process. The id that should be used to trigger this job is the record_id, like here:
https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/covid_vaccine/app/jobs/covid_vaccine/submission_job.rb#L10

patrickvinograd: (i.e just the standard activerecord `id` field for the RegistrationSubmission object

modules/covid_vaccine/app/jobs/covid_vaccine/confirmation_email_job.rb
```diff
@@ -0,0 +1,56 @@
...
+      if submission.nil?
+        log_message_to_sentry('No SID found!', :warn, { email: email, sid: sid })
+        return
+      end
+      return if submission.email_confirmation_id.present?
+
+      if Flipper.enabled?(:vanotify_service_enhancement)
+        notify_client ||= VaNotify::Service.new(Settings.vanotify.services.va_gov.api_key)
+        template_id ||= Settings.vanotify.services.va_gov.template_id.covid_vaccine_registration
+      else
```
patrickvinograd, line 26: Remove this else block - by the time we merge this we can just make sure we are putting the template ID in under the new preferred stanza in settings.yml.

config/settings.yml
```diff
@@ -838,6 +839,7 @@ vanotify:
     connection_template: fake_template_id
     disconnection_template: fake_template_id
     covid_vaccine_registration: fake_template_id
+    covid_vaccine_confirmation: fake_template_id
```
patrickvinograd, line 842: You can remove this since it's the obsolete settings location and just use the new preferred location.

patrickvinograd (changes requested):
As mentioned in a comment, `sid` will always be nil when this job is triggered, since this confirmation will be sent before we pass to vetext. So base things off of the record ID. And there will be no sid to pass as a personalization param to the template. 

I'm not sure - is `reference` mandatory for the VANotify API? That's another place where we are currently passing the SID.

modules/covid_vaccine/app/jobs/covid_vaccine/confirmation_email_job.rb
saneshark: oh, for some reason looking at the schema I thought we had replaced id with sid and we were just going to generate something randomish.

ericboehs (approved):

saneshark:
merged separately

PR #6289, opened 2021-03-25
[API-5904] - Deprecate ClaimsApi "schema" and "validate" endpoints

jw81:
## Description of change

Deprecates the "schema" and "validate" endpoints for both v0 and v1 of ClaimsApi.
- Updates Swagger to visually indicate that the endpoints are deprecated.
- Adds logic to return `Deprecation` and optional `Link` response headers

## Original issue(s)
[API-5904](https://vajira.max.gov/browse/API-5904)


modules/claims_api/app/controllers/concerns/claims_api/endpoint_deprecation.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module ClaimsApi
+  module EndpointDeprecation
+    extend ActiveSupport::Concern
+
+    included do
+      V0_DEV_DOCS = 'https://developer.va.gov/explore/benefits/docs/claims?version=0.0.1'
+      V1_DEV_DOCS = 'https://developer.va.gov/explore/benefits/docs/claims?version=current'
```
SlakrHakr, line 9: might be worth putting these in Settings instead?

SlakrHakr, line 9: i wonder if its worth using
https://developer.va.gov/explore/benefits/docs/claims?version=1.0.0
as the url instead.

the above url still redirects to "current" it looks like, but should stay v1 if a v2 is introduced as the new "current" version.

jw81: Ah, ok.
Yeah, I had originally put `version=1.0.0` but when it redirected to "current", I wasn't sure if maybe that url would "break" whenever a new version comes out.
Updated [here](https://github.com/department-of-veterans-affairs/vets-api/pull/6289/commits/8aa37123798926ef1d594dbe486052c0d97392d5).

jw81: Do you mean storing them [here](https://github.com/department-of-veterans-affairs/vets-api/blob/master/config/settings.yml#L414)?

thilton-oddball (approved):
Looks good!

modules/claims_api/app/controllers/concerns/claims_api/endpoint_deprecation.rb
SlakrHakr: yes

SlakrHakr (approved):

PR #6293, opened 2021-03-26
Add db migration for expanded covid registration model

patrickvinograd:
## Description of change
Adds a new DB table/model for the expanded covid vaccine registration workflow.
Because these submissions wind up having a quite different processing model than the original covid vaccine registrations, we opted to fork the model. Key differences:

- We will be resolving some eligibility info and need to store a record of that. This may include lookup data from MPI/EMIS so we encrypt that attribute.
- We will be pushing the objects through various offline ETL processes so we need the `state` and `batch_id` parameters to keep track of where they are in that pipeline. 

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/22118

### DIFF
--- a/db/migrate/20210325230116_add_expanded_covid_registration.rb
+++ b/db/migrate/20210325230116_add_expanded_covid_registration.rb
@@ -1,0 +1,26 @@
+class AddExpandedCovidRegistration < ActiveRecord::Migration[6.0]
+  def change
+    create_table :covid_vaccine_expanded_registration_submissions, id: :serial do |t|
+      t.string "submission_uuid", null: false
+      t.string "vetext_sid"
+      t.boolean "sequestered", default: true, null: false
+      t.string "state"
+      t.string "email_confirmation_id"
+      t.string "enrollment_id"
+      t.string "batch_id"
+      t.string "encrypted_raw_form_data"
+      t.string "encrypted_raw_form_data_iv"
+      t.string "encrypted_eligibility_info"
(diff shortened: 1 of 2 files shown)


db/migrate/20210325230116_add_expanded_covid_registration.rb
```diff
@@ -0,0 +1,26 @@
+class AddExpandedCovidRegistration < ActiveRecord::Migration[6.0]
+  def change
+    create_table :covid_vaccine_expanded_registration_submissions, id: :serial do |t|
+      t.string "submission_uuid", null: false
+      t.string "vetext_sid"
+      t.boolean "sequestered", default: true, null: false
+      t.string "state"
+      t.string "email_confirmation_id"
+      t.string "enrollment_id"
```
saneshark, line 9: enrolled_at might be useful for audit purposes.

saneshark (approved):
LGTM but see comment

db/migrate/20210325230116_add_expanded_covid_registration.rb
patrickvinograd: Since the enrollment is asynchronous (via flat files), I don't know that a timestamp will be that meaningful. I will make sure the eligibility_info includes a timestamp of when the eligibility info was fetched from MPI/EMIS.

thilton-oddball (approved):
LGTM

PR #6295, opened 2021-03-26
Pv expanded vax api

patrickvinograd:
## Description of change
Adds controller for phase 2 of covid vaccine registrations. In this workflow the controller saves the submission and sends an email confirmation, but does no other upstream submission or processing - that will happen in a batch job to be implemented in a subsequent PR.

To be completed:
- [x] Trigger email confirmation job
- [ ] ~Update validation of incoming parameters to match those expected for phase 2 submissions.~

NOTE - will update expected input elements to match frontend schema in a separate PR.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/22118


patrickvinograd:
This rides on top of db migration PR https://github.com/department-of-veterans-affairs/vets-api/pull/6293 and will be rebased once that one is merged.

modules/covid_vaccine/app/models/covid_vaccine/v0/raw_expanded_form_data.rb
```diff
@@ -0,0 +1,26 @@
...
+      ATTRIBUTES = %w[email zip_code vaccine_interest birth_date].freeze
+      ZIP_REGEX = /\A^\d{5}(-\d{4})?$\z/.freeze
+
+      attr_accessor(*ATTRIBUTES)
+
+      validates :email, format: { with: URI::MailTo::EMAIL_REGEXP }
+      validates :vaccine_interest, presence: true
+      validates :zip_code, format: { with: ZIP_REGEX, message: 'should be in the form 12345 or 12345-1234' }
+      validates :birth_date, format: { with: /\A\d{4}-\d{2}-\d{2}\z/, message: 'should be in the form yyyy-mm-dd' },
+                             allow_blank: true
```
batemapf, line 17: why allow_blank here?

batemapf: ah guessing this is just pulled over from KMI and needs updating per your PR comment :-)

modules/covid_vaccine/app/controllers/covid_vaccine/v0/expanded_registration_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+
+      def create
+        raw_form_data = params[:registration]
+        record = CovidVaccine::V0::ExpandedRegistrationSubmission.create!({ submission_uuid: SecureRandom.uuid,
+                                                                            raw_form_data: raw_form_data })
+
+        formatted_date = record.created_at.strftime('%B %-d, %Y %-l:%M %P %Z').sub(/([ap])m/, '\1.m.')
+        CovidVaccine::ExpandedRegistrationEmailJob.perform_async(record.id,
+                                                                 raw_form_data[:email],
+                                                                 formatted_date)
```
saneshark, line 20: These attributes: `raw_form_data`, `created_at` should be available from the record. Why wouldn't we just load it from sidekiq? Is this just a performance tweak because raw_form_data is encrypted?

Any reason formatted_date couldn't live as a method on the job?

modules/covid_vaccine/app/jobs/covid_vaccine/expanded_registration_email_job.rb
```diff
@@ -0,0 +1,51 @@
...
+  class ExpandedRegistrationEmailJob
+    include Sidekiq::Worker
+    include SentryLogging
+    sidekiq_options expires_in: 1.day, retry: 2
+
+    STATSD_ERROR_NAME = 'worker.covid_vaccine_expanded_registration_email.error'
+    STATSD_SUCCESS_NAME = 'worker.covid_vaccine_expanded_registration_email.success'
+
+    def perform(record_id, email, datestring)
+      submission = CovidVaccine::V0::ExpandedRegistrationSubmission.find_by(id: record_id)
```
saneshark, line 16: If you use `find` instead of `find_by` it will raise an ActiveRecord::NotFound which would be just as good and automatically logged by handle_errors. My latest changes included this. It simplifies the guard clauses down to just `submission.email_confirmation_id.present?`

modules/covid_vaccine/app/jobs/covid_vaccine/expanded_registration_email_job.rb
```diff
@@ -0,0 +1,51 @@
...
+      if submission.nil?
+        log_message_to_sentry('No Record found!', :warn, { record_id: record_id, email: email })
+        return
```
saneshark, line 19: this guard clause could go away if you change `find_by` to `find`

modules/covid_vaccine/app/jobs/covid_vaccine/expanded_registration_email_job.rb
```diff
@@ -0,0 +1,51 @@
...
+      end
+      return if submission.email_confirmation_id.present?
+
+      notify_client ||= VaNotify::Service.new(Settings.vanotify.services.va_gov.api_key)
+      template_id ||= Settings.vanotify.services.va_gov.template_id.covid_vaccine_registration
```
saneshark, line 24: I'm not sure `||=` provides much value since this is the only place these variables are going to be initialized.

modules/covid_vaccine/app/jobs/covid_vaccine/expanded_registration_email_job.rb
```diff
@@ -0,0 +1,51 @@
...
+    rescue => e
+      handle_errors(e, submission.submission_uuid)
+    end
+
+    def handle_success(submission, notify_response)
+      submission.update!(email_confirmation_id: notify_response.id)
+      StatsD.increment(STATSD_SUCCESS_NAME)
+    end
+
+    def handle_errors(ex, submission_uuid)
```
saneshark, line 40: If you end up switching to `find` and for some reason wanted to log as a warning instead of error or if you wanted to ensure that Sidekiq doesn't retry for some reason you could rescue ActiveRecord::NotFound here and handle it differently.

modules/covid_vaccine/app/serializers/covid_vaccine/v0/expanded_registration_serializer.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module CovidVaccine
+  module V0
+    class ExpandedRegistrationSerializer < ActiveModel::Serializer
+      attribute :created_at
+
+      def id
+        nil
```
saneshark, line 9: I'm curious why id should be nil and not the record id?

modules/covid_vaccine/app/controllers/covid_vaccine/v0/expanded_registration_controller.rb
patrickvinograd: fair point, this implementation was path-dependent on using the other job as a starting point. I'll refactor so the only argument is the record_id.

modules/covid_vaccine/app/jobs/covid_vaccine/expanded_registration_email_job.rb
patrickvinograd: Sounds good. Changes the behavior slightly, such that it will retry vs return on a not-found ID (which should never happen regardless, but if it does fine to retry and capture in sentry)

modules/covid_vaccine/app/serializers/covid_vaccine/v0/expanded_registration_serializer.rb
patrickvinograd: i don't want to leak any record identifiers back to the unauthenticated client - especially not a sequential DB ID. The submission_uuid could be used here I guess, but also - what ever is going to be done with that value? nothing. Maybe some miniscule value for debugging.

modules/covid_vaccine/app/jobs/covid_vaccine/expanded_registration_email_job.rb
saneshark: yes and might actually be preferred if it's a result of race condition, but you could also handle ActiveRecord::NotFound and not re-raise if it becomes an issue later.

modules/covid_vaccine/app/jobs/covid_vaccine/expanded_registration_email_job.rb
patrickvinograd: I do need to rescue ActiveRecord::NotFound or else do safe navigation on `submission.submission_uuid`

modules/covid_vaccine/app/jobs/covid_vaccine/expanded_registration_email_job.rb
```diff
@@ -0,0 +1,54 @@
...
+      template_id = Settings.vanotify.services.va_gov.template_id.covid_vaccine_registration
+      email_address = submission.raw_form_data[:email]
+
+      notify_response = notify_client.send_email(email_address: email_address, template_id: template_id,
+                                                 personalisation: { 'date' => formatted_date(submission.created_at),
+                                                                    'confirmation_id' => submission.submission_uuid },
+                                                 reference: submission.submission_uuid)
+      handle_success(submission, notify_response)
+    rescue => e
+      handle_errors(e, submission&.submission_uuid)
```
patrickvinograd, line 29: Actually going to just change this to record_id

modules/covid_vaccine/app/jobs/covid_vaccine/expanded_registration_email_job.rb
```diff
@@ -0,0 +1,54 @@
...
+      submission = CovidVaccine::V0::ExpandedRegistrationSubmission.find(record_id)
+      return if submission.email_confirmation_id.present?
+
+      notify_client = VaNotify::Service.new(Settings.vanotify.services.va_gov.api_key)
+      template_id = Settings.vanotify.services.va_gov.template_id.covid_vaccine_registration
+      email_address = submission.raw_form_data[:email]
+
+      notify_response = notify_client.send_email(email_address: email_address, template_id: template_id,
+                                                 personalisation: { 'date' => formatted_date(submission.created_at),
+                                                                    'confirmation_id' => submission.submission_uuid },
```
saneshark, line 25: BTW, I'm not entirely sure if `confirmation_id` is the actual personalization key again, this was just an assumption I made.

saneshark (approved):
LGTM

modules/covid_vaccine/app/jobs/covid_vaccine/expanded_registration_email_job.rb
saneshark: yeah might be easier for debugging; it's also going to be more consistent with what ActiveRecord::NotFound will include based on my testing.

modules/covid_vaccine/app/jobs/covid_vaccine/expanded_registration_email_job.rb
patrickvinograd: It is, I connected to VANotify staging and got errors when I omitted it.

PR #6305, opened 2021-03-26
[6303] Update Jenkins parallel spec run so that modules directory is properly run

bosawt:
## Description of change
This PR updates the Jenkins parallel RSpec run so that the `modules` directory is included  in the spec run, turns out they weren't being run with the previous `parallel_specs` implementation`. Additionally, this adds back the code-climate reporting that was in the previous implementation

## Original issue(s)
department-of-veterans-affairs/vets-api/issues/6303

### DIFF
--- a/.rspec_parallel
+++ b/.rspec_parallel
@@ -5,4 +5,3 @@
 --format RspecJunitFormatter
 --out log/rspec<%= ENV['TEST_ENV_NUMBER'] %>.xml
 <% end %>
-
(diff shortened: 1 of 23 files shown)


bosawt:
@mviterna Hi Mark, specs in `modules` weren't being run for a while, I was able to fix them all with this change, except for a failure in `form_reloader_spec.rb`. It looks like a PR that removed `spec/support/vcr_cassettes/va_forms/gql_forms.yml` broke it. Can I add that back? Or would you like to look at the spec and fix it another way?

thilton-oddball (approved):
LGTM

PR #6314, opened 2021-03-27
output rspec xml for each test database

dillo:
## Description of change

- Output RSpec XML for each test database that we create on the Jenkins CI environment.
## Original issue(s)
department-of-veterans-affairs/va.gov-team#21632


thilton-oddball (approved):
Looks good!

PR #6316, opened 2021-03-27
Adds facility allowlist for covid_vaccine module

patrickvinograd:
## Description of change
We need to filter down the list of VAMCs that users can select from, based on a list provided by @batemapf, with possible future adjustments from VHA. 

Setting will be consumed in a follow-up PR within CODEOWNES-ed module.

- For now this setting will suffice for all environments; if it needs adjusting for any specific env we'll do so in the devops repo.


batemapf:
confirmed this include our one cerner facility in spokane (668). LGTM!

batemapf (approved):

thilton-oddball (approved):
LGTM

PR #6322, opened 2021-03-28
Updates vax expansion validation to match frontend

patrickvinograd:
## Description of change
Updates controller validations to match what is currently being sent by the frontend. 

## Original issue(s)

### DIFF
--- a/modules/covid_vaccine/app/models/covid_vaccine/v0/raw_expanded_form_data.rb
+++ b/modules/covid_vaccine/app/models/covid_vaccine/v0/raw_expanded_form_data.rb
@@ -5,16 +5,34 @@ module CovidVaccine
     class RawExpandedFormData
       include ActiveModel::Validations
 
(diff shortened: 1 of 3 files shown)


modules/covid_vaccine/app/models/covid_vaccine/v0/raw_expanded_form_data.rb
```diff
@@ -5,16 +5,34 @@ module V0
     class RawExpandedFormData
       include ActiveModel::Validations
 
-      ATTRIBUTES = %w[email zip_code vaccine_interest birth_date].freeze
+      ATTRIBUTES = %w[first_name middle_name last_name ssn birth_date birth_sex
```
saneshark, line 8: Why don't you just use %i instead of %w here and make these symbols so that you don't need to do #to_s on line 39 below?

saneshark (approved):
One minor improvement but LGTM

PR #6323, opened 2021-03-28
Changes facilities response on invalid zip

patrickvinograd:
## Description of change
Per agreement with frontend engineers, switching to return a 200 with an empty result for an unknown zip code. 

## Original issue(s)

Tested locally. 
```
curl http://localhost:3000/covid_vaccine/v0/facilities/00000

{"data":[]}
```


patrickvinograd:
We decided to handle the error on invalid zip on the frontend, so this is not needed.

PR #6326, opened 2021-03-28
Filters vaccine locations by allowlist in settings

patrickvinograd:
## Description of change
We expect that we will need to fine-tune the list of suggested facilities for vaccines based on a list provided by VHA. This PR ensures that only the list of facilities in Settings are eligible to be returned. This will allow us to update the list with a config-only deployment. 

## Original issue(s)

I added a test case that ensures that both numeric and alphanumeric facility codes will work correctly.


saneshark (approved):
LGTM

PR #6332, opened 2021-03-29
uses VANotify template for expanded registration

philherbert:
[Commit 1606376](https://github.com/department-of-veterans-affairs/devops/commit/16063769de866ae15a8f42a9d7a3201fc52bdbbb) added template IDs for the new registration email for expanded eligibility.

This commit fixes the `expanded_registration_email_job` (added in #6295) to use the new template ID, instead of the old registration template ID.


saneshark (approved):
LGTM

PR #6333, opened 2021-03-29
Refactor benefits claims api controllers with more separation between v0 and v1

SlakrHakr:
## Description of change
My grape poc didnt get much traction unfortunately: https://github.com/department-of-veterans-affairs/vets-api/pull/5856
So I went about trying to work through how i could improve, imo, our current basic rails api endpoints.
One thing I knew often confused me was how tied together v0 and v1 are. This work attempts to break them apart and simplify some of the logic within each of them.

My goal was to not change any of the specs, in order to (hopefully) prove that this refactor didnt actually touch the functionality of our black box. There was at least one change that removed the need for catching of an exception, which did result in a spec needing to change slightly however.


SlakrHakr:
We're now considering breaking out of vets-api. If we take that approach, we could potentially use grape and reference the poc i performed earlier.
Closing this for now as it would take a bit of a lift to get it across the finish line still.
May come back to it if breaking out of vets-api proves not possible at this time.

lib/common/exceptions/record_not_found.rb
```diff
@@ -9,12 +9,14 @@ module Exceptions
...
 
-      def initialize(id)
+      def initialize(id, detail: nil)
         @id = id
+        @detail = detail
       end
 
       def errors
-        Array(SerializableError.new(i18n_interpolated(detail: { id: @id })))
+        detail = @detail.presence || { id: @id }
```
jw81, line 18: I didn't know `.presence` was a thing!

modules/claims_api/app/controllers/claims_api/v0/claims_controller.rb
```diff
@@ -1,28 +1,66 @@
...
+    class ClaimsController < ClaimsApi::V0::ApplicationController
       def index
         claims = claims_service.all
         render json: claims,
                serializer: ActiveModel::Serializer::CollectionSerializer,
                each_serializer: ClaimsApi::ClaimListSerializer
-      rescue => e
-        log_message_to_sentry('Error in claims v0',
+      rescue EVSS::ErrorMiddleware::EVSSError => e
+        log_message_to_sentry('Error in claims v1',
```
jw81, line 14: Probably mean `Error in claims v0` here right?

modules/claims_api/app/controllers/concerns/claims_api/mpi_verification.rb
```diff
@@ -5,25 +5,8 @@ module MPIVerification
     extend ActiveSupport::Concern
 
     included do
-      before_action :verify_mpi
```
jw81, line 8: I like this.  It seemed odd to me that the concern would be declaring a `before_action` to basically run itself.  It makes more sense (to me) to allow the controller that's pulling this in to decide when the logic should be ran.

`ClaimsApi::PoaVerification` concerns has that same odd pattern.

modules/claims_api/app/controllers/concerns/claims_api/mpi_verification.rb
```diff
@@ -5,25 +5,8 @@ module MPIVerification
...
-               status: :bad_request
-      end
-
-      private
-
-      def inspected_error
-        target_veteran.mpi&.response&.error&.inspect
-      rescue
-        nil
+        raise ::Common::Exceptions::ParameterMissing, 'MPI user' unless target_veteran.mpi_record?
```
jw81, line 9: Any value in keeping some of the "_Not enough Veteran information, functionality limited_" context?
I just wonder if a 400 with a message of "MPI User" is going to be enough to let a consumer know that they haven't supplied enough veteran info.

jw81 (approved):

lib/common/exceptions/record_not_found.rb
SlakrHakr: can thank rubocop's autofix functionality haha

modules/claims_api/app/controllers/claims_api/v0/claims_controller.rb
SlakrHakr: https://github.com/department-of-veterans-affairs/vets-api/pull/6333/commits/6268d295eb63ff60dc7c90fd6dfb44cd73d03ee5#diff-ac7547575b98e9f9b6187dfefb8ed4c4e0356f4cca2472e1499daaa13b670ea4

modules/claims_api/app/controllers/concerns/claims_api/mpi_verification.rb
SlakrHakr: youre right! made the same change to PoaVerification
https://github.com/department-of-veterans-affairs/vets-api/pull/6333/commits/6268d295eb63ff60dc7c90fd6dfb44cd73d03ee5

PR #6338, opened 2021-03-29
char4 code pass-through on appointment model

grossb1:
## Description of change

The work here is to ensure the `char4` code provided by VAMF associated with appointments is passed through to vets-website.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/21540

Associated vcr cassette was updated to reflect this change.

### DIFF
--- a/modules/vaos/app/serializers/vaos/v0/va_appointments_serializer.rb
+++ b/modules/vaos/app/serializers/vaos/v0/va_appointments_serializer.rb
@@ -9,7 +9,7 @@ module VAOS
 
       set_type :va_appointments
 
-      attributes :start_date, :sta6aid, :clinic_id, :clinic_friendly_name, :facility_id, :community_care
+      attributes :start_date, :sta6aid, :clinic_id, :clinic_friendly_name, :facility_id, :char4, :community_care
 
       attribute :vds_appointments do |object|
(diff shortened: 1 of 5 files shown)


modules/vaos/app/serializers/vaos/v0/va_appointments_serializer.rb
```diff
@@ -9,7 +9,7 @@ class VAAppointmentsSerializer
 
       set_type :va_appointments
 
-      attributes :start_date, :sta6aid, :clinic_id, :clinic_friendly_name, :facility_id, :community_care
+      attributes :start_date, :sta6aid, :clinic_id, :clinic_friendly_name, :facility_id, :char4, :community_care
```
jbalboni, line 12: Now that I look at this, we might be missing `:phone_only`, too (`phoneOnly` from MAS). I can create a separate ticket for that, unless you want to add it in this PR?

grossb1: I'll include it here.

thilton-oddball (commented):
I'm getting a failing spec on this. The property '#/data/attributes' did not contain a required property of 'phone_only'

PR #6349, opened 2021-03-29
upload all files

annaswims:
## Description of change
Rather than beginning the next upload after the previous has succeeded enqueue them all with a delay. That way if a document fails it won't prevent the others from being sent. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#20704


thilton-oddball (approved):
Looks good!

PR #6358, opened 2021-03-30
Fixing schema

mviterna:
Fix for the schema and sha256 Forms API

## Original issue(s)


bosawt (approved):
Does it matter that this is still defined in `form_swagger.rb` and the `va_forms/config/locals/en.yml` locations? If it passes tests I suppose I don't care for now but just wanted to mention it still existed in those spots.

PR #6359, opened 2021-03-30
Api 4750 lighthouse token validation ssoi support

dbrown-va:
## Description of change
API-4750 adds a V2 token validation endpoint for Lighthouse API's to support SSOi and Client Credentials auth scenarios where different identifiers were needed for the responses.

https://community.max.gov/display/VAExternal/Token+Validation+v2

## Original issue(s)
https://vajira.max.gov/browse/API-4750

- This is a new V2 endpoint that will not interfere with existing validation endpoint calls
- Testing includes unit testing and local testing with standard auth code, client credentials, and SSOi based access tokens


modules/openid_auth/README_V2.yml
```diff
@@ -0,0 +1,147 @@
...
+    ### Token Validation Request
+
+    Allows a third-party application to request the validated token and VA Identifier ICN for an individual:
+
+    1. Client Request: POST https://api.va.gov/internal/auth/v2/validation
+       * Provide the Bearer token as a header: `Authorization: Bearer <token>`
+       * Provide a body parameter for appropriate audience for validation
+         aud: https://api.va.gov/services/fhir
+
+    2. Status Response: A JSON API object with the individual's validated token and VA identifier ICN
```
beaugrantham-va, line 29: V2 returns more than just ICN... worth mentioning that here?

dbrown-va: Good catch.  I had updated the schema but not this blurb.

Due to lines of code restraints, I've broken out this documentation route to a separate PR.
https://github.com/department-of-veterans-affairs/vets-api/pull/6363

app/controllers/openid_application_controller.rb
```diff
@@ -35,8 +36,11 @@ def authenticate_token
     return false if token.blank?
 
     # issued for a client vs a user
-    if token.client_credentials_token?
+    if token.client_credentials_token? || ssoi_token?
       token.payload[:icn] = fetch_smart_launch_context if token.payload['scp'].include?('launch/patient')
+      if token.payload['scp'].include?('launch')
+        token.payload[:launch] = JSON.parse(Base64.decode64(fetch_smart_launch_context))
```
beaugrantham-va, line 42: Because we don't _enforce_ that launch must be b64, what is the behavior here if it isn't?

app/controllers/openid_application_controller.rb
```diff
@@ -35,8 +36,11 @@ def authenticate_token
     return false if token.blank?
 
     # issued for a client vs a user
-    if token.client_credentials_token?
+    if token.client_credentials_token? || ssoi_token?
```
beaugrantham-va, line 39: Would it be worth breaking these two conditionals apart for clarity?

The `launch/patient` logic conditionally applies only for client credentials.
The `launch` logic conditionally applies only for internal users.

app/controllers/openid_application_controller.rb
dbrown-va: Good catch.  The JSON parse would throw an exception.

I have added a check to first check if valid JSON, otherwise it is assumed to be a plaintext ICN.

beaugrantham-va (approved):

PR #6364, opened 2021-03-30
Added Redis call to Notification Service

bradenshipley:
## Description of change

This PR includes multiple formatting changes to the notification_service and it's test file, as well as changing how duplicate emails are caught before being sent. Since okta seems to create multiple redundant events (each with a unique UUID), it's necessary to check for equality of the email of the user who created the event, and the published time of the event.

## Original issue(s)

Tested locally and unit tests have been updated. This feature is still testing in staging, and won't affect production. A PR will be made to remove the `staging_flag` once all testing is complete


modules/apps_api/lib/apps_api/notification_service.rb
```diff
@@ -53,34 +51,46 @@ def parse_event(event)
...
+      # get all members of the notification_events set
+      members = Redis.current.smembers('apps_notification_events')
+      return false if members.nil?
+
+      # check if there is a member with the same email
+      # and time as our parsed_hash to detect duplicates
+      current_event = { 'email' => parsed_hash[:user_email], 'time' => parsed_hash[:options][:time] }
+      members.each do |member|
+        member_hash = Redis.current.hgetall(member)
+        already_handled = true if current_event.eql? member_hash
```
SlakrHakr, line 91: can this just return true here? instead of having the `already_handled` variable
i dont see a reason why the rest of the list of members needs to be iterated over if it finds an event that matches the event youre checking

modules/apps_api/lib/apps_api/notification_service.rb
```diff
@@ -89,16 +99,26 @@ def event_unsuccessful?(event)
          event['target'][0]['detailEntry']['subject'].nil?)
```
SlakrHakr, line 99: worth putting some nil checking in here?

modules/apps_api/spec/lib/notification_service_spec.rb
```diff
@@ -148,7 +148,7 @@
     # lib/va_notify/configuration.rb#initialize contains:
     # @notify_client ||= Notifications::Client.new(api_key, client_url)
     allow(Notifications::Client).to receive(:new).and_return(notification_client)
-    allow(notification_client).to receive(:send_email)
+    allow(notification_client).to receive(:send_email).and_return(true)
```
SlakrHakr, line 151: any reason not to use `expect` here instead of `allow`? just to make sure the calls to those methods actually happen

SlakrHakr (approved):

modules/apps_api/lib/apps_api/notification_service.rb
bradenshipley: can you rephrase this? I'm not quite sure what you're asking. The target will always have at least one item because an event is only created (of the event types we are looking for) when there is both a `subject` and a `target` in the event data. There may be more than one `target`, but never less. This nil check is to handle the below edge case from talking with okta.

> Not all disconnect events will have a target[0]['detailEntry']['subject'] present in the hash. This seems to happen when the user being disconnected has already been deleted from Okta. It's an edge case but it is safe to assume any events without this key can be ignored.

modules/apps_api/lib/apps_api/notification_service.rb
bradenshipley: This is a good catch, thanks. I rewrote this a couple times and that must have gotten missed.

modules/apps_api/lib/apps_api/notification_service.rb
SlakrHakr: ok, sounds good, i thought there may have been a strange edge case where that list of targets could be empty.
sounds like there isn't, fair enough! 👍🏻

PR #6370, opened 2021-03-31
Suppresses zip code validation for non-US addresses

patrickvinograd:
## Description of change
Updates covid vaccine expanded submission endpoint to only enforce zip/zip+4 format for USA addresses.

## Original issue(s)


saneshark (approved):
LGTM; is FE doing something similar?

batemapf (approved):
Should be a pretty edge-y edge case once we pull down the country option (happening today). The FE should already be validating ZIP and ZIP+4 formats. Of course someone could post directly to the API... maybe I'm missing something?

PR #6374, opened 2021-03-31
Remove user mpi

bramleyjl:
## Description of change

This is an updated pull request from #6160, moving direct-to-MPI calls to within the User model getter methods. This change is part of the larger change to consolidate User & Identity functionality in preparation for the planned Identity Module / Component.

## Original issue(s)


bosawt (approved):
Looks good!

modules/vaos/app/models/vaos/appointment_request_form.rb
```diff
@@ -82,17 +82,17 @@ def facility=(values_hash)
     # These values ought to be derived from MVI vs user input, except for inpatient and text_messaging_allowed
     def patient=(values_hash)
       @patient = {
-        display_name: "#{last_name}, #{first_name}",
-        first_name: first_name,
-        last_name: last_name,
+        display_name: "#{@user.last_name_mpi}, #{@user.first_name_mpi}",
```
ericboehs, line 85:
```suggestion
        display_name: [@user.last_name_mpi, @user.first_name_mpi].join(', '),
```

Since last name is nil checked this could be empty here resulting in `", John"`. Would the suggest code change work instead?

ericboehs (approved):
Small suggestion.

modules/vaos/app/models/vaos/appointment_request_form.rb
bosawt: Should mention that you'll need to call `compact` off the array to do what you're suggesting here

LanceGundersen (approved):
👍

PR #6390, opened 2021-03-31
Sftp client covid vaccine

saneshark:
## Description of change
SFTP client for Enrollment Service

## Original issue(s)


modules/covid_vaccine/app/services/covid_vaccine/v0/enrollment_handler.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module CovidVaccine
```
LindseySaari, line 3: Are there specs for this file?

patrickvinograd:
Kam it's not needed immediately, but when we get the the point of having a scheduled job for this, I added a setting for `job_enabled` under enrollment_service stanza, so that we can avoid triggering the job in dev/staging where there is no sftp connection. 
https://github.com/department-of-veterans-affairs/devops/pull/8832

PR #6392, opened 2021-04-01
Eligibility check phase for covid vaccine expanded registration

patrickvinograd:
## Description of change
* Adds state machine. Note initial state was changed to 'received' because there is a 'sequestered' column in the database and I thought that was potentially confusing.
* Adds job to perform eligibility calculation, record eligibility_info and update state of object.
* Adds rake task to print a summary of submission states.
* Adds rake task to trigger eligibility job for all submissions with state nil or received.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/22490

* Tested locally with eligible/ineligible submisions.
* This PR surfaced the need for an index to be added on the state column. That will be done in a separate migration PR.


LindseySaari (approved):

patrickvinograd:
Rebased master to get an unrelated test fix in

PR #6396, opened 2021-04-01
Mobile update json serialization gem

kreek:
## Description of change
Updates mobile non secure messaging json serialization to use the [jsonapi-serializer](https://github.com/jsonapi-serializer/jsonapi-serializer) gem which is the forked and maintained version of [fast_jsonapi](https://github.com/Netflix/fast_jsonapi). This also sets us up for the appointments and claims pagination as `jsonapi-serializer` has built in pagination tools that conform to the JSON API Spec way of doing pagination.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#22516


kreek:
closing as it seems like you can't have `fast_jsonapi` serializers and `jsonapi-serializer` in the same app (even if different engines) as the later causes serializers using the former to change their behavior WRT includes

PR #6402, opened 2021-04-01
Csv generator and sftp

saneshark:
## Description of change
Updates the CSV Generator and SFTP

## Original issue(s)


modules/covid_vaccine/app/services/covid_vaccine/v0/enrollment_service.rb
```diff
@@ -0,0 +1,47 @@
...
+        end
+      end
+
+      private
+
+      def remote_file_path(name)
+        "/#{name}"
+      end
+
+      def generated_file_name
```
patrickvinograd, line 30: Can you make it so DHS_load is a default param and if pass in something else like TEST_ it doesn't even include 'DHS_load', that way if their pattern matching is not great, a TEST file won't have that string anywhere in it.

patrickvinograd: Maybe that's over-thinking it

patrickvinograd (approved):
LGTM

modules/covid_vaccine/app/services/covid_vaccine/v0/enrollment_service.rb
saneshark: I think he said if it's prefixed with anything other than DHS_load it wont pick up automatically

patrickvinograd: nevermind, but we can make sure whatever rake tasks we put in place are safe w/r/t filenames

PR #6411, opened 2021-04-01
Pv vax expansion facility name mapping

patrickvinograd:
## Description of change
Works around a frontend bug where frontend is sending facility name instead of ID. Since we have a bunch of batched submissions awaiting processing, we need to solve this before we submit to enrollment system. 

For the most part this was fixed by adding a static reverse mapping from name to ID. The complexity was compounded because there are two pairs of VAMCs with the same name in different states. 

## Original issue(s)


ericboehs (approved):

saneshark (approved):
LGTM

PR #6413, opened 2021-04-01
Add error handling for VSOReloader, run cron job nightly

hinzed1127:
## Description of change

This does a few things:

- Adds a cron job to run every night and refreshing the VSO list
- Notifies slack and sentry of failures

## Original issue(s)
https://vajira.max.gov/browse/API-5868


modules/veteran/app/workers/veteran/vso_reloader.rb
```diff
@@ -6,8 +6,15 @@ module Veteran
   class VSOReloader < BaseReloader
     def perform
       array_of_organizations = reload_representatives
+
       # This Where Not statement is for removing anyone no longer on the lists pulled down from OGC
       Veteran::Service::Representative.where.not(representative_id: array_of_organizations).destroy_all
```
hinzed1127, line 11: @SlakrHakr I think this covers the part you were asking about where whether it wipes the slate clean or just removes outdated ones (seems like the latter)

SlakrHakr: perfect! i was a bit concerned if it just wiped everything and started clean since this could lead to a situation where the data is empty for a short period of time (or for a longer period of time if the data is wiped and then there was a failure in repopulating it). this is great!

SlakrHakr (approved):

SlakrHakr:
Looks like the new tests are failing in CI

PR #6422, opened 2021-04-02
Api 6002 - Refactor submissions#create to return submission data and signed url

ElleeB:
## Description of change
Instead of taking in actual uploaded documents, we will provide the consumer with an s3 url, after creating an evidence submission instance and, when possible, linking it to an existing Appeal.

## Original issue(s)
https://vajira.max.gov/browse/API-6002

Tests updated in controller spec

### DIFF
--- a/modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
+++ b/modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
@@ -5,6 +5,24 @@ module AppealsApi::V1
     module NoticeOfDisagreements
       class EvidenceSubmissionsController < AppealsApi::ApplicationController
         skip_before_action :authenticate
+        before_action :set_submission_attributes, if: -> { params[:nod_id] }
+
+        def create
+          submission = AppealsApi::EvidenceSubmission.create(@submission_attributes)
+          signed_url = submission.get_location
+
+          render json: { data:
+                          {
+                            attributes:
+                              {
+                                status: submission.status,
+                                id: submission.id,
+                                appeal_id: submission.supportable_id,
+                                appeal_type: submission.supportable_type
+                              },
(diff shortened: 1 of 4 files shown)


ElleeB:
I made updates to the model, so will also add a few tests to the model spec for the new methods.

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
```diff
@@ -17,14 +25,16 @@ def show
           render json: serialized.serializable_hash
         end
 
-        def create
-          status, error = AppealsApi::FileValidator.new(params[:document]).call
+        private
+
+        def set_submission_attributes
+          @appeal = AppealsApi::NoticeOfDisagreement.find_by(id: params[:nod_id])
```
cilestin, line 31: Let's memoize both `@appeal `and `@submission_attributes`. We shouldn't need to regenerate them after the first call.

modules/appeals_api/app/services/appeals_api/supporting_evidence/evidence_uploader.rb
```diff
@@ -1,47 +0,0 @@
-# frozen_string_literal: true
```
cilestin, line 1: I think we can keep these files around and let the `upload_complete` endpoint PR decide if it'll use them or not.

modules/appeals_api/app/services/appeals_api/supporting_evidence/temporary_storage_uploader.rb
```diff
@@ -1,44 +0,0 @@
-# frozen_string_literal: true
```
cilestin, line 1: Same here. Let's keep this and let `upload_complete` decide if it want so use them or not.

cilestin (changes requested):

cilestin (approved):

cilestin (approved):

PR #6429, opened 2021-04-05
bump oj to latest minor version

LindseySaari:
## Description of change
As part of the gem update series, this PR updates the oj gem(default group) to the latest minor version. Gems are being updated individually (separate PRs), in order to avoid issues if a roll back is needed.

A change was made in `3.9.1`  to raise an exception on an invalid time represented as a decimal in :object mode. See change to users_serialized.json (I'd like to verify this change with the Identity team)

https://github.com/ohler55/oj/blob/develop/CHANGELOG.md

## Original issue(s)
department-of-veterans-affairs/va.gov-team##22265

## Testing Completed
- [x] CI make target passing locally 
- [x] Reviewed gem change log updates


ericboehs:
@LindseySaari Do you still need to verify this with the identity team?

bosawt (approved):
Looks good! I've never seen time represented like that, so I tested it locally just to prove that we save the Time object as expected. These fields are from a serialized User object that's stored in redis, so:
`@redis.set "barf", User.new(last_signed_in: Time.now).to_json`
`@redis.get 'barf'`
`{\"uuid\":null,\"last_signed_in\":\"2021-04-06T19:39:59.993Z\",\"mhv_last_signed_in\":null}`
`last_signed_in` is stored and serialized as a proper Time object. I can't think of anywhere else we would force a decimal Time representation. So we should be good

LindseySaari:
> Looks good! I've never seen time represented like that, so I tested it locally just to prove that we save the Time object as expected. These fields are from a serialized User object that's stored in redis, so:
> `@redis.set "barf", User.new(last_signed_in: Time.now).to_json`
> `@redis.get 'barf'`
> Result: `{\"uuid\":null,\"last_signed_in\":\"2021-04-06T19:39:59.993Z\",\"mhv_last_signed_in\":null}`
> `last_signed_in` is stored and serialized as a proper Time object. I can't think of anywhere else we would force a decimal Time representation. So we should be good

Thanks so much for the review! @ericboehs This should be good to go 😸

ericboehs (approved):

PR #6433, opened 2021-04-05
Remove external 'User.va_profile' references

bramleyjl:
## Description of change

This change removes `va_profile` references from outside the User class as part of a larger refactor and move to keep specific User trait calls (Identity vs. MPI) contained within the User class to reduce complexity across vets-api. This change is intended only to simplify existing functionality and should not cause any noticeable changes.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#18740

New unit tests have been added for explicit MPI getter & setter methods.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -62,53 +62,10 @@ class User < Common::RedisStore
     pciu&.get_alternate_phone&.to_s
   end
 
-  def home_phone
-    mpi_profile&.home_phone
-  end
+  # Identity getter methods
 
-  def first_name
-    identity.first_name || (mhv_icn.present? ? first_name_mpi : nil)
-  end
-
-  def first_name_mpi
-    given_names&.first
-  end
-
-  def given_names
-    mpi_profile&.given_names
-  end
-
-  def full_name_normalized
-    {
-      first: first_name&.capitalize,
-      middle: middle_name&.capitalize,
-      last: last_name&.capitalize,
-      suffix: va_profile&.normalized_suffix
-    }
-  end
-
-  def ssn_normalized
-    ssn&.gsub(/[^\d]/, '')
-  end
-
-  def middle_name
-    identity.middle_name || (mhv_icn.present? ? mpi&.profile&.given_names.to_a[1..-1]&.join(' ').presence : nil)
-  end
-
-  def last_name
-    identity.last_name || (mhv_icn.present? ? mpi&.profile&.family_name : nil)
-  end
-
-  def last_name_mpi
-    mpi_profile&.family_name
-  end
-
-  def gender
-    identity.gender || (mhv_icn.present? ? mpi&.profile&.gender : nil)
-  end
-
-  def gender_mpi
-    mpi_profile&.gender
+  def birls_id
+    identity&.birls_id || mpi&.birls_id
   end
 
   # Returns a Date string in iso8601 format, eg. '{year}-{month}-{day}'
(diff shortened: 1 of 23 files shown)


modules/openid_auth/app/controllers/openid_auth/application_controller.rb
```diff
@@ -7,9 +7,9 @@ class ApplicationController < ::OpenidApplicationController
     skip_before_action :set_tags_and_extra_content, raise: false
 
     def validate_user
-      unless token.client_credentials_token? || token.ssoi_token?
```
bosawt, line 10: Was this `|| token.ssoi_token?` condition removed for a reason?

spec/models/user_spec.rb
```diff
@@ -366,6 +389,10 @@
           stub_mpi(mvi_profile)
         end
 
+        it 'fetches given_names from MPI' do
+          expect(user.given_names).to be(user.mpi.profile.given_names)
```
bosawt, line 393: If we're eventually going to make `mpi` a private function on `User`, you'll probably want to remove `user.mpi` calls in specs as well. Since you have the `mvi_profile` used to build `User` though, you should be able to get this by `mvi_profile.given_names`, etc (as you did above for the `vet360_id` spec)

spec/models/form_profile_spec.rb
```diff
@@ -9,7 +9,7 @@
   let(:user) { build(:user, :loa3) }
 
   before do
-    user.va_profile.suffix = 'Jr.'
+    user.mpi.profile.suffix = 'Jr.'
```
bosawt, line 12: If we could prevent this case of referring to `user.mpi`, that'd help us down the line, you can do that by adding:
`  let(:mvi_profile) { build(:mvi_profile, suffix: 'Jr.') }`
    
`before {
      stub_mpi(mvi_profile)
    }`

spec/request/letters_request_spec.rb
```diff
@@ -127,8 +127,8 @@
       end
 
       before do
-        user.va_profile.edipi = '1005079999'
-        user.va_profile.participant_id = '600039999'
+        user.mpi.profile.edipi = '1005079999'
```
bosawt, line 130: Should be able to prevent this `user.mpi` call similar to the recommendation above:
`let(:mvi_profile) { build(:mvi_profile, edipi: 'blahblah', participant_id: 'blahblah') }`

`before { stub_mpi(mvi_profile) }`

app/models/user.rb
```diff
@@ -174,6 +200,10 @@ def loa
   delegate :icn_with_aaid, to: :mpi
   delegate :vet360_id, to: :mpi
   delegate :search_token, to: :mpi
+  delegate :status, to: :mpi, prefix: true
```
bosawt, line 203: It's interesting how these delegate functions work with `prefix: true`. I didn't realize that it gives us functions like `mpi_status` to differentiate. It might be more concise to delegate all the other `*_mpi` functions, but I'll leave that up to you, not really a big deal either way

bosawt (commented):
Nice work! I think there was only one comment about non-test code, then a couple suggestions for the specs, but it's looking good!

I also tested:
- Auth flow
- User endpoint successfully returns
- Clicking on `My VA` failed, but I looked into it and I can't see how this PR would affect that, it's probably unrelated.

spec/models/form_profile_spec.rb
bramleyjl: I tried replacing this case but it broken a few dozen other tests in this file with a variety of different user profile attributes being stubbed incorrectly, my guess is due to the unique nature of it testing forms?

app/models/user.rb
bramleyjl: I can probably do that in the second PR for this story that adds some order to the User methods; I already had to cut it out to get under the 500 lines changed metric.

ericboehs (approved):

bosawt (dismissed):

bosawt (approved):
Looks good!

LindseySaari (dismissed):

LindseySaari (approved):

LindseySaari (approved):

bosawt (approved):

PR #6439, opened 2021-04-05
Maps birth sex values as expected by enrollment service upstream

patrickvinograd:
## Description of change
Updates CSV generation for to map values to "M","F",or empty string as expected by enrollment service. 

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/22643

### DIFF
--- a/modules/covid_vaccine/app/services/covid_vaccine/v0/expanded_registration_csv_generator.rb
+++ b/modules/covid_vaccine/app/services/covid_vaccine/v0/expanded_registration_csv_generator.rb
@@ -17,13 +17,19 @@ module CovidVaccine
         r.raw_form_data['preferred_facility']&.delete_prefix('vha_')
       end
 
+      birth_sex_proc = proc do |r|
+        next r.raw_form_data['birth_sex'][0] if %w[Male Female].include? r.raw_form_data['birth_sex']
+
+        nil
+      end
+
       MAPPER = {
         first_name: proc { |r| r.raw_form_data['first_name'] },
         middle_name: proc { |r| r.raw_form_data['middle_name'] },
         last_name: proc { |r| r.raw_form_data['last_name'] },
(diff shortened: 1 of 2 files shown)


batemapf (approved):
LGTM from a product standpoint. Really know nil about `nil`, but going to assume it is rendered as nothing in the caret delimited file :-)

patrickvinograd:
> LGTM from a product standpoint. Really know nil about `nil`, but going to assume it is rendered as nothing in the caret delimited file :-)

Yep, the spec verifies that nil gets mapped as empty e.g. `^123456789^^` (using the known ssn value as an anchor for the following empty field).

modules/covid_vaccine/app/services/covid_vaccine/v0/expanded_registration_csv_generator.rb
```diff
@@ -17,13 +17,19 @@ class ExpandedRegistrationCsvGenerator
         r.raw_form_data['preferred_facility']&.delete_prefix('vha_')
       end
 
+      birth_sex_proc = proc do |r|
+        next r.raw_form_data['birth_sex'][0] if %w[Male Female].include? r.raw_form_data['birth_sex']
```
saneshark, line 21: how about 

How about:

```suggestion
        { 'Male' => 'M', 'Female' => 'F' }.fetch(r.raw_form_data['birth_sex'][0].upcase, nil)
```

or 

```suggestion
        %w[M F]).find(r.raw_form_data['birth_sex'][0].upcase)
```

saneshark: might not even need the `upcase` if it's always one of those three values

saneshark (approved):
Nitpick comments to simplify / make more readable, but logic and specs are sound as is.

PR #6441, opened 2021-04-05
RuboCop ABC Size - Move to todo.yml

ericboehs:
Our .rubocop.yml set our ABC max limit very high (44). This moves it to the todo yml.


f1337 (approved):
Excellent! :metal:
Bill & Ted playing air guitar, saying 'Excellent!'

PR #6450, opened 2021-04-06
LOAST fixes

mviterna:
## Description of change

Fixes to make the Forms API LOAST complaint
 
## Original issue(s)
https://vajira.max.gov/browse/API-5982

Locally tested against LOAST

### DIFF
--- a/modules/va_forms/app/swagger/va_forms/forms/form_swagger.rb
+++ b/modules/va_forms/app/swagger/va_forms/forms/form_swagger.rb
@@ -41,8 +41,8 @@ module VAForms
               key :example, '2016-07-10'
               key :format, 'date'
             end
-            property :last_revised_on do
-              key :description, I18n.t('va_forms.field_descriptions.last_revised_on')
+            property :last_revision_on do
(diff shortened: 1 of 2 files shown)


gpbowman-git (approved):

modules/va_forms/app/swagger/va_forms/forms/form_swagger.rb
```diff
@@ -70,42 +74,49 @@ class FormSwagger
...
+              key :nullable, true
               key :example, 'benefit'
             end
             property :language do
               key :description, I18n.t('va_forms.field_descriptions.language')
               key :type, :string
               key :example, 'en'
             end
-            property :deleted_at do
+            property :deleted_at, type: %i[string null] do
```
alexandervalencia, line 109: I believe the type here should only be `string` which is defined on L111, so I think we can ditch the property setting here.

See: https://swagger.io/specification/#schema-object
> type - Value MUST be a string. Multiple types via an array are not supported.

```suggestion
            property :deleted_at do
```

PR #6470, opened 2021-04-06
Changed event type and updated service logic

bradenshipley:

## Description of change

This PR fixes a typo that checked the Okta System Logs for the `app.oauth2.as.token.revoke` event instead of `app.oauth2.as.consent.revoke`. This is the event created when a user disconnects an application on their Va.gov profile's `connected applications` section.

Because the event schema for `app.oauth2.as.consent.revoke` is slightly different than the service logic was setup to handle in the previous code, the `#parse_event` method has been changed to be uniform between both connections and disconnections. The tests have been changed to reflect this schema change.

The `#format_published_time` method has been changed to remove a trailing `.`.

## Original issue(s)


charleystran (approved):
:shipit:

PR #6485, opened 2021-04-07
Parallel specs GitHub actions

rileyanderson:
## Description of change
Add Parallel specs to GitHub actions


.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,65 @@
+name: Build and Test
+on: [push]
+jobs:
+  tests:
+    runs-on: ubuntu-latest
```
ericboehs, line 5: In order to use our beefy runner, you'll need to do this.
```suggestion
    runs-on: [self-hosted, linux]
```

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,65 @@
...
+
+      - name: Run Specs
+        env:
+          POSTGRES_USER: vets-api
+          POSTGRES_DB: vets-api_test
+          POSTGRES_PASSWORD: postgres
+          RAILS_ENV: test
+          POSTGRES_HOST: localhost
+          POSTGRES_PORT: ${{ job.services.postgres.ports[5432] }}
+          DATABASE_URL: "postgis://vets-api:postgres@127.0.0.1:5432/vets_api_test?pool=4"
```
ericboehs, line 64:
```suggestion
          DATABASE_URL: "postgis://vets-api:postgres@127.0.0.1:5432/vets_api_test?pool=4"
          TERM: xterm-256color
```

I believe if you add a TERM here it will colorize the output in the workflow.

.github/workflows/build_and_test.yml
ericboehs: Demian says he rebuilt the worker off of ubuntu20 so we might be able to use it again. Let's keep it out of this PR for now though.

.github/workflows/build_and_test.yml
rileyanderson: I'm not sure TERM actually does anything in this case.

ericboehs:
We should rename the workflows to meld together better in the PR review checks.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,70 @@
...
+jobs:
+  tests:
+    runs-on: ubuntu-20.04
+    env:
+      BUNDLE_ENTERPRISE__CONTRIBSYS__COM: ${{ secrets.BUNDLE_ENTERPRISE__CONTRIBSYS__COM }}
+      RAILS_ENV: test
+      "Settings.test_database_url": "postgis://vets-api:postgres@127.0.0.1:5432/vets_api_test"
+      "Settings.saml.cert_path": "spec/support/certificates/ruby-saml.crt"
+      "Settings.saml.key_path": "spec/support/certificates/ruby-saml.key"
+      "Settings.binaries.clamdscan": "clamscan"
```
ericboehs, line 12: Can we move these to the settings.yml for test?

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,49 @@
...
+        env:
+          POSTGRES_USER: vets-api
+          POSTGRES_DB: vets-api_test
+          POSTGRES_PASSWORD: postgres
+        ports:
+          - 5432:5432
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 15
+
+      redis:
+        image: circleci/redis:5.0-alpine
```
ericboehs, line 30: Does redis need ports opened? If not, can the ports be removed from postgres?
```suggestion
        image: circleci/redis:5.0-alpine
        ports: 6379:6379
```

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,74 @@
...
+      DISABLE_BOOTSNAP: true
+      RAILS_ENV: test
+      "Settings.test_database_url": "postgis://vets-api:postgres@127.0.0.1:5432/vets_api_test"
+      "Settings.saml.cert_path": "spec/support/certificates/ruby-saml.crt"
+      "Settings.saml.key_path": "spec/support/certificates/ruby-saml.key"
+      "Settings.binaries.clamdscan": "clamscan"
+      TERM: xterm-256color
+      GIT_COMMIT_SHA: ${{ github.sha }}
+      CC_TEST_REPORTER_ID: ${{secrets.CC_TEST_REPORTER_ID}}
+      ImageOS: ubuntu20
```
ericboehs, line 21: How does this work with a Debian image as specified on line 7?

rileyanderson: yeah it doesn’t 😄. need to change it

ericboehs (approved):

PR #6488, opened 2021-04-07
Api 6165 - Refactor to use serializer on submission create

ElleeB:
Refactored the serializer for our evidence submissions to include s3 location url, and additional custom methods to manage the response keys for supportable_type and supportable_id to read `appeal` instead of `supportable`. The change impacted both the rendering in the create and show endpoints.

Controller test updated
Serializer test updated

Ticket: https://vajira.max.gov/browse/API-6156

### DIFF
--- a/modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
+++ b/modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
@@ -5,23 +5,15 @@ module AppealsApi::V1
     module NoticeOfDisagreements
       class EvidenceSubmissionsController < AppealsApi::ApplicationController
         skip_before_action :authenticate
-        before_action :set_submission_attributes, if: -> { params[:nod_id] }
+        before_action :set_submission_attributes, only: :create
 
         def create
           submission = AppealsApi::EvidenceSubmission.create(@submission_attributes)
-          signed_url = submission.get_location
 
-          render json: { data:
-                          {
-                            attributes:
-                              {
-                                status: submission.status,
-                                id: submission.id,
-                                appeal_id: submission.supportable_id,
-                                appeal_type: submission.supportable_type
-                              },
-                            location: signed_url
-                          } }
+          render json: submission,
+                 serializer: AppealsApi::EvidenceSubmissionSerializer,
+                 key_transform: :camel_lower,
+                 render_location: true
         end
 
         def show
@@ -30,9 +22,10 @@ module AppealsApi::V1
             supportable_type: 'AppealsApi::NoticeOfDisagreement'
           )
 
-          serialized = AppealsApi::EvidenceSubmissionSerializer.new(submissions)
-
-          render json: serialized.serializable_hash
+          render json: submissions,
+                 each_serializer: AppealsApi::EvidenceSubmissionSerializer,
+                 key_transform: :camel_lower,
+                 render_location: false
         end
 
         private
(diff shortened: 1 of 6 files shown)


modules/appeals_api/app/serializers/appeals_api/evidence_submission_serializer.rb
```diff
@@ -1,13 +1,30 @@
 # frozen_string_literal: true
 
+require_dependency 'common/exceptions'
 require 'fast_jsonapi'
 
 module AppealsApi
-  class EvidenceSubmissionSerializer
-    include FastJsonapi::ObjectSerializer
+  class EvidenceSubmissionSerializer < ActiveModel::Serializer
```
yakschuss, line 7: Can you help me understand the goal behind this change?

I worry about moving away from a non-standard implementation, considering the cost of changing in the future.

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
```diff
@@ -42,8 +36,8 @@ def set_submission_attributes
           raise Common::Exceptions::RecordNotFound, params[:nod_id] unless @appeal
```
yakschuss, line 36: wouldn't params[:nod_id] would need to be changed here as well?

modules/appeals_api/app/serializers/appeals_api/evidence_submission_serializer.rb
ElleeB: How is this a non-standard implementation?

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
ElleeB: Why?

modules/appeals_api/app/serializers/appeals_api/evidence_submission_serializer.rb
```diff
@@ -1,13 +1,30 @@
...
+    type 'evidence_submission'
 
-    set_key_transform :camel_lower
-    attributes :status
-    set_type :evidenceSubmission
+    attributes :status, :code, :details, :appeal_type, :appeal_id, :location
+
+    delegate :code, :details, to: :object
```
cilestin, line 12: Delegating `details` means it could return a large amount of text from an error backtrace or something. Let's make sure it only returns the first 100 characters of the details by having the serializer truncate the output.

Also it should be `detail`, not `details`, to follow our other models - but that's a fix for another ticket :P

ElleeB: Yeah, the migration set `details`, so a ticket is in order

modules/appeals_api/app/serializers/appeals_api/evidence_submission_serializer.rb
yakschuss: The rest of vets-api uses FastjsonApi. If we diverge from that, we're supporting a unique implementation of serialization. When changes to serialization happen, we'll need to update our serialization individually

ElleeB: image
this is just a sample of the serializers in vets-api using AM Serializer as opposed to Fastjson

cilestin:
I've rolled most of this into #6566, so we can probably close this one now.

PR #6493, opened 2021-04-07
22461 refactor mobile caching to enable filtering/pagination

kreek:
## Description of change
Previously we were caching mobile appointments and claims as http response bodies (with data and meta wrappers). Upcoming pagination work requires that we can cache and read a year of appointments but only return a subset based on client request params. This will allow us to filter/paginate on a set rather than parsing JSON first.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#22461

* If the app were released we'd have cache collisions in prod. Other than UAT we don't have prod users. I'll check that no testing sessions are planned (either way the cache will be clear once a user's session expires)


jperk51 (approved):
LGTM

PR #6494, opened 2021-04-07
Scheduled workers for vaccine enrollment submission

patrickvinograd:
## Description of change
Adds scheduled workers to perform the batch-and-upload-to-ES process for covid vaccine records. 

Still TBD:
- [x] Test coverage
- [ ] Should we suppress the upload job in staging/dev since there's no SFTP connection there? I added a `job_enabled` setting in an earlier PR with this in mind. But initially I'd like to see it getting triggered in staging to verify. 

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/22673


config/sidekiq_scheduler.yml
```diff
@@ -214,3 +214,8 @@ Veteran::VSOReloader:
   cron: "0 2 * * * America/New_York"
   class: Veteran::VSOReloader
   description: Pulls down current list of VSOs and updates accordingly 
+
+CovidVaccine::ScheduledBatchJob:
+  cron: "0 6-19 * * MON-FRI America/New_York"
```
batemapf, line 219: Inferring from a previous convo that if we want to change this value to hourly or some other cadence, we can do so without a full vets-api deploy (i.e. just re-deploying the last release)?

patrickvinograd: No...because this file, while technically a config file, isn't overriden in the devops repo the way the main settings.yml file is.

patrickvinograd: Caveat: I have been known to mess up cron syntax before 🙈

batemapf: Touche. Correct phrase would have been "every hour, regardless of hour of day or day of week" maybe.

batemapf: https://crontab.guru/#0_6-19_*_*_MON-FRI seems to like it, fwiw.

batemapf:
re: PV's comment in the PR description:

>Should we suppress the upload job in staging/dev since there's no SFTP connection there? I added a job_enabled setting in an earlier PR with this in mind. But initially I'd like to see it getting triggered in staging to verify.

Thinking we need to update [this](https://github.com/department-of-veterans-affairs/devops/blob/1ac47e82d632a66a90e9488f910091e6cb5c4266/ansible/deployment/config/vets-api/staging-settings.local.yml.j2#L580) in devops config, and condition the sidekiq job on it somehow? I tend to agree that we can rest a little easier seeing the job run at least until it fails on missing SFTP info, even just for a couple of hours, and then change devops setting back.

batemapf (approved):
Approved, with caveat that we're still discussing the enabling of jobs in lower envs for testing

ericboehs (approved):

PR #6498, opened 2021-04-07
Removing staging flag from notification service

bradenshipley:
## Description of change

This PR removes the staging flag from the notification service, that prevented this code from running in any environment other than staging. The functionality this flag was wrapping has been tested inside the staging environment and is now ready to be used in all environments.
## Original issue(s)

The only settings file changes are the removal of the `staging_flag` from `settings.yml` and `test.yml`


yakschuss (approved):
🚀  - does this also need a PR to `/devops`

bradenshipley:
> 🚀 - does this also need a PR to `/devops`

a subsequent pr will be made after this one.

PR #6503, opened 2021-04-08
Mark.viterna/api 5986 500

mviterna:
## Description of change

Fixing a bug where calling FormShow will throw a 500 when a form is not found. This needs to be a 404
## Original issue(s)
https://vajira.max.gov/browse/API-5986

Tested locally


cshupp1 (approved):
Thanks for adding that unit test!

gpbowman-git (approved):

PR #6507, opened 2021-04-08
Add Act.Type to Token Validation Object V2

lukemarkwordtlibertyits:
## Description of change

Token Validation V2 Object will have information about who / what is using the token.

## Original issue(s)

https://vajira.max.gov/browse/API-6214

## Tests

**Patient Act.Type**

- [x] Non CC or SSOi tokens should have an act.type of `patient`

**System Act.Type**

- [x] CC tokens should have an act.type of `system`

**User Act.Type**

- [x] SSOi tokens should have an act.type of 'user'


.github/dependabot.yml
```diff
@@ -8,6 +8,7 @@ registries:
 updates:
```
lukemarkwordtlibertyits, line 8: this file got pulled in with the rebase

modules/mobile/app/models/mobile/v0/adapters/claims_overview.rb
```diff
@@ -12,7 +12,6 @@ def parse(list)
```
lukemarkwordtlibertyits, line 12: this file got pulled in with the rebase

modules/va_forms/app/controllers/va_forms/v0/forms_controller.rb
```diff
@@ -13,8 +13,13 @@ def index
```
lukemarkwordtlibertyits, line 13: this file got pulled in with the rebase

modules/va_forms/spec/requests/v0/forms_request_spec.rb
```diff
@@ -65,6 +65,11 @@
       expect(response).to match_response_schema('va_forms/form')
```
lukemarkwordtlibertyits, line 65: this file got pulled in with the rebase

modules/vaos/app/docs/vaos/v0/vaos.yaml
```diff
@@ -219,11 +219,139 @@ paths:
             application/json:
```
lukemarkwordtlibertyits, line 219: This file got pulled in with the rebase

modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
```diff
@@ -34,6 +35,8 @@ def validated_payload
           payload_object.launch[:patient] = @current_user.icn
         end
 
+        payload_object.act[:type] = 'system' if token.client_credentials_token?
+
```
dbrown-va, line 39: One thought to simplify things is move the `if token.client_credentials_token?` check up to line 31 and return the payload similar to the ssoi_token? block.

Then that `unless` block could be reduced to a simple `if payload_object[:icn].nil?`.

lukemarkwordtlibertyits: I think the issue here is that it is possible for a non client_credentials token to have null ICN. In that case we would be setting the wrong act.type. Or we would put that check about the null icn check and we would have dubplicated code between the client_credentials token statement and the null icn statement.

dbrown-va (approved):
Confirmed the new `client_credentials_token?` block.

esubocz-va (approved):
Confirmed all token use cases.

PR #6508, opened 2021-04-08
Remove rubocop:disable Metrics/AbcSize in mobile claims adapter

kreek:
## Description of change
Seems like briefly yesterday afternoon that cop was needed but now it's not (and is blocking other PRs).

## Original issue(s)
N/A


kreek:
dup

PR #6510, opened 2021-04-08
set insecure-external-code-execution config

LindseySaari:
Setting config for dependabot Gemfile code execution 

"Dependabot refused to execute external code. 
Dependabot blocked external code required for this update. Please set insecure-external-code-execution: allow in the config if you trust all dependencies' supply chain."


ericboehs (approved):
Approving with squinty suspicious eyes.

Let's just test this out. I don't think we should keep this long term.

PR #6527, opened 2021-04-09
Api 6001/add status audit log

yakschuss:
[API-6001](https://vajira.max.gov/browse/API-6001)

This PR adds the necessary tables to produce an internal audit log of status updates created within the AppealApi system. 

It also adds an EventSubscription table to handle the event-sourcing communication of the status updates.

There's a possibility that although status updates are contained to the appeals system, it could grow large enough over time that it should be reduced in size. For now, there's no immediate reason to push it elsewhere.


db/schema.rb
```diff
@@ -104,6 +112,17 @@
     t.string "source"
   end
 
+  create_table "appeals_api_status_updates", force: :cascade do |t|
+    t.string "from"
+    t.string "to"
+    t.string "appeal_type"
```
cilestin, line 118: When we still thought EvidenceSubmission would have statuses under our control, I figured this table would store those as well. What're your thoughts on making this a little more generic so that it can track any model w/ a status field we want?

yakschuss: I have no issue with that. What do you have in mind?

db/schema.rb
```diff
@@ -104,6 +112,17 @@
...
   end
 
+  create_table "appeals_api_status_updates", force: :cascade do |t|
+    t.string "from"
+    t.string "to"
+    t.string "statusable_type"
+    t.string "statusable_id"
+    t.datetime "status_update_time"
+    t.datetime "created_at", precision: 6, null: false
+    t.datetime "updated_at", precision: 6, null: false
```
cilestin, line 122: Not sure we need all these timestamps. `status_update_time` is a copy of `created_at`, right? And we shouldn't ever be updating these so `updated_at` seems redundant, but since it's part of the rails default timestamps I don't mind it staying.

What about using `created_at` instead, and if necessary aliasing `status_update_time` to `created_at`?

yakschuss: I think the idea is that the event propagation can be eventually consistent. `status_update_time` is the generated timestamp at the time of the actual update, whereas the created_at and updated_at (as you said) are the default timestamps that relate to the audit log record itself.

SlakrHakr (approved):

cilestin (approved):

PR #6543, opened 2021-04-09
API-6012 - Update hlr updater job

nathanbwright:
## Description of change
This PR changes the statuses that are being targeted for the HLR status updater job. Our previous set of statuses wasn't covering the correct ones.

## Original issue(s)
https://vajira.max.gov/browse/API-6012


cilestin (approved):
Woohoo! Thanks!

PR #6544, opened 2021-04-09
Covid Vaccination -  Add enhanced eligibility registration service

stephenBarrs:
## Description of change
This PR adds the `expanded_service_registration` class and associated tests.  The purpose of this class is to pull data from the `covid_vaccine_expanded_registration_submissions` table and send it to an external system (VeText API).  The data is being sent to the same API endpoint as the original `service_registration` class sends to.  We opted to create a second service class as the data is different and requires different transformation prior to sending to VeText.  This solution also utilizes an AASM state machine and this class moves the state: 
* `from enrollment_pending to enrollment_complete` if the user is found in the MVI lookup
* `from enrollment_complete to registered` if the send to VeText is successful
* `from enrollment_pending to enrollment_out_of_band` if there is no preferred facility that the user can be registered at

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/22674

This Service will be invoked from the `CovidVacccine::ExpandedSubmissionJob()` job.  Data already exists in the relevant table in the DB so there will need to be coordinated execution of this job in batches to process existing data.  In the future the job will be executed at a TBD cadence. 

rspec tests have been implemented to support the `expanded_registration_service` class
test have not been completed for CovidVacccine::ExpandedSubmissionJob()


modules/covid_vaccine/app/services/covid_vaccine/v0/expanded_registration_service.rb
```diff
@@ -0,0 +1,144 @@
...
+        submission.update!(vetext_sid: response[:sid], form_data: attributes, state: 'registered', eligibility_info: elig_info_icn)
+        submit_confirmation_email(attributes[:email], submission.created_at, response[:sid])
+        submission
+      end
+      
+      def submit_confirmation_email(email, date, sid)
+        return if email.blank?
+        
+        formatted_date = date.strftime('%B %-d, %Y %-l:%M %P %Z').sub(/([ap])m/, '\1.m.')
+        CovidVaccine::ExpandedRegistrationEmailJob.perform_async(email, formatted_date, sid)
```
saneshark, line 70: Is this the right mailer? I think this is the one we were using when the form is received.

modules/covid_vaccine/app/jobs/covid_vaccine/expanded_submission_job.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+require 'sentry_logging'
+
+module CovidVaccine
+  class ExpandedSubmissionJob
+    include Sidekiq::Worker
+    include SentryLogging
+
```
saneshark, line 9: I think we want to make sure that we're not retrying.

Also I'm not entirely sure how the MVI errors will be handled since it didn't look like you were raising an exception.

saneshark (commented):
Left some feedback plus one nitpick comment.

modules/covid_vaccine/app/services/covid_vaccine/v0/expanded_registration_service.rb
saneshark: maybe this code was already here for submission received, but i want to emphasize we're not notifying veterans beyond submission received which i think was handled in earlier PRs. We do not need to notify a veteran that they have "registered."

modules/covid_vaccine/app/services/covid_vaccine/v0/expanded_registration_service.rb
```diff
@@ -0,0 +1,144 @@
...
+        
+        vetext_attributes = {}
+        
+        # preferred facility will either be in eligibility_info, or raw_form_data. If its in neither one,
+        # for the purposes of this register method we should not be fetching facilities and trying to reconcile;
+        # instead we will set the state to :enrollment_out_of_band
+        
+        # Get the preferred facility here as it is needed in MPI lookup
+        # facility = submission&.eligibility_info&['preferred_facility'] || raw_form_data['preferred_facility'].delete_prefix('vha_')
+        facility = submission&.eligibility_info&.dig('preferred_facility') || raw_form_data['preferred_facility'].delete_prefix('vha_')
```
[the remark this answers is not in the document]
stephenBarrs: @saneshark I was not able to get this to work using &.fetch but it does work with .dig.  Would be very happy to see how to make this work in a better way if you want to push up a change. Thanks!

modules/covid_vaccine/app/jobs/covid_vaccine/expanded_submission_job.rb
stephenBarrs: add raising exceptions for both MVI cases and for the no preferred facility use case.   I need to figure out what the default behavior is wrt retries as I dont think we want to retry for any of these exceptions.  For the no facility found exception the state in the DB is updated to `enrollment_out_of_band` to reflect that manual intervention is required.  For the MVI exception cases we can let the job pick up the record on next execution to try again.

ericboehs (approved):

saneshark (approved):
LGTM

PR #6547, opened 2021-04-09
Add DecisionReview PersonalInformationLog REPL

edmangimelli:
I took all of my helper methods that I use for debugging Decision Reviews and put them into a file and, with `bin/rails runner rakelib/decision_review_repl.rb`, you can use a tiny REPL that has all of these methods and classes preloaded. This PR also adds a tiny PersonalInformationLog REPL which simply makes querying PersonalInformationLogs a little easier (see the spec for usage examples).


edmangimelli:
emphasis on the low-priority tag

ericboehs (approved):

PR #6440, opened 2021-04-05
[API-5901] - Provide consistent error response structures in ClaimsApi endpoints

jw81:
## Description of change
Removes multiple areas where JSON error response structures were built manually.  Replaces them with logic to raise the appropriate exception.  Utilizes the existing [ExceptionHandling](https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/controllers/concerns/exception_handling.rb) concern to handle converting raised exceptions into proper JSON responses.

## Original issue(s)
[API-5901](https://vajira.max.gov/browse/API-5901)

PR #6554, opened 2021-04-12
Update error responses to proper status in claims

hinzed1127:
## Description of change

There's a couple of endpoints that are returning either the wrong error codes (422, 404, etc) or returning a 4xx error instead of a normal 200 (eg, claims). This updates those endpoints to the proper semantic responses. I also found a few instances similar to #6440 where we could use the `Common::Exceptions` responses instead of our own on the fly

### Specific Changes

- claims with EVSS errors are now returned as `200`s with an added `evss_errors` property in the response
- `422`s in the JSON format validations now utilize the `::Common::Exceptions::UnprocessableEntity` class

## Original issue(s)
https://vajira.max.gov/browse/API-5903


modules/claims_api/spec/requests/v1/claims_request_spec.rb
```diff
@@ -166,34 +166,30 @@
 
       it 'shows a single errored Claim with an error message', run_at: 'Wed, 13 Dec 2017 03:28:23 GMT' do
         with_okta_user(scopes) do |auth_header|
-          create(:auto_established_claim,
+          create(:auto_established_claim, :status_errored_with_evss_response,
```
hinzed1127, line 169: for some reason, this test returns encounters a `404`, while the next one works just fine. The only difference I can see between the 2 is adding the EVSS error property to the mocked claim. Any ideas on how to fix this?

SlakrHakr: this is the error i see when i step through the logic that test is testing:
```
#<NameError: undefined local variable or method `claim' for #<ClaimsApi::AutoEstablishedClaimSerializer:0x000055d76a2e6958>>
```

modules/claims_api/app/serializers/claims_api/auto_established_claim_serializer.rb
```diff
@@ -4,10 +4,27 @@ module ClaimsApi
...
+
     type :claims_api_claim
 
     def id
       object&.id
     end
+
+    def evss_errors
+      if object.evss_response&.any?
+        format_evss_errors(claim.evss_response['messages'])
```
SlakrHakr, line 17: this is the line that error below is happening on it looks like
https://github.com/department-of-veterans-affairs/vets-api/pull/6554/files#r611799757

SlakrHakr: ah, ok, this makes sense, the other test that is working doesnt have any evss errors listed so it never gets into this part of the serailizer

hinzed1127: ahh yup, forgot to switch `claim` to `object` in the serializer

ericboehs (dismissed):

jw81 (approved):

modules/claims_api/app/controllers/claims_api/v0/forms/disability_compensation_controller.rb
```diff
@@ -82,6 +82,8 @@ def upload_supporting_documents
           validate_documents_page_size
 
           claim = ClaimsApi::AutoEstablishedClaim.get_by_id_or_evss_id(params[:id])
+          raise ::Common::Exceptions::ResourceNotFound.new(detail: 'Resource not found') unless claim
```
SlakrHakr, line 85: Might say "Claim not found" for a detail here

modules/claims_api/app/controllers/concerns/claims_api/json_format_validation.rb
```diff
@@ -20,13 +19,8 @@ def validate_json_format
       end
 
       def render_body_is_not_a_hash_error(body)
```
SlakrHakr, line 21: is this method used elsewhere? might be worth moving this implementation up into line 10 and getting rid of the method.
if it is reused, or having the method is just preferred, maybe we can rename it to something that doesnt start with "render"? since its not actually rendering anything any more.

SlakrHakr (changes requested):
I might be missing them but i dont see any swagger doc changes? i believe there should be a few with these changes here.

claims index endpoints can still return a 404 in certain circumstances but should be an empty list. just based on the criteria for this ticket.

again, i might have missed it but i dont see any changes to authorization logic returning a 403 instead of a 401 if the representative is correctly authenticated but doesnt have authorization to interact with the particular Veteran.

hinzed1127:
Closing this for now. This doesn't add any explicit value in the existing endpoints and is something to be more completely addressed in forthcoming claims v2 endpoints. 

I've documented some additional thoughts and a proposed solution [here](https://community.max.gov/display/VAExternal/Error+responses+for+asynchronous+claims+workflows)

PR #6556, opened 2021-04-12
Accelerate vaccine enrollment job to every 15m

batemapf:
## Description of change
To take advantage of increased functionality of the external service this job is associated with, accelerate running of this job to every 15 minutes.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/23028

### DIFF
--- a/config/sidekiq_scheduler.yml
+++ b/config/sidekiq_scheduler.yml
@@ -216,6 +216,6 @@ Veteran::VSOReloader:
   description: Pulls down current list of VSOs and updates accordingly 
 
 CovidVaccine::ScheduledBatchJob:
-  cron: "0 6-19 * * MON-FRI America/New_York"
+  every: "15m"
   class: CovidVaccine::ScheduledBatchJob
   description: Batches received vaccine records and schedules upload to enrollment service.


config/sidekiq_scheduler.yml
```diff
@@ -216,6 +216,6 @@ Veteran::VSOReloader:
   description: Pulls down current list of VSOs and updates accordingly 
 
 CovidVaccine::ScheduledBatchJob:
-  cron: "0 6-19 * * MON-FRI America/New_York"
+  cron: "15 0-23 * * mon-sun America/New_York"
```
ericboehs, line 219:
```suggestion
  cron: "15 0-23 * * * America/New_York"
```

batemapf: @ericboehs certainly agree that the cron wildcards for hours and weekday are much more syntactically pleasant. Challenge is getting those wildcards to evaluate to anything useful for testing purposes. With the less syntactically pleasant but more verbose cron expression, [there is a bit more for the tests to chew on](https://github.com/department-of-veterans-affairs/vets-api/pull/6556/files#diff-3ee178cc87004e38096c1a7a2e51f3a87b145c80e7a1dda7636e3e271e01907cR25).

Alternatively, could get rid of `cron` all together and instead just use `every: '15m'`, which `sidekiq-scheduler` is set up to do. But haven't dug deep enough to know exactly how to write a test around that.

saneshark: I think Eric's suggestion is a bit more common while yours appears a bit more explicit. 

As for specs, you're really just testing the nuts and bolts of sidekiq-cron if you get any deeper than verifying the configuration itself.

patrickvinograd: I don't understand why we even feel the need to have tests around the sidekiq scheduler configuration. Delete those test cases and use the common-sense configuration syntax.

patrickvinograd: And then make sure the PR and commit message express the desired intent so somebody looking at `git blame` can understand what was intended. Not everything needs to be a unit test.

batemapf: Heh, happily will dial back the tests :-)

stephenBarrs (approved):
LGTM

ericboehs (approved):

stephenBarrs (approved):

LindseySaari (approved):

PR #6560, opened 2021-04-12
Tiny Refactor - Open Module Manually So That File Load Order Isn't Important (DecisionReview::Schemas)

edmangimelli:
This is actually a [***super*** tiny refactor](https://github.com/department-of-veterans-affairs/vets-api/pull/6560#discussion_r611831085).

### DIFF
--- a/lib/decision_review/schemas.rb
+++ b/lib/decision_review/schemas.rb
@@ -1,63 +1,65 @@
 # frozen_string_literal: true
 
-module DecisionReview::Schemas
-  nod_create_request_schema_json_string = File.read(
-    Rails.root.join('lib', 'decision_review', 'schemas', 'NOD_create_request_body_schema.json')
-  )
+module DecisionReview
+  module Schemas
+    nod_create_request_schema_json_string = File.read(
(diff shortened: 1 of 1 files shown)


lib/decision_review/schemas.rb
```diff
@@ -1,63 +1,65 @@
 # frozen_string_literal: true
 
-module DecisionReview::Schemas
-  nod_create_request_schema_json_string = File.read(
-    Rails.root.join('lib', 'decision_review', 'schemas', 'NOD_create_request_body_schema.json')
-  )
+module DecisionReview
+  module Schemas
```
edmangimelli, line 4: going from `::` to 2 lines is the only change, this is a really unfortunate diff

ericboehs: "Hide whitespace changes" under the cog.

ericboehs (approved):

lib/decision_review/schemas.rb
edmangimelli: 😮

PR #6563, opened 2021-04-12
Api 6001/audit logging event sourcing

yakschuss:
[API-6001](https://vajira.max.gov/browse/API-6001)

This PR adds functionality for Status Update audit logging, to track meta data around statuses changing in our API.

It does this by using an lightweight EventSourced model to decouple the event from the effect.

‼️ This PR is dependent on [this PR](https://github.com/department-of-veterans-affairs/vets-api/pull/6527) ‼️

### DIFF
--- a/modules/appeals_api/app/services/appeals_api/events/status_updated.rb
+++ b/modules/appeals_api/app/services/appeals_api/events/status_updated.rb
@@ -4,26 +4,26 @@ module AppealsApi
 
       def initialize(opts)
         @opts = opts
-        return false unless has_required_keys?
+        raise InvalidKeys unless has_required_keys?
       end
 
       def hlr_status_updated
         AppealsApi::StatusUpdate.create!(
-          from: opts[:from],
-          to: opts[:to],
-          status_update_time: opts[:status_update_time],
-          appeal_id: opts[:appeal_id],
-          appeal_type: 'AppealsApi::HigherLevelReview',
+          from: opts['from'],
+          to: opts['to'],
+          status_update_time: opts['status_update_time'],
+          statusable_id: opts['statusable_id'],
+          statusable_type: 'AppealsApi::HigherLevelReview',
         )
       end
 
       def nod_status_updated
         AppealsApi::StatusUpdate.create!(
-          from: opts[:from],
-          to: opts[:to],
-          status_update_time: opts[:status_update_time],
-          appeal_id: opts[:appeal_id],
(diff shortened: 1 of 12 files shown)


yakschuss:
> 1 Error
> 🚫	
> Modified files in `db/` should be the only files checked into this PR.
> 
> File Summary
> Database migrations do not run automatically with vets-api deployments. Application code must always be
> backwards compatible with the DB, both before and after migrations have been run. For more info:
> 
> * [Guidance on Safe DB Migrations](https://github.com/ankane/strong_migrations#checks)
> * [`vets-api` Deployment Process](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/engineering/deployment.md)
> 
> 1 Warning
> ⚠️	This PR changes `299` LoC (not counting whitespace/newlines).
> In order to ensure each PR receives the proper attention it deserves, we recommend not exceeding
> `200`. Expect some delays getting reviews.
> 
> File Summary
> Big PRs are difficult to review, often become stale, and cause delays.
> 
> Generated by 🚫 [Danger](https://danger.systems/)

this will be fixed when the dependent PR is merged into master.

modules/appeals_api/app/services/appeals_api/events/status_updated.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module AppealsApi
+  module Events
+    class StatusUpdated
+      def initialize(opts)
```
SlakrHakr, line 6: i see some of these options are required, any particular reason to not use keyword arguments here?

yakschuss: StatusUpdated (or any other callback class) would never actually be called directly from the application - it's proxied from Events::Handler. As a result, the interface requires opts, but the error would be functionally equivalent in a scenario where it's not included - instead of 'missing keyword opts' (which, as I write this, sounds better), it'd be `given 1 argument, expected 0`. I'll add it.

SlakrHakr (approved):

modules/appeals_api/app/services/appeals_api/events/status_updated.rb
yakschuss: Actually, in the interface where the method is called, it is added as a kwarg.

`AppealsApi::Events::Handler.new(opts: opts, event_type: :event_type).handle!` requires that opts is included at the public interface level.

PR #6565, opened 2021-04-12
Api 6123 enhance aud token validation

esubocz-va:
## Description of change

Enhancement to audience validation added to V2. Now allows for multiple audiences in request. Adds an optional `strict` parameter to specify whether or not to allow the audience to validate against the default.

## Original issue(s)
https://vajira.max.gov/browse/API-6123

sample request: 
```
curl -X POST \
  http://localhost:3000/internal/auth/v2/validation \
  -H "Authorization: Bearer $token" \
  -H "apiKey: ${KEY}" \
  --data-urlencode "aud[]=https://sandbox-api.va.gov/services/fhir" \
  --data-urlencode "aud[]=https://sandbox-api.va.gov/services/clinical-fhir" \
  --data "strict=true"
```

Does not affect V1, no V2 breaking changes. 

Unit tests have been added.


lukemarkwordtlibertyits:
# Tests

**Multiple Aud Parameters, matches single token Aud**

- [ ] Should return valid response

**Single Aud Parameter, matches token Aud**

- [ ] Should return valid response

**Multiple Aud parameters, no match**

- [ ] Should return invalid response

**Single Aud parameter, no match**

- [ ] Should return invalid response

**No Aud parameter**

- [ ] should return invalid response

**Default Aud, data-strict=false**

- [ ] Should return valid response

**Default Aud, data-strict=true**

- [ ] Should return invalid response

lukemarkwordtlibertyits:
I have a couple of questions about the functionality:

1. If strict is off and no aud parameter is provided, should the response be a 401 or 200?
2. For this to work the isolated default audience still needs to match the token's audience, is this desirable? It seems that the "match any of the audiences provided in the call" specification does not apply, because only the default audience provided in the vets-api config will work.

modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
```diff
@@ -6,14 +6,42 @@
...
+        if fetch_strict == 'true'
+          if aud.nil?
+            false
+          else
+            aud.include?(token.payload['aud'])
+          end
+        elsif aud.nil?
+          token.payload['aud'] == Settings.oidc.isolated_audience.default
+        else
+          [Settings.oidc.isolated_audience.default, aud].include?(token.payload['aud'])
```
lukemarkwordtlibertyits, line 36: This will need to be called before the `valid_token` method in token.rb is called for it to have its intended results.

lukemarkwordtlibertyits (changes requested):

lukemarkwordtlibertyits (approved):
All tests are now passing

modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
esubocz-va: when aud was an array the `.include?` method was trying to process an array inside an array, which doesn't work. solution creates a one-level array including all aud values as well as the default aud value.

dbrown-va (approved):
A note, the V1 token validation will also support the array of `@aud` values due to the changes inside token.rb, but it is backwards compatible so existing callers will not be effected.

PR #6566, opened 2021-04-13
Api 6346 Update Evidence submission create & show

cilestin:
## Description of change
Updates the EvidenceSubmission create & show endpoints to follow the research we've done with the Matsumoto team to talk to VBADocuments::UploadSubmission directly.

❗ ❗ ~Requires #6571 (db migrations)~  ❗ ❗ 

## Original issue(s)
https://vajira.max.gov/browse/API-6346
https://vajira.max.gov/browse/API-6221

Will need to follow up with other PRs to remove code that is no longer needed.

### DIFF
--- a/modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
+++ b/modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
@@ -5,34 +5,26 @@ module AppealsApi::V1
     module NoticeOfDisagreements
       class EvidenceSubmissionsController < AppealsApi::ApplicationController
         skip_before_action :authenticate
-        before_action :set_submission_attributes, if: -> { params[:nod_id] }
+        before_action :set_submission_attributes, only: :create
 
         def create
-          submission = AppealsApi::EvidenceSubmission.create(@submission_attributes)
-          signed_url = submission.get_location
+          upload = VBADocuments::UploadSubmission.create! consumer_name: 'appeals_api_nod_evidence_submission'
+          submission = AppealsApi::EvidenceSubmission.create! @submission_attributes.merge(upload_submission: upload)
 
-          render json: { data:
-                          {
-                            attributes:
-                              {
-                                status: submission.status,
-                                id: submission.id,
-                                appeal_id: submission.supportable_id,
-                                appeal_type: submission.supportable_type
-                              },
-                            location: signed_url
-                          } }
+          render json: submission,
+                 serializer: AppealsApi::EvidenceSubmissionSerializer,
+                 key_transform: :camel_lower,
+                 render_location: true
         end
 
         def show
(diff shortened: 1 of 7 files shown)


modules/appeals_api/db/migrate/20210412175709_add_upload_submission_to_evidence_submissions.rb
```diff
@@ -0,0 +1,16 @@
+class AddUploadSubmissionToEvidenceSubmissions < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
```
ElleeB, line 2: TIL. What does this do?

modules/appeals_api/db/migrate/20210412175709_add_upload_submission_to_evidence_submissions.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  def change
+    safety_assured do
```
ElleeB, line 5: TIL. What does this do?

ElleeB (commented):
LGTM! Do we need to separate out the migration from this PR into its own?

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
```diff
@@ -5,34 +5,26 @@ module DecisionReviews
...
-                            attributes:
-                              {
-                                status: submission.status,
-                                id: submission.id,
-                                appeal_id: submission.supportable_id,
-                                appeal_type: submission.supportable_type
-                              },
-                            location: signed_url
-                          } }
+          upload = VBADocuments::UploadSubmission.create consumer_name: 'appeals_api_nod_evidence_submission'
```
yakschuss, line 11: I generally prefer the 'bang' version when creating records at the controller level, as it blows up early if something goes wrong. What do you think?

`create!`

modules/appeals_api/db/migrate/20210412175709_add_upload_submission_to_evidence_submissions.rb
cilestin: strong migrations recommended it. It's similar to Prism's old "run_unsafe_migrations" where we're indicating that we know the risks of the operations we're running and that they'll be fine to perform.

Since our table is empty, we're pretty safe ;)

modules/appeals_api/db/migrate/20210412175709_add_upload_submission_to_evidence_submissions.rb
cilestin: strong migrations recommended it due to adding the upload_submission reference and adding the index, since that's a blocking operation.

cilestin:
Database migration errors will be fixed when #6571 is merged

nathanbwright (approved):
Looks good. I think it'd be good to add some additional specs but that can come later.

PR #6572, opened 2021-04-13
API-6346 EvidenceSubmission backwards compat before migrations

cilestin:
## Description of change
Backwards compatibility changes for running the migrations in #6571 

## Original issue(s)
https://vajira.max.gov/browse/API-6346


yakschuss (approved):
🚀  Thanks for getting rid of that stuff! I was intending to open a PR, didnt' realize it was blocking us.

PR #6574, opened 2021-04-13
23096 Remove Linting & Tests from Jenkinsfile

cbtkc:
## Description of change
Linting and specs have been moved to Github Actions. Removing them from the Jenkinsfile.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#23096


Jenkinsfile
```diff
@@ -29,63 +29,6 @@ pipeline {
       }
     }
 
-    stage('Build Docker Images'){
-      steps {
-        withCredentials([string(credentialsId: 'sidekiq-enterprise-license', variable: 'BUNDLE_ENTERPRISE__CONTRIBSYS__COM')]) {
-          sh 'env=$RAILS_ENV make build'
-        }
-      }
-    }
```
ericboehs, line 38: It might be good to still leave this step since we aren't building docker in GHA yet.

Jenkinsfile
```diff
@@ -29,63 +29,6 @@ pipeline {
...
-    stage('Lint All Files') {
-      when { branch 'master' }
-      steps {
-        sh 'env=$RAILS_ENV make lint'
-      }
-    }
-
-    stage('Security Scan') {
-      steps {
-        sh 'env=$RAILS_ENV make security'
```
ericboehs, line 72: Should we remove the associated lint/security rake files? Or should we use them in GHA instead?

Jenkinsfile
cbtkc: The built image here was being used for testing and linting yes? That's why I removed it, I didn't see any other stage that would be using it. If I'm wrong and there is a stage that requires it, let me know.

Jenkinsfile
cbtkc: Interesting question. Is there a reason why they weren't used within GHA? I think I'm indifferent on this particular point, removing them since they aren't used would be the quickest solution.

Jenkinsfile
ericboehs: Right. I'm not saying the step is required but it would be good for a process to continue to run and report to us if a `make build` fails. We haven't moved that to GHA yet so until we do, we should probably continue testing it in Jenkins.

ericboehs: Perhaps the Build AMI stage already does that and would be sufficient?

Jenkinsfile
cbtkc: I think I'd prefer to leave them in though for now. They could be used within GHA.

dillo (approved):

ericboehs:
Closing until we're ready to remove Jenkins testing.

PR #6578, opened 2021-04-14
23173 mobile push logging metrics

jperk51:
## Description of change
Added success and failure counters for both mobile push notifications request endpoints and VEText push service methods. Also added some info logs to mobile request endpoints

## Original issue(s)
department-of-veterans-affairs/va.gov-team#23137


modules/mobile/app/controllers/mobile/v0/push_notifications_controller.rb
```diff
@@ -15,22 +14,30 @@ def register
           params[:device_name] || params[:osName]
         )
 
+        Rails.logger.info('VEText Push service register success', app_name: params[:app_name])
```
kreek, line 17: Since these log lines in the controller are logging on the service success could they be in the service itself? (for `get_prefs`, `set_pref`, and `send_notification` too) 

The controller has an automated log line for each action so we have insight on its status.

kreek (dismissed):
Comment about moving some logging into the service, but otherwise LGTM 🚀

config/initializers/statsd.rb
```diff
@@ -226,3 +226,31 @@
...
+VEText::Service.statsd_count_success :get_preferences,
+                                     'vetext_push.get_prefs'
+VEText::Service.statsd_count_success :set_preference,
+                                     'vetext_push.set_pref'
+VEText::Service.statsd_count_success :send_notification,
+                                     'vetext_push.send_notification'
+VEText::Service.statsd_count_success :app_sid,
+                                     'vetext_push.app_lookup'
+
+StatsD.increment('vetext_push.register.success', 0)
```
LindseySaari, line 243: Kinda nitpicky but wondering if you could pull out 'vetext_push' into a STATSD_KEY_PREFIX constant on the VEText::Service class. This seems to be a pattern amongst other metrics in the initializer

jperk51: Will do. Thanks

kreek (dismissed):
LGTM 🚀

LindseySaari (approved):

PR #6587, opened 2021-04-15
Mobile push get prefs docs fix

jperk51:
## Description of change
Added a response yml I missed initially. Doc only


kreek (approved):
LGTM 🚀

PR #6588, opened 2021-04-15
update eligibilty_info check to use string instead of object key

stephenBarrs:
## Description of change
This PR:
* fixes a bug discovered when processing records where a lookup for the eligibility_info['preferred_facility'] was using an object (`:preferred_facility`) instead of a string (`'preferred_facility'`).
* Adds detail to the Exception raised when look up fails in MPI

The [test case factory trait](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/covid_vaccine/spec/factories/expanded_registration_submissions.rb#L102) that injects the preferred_facility into eligibility_info has been updated

This path is tested with this test case: 
* https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/covid_vaccine/spec/services/covid_vaccine/v0/expanded_registration_service_spec.rb#L124

The test cases for the MPI lookup exceptions are:
* [Profile Not Found](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/covid_vaccine/spec/services/covid_vaccine/v0/expanded_registration_service_spec.rb#L180)
* [Facility Not Found](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/covid_vaccine/spec/services/covid_vaccine/v0/expanded_registration_service_spec.rb#L187)

### DIFF
--- a/modules/covid_vaccine/app/services/covid_vaccine/v0/expanded_registration_service.rb
+++ b/modules/covid_vaccine/app/services/covid_vaccine/v0/expanded_registration_service.rb
@@ -142,7 +142,7 @@ module CovidVaccine
           submission: id,
           submission_date: date
         )
-        raise Common::Exceptions::RecordNotFound, "#{self.class.name}:Error in MPI Lookup: #{error}"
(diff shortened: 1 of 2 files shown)


patrickvinograd (approved):
LGTM

PR #6590, opened 2021-04-15
removed array data at front end request

jperk51:
## Description of change
Removed some array data from get push preferences response array based on front end feedback


kreek (approved):
LGTM 🚀

PR #6594, opened 2021-04-15
[6577] Using MVI COMP1 RMS setting for MPI calls by default

bosawt:
## Description of change
This change makes it so that we call MPI using the MVI.COMP1.RMS setting by default. This makes the request and asks for additional information about relationships of the queried User

## Original issue(s)
department-of-veterans-affairs/vets-api/issues/6577

- We need to confirm that calling MVI.COMP1.RMS instead of MVI.COMP2 does not regress any data. According to the MPI docs, MVI.COMP2 returns user information, and additionally historical icn data. If MVI.COMP1.RMS does not return this information, we may have to call MVI.COMP2 specifically in the cases that we need historical icn. Additionally need to see if this affects any other fields
- Will still need to update the mock data repo to reflect the updated call


joeniquette (dismissed):
I'm approving this based on my understanding of how the code will work, but certainly want to watch how this performs/test in staging prior to a production push.

jimiray (approved):

rileyanderson (approved):

PR #6596, opened 2021-04-16
Api 6066 validation smart launch failure enhancement

esubocz-va:
## Description of change

Adds launch context validation. If smart launch lookup !200, token validation will fail with 401.

## Original issue(s)
https://vajira.max.gov/browse/API-6066

This update is added to both V1 and V2 validation.


app/controllers/openid_application_controller.rb
```diff
@@ -48,9 +44,10 @@ def authenticate_token
         token.payload[:icn] = launch
         token.payload[:launch] = { patient: launch } unless launch.nil?
       end
-      if token.payload['scp'].include?('launch')
-        launch = fetch_smart_launch_context
-        token.payload[:launch] = base64_json?(launch) ? JSON.parse(Base64.decode64(launch)) : { patient: launch }
+      if token.payload['scp'].include?('launch') && !fetch_smart_launch_context.nil?
+        cond = base64_json?(launch)
```
beaugrantham-va, line 48: Where does `launch` get set?

app/controllers/openid_application_controller.rb
```diff
@@ -48,9 +44,10 @@ def authenticate_token
         token.payload[:icn] = launch
         token.payload[:launch] = { patient: launch } unless launch.nil?
       end
-      if token.payload['scp'].include?('launch')
-        launch = fetch_smart_launch_context
-        token.payload[:launch] = base64_json?(launch) ? JSON.parse(Base64.decode64(launch)) : { patient: launch }
+      if token.payload['scp'].include?('launch') && !fetch_smart_launch_context.nil?
```
beaugrantham-va, line 47: These is a block of `launch` logic right above this too ^.

app/controllers/openid_application_controller.rb
```diff
@@ -48,9 +44,10 @@ def authenticate_token
...
+        cond = base64_json?(launch)
+        cond ? JSON.parse(Base64.decode64(launch)) : { patient: launch }
```
beaugrantham-va, line 49: Can we get a better variable name?

modules/openid_auth/app/controllers/openid_auth/application_controller.rb
```diff
@@ -19,5 +19,9 @@ def validate_user
     def fetch_aud
       params['aud']
     end
+
+    def validate_launch_context
+      raise error_klass('Invalid launch context') if fetch_smart_launch_context.nil?
```
beaugrantham-va, line 24: Do we end up calling the launch service multiple times?

modules/openid_auth/app/controllers/openid_auth/v1/validation_controller.rb
```diff
@@ -6,7 +6,7 @@
 module OpenidAuth
   module V1
     class ValidationController < ApplicationController
-      before_action :validate_user
+      before_action :validate_user, :validate_launch_context
```
beaugrantham-va, line 9: How would a patient token pass this check?

esubocz-va: token type check added

app/controllers/openid_application_controller.rb
```diff
@@ -139,13 +138,14 @@ def fetch_aud
   def fetch_smart_launch_context
     response = RestClient.get(Settings.oidc.smart_launch_url,
                               { Authorization: 'Bearer ' + token.token_string })
-    unless response.nil? || response.code != 200
+    raise error_klass('Invalid launch context') if response.nil?
```
bdgrider-va, line 141: This is more of a me learning ruby comment-Is this common or the expected style in ruby, particularly when the thing that would usually come after **if** is a one liner? I would normally say for consistency to change it to a 
if _____
    then ____
but I have a feeling this is not uncommon for this situation.

bdgrider-va (dismissed):
Verified rspec tests pass.

app/controllers/openid_application_controller.rb
esubocz-va: yah it's not uncommon in ruby. I think the linter would have picked it up if it was also.

lukemarkwordtlibertyits:
I have verified the following:
```
 ✓ Valid Token. Patient Scp. Invalid ICN. Should return !200 (401)
 ✓ Valid Token. With Launch/Patient scp. Patient Scp. Invalid ICN. Should return !200 (401)
```
Still am curious as to how regular `patient` act type tokens are going to be handled. Will they also throw 401s?

modules/openid_auth/spec/requests/validation_request_spec_v2.rb
```diff
@@ -287,17 +287,15 @@
   context 'with client credentials jwt and failed launch lookup' do
     before do
       allow(JWT).to receive(:decode).and_return(client_credentials_jwt)
-      allow(RestClient).to receive(:get).and_return(failed_launch_response)
+      allow(RestClient).to receive(:get).and_raise(failed_launch_response)
     end
 
     it 'v2 POST returns true if the user is a veteran' do
```
beaugrantham-va, line 293: Update this test description to reflect the expected behavior (returns a 401 when launch lookup fails).

modules/openid_auth/spec/requests/validation_request_spec.rb
```diff
@@ -204,16 +206,15 @@
   context 'with client credentials jwt and failed launch lookup' do
     before do
       allow(JWT).to receive(:decode).and_return(client_credentials_jwt)
-      allow(RestClient).to receive(:get).and_return(failed_launch_response)
+      allow(RestClient).to receive(:get).and_raise(failed_launch_response)
     end
 
     it 'v1 POST returns true if the user is a veteran' do
```
beaugrantham-va, line 212: Update this test description to reflect the expected behavior (returns a 401 when launch lookup fails).

app/controllers/openid_application_controller.rb
```diff
@@ -139,13 +138,14 @@ def fetch_aud
...
+
+    if response.code == 200
       json_response = JSON.parse(response.body)
       json_response['launch']
     end
-  rescue => e
-    log_message_to_sentry('Error retrieving smart launch context for OIDC token: ' + e.message, :error)
-    nil
+  rescue
+    raise error_klass('Invalid launch context')
```
beaugrantham-va, line 148: I think we should continue to log the error. It should be unexpected for this to occur (outside of our dev/staging/sandbox environment) and having a report in sentry to track down the root cause would be beneficial.

beaugrantham-va (changes requested):

esubocz-va:
> I have verified the following in both V1 and V2:
> 
> ```
>  ✓ Valid Token. Patient Scp. Valid ICN. Should return 200
>  ✓ Valid Token. With Launch/Patient scp. Patient Scp. Valid ICN. Should return 200
>  ✓ Valid Token. Patient Scp. Invalid ICN. Should return !200 (401)
>  ✓ Valid Token. With Launch/Patient scp. Patient Scp. Invalid ICN. Should return !200 (401)
> ```
> 
> Still am curious as to how regular `patient` act type tokens are going to be handled. Will they also throw 401s?
> 
> I suppose the function will be the same as before this PR. Will look into it more tomorrow.

regular tokens do not do a smart launch lookup, so they are not affected by these changes and will return a 200 response.

beaugrantham-va (approved):

PR #6599, opened 2021-04-16
Covid Vaccine expanded submission automate sending data to VeText server

stephenBarrs:
## Description of change
This PR will add a new job that collects the records with state=`enrollment_pending`, does an MVI lookup for ICN and Location matching and if successful sends to VeText.  

## Original issue(s)

I am not sure the way I am collecting the data from the DB model is the most efficient. 
Is there any concern with race conditions as [ScheduledBatchJob](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/covid_vaccine/app/jobs/covid_vaccine/scheduled_batch_job.rb) is also running every 15 minutes and updates state to `enrollment_pending`.  Should these jobs be staggered?

Test specs implemented for submission job and scheduled submission job


modules/covid_vaccine/app/jobs/covid_vaccine/expanded_scheduled_submission_job.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    # Sorting by DESC here because currently if a record fails MPI lookup, it remains in state=enrollment_pending,
+    # so starting at the beginning we may never see the new records as only the MPI error records would process due to limit
+    # with anticipated volume and execution every 15 minutes we should be able to process all new records 
+    # todo: implement task to find and deal with MPI error records and enrollment_out_of_bounds records
+      CovidVaccine::V0::ExpandedRegistrationSubmission.where(state: 'enrollment_pending').order('created_at DESC').limit(1000).map do |submission|
+        CovidVaccine::ExpandedSubmissionJob.perform_async(submission.id)
+      end
+
+      count = CovidVaccine::V0::ExpandedRegistrationSubmission.where(state: 'enrollment_pending').count
```
kreek, line 25: Does getting the count here depend on the `CovidVaccine::ExpandedSubmissionJob`s above being done? Due to their async-ness that's not guaranteed.

stephenBarrs: @kreek yep it sure does.  great point and that count will not work.  maybe we can have a separate scheduled job that runs a couple of times a day to report status of all states in DB (can reuse the [state_summary ](https://github.com/department-of-veterans-affairs/vets-api/blob/master/rakelib/covid_vaccine.rake#L4)rake task code)

config/sidekiq_scheduler.yml
```diff
@@ -223,3 +223,8 @@ CovidVaccine::ScheduledBatchJob:
   cron: "*/15 * * * * America/New_York"
   class: CovidVaccine::ScheduledBatchJob
   description: Batches received vaccine records and schedules upload to enrollment service.
+
+  CovidVaccine::ExpandedScheduledSubmissionJob:
+  cron: "*/15 * * * * America/New_York"
```
batemapf, line 228: Maybe schedule this to run 7.5. minutes after the ScheduledBatchJob immediately above? If you use[ the `every` syntax](https://github.com/moove-it/sidekiq-scheduler#schedule-types), you should be able to achieve without a gnarly `cron` statement. Also might be worth looking at [`first_at` or `first_in` syntax ](https://github.com/jmettraux/rufus-scheduler#first_at-first_in-first-first_time) if you want to give yourself some buffer after prod deploy..

modules/covid_vaccine/app/jobs/covid_vaccine/expanded_scheduled_submission_job.rb
batemapf: +1 I think if we want regular counts of status showing up in logs, that is a separate process. Maybe the only logging here beyond exceptions is the number of records iterated over on ln 21-23.

modules/covid_vaccine/app/jobs/covid_vaccine/expanded_scheduled_submission_job.rb
```diff
@@ -0,0 +1,39 @@
...
+      Rails.logger.info('Covid_Vaccine Expanded_Scheduled_Submission Complete: enrollment_pending count', count: count)
+
+    #   the subtask being called does not raise errors, but keeping this in case some other issue causes this job to fail
+    rescue => e
+        handle_errors(e)
+    end
+
+    def handle_errors(ex)
+      Rails.logger.error('Covid_Vaccine Expanded_Scheduled_Submission: Failed')
```
batemapf, line 34: Confirming that when raised, this will include the record `id` in the log? Would be very helpful for debugging.

config/sidekiq_scheduler.yml
stephenBarrs: @batemapf My concern with using the `first_at` or `first_in` as a prod buffer is that if or when the sidekiq process restarts, the delay would occur again.  if everry time sidekiq restarts (no idea how often that happens) we wait 2days to run, then we might inadvertently cause delays in the future, unless we remove the delay once we confirm everything works

modules/covid_vaccine/app/jobs/covid_vaccine/expanded_scheduled_submission_job.rb
stephenBarrs: @batemapf I added logging with record_id  to [expanded_submission_job](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/covid_vaccine/app/jobs/covid_vaccine/expanded_submission_job.rb), which this Job invokes for each record.  we need to capture the individual record processing errors there.

stephenBarrs: The logging actually happens in the service class when the error occurs, so this is handled already.

modules/covid_vaccine/app/jobs/covid_vaccine/expanded_scheduled_submission_job.rb
stephenBarrs: @batemapf have not implemented logging to indicate number of records iterated over yet. Can look into that with a separate smaller PR

LindseySaari (approved):

LindseySaari (dismissed):

LindseySaari (approved):

PR #6605, opened 2021-04-19
Ensure source tracked

ElleeB:
## Description of change
We added `source` to our nod, hlr, and evidence submission models but never persisted the attribute upon creation. This is the fix.

## Original issue(s)
https://vajira.max.gov/browse/API-6414

### DIFF
--- a/modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
+++ b/modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
@@ -34,7 +34,7 @@ module AppealsApi::V1
           raise Common::Exceptions::RecordNotFound, params[:nod_id] unless @appeal
 
           @submission_attributes ||= {
-            source: request.headers['X-Consumer-Username'],
+            source: params['headers']['X-Consumer-Username'],
             supportable_id: params[:nod_id],
             supportable_type: 'AppealsApi::NoticeOfDisagreement'
           }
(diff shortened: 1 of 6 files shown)


modules/appeals_api/spec/requests/v1/notice_of_disagreements_controller_spec.rb
```diff
@@ -26,6 +26,8 @@ def base_path(path)
     context 'creates an NOD and persists the data' do
       it 'with all headers' do
         post(path, params: @data, headers: @headers)
+        nod = AppealsApi::NoticeOfDisagreement.last
```
yakschuss, line 29: Could you pull the id of the NOD from the response instead of inferring it here?

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
```diff
@@ -34,7 +34,7 @@ def set_submission_attributes
           raise Common::Exceptions::RecordNotFound, params[:nod_id] unless @appeal
 
           @submission_attributes ||= {
-            source: request.headers['X-Consumer-Username'],
+            source: params['headers']['X-Consumer-Username'],
```
cilestin, line 37: shouldn't this be `headers['X-Consumer-Username']` like it is for the HigherLevelReviewsController?

modules/appeals_api/spec/requests/v1/notice_of_disagreements_controller_spec.rb
cilestin: Agreed, there's no guarantee that `.last` will actually return the most recent entry from the database, since without an ORDER clause the database makes no guarantees about order.

`parsed` should have the guid already 👍

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
ElleeB: hlr's controller has a method for parsing the headers
`  def headers
    HEADERS.reduce({}) do |acc, header_key|
      header_value = request.headers[header_key]

      header_value.nil? ? acc : acc.merge({ header_key => header_value })
    end
  end`
kinda doesn't seem worth it here?

yakschuss (approved):

PR #6611, opened 2021-04-19
API-6451: Prevent issue with updating POA if representative does not have an ICN

SlakrHakr:
## Description of change
Sets a default value for external_uid (used for tracking within BGS) if the representative in question does not have an ICN.

## Original issue(s)
Found during this work: https://vajira.max.gov/browse/API-6451
Related sentry error can be found here: http://sentry.vfs.va.gov/organizations/vsp/issues/30399/?environment=sandbox&project=3&query=is%3Aunresolved+level%3Aerror+transaction%3A%22%2AClaimsApi%3A%3A%2A%22&statsPeriod=14d

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,5 @@
 # Vets API
 
-
 [![Build Status](http://jenkins.vfs.va.gov/buildStatus/icon?job=testing/vets-api/master)](http://jenkins.vfs.va.gov/job/builds/job/vets-api/)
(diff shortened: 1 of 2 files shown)


jw81 (approved):
LGTM, but it also looks like Mr. Rubocop has an objection.

jw81 (approved):

jw81 (approved):

PR #6616, opened 2021-04-20
Migration/add metadata to notice of disagreement

ElleeB:
## Description of change
In order to hold onto a notice of disagreement's `boardReviewOption` _(lane)_ after a PII purge, we are adding a column containing metadata in the form of a json blob. The review/lane option, among other data, will help validate evidence submissions.

related to: https://vajira.max.gov/browse/API-6166

### DIFF
--- a/modules/appeals_api/db/migrate/20210420131905_add_metadata_to_notice_of_disagreement.rb
+++ b/modules/appeals_api/db/migrate/20210420131905_add_metadata_to_notice_of_disagreement.rb
@@ -1,0 +1,5 @@
+class AddMetadataToNoticeOfDisagreement < ActiveRecord::Migration[6.0]
+  def change
+    add_column :appeals_api_notice_of_disagreements, :metadata, :jsonb
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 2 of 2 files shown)


modules/appeals_api/db/migrate/20210420131905_add_metadata_to_notice_of_disagreement.rb
```diff
@@ -0,0 +1,5 @@
+class AddMetadataToNoticeOfDisagreement < ActiveRecord::Migration[6.0]
+  def change
+    add_column :appeals_api_notice_of_disagreements, :metadata, :jsonb
```
yakschuss, line 3: Is this supposed to be a catchall for any data we 'might' need for NOD? Wouldn't it be better to just add 'boardReviewOption' and subsequent fields as they're needed?

ElleeB: The idea was kind of a catchall. Eventually we could put the date from which we want to calculate the 90-day submission window for evidence submission. I'm fine either way. It was a decision Drew and I made in light of the revelation that we are not really going to validate at this point and let all evidence submissions through, but that eventually we will want to validate, but may not be able to hold onto PII for the 90 days.

ElleeB:
closing in favor of: https://github.com/department-of-veterans-affairs/vets-api/pull/6620

PR #6623, opened 2021-04-20
Communications preferences batch update api

lihanli:
## Description of change
add batch update method to communication preferences controller. remove create and update apis because frontend says they will only be using the batch update method.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/23065

swagger docs updated


LindseySaari (approved):
👍

PR #6624, opened 2021-04-21
Only perform the enrollment upload job if the setting is enabled

kreek:
## Description of change
In staging there's no SFTP connection so the enrollment upload job shouldn't run there. This PR updates the scheduled batch job to only kick off an enrollment job if the covid_vaccine enrollment_service job_enabled setting is true.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#23535


LindseySaari (approved):

batemapf (approved):
🚀

PR #6625, opened 2021-04-21
Skip CSRF check on KMI opt_out endpoint

patrickvinograd:
## Description of change
Disabled CSRF protection on the covid KMI opt_out endpoint which is triggered via an email opt out link. Working theory is that since it essentially triggers the API request immediately, there may be a race condition vs. getting the CSRF token during page load. 

However it's also the case that the frontend was not passing the CSRF token at all, as remedied here: https://github.com/department-of-veterans-affairs/vets-website/pull/16880
But there is little value in CSRF protection on this particular endpoint so it may be safest to disable it. 

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/23542

### DIFF
--- a/modules/covid_vaccine/app/controllers/covid_vaccine/v0/registration_controller.rb
+++ b/modules/covid_vaccine/app/controllers/covid_vaccine/v0/registration_controller.rb
@@ -9,6 +9,7 @@ module CovidVaccine
       include IgnoreNotFound
 
       before_action :validate_raw_form_data, only: :create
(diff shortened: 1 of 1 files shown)


patrickvinograd:
Local testing indicates that it is timing-dependent whether the CSRF token has been set correctly by the time the unsubscribe API request happens when triggered from an email link (or equivalently in a browser with a fresh cookie state. So I think removing the CSRF enforcement from opt_out makes sense.

stephenBarrs (approved):
LGTM

PR #6626, opened 2021-04-21
[6412] Refactoring User model to remove MPI circular dependency

bosawt:
## Description of change
This change refactors the MPI service and user model so that the `UserIdentity` set of attributes is passed to the MPI service instead of the entire User model. This should help with circular dependencies where MPI needs fields on the `User` model that then require MPI.

## Original issue(s)
department-of-veterans-affairs/vets-api/issues/6412

* There shouldn't be any functional changes with this PR

## Testing Done
* Logged in a user and confirmed that MPI is still called correctly and returns the expected profile
* Called the `AddPerson` route to update user information and confirmed that MPI is called correctly

### DIFF
--- a/app/controllers/v0/profile/personal_informations_controller.rb
+++ b/app/controllers/v0/profile/personal_informations_controller.rb
@@ -21,7 +21,7 @@ module V0
       #   }
       #
       def show
-        response = MPIData.for_user(@current_user).profile
+        response = @current_user.mpi.profile
 
         handle_errors!(response)
 
--- a/app/models/mpi_data.rb
+++ b/app/models/mpi_data.rb
@@ -16,25 +16,28 @@ class MPIData < Common::RedisStore
   REDIS_CONFIG_KEY = :mpi_profile_response
   redis_config_key REDIS_CONFIG_KEY
 
-  # @return [User] the user to query MVI for.
-  attr_accessor :user
+  # @return [UserIdentity] the user identity object to query MVI for.
+  attr_accessor :user_identity
 
-  # Creates a new MPIData instance for a user.
+  # Creates a new MPIData instance for a user identity.
   #
-  # @param user [User] the user to query MVI for
+  # @param user [UserIdentity] the user identity to query MVI for
   # @return [MPIData] an instance of this class
-  def self.for_user(user)
-    MPIData.new(user: user)
+  def self.for_user(user_identity)
+    mvi = MPIData.new
+    mvi.user_identity = user_identity
+    mvi
   end
 
   # Queries MPI specifically to retrieve historical icn data.
   #
   # @param user [User] the user to query MVI for
   # @return [MPIData] an instance of this class
-  def self.historical_icn_for_user(user)
-    return nil unless user.loa3?
+  def self.historical_icn_for_user(user_identity)
+    return nil unless user_identity.loa3?
 
-    mpi_data_instance = MPIData.new(user: user)
+    mpi_data_instance = MPIData.new
+    mpi_data_instance.user_identity = user_identity
     mpi_data_instance.mvi_get_person_historical_icns
   end
 
@@ -98,7 +101,7 @@ class MPIData < Common::RedisStore
   #
   # @return [MPI::Models::MviProfile] patient 'golden record' data from MVI
   def profile
-    return nil unless user.loa3?
+    return nil unless user_identity.loa3?
 
     mvi_response&.profile
   end
@@ -107,7 +110,7 @@ class MPIData < Common::RedisStore
   #
   # @return [String] the status of the last MVI response
   def status
-    return MPI::Responses::FindProfileResponse::RESPONSE_STATUS[:not_authorized] unless user.loa3?
+    return MPI::Responses::FindProfileResponse::RESPONSE_STATUS[:not_authorized] unless user_identity.loa3?
 
     mvi_response.status
   end
@@ -116,7 +119,7 @@ class MPIData < Common::RedisStore
   #
   # @return [Common::Exceptions::BackendServiceException]
   def error
-    return Common::Exceptions::Unauthorized.new(source: self.class) unless user.loa3?
+    return Common::Exceptions::Unauthorized.new(source: self.class) unless user_identity.loa3?
 
     mvi_response.try(:error)
   end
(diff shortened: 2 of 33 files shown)


jimiray (dismissed):

app/controllers/v0/profile/personal_informations_controller.rb
```diff
@@ -21,7 +21,7 @@ class PersonalInformationsController < ApplicationController
       #   }
       #
       def show
-        response = MPIData.for_user(@current_user).profile
+        response = @current_user.mpi.profile
```
bramleyjl, line 24: If we're going to expose `User.mpi.profile` to places outside of the User model do we want to think about making the `mpi_profile` method public and relying on it as the standard way to directly access a User's MPI profile?

bramleyjl (approved):

app/controllers/v0/profile/personal_informations_controller.rb
bosawt: That's a good point, although I wonder if making `mpi_profile` public, we should finally make the push to make `mpi` off the `User` class private. I'm thinking of keeping this for this PR, but adding a story to make `User.mpi` private

rileyanderson (dismissed):

bosawt:
@annaswims I was wondering if you wouldn't mind reviewing this as well. It is an attempt to straighten out the `User` <-> `MPI` relationship, so that we no longer pass a `User` object when querying MPI. I'm hoping this will resolve issues similar to that infinite dependency problem you dealt with before.

jimiray (dismissed):

bramleyjl (dismissed):

bosawt:
@rileyanderson Can I get a re-review when you get a chance? Just needed to rebase master

annaswims:
I don't have time to look today, but I'd love to take a look at this before it's merged.  This reaches significantly into lighthouse code, so I'd suggest an explicit approval from them too. It's a bit strange to me to move towards reliance on UserIdentity (provided by identity providers) when lighthouse doesn't use UserIdentity

bosawt:
> I don't have time to look today, but I'd love to take a look at this before it's merged. This reaches significantly into lighthouse code, so I'd suggest an explicit approval from them too. It's a bit strange to me to move towards reliance on UserIdentity (provided by identity providers) when lighthouse doesn't use UserIdentity

Sounds good, I'll add a Lighthouse representative.
One thing to consider about the transition to `UserIdentity` here, I agree with you, and I'm planning on making it so MPI is called just with an attribute hash, or with a specific class that deals with the attributes, `UserIdentity` was just a convenient stepping stone to that point, and I didn't want to do too much stuff in this PR

SlakrHakr (dismissed):

dbrown-va (approved):
Confirmed Lighthouse token validation and Veteran Verification & Confirmation API functionality from the branch.

annaswims:
could you update app/policies/mpi_policy.rb ?  Presumably it's effectively the same, but it would be more clear if we referenced user_identity there now.

annaswims:
Not 100% related to this, but I'd think [MPI should be queryable](https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/policies/mpi_policy.rb#L9) if the users has an EDIPI even if they don't have all the attributes.

annaswims:
I apologize, but I really don't have time to do a deep dive on this.  The one scenario that almost necessitates a circular dependency is "add person".

Say, a person logs in with MyHealtheVe as LOA3, that means we only have their ICN from the identity provider (no other PII).
We look them up in MPI and see that they have some but not all the necessary ID's to start form 526 (they might be missing `BIRLS id` and `participant id`).  We can get` BIRLS id`  and `participant id` added to MPI by calling "MPI add person", but only if we know their SSN and EDIPI that we got from MPI. 

To be clear, in the way we use "MPI add person" it doesn't actually add a person to MPI it only adds a `BIRLS id` and `participant id` to an existing record.

We have stats on MPI add person success here.  They're not great (in terms of success rate) but maybe it could be a next step for your team. http://grafana.vfs.va.gov/d/000000062/mvi?viewPanel=16&orgId=1. I think those stats will be important to keep an eye on as this rolls out.



in summary: I think we must use user (with data from MPI) to call MPI  add person

bosawt:
> Not 100% related to this, but I'd think [MPI should be queryable](https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/policies/mpi_policy.rb#L9) if the users has an EDIPI even if they don't have all the attributes.

👍 I added a ticket on the backlog to do this

bosawt:
> I apologize, but I really don't have time to do a deep dive on this. The one scenario that almost necessitates a circular dependency is "add person".
> 
> Say, a person logs in with MyHealtheVe as LOA3, that means we only have their ICN from the identity provider (no other PII).
> We look them up in MPI and see that they have some but not all the necessary ID's to start form 526 (they might be missing `BIRLS id` and `participant id`). We can get` BIRLS id` and `participant id` added to MPI by calling "MPI add person", but only if we know their SSN and EDIPI that we got from MPI.
> 
> To be clear, in the way we use "MPI add person" it doesn't actually add a person to MPI it only adds a `BIRLS id` and `participant id` to an existing record.
> 
> We have stats on MPI add person success here. They're not great (in terms of success rate) but maybe it could be a next step for your team. http://grafana.vfs.va.gov/d/000000062/mvi?viewPanel=16&orgId=1. I think those stats will be important to keep an eye on as this rolls out.
> 
> in summary: I think we must use user (with data from MPI) to call MPI add person

I went back and revisited this case just to make sure I didn't miss anything about it. I do add edipi, icn_with_aaid, search_token to a user_identity object that is then passed to the Mpi AddPerson call. Since you mentioned `SSN` as well I added that to be safe:
`  def mpi_add_person
    add_person_identity = identity
    add_person_identity.edipi = edipi
    add_person_identity.ssn = ssn
    add_person_identity.icn_with_aaid = icn_with_aaid
    add_person_identity.search_token = search_token
    mpi.add_person(add_person_identity)
  end`

I still think that Grafana link will be super useful, I'll be watching that whenever I merge this to ensure it doesn't spike. I think this is good to go now, but I'll be monitoring closely for a few days after merge regardless

joeniquette (approved):

jimiray (dismissed):

thilton-oddball (approved):
This all looks good

jimiray (approved):

thilton-oddball (approved):

PR #6628, opened 2021-04-21
Api 6218 charon prep

crolarlibertyva:
## Description of change
Adds prescreening of the token payload that occurs prior to the call to the Charon service.

## Original issue(s)
https://vajira.max.gov/browse/API-6218

- This update is added to V2 validation. This change does not include the call to the Charon service yet.
- Related to https://github.com/department-of-veterans-affairs/devops/pull/9074


modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
```diff
@@ -81,6 +92,24 @@ def setup_structure
...
+
+      def charon_token_screen?(payload_object)
+        vid = payload_object.act[:vista_id]
+        sta3n = payload_object.launch['sta3n']
+        return false unless !vid.nil? && !sta3n.nil?
+
+        parsed_sta3n = vid.match(/\d{3}[A-Z]*\|\d+\^[A-Z]{2}\^\d{3}[A-Z]*\^[A-Z]{5}\|[A-Z]{1}/)
+        return false unless parsed_sta3n
+
+        parsed_sta3n = parsed_sta3n[0].match(/\d{3}|/)
```
beaugrantham-va, line 104: We're told that _most_ users will only have a single vista_id, but it's possible for there to be more than one (as with the test user). I think we need to check them all.

The direction in the story (and other conversations) may have not implied this, and it was probably my bad.

> Ensure that `act.vista_id` contains an entry for `launch.sta3n`

modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
```diff
@@ -11,7 +11,11 @@ class ValidationController < ApplicationController
       def index
         render json: validated_payload, serializer: OpenidAuth::ValidationSerializerV2
       rescue => e
-        raise Common::Exceptions::InternalServerError, e
+        if e.detail.eql?('Invalid request')
```
beaugrantham-va, line 14: Is this necessary? I don't recall seeing this pattern for other 4xx errors.

crolarlibertyva: The endpoint returned 500 if I didn't add the line.

modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
crolarlibertyva: It was clear and forgot to finish a redo. My original implementation covered this but it was much looser check.

modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
crolarlibertyva: I took out the rescue altogether to address this.

modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
```diff
@@ -81,6 +86,28 @@ def setup_structure
         end
         payload_object
       end
+
+      def charon_token_screen?(payload_object)
+        act_vid = payload_object.act[:vista_id]
```
lukemarkwordtlibertyits, line 91: This is just a preference so take it or leave it, but i think vista_id is short enough where the abbreviation to vid is not necessary.

lukemarkwordtlibertyits (changes requested):

modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
```diff
@@ -81,6 +86,28 @@ def setup_structure
...
+        sta3n = payload_object.launch['sta3n']
+        return false unless !act_vid.nil? && !sta3n.nil?
```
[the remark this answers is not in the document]
crolarlibertyva: There has to be launch scope for those token issued for certain particular audiences. because the launch context is where the station id (used as a charon parameter) comes from.  Not all ssoi tokens will need additional validation. But any ssoi token with an audience flagged as one that should need additional charon validation is not authorized if it doesn't have a launch context.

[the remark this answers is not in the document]
lukemarkwordtlibertyits: Ok so in that case the response will be as follows:

```json
{"errors":[{"title":"Not authorized","detail":"Invalid request","code":"401","status":"401"}]}
```

By saying "not all ssoi tokens will need additional validation" you are saying they will not hit the validation endpoint?

modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
crolarlibertyva: Will change to make more readable.

modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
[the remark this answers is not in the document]
crolarlibertyva: no. but rather not go through additional checks, using charon during validation.

lukemarkwordtlibertyits:
Some regression Tests I have ran on this branch:

```sh
✓ SSOI Token. Valid. With Launch scp. Return 200 with appropriate fields
```

lukemarkwordtlibertyits (dismissed):
Approved assuming the following Specs are desired:

1. SSOI Token with Launch Scope and Valid Charon Audience. 
	- Returns valid token response.
2. SSOI Token with Launch Scope and Invalid Charon Audience.
	- Returns valid token response.
3. SSOI Token without Launch Scope and valid Charon Audience.
	- returns 401
```json
{"errors":[{"title":"Not authorized","detail":"Invalid request","code":"401","status":"401"}]}
```

4. SSOI Token withouth Launch Scope and invalid Charon Audience.
	- Returns valid token response.

modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
```diff
@@ -81,6 +86,28 @@ def setup_structure
...
+      def charon_token_screen?(payload_object)
+        act_vista_id = payload_object.act[:vista_id]
+        sta3n = payload_object.launch['sta3n']
+        return false unless !act_vista_id.nil? && !sta3n.nil?
+
+        vista_ids = act_vista_id.scan(/\d{3}[A-Z]*\|\d+\^[A-Z]{2}\^\d{3}[A-Z]*\^[A-Z]{5}\|[A-Z]{1}/)
+        return false unless vista_ids
+
+        vista_ids.each do |vista_id|
+          parsed_sta3n = vista_id.match(/\d{3}|/)
```
beaugrantham-va, line 99: Let me reach out to the team that provided me this regex and ask about valid `sta3n` values.

They didn't provide regex groupings, but assuming that `|` and `^` are delimiters, I'm wondering if this regex should actually be `\d{3}[A-Z]*`.

I'll let you know.

modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
```diff
@@ -81,6 +86,28 @@ def setup_structure
...
+          no_match =  !sta3n.to_s.eql?(parsed_sta3n.to_s)
+          return true unless no_match
```
beaugrantham-va, line 101: Are these two lines (100, 101) driven by linting rules? The double inverted logic is quite hard to read. I had to break out `irb` to make sure I understood this.

`return true if sta3n.to_s.eql?(parsed_sta3n.to_s)` is a lot cleaner.

modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
```diff
@@ -81,6 +86,28 @@ def setup_structure
         end
         payload_object
       end
+
+      def charon_token_screen?(payload_object)
```
beaugrantham-va, line 90: Can this be renamed to a boolean style method (or docs added, or both) to make it clear what to expect from this?

What does it mean for a value of `true` or `false` to be returned? I had to dig into the logic to understand it.

modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
```diff
@@ -10,8 +10,6 @@ class ValidationController < ApplicationController
 
       def index
         render json: validated_payload, serializer: OpenidAuth::ValidationSerializerV2
-      rescue => e
-        raise Common::Exceptions::InternalServerError, e
```
beaugrantham-va, line 14: I'd like to get @dbrown-va to get some eyes on this. I'm not sure how this affects unhandled/runtime errors.

beaugrantham-va:
Can you open and link a PR for the real config changes?

* Enabled in staging, prod
* Disabled dev, sandbox

modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
crolarlibertyva: I will change to make sure. It's safer to go with your recommendation for now and I'll adjust per feedback.

modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
crolarlibertyva: This part is changed in in the part2 changes.

modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
crolarlibertyva: The change I made doesn't include the last "\|" since so the parsed section wouldn't include that delimiter.

modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
dbrown-va: I think this removal will allow for unexpected exceptions to be raised in a less controlled fashion.

Earlier in the PR thread I see there was the `e.detail.eql?('Invalid request')` logic inside the rescue block.

Checking out the branch now to take a look at the different scenarios that could occur.

crolarlibertyva: That original one would need to check e.detail.is_nil? if it were still there.

dbrown-va: I think checking for the `Common::Exceptions::TokenValidationError` error type which is set with the `error_klass` method would provide the most consistency.  Then ensure any other unexpected errors are returned at least as the InternalServerError rather than a stack trace.

dbrown-va (dismissed):
I think there is still opportunity to clean up the re-raised exception to prevent stack traces returned in the response, but that was existing prior to this charon addition.

Should not merge until the config PR is approved and merged.

dbrown-va (approved):
Confirmed Devops PR for settings was merged.

PR #6630, opened 2021-04-21
Set more appropriate log level for KMI MVI records not found

batemapf:
## Description of change
Comment out raising of an exception so we stop triggering a Sidekiq job to log a Sentry error every time a record is not found. Not finding a record is an expected and acceptable outcome. We want to know when it happens for evaluation purposes, but we don't need an error throw every time.

Tracking this [over here for now](https://github.com/department-of-veterans-affairs/va.gov-team/issues/23580).


LindseySaari (approved):

stephenBarrs (dismissed):
LGTM

stephenBarrs (dismissed):
LGTM

stephenBarrs (approved):
LGTM

LindseySaari (approved):

PR #6638, opened 2021-04-21
Mobile appointments record count/data filter update

kreek:
## Description of change
Adds pagination meta data to the appointments list response. Also fixes an issue where cached and fresh appointments weren't being filtered by date in the same way.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#23605


jperk51 (approved):
LGTM

PR #6645, opened 2021-04-22
Feature toggles endpoint refactor

LindseySaari:
## Description of change
It was brought to our attention that IE11 has a URL character limit of 2087 chars. If the character limit is surpassed when calling the `/v0/feature_toggles` endpoint when passing in long toggle names via params, nothing comes back in IE11. 

[Related Slack Conversation](https://dsva.slack.com/archives/CJRQ85PQB/p1619054544454400)

This PR updates the `/v0/feature_toggles` (index action) with a case statement to return all feature toggles. Since the frontend uses a variety of camel and snake casing with no consistent pattern, toggles are returned in both formats. Since the new option won't have any params, we won't know how they are expecting it and uniformity is need to respect camel and snake casing styles.

**The existing method of passing in features params is still enabled to support backwards compatibility until the frontend can make changes.**

Examples of snake & camel casing mixtures on frontend:

1. https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/applications/static-pages/health-care-manage-benefits/fixtures/feature-toggles/disabled.json
2. https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/applications/static-pages/health-care-manage-benefits/fixtures/feature-toggles/enabled.json
3. https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/platform/utilities/feature-toggles/featureFlagNames.js

By removing the need to pass the feature toggles via URL params, this will resolve the IE11 URL character limitation

## Original issue(s)
department-of-veterans-affairs/va.gov-team#23511

- [x] Passing specs
- [x] Camel and snake casing covered via specs


app/controllers/v0/feature_toggles_controller.rb
```diff
@@ -6,22 +6,34 @@ class FeatureTogglesController < ApplicationController
     skip_before_action :authenticate
     before_action :load_user
 
+    # rubocop:disable Metrics/AbcSize
```
ericboehs, line 9: Move this to the todo yml.

LindseySaari: On it 👍

app/controllers/v0/feature_toggles_controller.rb
```diff
@@ -12,16 +12,26 @@ def index
...
         end
       else
         features = []
+
+        # returning both camel and snakecase for uniformity on FE
+        FLIPPER_FEATURE_CONFIG['features'].collect do |feature_name, values|
+          features << { name: feature_name.camelize(:lower),
+                        value: Flipper.enabled?(feature_name, actor(values['actor_type'])) }
+          features << { name: feature_name, value: Flipper.enabled?(feature_name, actor(values['actor_type'])) }
+        end
```
ericboehs, line 27: Instead of returning both, could we base it on the `X-Key-Inflection` header?

mchelen-gov: it doesn't look like the [FE code](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/platform/utilities/feature-toggles/flipper-client.js#L8-L10) currently sets that header, @timwright12 does that seem like a reasonable thing to include?

ericboehs: They won't have to include it but the feature mappings will need to be updated because they include both snack case and camelCase arbitrarily.

LindseySaari: @timwright12 confirmed that the FE doesn't currently set the header

ericboehs (approved):

PR #6653, opened 2021-04-22
Add covid_vaccine enrollment_service job_enabled flag to default settings

kreek:
## Description of change
Add covid_vaccine enrollment_service job_enabled flag to default settings. Most likely the cause of a spec that passed locally on the merge check but failed during deployment.

## Original issue(s)
N/A


ericboehs (approved):

ericboehs:
Thanks for the quick fix on this!

PR #6657, opened 2021-04-22
Adds primary_eligibility attribute as part of enrollment_status on healthcare_applications_controller

grossb1:
## Description of change
Adds `primary_eligibility` attribute as part of enrollment_status on `healthcare_applications_controller`

This is needed by VAOS direct scheduling to ensure we’re only allowing veterans and not spouses, caretakers, etc who might have recently received a vaccine from the VA.

## Original issue(s)
@grossb1 can you update with ticket number?

### DIFF
--- a/app/models/health_care_application.rb
+++ b/app/models/health_care_application.rb
@@ -96,10 +96,9 @@ class HealthCareApplication < ApplicationRecord
       )
 
       ee_data.slice(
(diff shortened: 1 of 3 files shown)


saneshark (dismissed):
LGTM

LindseySaari (approved):

PR #6659, opened 2021-04-22
Revert "Communications preferences batch update api"

lihanli:
Reverts department-of-veterans-affairs/vets-api#6623
VA profile team told us in a meeting today that the batch update API will not be production ready in the near future, so we are going to use the single update APIs for now


LindseySaari (approved):
👍

LindseySaari (approved):

PR #6667, opened 2021-04-23
Rely on sidekiq retry and add logging for duplicate HLRPDFSubmitJob

yakschuss:
[API-6424](https://vajira.max.gov/browse/API-6424)

This PR:

HLRPDFSubmitJob is potentially sending duplicate requests to CentralMail. We _think_ that the reason is because we're both relying on CentralMail::Utilities#retry_errors, which manually schedules another job, and then some failure (for which no logging exists) then raises and is retried automatically by sidekiq.

This PR attempts to add some logging to both error states, and removes the dependency on CentralMail::Utilities.

Potential Issue - this change will mean that any UploadError due to upstream outages (DOC201) will need to manually be retried, as the HLR will be in 'error' state. (which is kinda what it's for anyway)

This issue will present for NOD as well, assuming this is the problem.

CC: @nathanbwright (co-authored)

### DIFF
--- a/modules/appeals_api/app/workers/appeals_api/higher_level_review_pdf_submit_job.rb
+++ b/modules/appeals_api/app/workers/appeals_api/higher_level_review_pdf_submit_job.rb
@@ -14,23 +14,23 @@ module AppealsApi
     include Sidekiq::MonitoredWorker
     include CentralMail::Utilities
 
-    def perform(higher_level_review_id, retries = 0)
+    def perform(higher_level_review_id)
       higher_level_review = AppealsApi::HigherLevelReview.find(higher_level_review_id)
 
       begin
-        @retries = retries
         stamped_pdf = AppealsApi::PdfConstruction::Generator.new(higher_level_review).generate
         higher_level_review.update!(status: 'submitting')
         upload_to_central_mail(higher_level_review, stamped_pdf)
         File.delete(stamped_pdf) if File.exist?(stamped_pdf)
       rescue => e
         higher_level_review.update!(status: 'error', code: e.class.to_s, detail: e.message)
+        Rails.logger.info("#{self.class} error: #{e}")
         raise
       end
     end
 
     def retry_limits_for_notification
-      [6, 10]
+      [2, 5]
     end
 
     def notify(retry_params)
@@ -57,8 +57,7 @@ module AppealsApi
       process_response(CentralMail::Service.new.upload(body), higher_level_review)
       log_submission(higher_level_review, metadata)
     rescue AppealsApi::UploadError => e
-      e.detail = "#{e.detail} (retry attempt #{@retries})"
-      retry_errors(e, higher_level_review)
+      handle_upload_error(higher_level_review, e)
     end
 
     def process_response(response, higher_level_review)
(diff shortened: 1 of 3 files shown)


modules/appeals_api/app/workers/appeals_api/higher_level_review_pdf_submit_job.rb
```diff
@@ -75,5 +74,27 @@ def receive_date(higher_level_review)
...
+                        'uuid' => higher_level_review.id,
+                        'code' => e.code,
+                        'detail' => e.detail)
+    end
+
+    def handle_upload_error(higher_level_review, e)
+      log_upload_error(higher_level_review, e)
+      higher_level_review.update(status: 'error', code: e.code, detail: e.detail)
+
+      if e.code == 'DOC201' || e.code == 'DOC202'
```
ElleeB, line 92: Is it that any other error codes are related to the submitted content?

yakschuss: Yeah. The 2– prefix is for issues with the CMP server. You can see a list of the codes here:
[lib/centrail_mail/upload_error.rn](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/central_mail/upload_error.rb)

nathanbwright (approved):
👍

modules/appeals_api/app/workers/appeals_api/higher_level_review_pdf_submit_job.rb
cilestin, line 92: I feel like we should notify in this case as well -feels kinda bad form to wait for our daily error report to know that this stalled out and needs manual intervention.

yakschuss: good point

nathanbwright (approved):
Looks good!

modules/appeals_api/lib/appeals_api/sidekiq_retry_notifier.rb
```diff
@@ -35,6 +35,12 @@ def request_headers
         }
       end
 
+      def retried_at(retried_time)
+        return 'was not retried' unless retried_time
```
SlakrHakr, line 39: bit of a nitpick, but if the flow hits this line, it looks like the message will read like this:
```
This job failed at: {current-time}, and was retried at: was not retried
```
would it make more sense to have this `retried_at` method return the last bit of the above sentence instead? so
```
      def retried_at(retried_time)
        return 'was not retried' unless retried_time
        
        "was retried at: #{Time.zone.at(retried_time)}"
      end
```

i dont see anywhere else this method is called that this change might break at least

SlakrHakr (dismissed):

modules/appeals_api/lib/appeals_api/sidekiq_retry_notifier.rb
yakschuss: Yeah, that's better. I'll change it! Thanks

nathanbwright (approved):
Looks good to me!

cilestin (approved):
:shipit:

PR #6668, opened 2021-04-23
Api 5348/status simulation

nathanbwright:
## Description of change
This PR adds the ability to simulate an appeal or evidence submission's status *if* the correct header is passed in and the env is `development`, `staging`, or `sandbox`.

Questions:
- Should the we silently ignore invalid statuses (current implementation)?
- Do we think `StatusSimulator` should live in its own file?
- Would it be preferable to change the concern specs to match the `RSpec/PredicateMatcher` style? https://docs.rubocop.org/rubocop-rspec/2.2/cops_rspec.html#rspecpredicatematcher I personally don't prefer this style but I'm willing to change them.

## Original issue(s)
https://vajira.max.gov/browse/API-5348


modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/higher_level_reviews_controller.rb
```diff
@@ -37,6 +38,9 @@ def schema
   end
 
   def show
+    if status_simulation_reqested? && status_simulation_allowed?
```
yakschuss, line 41: typo here - 'requested'

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/higher_level_reviews_controller.rb
```diff
@@ -37,6 +38,9 @@ def schema
...
+      @higher_level_review = status_simulation_for(@higher_level_review)
```
yakschuss, line 42: I like this pattern, but my sense is that the method name `status_simulation_for` doesn't indicate that the return value is a wrapped object. What do you think?

modules/appeals_api/app/controllers/concerns/appeals_api/status_simulation.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module AppealsApi
+  class StatusSimulator < SimpleDelegator
+    attr_reader :status
+
+    def status=(status)
```
yakschuss, line 7: Where is this method used?

yakschuss (dismissed):
Looks really good!

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/higher_level_reviews_controller.rb
nathanbwright: Yeah, that is a good point... Do you have an idea for a more fitting name?

modules/appeals_api/app/controllers/concerns/appeals_api/status_simulation.rb
nathanbwright: This is used on line 31 of this file

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
```diff
@@ -22,6 +24,12 @@ def show
           submission = AppealsApi::EvidenceSubmission.find_by(guid: params[:id])
           raise Common::Exceptions::RecordNotFound, params[:id] unless submission
 
+          # rubocop:disable Style/IfUnlessModifier
```
SlakrHakr, line 27: why disable this instead of just putting the if inline like im assuming its wanting?

nathanbwright: Yeah, that is a good question. I did it for two reasons... One is that it matches the shape of the other classes call sites and I like keeping things looking the same if possible. And the other reason is that I don't think having a `=`, `if`,  and`&&` all on one line is very readable...

```
submission = status_simulation_for(submission) if status_simulation_requested? && status_simulation_allowed?
```

Thoughts?

modules/appeals_api/app/swagger/appeals_api/v1/api_description.md
```diff
@@ -43,6 +43,16 @@ The submission statuses begin with pending and end with success.
 | success   | The centralized mail portal, Digital Mail Handling System (DHMS), has received the data. |
 | error   | An error occurred. See the error code and message for further information. |
 
+#### Status Simulation
+
+Sandbox test submissions do not progress to Central Mail / VBMS.  In the lower environments (i.e. Sandbox or Staging), the final status of a submission is either `submitted` (for NOD) or `received` (for HLR). In the lower environments, we allow passing a `Status-Simulation` header on the show endpoints so that you can simulate the other statuses.
```
yakschuss, line 48: Based on my E2E testing, this is factually incorrect. What actually happens is that it hits the environment, but CMP (and VBMS, I guess) does receive it, but doesn't take action with it. (I know this is the language used by VBA, but i'm wondering if we should clarify.)

yakschuss (approved):
Really good work!

modules/appeals_api/app/serializers/appeals_api/evidence_submission_serializer.rb
```diff
@@ -14,10 +14,7 @@ def id
       object.guid
     end
 
-    def status
-      object.upload_submission.status
-    end
-
+    delegate :status, to: :object
```
cilestin, line 17: 👍

cilestin (approved):
Very nice! 🚀

PR #6673, opened 2021-04-26
Mobile claims overview completed filter

jperk51:
## Description of change
Added a new optional query param for the claims and appeals overview mobile endpoint. The new param will allow consumer to filter results to show only open record, only completed records, or all records. Tests have been added and docs have been updated.


kreek (approved):
LGTM 🚀

PR #6676, opened 2021-04-26
API-6280 - SSOi session caching

lukemarkwordtlibertyits:
Caches Okta User Profile, so the validation will not need to call Okta for every SSOI validation call.

## Description of change

- Added a profile attribute to the OpenId session

## Original issue(s)

https://vajira.max.gov/browse/API-6280

To the end user this PR will not create any changes. The best way I have found to confirm caching functionality is through debugging.


lukemarkwordtlibertyits:
Regarding the Github action Failures, there appears to be some trouble shooting going on and it is not ready to be an authority yet. Jenkins CI should be trusted for now. See the following [thread](https://dsva.slack.com/archives/CBU0KDSB1/p1619470804460300).

app/controllers/openid_application_controller.rb
```diff
@@ -125,8 +133,8 @@ def fetch_profile(uid)
     end
   end
 
-  def build_session(ttl)
-    session = Session.new(token: token.to_s, uuid: token.identifiers.uuid)
+  def build_session(ttl, profile)
```
dbrown-va, line 136: Rather than storing the entire Okta profile response here as an object, I think we should minimize the profile down to only the necessary information.

That should only be:
- last_login_type
- icn
- sec_id
- vista_id
- npi

app/controllers/openid_application_controller.rb
```diff
@@ -37,29 +37,37 @@ def authenticate_token
...
+
+  def populate_payload_for_launch_patient_scope
+    launch = fetch_smart_launch_context
+    token.payload[:icn] = launch
+    token.payload[:launch] = { patient: launch } unless launch.nil?
+  end
+
   def populate_ssoi_token_payload(profile)
     token.payload['last_login_type'] = 'ssoi'
     token.payload['icn'] = profile.attrs['icn']
```
dbrown-va, line 73: Semi-related to this PR, I think it would be good to get the sec id from the profile, rather than the token.payload['sub'].  https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb#L52

`token.payload['sec_id'] = profile.attrs['SecID']`

dbrown-va (approved):

app/controllers/openid_application_controller.rb
```diff
@@ -106,7 +115,10 @@ def establish_session(profile)
     ttl = token.payload['exp'] - Time.current.utc.to_i
     user_identity = OpenidUserIdentity.build_from_okta_profile(uuid: token.identifiers.uuid, profile: profile, ttl: ttl)
     @current_user = OpenidUser.build_from_identity(identity: user_identity, ttl: ttl)
-    @session = build_session(ttl)
+    @session = build_session(ttl,
+                             Okta::UserProfile.new({ 'last_login_type' => profile['last_login_type'],
+                                                     'SecID' => profile['SecID'], 'VistaId' => profile['VistaId'],
+                                                     'npi' => profile['npi'], 'icn' => profile['icn'] }))
     @session.save && user_identity.save && @current_user.save
```
dbrown-va, line 122: One thing to keep in mind, currently the `derived_loa` method in user_profile.rb doesn't handled SSOi.  Which results in the user_identity object failing validation and not saving to the cache.  And with that failure the `@current_user` save isn't attempted either.

For token validation by itself, not really a problem.  But something to keep in mind in case in the future an SSOi user wanted to use Veteran Verification API.  In that case, the `@current_user` build up is required to have the necessary id's to call underlying service endpoints.

PR #6680, opened 2021-04-26
I think this fixes it 😬

yakschuss:
The `before do` request was creating some intermittent failure modes locally when the request would return a 500 or just not process for every example.


ElleeB:
Thank you!

ElleeB (approved):
Thank you!

PR #6681, opened 2021-04-26
Cris.shupp api 6651 duplicate uploads

cshupp1:
## Description of change

We are adding in an application-level lock to ensure an upload submission never occurs more than once.

## Original issue(s)
https://vajira.max.gov/browse/API-6651

A detailed RSPEC test is being added that has sub-processes to test the application lock.  If the lock is removed the tests do fail!!


gpbowman-git (approved):
looks good. nice work!

PR #6688, opened 2021-04-27
Mobile claim type metric update

jperk51:
## Description of change
Added statsd call to claim request so we start seeing data for claim type viewed


kreek (approved):
LGTM 🚀

PR #6693, opened 2021-04-27
API-6116 Persist board_review_hearing first pass

ElleeB:
THIS IS NOW INCLUDED IN THIS PR: https://github.com/department-of-veterans-affairs/vets-api/pull/6701
THIS PR WILL BE CLOSED

### DIFF
--- a/modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements_controller.rb
+++ b/modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements_controller.rb
@@ -79,7 +79,8 @@ class AppealsApi::V1::DecisionReviews::NoticeOfDisagreementsController < Appeals
     @notice_of_disagreement = AppealsApi::NoticeOfDisagreement.new(
       auth_headers: headers,
       form_data: @json_body,
-      source: headers['X-Consumer-Username']
+      source: headers['X-Consumer-Username'],
+      board_review_option: @json_body['data']['attributes']['boardReviewOption']
     )
     render_model_errors unless @notice_of_disagreement.validate
   end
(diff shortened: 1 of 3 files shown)


modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements_controller.rb
```diff
@@ -79,7 +79,8 @@ def new_notice_of_disagreement
     @notice_of_disagreement = AppealsApi::NoticeOfDisagreement.new(
       auth_headers: headers,
       form_data: @json_body,
-      source: headers['X-Consumer-Username']
+      source: headers['X-Consumer-Username'],
+      board_review_option: @json_body['data']['attributes']['boardReviewOption']
```
yakschuss, line 83: It's unfortunate that we need this, but there's no defensive measure here in case of `nil`. What happens if `['data']['attributes']['boardRevviewOption']` is nil?

ElleeB: right right right blah!

ElleeB: Well, wait, if board review is nil...schema error...it's required and will get caught in create when it goes to save the object.

ElleeB: It validates the JSON schema before `new_notice_of_disagreement`, so it would never get to this point and be nil.

yakschuss (dismissed):

modules/appeals_api/spec/models/notice_of_disagreement_spec.rb
```diff
@@ -121,13 +124,15 @@
         veteran_data.delete('address')
       end
 
-      it do
-        expect(notice_of_disagreement.zip_code_5).to eq '00000'
-      end
+      it { expect(notice_of_disagreement.zip_code_5).to eq '00000' }
```
ElleeB, line 127: a little change for the sake of consistency

ElleeB:
The content of this PR is not here: #6701

PR #6695, opened 2021-04-27
Remove unique md5 requirement for POA records

hinzed1127:
### Background Context

We initially added the POA `status` to the md5 hash to make it pass validation for uniqueness via this flow:

- set to rep A
- set to rep B
- set back to rep A

However, this still fails on this workflow after a few go-rounds, since there's only a handful of status options it will have:

    set to rep A (with status `pending` -> 'updated`)
    set to rep B (with status `pending` -> `updated`)
    set back to rep A (with status `pending` -> failing on uniqueness going to `updated`)
    set back to rep B (with status `pending` -> failing on uniqueness going to `updated`)
    set back to rep A ( failing on uniqueness for `pending` and `updated`

Since the lookups can happen with just the header_md5, I'm not sure if the md5 is even needed at all now, but at least removing the uniqueness requirement avoids validation failures

## Original issue(s)
https://vajira.max.gov/browse/API-6192

### DIFF
--- a/modules/claims_api/app/models/claims_api/power_of_attorney.rb
+++ b/modules/claims_api/app/models/claims_api/power_of_attorney.rb
@@ -13,17 +13,18 @@ module ClaimsApi
 
     PENDING = 'pending'
     UPDATED = 'updated'
+    SUBMITTED = 'submitted'
     ERRORED = 'errored'
 
-    before_validation :set_md5
-    validates :md5, uniqueness: true
+    before_save :set_md5
 
     def self.find_using_identifier_and_source(source_name:, id: nil, header_md5: nil, md5: nil)
       primary_identifier = {}
       primary_identifier[:id] = id if id.present?
       primary_identifier[:header_md5] = header_md5 if header_md5.present?
       primary_identifier[:md5] = md5 if md5.present?
-      poas = ClaimsApi::PowerOfAttorney.where(primary_identifier).order(:created_at)
+      # it's possible to have duplicate POAs, so be sure to return the most recently created match
+      poas = ClaimsApi::PowerOfAttorney.where(primary_identifier).order(created_at: :desc)
       poas = poas.select { |poa| poa.source_data['name'] == source_name }
       return nil if poas.blank?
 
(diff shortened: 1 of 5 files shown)


modules/claims_api/spec/requests/v0/power_of_attorney_request_spec.rb
```diff
@@ -49,15 +49,6 @@
         expect(parsed['data']['type']).to eq('claims_api_power_of_attorneys')
         expect(parsed['data']['attributes']['status']).to eq('pending')
       end
-
-      it 'returns the same successful response with all the data' do
```
hinzed1127, line 53: this test (and the same test in v1) is expected to fail now, since a _new_ record will get created. should I add a different test in its place (like the scenario described in the description) or not worry about it?

modules/claims_api/app/models/claims_api/power_of_attorney.rb
```diff
@@ -13,17 +13,18 @@ class PowerOfAttorney < ApplicationRecord
 
     PENDING = 'pending'
     UPDATED = 'updated'
+    SUBMITTED = 'submitted'
```
hinzed1127, line 16: this was the only state not captured with a constant, so I just added that here, as well

rileyanderson (approved):

modules/claims_api/spec/requests/v0/power_of_attorney_request_spec.rb
SlakrHakr: i think this is fine to just remove personally

modules/claims_api/app/models/claims_api/power_of_attorney.rb
```diff
@@ -13,17 +13,18 @@ class PowerOfAttorney < ApplicationRecord
...
     ERRORED = 'errored'
 
-    before_validation :set_md5
-    validates :md5, uniqueness: true
+    before_save :set_md5
```
SlakrHakr, line 19: why change this to before_save?

SlakrHakr (approved):

modules/claims_api/app/models/claims_api/power_of_attorney.rb
hinzed1127: since this removes the md5 uniqueness validation, I don't think there's any validations that explicitly get run (unless there are implicit ones that still do?), so I figured just explicitly doing it pre-save works just as well

PR #6702, opened 2021-04-27
API-6718 - Update NOD header schema

nathanbwright:
## Description of change
This updates the naming conventions (for the request headers) used in the NOD. Previously we were using `X-VA-Veteran-First-Name` instead of how the HLR is structured `X-VA-First-Name`.

## Original issue(s)
https://vajira.max.gov/browse/API-6718

### DIFF
--- a/modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
+++ b/modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
@@ -32,19 +32,19 @@ module AppealsApi
     end
 
     def veteran_first_name
-      header_field_as_string 'X-VA-Veteran-First-Name'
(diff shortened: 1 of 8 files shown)


yakschuss (approved):
Looks great!

ElleeB (approved):
This looks great. The only consideration it brings up for me is the reintroduction of Claimant to the equation. For example, how will we differentiate Veteran name vs Claimant name? Food for thought.

nathanbwright:
> This looks great. The only consideration it brings up for me is the reintroduction of Claimant to the equation. For example, how will we differentiate Veteran name vs Claimant name? Food for thought.

Yes, that is a good point that we'll need to keep in mind.

PR #6706, opened 2021-04-28
update NOD headers

annaswims:
Lighthouse is changing header names in pr #6702 so we are making the corresponding chnges.  The VCR cassettes already have the names of the new headers.


annaswims:
I think https://github.com/department-of-veterans-affairs/vets-api/pull/6707 will need to be merged first so that brakeman will pass.

nathanbwright (approved):
👍

PR #6709, opened 2021-04-28
Full vertical integration for get_appointments for the new VAOS service

grossb1:
## Description of change

Added full vertical integration for get_appointments for the new VAOS service. Once the service is hosted in staging, VCR cassettes will be officially recorded and further unit testing will commence.

## Original issue(s)


grossb1:
@department-of-veterans-affairs/vfs-vaos Is there a way to relax the character limit requirement?  We're adding a new swagger file.

modules/vaos/config/routes.rb
```diff
@@ -43,7 +43,9 @@
   end
 
   namespace :v2, defaults: { format: :json } do
+    get '/appointments', to: 'appointments#index'
     get '/locations/:location_id/clinics', to: 'clinics#index'
     get '/patients/', to: 'patients#index'
+    end
```
rileyanderson, line 49: Extra `end` here not needed

rileyanderson (changes requested):

modules/vaos/config/routes.rb
grossb1: ack, didn't rerun my tests locally after merging.

Dangerfile
```diff
@@ -48,7 +48,7 @@ module VSPDanger
   class ChangeLimiter
     EXCLUSIONS = %w[
       *.csv *.json *.tsv *.txt Gemfile.lock app/swagger modules/mobile/docs spec/fixtures/ spec/support/vcr_cassettes/
-      modules/mobile/spec/support/vcr_cassettes/ db/seeds
+      modules/mobile/spec/support/vcr_cassettes/ db/seeds modules/vaos/app/docs
```
ajmagdub, line 51: If the size of the PR is just a warning this change is probably best to leave out.

grossb1: It was a recommendation made a member of the vfs platform team.

modules/vaos/app/controllers/vaos/v2/appointments_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+      def appointments_service
+        VAOS::V2::AppointmentsService.new(current_user)
+      end
+
+      def appointments
+        @appointments ||=
+          appointments_service.get_appointments(start_date, end_date, pagination_params)
+      end
+
+      def validate_params
```
ajmagdub, line 25: Normally require and permit are used for parameter presence validation. Any reason that was used?

ajmagdub (dismissed):
left a couple of comments otherwise LGTM.

modules/vaos/app/controllers/vaos/v2/appointments_controller.rb
grossb1: I was wondering that myself, it was pulled from the original V0 controller but can probably be changed.  @saneshark ?

Dangerfile
ajmagdub: good, just wanted to make sure the platform team would complain about it.

modules/vaos/app/controllers/vaos/v2/appointments_controller.rb
grossb1: Going to push this for now, because it's blocking other work.  Will revisit ASAP.

ajmagdub (approved):
LGTM

modules/vaos/app/controllers/vaos/v2/appointments_controller.rb
saneshark: I think it might be worth re-evaluating the logic here. Not super critical, but nice to have.

only downside is that we don't surface all the errors and instead just the first one we ran into.

saneshark (approved):
LGTM

rileyanderson (approved):
LGTM

PR #6710, opened 2021-04-28
Covid Vaccine - remove enrollment_out_of_band and process records with no preferred facility

stephenBarrs:
## Description of change
This PR removes a pre-existing state in the DB model state machine of `enrollment_out_of_band`. This state was previously used to indicate that there was no `preferred_facility` included with the submitted form data, which can occur if the zip code lookup did not find any VAMCs nearby.  These records will now be processed in the same way as other records that fail to match an entered preferred_facility with the users facilities association array in MPI: after 24 hours these records will be sent to the VeText backend with status of `registered_no_icn` if they were not found in MPI at all, and `registered_no_facility` if they were found in MPI.  

Test cases have been updated to remove reference to the `enrollment_out_of_band` state and to include tests with created_at dates both more and less than 24 hours for test data that does not include a `preferred_facility`

Existing records in the production database with state = `enrollment_out_of_band` will need to be migrated to state = `enrollment_pending`.  This is a small subset of records, currently less than 400 total.


modules/covid_vaccine/spec/factories/expanded_registration_submissions.rb
```diff
@@ -90,6 +90,10 @@
       }
     end
 
+    # trait :enrollment_out_of_band do
```
rileyanderson, line 93: Is this meant to be commented out? If so, should it just be removed?

batemapf (dismissed):

modules/covid_vaccine/spec/factories/expanded_registration_submissions.rb
stephenBarrs: thanks yep removed those comments.

rileyanderson (approved):
LGTM. Just will need to merge master in to fix the failing pipelines

PR #6714, opened 2021-04-28
[API-6441] - Build ClaimsAPI v1 docs from specs

jw81:
## Description of change

This PR attempts to use [rswag](https://github.com/rswag/rswag) to build our OpenApi/Swagger docs from specs.  If we choose to proceed with this work, we should be able to delete a good amount of our old Swagger generation files that are spread throughout the codebase.

## Original issue(s)
[API-6441](https://vajira.max.gov/browse/API-6441)

This is the command to invoke `rswag` and have it generate OpenApi/Swagger docs from specs.
`rake rswag:specs:swaggerize PATTERN="modules/claims_api/spec/requests/**/*_spec.rb" SWAGGER_DRY_RUN=0`

A new `Setting` has been added to allow us to toggle between showing the old `developer-docs` and the new rswag-generated docs.  That logic is specifically defined [here](https://github.com/department-of-veterans-affairs/vets-api/pull/6714/files#diff-3281bf3430dd7f2d6520d5115d44184ac8ec8655293f59d5e18d59b5170bf281R30).  If running locally, that `Settings.claims_api.rswag_docs.enabled` setting will need to be defined in `vets-api/config/settings.local.yml`.  There is an [accompanying PR](https://github.com/department-of-veterans-affairs/devops/pull/9136) that enables this new setting for just the `Sandbox` environment.

This PR will absolutely trigger the `danger` bot logic and will require override authorization if we choose to move forward with this implementation.  Hopefully that won't be a big deal since this is a docs replacement and not actual implementation changes.


modules/claims_api/spec/requests/v1/rswag_claims_request_spec.rb
```diff
@@ -0,0 +1,245 @@
+# frozen_string_literal: true
+
+require 'swagger_helper'
+require 'rails_helper'
+require_relative '../../support/swagger_shared_components'
+
+describe 'EVSS Claims management' do  # rubocop:disable RSpec/DescribeClass
+  path '/claims' do
```
SlakrHakr, line 8: is there anything we can easily do with the response for this endpoint in doc? its just gigantic so it takes a second for the dropdown on the frontend to open

jw81: Yeah, this is one of those pieces that I wasn't exactly sure how we wanted to handle.
My initial thought was to go to the VCR recording and figure out how to manually cut down the response payload, but I wasn't certain if that's the approach we'd want to take.

SlakrHakr: could at the very least have a second vcr recording that is a single claim in the list for this specific test no?

SlakrHakr:
I used http://jsondiff.com/ just now to diff the openapi docs in this pr against what is currently in production and realzied the schema and validate endpoints are missing from these. We're going to need those if we want to push this out to production i think.

modules/claims_api/spec/requests/v1/rswag_intent_to_file_request_spec.rb
```diff
@@ -0,0 +1,319 @@
...
+      let(:'X-VA-Birth-Date') { '1986-05-06T00:00:00+00:00' }
+      let(:Authorization) { 'Bearer token' }
+
+      parameter name: :data, in: :body, required: true, schema: {
+        type: :object,
+        properties: {
+          data: {
+            type: :hash,
+            description: 'JSON API Payload of Veteran being submitted',
+            example: 'DO WE WANT AN EXAMPLE HERE???'
```
SlakrHakr, line 39: yeah, mostly since our current doc has one, and the example is used to provide an example curl call

SlakrHakr:
it looks like none of the responses have schema showing in the developer portal?

modules/claims_api/spec/requests/v1/rswag_intent_to_file_request_spec.rb
SlakrHakr: weirdly, the schema for this endpoint request looks strange, like its looking at the example to build off of or something? that doesnt really make sense but the request schema does look off for whatever reason

jw81:
> I used http://jsondiff.com/ just now to diff the openapi docs in this pr against what is currently in production and realzied the schema and validate endpoints are missing from these. We're going to need those if we want to push this out to production i think.

Yeah, I completely ignored them because they've been deprecated, but I guess you're right that if we want this to be an apples-to-apples swap, then we'll probably have to include them.

jw81:
> it looks like none of the responses have schema showing in the developer portal?

This should be taken care of now.

modules/claims_api/spec/requests/v1/rswag_intent_to_file_request_spec.rb
jw81: This should be fixed now.

modules/claims_api/spec/requests/v1/rswag_claims_request_spec.rb
jw81: This has been handled [here](https://github.com/department-of-veterans-affairs/vets-api/pull/6714/commits/a4f08fabc42887e1d74da6eb0dbfcb6f5d5164f8).

modules/claims_api/app/swagger/v1/swagger.json
```diff
@@ -0,0 +1,7171 @@
...
+            "required": false,
+            "description": "Veteran birthdate if consumer is representative",
+            "schema": {
+              "type": "string"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "claim response",
```
SlakrHakr, line 92: production has schema showing in the developer portal for responses but i dont see any for this rswag stuff, is schema for responses not possible with rswag?

modules/claims_api/config/schemas/2122.json
```diff
@@ -5,72 +5,6 @@
   "required": [
     "serviceOrganization"
   ],
-  "definitions": {
```
SlakrHakr, line 8: more curious than anything, why move away from these definitions defined up here? im guessing rswag / the developer portal didnt like it this way or something?

spec/swagger_helper.rb
```diff
@@ -0,0 +1,145 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.configure do |config|
+  # Specify a root folder where Swagger JSON files are generated
+  # NOTE: If you're using the rswag-api to serve API descriptions, you'll need
+  # to ensure that it's configured to serve Swagger from the same folder
+  config.swagger_root = Rails.root.join('modules', 'claims_api', 'app', 'swagger').to_s
```
SlakrHakr, line 9: how does this work if another module wanted to use rswag? since this file isnt within our module

SlakrHakr: well.... lame... i don't see a way to do this for multiple modules... this might just have to be good enough for now i guess

SlakrHakr: ok, i think this is fine for now. it looks like we could make swagger_root root of vets-api and then define multiple modules within this file, if im reading it correctly.

modules/claims_api/app/swagger/v1/swagger.json
SlakrHakr: alright, well.... i found out what needs to be done to get schema to show up, but its not a quick and easy thing :(
don't want to commit what i have but you essentially have to define the schema for each response, either right in the spec or in the "components" section and reference them in the spec.

thats the bad news, we have to define structures for each response.
good news is that the tests do fail if the expected schema does not match what is in the response of the spec, so thats good at least.

https://github.com/rswag/rswag#getting-started
Step 3 shows an example of defining the schema within the spec itself

https://github.com/rswag/rswag#referenced-parameters-and-schema-definitions
shows an example of referencing a defined schema component.

SlakrHakr (approved):

SlakrHakr (approved):

PR #6715, opened 2021-04-28
Returns 404 when virtual_agent_token toggle is turned off; returns 200 when turned on

anitadewitt-va:
## Original issue(s)
department-of-veterans-affairs/va-virtual-agent#172

we added a feature toggle called virtual_agent_token

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -181,7 +181,7 @@
         steps:
             - run:
                 name: Audit bundled gems for CVE issues
-                command: bundle exec bundle-audit check --update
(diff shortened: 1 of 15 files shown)


rileyanderson (changes requested):
Looks good. Just need to address the Rubocop error then should be good to go

rileyanderson (changes requested):
~Can you merge master into your branch before we merge this? This will help clean up all those extraneous commits and changes that do not need to go through with this again~

rileyanderson (approved):
Actually looks good. Apparently it is a GitHub UI issue where they erroneously show changes already in master but it looks like it fixed itself

PR #6716, opened 2021-04-28
Cache SM triage teams, folders and folder messages

narin:
## Description of change
Adds Redis caching for Triage Teams, Folders, and Folder Messages in Secure Messaging using a RedisCaching concern in preparation for flagship mobile launch.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#23333

* Redis namespace for SM is currently being shared between mobile and website 
* Future plan is to refactor SM out of lib and use the module/proxy pattern similar to mobile appointments and claims so we can have separate namespaces and reuse the existing RedisCaching concern in `modules/mobile/app/models/mobile/v0/concerns/redis_caching.rb`


lib/sm/client.rb
```diff
@@ -108,22 +108,35 @@ def delete_folder(id)
...
+          page_data = perform(:get, path, nil, token_headers).body
+          json[:data].concat(page_data[:data])
+          json[:metadata].merge(page_data[:metadata])
+          break unless page_data[:data].size == MHV_MAXIMUM_PER_PAGE
+
+          page += 1
+        end
+        messages = Common::Collection.new(Message, json)
+        Message.set_cached(cache_key, messages)
+        messages
```
narin, line 136: Thinking I might also be able to use the `get_cached_or fetch_data` helper method I created below with `yield` somehow, but still getting a grasp of how to do that.

spec/lib/sm/client/triage_teams_spec.rb
```diff
@@ -18,7 +18,7 @@
     end
 
     it 'gets a collection of triage team recipients', :vcr do
-      folders = client.get_triage_teams
+      folders = client.get_triage_teams('1234', false)
```
kreek, line 21: Are there specs for when there's already cached data? Should be able to use the classes that use the concern in a before hook to pre-cache.

app/models/folder.rb
```diff
@@ -5,6 +5,9 @@
 # Folder model
 class Folder < Common::Base
   include ActiveModel::Validations
+  include RedisCaching
```
kreek, line 8: This ends up being a nice light weight way to add caching without affecting all `Common::Base` children 👍

kreek (commented):
LGTM so far, couple minor comments, plus it'd be good to get specs for when there's already cached data.

app/controllers/v0/folders_controller.rb
```diff
@@ -3,7 +3,8 @@
 module V0
   class FoldersController < SMController
     def index
-      resource = client.get_folders
+      use_cache = params[:useCache] == 'true'
```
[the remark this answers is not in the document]
narin: Good idea.  Will implement this.

kreek:
@patrickvinograd when using IAM as the service provider we don't get a similar id (or not one we could match to the ID.me uuid). Because IAMUsers extend regular User models they need a uuid to use as their redis key. The solution was to generate a UUID for IAM users via their sec_id and icn `Digest::UUID.uuid_v5(@iam_sec_id, @iam_icn)`.

For the caching if it's left as is there could be two caches per user. As you mention, this could be fine 😃. If we wanted more of a distinction, and not cache for .gov, we could have a mobile only redis config and move the caching logic into a proxy service that sits between the mobile controllers and vets-api core service client (lib client). Appointments and Claims have mobile service proxies, mostly for adapting data, their caching logic is in their controllers but it could easily move to their proxies if we wanted to be consistent.

patrickvinograd:
Ok, that works for me, I'm not concerned about the duplication at this point.

kreek (approved):

thilton-oddball (approved):
This all looks good

PR #6718, opened 2021-04-29
Blanking out form detail URL when a form is unpublished

mviterna:
## Original issue(s)
https://vajira.max.gov/browse/API-6703

No to all

Local testing


gpbowman-git (approved):
looks good. Just align the end with the if statement (unless the linter did that to you)

mviterna:
> looks good. Just align the end with the if statement (unless the linter did that to you)

Oh, that's definitely the linter's work...

PR #6719, opened 2021-04-29
Api 6766/add disagreement reason to contestable issues

yakschuss:
[API-6766](https://vajira.max.gov/browse/API-6766)

This PR adds the field 'disagreementReason' to our contestableIssues schema for Notice Of Disagreement.

This field is optional, and based on feedback that would allow the BVA to process NOD's more accurately and with more expediency.


nathanbwright:
Do you think we should add a disagreement reason to the overflow table for pdf test?

modules/appeals_api/app/services/appeals_api/pdf_construction/notice_of_disagreement/v1/structure.rb
```diff
@@ -179,6 +168,34 @@ def additional_pages?
           form_data.hearing_type_preference.present? ||
             form_data.contestable_issues.count > 5
         end
+
+        # rubocop:disable Metrics/MethodLength
+        def insert_issues_into_text_boxes(pdf, text_opts)
```
cilestin, line 173: The way this is written it seems like if both the issue text & disagreementReason are long, the two bounding boxes will overlap each other before they start autosizing.

I think we should detect if an issue has a disagreement reason and change the bounding boxes between the reason & disagreement so there isn't a possibility of overlap.

We'd need to change the `ypos` & `height` of each depending on whether `disagreementReason` is present/blank.

Let's also double check the documentation in case there's character limits (maxLength) there and adjust them if need be (unless that's already different ticket).

cilestin (changes requested):

modules/appeals_api/app/services/appeals_api/pdf_construction/notice_of_disagreement/v1/structure.rb
yakschuss: Discussed on zoom.

cilestin (approved):

PR #6731, opened 2021-04-30
Move all status updates to happen within the HLR and NOD models

nathanbwright:
## Description of change
This PR moves all the status update calls for HLRs and NODs into the models themselves so that we can handle additional logic surrounding status updates more easily.

## Original issue(s)
https://vajira.max.gov/browse/API-6650


yakschuss (approved):
🚀

cilestin (dismissed):

cilestin (approved):

PR #6743, opened 2021-05-03
Removing URL when unpublished

mviterna:
## Original issue(s)
https://vajira.max.gov/browse/API-6703

No to all

Local testing

### DIFF
--- a/modules/va_forms/app/workers/va_forms/form_reloader.rb
+++ b/modules/va_forms/app/workers/va_forms/form_reloader.rb
@@ -91,7 +91,8 @@ module VAForms
         if form['entityPublished']
           "#{FORM_BASE_URL}#{form.dig('entityUrl', 'path')}"
         else
-          '' end
+          ''
+        end
       mapped
     end
 
--- a/modules/va_forms/app/workers/va_forms/form_reloader.rb
+++ b/modules/va_forms/app/workers/va_forms/form_reloader.rb
@@ -87,12 +87,7 @@ module VAForms
         benefit_categories: map_benefit_categories(form['fieldBenefitCategories']),
         va_form_administration: form.dig('fieldVaFormAdministration', 'entity', 'entityLabel')
       }
-      mapped[:form_details_url] =
-        if form['entityPublished']
-          "#{FORM_BASE_URL}#{form.dig('entityUrl', 'path')}"
-        else
-          ''
-        end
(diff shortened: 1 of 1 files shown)


modules/va_forms/app/workers/va_forms/form_reloader.rb
```diff
@@ -87,7 +87,11 @@ def init_attributes(form)
         benefit_categories: map_benefit_categories(form['fieldBenefitCategories']),
         va_form_administration: form.dig('fieldVaFormAdministration', 'entity', 'entityLabel')
       }
-      mapped[:form_details_url] = "#{FORM_BASE_URL}#{form.dig('entityUrl', 'path')}" if form['entityPublished']
+      mapped[:form_details_url] =
+        if form['entityPublished']
+          "#{FORM_BASE_URL}#{form.dig('entityUrl', 'path')}"
+        else
+          '' end
```
rileyanderson, line 94: Just a style change here, the `end` should be on the next line 

```ruby
if form['entityPublished']
  "#{FORM_BASE_URL}#{form.dig('entityUrl', 'path')}"
else
  ''
end
```

rileyanderson (changes requested):

modules/va_forms/app/workers/va_forms/form_reloader.rb
mviterna: I was thinking Rubocop forced me to do it that way, but I could have dreamed that up. Let me try it again...

mviterna: @rileyanderson Yeah, if I add a newline, rubocop says the method is too long. Do you have a preference how you want to get past that? If I put it all on one line, it looks too long to me

rileyanderson: Ah yeah it looks like rubocop is catching too many lines in the method 😞 . You could use a ternary operator here to keep method lines down and to keep styling consistent 

```ruby
mapped[:form_details_url] = form['entityPublished'] ? "#{FORM_BASE_URL}#{form.dig('entityUrl', 'path')}" : ''
```

mviterna: done

rileyanderson (approved):
LGTM

PR #6746, opened 2021-05-03
Add POA code check for /active POA endpoints

hinzed1127:
## Description of change

We want to make sure only representatives with accredited POA codes can check the active POA for a veteran. This adds in that validation.

## Original issue(s)
https://vajira.max.gov/browse/API-6464

Might wanna add a few tests, too


modules/claims_api/app/controllers/claims_api/v0/forms/power_of_attorney_controller.rb
```diff
@@ -79,6 +79,8 @@ def status
         def active
           power_of_attorney = ClaimsApi::PowerOfAttorney.find_using_identifier_and_source(header_md5: header_md5,
                                                                                           source_name: source_name)
+
+          validate_poa_code_for_current_user!(current_poa_code)
```
SlakrHakr, line 83: i might be misunderstanding but im not sure this is what we need, i think something more like
```
# method in concerns/poa_verfication
# https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/app/controllers/concerns/claims_api/poa_verification.rb#L30
def validate_user_is_accredited! 
    representative = ::Veteran::Service::Representative.for_user(first_name: @current_user.first_name,
                                                                                                     last_name: @current_user.last_name)
    raise ::Common::Exceptions::Forbidden if representative.blank?
end


validate_user_is_accredited! if header_request?
```

what is currently here, i think, only allows the current representative of a Veteran to use this endpoint. what is described in the ticket is to allow any accredited representative to use this endpoint for any Veteran.

modules/claims_api/app/controllers/claims_api/v0/forms/power_of_attorney_controller.rb
```diff
@@ -100,6 +102,14 @@ def schema
 
         private
 
+        def current_poa_code
+          current_poa.try(:code)
+        end
+
+        def current_poa
+          @current_poa ||= BGS::PowerOfAttorneyVerifier.new(target_veteran).current_poa
+        end
```
SlakrHakr, line 111: i dont think these are needed if
https://github.com/department-of-veterans-affairs/vets-api/pull/6746/files#r625204314
is addressed

SlakrHakr (changes requested):
i dont think the current logic maps to what is being asked for in the ticket

modules/claims_api/app/controllers/claims_api/v0/forms/power_of_attorney_controller.rb
hinzed1127: ahh ok I see now. I thing I was mixed up on the semantics of the `for_user` method and _who_ the check was actually being applied to

modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
```diff
@@ -87,6 +87,8 @@ def status
         #
         # @return [JSON] Last POA change request through Claims API
         def active
+          validate_user_is_accredited! if header_request?
```
hinzed1127, line 90: If I understand the logic better now, it doesn't seem like there needs to be this same check in v0 because a veteran  checking their own active POA doesn't need this same validation, right?

modules/claims_api/spec/requests/v1/power_of_attorney_request_spec.rb
```diff
@@ -202,6 +209,16 @@
           end
         end
       end
+
+      context 'when a non-accredited representative and non-veteran request active power of attorney' do
```
hinzed1127, line 213: similar question as above. since this only validates non-veteran requests, do we need to worry about adding `/active/` tests for veteran requests? I don't think there's any in this file, so the only _new_ test I added was a specific failing case for the accredited validation

modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
markgreenburg: Well, v0 doesn't really have any knowledge of a particular user and there is explicit trust in the system that has production access. So this check would only be possible to perform on v1.

modules/claims_api/spec/requests/v1/power_of_attorney_request_spec.rb
markgreenburg: Yeah, I mean, for v1 specifically we should be validating that the Veteran making the request is the same Veteran as the subject of the request. For v0...this doesn't apply, since Veterans and representatives can't use v0.

SlakrHakr (approved):

rileyanderson (approved):

PR #6748, opened 2021-05-03
Mobile claims overview date filter update

jperk51:
## Description of change
Changed claims and appeals overview date filter so not providing a start date with the request will cause the default start date to be Jan 1 1700. Updated tests and docs to reflect this change.


rileyanderson (approved):

modules/mobile/app/controllers/mobile/v0/claims_and_appeals_controller.rb
```diff
@@ -99,7 +99,7 @@ def paginate(list, params, _show_completed)
 
       def fill_missing_params(params)
         [
-          params[:startDate] || (DateTime.now.utc.beginning_of_day - 1.year).iso8601,
+          params[:startDate] || DateTime.new(1700).iso8601,
```
kreek, line 102: I know we're looking for a date that's far back in time but could it be one after the founding of the VA or the first computer 😀 Not sure how far back the current records go? Maybe the start date could be 10-25 years ago? 

Or, since now we basically want all the records, do we need date params at all?

PR #6755, opened 2021-05-03
Api 6411/status update event

yakschuss:
[API-6411](https://vajira.max.gov/browse/API-6411)

This PR emits a status updated event in the [newly created](https://github.com/department-of-veterans-affairs/vets-api/pull/6731) `update_status!` interface for both NOD and HLR.

As a result, it's also dependent on that PR.

Potential Issues: As of right now, it seems that the module namespaced initializer isn't working as expected, and the EventSubscriptions are not being created locally. Will continue to investigate, this won't break any of the code that exists.


cilestin (dismissed):

nathanbwright (dismissed):
Looks good!

cilestin (approved):

PR #6765, opened 2021-05-04
migrations to go with #6764 - Ready for review, despite status

annaswims:
migrations to go with #6764

### DIFF
--- a/app/models/appeal_submission.rb
+++ b/app/models/appeal_submission.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class AppealSubmission < ApplicationRecord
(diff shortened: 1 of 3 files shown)


app/models/appeal_submission.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class AppealSubmission < ApplicationRecord
+  self.ignored_columns = ['board_review_otpion']
```
annaswims, line 4: I know dangerbot is going to complain about this line but `strong_migrations` insisted on it.  So, botfight! 🤖 💥

rileyanderson (approved):

PR #6701, opened 2021-04-27
API-6116 Add evidence submission validation

ElleeB:
Step 1
In order to validate our nod evidence submissions, we need to persist a nod's board_review_option upon its creation in order to check that it's marked as evidence_submission.

Step 2
Validate the board_review_option for incoming evidence submission.

**Step 3 In Progress**
Validate agains NOD's Veteran SSN _if_ PII still available

---

This PR is a follow up to the recent addition of board_review_option to our nod model schema.

Relates to https://vajira.max.gov/browse/API-6166

Updates made to the multiple specs with the addition of the new NOD attribute.

Also includes fix for: https://vajira.max.gov/browse/API-6840

PR #6769, opened 2021-05-04
Fix bug with header parsing on EvidenceSubmission create

cilestin:
## Description of change
The specs were creating a `header` key on the params rather than actually sending them in as headers to the request. We then built the functionality of extracting the `source` from this incorrect key.

Updated to send the headers properly and extract the `source` from the request headers.

## Original issue(s)
https://vajira.max.gov/browse/API-6840

### DIFF
--- a/modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
+++ b/modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
@@ -39,7 +39,7 @@ module AppealsApi::V1
 
         def set_submission_attributes
           @submission_attributes ||= {
-            source: params['headers']['X-Consumer-Username'],
+            source: request.headers['X-Consumer-Username'],
             supportable_id: params[:nod_id],
             supportable_type: 'AppealsApi::NoticeOfDisagreement'
           }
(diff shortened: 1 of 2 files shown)


yakschuss (dismissed):
🚀

nathanbwright (approved):
👍

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
```diff
@@ -39,7 +39,7 @@ def set_notice_of_disagreement
 
         def set_submission_attributes
           @submission_attributes ||= {
-            source: params['headers']['X-Consumer-Username'],
+            source: request.headers['X-Consumer-Username'],
```
ElleeB, line 42: I'm getting nil for `request.headers` when I try this locally.

yakschuss: I think the dot notation and the parameter notation both work for the request object.

cilestin: We are using dot notation elsewhere in our controllers so I think we can stick with that.

cilestin:
Closing since #6701 incorporates it.

PR #6770, opened 2021-05-04
rubocop todos

SlakrHakr:
## Description of change
Trying to resolve previously marked todo items within rubocop

There is a lot here but the vast majority of these changes were auto-corrected by rubocop and then just reviewed by me.


batemapf (approved):

jimiray (dismissed):
HOORAY for string interpolation and getting rid of "dig", this was something I wondered about for a while.

kreek:
Could this be broken up by product, or maybe by cop? Seems fairly innocuous but it touches basically the entire API.

SlakrHakr:
> Could this be broken up by product, or maybe by cop? Seems fairly innocuous but it touches basically the entire API.

I actually planned on doing it in smaller chunks but after doing just a few of them i was already over the 500 loc limit so i figured id just do a few more. I can try to break them up if people don't feel comfortable with this large pr though.

stephenBarrs (dismissed):

SlakrHakr:
closing this for now as its proving too big and spread across the codebase to get enough people to sign off before something else is pushed to master that breaks a rubocop rule.

will probably open up smaller prs for this stuff in the future

PR #6771, opened 2021-05-04
Api 6669/91 day validation evidence window

yakschuss:
[API-6669](https://vajira.max.gov/browse/API-6669)

This PR adds a check ensuring that the NOD is within the submission window for evidence submission (90 days)


cilestin (approved):
LGTM!

PR #6705, opened 2021-04-28
API-6425 Authorization check with the Charon service where applicable

crolarlibertyva:
## Description of change

Provide additional screening for authorization for tokens issued through SSOI. This is done with a call the the internal Charon service. 

## Original issue(s)

https://vajira.max.gov/browse/API-6425

## Things to know about this PR

This PR should  follow after:

- https://github.com/department-of-veterans-affairs/vets-api/pull/6628
- https://github.com/department-of-veterans-affairs/devops/pull/9074

PR #6789, opened 2021-05-06
API-6838 Response body charon errors

dbrown-va:
## Description of change
As a follow up to earlier token validation PR's, this adds more details to the non-200 responses from the token validation service.

## Original issue(s)
- https://vajira.max.gov/browse/API-6838
- https://github.com/department-of-veterans-affairs/vets-api/pull/6705

To return both 401 and 500 responses from the token validation service, the Common::Exceptions::TokenValidationError error transform to the Common::Exceptions::Unauthorized error has been removed.

Rspec tests have been added for 200, 401, and 500 response types.  Additionally, existing tests for other 401 responses were confirmed working and a default 401 was set for the TokenValidationError response.

### DIFF
--- a/modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
+++ b/modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
@@ -142,7 +142,7 @@ module OpenidAuth
                                   { params: { duz: duz, site: site } })
         response
       rescue => e
-        raise Common::Exceptions::TokenValidationError.new(detail: 'Failed validation with Charon.', status_code: 500)
+        raise Common::Exceptions::TokenValidationError.new(status: 500, code: 500, detail: 'Failed validation with Charon.')
       end
     end
   end
--- a/modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
+++ b/modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
@@ -13,9 +13,6 @@ module OpenidAuth
         render json: validated_payload, serializer: OpenidAuth::ValidationSerializerV2
       rescue => e
         case e
-        when RestClient::ExceptionWithResponse
-          status_code = e.response.code >= 500 ? 503 : 401
-          render status: status_code
         when Common::Exceptions::TokenValidationError
           raise e
         else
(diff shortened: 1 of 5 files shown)


modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
```diff
@@ -136,12 +131,25 @@ def should_validate_with_charon?(aud)
...
+          raise error_klass('Charon menu-code: ' + json_response['value'])
+        else
+          raise Common::Exceptions::TokenValidationError.new(
+            status: 500, code: 500, detail: 'Failed validation with Charon.'
+          )
+        end
       rescue => e
-        log_message_to_sentry('Failed validation with Charon', :error, body: e.message)
-        raise e
+        log_message_to_sentry('Error retrieving charon context for OIDC token: ' + e.message, :error)
```
dbrown-va, line 149: @beaugrantham-va, thoughts on logging unexpected charon responses to sentry?

This is similar to what we do fetching the launch context.

modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
```diff
@@ -136,12 +131,25 @@ def should_validate_with_charon?(aud)
...
-        response = RestClient.get(Settings.oidc.charon.endpoint,
-                                  { params: { duz: duz, site: site } })
-        response.code == 200
+        RestClient.get(Settings.oidc.charon.endpoint, { params: { duz: duz, site: site } })
+      rescue RestClient::ExceptionWithResponse => e
+        case e.response.code
+        when 400
+          json_response = JSON.parse(e.response.body)
+          raise error_klass(json_response['message'])
+        when 401
```
beaugrantham-va, line 140: The Charon docs also list 403 as a valid response code.

modules/openid_auth/app/controllers/openid_auth/v2/validation_controller.rb
beaugrantham-va: Yes, this seems appropriate.

beaugrantham-va (approved):

crolarlibertyva (approved):

PR #6796, opened 2021-05-06
Raise an error if file uploads aren't present for file upload controllers.

annaswims:
## Description of change
The file upload endpoints respond with 200 if there's no file. We'll make it 400 instead. 

## Original issue(s)


spec/controllers/concerns/form_attachment_create_spec.rb
```diff
@@ -43,14 +46,26 @@
...
+        'code' => '103',
+        'status' => '400'
+      )
+    end
+
+    def expect_form_attachment_creation
       form_attachment = build(attachment_factory_id, guid: form_attachment_guid)
 
       expect(form_attachment_model).to receive(:new) do
-        expect(form_attachment).to receive(:set_file_data!).with(
```
annaswims, line 50: @kevinmirc Mind if I delete this expectation?  `req_params[param_namespace][:file_data]` is a ` Rack::Test::UploadedFile` but `filtered_params[:file_data]` is a `ActionDispatch::Http::UploadedFile`

kevinmirc: That's fine by me. If I can think of a better way to set that expected param, I'll just do a follow-up PR for it later. Thanks!

kevinmirc (approved):

rileyanderson (approved):

PR #6797, opened 2021-05-06
add versioning parameter to pdf submit job

yakschuss:
[API-6894](https://vajira.max.gov/browse/API-6894)

This optional parameter for PDF submit job allows for multiple versions of pdf generation to occur

### DIFF
--- a/modules/appeals_api/app/workers/appeals_api/notice_of_disagreement_pdf_submit_job.rb
+++ b/modules/appeals_api/app/workers/appeals_api/notice_of_disagreement_pdf_submit_job.rb
@@ -14,13 +14,13 @@ module AppealsApi
     include Sidekiq::MonitoredWorker
     include CentralMail::Utilities
 
-    def perform(id, retries = 0)
+    def perform(id, retries = 0, version = 'V1')
       @retries = retries
       notice_of_disagreement = NoticeOfDisagreement.find(id)
 
       begin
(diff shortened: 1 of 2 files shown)


modules/appeals_api/app/workers/appeals_api/notice_of_disagreement_pdf_submit_job.rb
```diff
@@ -14,13 +14,13 @@ class NoticeOfDisagreementPdfSubmitJob
     include Sidekiq::MonitoredWorker
     include CentralMail::Utilities
 
-    def perform(id, retries = 0)
+    def perform(id, retries = 0, version = 'V1')
```
cilestin, line 17: I thought we got rid of the retries already? Or was that just for the NOD submit job? :\

ElleeB: Can you guys help me remember why we removed retries from HLR but not NOD? Is that something we're going to do but haven't yet, or is there a reason for one and not the other?

ElleeB: nm found the original ticket for the HLR retries.

SlakrHakr (approved):

cilestin (approved):

PR #6812, opened 2021-05-07
API-6460: Prefill correct POA form

SlakrHakr:
## Description of change
Benefits Claims should prefill the correct 21-22 pdf, on a POA change submission, based on the Veteran's representative.
If the representative is an individual, our system should submit a 21-22A.
If the representative is part of an organization, our system should submit a 21-22. 

## Original issue(s)
https://vajira.max.gov/browse/API-6460

Existing tests pass and new tests added.
Generated pdfs manually checked for accuracy.

Note: I had to restructure a good portion of this existing logic to add this new concept of filling out different pdf's.

Examples of pdfs generated from specs:
[25f54146-c758-45e5-bbd5-5fba744f4507_final.pdf](https://github.com/department-of-veterans-affairs/vets-api/files/6444439/25f54146-c758-45e5-bbd5-5fba744f4507_final.pdf)
[f68d7629-65bb-4650-8c95-b4b5b52532e8_final.pdf](https://github.com/department-of-veterans-affairs/vets-api/files/6444442/f68d7629-65bb-4650-8c95-b4b5b52532e8_final.pdf)


modules/claims_api/spec/workers/poa_form_builder_job_spec.rb
```diff
@@ -29,7 +30,7 @@
         },
         phone: {
           areaCode: '555',
-          ohoneNumber: '5551337'
+          phoneNumber: '5551337'
```
SlakrHakr, line 33: this, and the similar spelling mistake below might be my favorite thing about this pr.
i spent far too long trying to understand why this information wasn't being populated into the pdf before i discovered this 😞

modules/claims_api/spec/support/pdf_matcher.rb
```diff
@@ -4,7 +4,16 @@
...
+    actual_reader = PDF::Reader.new(actual)
+    actual_pages = actual_reader.pages.size
+    actual_text = actual_reader.pages.map(&:text)
+
+    expected_reader = PDF::Reader.new(expected)
+    expected_pages = expected_reader.pages.size
+    expected_text = expected_reader.pages.map(&:text)
+
+    # Check both page count and text content in case there are extraneous blank pages
+    actual_pages == expected_pages && actual_text == expected_text
```
SlakrHakr, line 16: blatantly stolen from 
https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/appeals_api/spec/support/pdf_matcher.rb#L16
to resolve the error i was seeing: "PDF does not contain EOF marker"

might be valuable to move that helper file outside modules for all to share but didn't for now considering how much is already changing with this pr.

modules/claims_api/lib/claims_api/poa_pdf_constructor/organization.rb
```diff
@@ -0,0 +1,95 @@
...
+        ]
+      end
+
+      # rubocop:disable Layout/LineLength
+      def page2_options(data)
+        base_form = 'F[0].Page_2[0]'
+        {
+          "#{base_form}.SocialSecurityNumber_FirstThreeNumbers[0]": data.dig('veteran', 'ssn')[0..2],
+          "#{base_form}.SocialSecurityNumber_SecondTwoNumbers[0]": data.dig('veteran', 'ssn')[3..4],
+          "#{base_form}.SocialSecurityNumber_LastFourNumbers[0]": data.dig('veteran', 'ssn')[5..9],
```
jw81, line 36: I'm prob just missing something, but can the 3rd segment of an SSN have 5 digits?
I'm just thinking that the range `5..9` is looking to pull out 5 digits and AFAIK there's only 4 digits in the 3rd segment of an SSN.

SlakrHakr: this is something that was copied from what existed so this had me wondering if i was missing something and this was trying to protect against someone with 10 digits in their ssn or something haha.

i tried it out in console and it looks like it just gets you the last 4 even though the last 5 were asked for.

but, i tried to look into it and make sure everyone always only has 9 digits in their ssn, and it sounds like that is the case, so fixed it here:
https://github.com/department-of-veterans-affairs/vets-api/pull/6812/commits/1c17d2e63fb4ef145a96eb92ad9890561941eea7

jw81 (approved):

PR #6836, opened 2021-05-10
24402 trouble shooting fsr pdf download

dillo:
## Description of change

- Adding exception handling to figure out why the FSR form download link is not working. We'll be using the dev environment to try and troubles shoot the issue. If we cannot conduct this in development due to not having proper test users we'll have to shift our focus to the production environment next.
- Adding more exception handling to critical paths
- Download the pdf inline
- 
## Original issue(s)
department-of-veterans-affairs/va.gov-team#24402


rileyanderson (approved):
LGTM

PR #6842, opened 2021-05-11
23942 allow variable date range

jperk51:
## Description of change
Allow users to request appointments with a date range larger than +/- 1 year

## Original issue(s)
department-of-veterans-affairs/va.gov-team#23942


rileyanderson (dismissed):

modules/mobile/app/controllers/mobile/v0/appointments_controller.rb
```diff
@@ -66,7 +72,7 @@ def fetch_cached_or_service(validated_params)
                                  # fetch a two year range these are later filtered by start and end date params
                                  # from the request
                                  appointments, errors = appointments_proxy.get_appointments(
-                                   start_date: one_year_ago, end_date: one_year_from_now
+                                   start_date: validated_params[:start_date], end_date: validated_params[:end_date]
```
kreek, line 75: If the params are less than a two year range here I think it won't cache enough of a range? e.g. six months ago -> today's date makes it though the `use_cache?` gate but then later if the range is today -> six months in the future there will be cached data but not for those dates. I we add min and max I think we're safe `[validated_params[:start_date], one_year_ago].min` and `[validated_params[:end_date], one_year_from_now].max`

jperk51: Make sense. My assumption was that the first time this endpoint is called for a given user session the start and end dates would be nil so we would always start with the +/- 1 year in the cache, but that will ensure it so I'll go ahead and add it.

kreek (approved):
LGTM 🚀

PR #6843, opened 2021-05-11
v2 documentation

yakschuss:
[API-6900](https://vajira.max.gov/browse/API-6900)

This PR adds the ability for our docs endpoints to serve a V2 docset

### DIFF
--- a/modules/appeals_api/app/controllers/appeals_api/metadata_controller.rb
+++ b/modules/appeals_api/app/controllers/appeals_api/metadata_controller.rb
@@ -92,7 +92,7 @@ module AppealsApi
     end
 
     def decision_reviews_versions
-      if Settings.modules_appeals_api.documentation.path_enabled_flag
+      if v2_enabled?
         [
           decision_reviews_v1.merge({ status: VERSION_STATUS[:previous] }),
           decision_reviews_v2
@@ -122,6 +122,10 @@ module AppealsApi
         path: '/services/appeals/docs/v2/decision_reviews',
         healthcheck: '/services/appeals/v2/healthcheck'
       }
+    end
+
+    def v2_enabled?
+      Settings.modules_appeals_api.documentation.path_enabled_flag
     end
   end
 end
--- a/modules/appeals_api/app/controllers/appeals_api/metadata_controller.rb
+++ b/modules/appeals_api/app/controllers/appeals_api/metadata_controller.rb
@@ -11,15 +11,7 @@ module AppealsApi
     def decision_reviews
       render json: {
         meta: {
-          versions: [
-            {
-              version: '1.0.0',
-              internal_only: true,
-              status: VERSION_STATUS[:current],
-              path: '/services/appeals/docs/v1/decision_reviews',
-              healthcheck: '/services/appeals/v1/healthcheck'
-            }
-          ]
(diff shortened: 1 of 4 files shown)


modules/appeals_api/app/controllers/appeals_api/metadata_controller.rb
```diff
@@ -98,5 +90,38 @@ def upstream_service_details(service_name, health_checker, time)
...
+          decision_reviews_v1
+        ]
+      end
+    end
+
+    def decision_reviews_v1
+      {
+        version: '1.0.0',
+        internal_only: true,
+        status: VERSION_STATUS[:current],
```
SlakrHakr, line 111: im guessing one of these (probably this one?) should be marked `previous` instead of `current`?

modules/appeals_api/app/controllers/appeals_api/metadata_controller.rb
```diff
@@ -98,5 +90,38 @@ def upstream_service_details(service_name, health_checker, time)
...
+      }
+    end
+
+    def decision_reviews_v2
+      {
+        version: '2.0.0',
+        internal_only: true,
+        status: VERSION_STATUS[:current],
+        path: '/services/appeals/docs/v2/decision_reviews',
+        healthcheck: '/services/appeals/v1/healthcheck'
```
SlakrHakr, line 123: probably want to have a v2 healthcheck url here, even if it 100% maps to the same v1 logic

modules/appeals_api/app/controllers/appeals_api/metadata_controller.rb
SlakrHakr: ah, i just noticed it overrides the status above on line 97...
hmmmm, i dont personally love this (just because it wasnt immediately obvious to me) but i do get why you did it.
whatever works 👍🏻

SlakrHakr: i almost think it would be better, imo, if this just also checked that setting down here inline.
would mean checking the feature flag in 2 places, but would read better to me personally i think

SlakrHakr (dismissed):

modules/appeals_api/app/controllers/appeals_api/metadata_controller.rb
yakschuss: hmm... i'm not a huge fan of putting conditionals inside hashes. Unless you meant something else?

modules/appeals_api/app/controllers/appeals_api/metadata_controller.rb
yakschuss: That's a good point!

nathanbwright (approved):
Looks good!

modules/appeals_api/app/controllers/appeals_api/metadata_controller.rb
yakschuss: @SlakrHakr We're going to do away with the merge anyway as per our PM, so resolving this here

nathanbwright (dismissed):

nathanbwright (dismissed):

nathanbwright (approved):

PR #6844, opened 2021-05-11
Facilities locator Covid online scheduling availability client

darkbushido:
## Description of change

A new client to talk to the mobile API in an unauthenticated.
I know the logic is unpleasant but the presence of a key is all I have.
If the key is there then covid online scheduling is available, if it's missing it isn't available.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#21734


rileyanderson (approved):

darkbushido:
> Do we need a feature flag for this?

This part no, this is just the code to get the status. the next or will need it because it uses it

PR #6845, opened 2021-05-11
[API-7020/7021] - Add ClaimsApi Veteran Identifier Endpoint

jw81:
## Description of change

Adds the V2 Veteran-Identifier endpoint for initial usability testing.  This PR also includes the work for creating the new developer docs as required in [API-7021](https://vajira.max.gov/browse/API-7021).

## Original issue(s)
[API-7020](https://vajira.max.gov/browse/API-7020) and [API-7021](https://vajira.max.gov/browse/API-7021)

## Relations
[This PR](https://github.com/department-of-veterans-affairs/devops/pull/9154) handles enabling the v2 docs in the `dev` environment.


modules/claims_api/app/controllers/claims_api/metadata_controller.rb
```diff
@@ -6,14 +6,21 @@ class MetadataController < ::ApplicationController
...
       render json: {
         meta: {
           versions: [
             {
-              version: '1.0.0',
+              version: '2.0.0',
               internal_only: false,
               status: VERSION_STATUS[:current],
+              path: '/services/claims/docs/v2/api',
+              healthcheck: '/services/claims/v2/healthcheck'
```
SlakrHakr, line 18: doesnt this need to be wrapped in a feature flag so it doesnt show in sandbox/production?

modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+  module V2
+    class VeteranIdentifierController < ApplicationController
+      skip_before_action :authenticate, only: %i[show]
+      ICN_FOR_TEST_USER = '1012667145V762142'
+
+      def show
+        headers_to_validate = %w[Authorization X-VA-SSN X-VA-First-Name X-VA-Last-Name X-VA-Birth-Date]
+        validate_headers(headers_to_validate)
+
+        render json: { icn: ICN_FOR_TEST_USER }
```
SlakrHakr, line 15: im concerned about having "icn" explicitly here. would rather is be just simply "id" in case we ever want/need to change that identifier in the future

modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require 'evss/error_middleware'
+
+module ClaimsApi
+  module V2
+    class VeteranIdentifierController < ApplicationController
+      skip_before_action :authenticate, only: %i[show]
+      ICN_FOR_TEST_USER = '1012667145V762142'
```
SlakrHakr, line 9: does it make more sense to have this value be randomly generated each call? i can sort of see benefits either way i guess.

modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+
+      def show
+        headers_to_validate = %w[Authorization X-VA-SSN X-VA-First-Name X-VA-Last-Name X-VA-Birth-Date]
+        validate_headers(headers_to_validate)
```
SlakrHakr, line 13: this isnt taking these values through the header. it needs to be a post with a post body of these 4 items.

modules/claims_api/app/controllers/claims_api/metadata_controller.rb
SlakrHakr, line 18: lets make this be `/services/benefits/docs/v2/api` and `/services/benefits/v2/healthcheck`
easiest way to do this that ive found can be seen here:
https://github.com/department-of-veterans-affairs/vets-api/blob/36672a7cc8d6e8247cf597034263a1c45cd4a55e/config/routes.rb#L354

modules/claims_api/config/routes.rb
```diff
@@ -59,6 +59,12 @@
     end
   end
 
+  namespace :v2, defaults: { format: 'json' } do
+    mount OkComputer::Engine, at: '/healthcheck'
+
+    get '/veteran-identifier', to: 'veteran_identifier#show'
```
SlakrHakr, line 65: i feel relatively strongly we should completely get away from sending anything in via headers that is not the authentication token.

that being said, i think this needs to be a post with the 4 Veteran values in the payload then.

that being said, if we're worried about it not being semantic, we could always use Google's "custom verb" concept and do something like this:
```
/veteran-identifier:find
```
or 
```
/veteran-identifier:search
```

SlakrHakr, line 65: possible a nitpick but i wonder if this should be `veteran-id` instead of `veteran-identifier`

it really depends on what we want to put in other endpoints like
```
/v2/veterans/{veteranIdentifier}/claims
```
or 
```
/v2/veterans/{veteranId}/claims
```

the short "id" is just something im more used to i guess, but either way as long as theyre consistent is my larger opinion.

SlakrHakr (changes requested):
left a few comments

modules/claims_api/app/controllers/claims_api/metadata_controller.rb
```diff
@@ -6,14 +6,21 @@ class MetadataController < ::ApplicationController
     skip_after_action :set_csrf_header
     skip_before_action(:authenticate)
 
-    def index
+    def index # rubocop:disable Metrics/MethodLength
```
SlakrHakr, line 9: could do something like Jack does here:
https://github.com/department-of-veterans-affairs/vets-api/pull/6843/files#diff-19ec423d0f85fbfb55c61ced75a8e8835420454ab372250da883cfb10e8f057cR94
to get rid of this rubocop disable

modules/claims_api/app/controllers/claims_api/metadata_controller.rb
jw81: Yeah, I threw the docs stuff in this PR branch and ran out of time yesterday to get this figured out.
Will be adding it today.

modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
jw81: That's fair.  I'll switch it.

modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
jw81: eh, we had talked about that yesterday during standup and Mark said he'd rather we just return the ICN of a test user.

modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
jw81: > this isnt taking these values through the header

Hmmm the header values are coming through properly.  Are you saying that the endpoint shouldn't be expecting headers?

modules/claims_api/config/routes.rb
jw81: I'm fine with `veteran-id`.  I'll switch it.

modules/claims_api/app/controllers/claims_api/metadata_controller.rb
jw81: Changed [here](https://github.com/department-of-veterans-affairs/vets-api/pull/6845/commits/21b6bf1d705a060801414ee164315acd8de4a636).

modules/claims_api/config/routes.rb
jw81: Switched to a POST [here](https://github.com/department-of-veterans-affairs/vets-api/pull/6845/commits/c416a5de634252f96c147b77693ae5a71af5b2e7).

modules/claims_api/app/swagger/claims_api/forms/veteran_identifier_example.json
```diff
@@ -0,0 +1,8 @@
+{
+  "attributes": {
+    "veteranSSN": "796130115",
+    "veteranFirstName": "Tamara",
+    "veteranLastName": "Ellis",
+    "veteranBirthDate": "1967-06-19"
```
SlakrHakr, line 6: can we just do:
- ssn
- firstName
- lastName
- birthdate

for these?

modules/claims_api/app/swagger/claims_api/forms/veteran_identifier_example.json
```diff
@@ -0,0 +1,8 @@
+{
+  "attributes": {
```
SlakrHakr, line 2: any chance we could at least consider not having this wrapped with both attributes and data?
im not even sure where its being wrapped by data actually...

modules/claims_api/app/swagger/claims_api/forms/veteran_identifier_example.json
jw81: Changed [here](https://github.com/department-of-veterans-affairs/vets-api/pull/6845/commits/a3aae693d14ff06137846080cac3634f9c2a8c5d).

modules/claims_api/app/swagger/claims_api/forms/veteran_identifier_example.json
jw81: This one has me stumped.
I'm not sure where the requirement for the `data` -> `attributes` nesting structure is coming from.
It's not defined in my schema definition file.

modules/claims_api/app/controllers/claims_api/metadata_controller.rb
jw81: Here's [my version](https://github.com/department-of-veterans-affairs/vets-api/pull/6845/commits/0c43cd2c7bd1a342c9c93e16c178de37003e9136).

modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
SlakrHakr: sorry for the confusion! yep! post is good 👍🏻

modules/claims_api/config/routes.rb
SlakrHakr: how do you feel about using something like `/veteran-id:search` or `/veteran-id:find` ? 

`/veteran-id` being a post just kind of sucks because its not creating a identifier or anything, but it cant be a get with the data we're passing in be in the path

jw81: Eh, it's prob due to a lack of familiarity with that pattern, but I don't really like it.  It doesn't feel very "RESTy" to me.

modules/claims_api/app/controllers/claims_api/v2/base.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+require 'claims_api/form_schemas'
+
+module ClaimsApi
+  module V2
+    class Base < ClaimsApi::V2::ApplicationController
+      private
```
SlakrHakr, line 8: should these be `protected` since they're being accessed in classes that extend this class?

modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+require 'evss/error_middleware'
+
+module ClaimsApi
+  module V2
+    class VeteranIdentifierController < ClaimsApi::V2::Base
+      skip_before_action :authenticate, only: %i[create]
+      ICN_FOR_TEST_USER = '1012667145V762142'
+      FORM_NUMBER = 'VETERAN_IDENTIFIER'
```
SlakrHakr, line 10: can we rename this to be something more generic? this isnt a form...

modules/claims_api/app/controllers/claims_api/v2/base.rb
jw81: Eh, I was just following the same pattern used in [v1](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/app/controllers/claims_api/v1/forms/base.rb#L23).

modules/claims_api/config/routes.rb
SlakrHakr: a post that doesnt create something isnt super restful either though :(

https://cloud.google.com/apis/design/custom_methods is doc on this colon concept

modules/claims_api/config/routes.rb
```diff
@@ -59,6 +59,12 @@
     end
   end
 
+  namespace :v2, defaults: { format: 'json' } do
+    mount OkComputer::Engine, at: '/healthcheck'
+
+    post '/veteran-id', to: 'veteran_identifier#create'
```
SlakrHakr, line 65: linking back to this discussion:
https://github.com/department-of-veterans-affairs/vets-api/pull/6845#discussion_r632018043

bit of a nitpick probably, but does it make more sense to call this `search` or `find` instead of `create` ?

modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
jw81: `FORM_NUMBER` is what `v0` and `v1` use for identifying what schema file name to use. ([ref](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/app/controllers/claims_api/v1/forms/base.rb#L20))
We can change this `v2` implementation to a diff name if we need to, but I err on the side of staying consistent.

modules/claims_api/config/schemas/veteran_identifier.json
```diff
@@ -0,0 +1,26 @@
...
+  "type": "object",
+  "required": [
+    "ssn",
+    "firstName",
+    "lastName",
+    "birthdate"
+  ],
+  "properties": {
+    "ssn": {
+      "type": "string"
```
SlakrHakr, line 13: should we have some validation on the string submitted here?
https://github.com/department-of-veterans-affairs/vets-api/blob/3d694a147d970392927066b244eae42135ec334d/modules/claims_api/app/controllers/claims_api/v2/veterans.rb#L21 is an example of what we could require for ssn

modules/claims_api/spec/requests/v2/veteran_identifier_request_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+        expect(response.status).to eq(422)
+      end
+    end
+  end
+
+  context 'when auth header is not present' do
+    let(:headers) { nil }
+
+    describe 'veteran identifier' do
+      it 'returns a 400 error code' do
```
SlakrHakr, line 36: shouldnt this be a 401 in this case?

modules/claims_api/spec/requests/v2/veteran_identifier_request_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+        expect(response.status).to eq(200)
+      end
+    end
+  end
+
+  context 'when body params are not present' do
+    let(:data) { nil }
+
+    describe 'veteran identifier' do
+      it 'returns a 422 error code' do
```
SlakrHakr, line 25: shouldnt this be a 400 in this case?

modules/claims_api/spec/requests/v2/veteran_identifier_request_spec.rb
jw81: If it was checking that the token was valid, then yeah it would be a 401.
But that logic is only checking that an auth header is provided.
If it's not included, that existing logic will return a 400 because it doesn't have the required header.

modules/claims_api/spec/requests/v2/veteran_identifier_request_spec.rb
jw81: JSON schema validation will raise an Unprocessable Entity exception which will return a 422 if any of the required params are missing.

modules/claims_api/config/routes.rb
SlakrHakr: this might be an idea for a route we could be agree on
https://stackoverflow.com/a/18933902/3013472

essentially the endpoint "creates a search" so in this case it would be something like
```
POST   /veteran-id/searches
```
that returns either the simple `id` we have now, or the concept shown in that stackoverflow link is to also return information about the search, so something like this maybe
```
{
  terms: {
    ssn: 'fsfds',
    firstName: 'fsdfdsf',
    lastName: 'fsdgfdgdf',
    birthdate: 'fdsgdfg'
  },
  result: {
    id: '21332432432'
  } 
}
```
^^^^ response structure might be a bit overkill though

modules/claims_api/config/routes.rb
jw81: Alright, you wore me down.
Custom method it is.  😂 
[Changed Here](https://github.com/department-of-veterans-affairs/vets-api/pull/6845/commits/b133d9f2bd7591480bc9d243e80e7cd1a4c3ac75)

modules/claims_api/config/routes.rb
jw81: Alright, you wore me down.
Custom method it is.  😂 
[Changed Here](https://github.com/department-of-veterans-affairs/vets-api/pull/6845/commits/b133d9f2bd7591480bc9d243e80e7cd1a4c3ac75)

modules/claims_api/config/schemas/veteran_identifier.json
jw81: [Added here](https://github.com/department-of-veterans-affairs/vets-api/pull/6845/commits/13e7dab71e68ae8ea80874ac5158c72df10f2997).

modules/claims_api/spec/requests/v2/veteran_identifier_request_spec.rb
SlakrHakr: sorry, what i meant was not what is currently implemented but what is expected, and a 400 for invalid input is more what i expect from a rest api. it is a "bad request"

modules/claims_api/spec/requests/v2/veteran_identifier_request_spec.rb
SlakrHakr: > But that logic is only checking that an auth header is provided.
i get it. all im saying is i wouldnt expect a 400 from an api that i didnt provide an auth token but it was required. i would expect it to tell me 401 because im "not authorized" whether i didnt provide an auth token at all or if i provided an invalid one

modules/claims_api/config/routes.rb
SlakrHakr: hahahah but :(
im sorry :( :( :( 
but thank you ❤️

modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
SlakrHakr: i get that, but v0 and v1 only had posts within the context of "forms"
id argue this should be changed for v2 because its more than just forms in this version

modules/claims_api/app/controllers/claims_api/v2/base.rb
SlakrHakr: fair enough, i think protected is "more valid" maybe? but not the end of the world as long as it functions 👍🏻

modules/claims_api/config/routes.rb
SlakrHakr: 😢 but 💯

modules/claims_api/spec/requests/v2/veteran_identifier_request_spec.rb
jw81: Hmmmm, this definitely feels like a grey area.
400 makes sense because you didn't give it the thing it needed to do it's job (part of which is checking auth).
401 also makes sense because if it couldn't do its job to make sure you're auth'd, then you're not auth'd.

modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
SlakrHakr: looking at what all this extends, im now wondering if i was just misunderstand form in this context?
the name "form" is at the core of vets-api, so i wonder if this is more like "form == request" sort of?
either way, this is probably not worth trying to rename since it is in a core class.

SlakrHakr (approved):

SlakrHakr (approved):

PR #6846, opened 2021-05-11
Refactor code and detail delegation

ElleeB:
While testing, claims and appeals hit our #show endpoint and received a `status: error` but no `code` or `detail`.

This PR delegates `AppealsApi::EvidenceSubmission` `code` and `detail` to `VBADocuments::UploadSubmission` in both the model and serializer so that when serialized, or called, the attributes inherits from the associated vba upload record.

Dependent migration PR: https://github.com/department-of-veterans-affairs/vets-api/pull/6826

Ticket: https://vajira.max.gov/browse/API-7009

### DIFF
--- a/modules/appeals_api/app/serializers/appeals_api/evidence_submission_serializer.rb
+++ b/modules/appeals_api/app/serializers/appeals_api/evidence_submission_serializer.rb
@@ -10,14 +10,16 @@ module AppealsApi
 
     attributes :id, :status, :code, :detail, :appeal_type, :appeal_id, :location, :created_at, :updated_at
 
+    delegate :status, to: :object
+    delegate :code, to: :object
+
     def id
       object.guid
     end
 
-    delegate :status, to: :object
-    delegate :code, to: :object
-
     def detail
+      return unless object.detail
+
       details = object.detail.to_s
       details = "#{details[0..MAX_DETAIL_DISPLAY_LENGTH - 1]}..." if details.length > MAX_DETAIL_DISPLAY_LENGTH
       details
--- a/modules/appeals_api/spec/factories/evidence_submissions.rb
+++ b/modules/appeals_api/spec/factories/evidence_submissions.rb
@@ -6,13 +6,18 @@ FactoryBot.define do
     guid { SecureRandom.uuid }
     association :supportable, factory: :notice_of_disagreement
     upload_submission { create(:upload_submission, guid: SecureRandom.uuid) } # set the guid to pass uniqueness check
+  end
 
-    trait :with_detail do
-      detail { SecureRandom.alphanumeric(150) }
-    end
-
-    trait :with_code do
-      code { 404 }
-    end
+  factory :evidence_submission_with_error, class: 'AppealsApi::EvidenceSubmission' do
+    sequence(:id) { |n| n }
+    guid { SecureRandom.uuid }
+    association :supportable, factory: :notice_of_disagreement
+    upload_submission {
+      create(:upload_submission,
+             guid: SecureRandom.uuid,
+             status: 'error',
+             code: '404',
+             detail: SecureRandom.alphanumeric(150))
+    }
   end
 end
--- a/modules/appeals_api/spec/serializers/evidence_submission_serializer_spec.rb
+++ b/modules/appeals_api/spec/serializers/evidence_submission_serializer_spec.rb
@@ -4,10 +4,8 @@ require 'rails_helper'
 require AppealsApi::Engine.root.join('spec', 'spec_helper.rb')
 
 describe AppealsApi::EvidenceSubmissionSerializer do
-  let(:notice_of_disagreement) { create(:notice_of_disagreement) }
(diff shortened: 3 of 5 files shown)


modules/appeals_api/app/serializers/appeals_api/evidence_submission_serializer.rb
```diff
@@ -10,17 +10,18 @@ class EvidenceSubmissionSerializer < ActiveModel::Serializer
 
     attributes :id, :status, :code, :detail, :appeal_type, :appeal_id, :location, :created_at, :updated_at
 
+    delegate :status, to: :object
+    delegate :code, to: :object
+    delegate :detail, to: :object
```
cilestin, line 15: Delegating `detail` doesn't do anything if you later supply a local `#detail` method (your method will override the delegation). You can remove this delegate.

cilestin (changes requested):
Looks good! Just one easy change

modules/appeals_api/app/serializers/appeals_api/evidence_submission_serializer.rb
```diff
@@ -10,17 +10,17 @@ class EvidenceSubmissionSerializer < ActiveModel::Serializer
...
     end
 
-    delegate :status, to: :object
-    delegate :code, to: :object
-
     def detail
       details = object.detail.to_s
       details = "#{details[0..MAX_DETAIL_DISPLAY_LENGTH - 1]}..." if details.length > MAX_DETAIL_DISPLAY_LENGTH
-      details
+      details.presence
```
yakschuss, line 23: What edge case was this protecting against?

modules/appeals_api/spec/factories/evidence_submissions.rb
```diff
@@ -6,13 +6,18 @@
     guid { SecureRandom.uuid }
     association :supportable, factory: :notice_of_disagreement
     upload_submission { create(:upload_submission, guid: SecureRandom.uuid) } # set the guid to pass uniqueness check
+  end
 
-    trait :with_detail do
```
yakschuss, line 10: This trait is failing the Migration PR. You may want to remove this in it's own PR, and get that merged ahead of time.

modules/appeals_api/spec/serializers/evidence_submission_serializer_spec.rb
```diff
@@ -31,24 +25,44 @@
...
-  it 'includes :code' do
-    expect(rendered_hash[:code]).to eq evidence_submission.code
+  it 'includes :updated_at' do
+    expect(rendered_hash[:updated_at]).to eq evidence_submission.updated_at
   end
 
-  it "truncates :detail value if longer than #{described_class::MAX_DETAIL_DISPLAY_LENGTH}" do
-    max_length_plus_ellipses = described_class::MAX_DETAIL_DISPLAY_LENGTH + 3
-    expect(rendered_hash[:detail].length).to eq(max_length_plus_ellipses)
+  it 'does not include any extra attributes' do
```
yakschuss, line 32: Hmmm.. I'm not sure how I feel about this test. On the one hand, it lets us know when anything changes, but it's also not really handling any code path.

modules/appeals_api/app/serializers/appeals_api/evidence_submission_serializer.rb
ElleeB: Coercing empty string to nil. May not be necessary. Let me know what you think.

modules/appeals_api/spec/factories/evidence_submissions.rb
ElleeB: I was going to merge this and then the migration. Should I not wait?

modules/appeals_api/app/serializers/appeals_api/evidence_submission_serializer.rb
yakschuss: I think JSON converts nil to “”, not null. Could be wrong.

yakschuss: Nevermind, I'm wrong.

modules/appeals_api/spec/factories/evidence_submissions.rb
yakschuss: I think since the migration is just removing columns, that should be fine.

modules/appeals_api/app/serializers/appeals_api/evidence_submission_serializer.rb
nathanbwright, line 23: Personally I think that a guard statement would make sense here (maybe something like `return unless object.detail`). Otherwise you have to read the entire method and remember how `String#presence` works to know the outcome for what should be the majority of cases. Thoughts?

cilestin (dismissed):

nathanbwright (dismissed):
🚀

cilestin (approved):

cilestin (approved):

PR #6849, opened 2021-05-11
API-6895 HLR V2 structure

cilestin:
## Description of change
Adds the beginnings of pdf generation for v2 (Apr 2021) of the 200996 form. Fills in the fields we care to fill in using pdf forms while skipping other, more manual data placement. 

## Original issue(s)
https://vajira.max.gov/browse/API-6895
https://vajira.max.gov/browse/API-6896

* Not indented to be 100% complete for all fields
* Worked with the existing v1 HLR schema as far as it could to ensure the fields were mapped properly, but is expected that the schema will change for v2 and things will need to be updated at that time.
* Is missing anything that requires manual autosize text box placement like veteran & rep name
* Has some TODOs that will become tickets later
* ~Missing informal conference time selection, since the v1 schema can't handle its new use case~


ElleeB (dismissed):
LGTM

modules/appeals_api/app/models/appeals_api/higher_level_review.rb
```diff
@@ -132,11 +132,23 @@ def veteran_phone_number
...
+    def veteran_homeless?
+      form_data&.dig('data', 'attributes', 'veteran', 'homeless')
+    end
+
     # 11. E-MAIL ADDRESS
     def email
       veteran&.dig('emailAddressText').to_s.strip
     end
 
+    def email_v2
```
yakschuss, line 148: I'm not sure that I know how to solve this, but I feel like it stinks that we need to do this.

modules/appeals_api/app/services/appeals_api/pdf_construction/higher_level_review/pages/additional_issues_table.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module AppealsApi
+  module PdfConstruction
+    module HigherLevelReview
+      module Pages
+        class AdditionalIssuesTable
+          MAX_ISSUES_ON_FIRST_PAGE = 13
```
yakschuss, line 8: Is the only difference between the already existing AdditionalIssues class and this the MAX_ISSUES_ON_FIRST_PAGE?

Maybe we can move this to Structure::MAX_ISSUES_ON_FIRST_PAGE and have one file?

modules/appeals_api/app/services/appeals_api/pdf_construction/higher_level_review/pages/additional_issues_table.rb
```diff
@@ -0,0 +1,41 @@
...
+          attr_accessor :pdf, :form_data
+
+          def extra_issues?
+            form_data.contestable_issues.count > MAX_ISSUES_ON_FIRST_PAGE
+          end
+
+          def extra_issues_text
+            form_data.contestable_issues.drop(MAX_ISSUES_ON_FIRST_PAGE).map do |issue|
+              "Issue: #{issue['attributes']['issue']} - Decision Date: #{issue['attributes']['decisionDate']}"
+            end.join("\n")
```
yakschuss, line 35: This part is different as well, with the lack of reverse, but i'm not sure it necessitates an entirely new class.

yakschuss: Perhaps we can decide to reverse at initialization

modules/appeals_api/app/models/appeals_api/higher_level_review.rb
cilestin: Very much agreed. I also played with passing a version into the `email` and found that more distasteful :(

modules/appeals_api/app/services/appeals_api/pdf_construction/higher_level_review/pages/additional_issues_table.rb
cilestin: It is for now, but there's a TODO later in the file to change this to output a table instead of just text output, so I felt like having it be its own file made sense.

modules/appeals_api/app/services/appeals_api/pdf_construction/higher_level_review/pages/additional_issues_table.rb
cilestin: Same as my previous comment here - the eventual desire is for this file to become a tabular output instead of a list, I just didn't want to tackle that output in this PR.

nathanbwright (approved):
Looks like a good initial implementation to me!

PR #6851, opened 2021-05-11
New VAOS - Record VCR Cassettes for `clinics` end-point

ajmagdub:
## Description of change

Update VCR cassettes to reflect data retrieved from the hosted new VAOS service. Refactored systems service to reduce the coupling between Rails session and service calls. Added camel schema checks.

Also refactored get available slots method in the systems service class to reduce session coupling.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#24581

updated rspecs.


modules/vaos/app/controllers/vaos/v2/slots_controller.rb
```diff
@@ -4,7 +4,10 @@ module VAOS
   module V2
     class SlotsController < VAOS::V0::BaseController
       def index
-        response = systems_service.get_available_slots(slots_params)
+        response = systems_service.get_available_slots(location_id: location_id,
+                                                       clinic_id: clinic_id,
+                                                       start_dt: start_dt,
+                                                       end_dt: end_dt)
```
cferris32, line 10: I was wondering why these params got split out to be individually passed in? (if it works it works, I've just had our code checks raise concern when I pass in a large amount of method params before which is why I'm wondering)

cferris32:
LGTM with the addition of the one comment I left which was more of a general curiosity.

modules/vaos/app/controllers/vaos/v2/slots_controller.rb
ajmagdub: they were split to uncouple the http origin from the service layer. The controller knows that the data comes from an http session. The service layer should not need to know that the data came from ActionController::Parameters.  

In Ruby named parameters can also be passed in as a Hash, in addition to being passed in individually.

cferris32: Gotcha that all makes sense, thanks for the insight

cferris32 (approved):

PR #6853, opened 2021-05-12
updated table  formatting

gpbowman-git:
updating table formatting

## Original issue(s)
https://vajira.max.gov/browse/API-6861


cshupp1 (dismissed):

mviterna (approved):
deleteme is missing an _ but probably not worth changing if you've done all your testing

PR #6856, opened 2021-05-12
Api 6841 cache charon responses

lukemarkwordtlibertyits:
# Description

Token Validation V2 should cache charon responses instead of calling charon ever time.

# To ensure functionality

- [x] Unit Tests Passing
- [x] Regression Tests Passing
- [x] Debug to ensure caching is working


crolarlibertyva (dismissed):
Confirmed caching works for a valid token and an unauthorized one, once the responses are acquired in an initial call to charon.

dbrown-va (approved):

PR #6857, opened 2021-05-12
[22732] adds message log when failing to look up edipi

bramleyjl:
## Description of change
This change adds a new, temporary, Sentry logging call in the case of an EVSS `getCurrentInfo` call that is to be performed without the needed EDIPI. We suspect that the failure to produce an EDIPI is due to this method (`request_headers`) being called on a user that has either not been authenticated to LOA3 or is missing their mhv_icn.

## Original issue(s)

This PR should not change any existing functionality.

### DIFF
--- a/lib/evss/vso_search/service.rb
+++ b/lib/evss/vso_search/service.rb
@@ -31,6 +31,13 @@ module EVSS
       def request_headers(additional_headers)
         ssn = additional_headers.key?('va_eauth_pnid') ? additional_headers['va_eauth_pnid'] : @user.ssn
         edipi = additional_headers.key?('va_eauth_dodedipnid') ? additional_headers['va_eauth_dodedipnid'] : @user.edipi
+        if edipi.nil?
+          user_info = {
+            loa3: @user.loa3?,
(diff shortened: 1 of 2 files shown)


thilton-oddball (approved):
LGTM

lib/evss/vso_search/service.rb
```diff
@@ -31,6 +31,13 @@ def get_current_info(additional_headers = {})
       def request_headers(additional_headers)
         ssn = additional_headers.key?('va_eauth_pnid') ? additional_headers['va_eauth_pnid'] : @user.ssn
         edipi = additional_headers.key?('va_eauth_dodedipnid') ? additional_headers['va_eauth_dodedipnid'] : @user.edipi
+        if edipi.nil?
+          user_info = {
+            loa3: @user.loa3?,
+            mhv_icn: !@user.mhv_icn.nil?
```
bosawt, line 37: The goal is to just get boolean values for both of these, right? You can also do `@user.mhv_icn.present?`, but if it's temporary I guess it doesn't matter

bosawt (approved):

PR #6860, opened 2021-05-12
21734 show covid 19 online scheduling to the fe

darkbushido:
## Description of change

Adding a new attribute to the facilities module.
This is a temporary measure until lighthouse has the covid online scheduling availability in their database.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#21734

* New Feature Flag `facilities_locator_mobile_covid_online_scheduling`


modules/facilities_api/app/controllers/facilities_api/v1/va_controller.rb
```diff
@@ -11,6 +11,12 @@ class V1::VAController < ApplicationController
     def index
       api_results = api.get_facilities(lighthouse_params)
 
+      if Flipper.enabled?(:facilities_locator_mobile_covid_online_scheduling) && covid_mobile_params?
+        api_results.each do |api_result|
+          api_result.tmp_covid_online_scheduling = mobile_api_get_by_id(api_result.id)
```
mpelzsherman, line 16: Does `mobile_api_get_by_id` return a boolean? Do we need to cast it or something?

modules/facilities_api/app/controllers/facilities_api/v1/va_controller.rb
```diff
@@ -51,5 +56,17 @@ def serializer
     def resource_path(options)
       v1_va_index_url(options)
     end
+
+    def mobile_api
+      FacilitiesApi::V1::MobileCovid::Client.new
+    end
+
+    def mobile_api_get_by_id(id)
```
mpelzsherman, line 64: Using a more self-documenting name here would improve code readablility, e.g. `get_ covid_online_scheduling_available_from_mobile_api`.

mpelzsherman:
Have we considered using the same field name and data structure we will eventually use for Lighthouse so we don't need to change our FE code once the Lighthouse data becomes available?

Is there some reason the "tmp" attribute needs to persist after Lighthouse is working?

modules/facilities_api/app/controllers/facilities_api/v1/va_controller.rb
darkbushido: its aliasing `mobile_api.direct_booking_eligibility_criteria_by_id(id).covid_online_scheduling_available?`

darkbushido (commented):
> Have we considered using the same field name and data structure we will eventually use for Lighthouse so we don't need to change our FE code once the Lighthouse data becomes available?
> 
> Is there some reason the "tmp" attribute needs to persist after Lighthouse is working?

I don't think we should. `detailedServices` is an array that should contain the covid structure but currently isn't always populated and I don't want to try to fake a structure when its not there.

modules/facilities_api/app/controllers/facilities_api/v1/va_controller.rb
mpelzsherman: Not sure I follow but this is a minor concern.

darkbushido: I was just making a shorter method name since the other one is descriptive. I can get rid of it but it's just an alias to a long method chain.

thilton-oddball (approved):
This all looks good!

PR #6866, opened 2021-05-12
Mobile appointments use clinic address if available

kreek:
## Description of change
Some clinics have addresses that are different than their associated facility. Updates the VA appointments adapter to use the  clinic address first if available.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#24865

Tested on a review instance


jperk51 (approved):
LGTM

PR #6874, opened 2021-05-13
[API-6461] Enable VBMS Access on successful POA submittals

hinzed1127:
If the contents of a POA request match requirements, enable VBMS access so reps can have full tracking access

## Original issue(s)
https://vajira.max.gov/browse/API-6461

### DIFF
--- a/modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
@@ -47,6 +47,7 @@ module ClaimsApi
 
           # This job only occurs when a Veteran submits a PoA request, they are not required to submit a document.
           ClaimsApi::PoaUpdater.perform_async(power_of_attorney.id) unless header_request?
+          ClaimsApi::VBMSUpdater.perform_async(power_of_attorney.id)
           data = power_of_attorney.form_data
           ClaimsApi::PoaFormBuilderJob.perform_async(power_of_attorney.id) if data['signatures'].present?
 
@@ -119,6 +120,10 @@ module ClaimsApi
         end
 
         private
+
+        def enable_vmbs_access?
+          form_attributes['recordConsent'] && (form_attributes['consentLimits'].nil? || form_attributes['consentLimits'].empty?)
+        end
 
         def current_poa_begin_date
           return nil if current_poa.try(:begin_date).blank?
--- a/modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
@@ -47,7 +47,7 @@ module ClaimsApi
 
           # This job only occurs when a Veteran submits a PoA request, they are not required to submit a document.
           ClaimsApi::PoaUpdater.perform_async(power_of_attorney.id) unless header_request?
-          ClaimsApi::VBMSUpdater.perform_async(power_of_attorney.id) if enable_vmbs_access?
+          ClaimsApi::VBMSUpdater.perform_async(power_of_attorney.id, target_veteran) if enable_vmbs_access?
           data = power_of_attorney.form_data
           ClaimsApi::PoaFormBuilderJob.perform_async(power_of_attorney.id) if data['signatures'].present?
 
@@ -122,7 +122,7 @@ module ClaimsApi
         private
 
         def enable_vmbs_access?
-          form_attributes['recordConsent'] && (form_attributes['consentLimits'].nil? || form_attributes['consentLimits'].empty?)
+          form_attributes['recordConsent'] && form_attributes['consentLimits'].blank?
         end
 
         def current_poa_begin_date
--- a/modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
@@ -47,7 +47,7 @@ module ClaimsApi
 
           # This job only occurs when a Veteran submits a PoA request, they are not required to submit a document.
           ClaimsApi::PoaUpdater.perform_async(power_of_attorney.id) unless header_request?
-          ClaimsApi::VBMSUpdater.perform_async(power_of_attorney.id, target_veteran) if enable_vmbs_access?
+          ClaimsApi::VBMSUpdater.perform_async(power_of_attorney.id, target_veteran.participant_id) if enable_vmbs_access?
           data = power_of_attorney.form_data
           ClaimsApi::PoaFormBuilderJob.perform_async(power_of_attorney.id) if data['signatures'].present?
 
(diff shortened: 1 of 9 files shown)


modules/claims_api/app/workers/claims_api/vmbs_updater.rb
```diff
@@ -0,0 +1,30 @@
...
+        include Sidekiq::Worker
+    
+        def perform(power_of_attorney_id)
+          poa_form = ClaimsApi::PowerOfAttorney.find(power_of_attorney_id)
+          service = BGS::Services.new(
+            external_uid: poa_form.external_uid,
+            external_key: poa_form.external_key
+          )
+          service.vet_record.update_birls_record(
+            participant_id: form_data.form_data['participant_vet_id'] || target_veteran.participant_id,
```
hinzed1127, line 23: I believe this is the right id in this case, but I've been burned on that before, so calling it out here for validation

modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
```diff
@@ -120,6 +121,10 @@ def validate
 
         private
 
+        def enable_vmbs_access?
```
SlakrHakr, line 124: is this supposed to be called on line 50 above? i dont see it being called anywhere in this file currently.

modules/claims_api/app/workers/claims_api/vmbs_updater.rb
SlakrHakr: seems strange to use a Veteran id from the request payload i guess. should we not just use the the participant_id of the target_veteran always?

modules/claims_api/app/workers/claims_api/vmbs_updater.rb
```diff
@@ -0,0 +1,30 @@
...
+      class VBMSUpdater
+        include Sidekiq::Worker
+    
+        def perform(power_of_attorney_id)
+          poa_form = ClaimsApi::PowerOfAttorney.find(power_of_attorney_id)
+          service = BGS::Services.new(
+            external_uid: poa_form.external_uid,
+            external_key: poa_form.external_key
+          )
+          service.vet_record.update_birls_record(
```
SlakrHakr, line 22: i might have missed something but i thought we were needing to use the "update_poa_access" method of BGS?
https://vajira.max.gov/browse/API-6461?focusedCommentId=890592&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-890592

if this is also calling "update_birls_record" can we not just change the existing poa_updater job?
https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/app/workers/claims_api/poa_updater.rb#L16

SlakrHakr (changes requested):
couple of concerns

modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
hinzed1127: oof yeah I accidentally had stuff in v0 at first and must have forgotten it moving things over

modules/claims_api/app/workers/claims_api/vmbs_updater.rb
hinzed1127: more copy-paste nonsense. I even updated it with the right arguments but the wrong method 🤦‍♂️ correcting that too

modules/claims_api/app/workers/claims_api/vmbs_updater.rb
hinzed1127: yeah you're right, I don't even think we accept those arguments for POA typically, anyway. looks like just ITF. removing the form data

modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
```diff
@@ -120,6 +121,10 @@ def validate
...
+          form_attributes['recordConsent'] && (form_attributes['consentLimits'].nil? || form_attributes['consentLimits'].empty?)
```
SlakrHakr, line 125: this
```
(form_attributes['consentLimits'].nil? || form_attributes['consentLimits'].empty?)
```
can be replaced with this
```
form_attributes['consentLimits'].blank?
```

its a rails method (opposite of `.present?`) that checks for
- nil
- empty string
- empty array
really anything that is "not something"

modules/claims_api/app/workers/claims_api/vmbs_updater.rb
```diff
@@ -0,0 +1,30 @@
...
+        include Sidekiq::Worker
+    
+        def perform(power_of_attorney_id)
+          poa_form = ClaimsApi::PowerOfAttorney.find(power_of_attorney_id)
+          service = BGS::Services.new(
+            external_uid: poa_form.external_uid,
+            external_key: poa_form.external_key
+          )
+          service.corporate_update.update_poa_access(
+            participant_id: target_veteran.participant_id,
```
SlakrHakr, line 23: i dont think target_veteran is accessible within a background job context?

modules/claims_api/app/workers/claims_api/vmbs_updater.rb
```diff
@@ -0,0 +1,30 @@
...
+            poa_code: form_data.form_data.dig('serviceOrganization', 'poaCode'),
```
SlakrHakr, line 24: is this supposed to be `poa_form.form_data`?

SlakrHakr (changes requested):
lets get some specs of this vbms_updater job class as well

modules/claims_api/app/workers/claims_api/vmbs_updater.rb
hinzed1127: ahh yeah, good catch

modules/claims_api/app/workers/claims_api/vmbs_updater.rb
hinzed1127: I think you're right, passing into the method instead

SlakrHakr: im pretty sure this won't work either, because of all the issues we've had around sidekiq. Sidekiq only allows super simple arguments.

SlakrHakr: it'll most likely have to be accessed through the database record some way. stored with the auth_headers perhaps.

hinzed1127:
The only thing I can think of maybe adding here (or later on) is something in the POA model to add some kind of indicator in something we store to confirm that VBMS access has been granted to a given POA, similar to how the POA updater job [updates the POA form `status` property](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/app/workers/claims_api/poa_updater.rb#L21-L22)

SlakrHakr:
This will require the following pr for the bgs-ext gem to go in first:
https://github.com/department-of-veterans-affairs/bgs-ext/pull/60

SlakrHakr (dismissed):

SlakrHakr (dismissed):

thilton-oddball (approved):
Looks good! Ship it!

SlakrHakr (approved):

PR #6876, opened 2021-05-13
Api 6874 cache smart launch responses

esubocz-va:
## Description of change

Cache smart launch responses for token token validation. Edit session model to not require uuid, as system tokens do not have uuids.

## Original issue(s)
https://vajira.max.gov/browse/API-6874

update to sessions model required changes to classes outside of token validation - updates have been made to prevent a breaking change.

Used `redis-commander` to confirm caching for regular, cc, and ssoi tokens. All unit tests pass.


crolarlibertyva (dismissed):
LGTM and Confirmed desired caching through debug traces.

app/models/session.rb
```diff
@@ -19,8 +19,9 @@ class Session < Common::RedisStore
   attribute :ssoe_transactionid
   attribute :profile
   attribute :charon_response
+  attribute :launch
+  attribute :uuid
```
beaugrantham-va, line 23: It might be worth a comment here, indicating that `uuid` is optional since a session may be for a system (vs a user). I worry that someone may revert this change without as thorough of a review... breaking us.

app/controllers/openid_application_controller.rb
```diff
@@ -54,18 +54,30 @@ def authenticate_token
...
 
+    return false if @session.uuid.nil?
+
     @current_user = OpenidUser.find(@session.uuid)
   end
 
   def populate_payload_for_launch_scope
-    launch = fetch_smart_launch_context
-    token.payload[:launch] = base64_json?(launch) ? JSON.parse(Base64.decode64(launch)) : { patient: launch }
+    if @session.launch.nil?
```
lukemarkwordtlibertyits, line 63: This is going to break client_credential tokens with launch in the payload because the session never gets instantiated in the first place.

lukemarkwordtlibertyits (changes requested):

app/controllers/openid_application_controller.rb
lukemarkwordtlibertyits: Here's how to get the breaking token
```sh
docker run \
  vasdvp/lighthouse-auth-utils:latest auth-cc \
  --client-id="$CC_CLIENT_ID" \
  --client-secret="$CC_CLIENT_SECRET" \
  --authorization-url="$SANDBOX_AUTHORIZATION_URL" \
  --scope="launch" \
  --launch="123V456" | jq .access_token | tr -d '\"'
```

esubocz-va: 👍 added cc session builder for launch scope to match that of launch/patient scope.

app/controllers/openid_application_controller.rb
```diff
@@ -54,18 +54,33 @@ def authenticate_token
...
-    token.payload[:launch] = base64_json?(launch) ? JSON.parse(Base64.decode64(launch)) : { patient: launch }
+    @session = Session.find(token)
+    if @session.nil?
+      ttl = token.payload['exp'] - Time.current.utc.to_i
+      launch = fetch_smart_launch_context
+      @session = build_cc_session(ttl, launch)
+      @session.save
+    end
+    token.payload[:launch] =
+      base64_json?(@session.launch) ? JSON.parse(Base64.decode64(@session.launch)) : { patient: @session.launch }
```
lukemarkwordtlibertyits, line 71: SSOI Tokens will make it to here and break. The launch is not stored in their session.

lukemarkwordtlibertyits (changes requested):

app/controllers/openid_application_controller.rb
esubocz-va: added support for storing launch in session for ssoi tokens.

lukemarkwordtlibertyits (approved):
Confirmed caching and the following test cases:

```
```

app/controllers/openid_application_controller.rb
```diff
@@ -54,18 +54,34 @@ def authenticate_token
...
   def populate_payload_for_launch_patient_scope
-    launch = fetch_smart_launch_context
-    token.payload[:icn] = launch
-    token.payload[:launch] = { patient: launch } unless launch.nil?
+    analyze_redis_launch_context
+    token.payload[:icn] = @session.launch
+    token.payload[:launch] = { patient: @session.launch } unless @session.launch.nil?
+  end
+
+  def analyze_redis_launch_context
```
dbrown-va, line 74: I think this method is worth a comment highlighting how the session is already in place for SSOi but not for CC.

dbrown-va (approved):
Thank you for adding the session based tests.

thilton-oddball (approved):
Looks good!

bosawt (approved):
Identity related files look good!

PR #6877, opened 2021-05-13
Set Status of 674 to Manual

kathleencrawford:
## Description of change

During integration testing with our partners, a couple issues have come up and we have been asked to set the status to `MANUAL_VAGOV` for certain scenarios.  We were already doing that for some use cases already, so this PR adds two more checks: 
 - when submitting a 674 (school age child attending college)
 - when adding a spouse but the marriage type is not `CEREMONIAL`

## Original issue(s)
department-of-veterans-affairs/va.gov-team#24624

This is an update to existing functionality, and just sets the status for certain use cases.  No new settings or logging have been added.

- [x] Local testing
- [x] Staging testing planned
- [x] Customer integration testing planned


thilton-oddball (approved):
LGTM

PR #6900, opened 2021-05-13
Updated local login redirects to use port 3002.

U-DON:
## Original issue(s)


U-DON:
@thilton-oddball: This might not be needed, so don't look too much into the specs yet. Basically, this is predicated on whether or not the site served at 3001 is enough to enable a locally authenticated experience, but still investigating to confirm.

PR #6924, opened 2021-05-14
[WIP] [6835] removes remaining user.mpi calls

bramleyjl:
## Description of change
This change removes remaining `User.mpi` calls and minimizes `User.mpi_profile` calls, leaving only a few for unit tests and rake tasks.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#6835


jimiray (approved):

spec/models/form_profile_spec.rb
```diff
@@ -9,7 +9,7 @@
   let(:user) { build(:user, :loa3) }
 
   before do
-    user.mpi.profile.suffix = 'Jr.'
+    user.set_mpi_profile('suffix', 'Jr.')
```
bosawt, line 12: We can set this up up in a better way, I can help show you how

spec/request/letters_request_spec.rb
```diff
@@ -169,8 +169,8 @@
       end
 
       before do
-        user.mpi.profile.edipi = '1005079124'
-        user.mpi.profile.participant_id = '600036159'
+        user.set_mpi_profile('edipi', '1005079124')
```
bosawt, line 172: I can help show how to set this test up in a better way

bosawt (changes requested):
Mostly looks good, I can help you fix the specs so we don't need a function to override MPI values in User

bosawt (approved):
Looks good!

PR #6929, opened 2021-05-17
API-6901 - update hlr docs

nathanbwright:
## Description of change
This is the first of several PRs to update our docs for HLR v2 (split to keep the line count down).

## Original issue(s)
https://vajira.max.gov/browse/API-6901


modules/appeals_api/spec/fixtures/valid_200996_v2.json
```diff
@@ -0,0 +1,91 @@
...
+        "timezone": "America/Chicago"
+      },
+      "ssoSsocOptIn": true,
+      "hlrCreateInformalConferenceContact": [
+        "veteran"
+      ],
+      "informalConferenceTime": [
+        "800-1200 ET"
+      ],
+      "informalConferenceRep": {
```
ElleeB, line 24: I think we changed this to `representative` in the updated schema

modules/appeals_api/spec/fixtures/valid_200996_v2.json
```diff
@@ -0,0 +1,91 @@
...
+        "phone": {
+          "countryCode": "34",
+          "areaCode": "555",
+          "phoneNumber": "8001111"
+        },
+        "emailAddressText": "josie@example.com",
+        "timezone": "America/Chicago"
+      },
+      "ssoSsocOptIn": true,
+      "hlrCreateInformalConferenceContact": [
```
ElleeB, line 18: And I believe this should be `informalConferenceContact`

modules/appeals_api/spec/fixtures/valid_200996_v2.json
```diff
@@ -0,0 +1,91 @@
+{
+  "data": {
+    "type": "higherLevelReview",
+    "attributes": {
+      "informalConference": true,
+      "benefitType": "compensation",
+      "veteran": {
```
ElleeB, line 7: I think we need to include `homeless` here within the `veteran` object

cilestin (approved):
These look good! There will probably be some changes necessary later as we go through the new schema more carefully, but I think keeping these in sync won't be too difficult from here on out.

🚀

PR #6937, opened 2021-05-17
Mobile claims overview caching update

jperk51:
## Description of change
Updated claims overview logic so the service is requested when the cache is empty then it sets the cache. This matches the pattern for appointments now. Should also resolve an error that's been popping up in sentry of trying to filter nil.


kreek (approved):
LGTM 🚀

PR #6938, opened 2021-05-17
API-7126: Previous/active POA still not working as expected in sandbox

SlakrHakr:
## Description of change
Sorting issue with previous poa codes list.

## Original issue(s)
https://vajira.max.gov/browse/API-7126

Tested logic change in sandbox first to understand and verify.
Added spec to test for the correct sort.

### DIFF
--- a/modules/veteran/spec/models/veteran/user_spec.rb
+++ b/modules/veteran/spec/models/veteran/user_spec.rb
@@ -9,7 +9,7 @@ describe Veteran::User do
     it 'initializes from a user' do
       VCR.use_cassette('bgs/claimant_web_service/find_poa_by_participant_id') do
         allow_any_instance_of(BGS::OrgWebService).to receive(:find_poa_history_by_ptcpnt_id)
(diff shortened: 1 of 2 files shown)


modules/veteran/spec/models/veteran/user_spec.rb
```diff
@@ -9,7 +9,7 @@
     it 'initializes from a user' do
       VCR.use_cassette('bgs/claimant_web_service/find_poa_by_participant_id') do
         allow_any_instance_of(BGS::OrgWebService).to receive(:find_poa_history_by_ptcpnt_id)
-          .and_return({ person_poa_history: { person_poa: [{ begin_date: Time.zone.now, legacy_poa_cd: '033' }] } })
+          .and_return({ person_poa_history: { person_poa: [{ begin_dt: Time.zone.now, legacy_poa_cd: '033' }] } })
```
SlakrHakr, line 12: i spent an embarrassing amount of time thoroughly confused why the sort wasn't working locally before i discovered this key was incorrect :(

thilton-oddball (approved):
Looks good

PR #6941, opened 2021-05-17
[API-7226] - Adds 404 logic to ClaimsApi veteran id endpoint prototype

jw81:
## Description of change

Adds logic to the new `ClaimsApi` V2 veteran-id endpoint to return a 404 if the received data does not match the test veteran's data.

## Original issue(s)
[API-7226](https://vajira.max.gov/browse/API-7226)

### DIFF
--- a/modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
@@ -14,8 +14,9 @@ module ClaimsApi
         raise ::Common::Exceptions::Unauthorized if request.headers['Authorization'].blank?
 
         validate_request
+        veteran = find_veteran(params)
 
-        render json: { id: ICN_FOR_TEST_USER }
+        render json: { id: veteran[:id] }
       end
 
       private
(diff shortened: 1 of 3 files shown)


modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
```diff
@@ -44,6 +45,34 @@ def validate_birthdate!(date_str)
...
+      def find_veteran(params)
+        test_veteran_data = {
+          id: ICN_FOR_TEST_USER,
+          ssn: '796130115',
+          firstName: 'Tamara',
+          lastName: 'Ellis',
+          birthdate: '1967-06-19'
+        }
+
+        unless params[:ssn] == test_veteran_data[:ssn]
```
SlakrHakr, line 58: more out of curiosity, why do `casecmp?` for the names down below but not here?

SlakrHakr (approved):

modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
jw81: Because this is a social security number.
There shouldn't be any case sensitivity issues with digits.  (unless I'm missing something, haha 😂 )

PR #6944, opened 2021-05-18
API-6902 Add deprecated key to v1 hlr swagger routes

ElleeB:
In preparation for rollout of v2.

Add deprecation key:value pair to trigger the fading and crossing out of these v1 routes in our developer portal docs.

### DIFF
--- a/modules/appeals_api/app/swagger/appeals_api/v1/higher_level_reviews_controller_swagger.rb
+++ b/modules/appeals_api/app/swagger/appeals_api/v1/higher_level_reviews_controller_swagger.rb
@@ -5,6 +5,7 @@ class AppealsApi::V1::HigherLevelReviewsControllerSwagger
 
   OBJ = :object
   HLR_TAG = ['Higher-Level Reviews'].freeze
+  DOC_VISIBILITY_SETTING = Settings.modules_appeals_api.documentation.path_enabled_flag
 
   read_file = ->(path) { File.read(AppealsApi::Engine.root.join(*path)) }
   read_json = ->(path) { JSON.parse(read_file.call(path)) }
@@ -68,6 +69,7 @@ class AppealsApi::V1::HigherLevelReviewsControllerSwagger
 
   swagger_path '/higher_level_reviews' do
     operation :post, tags: HLR_TAG do
+      key :deprecated, DOC_VISIBILITY_SETTING
       key :operationId, 'postHigherLevelReviews'
       key :summary, 'Creates a new Higher-Level Review.'
       desc = 'Submits a Decision Review request of type *Higher-Level Review*. This endpoint is the same as ' \
@@ -84,6 +86,7 @@ class AppealsApi::V1::HigherLevelReviewsControllerSwagger
 
   swagger_path '/higher_level_reviews/{uuid}' do
     operation :get, tags: HLR_TAG do
+      key :deprecated, DOC_VISIBILITY_SETTING
       key :operationId, 'getHigherLevelReview'
       key :summary, 'Shows a specific Higher-Level Review. (a.k.a. the Show endpoint)'
       key :description, 'Returns all of the data associated with a specific Higher-Level Review.'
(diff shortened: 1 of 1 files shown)


modules/appeals_api/app/swagger/appeals_api/v1/higher_level_reviews_controller_swagger.rb
```diff
@@ -68,6 +68,7 @@ class AppealsApi::V1::HigherLevelReviewsControllerSwagger
 
   swagger_path '/higher_level_reviews' do
     operation :post, tags: HLR_TAG do
+      key :deprecated, true
```
nathanbwright, line 71: What do you think about adding some logic so these only show up as deprecated in dev?

ElleeB: Ah! Ok. I'm going to attempt now but may reach out with questions!

ElleeB:
**Question:**
Test for this via new`v2/docs_controller_spec`?

modules/appeals_api/app/swagger/appeals_api/v1/higher_level_reviews_controller_swagger.rb
```diff
@@ -3,6 +3,8 @@
 class AppealsApi::V1::HigherLevelReviewsControllerSwagger
   include Swagger::Blocks
 
+  env_deprecation_state = Rails.env == 'dev' ? true : false
```
yakschuss, line 6: can we change this to `!Settings.modules_appeals_api.documentation.path_enabled?`

yakschuss: Unless you have a better way of removing the hard coded assumption about which environment we're deprecating in.

yakschuss: path enabled returns a true or false, which we're using to turn v2 on in staging/dev/sandbox. 

If we're enabling v2 (true return value), then we're also deprecating v1 paths (false return value) - so we'd just invert the flag we're already using to determine the enabling of the documentation.

yakschuss: I don't know if @nathanbwright was referring specifically to dev rather than `lower` environments. Have to ask him

ElleeB: Perhaps we need to decide as a team how we want to roll this out.

modules/appeals_api/app/swagger/appeals_api/v1/higher_level_reviews_controller_swagger.rb
ElleeB: need more clarification

nathanbwright: Sorry I should have been clearer... I was talking about only showing these as deprecated in the lower envs.

So if you had something like this at the top of the file:

`LOWER_ENV_DOCS = !Settings.modules_appeals_api.documentation.path_enabled_flag`

then each of these deprecated places could be something like this:

`key :deprecated, true if LOWER_ENV_DOCS`

The const could named whatever you think makes sense.

ElleeB: OK. I think I got it. I was completely forgetting the fact that I was working on a v1 controller. Good times.

modules/appeals_api/app/swagger/appeals_api/v1/higher_level_reviews_controller_swagger.rb
cilestin: Not related to the discussion, but - reminder that the Rails environment is always `production` on all our servers. `Settings.vsp_environment` is used to determine the server environment.

modules/appeals_api/app/swagger/appeals_api/v1/higher_level_reviews_controller_swagger.rb
```diff
@@ -68,6 +68,7 @@ class AppealsApi::V1::HigherLevelReviewsControllerSwagger
 
   swagger_path '/higher_level_reviews' do
     operation :post, tags: HLR_TAG do
+      key :deprecated, Settings.modules_appeals_api.documentation.path_enabled_flag
```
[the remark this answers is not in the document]
ElleeB: ya

cilestin (approved):

nathanbwright (approved):
👍

PR #6945, opened 2021-05-18
stub for pg_stat_statements

LindseySaari:
## Description of change
This stubs the call to pg_stat_statements used from within PgHero (for the PgHero jobs) in the test suite. Developers that are running the native flow for vets-api will run into the issue `PgHero::NotEnabled: Query stats not enabled` and will have to enable some [postgres config settings](https://github.com/ankane/pghero/blob/master/guides/Query-Stats.md#installation), otherwise this test will fail locally. It's confirmed that pg_stat_statements are enabled in all of our envs ([dev](https://github.com/department-of-veterans-affairs/devops/blob/badb0d95782c9ae91084247995287f478f08c2df/terraform/environments/dsva-vagov-dev/main.tf), [staging](https://github.com/department-of-veterans-affairs/devops/blob/badb0d95782c9ae91084247995287f478f08c2df/terraform/environments/dsva-vagov-staging/main.tf) [prod](https://github.com/department-of-veterans-affairs/devops/blob/4e0904f8fd98a8cd5a3e4fcf7c10046cfb473d46/terraform/environments/dsva-vagov-prod/main.tf)) via terraform, so unless needing to specifically work with pg_stat_statements, it doesn't need to be enabled.

If developers need to work with pg_stat_statements locally, it's enabled via [docker](https://github.com/department-of-veterans-affairs/vets-api/blob/master/docker-compose.test.yml#L5) in the test env or it can be enabled via [postgres config settings](https://github.com/ankane/pghero/blob/master/guides/Query-Stats.md#installation) locally.

This PR also updates the native.md documentation for instructions on how to enable pg_stat_statements locally

## Original issue(s)
department-of-veterans-affairs/va.gov-team#24842


docs/setup/native.md
```diff
@@ -54,6 +54,26 @@ Vets API requires:
...
 1. Run `bin/setup` to setup the database and start the server.
 
+### pg_stat_statements
+If you have trouble enabling query stats from the PgHero dashboard, try enabling it manually
+
+Add the lines below to your main postgresql.conf file
+
+On Mac it should be located somewhere similiar to the following:
+
+`/Users/[your username]/Library/Application Support/Postgres/var-12/postgresql.conf`
```
ericboehs, line 64:
```suggestion
`~/Library/Application Support/Postgres/var-12/postgresql.conf`
```

ericboehs: This will only be the location if you're using Postgres.conf.

docs/setup/native.md
```diff
@@ -54,6 +54,26 @@ Vets API requires:
...
+shared_preload_libraries = 'pg_stat_statements'
+pg_stat_statements.track = all
+pg_stat_statements.max = 10000
+track_activity_query_size = 2048
+```
+Then from the command line:
+`psql -U postgres`
+
+Next from the psql console:
+`CREATE extension pg_stat_statements;`
```
ericboehs, line 75: We should have a migration to do this instead of having non-standard procedures to follow.

LindseySaari: I can remove this piece I think. We do have a migration already https://github.com/department-of-veterans-affairs/vets-api/blob/master/db/migrate/20210507122840_add_stats_extension.rb

LindseySaari: Fixed ✔️

thilton-oddball (approved):
Looks good!

PR #6950, opened 2021-05-18
Faraday major upgrade

rileyanderson:
## Description of change
As a more cautious approach to update our production gems, the upgrade path was split into three sub-steps. It was discussed to break out each gem into a ticket individually, as the risk level increases. This ticket details the upgrade of the faraday gem (production group) to its latest major version.

This PR updates `faraday` `faraday_middleware` and `sentry-raven` as these all need to be updated concurrently due to dependency constraints  

## Original issue(s)
department-of-veterans-affairs/va.gov-team#24430
department-of-veterans-affairs/va.gov-team#24420
department-of-veterans-affairs/va.gov-team#24418


thilton-oddball (approved):
This all looks good. Nice work!

PR #6951, opened 2021-05-18
Self-host Lint/Security GHA Job

ericboehs:


ericboehs:
Closing in favor of trying to move everything TO cloud runners (rather than away).

PR #6957, opened 2021-05-18
Mobile multi image upload param reading fix

jperk51:
## Description of change
Fixed issue with accessing params for claims file upload. Since single file is form data and multi-image is json I split the common code into a private method and passed in the params accessing them the correct way for each endpoint. Also fixed a small typo so now multi-image will accept `files` for the array of images as it should have. 

Updated tests to mirror these changes.


kreek (approved):
LGTM 🚀

PR #6970, opened 2021-05-19
Converting the type parameter to separate Routes

darkbushido:
## Description of change

Converting the `type` param into separate routes.

We scrub all of the params from logs so it's difficult to tell what `type` of search we are handling.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#22183


thilton-oddball (approved):
Looks good

PR #6971, opened 2021-05-19
Disabling cache_classes in the test env

darkbushido:
## Description of change

The current rails template disables this if we are installing spring
This will make repeated test runs faster if your using spring but should
not effect your first run

[Rails test.rb Generator](https://github.com/rails/rails/blob/main/railties/lib/rails/generators/rails/app/templates/config/environments/test.rb.tt#L12-L17)

[Discord Thread](https://dsva.slack.com/archives/CBU0KDSB1/p1620229594363700)


ericboehs:
I tried this and it failed the build too. Are you still working on this?

ericboehs:
Please re-open if you'd like to continue on this.

PR #6972, opened 2021-05-19
New VAOS facilities endpoint complete with tests

grossb1:
## Description of change

This work is related to updating vets-api with a new set of end-points that will leverage both MAS 2.0 and the new Mobile Facilities Service.  Specifically, this ticket address the new `/facilities` end-point.

## Original issue(s)

Both request and service specs have been written to cover this change.


cferris32 (approved):
LGTM

modules/vaos/spec/services/v2/mobile_facility_service_spec.rb
```diff
@@ -31,4 +31,58 @@
...
+    end
+
+    context 'with a facility id and children true' do
+      it 'returns a configuration' do
+        VCR.use_cassette('vaos/v2/mobile_facility_service/get_facilities_with_children_200',
+                         match_requests_on: %i[method uri]) do
+          params = {
+            ids: %w[688],
+            children: true
+          }
```
saneshark, line 56: typically something like this would be in a let block within the context

modules/vaos/spec/services/v2/mobile_facility_service_spec.rb
```diff
@@ -31,4 +31,58 @@
...
+      end
+    end
+
+    context 'when the upstream server returns a 500' do
+      it 'raises a backend exception' do
+        VCR.use_cassette('vaos/v2/mobile_facility/get_facilities_500',
+                         match_requests_on: %i[method uri]) do
+          params = {
+            ids: %w[688]
+          }
```
saneshark, line 80: `let(:params) { { ids: %w[688] } }`

saneshark (approved):
LGTM, but left some nitpick comments

thilton-oddball (approved):
LGTM

PR #6979, opened 2021-05-20
Rails 6.1 upgrade

LindseySaari:
## Description of change
This PR upgrades to the latest Rails version 6.1. There were various deprecation and breaking changes introduced between the 6.0 and 6.1 upgrade. Listed below are the changes that were addressed in the PR:

* ActiveModel Error Objects
    * New in 6.1, ActiveModel Error is now an object. ex: my_object.errors is now an array of objects vs a hash
    * Various sytax updates for custom error handling across the entire repository. 
    * Example errors[:some_attribute] << vs errors.add(:some_attribute]
* Fixture Paths
    * Removed the prepended files/ directory from file_fixture_uploads calls in the code. If fixtures are placed within the /files directory, the prepended /files directory is auto configured. 
    * app/controllers/concerns/exception_handling.rb has to use the explicit to_h method instead of to_hash due to object.errors now being an array

* Active Storage Blob migrations
    * These are new for Rails 6.1 and will need to go out with the gem upgrade (currently failing danger)
    * Some of the specs were failing with `NoMethodError (undefined method 'service_name' for #<ActiveStorage::Blob:0x0000558b5edfbeb8>)`. db migrations were needed for the ActiveStorage blobs (specifically to add the service_name column) Used the solution suggested [https://github.com/rails/rails/issues/40895#issuecomment-748811171](https://)

* Private headers method rename
    * Issue with private method `'headers'` in various controllers. Renamed the methods to `request_headers`
    * Believe that the private headers method was overlapping/overwriting the controller built in ‘headers’ method ex: headers['X-Session-Expiration'] 

* Environment config deprecation change
    * config/environments/test.rb 
    * DEPRECATION WARNING:` action_view.raise_on_missing_translations` is deprecated and will be removed in Rails 6.2. Set `i18n.raise_on_missing_translations` instead.

## Original issue(s)


modules/appeals_api/spec/controllers/v1/contestable_issues_controller_spec.rb
```diff
@@ -8,10 +8,8 @@
     subject { described_class.new.send(:get_contestable_issues_from_caseflow).body }
 
     before do
-      allow_any_instance_of(described_class).to receive(:headers).and_return({
-                                                                               'X-VA-SSN' => '872958715',
-                                                                               'X-VA-Receipt-Date' => '2019-12-01'
-                                                                             })
+      expected_headers = { 'X-VA-SSN' => '872958715', 'X-VA-Receipt-Date' => '2019-12-01' }
+      allow_any_instance_of(described_class).to receive(:request_headers).and_return(expected_headers)
```
LindseySaari, line 12: This was split up to avoid rubocop line length

app/controllers/concerns/exception_handling.rb
```diff
@@ -80,7 +80,7 @@ def report_original_exception(exception)
   end
 
   def report_mapped_exception(exception, va_exception)
-    extra = exception.respond_to?(:errors) ? { errors: exception.errors.map(&:to_hash) } : {}
```
LindseySaari, line 83: errors are no longer a hash (now an object) so the explicit method call

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements_controller.rb
```diff
@@ -93,9 +93,9 @@ def render_model_errors
   end
 
   def model_errors_to_json_api
-    errors = @notice_of_disagreement.errors.to_h.map do |source, data|
```
LindseySaari, line 96: Need to call explicit to_h method here too

modules/vaos/spec/models/cancel_form_spec.rb
```diff
@@ -10,7 +10,7 @@
 
     it 'validates presence of required attributes' do
       expect(subject).not_to be_valid
-      expect(subject.errors.keys).to contain_exactly(:appointment_time, :cancel_code, :facility_id)
```
LindseySaari, line 13: keys is deprecated, instead use attribute_names

LindseySaari:
@cbtkc This is ready for review, it's just red because there's /db changes in the same PR

thilton-oddball (commented):
This all looks good to me. The only two comments I have are that it looks like there are 59 rubocop violations that are all auto-correctable and then I wasn't sure if we want to split out the migrations into a separate PR or if they both have to be in this same PR.

kreek (approved):
Looks good for the mobile api change

LindseySaari:
> This all looks good to me. The only two comments I have are that it looks like there are 59 rubocop violations that are all auto-correctable and then I wasn't sure if we want to split out the migrations into a separate PR or if they both have to be in this same PR.

Yes, I think that the migrations/schema changes will have to go out in this same PR because they're new in 6.1 and there was a failing spec related to the migration changes too. I will take a look at the rubocop violations.

cbtkc (approved):
Looks good!

Gemfile
```diff
@@ -37,15 +37,15 @@ gem 'sass-rails', '>= 6'
 # Anchored versions, do not change
 gem 'puma', '~> 4.3.8'
 gem 'puma-plugin-statsd', '~> 1.2.1'
-gem 'rails', '~> 6.0.3'
+gem 'rails', '~> 6.1.3.2'
```
ericboehs, line 40:
```suggestion
gem 'rails', '~> 6.1.3'
```

I'd recommend locking just to the patch.

app/models/concerns/redis_form.rb
```diff
@@ -50,7 +50,10 @@ def save
 
   def form_matches_schema
     if form.present?
-      errors[:form].concat(JSON::Validator.fully_validate(VetsJsonSchema::SCHEMAS[self.class::FORM_ID], parsed_form))
+      validations = JSON::Validator.fully_validate(VetsJsonSchema::SCHEMAS[self.class::FORM_ID], parsed_form)
+      validations.each do |v|
+        errors.add(:form, v.to_s)
+      end
```
ericboehs, line 56: Before the method would return `errors[:form]`. Now it's returning the enumerable. I don't know if this will affect anything but here's how you should be able to do what you're asking without doing the gross `v = foo; f.each {}; v` dance.

```suggestion
      JSON::Validator.fully_validate(VetsJsonSchema::SCHEMAS[self.class::FORM_ID], parsed_form).tap do |v|
        errors.add(:form, v.to_s)
      end
```

app/models/saved_claim.rb
```diff
@@ -79,13 +79,16 @@ def form_is_string
...
-    errors[:form] << 'must be a json string' unless form_is_string
+    errors.add(:form, :invalid_format, message: 'must be a json string') unless form_is_string
   end
 
   def form_matches_schema
     return unless form_is_string
 
-    errors[:form].concat(JSON::Validator.fully_validate(VetsJsonSchema::SCHEMAS[self.class::FORM], parsed_form))
+    validations = JSON::Validator.fully_validate(VetsJsonSchema::SCHEMAS[self.class::FORM], parsed_form)
+    validations.each do |v|
```
ericboehs, line 89: `tap`?

app/models/concerns/temp_form_validation.rb
```diff
@@ -18,7 +18,10 @@ def parsed_form
 
   def form_matches_schema
     if form.present?
-      errors[:form].concat(JSON::Validator.fully_validate(VetsJsonSchema::SCHEMAS[self.class::FORM_ID], parsed_form))
+      validations = JSON::Validator.fully_validate(VetsJsonSchema::SCHEMAS[self.class::FORM_ID], parsed_form)
+      validations.each do |v|
```
ericboehs, line 22: `tap`?

app/models/saved_claim/education_benefits/va_1990s.rb
```diff
@@ -7,6 +7,9 @@ class SavedClaim::EducationBenefits::VA1990s < SavedClaim::EducationBenefits
   def form_matches_schema
     return unless form_is_string
 
-    errors[:form].concat(JSON::Validator.fully_validate(VetsJsonSchema::SCHEMAS['VRRAP'], parsed_form))
+    validations = JSON::Validator.fully_validate(VetsJsonSchema::SCHEMAS['VRRAP'], parsed_form)
+    validations.each do |v|
```
ericboehs, line 11: `tap`?

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/base_contestable_issues_controller.rb
```diff
@@ -35,8 +35,14 @@ def index
 
   attr_reader :caseflow_response, :backend_service_exception
 
+  def request_headers
+    EXPECTED_HEADERS.reduce({}) do |hash, key|
+      hash.merge(key => request.headers[key])
+    end.compact
```
ericboehs, line 41: I think `.slice` would work here instead of this craziness.

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements_controller.rb
ericboehs: Does this still need to happen?

ericboehs (commented):
It would be good to go through [RailsDiff](https://railsdiff.org/6.0.0/6.1.3.2) and make sure we have all the changes add.

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements_controller.rb
LindseySaari: We're now iterating over an array of error objects

LindseySaari: See line 97

app/models/concerns/redis_form.rb
```diff
@@ -50,7 +50,9 @@ def save
 
   def form_matches_schema
     if form.present?
-      errors[:form].concat(JSON::Validator.fully_validate(VetsJsonSchema::SCHEMAS[self.class::FORM_ID], parsed_form))
+      JSON::Validator.fully_validate(VetsJsonSchema::SCHEMAS[self.class::FORM_ID], parsed_form).each do |v|
```
ericboehs, line 53: Each still returns an iterator instead of the original object.

```suggestion
      JSON::Validator.fully_validate(VetsJsonSchema::SCHEMAS[self.class::FORM_ID], parsed_form).tap do |v|
```

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/base_contestable_issues_controller.rb
```diff
@@ -35,8 +35,12 @@ def index
 
   attr_reader :caseflow_response, :backend_service_exception
 
+  def request_headers
+    EXPECTED_HEADERS.index_with { |key| request.headers[key] }.compact
```
ericboehs, line 39: Still feel like this should work:
```suggestion
    request.headers.slice EXPECTED_HEADERS
```

LindseySaari:
@ericboehs I had to revert the .tap changes, which caused test failures

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/base_contestable_issues_controller.rb
LindseySaari: This change caused some test failures @ericboehs, I reverted back to the original

ericboehs:
> @ericboehs I had to revert the .tap changes, which caused test failures

Dang. :(

LindseySaari:
@cbtkc @thilton-oddball @rileyanderson Can we get a re-review on this please? Also, still waiting on a re-review from @department-of-veterans-affairs/vfs-vaos

rileyanderson (approved):
LGTM 🚀

cbtkc (approved):
This looks good to me. Great work

LindseySaari:
@department-of-veterans-affairs/vfs-vaos can we get a review on this again please! Thanks! @kreek

PR #6981, opened 2021-05-20
Clean up lint checks in GHA

ericboehs:
- Alphabetize most yaml keys (leaving `name:` at the top, and `user` before `pass`).
- Removed unneeded system packages install for linting/security
- Moved some ENV vars into the specific step that needed it
- Removed redis (wasn't being used)
- Fix deprecation warning for `comment_on_pr` in Test Results action
- Looked into merging Danger workflow. We could but then everything would run twice because we'd need to add `pull_request` and `push` to our `on:` block.
- Move result publishing to its own job that runs on GH cloud runners since our self-hosted runners have SSL issues due to MITM cert.
- Remove Code Climate in favor of SimpleCov (GH artifacts and coverage report).

### DIFF
--- a/.github/workflows/code_checks.yml
+++ b/.github/workflows/code_checks.yml
@@ -35,7 +35,6 @@ jobs:
     name: Test
     env:
       BUNDLE_ENTERPRISE__CONTRIBSYS__COM: ${{ secrets.BUNDLE_ENTERPRISE__CONTRIBSYS__COM }}
-      CC_TEST_REPORTER_ID: ${{secrets.CC_TEST_REPORTER_ID}}
       CI: true
       DATABASE_URL: "postgis://vets-api:postgres@127.0.0.1:5432/vets_api_test?pool=4"
       DISABLE_BOOTSNAP: true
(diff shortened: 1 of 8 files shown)


.github/workflows/code_checks.yml
```diff
@@ -27,79 +22,108 @@ jobs:
...
         run: bundle exec parallel_test -n 14 -e 'bundle exec rake db:reset'
 
-      - name: Setup Code Climate Reporting
+      - name: Run Specs
         run: |
-          curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
-          chmod +x ./cc-test-reporter
-          ./cc-test-reporter before-build
+          bundle exec parallel_rspec spec/ modules -n 14 -o '--color --tty'
+          sed -i 's/"line"/"covered_percent"/g' coverage/.last_run.json
```
ericboehs, line 76: Had to add this line as simplecov changed it's last run json syntax.

LindseySaari:
Are we able to move code climate? I'm wondering if any product folks use it for the QASP report?

.github/workflows/code_checks.yml
```diff
@@ -27,79 +22,108 @@ jobs:
...
-      ImageOS: ubuntu20
-      NODE_EXTRA_CA_CERTS: '/etc/ssl/certs/VA-Internal-S2-RCA1-v1.cer.pem' # Needed for self-hosted runners
+    runs-on: asg
 
     services:
       postgres:
         image: postgis/postgis:11-2.5
         env:
-          POSTGRES_USER: vets-api
           POSTGRES_DB: vets-api_test
```
rileyanderson, line 53: Should we rename this `vets_api_test` so it's consistent with  line 45?

ericboehs: Turns out this line does nothing.

thilton-oddball (approved):
LGTM

PR #6987, opened 2021-05-20
Add last name and last four social to hlr v2 form

ElleeB:
Add Veteran last name/SSN stamp to top of new HLR form while leaving the original as is.

Suggest creating another ticket at some point to look into a stamping module.


modules/appeals_api/spec/services/appeals_api/pdf_construction/notice_of_disagreement/v1/form_data_spec.rb
```diff
@@ -105,16 +105,6 @@ module V1
             it { expect(form_data.contestable_issues).to eq(notice_of_disagreement.form_data.dig('included')) }
           end
 
-          describe '#stamp_text' do
```
cilestin, line 108: Even if we removed `#stamp_text` we should still move this test somewhere instead of removing it outright, IMO.

ElleeB: Yeah. I tried to move it but was unexpectedly stumped. Still working on it.

modules/appeals_api/app/services/concerns/appeals_api/stamp_pdf.rb
```diff
@@ -0,0 +1,69 @@
...
+      end
+
+      private
+
+      def stamp_text(veteran_last_name, veteran_ssn)
+        "#{veteran_last_name.truncate(35)} - #{veteran_ssn.last(4)}"
+      end
+
+      def v1_hlr?
+        # TODO: remove this method and the conditional logic reliant upon it once V1 HLR goes bye bye
```
yakschuss, line 63: It could be a year or more before we actually remove the first version of HLR.. IMO this is a pretty tight dependency on a specific version of a specific form, rather than a generic stamping module. 

Can you help me understand what problem this refactor is solving?

ElleeB: We can't stamp the the v1 hlr with last name and social, like we're stamping nod and v2 hlr.

ElleeB: It's generic for hlr v2, nod v1, and eventually for sc

yakschuss: I'm trying to understand how the introduction of this pattern solves that issue

ElleeB: I really don't mind distributing all of this in the individual appeal structures, but there was a comment in the code stating that upon v2 hlr, the expectation was to refactor stamping to be type-of-appeal independent, so I went down that road. I can rewind and go down the other (each pdf structure holds onto it's own stamp_text) if everyone agrees.

yakschuss: As it stands right now, the Structure file of each form is what contains the unique business logic of that specific form. I think it might make sense to include this module in the structure file rather than the generator, and just not call the stamp method from the module inside v1 hlr, instead of doing the conditional logic inside the module itself.

ElleeB: Got it.

yakschuss (approved):

PR #6991, opened 2021-05-21
[24702] Adding fallback to BGS service when user.relationships is cal…

bosawt:
…led and MPI response has no relationship data

## Description of change
This change adds a call to the BGS dependent service to get information about a user's dependents when `user.relationships` is called and there is no relationship data from the saved MPI response (first we attempt to query MPI for relationship data, if that doesn't exist, we query BGS for relationship data)

## Original issue(s)
department-of-veterans-affairs/va.gov-team/issues/24702

- This shouldn't appear to have any functional impact on production at this point
- To test this, I added the line: `relationship: user.relationships` in the `profile` function of `models/profile.rb`, and logged in with user `vets.gov.user+36@gmail.com` (this user has a BGS mock with relationship data).
- Once that line is added, we can sign in and inspect the response from the call to `user`, there should be a `relationship` field with populated data

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -450,7 +450,7 @@ class User < Common::RedisStore
   end
 
   def relationships
-    @relationships ||= mpi_profile_relationships.map { |relationship| relationship_hash(relationship) }
+    @relationships ||= get_relationships_array
   end
 
   def mpi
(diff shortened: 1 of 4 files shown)


joeniquette (commented):
I reviewed this as best as I could, also scanned for "bugs" and "code smells" with sonarqube, no issues detected. I expect a backend engineer from our team will take a pass at this soon.

jimiray (dismissed):

thilton-oddball (approved):
LGTM

jimiray (approved):

PR #6993, opened 2021-05-21
hide zero value attachment sizes

Teesh:
## Description of change
Added a check to SM attachment serializer to not display zero attachment size

## Original issue(s)
department-of-veterans-affairs/va.gov-team#25032

This is a patch to not show additional info that always returns zero from MHV. Eventually, this issue should be escalated to MHV for a fix at the source and this patch should be reverted

Ran the code in review instance and tested POSTs through Postman to make new messages with attachments


modules/mobile/app/serializers/mobile/v0/attachment_serializer.rb
```diff
@@ -8,7 +8,7 @@ class AttachmentSerializer < ActiveModel::Serializer
       attribute :id
       attribute :name
       attribute :message_id
-      attribute :attachment_size
+      attribute :attachment_size, if: -> { object.attachment_size.positive? }
```
narin, line 11: Looks good but we should add some unit tests.  Also, is there any scenario where `attachment_size` could be null/nil?

Teesh: If it was nil, positive would still test false and we would still not be included. Is that not what we want to do?

narin: I think if the value is `nil` and you try to run `.positive?` on it, it'll throw `NoMethodError (undefined method 'positive?' for nil:NilClass)`

narin: I think if you do this it'll handle a `nil` scenario gracefully: `object.attachment_size&.positive?`.

Source: https://stackoverflow.com/questions/12425465/ruby-check-if-nil-before-calling-method

modules/mobile/spec/request/messages_request_spec.rb
```diff
@@ -137,6 +137,7 @@
           expect(response.body).to be_a(String)
           expect(JSON.parse(response.body)['data']['attributes']['subject']).to eq('CI Run')
           expect(JSON.parse(response.body)['data']['attributes']['body']).to eq('Continuous Integration')
+          expect(JSON.parse(response.body)['data']['attributes']['attachment_size']).to be_positive.or be_nil
```
narin, line 140: If you take a look at a sample response below, you'll see that the attachments are returned in the `included` array.  So this probably has to be something like `expect(JSON.parse(response.body)['included'][0]['attributes']['attachment_size']).to be_positive.or be_nil`
```
{
    "data": {
        "id": "1958827",
        "type": "messages",
        "attributes": {
            "message_id": 1958827,
            "category": "TEST_RESULTS",
            "subject": "Your lab results",
            "body": "test with attachment 2",
            "attachment": true,
            "sent_date": "2021-06-03T17:26:52.000Z",
            "sender_id": 523757,
            "sender_name": "FREEMAN, MELVIN  V",
            "recipient_id": 1918816,
            "recipient_name": "VA Flagship mobile applications interface 2_DAYT29",
            "read_receipt": null
        },
        "relationships": {
            "attachments": {
                "data": [
                    {
                        "id": "1958826",
                        "type": "attachments"
                    }
                ]
            }
        },
        "links": {
            "self": "http://2811b521c3f13d58f79b6ec3c8735f1e-api.review.vetsgov-internal/mobile/v0/messaging/health/messages/1958827"
        }
    },
    "included": [
        {
            "id": "1958826",
            "type": "attachments",
            "attributes": {
                "name": "bob.jpg",
                "message_id": 1958827,
            },
            "links": {
                "download": "http://2811b521c3f13d58f79b6ec3c8735f1e-api.review.vetsgov-internal/mobile/v0/messaging/health/messages/1958827/attachments/1958826"
            }
        }
    ]
}
```

narin (approved):
LGTM 🚀

PR #6995, opened 2021-05-21
[24862] makes User mpi and mpi_profile methods private

bramleyjl:
## Description of change

This change removes the last few calls of the User model's `mpi` method and makes it private. It also returns the `mpi_profile` method to private as well and cleans up some testing surrounding these two changes.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#24862

Affected unit specs update and manual login/logout workflow.

### DIFF
--- a/app/controllers/v0/profile/personal_informations_controller.rb
+++ b/app/controllers/v0/profile/personal_informations_controller.rb
@@ -7,22 +7,13 @@ module V0
 
       # Fetches the personal information for the current user.
       # Namely their gender and birth date.
-      #
-      # @return [Response] Sample response.body:
-      #   {
-      #     "data" => {
-      #       "id"         => "",
-      #       "type"       => "mvi_models_mvi_profiles",
-      #       "attributes" => {
-      #         "gender"     => "M",
-      #         "birth_date" => "1949-03-04"
-      #       }
-      #     }
-      #   }
-      #
       def show
-        response = @current_user.mpi.profile
-
+        response = OpenStruct.new({
+                                    'id': @current_user.account_uuid,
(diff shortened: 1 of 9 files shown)


cbtkc (dismissed):
Looks good

LindseySaari:
Looks like some linting issues. Can take another 👀  once that's resolved

patrickvinograd:
Please hold on merging this until the mobile team has a chance to review. I will follow up on it this morning.

patrickvinograd:
I think we're all clear from mobile side, though I'd advocate for merging this after today's deploy so we have some time to observe it in staging.

ericboehs (approved):

app/controllers/v0/profile/personal_informations_controller.rb
```diff
@@ -7,22 +7,13 @@ class PersonalInformationsController < ApplicationController
...
-      #       }
-      #     }
-      #   }
-      #
       def show
-        response = @current_user.mpi.profile
-
+        response = OpenStruct.new({
+                                    'id': @current_user.account_uuid,
+                                    'type': 'mvi_models_mvi_profiles',
```
bosawt, line 13: Setting `type` here doesn't do anything, it gets overridden when serialized (it'll be `open_structs` to match the type of the class that we serialize). There are ways we can fix it, but I don't think the frontend cares what the `type` is here, so I think we should just remove this line and let the type be `open_structs`)

bosawt (approved):
I logged in a user and made a curl to the `personal_inormations` route, everything looks good! 

I have a couple quick comments that are really just cleaning up the code a bit.

PR #6996, opened 2021-05-21
API-6614: Ensure upstream_healthcheck does not timeout for any upstream dependencies

SlakrHakr:
## Description of change
Attempts to ensure all healthcheck calls return a response instead of timing out. This helps us always know which upstream dependency is having issues.

## Original issue(s)
https://vajira.max.gov/browse/API-6614

WSDL path for the VBMS healthcheck was added due to base url check not working in production (though it worked fine in sandbox).
WSDL path addition allows healthcheck to work in both sandbox and production.

### DIFF
--- a/modules/claims_api/config/initializers/okcomputer.rb
+++ b/modules/claims_api/config/initializers/okcomputer.rb
@@ -76,7 +76,9 @@ end
 
 class VbmsCheck < BaseCheck
   def check
-    response = Faraday::Connection.new.get(Settings.vbms.url) { |request| request.options.timeout = 20 }
+    connection = Faraday::Connection.new
+    connection.options.timeout = 10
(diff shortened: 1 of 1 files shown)


modules/claims_api/config/initializers/okcomputer.rb
```diff
@@ -76,7 +76,9 @@ def name
 
 class VbmsCheck < BaseCheck
   def check
-    response = Faraday::Connection.new.get(Settings.vbms.url) { |request| request.options.timeout = 20 }
+    connection = Faraday::Connection.new
+    connection.options.timeout = 10
```
SlakrHakr, line 80: moved to this structure for setting the timeout because the previous method (line 79 that is removed here) was not actually forcing a timeout like it should.

jw81 (approved):

PR #6999, opened 2021-05-21
Fixes nil get/set case in mobile caching library

patrickvinograd:
## Description of change
Fixes bug with setting/getting nil values in mobile module's redis caching mixin. 

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/25101


thilton-oddball (approved):
LGTM

app/models/concerns/redis_caching.rb
```diff
@@ -19,8 +19,12 @@ def get_cached(key)
     end
 
     def set_cached(key, data)
-      @redis.set(key, data.to_json)
-      @redis.expire(key, @redis_ttl)
+      if data
+        @redis.set(key, data.to_json)
+        @redis.expire(key, @redis_ttl)
+      else
+        Rails.logger.debug('Nil data for redis caching')
```
patrickvinograd, line 26: I'm not sure we'll ever see `debug` level logging in staging/prod like environments if we are trying to chase this down. info might be better. And I would change the message to something like "Attempted to set nil data in redis cache". But not a huge deal, the meaning is clear enough.

thilton-oddball (approved):

PR #7000, opened 2021-05-22
Handle empty payment dates on FSR form pre-fill

kevinmirc:
## Description of change

BGS payments will sometimes have a `:payment_date` of `nil`. This would cause the sort function to raise an error.

Update the filtering of payments to exclude any payments with a `nil` `:payment_date`.

## Summary
- [x] Add test suite for existing functionality in `DebtManagementCenter::PaymentsService`
- [x] Exclude payments that don't contain a `:payment_date` from the filtered lists of payments (in `#compensation_and_pension` and `#education`)
- [x] Add yard documentation specifying what `#compensation_and_pension` and `#education` returns.

## Original issue(s)
- _(ticket coming soon)_
- sentry: http://sentry.vfs.va.gov/organizations/vsp/issues/26275/?environment=production&project=3 
- slack: https://dsva.slack.com/archives/CBU0KDSB1/p1621536803181100

This error would prevent end-users from being able to start the FSR form. Because this function is called on `InProgressForm` creation, the entire form would fail to load, and a generic error was shown to the user.

We still have no confirmation that a `nil` `:payment_date` means it's pending, but when viewing production data we were able to confirm that it **can** be `nil`, and that only a single payment in this data sample was `nil`. This suggests that the payment could be a pending one. In the future, we may want to consider using the `:scheduled_date` instead (which may never be `nil`).

If another team/product needs the full list of payments, additional methods should be implemented, and these existing methods should have `_processed` appended to their name.


thilton-oddball (approved):
LGTM

PR #7006, opened 2021-05-24
undefined method `mpi_icn' for #<ClaimsApi::Veteran:0x000055f83e6cf160>

SlakrHakr:
## Description of change
Undefined method issue cropped up over the weekend, after a deploy on Friday it looks like.
I believe this was the culprit: https://github.com/department-of-veterans-affairs/vets-api/pull/6924

## Original issue(s)
http://sentry.vfs.va.gov/share/issue/bbb51c4a609043a5a9712a61067cb30a/

Ideally this should go out with today's deploy because it has rendered one of our endpoints completely broken :(

### DIFF
--- a/modules/claims_api/spec/models/veteran/service/user_spec.rb
+++ b/modules/claims_api/spec/models/veteran/service/user_spec.rb
@@ -1,0 +1,55 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe Veteran::User do
+  context 'initialization' do
+    let(:user) do
+      ClaimsApi::Veteran.new(
+        uuid: '123456789',
+        ssn: '123456789',
+        first_name: 'Firstname',
+        last_name: 'Lastname',
+        va_profile: ClaimsApi::Veteran.build_profile('1970-01-01'),
+        last_signed_in: Time.now.utc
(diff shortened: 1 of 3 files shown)


modules/claims_api/spec/models/veteran/service/user_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    let(:user) do
+      ClaimsApi::Veteran.new(
+        uuid: '123456789',
+        ssn: '123456789',
+        first_name: 'Firstname',
+        last_name: 'Lastname',
+        va_profile: ClaimsApi::Veteran.build_profile('1970-01-01'),
+        last_signed_in: Time.now.utc
+      )
+    end
```
SlakrHakr, line 16: i pretty much just copied the test from here:
https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/veteran/spec/models/veteran/user_spec.rb#L7

but send in our ClaimsApi::Veteran instance instead.

i was able to recreate the undefined method error that way and then see it resolve when fixed.

bramleyjl (approved):

jw81 (approved):

PR #7008, opened 2021-05-24
Feature flag for notifications triggered for connected accounts in lower environments

bradenshipley:
## Description of change

This PR prevents notifications being sent out when a user connects or disconnects an application from/to their VA Profile Connected Accounts. This feature was enabled for the lower environments for testing purposes, and is now causing rate limit warnings in Okta when large batches of zombie accounts/apps are being deleted in Okta. There's no need for emails to be sent in staging/testing so this feature can be disabled outside of Production without issue.

## Original issue(s)
slack thread discussing the need for a feature flag in lower environments
https://lighthouseva.slack.com/archives/C01SABWSGRY/p1621863254056700
devops pr adding environment variable https://github.com/department-of-veterans-affairs/devops/pull/9222

Unit test has been added to `notification_service_spec.rb` to test for early return.

### DIFF
--- a/config/features.yml
+++ b/config/features.yml
@@ -46,6 +46,10 @@ features:
     actor_type: user
     description: Feature gates the certificate of eligibility application
     enable_in_development: true
+  connected_applications_notif_service:
+    actor_type: cookie_id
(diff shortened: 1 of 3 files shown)


charleystran (approved):
LGTM

PR #7009, opened 2021-05-24
[22732] adds check for empty string to EVSS service request_headers

bramleyjl:
## Description of change
Fix to catch empty string `va_eauth_dodedipnid` headers that were being incorrectly used as a reference for a user's EDIPI; this is causing [500 errors](http://sentry.vfs.va.gov/organizations/vsp/issues/31328/?environment=production&project=3&query=is%3Aunresolved+message%3A%22EducationForm%3A%3AProcess10203EVSSError%22&statsPeriod=14d) to occur within EVSS when calls are made to their service with a blank EDIPI.

The `presence` method should check both for the existence of the appropriate header within the hash as well as the existence of a truthy value for that header.

## Original issue(s)

Some Sentry logging added to the EVSS service has been maintained; if the related errors stop occurring it will be removed.

### DIFF
--- a/lib/evss/vso_search/service.rb
+++ b/lib/evss/vso_search/service.rb
@@ -36,7 +36,7 @@ module EVSS
             loa3: @user.loa3?,
             mhv_icn: !@user.mhv_icn.nil?,
             edipi_type: additional_headers['va_eauth_dodedipnid'].class,
-            edipi_length: additional_headers['va_eauth_dodedipnid'].class.length
+            edipi_length: additional_headers['va_eauth_dodedipnid']&.length
           }
           log_message_to_sentry('Failed to find EDIPI, EVSS service call will not succeed', :warn, user_info)
         end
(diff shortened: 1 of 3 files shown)


lib/evss/vso_search/service.rb
```diff
@@ -29,12 +29,14 @@ def get_current_info(additional_headers = {})
...
-        edipi = additional_headers.key?('va_eauth_dodedipnid') ? additional_headers['va_eauth_dodedipnid'] : @user.edipi
+        ssn = additional_headers['va_eauth_pnid'].presence || @user.ssn
+        edipi = additional_headers['va_eauth_dodedipnid'].presence || @user.edipi
         if edipi.nil?
           user_info = {
             loa3: @user.loa3?,
-            mhv_icn: !@user.mhv_icn.nil?
+            mhv_icn: !@user.mhv_icn.nil?,
+            edipi_type: additional_headers['va_eauth_dodedipnid'].class,
+            edipi_length: additional_headers['va_eauth_dodedipnid'].class.length
```
thilton-oddball, line 39: I think this line will always result in an undefined method error. Should it maybe be:

`edipi_length: additional_headers['va_eauth_dodedipnid']&.length`

thilton-oddball (approved):
Looks good!

PR #7010, opened 2021-05-24
[6901] Adding ability to query MPI with a relationship object and ret…

bosawt:
…urn a full MPI Data object

## Description of change
This PR adds a function on the UserRelationship object that will potentially query MPI with the user relationship attributes to return a full MPI data object

## Original issue(s)
department-of-veterans-affairs/vets-api/issues/6901

- Nothing uses this functionality yet, so to test this you'll have to set a breakpoint somewhere in the user object. You should then be able to call `user.relationships` to get the relationships for the user, and then call `get_full_attributes` on one of the objects from that array. Test user `vets.gov.user+36@gmail.com` has BGS dependents mocked that can help with testing.

- To test this, I signed in with `vets.gov.user+36@gmail.com`, confirmed there were returned relationships with `user.relationships`, and then confirmed that I got back an MPI response when calling `user.relationships.first.get_full_attributes`

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -460,6 +460,8 @@ class User < Common::RedisStore
   private
 
   def get_relationships_array
+    return unless loa3?
+
     mpi_profile_relationships || bgs_relationships
   end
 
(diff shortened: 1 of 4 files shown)


jimiray (approved):

bramleyjl (approved):
LGTM

thilton-oddball (approved):
LGTM

PR #7019, opened 2021-05-25
Add MPI id theft flag to user model

patrickvinograd:
## Description of change
Adds the MPI identity theft flag to the MPI profile parser and from there to the user model.

The intent is that this flag can be used in scenarios like Mobile OAuth where we don't authenticate as frequently, to block user access. That change would come as a separate PR. 

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/25057 describes some of the impetus

Tested with a staging user for whom we had the identity flag set in MPI's SQA environment.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -273,6 +273,7 @@ class User < Common::RedisStore
   delegate :icn_with_aaid, to: :mpi
   delegate :vet360_id, to: :mpi
   delegate :search_token, to: :mpi
+  delegate :id_theft_flag, to: :mpi
   delegate :status, to: :mpi, prefix: true
   delegate :error, to: :mpi, prefix: true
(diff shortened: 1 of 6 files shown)


kreek (dismissed):
LGTM 🚀

patrickvinograd:
@bosawt I know you've been making a bunch of changes to MPI-related code, wanted to get your eyes on this

bosawt (approved):
Looks good! I don't think the changes we recently made will impact this or vice versa either, so should be good to go!

I didn't see `ID_THEFT` in the MPI service doc, not really your responsibility but would be good to see IAM update the service doc with a description of that.

cbtkc (approved):
Looks good

PR #7023, opened 2021-05-25
Rails 6.1 upgrade sql deprecation changes

LindseySaari:
## Description of change
The Rails minor upgrade (to 6.1) caused some specs to fail with `ActiveRecord::UnknownAttributeReference Query method called with non-attribute argument(s)`. In the log for rails 6.0, you can see `DEPRECATION WARNING: Dangerous query method (method whose arguments are used as raw SQL) called with non-attribute argument(s)` See [here](http://jenkins.vfs.va.gov/blue/rest/organizations/jenkins/pipelines/testing/pipelines/vets-api/branches/PR-6979/runs/18/nodes/56/steps/57/log/?start=0) for more details. The suggestion is to wrap in Arel.sql. This PR is prep work for the larger Rails 6.1 upgrade. It's conceptually easier to break the changes that can be broken apart from the Rails 6.1 upgrade into smaller PRs to reduce review time and complexity.

## Original issue(s)


thilton-oddball (approved):
LGTM

PR #7027, opened 2021-05-25
Push notification log update

jperk51:
## Description of change
Changed logging from debug to info so it will actually show up in AWS cloudwatch


thilton-oddball (approved):
LGTM

PR #7030, opened 2021-05-25
Rails 6.1 migrations and schema update

LindseySaari:
## Description of change
This PR makes some necessary migrations for the Rails 6.1 upgrade. Some of the specs were failing with `NoMethodError (undefined method 'service_name' for #<ActiveStorage::Blob:0x0000558b5edfbeb8>)`. db migrations were needed for the ActiveStorage blobs (specifically to add the service_name column) Used the solution suggested [here](https://github.com/rails/rails/issues/40895#issuecomment-748811171)

## Original issue(s)


thilton-oddball (approved):
LGTM

PR #7031, opened 2021-05-25
add versioning to appeals

yakschuss:
[API-7380](https://vajira.max.gov/browse/API-7380)

This PR adds necessary fields to both NOD and HLR.

This will need to be backfilled, but before writing the migration to do so, we want to add the process to the Generator directly so that there are no orphaned records.


yakschuss:
> Looks good. My only question would be naming conventions... I would have initially thought `api_version`/`pdf_version` and would be curious to hear your logic for adding `fill` in.

I guess my thinking was that in theory the actual version of the PDF form could differ from the structure and business logic we use to generate it. That's kind of farfetched, though. I'll change it.

nathanbwright (approved):

nathanbwright (approved):
🚀

PR #7032, opened 2021-05-25
Api 7380/update generator to add versionsb

yakschuss:
[API-7380](https://vajira.max.gov/browse/API-7380)

This PR adds the storage of pdf fill version and api versions to HLR/NOD. It's dependent on [this](https://github.com/department-of-veterans-affairs/vets-api/pull/7031) PR.

We'll add a backfilling PR that's dependent on this PR next.

### DIFF
--- a/modules/appeals_api/app/services/appeals_api/pdf_construction/generator.rb
+++ b/modules/appeals_api/app/services/appeals_api/pdf_construction/generator.rb
@@ -7,6 +7,7 @@ module AppealsApi
     class Generator
       def initialize(appeal, version: 'V1')
         @appeal = appeal
+        appeal.update(pdf_fill_version: version)
         @structure = appeal.pdf_structure(version)
       end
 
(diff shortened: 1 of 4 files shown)


modules/appeals_api/app/services/appeals_api/pdf_construction/generator.rb
```diff
@@ -7,6 +7,7 @@ module PdfConstruction
     class Generator
       def initialize(appeal, version: 'V1')
         @appeal = appeal
+        appeal.update(pdf_fill_version: version)
```
nathanbwright, line 10: Maybe I'm missing something here, but why are we updating the appeal itself in the pdf generator?

yakschuss: This version is a direct reference to the version of pdf that we fill for that specific appeal. This is where the generator utilizes the version number to apply the correct structure (pdf version) to the pdf generation. So this is where we store that same version, at the time of generating the PDF.

yakschuss: We could (in theory) do it elsewhere, but this kind of circumvents the edge case of us using a different PDF version from the API version.

cilestin:
Looks like there are lots of failing specs related to this change ☹️

cilestin (approved):

yakschuss:
Waiting to merge this once migrations are run, as this WILL break without them

PR #7039, opened 2021-05-26
API-7106 update v2 hlr contact info

ElleeB:
Updates via this PR:

1. HLR V2 accepts veteran address and prints on form when address provided, else prints 'USE ADDRESS ON FILE' + zip
2. Combines three lines of address data into one line for the form.
3. Adds and updates spec files as necessary


modules/appeals_api/config/schemas/v2/200996.json
```diff
@@ -134,6 +134,12 @@
     "hlrCreateAddress": {
       "type": "object",
       "properties": {
+        "addressLine1":            { "type": "string" },
+        "addressLine2":            { "type": "string" },
+        "addressLine3":            { "type": "string" },
+        "city":                    { "type": "string" },
+        "stateCode":               { "$ref": "#/definitions/hlrCreateStateCode" },
+        "countryName":             { "type": "string" },
```
cilestin, line 142: We should probably keep this as `countryCodeISO2` and ensure it only allows two characters, since we likely don't want to do the translation from a roughly-arbitrary string into an [ISO 2-character country code](https://www.nationsonline.org/oneworld/country_code_list.htm), so let's have our consumers pass in the code instead.

If we need `countryName` in the future, we can update to take that in without the need to update the API version.

modules/appeals_api/spec/fixtures/valid_200996_v2_extra.json
```diff
@@ -17,6 +24,7 @@
           "email": "josie@example.com",
           "timezone": "America/Chicago"
         },
+        "ssoSsocOptIn": true,
```
cilestin, line 27: This is the wrong key. Should be `socOptIn`

modules/appeals_api/spec/models/higher_level_review_spec.rb
```diff
@@ -114,52 +115,10 @@
     it('matches header') { is_expected.to eq auth_headers['X-VA-Insurance-Policy-Number'] }
   end
 
-  describe '#number_and_street' do
```
cilestin, line 117: Since these are shared methods between v1 and v2, we should keep them in this area and not move them into  a v2 context.

cilestin (changes requested):

cilestin:
PS:

If there are enough differences between the v1 model methods and v2, then the v2 versions should be transferred into V2::FormData methods instead, since that's the spot for making the distinction between versions. I made this mistake by writing an HigherLevelReview#email_v2 that still needs refactored away. It should have just been `V2::FormData#email` and worked with the model's `form_data` directly.

modules/appeals_api/config/schemas/v2/200996.json
```diff
@@ -172,6 +178,19 @@
...
+    "hlrCreateStateCode": {
+      "type": "string",
+      "enum": [
+        "AL", "AK", "AS", "AZ", "AR", "AA", "AE", "AP", "CA", "CO",
+        "CT", "DE", "DC", "FM", "FL", "GA", "GU", "HI", "ID", "IL",
+        "IN", "IA", "KS", "KY", "LA", "ME", "MH", "MD", "MA", "MI",
+        "MN", "MS", "MO", "MT", "NE", "NV", "NH", "NJ", "NM", "NY",
+        "NC", "ND", "MP", "OH", "OK", "OR", "PW", "PA", "PR", "RI",
+        "SC", "SD", "TN", "TX", "UT", "VT", "VI", "VA", "WA", "WV",
+        "WI", "WY"
```
cilestin, line 192: 👍

modules/appeals_api/app/services/appeals_api/pdf_construction/higher_level_review/v2/structure.rb
```diff
@@ -25,13 +25,11 @@ def form_fill
...
             form_fields.insurance_policy_number => form_data.insurance_policy_number,
-            form_fields.mailing_address_street => 'USE ADDRESS ON FILE',
-            form_fields.mailing_address_unit_number => '',
-            form_fields.mailing_address_city => '',
-            form_fields.mailing_address_state => '',
-            form_fields.mailing_address_country => '',
-            form_fields.mailing_address_zip_first_5 => '',
-            form_fields.mailing_address_zip_last_4 => '',
+            form_fields.mailing_address_street => form_data.number_and_street,
+            form_fields.mailing_address_city => form_data.city,
```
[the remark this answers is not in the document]
ElleeB: I haven't gotten there yet. I'm just starting with the whiteout and autosizing now. Wanted to get the values combined and printed properly first.

modules/appeals_api/config/schemas/v2/200996.json
ElleeB: K!

modules/appeals_api/spec/fixtures/valid_200996_v2_extra.json
ElleeB: On it

modules/appeals_api/app/models/appeals_api/higher_level_review.rb
```diff
@@ -282,10 +274,6 @@ def veterans_timezone
       veteran&.dig('timezone').presence&.strip
     end
 
-    def header_field_as_string(key)
```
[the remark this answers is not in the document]
ElleeB: this is the model..not versioned

modules/appeals_api/app/models/appeals_api/higher_level_review.rb
```diff
@@ -87,44 +87,40 @@ def birth_yyyy
...
-      header_field_as_string 'X-VA-Insurance-Policy-Number'
+      auth_headers.dig('X-VA-Insurance-Policy-Number')
     end
 
     # 7. CLAIMANT'S NAME
     # 8. CLAIMANT TYPE
 
     # 9. CURRENT MAILING ADDRESS
     def number_and_street
-      address_field_as_string 'addressLine1'
```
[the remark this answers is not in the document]
ElleeB: Everything passes. The values returned are the same. I don't know if that was holdover from another implementation

modules/appeals_api/app/models/appeals_api/higher_level_review.rb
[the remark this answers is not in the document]
ElleeB: Everything passes. The values returned are the same. I don't know if that was holdover from another implementation

modules/appeals_api/spec/factories/higher_level_reviews.rb
```diff
@@ -44,16 +67,13 @@
...
     id { SecureRandom.uuid }
     auth_headers do
       JSON.parse File.read "#{::Rails.root}/modules/appeals_api/spec/fixtures/valid_200996_headers.json"
     end
     form_data do
-      JSON.parse File.read "#{::Rails.root}/modules/appeals_api/spec/fixtures/valid_200996_minimum.json"
-    end
-    trait :status_received do
-      status { 'received' }
+      JSON.parse File.read "#{::Rails.root}/modules/appeals_api/spec/fixtures/valid_200996_minimum_v2.json"
```
cilestin, line 76: Looks like there's a bug on minimum the minimum pdf where the single issue is being shown twice :( We'll have to fix that up, but I'll open a ticket for it so it doesn't stall this PR.

cilestin (approved):
Awesome! 🚀

modules/appeals_api/spec/factories/higher_level_reviews.rb
ElleeB: ok!

PR #7040, opened 2021-05-26
Revert rails 6.1 migrations

LindseySaari:
## Description of change
These migrations aren't available until Rails 6.1. Rolling back and deleting until 6.1 

## Original issue(s)


bramleyjl (dismissed):

db/schema.rb
```diff
@@ -10,10 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-
-ActiveRecord::Schema.define(version: 2021_05_25_183405) do
-
-
+ActiveRecord::Schema.define(version: 2021_05_10_202141) do
```
rileyanderson, line 13: This needs to be the schema version that is currently in master. There was another change after yours `2021_05_25_183405`

rileyanderson (changes requested):
I think we need to fix the schema version. It should stay the same since there was a change after yours

rileyanderson (approved):
👍

db/schema.rb
LindseySaari: Oopsie copy/paste error, fixed ✔️

PR #7045, opened 2021-05-26
[API-7232] - Adds "dev" as a selectable option for ClaimsApi v2 docs

jw81:
## Description of change

Adds the "development" environment as a selectable option on the v2 curl generation docs.

## Original issue(s)
[API-7232](https://vajira.max.gov/browse/API-7232)

### DIFF
--- a/modules/claims_api/app/swagger/claims_api/v2/swagger_root.rb
+++ b/modules/claims_api/app/swagger/claims_api/v2/swagger_root.rb
@@ -38,22 +38,6 @@ module ClaimsApi
           end
         end
 
-        server do
-          key :url, 'https://sandbox-api.va.gov/services/benefits/{version}'
-          key :description, 'VA.gov API sandbox environment'
-          variable :version do
-            key :default, 'v2'
-          end
-        end
-
-        server do
-          key :url, 'https://api.va.gov/services/benefits/{version}'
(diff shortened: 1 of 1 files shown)


modules/claims_api/app/swagger/claims_api/v2/swagger_root.rb
```diff
@@ -30,6 +30,14 @@ class SwaggerRoot
           key :description, 'Allows authenticated and authorized users to access claims data for a given Veteran. No data is returned if the user is not authenticated and authorized.'
         end
 
+        server do
```
SlakrHakr, line 33: i would almost just remove sandbox and prod entirely from this, so the user cant accidentally pick a different environment.
we're going to have to edit this list before it goes all the way out to production anyways, to remove dev from this list.

SlakrHakr (approved):

modules/claims_api/app/swagger/claims_api/v2/swagger_root.rb
jw81: Oh, good call!
[Removed here](https://github.com/department-of-veterans-affairs/vets-api/pull/7045/commits/0fdb563f08aa3af8cce7c5c399a3228967cd92d0).

SlakrHakr (approved):

PR #7046, opened 2021-05-26
Add in validation for the dates on service periods

hinzed1127:
## Description of change

`activeDutyBeginDate` should always occur _before_ `activeDutyEndDate`. This adds in validation to make sure that is the case

## Original issue(s)
https://vajira.max.gov/browse/API-4044


modules/claims_api/app/models/claims_api/auto_established_claim.rb
```diff
@@ -132,6 +133,25 @@ def log_special_issues
...
+      service_periods.each do |service_period|
+        start_date = if service_period['activeDutyBeginDate'].present?
+                       Date.parse(service_period['activeDutyBeginDate'])
+                     end
+        end_date = (Date.parse(service_period['activeDutyEndDate']) if service_period['activeDutyEndDate'].present?)
+
+        if start_date.present? && end_date.blank?
+          next
+        elsif start_date.blank?
+          errors.add :activeDutyBeginDate, 'must be present'
```
hinzed1127, line 148: @SlakrHakr don't some of the required validations occur elsewhere? I thought that the `required` fields aspect gets enforced by JSON schema

SlakrHakr: probably, i was just trying to be as thorough as possible.
also, whos to say sometime in the future this record is edited outside of the workflow we expect today, not the end of the world to come all bases in a couple different places i dont think

SlakrHakr (approved):

PR #7056, opened 2021-05-27
Api 7461 lighthouse ssoe token validation support

dbrown-va:
## Description of change
Adding 2 small changes to support Lighthouse SSOe integration.

1. Adds support for `profile['uuid']` with fallback to the `token.identifiers.uuid`
2. Updates the `derived_loa` method to handle when SSOe traits are used to populate the profile

## Original issue(s)
https://vajira.max.gov/browse/API-7461

Behavior should 100% reverse compatible with non-SSOe logins.  Redis cache entries are based on the uuid so different IDP uses will not result in multiple cache entries as long as the same underlying login method is used.


esubocz-va (approved):
Confirmed backwards compatibility & confirmed new functionality with debugger.

LGTM

PR #7060, opened 2021-05-27
Api 7333/rep phone number should accept intl

yakschuss:
[API-7333](https://vajira.max.gov/browse/API-7333)

This PR adds support for representative international phone number, as well as domestic extensions using autosizing.

Screen Shot 2021-05-27 at 4 30 12 PM
Screen Shot 2021-05-27 at 4 30 39 PM


cilestin (approved):
Looks good! One thing I think I missed is that the phone's countryCode is optional, so may not always have a value (in which case we should assume a US code). I'll make a ticket to follow up on that. NOD doesn't require it either.

PR #7062, opened 2021-05-27
Pause Chapter 31 Submissions

kathleencrawford:
## Description of change

The VR&E pilot is being paused and needs to stop data from being sent to the CMS for an unknown period of time.  I've removed some code to prevent submissions going to their system.  I've also commented out some lines of code in the specs - I know we don't love doing this, but we are going to be putting it back so didn't want to remove them completely.
## Original issue(s)
department-of-veterans-affairs/va.gov-team#25334

This is a minor PR to temporarily prevent Chapter 31 submissions from being sent to the VR&E system for processing.

- [x] Local testing
- [x] Specs updated
- [x] Staging testing planned


cbtkc (approved):
Understandable solution given the business requirements. Looks good.

PR #7064, opened 2021-05-27
Fix missing iam_user icn due to user mpi refactor

kreek:
## Description of change
A recent [refactor](https://github.com/department-of-veterans-affairs/vets-api/pull/6626) to how the base User accesses MPI caused IAMUsers to always return `nil` for `icn`. Previously the user object was used to make MPI queries, in the [updated code](https://github.com/department-of-veterans-affairs/vets-api/pull/6626/files#diff-4bba228da8ba9f87a8f86f8773674025574063190be9aff69fb5b7b0f6f75ccfR134) the user identity is. Since the iam user object had an alias for `mhv_icn` pointing to `iam_icn` but the iam identity object did not MPI queries would fail.

This PR changes the iam user identity to no longer have separete `icn` and `iam_icn` attributes which originally was a defensive choice in case they were different. Since the lookup happens by `iam_icn` in either case I think we can set `icn` in the user identity to that directly. The `mhv_icn` alias is still needed for other parts of the code.

## Original issue(s)
N/A

### DIFF
--- a/app/models/iam_user.rb
+++ b/app/models/iam_user.rb
@@ -13,13 +13,13 @@ class IAMUser < ::User
   redis_key :uuid
 
   attribute :expiration_timestamp, Integer
-  attribute :iam_icn, String
+  attribute :icn, String
   attribute :iam_edipi, String
   attribute :iam_sec_id, String
   attribute :iam_mhv_id, String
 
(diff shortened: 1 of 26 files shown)


patrickvinograd (dismissed):
LGTM

cbtkc (approved):
Looks good

PR #7076, opened 2021-05-28
API-7416 character limits for Veteran first and last names

ElleeB:
1. Added printing of middle initial, which had not been printing (name of field on pdf had changed)
2. Discovered how many autosized characters fit for Veteran first and last names (takes into account the digital signature)
3. Entered limits in schema
4. Updated fixtures: expected pdfs

Ticket: https://vajira.max.gov/browse/API-7416


modules/appeals_api/app/services/appeals_api/pdf_construction/higher_level_review/v2/form_fields.rb
```diff
@@ -4,6 +4,10 @@ module AppealsApi
   module PdfConstruction
     module HigherLevelReview::V2
       class FormFields
+        def middle_initial
```
ElleeB, line 7: This field is not auto filled and needs to be printed in the pdf form field, which had a name change with the new version.

modules/appeals_api/app/services/appeals_api/pdf_construction/higher_level_review/v2/structure.rb
```diff
@@ -17,6 +17,7 @@ def form_fill
           options = {
             # Section I: Vet's ID
             # Veteran name is filled out through autosize text box, not pdf fields
+            form_fields.middle_initial => form_data.middle_initial,
```
ElleeB, line 20: See above comment.

modules/appeals_api/spec/factories/higher_level_reviews.rb
```diff
@@ -70,7 +70,7 @@
   factory :minimal_higher_level_review_v2, class: 'AppealsApi::HigherLevelReview' do
     id { SecureRandom.uuid }
     auth_headers do
-      JSON.parse File.read "#{::Rails.root}/modules/appeals_api/spec/fixtures/valid_200996_headers.json"
+      JSON.parse File.read "#{::Rails.root}/modules/appeals_api/spec/fixtures/valid_200996_headers_minimum.json"
```
ElleeB, line 73: Updated so we could test hlr when no middle initial provided.

modules/appeals_api/config/schemas/v2/200996_headers.json
```diff
@@ -49,7 +49,10 @@
     "X-VA-Last-Name": {
       "allOf": [
```
cilestin, line 50: I think that since you're not doing any references that you could get away with removing `allOf` and just combining all the keys into `X-VA-Last-Name: {...}`

ElleeB: Refactoring breaks it :(

ElleeB: Create a separate ticket to look into the structure?

cilestin (approved):

PR #7077, opened 2021-05-28
[7075] Fix recent regression that broke MPI AddPerson calls by not pr…

bosawt:
…operly updating user identity with orchestrated search MPI response

## Description of change
This PR properly updates the UserIdentity structure when an Orchestrated Search from MPI is queried, then calls MPI AddPerson with the updated profile information. Previously we had introduced a PR that was breaking AddPerson calls, because we were not using the `search_token` from the Orchestrated Search response, this meant we were using a `search_token` from a previous MPI query, and it was timing out before the AddPerson call was completed

## Original issue(s)
department-of-veterans-affairs/vets-api/issues/7075

- Confirmed that standard login is unaffected with this change
- It's difficult to test AddPerson functionality on local/staging, but but spec coverage isn't bad. Should at least confirm that this does not regress things further

### DIFF
--- a/app/models/mpi_data.rb
+++ b/app/models/mpi_data.rb
@@ -139,10 +139,11 @@ class MPIData < Common::RedisStore
   # call is made. The response is recached afterwards so the new ids can be accessed on the next call.
   #
   # @return [MPI::Responses::AddPersonResponse] the response returned from MPI Add Person call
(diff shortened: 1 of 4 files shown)


jimiray (approved):

cbtkc (approved):
Looks good

PR #7080, opened 2021-06-01
SM Folder Messages Page Size Fix

narin:
## Description of change
MHV has updated their folder messages endpoint with validation that will not allow the `page_size` to be greater than the number of messages in the folder, which is causing issues because we had `page_size` set to a constant 250.

This PR updates our method to use the `folder_count` as the page size if it is less than the max size of 250, and updates the looping logic to use the proper number of pages since we can calculate it using the folder count.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#21454

### DIFF
--- a/app/controllers/v0/messages_controller.rb
+++ b/app/controllers/v0/messages_controller.rb
@@ -5,7 +5,7 @@ module V0
     include Filterable
 
     def index
-      resource = client.get_folder_messages(@current_user.uuid, params[:folder_id].to_s, use_cache?)
+      resource = client.get_folder_messages(@current_user.uuid, params[:folder_id].to_s)
       raise Common::Exceptions::RecordNotFound, params[:folder_id] if resource.blank?
 
       resource = params[:filter].present? ? resource.find_by(filter_params) : resource
--- a/lib/sm/client.rb
+++ b/lib/sm/client.rb
@@ -137,7 +137,6 @@ module SM
           json = perform(:get, path, nil, token_headers).body
         else
           # If total count > 250, we'll have to loop multiple times using the MHV_MAXIMUM_PER_PAGE
-          # In mose cases, there'll be less than 250 messages in a given folder, so this will only run once
           page = 1
           total_pages = (total_count / MHV_MAXIMUM_PER_PAGE.to_f).ceil
           json = { data: [], errors: {}, metadata: {} }
--- a/lib/sm/client.rb
+++ b/lib/sm/client.rb
@@ -118,26 +118,40 @@ module SM
     #
     # @return [Common::Collection]
     #
+    # rubocop:disable Metrics/MethodLength
     def get_folder_messages(user_uuid, folder_id, use_cache)
       cache_key = "#{user_uuid}-folder-messages-#{folder_id}"
+
+      # Get folders and count so we can determine max page size
+      folders = get_folders(user_uuid, true)
+      current_folder = folders.find_first_by('id' => { 'eq' => folder_id })
+      folder_count = current_folder[:count]
+
       get_cached_or_fetch_data(use_cache, cache_key, Message) do
-        page = 1
-        json = { data: [], errors: {}, metadata: {} }
+        if folder_count > MHV_MAXIMUM_PER_PAGE
+          # If folder_count is greater than the MHV max allowed, loop at increments of the max
+          page = 1
+          json = { data: [], errors: {}, metadata: {} }
+          loop do
+            path = "folder/#{folder_id}/message/page/#{page}/pageSize/#{MHV_MAXIMUM_PER_PAGE}"
+            page_data = perform(:get, path, nil, token_headers).body
+            json[:data].concat(page_data[:data])
+            json[:metadata].merge(page_data[:metadata])
+            break unless page_data[:data].size == MHV_MAXIMUM_PER_PAGE
 
-        loop do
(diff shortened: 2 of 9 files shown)


lib/sm/client.rb
```diff
@@ -118,32 +118,47 @@ def delete_folder(id)
...
-        json = { data: [], errors: {}, metadata: {} }
 
-        loop do
-          path = "folder/#{folder_id}/message/page/#{page}/pageSize/#{MHV_MAXIMUM_PER_PAGE}"
-          page_data = perform(:get, path, nil, token_headers).body
-          json[:data].concat(page_data[:data])
-          json[:metadata].merge(page_data[:metadata])
-          break unless page_data[:data].size == MHV_MAXIMUM_PER_PAGE
+      # Get folders and count so we can determine max page size
+      folders = get_folders(user_uuid, true)
```
narin, line 126: In most cases, we should already have these folders in our Redis cache

spec/request/folders_request_spec.rb
```diff
@@ -139,55 +139,64 @@
     describe 'nested resources' do
       it 'gets messages#index' do
         VCR.use_cassette('sm_client/folders/nested_resources/gets_a_collection_of_messages') do
-          get "/v0/messaging/health/folders/#{inbox_id}/messages"
+          VCR.use_cassette('sm_client/folders/gets_a_collection_of_folders') do
+            get "/v0/messaging/health/folders/#{inbox_id}/messages"
+          end
         end
```
narin, line 145: My changes to `client.rb` above seem to be working, but due to the additional `get_folders` call, it is messing up these pagination related unit tests.  I think I'm messing up somewhere with the nesting of the VCR cassettes. Could use some extra ey

app/controllers/v0/messages_controller.rb
```diff
@@ -5,7 +5,7 @@ class MessagesController < SMController
     include Filterable
 
     def index
-      resource = client.get_folder_messages(@current_user.uuid, params[:folder_id].to_s, use_cache?)
+      resource = client.get_folder_messages(@current_user.uuid, params[:folder_id].to_s)
```
narin, line 8: `get_folder_messages` will now determine whether to use cache or not based on if the length of our cached_messages === total folder count or not. See below.

spec/lib/sm/client/folders_spec.rb
```diff
@@ -41,20 +41,24 @@
...
-        expect(messages.data.size).to eq(10)
+        VCR.use_cassette('sm_client/folders/gets_a_single_folder') do
+          # set the max pages to 1 for testing purposes
+          stub_const('SM::Client::MHV_MAXIMUM_PER_PAGE', 2)
+          # There are 10 records, 2 per page, so it should loop 6 times making requests
+          expect(client).to receive(:perform).and_call_original.exactly(6).times
+          messages = client.get_folder_messages('1234', folder_id)
+          expect(messages).to be_a(Common::Collection)
+          expect(messages.data.size).to eq(10)
+        end
```
narin, line 52: Wrap all "get messages" tests with this `gets_a_single_folder` cassette since we will always make the `get_folder` call before getting messages.

lib/sm/client.rb
```diff
@@ -118,26 +118,45 @@ def delete_folder(id)
...
 
+        # If total count > 250, we'll have to loop multiple times using the MHV_MAXIMUM_PER_PAGE
+        # In mose cases, there'll be less than 250 messages in a given folder, so this will only run once
         loop do
-          path = "folder/#{folder_id}/message/page/#{page}/pageSize/#{MHV_MAXIMUM_PER_PAGE}"
+          path = "folder/#{folder_id}/message/page/#{page}/pageSize/#{page_size}"
           page_data = perform(:get, path, nil, token_headers).body
           json[:data].concat(page_data[:data])
           json[:metadata].merge(page_data[:metadata])
-          break unless page_data[:data].size == MHV_MAXIMUM_PER_PAGE
```
narin, line 132: Now that we have the total count, we know the number of pages and if there is a next page instead of making a guess with the page size.

lib/sm/client.rb
```diff
@@ -118,26 +118,45 @@ def delete_folder(id)
...
     # @return [Common::Collection]
     #
-    def get_folder_messages(user_uuid, folder_id, use_cache)
+    # rubocop:disable Metrics/MethodLength
+    def get_folder_messages(user_uuid, folder_id)
       cache_key = "#{user_uuid}-folder-messages-#{folder_id}"
+
+      # MHV API doesn't allow for a page_size > total number of messages in folder, so
+      # we get the folder's message count here
+      current_folder = get_folder(folder_id)
```
narin, line 127: @kreek if `get_folder()` fails for here some reason, do I need to throw an exception to prevent the code below from executing, or will rails handle that for me?

kreek: @narin dug into the code and I think it depends on the error. If it's because the request failed I think it will be handled for you, otherwise no (it will always be handled but you'll get a 500 response). However, there's a way to find out for sure 😄 Write a spec for when `get_folder` fails. Take (or make) a VCR cassette where that call fails and check to see if it's gracefully handled. If you wan to test a non HTTP error you could do something like `allow(Folder).to receive(:new).and_raise(StandardError)` before the expectation.

narin: Thanks. Does what I did at `spec/request/swagger_spec.rb:1148` cover the first scenario you mentioned?

narin:
This issue ended up being fixed on MHV's side. Closing out this PR for now.

PR #7081, opened 2021-06-01
Appeals api & aws-sdk-s3 gem rollback 

LindseySaari:
## Description of change
#7020 was committed/merged last week resulting in test suite failures related to #7025. #7025 caused the build to fail, which was rolled back. Most likely, #7025 wasn't up to date with master before it was merged it (master contained the latest aws-sdk-s3 gem). This PR reverts the aws-sdk-s3 gem update which looks to be the culprit as to why we saw test failures in #7025. 

Additionally: The failing spec reference some [s3 settings](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/appeals_api/spec/requests/v1/notice_of_disagreements/evidence_submissions_controller_spec.rb#L14-L23), so we're thinking it has something to do with the test suite failures in relation to the gem upgrade

## Original issue(s)


cbtkc (approved):
Looks good

PR #7088, opened 2021-06-01
Move helpers inside task to unblock other teams

dillo:
- Moving helpers and constants inside the tasks for now so that other teams are not blocked by rake-related issues. Will implement a better solution in the near future.
## Original issue(s)


cbtkc (approved):
LGTM. Will need to DRY up code at a later date.

PR #7089, opened 2021-06-01
Update vets-json-schema

dmccurdybah:
## Description of change
Update vets-json-schema repo to 2.5.0

## Original issue(s)
department-of-veterans-affairs/va.gov-team#25122

PR #7094, opened 2021-06-01
Update 5490 spool files

dmccurdybah:
## Description of change
Update 5490 spool file to include new language updates based on the new `sponsorStatus` field

## Original issue(s)
department-of-veterans-affairs/va.gov-team#25122

- requires PR [7089](https://github.com/department-of-veterans-affairs/vets-api/pull/7089) to update vets-json-schema to be merged in first


shawkey-daniel (approved):
LGTM

delli-gatti-michael (approved):

app/workers/education_form/forms/va_5490.rb
```diff
@@ -65,5 +65,25 @@ def previous_benefits
...
     end
+
+    def veteran_date_of_death_label
+      if @applicant.benefit != 'chapter33' || !Flipper.enabled?(:edu_form5490_updates) || @applicant.sponsorStatus.nil?
+        return 'Date of death or MIA/POW:'
+      end
+
+      label = ''
+
+      case @applicant.sponsorStatus
```
cbtkc, line 76: label variable isn't needed, you can just have the string returned after the case is evaluated. If you'd like to keep the variable however, another acceptable method is instead of instantiating label as an empty string, you could set it equal to the evaluation of `case @applicant.sponsorStatus`

cbtkc (changes requested):

app/workers/education_form/forms/va_5490.rb
dmccurdybah: Thanks! I just assumed rubocop would complain.

cbtkc: No worries, everything else looks good

cbtkc (approved):
👍

PR #7096, opened 2021-06-01
Uses id theft flag to block mobile session refresh

patrickvinograd:
## Description of change
Uses the id_theft_flag attribute set in MPI (and newly available in user record) to prevent mobile OAuth sessions from being established. 

This is warranted for mobile because otherwise flagged users would be able to continue using the OAuth long-lived session refresh mechanism for up to 45 days. This is not an issue for SAML-based auth on VA.gov because re-authentication is required every time a session expire, and authentication is blocked by the id theft flag. 

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/25057

* I plan to re-test this but it will require the IAM team to remove then re-add the id theft flag for a user so we can log in to the app, block them, and then make sure session refresh is blocked.

### DIFF
--- a/app/models/iam_user_identity.rb
+++ b/app/models/iam_user_identity.rb
@@ -44,7 +44,8 @@ class IAMUserIdentity < ::UserIdentity
       iam_mhv_id: valid_mhv_id(iam_profile[:fediam_mhv_ien]),
       last_name: iam_profile[:family_name],
       loa: { current: loa_level, highest: loa_level },
-      middle_name: iam_profile[:middle_name]
+      middle_name: iam_profile[:middle_name],
+      sign_in: { service_name: "oauth_#{iam_auth_n_type}", account_type: iam_profile[:fediamassur_level] }
     )
 
     StatsD.increment('iam_ssoe_oauth.auth_type', tags: ["type:#{iam_auth_n_type}"])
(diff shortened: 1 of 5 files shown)


app/models/iam_user_identity.rb
```diff
@@ -44,7 +44,8 @@ def self.build_from_iam_profile(iam_profile)
       iam_mhv_id: valid_mhv_id(iam_profile[:fediam_mhv_ien]),
       last_name: iam_profile[:family_name],
       loa: { current: loa_level, highest: loa_level },
-      middle_name: iam_profile[:middle_name]
+      middle_name: iam_profile[:middle_name],
+      sign_in: { service_name: "oauth_#{iam_auth_n_type}", account_type: iam_profile[:fediamassur_level] }
```
patrickvinograd, line 48: This is somewhat unrelated but will populate the credential and LOA tags in Sentry for mobile instead of having them say "not-logged-in"

lib/iam_ssoe_oauth/session_manager.rb
```diff
@@ -65,7 +67,7 @@ def build_identity(iam_profile)
 
     def build_session(access_token, user_identity)
       @session = IAMSession.new(token: access_token, uuid: user_identity.uuid)
-      @session.save
+      @session
```
patrickvinograd, line 70: Idea is to not save `session` until *after* validating, since otherwise a subsequent request with the token would find the cached session.

kreek: Like the change to saving after validating 👍 RubyMine is telling me you don't need line 70 as 69 also returns the session object, but maybe it's more clear to humans as is? 🤷‍♂️

kreek (approved):
LGTM 🚀

cbtkc (approved):
Looks good

PR #7099, opened 2021-06-01
API-7420 check address char limits and confirm schema hlr v2

ElleeB:
The work for this PRs related ticket had already been completed in two previous PRs, however, looking into the schema one last time revealed missing specs. In addition, there were some inconsistencies in our schema and our fixtures. This PR attempts to clean these up.

Ticket: https://vajira.max.gov/browse/API-7420

In relation to the original acceptance criteria, which had already been completed, here are the Veteran address values appearing in our dev portal docs:

Screen Shot 2021-06-01 at 10 55 45 AM

### DIFF
--- a/config/features.yml
+++ b/config/features.yml
@@ -392,3 +392,7 @@ features:
     actor_type: user
     description: Form 10203 updates in support of Isakson and Roe legislation
     enable_in_development: true
+  enable_check_in_experience:
+      actor_type: user
+      description: Enables the health care check-in experiences
+      enable_in_development: true


modules/appeals_api/config/schemas/v2/200996.json
```diff
@@ -48,7 +48,7 @@
         "veteran":                            { "$ref": "#/definitions/hlrCreateVeteran" },
         "informalConferenceContact":          { "$ref": "#/definitions/hlrCreateInformalConferenceContact" },
         "informalConferenceTime":             { "$ref": "#/definitions/hlrCreateInformalConferenceTime" },
-        "representative":                     { "$ref": "#/definitions/hlrCreateRepresentative" },
+        "informalConferenceRep":              { "$ref": "#/definitions/hlrCreateRepresentative" },
```
ElleeB, line 51: I've returned this key to its original name in order to get the schema in working order. Changing it caused confusion and would require unnecessary complexity in order to refactor given the number of methods on the HLR model that rely on this key. 

I will create a ticket for someone to get in here and spend time refactoring so we can move ahead with `representative` if we want.

modules/appeals_api/spec/fixtures/valid_200996_v2_extra.json
```diff
@@ -1,154 +1,153 @@
 {
```
ElleeB, line 1: Sorry for the long commit change. This file's indentation appears to have changed. The only actual content change is at line#31 where I've inserted the key `name` to correctly form the representative's name object.

modules/appeals_api/spec/support/schema_helpers.rb
```diff
@@ -15,4 +15,19 @@ def read_schema(filename, schema_version = 'v1')
       )
     )
   end
+
+  def read_schema_v2(filename, schema_version = 'v2')
```
ElleeB, line 19: helper method to point to the appropriate schema file to test against for the schema spec.

yakschuss: Why are we passing in the schema version as an argument if the method name is already constrained to the version?

ElleeB: The method name doesn't constrain it to the version...the schema version variable is passed into the path that the method reads in order to pull the schema from the correct directory/file. Am I missing something?

ElleeB: Never mind. I see lol.

modules/appeals_api/app/models/appeals_api/higher_level_review.rb
```diff
@@ -283,6 +283,7 @@ def veteran_phone_is_not_too_long
 
     # validation
     def informal_conference_rep_name_and_phone_number_is_not_too_long
+      return if pdf_version == 'v2'
```
ElleeB, line 286: this is my thought process...not the best solution, but we need to skip some of the model validations for v2

PR #7101, opened 2021-06-01
[7028] Logging message to Sentry if MPI request is missing necessary …

bosawt:
…data

## Description of change
This PR just adds a message to Sentry if there are missing fields in an MPI request. We should probably catch these kinds of issues upstream but it's possible we could have cached the MPI response, so a user would be able to access it even if we temporarily didn't have access to certain data.

## Original issue(s)
department-of-veterans-affairs/vets-api/issues/7028

- Sentry warning will have form like this: `[MPI Data] Request error: required values are missing for keys: [:birth_date, :ssn]`

- I confirmed this message was raised when a user with incomplete Identity data signed in

### DIFF
--- a/app/models/mpi_data.rb
+++ b/app/models/mpi_data.rb
@@ -6,12 +6,14 @@ require 'mpi/orch_search_service'
 require 'common/models/redis_store'
 require 'common/models/concerns/cache_aside'
 require 'mpi/constants'
+require 'sentry_logging'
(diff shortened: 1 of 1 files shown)


joeniquette (approved):

jimiray (approved):

cbtkc (approved):
👍

PR #7102, opened 2021-06-01
24909 mobile discovery minimum version/maintenance windows

kreek:
## Description of change
Updates the mobile discovery endpoint to include a minimum app version supported by the API and a list of services affected by upstream maintenance windows. Maintenance windows are implemented as a service graph where the leaf service nodes can be put into maintenance if any of their upstream dependencies are down. 

e.g. below Claims, Direct Deposit, and Letters depend on EVSS, which depends on BGS, so if either BGS or EVSS are down then those services would show up in the list of maintenance windows. If a service is in maintenance due to multiple upstream dependencies being down then it uses the earliest start time and the latest end time for its window.

diagram

## Original issue(s)
department-of-veterans-affairs/va.gov-team#24909

* api and auth urls, web views and minimum version,  are all defined in the `settings.yml` file


modules/mobile/app/controllers/mobile/discovery_controller.rb
```diff
@@ -6,12 +6,52 @@ module Mobile
...
+      %i[emis military_service_history],
+      %i[evss claims],
+      %i[evss direct_deposit_benefits],
+      %i[evss letters_and_documents],
+      %i[idme auth_idme],
+      %i[mhv auth_mhv],
+      %i[mhv secure_messaging],
+      %i[vaos appointments],
+      %i[vet360 user_profile_update]
+    )
```
kreek, line 29: This probably needs tweaking to get all the services dependencies correct.

jjulian:
Fantastic graphic! Seems like defining the graph may be the trickiest part of this.

lexicalninja:
I assume the FE will need to come up with content for this time of error and should not depend on the `description`?

config/settings.yml
```diff
@@ -1157,3 +1157,8 @@ bid:
 # Settings for Test User Dashboard modules
 test_user_dashboard:
   env: dev
+
+# Settings for mobile module api. Currently unused. They require being added to the devops repo settings to be used.
```
jperk51, line 1161: Should we delete these settings and add them back when they are actually being used by the API?

jperk51 (approved):
LGTM

PR #7103, opened 2021-06-01
Derives multifactor correctly for mobile OAuth

patrickvinograd:
## Description of change
Mobile OAuth is not brokered through ID.me, so MHV/DSL credentials
by definition to not have multifactor authentication present. This
change ensures that that attribute is derived correctly for all
credential types available for OAuth.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/20190

@kreek I noticed that `authorized_services` for payment is derived only from "can access evss" and not also "can access ppiu" - so it's now inaccurate in this case. The payment controller itself is correctly authorized. Do we want to make `authorized_services` smarter so it can check both of those policies?

### DIFF
--- a/app/models/iam_user_identity.rb
+++ b/app/models/iam_user_identity.rb
@@ -11,6 +11,7 @@ class IAMUserIdentity < ::UserIdentity
 
   PREMIUM_LOAS = [2, 3].freeze
   UPGRADE_AUTH_TYPES = %w[DSL MHV].freeze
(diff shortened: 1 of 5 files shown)


kreek:
@patrickvinograd yes to making `authorized_services` smarter. I'll make a ticket for it to accept multiple policies.

kreek (approved):
LGTM 🚀

cbtkc (approved):
Looks good

PR #7105, opened 2021-06-02
Fixes NameError bug in mobile claims endpoint

patrickvinograd:
## Description of change
Fixes NameError observed in Sentry here:
http://sentry.vfs.va.gov/organizations/vsp/issues/14952/?query=is%3Aunresolved+assigned%3A%23mobile-api-team&statsPeriod=14d

Looks like this was refactored out of the controller at some point and still had references to `@current_user` and `params`. 
Then, the spec was effectively hidden from CI because it was named `.spec.rb` instead of `_spec.rb`. 

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/25260


kreek (approved):
LGTM 🚀

jperk51 (approved):
Nice catch

PR #7111, opened 2021-06-02
25428 decompose the registry method

dillo:
- Break the registry methods hash into their own relevant hashes
- Making it easy to add other parameters to the registries
- Seeing the potential to extract registries into their own objects in the future
- Add the ability to change the resource count
## Original issue(s)
department-of-veterans-affairs/va.gov-team#25428

- [x] RSpec test suite passing


mdewey (approved):
LGTM

PR #7123, opened 2021-06-02
Transliterate names for lighthouse

annaswims:
## Description of change
the benefits api's upstream partner only accepts some characters.  But, they use a fuzzy match so it should be ok. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#25163


cbtkc (approved):
looks good

PR #7130, opened 2021-06-03
Claims appeals remove timezone shift from dates

jperk51:
## Description of change
Removed timezone shift from claims and appeals updatedAt date. It was defaulting to 00:00:00 +0000 since there isn't any indicator on the original data and this was causing date shifting issues depending on the timezone of the FE.


kreek (approved):
LGTM 🚀

kreek (approved):

kreek (approved):
LGTM 🚀

PR #7132, opened 2021-06-03
Submit VA Form 28-1900 to Central Mail

kathleencrawford:
## Description of change

When VA Form 28-1900 gets submitted from va.gov, an email is sent to the Veteran Readiness and Employment (VR&E) team notifying them of a submission.  This email includes the user's participant ID.  About 2-3% of the time, there is a submission that does not have a participant ID so we have been asked to temporarily route those submissions to central mail.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#25004

We are re-using code that is already implemented for other claims (we already send other claims to central mail), so no new logging or settings were necessary. 

- [x] Local testing
- [x] Specs updated
- [x] Staging testing planned


cbtkc (commented):
Is there a spec we could add to test and verify that veterans without a participant_id are indeed having their claims sent to central mail?

kathleencrawford:
> Is there a spec we could add to test and verify that veterans without a participant_id are indeed having their claims sent to central mail?

great catch @cbtkc!  please note the new changes i've made in `spec/controllers/v0/veteran_readiness_employment_claims_controller_spec.rb`

cbtkc (approved):
LGTM

PR #7148, opened 2021-06-03
API-7341 - HLR v2 statuses

nathanbwright:
## Description of change
This PR adds support for the HLR model to have two different sets of statuses depending on the `api_version`.

## Original issue(s)
https://vajira.max.gov/browse/API-7341

### DIFF
--- a/modules/appeals_api/app/models/concerns/appeals_api/hlr_status.rb
+++ b/modules/appeals_api/app/models/concerns/appeals_api/hlr_status.rb
@@ -1,10 +1,18 @@
 # frozen_string_literal: true
 
+require 'appeals_api/central_mail_updater'
+
 module AppealsApi
   module HlrStatus
     extend ActiveSupport::Concern
 
-    STATUSES = %w[pending submitting submitted processing error uploaded received success expired].freeze
+    V1_STATUSES = %w[pending submitting submitted processing error uploaded received success expired].freeze
+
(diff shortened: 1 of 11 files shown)


modules/appeals_api/app/models/concerns/appeals_api/hlr_status.rb
```diff
@@ -1,10 +1,18 @@
...
+    V1_STATUSES = %w[pending submitting submitted processing error uploaded received success expired].freeze
+
+    V2_INTERNAL_STATUSES = %w[pending submitting submitted].freeze
+    V2_STATUSES = [*V2_INTERNAL_STATUSES, *CentralMailUpdater::CENTRAL_MAIL_STATUSES].uniq.freeze
+
+    # used primarly for reporting
+    STATUSES = [*V1_STATUSES, *V2_STATUSES].uniq.freeze
 
     RECEIVED_OR_PROCESSING = %w[received processing].freeze
     COMPLETE_STATUSES = %w[success error].freeze
```
cilestin, line 18: Shouldn't we also add `caseflow` to the completed statuses, since it's "one beyond" `success` for HLR V2?

nathanbwright: Great point! Thanks for catching this.

cilestin (approved):

PR #7152, opened 2021-06-04
Fix 10-10CG PDF download (file reference) bug

kevinmirc:
Fixes: department-of-veterans-affairs/va.gov-team#24989

## Description of change

Breakman (`Brakeman -t FileAccess`) will not allow us to use an attribute (or return value of a method) of an instance that was built with user-provided params, as a file name sent in the response.

To avoid the breakman warning, we interpreted the file path. This resulted in errors reading the file because PDF Fill will provide a different name for PDFs that have an "extras" file (`{guid}_final.pdf` instead of `{guid}.pdf`).

This change allows us to get the file name from PDF Fill while also not violating the breakman warning.

## Original issue(s)
- Ticket: https://github.com/department-of-veterans-affairs/va.gov-team/issues/24989
- Related Slack Threads:
  - https://dsva.slack.com/archives/CBU0KDSB1/p1595341676422700
  - https://dsva.slack.com/archives/CBU0KDSB1/p1601653841002800


cbtkc (approved):
Looks good

PR #7155, opened 2021-06-04
add default country code for HLR v2

yakschuss:
[API-7578](https://vajira.max.gov/browse/API-7578)

This PR adds a default countryCode for HLR v2 if it doesn't exist

The ticket lists adding a requirement for countryCode on the schema, but I'm not sure if requiring a value and then providing a default makes sense. Will discuss in standup


cilestin:
Since this is already dealing with phone numbers, I think the regex for phoneNumberExt is incorrect, as we're automatically supplying the "x" in front of the number for them. Should probably just allow `^[0-9]{1,10}$`

Additionally, let's add `phoneNumberExt` support for the veteran phone number like we did for the ICRep - where it adds the extension to the side of the Veteran's phone number boxes (will require a new bounding box).

cilestin (approved):

PR #7158, opened 2021-06-04
remove backend review groups from modules

cbtkc:
## Description of change
Code owned by teams nested under modules should be allowed to review their own code without other groups approval

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,14 +6,14 @@
 *       @department-of-veterans-affairs/backend-review-group @department-of-veterans-affairs/va-api-engineers
 
 # Experiment for isolating VFS teams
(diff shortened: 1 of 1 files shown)


LindseySaari:
By removing backend-review-group from modules, I'm thinking of the case where the build fails and we need to make a change to a module quickly. Would we need an approval from the module team? Wondering if this may be a blocker in cases where there's a breaking change.

ericboehs:
Yeah. I think we should close this as I just ran into an issue with https://github.com/department-of-veterans-affairs/vets-api/pull/7323. We can't merge code because vsp-identity solely owns some of the changed files.

**Unless**, we want to change it so that backend-review-group can dismiss code review approvals so that we can merge any PR. 🤔 

Also, Cameron, each team added per line isn't an `AND`, it's an `OR`, meaning only one person from one of those teams has to approve. (Unless I'm not understanding it correctly.)

PR #7164, opened 2021-06-07
rswag: creation_date can be null

SlakrHakr:
## Description of change
Appears that the creation_date field in the 0966/active response can be null.
Updated the spec and re-generated openapi documentation.


gpbowman-git (approved):
LGTM

cshupp1 (approved):

PR #7166, opened 2021-06-07
display title implementation

jperk51:
## Description of change
For appeals web creates a custom display title based on the appeal type and the program area, we want the api to do the same then return that to the front end. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#25755

Docs and tests have been updated to match this change


kreek (dismissed):
LGTM 🚀

kreek (approved):
LGTM 🚀

PR #7171, opened 2021-06-08
Reject additional properties not on claims schemas in requests

hinzed1127:
## Description of change

Adds in [`additionalProperties`](https://json-schema.org/understanding-json-schema/reference/object.html#additional-properties) to every eligible field on claims endpoint schemas. This throws an error for any request with another property undefined on our schemas.

## Original issue(s)
https://vajira.max.gov/browse/API-6772

- Added tests on all 3 request specs that makes sure validation works as expected
- JSON-schemer offers a method in its [Errors Module](https://github.com/davishmcclurg/json_schemer/pull/67) to generate more human readable errors. It might be worth seeing if this could be expanded further with [our error formatting in vets-api](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/json_schema/json_api_missing_attribute.rb)


lib/json_schema/json_api_missing_attribute.rb
```diff
@@ -12,8 +12,10 @@ def initialize(details)
     def to_human(detail)
       if detail['type'] == 'required'
         "The property #{to_source(detail)} did not contain the required key #{detail['details']['missing_key']}"
+      elsif detail['type'] == 'schema' && detail['schema_pointer'].end_with?('additionalProperties')
```
hinzed1127, line 15: it was hard for me to tell looking at the validation gem if `additionalProperties` would be the only type of validation with type `schema`, but adding in the `schema_pointer` check as well, should be sufficient to ensure that we're catching the right error

lib/json_schema/json_api_missing_attribute.rb
```diff
@@ -12,8 +12,10 @@ def initialize(details)
...
+        "The property #{to_source(detail)} is not defined on the schema. Additional properties are not allowed"
```
hinzed1127, line 16: Open to other wording, this is just what I came up with

modules/claims_api/spec/requests/v1/intent_to_file_request_spec.rb
```diff
@@ -162,6 +162,21 @@
       end
     end
 
+    it 'doesn\'t allow additional fields' do
+      with_okta_user(scopes) do |auth_header|
+        params = data
+        params[:data][:attributes]['someBadField'] = 'someValue'
+
+        post "#{path}/validate", params: params.to_json, headers: headers.merge(auth_header)
```
jw81, line 170: Just curious, but why test this in the `/validate` context?
That endpoint is "technically" deprecated right now and if we were to eventually remove the endopint altogether, we'd lose this test.

jw81:
Any reason to not add these tests to the "rswag" version of the request specs?

SlakrHakr (approved):

modules/claims_api/spec/requests/v1/intent_to_file_request_spec.rb
hinzed1127: ahh that was just an oversight. I think this was one of the last tests I added and just wasn't paying attention since it was mostly copy-paste-modify at that point. I'll move into into the standard post where it makes more sense

hinzed1127:
> Any reason to not add these tests to the "rswag" version of the request specs?

I just added it to the disability request spec as one of multiple exampless. I don't think it necessarily needs to be in all of the specs since it's a different version of the same error we're already generating in other tests. I wanted to specifically test the error message in the non-rswag tests in case that were to break down the line, though.

SlakrHakr (approved):

PR #7172, opened 2021-06-08
Fix Statuses parameter not working on /vaos/v2/appointments GET

ajmagdub:
## Description of change

There's a statuses parameter on the GET for the appointments fetch in the VAOS service, but passing it to /vaos/v2/appointments doesn't seem to have any effect on the results returned.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#25800
RSPECs for submitting multiple statuses will need to implemented at a later date because the VAOS Service currently cannot process that.


ajmagdub:
vaos service returned an error when sent more than one status. test for submitting more than one status will have to be implemented when the vaos service can accept more than one status. Created [VAOSR-2005](https://issues.mobilehealth.va.gov/browse/VAOSR-2005)

LindseySaari (approved):

grossb1 (approved):

PR #7178, opened 2021-06-08
API-7743 - Add details about evidence submission S3 upload

nathanbwright:
## Description of change
Adds additional details to our evidence submission docs - including a schema that shows the regex for Veteran name.

## Original issue(s)
https://vajira.max.gov/browse/API-7743


modules/appeals_api/app/swagger/appeals_api/v1/schemas/notice_of_disagreements.rb
```diff
@@ -400,6 +400,61 @@ class NoticeOfDisagreements
...
+            key :example, '316'
+            key :description, 'VBA form number of the document'
+          end
+
+          property :businessLine do
+            key :type, :string
+            key :example, 'BVA'
+            key :enum, %i[BVA]
+            key :description,
+                <<~DESCRIPTION
```
yakschuss, line 453: Why is this string formatting necessary?

yakschuss (dismissed):
🚀

SlakrHakr (approved):

modules/appeals_api/app/swagger/appeals_api/v1/schemas/notice_of_disagreements.rb
nathanbwright: It was a left over from the copy/paste where they had a multi line list.

modules/appeals_api/app/swagger/appeals_api/v1/notice_of_disagreements_controller_swagger.rb
```diff
@@ -641,6 +641,7 @@ class AppealsApi::V1::NoticeOfDisagreementsControllerSwagger
   swagger_path '/path' do
     operation :put, tags: NOD_TAG do
       key :operationId, 'putNoticeOfDisagreementEvidenceSubmission'
+      key :description, File.read(AppealsApi::Engine.root.join('app', 'swagger', 'appeals_api', 'v1', 'put_description.md'))
```
cilestin, line 644: I think with this inclusion we could remove the "example metadata.json file" section in our main doc body, or at least change it to point to this example at `/path`. What do you think?

yakschuss (approved):
🚀

modules/appeals_api/app/swagger/appeals_api/v1/notice_of_disagreements_controller_swagger.rb
nathanbwright: I think Kelly would like it to be in both spots... We can always remove it later if we decide too.

PR #7190, opened 2021-06-09
[22732] adds user to EVSS VSOSearch service instantiation

bramleyjl:
## Description of change
This PR changes the EVSS service instantiation to properly take a user as an argument so that the user's attribute values will be available as backup values in the event that `additional_headers' is missing an attribute value. It is a quick & partial reversal of [this PR](https://github.com/department-of-veterans-affairs/vets-api/pull/5793/files) that changed the EVSS service from being instantiated with a `User` instance to using authentication_headers.

For further information related to this change, see my writeup in the linked issue thread.

## Original issue(s)

Unit tests passed, manual testing of EVSS::VSOSearch service creation with only backup user object to confirm backup functionality is working.

### DIFF
--- a/lib/evss/vso_search/service.rb
+++ b/lib/evss/vso_search/service.rb
@@ -29,7 +29,7 @@ module EVSS
       private
 
       def request_headers(additional_headers)
-        edipi = additional_headers['va_eauth_dodedipnid'].presence || @account.edipi
+        edipi = additional_headers['va_eauth_dodedipnid'].presence || @account&.edipi
         ssn = additional_headers['va_eauth_pnid'].presence || @user.ssn
         {
           'ssn' => ssn,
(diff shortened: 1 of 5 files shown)


shawkey-daniel (approved):
LGTM

lib/evss/vso_search/service.rb
```diff
@@ -31,16 +31,6 @@ def get_current_info(additional_headers = {})
       def request_headers(additional_headers)
         ssn = additional_headers['va_eauth_pnid'].presence || @user.ssn
         edipi = additional_headers['va_eauth_dodedipnid'].presence || @user.edipi
```
LindseySaari, line 33: Could there ever be a chance here still that edipi could be nil?

bramleyjl: Yes, that's definitely a possibility and we changed the logic to pull from the account instead to avoid this.

LindseySaari (approved):

shawkey-daniel (approved):
LGTM

PR #7191, opened 2021-06-09
Adds spec for additional error case from MHV API

patrickvinograd:
## Description of change
While looking at this issue in Sentry: http://sentry.vfs.va.gov/organizations/vsp/issues/44/?environment=staging&project=3&query=is%3Aunresolved+mobile&statsPeriod=14d

I noticed that the MHV client middleware was unable to handle one response from MHV - a 400 with no message body. The upshot of this is the same - that we can't get an MHV account type for this user - but this avoids us writing a spec against an internal error in the client implementation - that `string not matched` was coming from the mhv_errors middleware, trying to treat an empty string as a Hash. 

This seems to happen in staging because of garbage MHV IDs in MPI, so I don't know if it's essential to fix but seems like it will help with diagnosability. 

## Original issue(s)

### DIFF
--- a/lib/common/client/middleware/response/mhv_errors.rb
+++ b/lib/common/client/middleware/response/mhv_errors.rb
@@ -15,6 +15,7 @@ module Common
           #
           def on_complete(env)
             return if env.success?
+            return unless env[:body].is_a?(Hash)
 
             env[:body]['code'] = env[:body].delete('errorCode')
             env[:body]['detail'] = env[:body].delete('message')
--- a/spec/services/mhv_account_type_service_spec.rb
+++ b/spec/services/mhv_account_type_service_spec.rb
@@ -5,12 +5,14 @@ require 'rails_helper'
 RSpec.describe MHVAccountTypeService do
   subject { described_class.new(user) }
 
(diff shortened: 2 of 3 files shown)


lib/common/client/middleware/response/mhv_errors.rb
```diff
@@ -15,6 +15,7 @@ class MHVErrors < Faraday::Response::Middleware
           #
           def on_complete(env)
             return if env.success?
+            return unless env[:body].is_a?(Hash)
```
patrickvinograd, line 18: There's no point in proceeding if this has not been turned into a hash by earlier middleware, it will just raise a ` string not matched` type error.

spec/services/mhv_account_type_service_spec.rb
```diff
@@ -5,12 +5,14 @@
 RSpec.describe MHVAccountTypeService do
   subject { described_class.new(user) }
 
+  let(:unknown_error) { 'BackendServiceException: {:status=>400, :detail=>nil, :code=>"VA900", :source=>nil}' }
```
patrickvinograd, line 8: This is the best we can do since MHV doesn't provide any error payload whatsoever, just a bare status, in some conditions.

LindseySaari (approved):

PR #7194, opened 2021-06-09
API-6698 Remove claimant method from form_data hlr v2

ElleeB:
Not yet supporting claimant on HLR v2. Removed a method in form_data referencing the attribute.

Ticket: https://vajira.max.gov/browse/API-6698


yakschuss (approved):
🚀

PR #7195, opened 2021-06-09
[API-7815] -  ClaimsApi swap v2 docs to rswag

jw81:
## Description of change

Uses `rswag` to generate the `ClaimsApi` prototype V2 swagger docs.

## Original issue(s)
[API-7815](https://vajira.max.gov/browse/API-7815)


modules/claims_api/app/swagger/claims_api/v1/swagger.json
```diff
@@ -134,10 +134,12 @@
                         "properties": {
                           "id": {
                             "type": "string",
-                            "description": "Claim ID from EVSS"
+                            "description": "Claim ID from EVSS",
+                            "example": "8347210"
```
SlakrHakr, line 138: why was this v1 swagger doc updated with a bunch of example values?

modules/claims_api/app/swagger/claims_api/v2/swagger.json
```diff
@@ -0,0 +1,599 @@
...
+  },
+  "tags": [
+    {
+      "name": "Veteran Identifier",
+      "description": "Allows authenticated veterans and veteran representatives to retrieve a veteran's id."
```
SlakrHakr, line 11: nit: should Veteran be capitalized here? i feel like ive always seen it capitalized and i took that to be a respect type thing?

modules/claims_api/app/swagger/claims_api/v2/swagger.json
```diff
@@ -0,0 +1,599 @@
...
+                          }
+                        }
+                      }
+                    }
+                  }
+                }
+              }
+            }
+          },
+          "404": {
```
SlakrHakr, line 288: i know this isnt necessarily part of this specific ticket, but can we take this opportunity to remove 404 from the possible responses this claims endpoint can respond with?

modules/claims_api/spec/fixtures/usability_testing_claims.json
```diff
@@ -4,35 +4,35 @@
             "id": "600098193",
             "type": "evss_claims",
             "attributes": {
-                "evssId": 600098193,
-                "dateFiled": "2017-05-02",
-                "minEstDate": null,
-                "maxEstDate": null,
+                "evss_id": 600098193,
```
SlakrHakr, line 7: why was this changed to snake_case?

modules/claims_api/app/swagger/claims_api/v2/swagger.json
```diff
@@ -0,0 +1,599 @@
...
+            "description": "Resource Not Found",
+            "content": {
+              "application/json": {
+                "example": {
+                  "errors": [
+                    {
+                      "title": "Resource not found",
+                      "detail": "Resource not found",
```
SlakrHakr, line 296: nit: not critical here but i think when we get to the real implementation for this we should have something like "Veteran not found" for the detail here. i remember one of our interviews mentioned it was odd to have a 404 for an endpoint that is supposed to return a list and i agreed with him at the time. Just now realizing this situation should only occur if no Veteran is found with the passed in identifier.

spec/support/schemas/claims_api/claims.json
```diff
@@ -12,10 +12,12 @@
         "properties": {
           "id": {
             "type": "string",
-            "description": "Claim ID from EVSS"
+            "description": "Claim ID from EVSS",
+            "example": "8347210"
```
SlakrHakr, line 16: same here, why are all these example values suddenly here?

modules/claims_api/spec/requests/v2/rswag_veteran_identifier_request_spec.rb
```diff
@@ -0,0 +1,140 @@
+# frozen_string_literal: true
+
+require 'swagger_helper'
+require 'rails_helper'
+require_relative '../../support/swagger_shared_components'
+
+describe 'Veteran Identifier', swagger_doc: 'v2/swagger.json' do # rubocop:disable RSpec/DescribeClass
```
SlakrHakr, line 7: shouldnt this swagger_doc value (but v1 instead) be added to all the existing v1 rswag spec files?

spec/support/schemas/claims_api/claims.json
jw81: Because I explicitly added them here.
I did a side-by-side comparison of what the v2 docs look like today in dev, vs. what I was developing.
What's deployed today (`swagger_blocks`) provides example values in the schema definitions.
So I went and added the same example values to the schemas so that `rswag` would generate docs that included the same example values.

modules/claims_api/app/swagger/claims_api/v1/swagger.json
jw81: [Because I updated the schema file to include the example values that already exist today.](https://github.com/department-of-veterans-affairs/vets-api/pull/7195/files#r649196331)

modules/claims_api/spec/fixtures/usability_testing_claims.json
jw81: Because [the schema file](https://github.com/department-of-veterans-affairs/vets-api/blob/master/spec/support/schemas/claims_api/claims.json#L25) that `rswag` is using to compare the response to, has the attributes defined using snake_case.  `rswag` fails the testing because stuff like `evssId` != `evss_id`.

modules/claims_api/spec/requests/v2/rswag_veteran_identifier_request_spec.rb
jw81: We don't have to because of [this](https://github.com/rswag/rswag#supporting-multiple-versions-of-api).
> "_By default, the paths, operations and responses defined in your spec files will be associated with the **first** Swagger document in swagger_helper.rb._"

We can explicitly define that `swagger_doc` param on v1 classes if we want.  We just don't "have" to.  🤷

modules/claims_api/app/swagger/claims_api/v2/swagger.json
jw81: Ah, I try to always capitalize those.  I must've missed it.
Will update.

modules/claims_api/spec/requests/v2/rswag_veteran_identifier_request_spec.rb
SlakrHakr: gotcha. its probably something we'll need to do if another team decides to start using rswag i feel like?

modules/claims_api/spec/fixtures/usability_testing_claims.json
SlakrHakr: gotcha, ok, yeah, we'll end up needing new schema files for v2 for that sort of stuff when we actually go and implement it.

SlakrHakr (dismissed):

modules/claims_api/spec/requests/v2/rswag_veteran_identifier_request_spec.rb
jw81: Fair enough.  
[Updated here.](https://github.com/department-of-veterans-affairs/vets-api/pull/7195/commits/57b3a2c69b68f1426f22ae630273eb3289690439)

SlakrHakr (approved):

PR #7198, opened 2021-06-10
8953 upgrade to sidekiq 6

cbtkc:
## Description of change
Upgrades sidekiq (6.1.3) and sidekiq related gems.

The `worker_class` arg passed to the `SidekiqStatsInstrumentation::ClientMiddleware` `call` method is an actual class and no longer a symbol. `Object.const_get()` is no longer required.

Retrieving named tags can be done through `Sidekiq.logger.named_tags`. `Sidekiq::Logging` has been renamed.

In Redis 4.3, `exists` will no longer return a boolean, but rather an integer. You can adopt this newer behavior by setting `Redis.exists_returns_integer` which has been done in this PR, as sidekiq 6 requires this newer behavior.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#8953


LindseySaari (approved):
Nice work!

lib/sidekiq_stats_instrumentation/client_middleware.rb
```diff
@@ -5,7 +5,7 @@
 module SidekiqStatsInstrumentation
   class ClientMiddleware
     def call(worker_class, _job, _queue, _redis_pool)
-      klass = Object.const_get(worker_class)
+      klass = worker_class
```
annaswims, line 8: this change caused 

http://sentry.vfs.va.gov/organizations/vsp/issues/40527/?project=3&query=transaction%3ASidekiq&statsPeriod=14d

PR #7201, opened 2021-06-10
API-7873 Handle row number constants in hlr v2 structure

ElleeB:
While wiring up the contestable issues soc date printing, it was pointed out that we would be better off using constants for number of rows in a column on each the first and second pages of contestable issues. Additionally, these constants should be managed by the hlr v2 `structure` file.

Ticket: https://vajira.max.gov/browse/API-7873

### DIFF
--- a/modules/appeals_api/app/services/appeals_api/pdf_construction/higher_level_review/v2/form_fields.rb
+++ b/modules/appeals_api/app/services/appeals_api/pdf_construction/higher_level_review/v2/form_fields.rb
@@ -4,9 +4,6 @@ module AppealsApi
   module PdfConstruction
     module HigherLevelReview::V2
       class FormFields
-        FIRST_PAGE_ISSUES_ROW_COUNT = 7
-        SECOND_PAGE_ISSUES_ROW_COUNT = 6
-
         def middle_initial
           'form1[0].#subform[2].Veteran_Middle_Initial1[0]'
         end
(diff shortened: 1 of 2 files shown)


modules/appeals_api/app/services/appeals_api/pdf_construction/higher_level_review/v2/form_fields.rb
```diff
@@ -4,8 +4,8 @@ module AppealsApi
   module PdfConstruction
     module HigherLevelReview::V2
       class FormFields
-        FIRST_PAGE_ISSUES_ROW_COUNT = 7
```
yakschuss, line 7: I don't think we should be assigning the value of the constant to the Structure constant. Can we just reference the constant from Structure where it's relevant?

modules/appeals_api/app/services/appeals_api/pdf_construction/higher_level_review/v2/form_fields.rb
```diff
@@ -170,6 +167,9 @@ def issue_decision_date_fields(index)
 
         # rubocop:disable Metrics/MethodLength
         def boxes
+          number_of_issues_on_second_pg = Structure::NUMBER_OF_ISSUES_SECOND_PAGE
```
ElleeB, line 170: adding these variables to avoid having to break up the blocks below into multiple lines since adding the constant knocks it over the max line length - easier to read this way imo.

cilestin (approved):

nathanbwright (approved):

PR #7204, opened 2021-06-10
Error when fetching scheduling configurations

ajmagdub:
## Description of change

Updated mobile facility service class to pass camel case facility ids and cc enabled query parameters to the VAOS Service GET `/facilities/v2/scheduling/configurations` endpoint.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#25903

When testing against the mobile facility service in SQA submitting a ccEnabled query parameter did not make a difference in the scheduling configuration resources returned. Regardless of what ccEnabled was both CommiunityCare = true and CommunityCare = false configurations were returned. 

Also arranging the order of the mobile facility service query parameters where the ccEnabled is the first query parameter causes the mobile facility service to return a 400 error (missing required parameter: facilityIds) even though facilityIds is being sent.


LindseySaari (approved):

modules/vaos/app/services/vaos/v2/mobile_facility_service.rb
```diff
@@ -22,10 +22,10 @@ def get_facilities(query_params, pagination_params = {})
...
       end
 
-      def get_scheduling_configurations(facility_ids, cc_enabled, pagination_params = {})
+      def get_scheduling_configurations(facility_ids, cc_enabled = nil, pagination_params = {})
         params = {
-          facility_ids: facility_ids,
-          cc_enabled: cc_enabled
+          facilityIds: facility_ids,
+          ccEnabled: cc_enabled
         }.merge(page_params(pagination_params)).compact
```
saneshark, line 29: why are we passing camelcase attributes? We have a middleware that converts snakecase to camelcase automatically. You should be able to work with snakecase everywhere in rails and just let the middleware do its thing.

modules/vaos/app/controllers/vaos/v2/scheduling_controller.rb
```diff
@@ -4,23 +4,20 @@ module VAOS
...
-      def configuration_params
-        params.require(:facility_ids)
-        params.permit(:cc_enabled, facility_ids: [])
-      end
-
       def mobile_facility_service
         VAOS::V2::MobileFacilityService.new(current_user)
       end
+
+      def csv_facility_ids
```
saneshark, line 17: csv? when did we start working with files instead of JSON? 🤔

saneshark (commented):
a few comments

modules/vaos/app/controllers/vaos/v2/scheduling_controller.rb
ajmagdub: these are for the query parameters we send to VAMF Mobile Facilities Service. The Mobile Facility Service requires the list to be comma separated values. We still work with JSON but in the request and response bodies.

modules/vaos/app/services/vaos/v2/mobile_facility_service.rb
ajmagdub: The original bug report was that the facility_ids query parameter sent to VAMF was not be sent as camel case (as required by VAMF). The changes were made to comply with the Mobile Facility Service swagger docs. 

Olive branch does not seem to change snake/camel case between vets website and vets api, but not between vets api and VAMF.

modules/vaos/app/controllers/vaos/v2/scheduling_controller.rb
saneshark: ah, my bad I thought it was a request body. This makes sense... i was thinking the faraday middleware handles those, but it does not.

modules/vaos/app/services/vaos/v2/mobile_facility_service.rb
saneshark: ah, my bad I thought it was a request body. This makes sense... i was thinking the faraday middleware handles those, but it does not.

modules/vaos/app/controllers/vaos/v2/scheduling_controller.rb
saneshark: k, go ahead and merge this -- but we've ran into this problem before and used FlatParamsEncoder with faraday to convert.

PR #7210, opened 2021-06-10
API-4183 - Clean up invalid json error

nathanbwright:
## Description of change
Instead of returning the entire invalid json body, return the text that states the body is invalid.

## Original issue(s)
https://vajira.max.gov/browse/API-4183


cilestin (approved):
Huzzah!

PR #7213, opened 2021-06-11
25916 mobile fix pre cache claims error handling

kreek:
## Description of change
Fixes claims and appointments error message parsing. Adds specs for each case. Removes some duplication in the claims request specs.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#25916


patrickvinograd (approved):
LGTM

PR #7215, opened 2021-06-11
WIP for RI

grossb1:
## Description of change

Permit validation has been updated to validate correctly against nested object/arrays.

## Original issue(s)

### DIFF
--- a/modules/vaos/app/controllers/vaos/v2/appointments_controller.rb
+++ b/modules/vaos/app/controllers/vaos/v2/appointments_controller.rb
@@ -58,8 +58,15 @@ module VAOS
       end
 
       def create_params
-        params.permit(:kind, :status, :location_id, :clinic, :reason, :slot, :contact,
-                      :service_type, :requested_periods)
+        params.permit(:kind,
+                      :status,
+                      :location_id,
+                      :clinic,
+                      :reason,
+                      :service_type,
+                      slot: %i[id start end],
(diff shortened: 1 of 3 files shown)


modules/vaos/app/controllers/vaos/v2/appointments_controller.rb
```diff
@@ -58,8 +58,15 @@ def appointment_params
...
-                      :service_type, :requested_periods)
+        params.permit(:kind,
+                      :status,
+                      :location_id,
+                      :clinic,
+                      :reason,
+                      :service_type,
+                      slot: %i[id start end],
+                      contact: [telecom: %i[type value]],
+                      requested_periods: %i[start end])
```
saneshark, line 69: This is everything according to the schema right? Is telecom the only form of contact info?

If the answer is "I am not sure" I might propose we just do `params.permit!` because the way we're deserializing a class and deserializing we're essentially preventing things such as ICN getting mass assigned anyway (we're loading it from the user object vs from the payload potentially, etc).

saneshark (dismissed):
LGTM

modules/vaos/app/controllers/vaos/v2/appointments_controller.rb
jbalboni: That looks like all the fields to me, but there will be three or four more fields to add later. In case that makes a difference.

saneshark (approved):

PR #7218, opened 2021-06-11
Make rswag more easily configurable across modules

jw81:
Just tinkering trying to find a way to make this more easily configurable/usable across different modules.


SlakrHakr (approved):

hinzed1127 (approved):

SlakrHakr:
not sure if this is still valuable considering appeals is working on using rswag now it looks like?
https://github.com/department-of-veterans-affairs/vets-api/pull/7484

PR #7219, opened 2021-06-11
API-6101 Lighthouse Token Validation - Revoke session if ICN mismatch

dbrown-va:
## Description of change
Lighthouse Token Validation - Warning prior to revoke session if ICN mismatch

If the Okta profile ICN does not match the MPI returned ICN, then adds method to call Okta to revoke the entire session (includes both access and refresh tokens) and set the session to nil forcing the 401 response.  For this PR, only logs warning to sentry to get an accurate expectation of these occurrences. 

This occurs in cases where an ICN is updated in some way and instead is listed in the historical ICN's portion)

Also reverts back to the token.identifiers.uuid for Redis cache key since the uuid for SSOe + DSLogon does not behave in the same was the Lighthouse SAML proxy method.

## Original issue(s)
https://vajira.max.gov/browse/API-6101

* Considering making this configurable to only apply in Dev to allow additional test prior to promotion
* Sentry logging to warn for future revocations was added to detect any anomalies


app/controllers/openid_application_controller.rb
```diff
@@ -50,13 +50,13 @@ def authenticate_token
     if token.client_credentials_token? || token.ssoi_token?
       populate_payload_for_launch_patient_scope if token.payload['scp'].include?('launch/patient')
       populate_payload_for_launch_scope if token.payload['scp'].include?('launch')
-
       return true
     end
 
     return false if @session.uuid.nil?
 
     @current_user = OpenidUser.find(@session.uuid)
```
beaugrantham-va, line 58: Does this need to explicitly return false since it's no longer the last line in the method?

app/controllers/openid_application_controller.rb
```diff
@@ -137,15 +137,22 @@ def establish_session(profile)
...
-  def uuid(profile)
-    profile['uuid'] || token.identifiers.uuid
+  # Ensure the Okta profile ICN continues to match the MPI ICN
+  # If mismatched, revoke in Okta, set @session to nil, and return false
+  # POA support (profile['icn'].nil?)
+  def confirm_icn_match(profile)
+    if profile['icn'].nil? || @current_user.icn == profile['icn']
+      true
+    else
+      Okta::Service.new.clear_user_session(token.identifiers.okta_uid)
```
beaugrantham-va, line 151: Is it worth adding a prometheus metric here so we can see historical occurrences without digging through logs?

app/controllers/openid_application_controller.rb
```diff
@@ -137,15 +137,22 @@ def establish_session(profile)
...
+      @session = nil
+      log_message_to_sentry('Profile ICN mismatch detected. Clearing user session.', :error)
```
beaugrantham-va, line 153: What is `:error` here?

app/controllers/openid_application_controller.rb
beaugrantham-va: Nevermind, I see that `clear_user_session` uses `with_monitoring` which should accomplish this.

app/controllers/openid_application_controller.rb
dbrown-va: Log level.  I went back and forth between error and warn since it is a scenario we are designing to handle vs an exception.  But the ICN mismatch will return a 401, so thought that more aligned with an error.

app/controllers/openid_application_controller.rb
dbrown-va: My thought was the `confirm_icn_match` should catch that if nil and return false as well.

However, I did just add a `&.` to ensure the check is nil safe.

app/controllers/openid_application_controller.rb
dbrown-va: Switching to log level `:warn` now that this will detect and log to sentry only, not actually clear the Okta user session.

lukemarkwordtlibertyits:
I have confirmed the logging functionality and a variety of regression tests

```
 ✓ Patient Token. Valid. With Launch/Patient scp. Return 200 with appropriate fields
```

app/models/openid_user_identity.rb
```diff
@@ -16,7 +16,7 @@ class OpenidUserIdentity < ::UserIdentity
   # @return [OpenidUserIdentity]
   def self.build_from_okta_profile(uuid:, profile:, ttl:)
     identity = new(
-      uuid: profile['uuid'] || uuid,
+      uuid: uuid,
```
beaugrantham-va, line 19: The code here was reverted but the comment on line 13 still references the fallback behavior.

beaugrantham-va (approved):

app/models/openid_user_identity.rb
dbrown-va: Removed!

beaugrantham-va (approved):

PR #7226, opened 2021-06-14
Allow download of 10-10cg with no veteran name

kevinmirc:
## Description of change

Downloading a pre-filled 10-10CG should not require fully valid 10-10cg data. Allow pdf download even if veteran's name is missing.

## Original issue(s)
- https://github.com/department-of-veterans-affairs/va.gov-team/issues/14287

### DIFF
--- a/app/controllers/v0/caregivers_assistance_claims_controller.rb
+++ b/app/controllers/v0/caregivers_assistance_claims_controller.rb
@@ -59,7 +59,10 @@ module V0
     end
 
     def file_name_for_pdf(veteran_data)
-      "10-10CG_#{veteran_data['fullName']['first']}_#{veteran_data['fullName']['last']}.pdf"
+      veteran_name = veteran_data.try(:[], 'fullName')
+      first_name = veteran_name.try(:[], 'first') || 'First'
+      last_name = veteran_name.try(:[], 'last') || 'Last'
+      "10-10CG_#{first_name}_#{last_name}.pdf"
     end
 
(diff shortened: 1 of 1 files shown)


app/controllers/v0/caregivers_assistance_claims_controller.rb
```diff
@@ -59,7 +59,10 @@ def initialize_claim
     end
 
     def file_name_for_pdf(veteran_data)
-      "10-10CG_#{veteran_data['fullName']['first']}_#{veteran_data['fullName']['last']}.pdf"
+      veteran_name = veteran_data.try(:[], 'fullName')
+      first_name = veteran_name.try(:[], 'first') || 'First'
+      last_name = veteran_name.try(:[], 'last') || 'Last'
```
LindseySaari, line 64: Is there a possibility that there could be multiple files named "10-10CG_First_Last.pdf" given that multiple results returned nil for first and last? Is there a way to make the filenames more unique?

kevinmirc: Yes, there could be duplicate names, but this is purely used to send the client a suggested name for the file they are downloading. Whenever we write/reference files locally, we use the claim's uuid.

LindseySaari: Ok great, thanks for clarifying!

LindseySaari (approved):

PR #7230, opened 2021-06-14
fix retry monitoring

yakschuss:
[API-8002](https://vajira.max.gov/browse/API-8002)

Sidekiq provides a nil value on the first retry check. We don't want to add +1 to 0 retries, so we return false early.

### DIFF
--- a/lib/sidekiq/retry_monitoring.rb
+++ b/lib/sidekiq/retry_monitoring.rb
@@ -13,6 +13,8 @@ module Sidekiq
     private
 
     def should_notify?(worker, job)
+      return false unless job['retry_count']
+
       # retry_count is incremented after all middlewares are called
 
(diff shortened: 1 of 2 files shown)


LindseySaari (approved):

cilestin (approved):

lib/sidekiq/retry_monitoring.rb
```diff
@@ -13,6 +13,8 @@ def call(worker, params, _queue)
     private
 
     def should_notify?(worker, job)
+      return false unless job['retry_count']
```
annaswims, line 16: @yakschuss / @LindseySaari Do you think that this is part of the problem I'm seeing here - https://github.com/department-of-veterans-affairs/va.gov-team/issues/26326

PR #7233, opened 2021-06-15
Update security properties for claims v1 swagger

hinzed1127:
## Description of change

This updates the [security definitions](https://swagger.io/docs/specification/authentication/oauth2/) in claims v1 swagger to properly reflect that it's using OAuth for authentication.

## Original issue(s)
https://vajira.max.gov/browse/API-7860

In and of itself, this doesn't give us anything extra within swagger ui. The developer portal doesn't currently have the OAuth flow implemented from swagger ui, and even when using https://editor.swagger.io/, I don't think that it'll redirect properly.

Screen Shot 2021-06-14 at 8 07 07 PM

This will require dev work on the portal and registering the portal as a valid OAuth consumer for in-browser requests to work properly. TBD if that's a route us or other teams are interested in pursuing, but this lays the groundwork for that.

### DIFF
--- a/modules/claims_api/app/swagger/claims_api/v1/swagger.json
+++ b/modules/claims_api/app/swagger/claims_api/v1/swagger.json
@@ -25,6 +25,24 @@
   ],
   "components": {
     "securitySchemes": {
+      "bearer_token": {
+        "type": "http",
+        "scheme": "bearer"
+      },
+      "productionOauth": {
+        "type": "oauth2",
+        "description": "This API uses OAuth 2 with the authorization code grant flow. [More info](https://developer.va.gov/explore/authorization?api=claims)",
+        "flows": {
+          "authorizationCode": {
+            "authorizationUrl": "https://api.va.gov/oauth2/authorization",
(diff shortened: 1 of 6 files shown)


modules/claims_api/app/swagger/claims_api/v1/swagger.json
```diff
@@ -25,11 +25,19 @@
   ],
   "components": {
     "securitySchemes": {
-      "bearer_token": {
-        "type": "http",
-        "name": "token",
-        "scheme": "bearer",
-        "bearer_format": "JWT"
+      "sandboxOauth": {
```
hinzed1127, line 28: This name is arbitrary and can be updated to something different if desired

LindseySaari (approved):

SlakrHakr (changes requested):
it breaks the curl generator on the dev portal :(
i thought i had tried everything but apparently i missed the: bearer token + curl generator feature
not sure how we can have these live side by side but this needs to be investigated, what is the portal needing for that bearer token + curl generator to function properly?

SlakrHakr (approved):

PR #7236, opened 2021-06-15
Adds metrics and logging for OAuth sessions

patrickvinograd:
## Description of change
Adds logging of obscured token and user info when new token is
presented. Adds obscured token to request logging. Adds statsd metrics
for new vs. refresh tokens for use in alerting.

More or less follows the approach used in VA.gov:
* Log a bunch of user info when session is first established. I'm open to not logging the secid directly but VA.gov does log ICN and I believe SECID as well. These are linkable PII but considered fair game for logging. 
* Log user UUID and obscured token in payload of each request.

I also reworked the statsd metric to allow us to distinguish between token issued immediately after authenticating vs. refresh tokens so we can define some alerts, and combined the credential type metric with that one. 

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/25580
https://github.com/department-of-veterans-affairs/va.gov-team/issues/25118

* Does add linkable PII - namely SECID - to logging on initial session establishment and on each refresh.

### DIFF
--- a/config/initializers/statsd.rb
+++ b/config/initializers/statsd.rb
@@ -235,6 +235,11 @@ StatsD.increment('iam_ssoe_oauth.inactive_session', 0)
 
 StatsD.increment('iam_ssoe_oauth.auth_type', 0)
 
+%w[IDME MHV DSL].each do |cred|
+  StatsD.increment('iam_ssoe_oauth.session', 0, tags: ['type:new', "credential:#{cred}"])
+  StatsD.increment('iam_ssoe_oauth.session', 0, tags: ['type:refresh', "credential:#{cred}"])
+end
+
 # init VEText Push Notifications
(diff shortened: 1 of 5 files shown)


config/initializers/statsd.rb
```diff
@@ -235,6 +235,11 @@
 
 StatsD.increment('iam_ssoe_oauth.auth_type', 0)
 
+%w[IDME MHV DSL].each do |cred|
```
LindseySaari, line 238: Can you reference the STATSD_OAUTH_SESSION_KEY constant here?

patrickvinograd: Done

LindseySaari (approved):

kreek (approved):
LGTM 🚀

PR #7241, opened 2021-06-16
Kevin.cranmer/api 7592 postman testing

KevinCranmer:
## Description of change
New automated Postman testing functionality added for the Benefits Intake and Forms APIs.

## Original issue(s)
https://vajira.max.gov/browse/API-7592
https://vajira.max.gov/browse/API-7808

Not affecting production code.


mviterna (approved):
Form API looks fine. I noticed that the SHA256 in the console does not match what is actually returned, but I think we were going to skip that test anyway, so it shouldn't matter. All tests seemed to pass

gpbowman-git (approved):
This looks good to me. I will pull this in from master and get it set up asap.

PR #7246, opened 2021-06-16
API-7795 legacy issues endpoint

ElleeB:
This is a WIP.
The result is a new controller created to manage the call to Caseflow, which returns a Veteran's legacy appeals.
Required params = either SSN or FILE_NUMBER.
The endpoint that we intend to call does not exist in Caseflow yet, so I've created a pretend cassette to test with.

Ticket: https://vajira.max.gov/browse/API-7795


modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/legacy_appeals_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    render_response(caseflow_response)
+  end
+
+  private
+
+  attr_reader :caseflow_response
+
+  def get_legacy_appeals_from_caseflow
+    @caseflow_response = Caseflow::Service.new.get_legacy_appeals headers: headers
```
yakschuss, line 25: We should probably add a timeout to this request, right? We can't be sure that the network request will always succeed

PR #7250, opened 2021-06-16
Update claims OAuth scopes for sandbox

hinzed1127:
## Original issue(s)


LindseySaari (approved):

LindseySaari (dismissed):
Looks like a flapping spec is failing, but these changes look good!

LindseySaari (approved):

SlakrHakr (changes requested):
This shouldn't be needed anymore.
We were finally able to prove that the existing scopes are all that is necessary for this to operate.
https://sandbox-api.va.gov/services/claims/docs/swagger/index.html

PR #7254, opened 2021-06-17
Mobile handle Faraday::ResourceNotFound errors when getting contact info

kreek:
## Description of change
Some calls to the Vet360 service from the IAMUser model are raising Faraday::ResourceNotFound errors which are in turn raising 500s rather than 404s. PR adds a rescue block in IAMUser to correctly map the error. This fix would be mobile specific, it could also be handled lower down in the `VAProfile::Service` or even the base `Common::Client`?

## Original issue(s)
http://sentry.vfs.va.gov/organizations/vsp/issues/22541/?environment=staging&project=3&query=is%3Aunresolved


app/models/iam_user.rb
```diff
@@ -68,4 +68,10 @@ def va_patient?
   def set_expire
     redis_namespace.expireat(REDIS_CONFIG[:iam_user][:namespace], expiration_timestamp)
   end
+  
+  def vet360_contact_info
+    super
+  rescue Faraday::ResourceNotFound => e
```
LindseySaari, line 74: Looks like a linting error here, but otherwise this looks good!

LindseySaari (approved):

PR #7255, opened 2021-06-17
Rails initializer class loading

thilton-oddball:
## Description of change
This PR gets rid of the deprecation warnings (that would eventually become errors) related to autoloading constants during initialization. By wrapping code in initializer files in 'Rails.application.reloader.to_prepare do' blocks, it eliminates the warning and allows any changes made to classes to get reloaded in the development environment.

Reloading does not reboot the application, and therefore code executed during initialization does not run again. So, if you reload, changes to classes in the initializer won't be reflected in that stale Class object because the initializer only runs on reboot. Therefore, code that references a reloadable class or module, but that is not executed again on reload, becomes stale.

Co-authored by: @LindseySaari


LindseySaari:
@thilton-oddball I think this may be ready to go, but I wasn't 100% sure on the rubocop changes

.rubocop.yml
```diff
@@ -87,7 +87,7 @@ Style/ZeroLengthPredicate:
   Enabled: false
 
 Layout/LineLength:
-  Max: 120
+  Max: 125
```
ericboehs, line 90: 80 and 120 are a pretty standard line lengths. Why the change to 125?

thilton-oddball: I can change it back to 120, I mainly just tweaked it because one of the lines that had to change ended up being like 122 and Rubocop was giving me grief.

modules/mobile/config/initializers/statsd.rb
```diff
@@ -88,17 +100,6 @@
 StatsD.increment('mobile.profile.link_account.success', 0)
 StatsD.increment('mobile.profile.link_account.failure', 0)
```
ericboehs, line 101: Having a hard time figuring out why this file has so many changes. It looks the same. 🤔

thilton-oddball: I was taking lines out of the `Rails.application.reloader.to_prepare do` block and then putting them back in trying to figure out what needed to be in there and what didn't, so I probably ended up getting things out of the order they were originally in.

ericboehs: Why not just put them all in there?

thilton-oddball: We certainly could I guess. I'll make that change and push it up.

.rubocop.yml
```diff
@@ -129,6 +129,7 @@ Metrics/BlockLength:
     - 'modules/*/spec/**/*.rb'
     - 'modules/*/app/swagger/**/*.rb'
     - 'modules/*/config/routes.rb'
+    - 'modules/mobile/config/initializers/statsd.rb'
```
ericboehs, line 132: Can you add this to the `.rubocop_todo.yml` instead via `rubocop --auto-gen-config`? You may have to run `rubocop -a` afterwards to correct any redundant disables.

ericboehs:
You forgot the `rubocop -a` 😺

thilton-oddball:
https://gfycat.com/bouncycarefuldormouse-patrick-stewart-star-trek-facepalm-picard

ericboehs (approved):

PR #7256, opened 2021-06-17
empty slot hash is dropped from post appointment request object

grossb1:
## Description of change

vaos service does not accept an empty object for the `slot` field in the post appointment request body.  The code has been updated to drop an empty `slot` object from the request body, should it exist

## Original issue(s)

Tests were added to verify that an empty `slot` object is dropped from the request

### DIFF
--- a/modules/vaos/spec/models/v2/appointment_form_spec.rb
+++ b/modules/vaos/spec/models/v2/appointment_form_spec.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe VAOS::V2::AppointmentForm, type: :model do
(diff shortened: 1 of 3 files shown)


modules/vaos/spec/models/v2/appointment_form_spec.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+# frozen_string_literal true
```
cferris32, line 3: Looks like this might be able to get removed due to line 1

grossb1: Good catch

ajmagdub (approved):
LGTM

saneshark (approved):
LGTM

PR #7262, opened 2021-06-17
API-8092 - Only display HLR v2 notice in lower envs

nathanbwright:
## Description of change
Only show the HLR `See v2...` notice in the lower envs.

## Original issue(s)
https://vajira.max.gov/browse/API-8092


cilestin (dismissed):

yakschuss (approved):
This does what it's supposed to do! I worry about the naming of `DOC_VISIBILITY_SETTING` in the context of deprecation and then the v2 notice... Does this read easily to you? I'm having trouble with it

nathanbwright:
> This does what it's supposed to do! I worry about the naming of `DOC_VISIBILITY_SETTING` in the context of deprecation and then the v2 notice... Does this read easily to you? I'm having trouble with it

Yes, I'm sure that the naming could be improved... What do you suggest?

cilestin (approved):

PR #7265, opened 2021-06-18
[API-7793] - ClaimsAPI add logic for finding Veteran ID

jw81:
## Description of change

Replaces the static prototype logic used during "usability testing" with real logic for returning a Veteran id.
Also introduces patterns for ~using [`jbuilder`](https://github.com/rails/jbuilder)~ and [`ActiveModel Validations`](https://api.rubyonrails.org/v6.1.3.1/classes/ActiveModel/Validations.html).

## Original issue(s)
[API-7793](https://vajira.max.gov/browse/API-7793)


modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
```diff
@@ -5,18 +5,19 @@
...
         veteran = find_veteran(params)
 
-        render json: { id: veteran[:id] }
+        raise ::Common::Exceptions::ResourceNotFound unless veteran_icn_found?(veteran)
+
+        user_is_the_veteran   = current_user_is_the_veteran?(veteran: veteran, user: @current_user)
+        user_is_a_veteran_rep = current_user_is_a_veteran_representative?(@current_user)
+        raise ::Common::Exceptions::Unauthorized unless user_is_the_veteran || user_is_a_veteran_rep
+
+        render json: { id: veteran.mpi.icn }
```
SlakrHakr, line 20: I wanted to investigate jbuilder when we started implementing all this stuff. Worth looking into for this simple endpoint just as a concept for future larger complex endpoints?
https://github.com/rails/jbuilder

modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
```diff
@@ -5,18 +5,19 @@
 module ClaimsApi
   module V2
     class VeteranIdentifierController < ClaimsApi::V2::ApplicationController
-      # TODO: REMOVE BEFORE IMPLEMENTATION
-      skip_before_action :authenticate, only: %i[find]
-      ICN_FOR_TEST_USER = '1012667145V762142'
-      # TODO: REMOVE BEFORE IMPLEMENTATION
-
       def find
         raise ::Common::Exceptions::Unauthorized if request.headers['Authorization'].blank?
```
SlakrHakr, line 9: Does this return a 401? Just making sure its a 401 and not a 403

modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
```diff
@@ -5,18 +5,19 @@
...
 
         validate_request
         veteran = find_veteran(params)
 
-        render json: { id: veteran[:id] }
+        raise ::Common::Exceptions::ResourceNotFound unless veteran_icn_found?(veteran)
+
+        user_is_the_veteran   = current_user_is_the_veteran?(veteran: veteran, user: @current_user)
+        user_is_a_veteran_rep = current_user_is_a_veteran_representative?(@current_user)
+        raise ::Common::Exceptions::Unauthorized unless user_is_the_veteran || user_is_a_veteran_rep
```
SlakrHakr, line 18: This should be a 403, maybe the ::Common::Exceptions::Forbidden class?
Because the user has authenticated fine, but they are forbidden from getting this Veteran's ID

modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
jw81: Oh dang it.  I should remove that completely right?  Because the usual auth logic is now back in place.

modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
jw81: Ah, that's a really good point!  
Good catch.

modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
```diff
@@ -5,18 +5,19 @@
...
     class VeteranIdentifierController < ClaimsApi::V2::ApplicationController
-      # TODO: REMOVE BEFORE IMPLEMENTATION
-      skip_before_action :authenticate, only: %i[find]
-      ICN_FOR_TEST_USER = '1012667145V762142'
-      # TODO: REMOVE BEFORE IMPLEMENTATION
-
       def find
         raise ::Common::Exceptions::Unauthorized if request.headers['Authorization'].blank?
 
         validate_request
```
SlakrHakr, line 11: These validations are good enough for this endpoint but I wanted to investigate something like ActiveModel::Validations eventually, especially for our larger more complex endpoints.
Worth using this simpler endpoint as a proof of concept?
https://blog.kalina.tech/2019/12/nested-api-parameter-validation-in-rails.html

modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
SlakrHakr: Actually, is this still necessary? shouldnt `authenticate` handle this use case and the use case where the auth token is there but invalid?

jw81: > Actually, is this still necessary? shouldnt `authenticate` handle this use case and the use case where the auth token is there but invalid?

Nope, it's not needed.
I just forgot to remove it.
[Removed here.](https://github.com/department-of-veterans-affairs/vets-api/pull/7265/commits/7328e7c37a8c4a6fe45649819419ccaf06020d30)

modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
jw81: Introduced and used "jbuilder" [here](https://github.com/department-of-veterans-affairs/vets-api/pull/7265/commits/729ce10f62d2a940f3f4e572e0e9a3f545543db8).

modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
jw81: Used the example link you provided to layout an initial pattern using `ActiveModel::Validations`.

modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
```diff
@@ -1,40 +1,34 @@
...
-
-        validate_request
         veteran = find_veteran(params)
 
-        render json: { id: veteran[:id] }
+        raise ::Common::Exceptions::ResourceNotFound unless veteran_icn_found?(veteran)
+
+        user_is_the_veteran   = current_user_is_the_veteran?(veteran: veteran, user: @current_user)
+        user_is_a_veteran_rep = current_user_is_a_veteran_representative?(@current_user)
+        raise ::Common::Exceptions::Forbidden unless user_is_the_veteran || user_is_a_veteran_rep
```
SlakrHakr, line 18: should these come before actually searching for the Veteran? i think they, logically, can function before at least.
im just wondering if there is a small vulnerability here where someone who is authenticated but not authorized could find out if a Veteran existed in our system by hitting this endpoint (they would get a 404 even if they werent either the Veteran themself or an accredited representative)

modules/claims_api/app/controllers/claims_api/v2/veteran_identifier_controller.rb
```diff
@@ -1,40 +1,34 @@
...
+      def validate_params
+        validator = ClaimsApi::V2::ParamsValidation::VeteranIdentifier.validator(params)
 
-      def validate_ssn!(ssn)
-        return if ssn.match?(/^\d{9}$/)
+        return if validator.valid?
 
-        raise ::Common::Exceptions::InvalidFieldValue.new('ssn', ssn)
+        validation_error = validator.errors.first
+        raise ::Common::Exceptions::InvalidFieldValue.new(validation_error.attribute, validation_error.type)
```
SlakrHakr, line 31: now that we're using this "validation" approach, i wonder if we can actually just pass that "validator" above to this common exceptions class instead:
https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/common/exceptions/validation_errors.rb#L9

jw81: Ended up having to create a `Common::Exceptions::ValidationErrorsBadRequest` class and using that because we wanted a validation failure to return a `400` error code, rather than the `422` that `Common::Exceptions::ValidationErrors` provides.
That work was done [here](https://github.com/department-of-veterans-affairs/vets-api/pull/7265/commits/f832e1e5c1b967fa8314375bba732d8d5d973911).

SlakrHakr (approved):

PR #7273, opened 2021-06-21
Moving PDF Job for Dependents Application Submissions

kathleencrawford:
## Description of change

During integration testing of the 686c/674 (Dependents Application) it was discovered that when a failure occurs and the job is claim is not established, the PDF of the form was still being generated and submitted to VBMS.  This PR changes the process flow so that the PDF only gets generated if a fatal exception does not occur.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#24891

This PR does not add any new functionality, it just changes the order of operations.  Spec tests have been updated, but no new configuration or settings have been added. 

- [x] Local testing
- [x] Specs updated
- [x] Staging testing planned


LindseySaari (approved):
💯

PR #7283, opened 2021-06-22
Contact info update email

lihanli:
## Description of change
send users notification emails when their contact info changes. email updates are sent to both the old and new email addresses.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/25865

tested in staging


app/models/user.rb
```diff
@@ -355,15 +355,19 @@ def vet360_contact_info
     @vet360_contact_info ||= VAProfileRedis::ContactInformation.for_user(self)
   end
 
+  def va_profile_email
+    vet360_contact_info&.email&.email_address
+  end
+
   def all_emails
-    vet360_email =
+    the_va_profile_email =
```
ericboehs, line 363: I don't feel like this rescue is necessary. Thoughts?

app/models/user.rb
```diff
@@ -355,15 +355,19 @@ def vet360_contact_info
...
       begin
-        vet360_contact_info&.email&.email_address
+        va_profile_email
       rescue
         nil
       end
 
-    [vet360_email, email]
+    [the_va_profile_email, email]
```
ericboehs, line 370:
```suggestion
    [va_profile_email, email]
```

I think you can just get rid of 363-369.

ericboehs (approved):

app/models/user.rb
lihanli: if a user isn't initialized in va profile then it will raise an error when you try to fetch their va profile contact info

bosawt: What throws the error in this case? is it the `VAProfileRedis::ContactInformation` class? Wondering if we should implement the error handling there, and then we'll be able to safely call `vet30_contact_info`

bosawt (approved):
Looks good!
Only feedback is similar to Eric's comment, wondering if the rescue for `va_profile_email` is necessary, and instead if we can move this error handling into the class responsible.  But I don't think that adversely impacts functionality either way so I'll leave it up to you if you want to change that.

app/models/user.rb
lihanli: https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/va_profile/contact_information/service.rb#L155
this is where the error is raised. the error handling would be a separate issue but you would need to look into whether or not it's a good idea to suppress all errors every time `VAProfileRedis::ContactInformation` is used

PR #7286, opened 2021-06-23
26500 Feature toggles cache to use Redis cache 

dillo:
## Description of change

- Switching over to Redis cache for feature toggles caching. We want to test this in the staging environment and gather relevant metrics before deciding if we wish to move forward with this in production or not.
- Currently, the vets-api is caching the feature toggle endpoint using ActiveSupport::Cache::MemoryStore(usually implemented in the development environment).
- The memory store cache is only available to the individual application process which happens to be running the vets-api service.
- Another server/container/location running the vets-api service will not have access to this memory store.
- Moving the caching strategy to use Redis will centralize the cache-store and allow all instances of the vets-api processes to gain access to the feature toggle cache.
- This strategy will also allow us to horizontally scale our caching strategy: The feature count and website traffic to va.gov will continue to grow in the future.
## Original issue(s)
department-of-veterans-affairs/va.gov-team#26500

- [x] RSpec tests written


annaswims:
I don't think this should be configured to be in our rails cache Redis, which I think is configured to be `last-in first-out (LIFO)`. It would be disastrous for stability if the values of the feature toggles were lost. 

While the query to get feature toggles values is one of our most frequent queries, it runs in under 1ms and we have no concerns of the database scaling.

dillo:
> I don't think this should be configured to be in our rails cache Redis, which I think is configured to be `last-in first-out (LIFO)`. It would be disastrous for stability if the values of the feature toggles were lost.
> 
> While the query to get feature toggles values is one of our most frequent queries, it runs in under 1ms and we have no concerns of the database scaling.

Thanks for your valued opinion @annaswims 👍

annaswims:
But if you really want it in Redis @dginther might have advice on a different redis that would be better.  We're teating the Sidekiq instance of redis as a permanent data store.

annaswims:
And, as I said to @dillo when we chatted yesterday I think the ultimate fix is that the front end is making way too many requests for the feature toggle values.  The front end is making API requests for every feature toggle value multiple times a page load.  Ideally, the front end would either 1) make an api call for only the relevant feature toggles OR 2) make an API call for all feature toggles and cache it for a while.

dillo:
@ericboehs The main reasons for wanting to **learn** more about caching with Redis in staging for feature toggles are as follows:
 
- The feature toggle endpoint was rendering our pilot Questionnaire app inoperable in a mobile browser in production for actual veterans participating in the trial. Turning off the calls from the UI seems to have fixed this problem. The new Check-In project UI was experiencing the same issue on mobile and sometimes even on desktops. Turning off the calls to the endpoint again seems to have fixed the issue where the user cannot access the main content/form. I want to see if our caching strategy is the issue and I want to test this hypothesis in Staging.
- Using MemoryStore is meant for small low traffic systems and is usually reserved for development and testing environments: https://guides.rubyonrails.org/caching_with_rails.html#activesupport-cache-memorystore According to Prometheus metrics in prod, the `feature_toggles` endpoint gets hit like 25 million times in 24 hrs.
- A puma cluster in production cannot share the same MemoryStore cache and I'm thinking because of this that the cache is duplicated across our EC2 instances? I'm not really sure, will need to learn more about this. 
- I want to conduct load testing using both types of caching in Staging as we're anticipating a large increase in traffic once the CheckIn project goes live. The Check-In initiative seeks to replace all Kiosks in the VA hospitals by moving all of its functionality to va.gov. The current estimate of the number of Check-Ins that would ensue as a result of this move is around 100 million or more Check-ins per year.
- I want to **learn** if moving to a Redis cache will help us handle this type of traffic better.
- Openly provide all the facts and data to everyone on the different teams so that a good decision can be made based on the facts.

ericboehs:
There's a lot to unpack here. Let's start with benchmarks for the existing features toggles for this app on mobile. How long are they taking to run? How many are there?

config/initializers/flipper.rb
```diff
@@ -12,13 +12,10 @@
...
 Flipper.configure do |config|
   config.default do
-    activerecord_adapter = Flipper::Adapters::ActiveRecord.new
-    cache = ActiveSupport::Cache::MemoryStore.new
-    # Flipper settings will be stored in postgres and cached in memory for 1 minute in production/staging
-    cached_adapter = Flipper::Adapters::ActiveSupportCacheStore.new(activerecord_adapter, cache, expires_in: 1.minute)
-    adapter = Rails.env.development? || Rails.env.test? ? activerecord_adapter : cached_adapter
+    adapter = Flipper::Adapters::ActiveRecord.new
+    cache = Rails.cache
+    adapter = Flipper::Adapters::ActiveSupportCacheStore.new(adapter, cache, expires_in: 5.minutes)
```
bosawt, line 17: For comparison purposes, while gathering metrics, should we keep this `expires_in` to 1 minute to match the old caching method? (Might make sense to move it up to 5, or higher, eventually but just for an equal comparison)

config/initializers/flipper.rb
```diff
@@ -12,13 +12,10 @@
 
 Flipper.configure do |config|
   config.default do
-    activerecord_adapter = Flipper::Adapters::ActiveRecord.new
-    cache = ActiveSupport::Cache::MemoryStore.new
-    # Flipper settings will be stored in postgres and cached in memory for 1 minute in production/staging
-    cached_adapter = Flipper::Adapters::ActiveSupportCacheStore.new(activerecord_adapter, cache, expires_in: 1.minute)
-    adapter = Rails.env.development? || Rails.env.test? ? activerecord_adapter : cached_adapter
```
bosawt, line 19: In your writeup in the description (which was very useful, by the way), you mention that we use caching in development: `Currently, the vets-api is caching the feature toggle endpoint using ActiveSupport::Cache::MemoryStore(usually implemented in the development environment).`.
However, according to this line, it looks it's actually opposite, we appear to use caching in all environments that aren't `development` or `test`

app/controllers/v1/feature_toggles_controller.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module V1
+  class FeatureTogglesController < ApplicationController
+    skip_before_action :authenticate
+    before_action :load_user
+
+    def index
```
bosawt, line 8: If we allow functionality to query just a single feature (which IMO is the best thing we could do to make this more efficient in general), I think we should put that functionality in a `show` route in this controller, to split it from the full `index` functionality

app/services/feature_toggles/bundle.rb
```diff
@@ -0,0 +1,133 @@
+# frozen_string_literal: true
+
+module FeatureToggles
```
bosawt, line 3: I like putting all this in a module outside of the controller, seems cleaner overall

bosawt (commented):
This PR looks good for your current intention of testing performance to see if this is a viable solution. I'm curious to know the performance improvement of caching in general, for example, the timing of the endpoint when everything is cached, versus when nothing is cached. I suspect we might not find a huge difference, and if so, the only solution likely will be to call this less on the frontend

dillo:
Thank you all for your input. I'm closing this for now since I believe that my approach to the topic and solution was rather abrupt and should have included more talks with the right people before any code was written.

ericboehs:
Sometimes it's easier to think in code. I definitely learned stuff from the talks this code started.

Don't be afraid to open PRs for discussion. Even when closed they are valuable tools.

PR #7296, opened 2021-06-24
Api 8102/filter contestable issues by decision date

yakschuss:
[API-8102](https://vajira.max.gov/browse/API-8102)

This API adds filtering (specifically for NOD) of contestable issue results and removes `ratingIssueSubjectText: null` results.


modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/base_contestable_issues_controller.rb
```diff
@@ -62,6 +64,18 @@ def decision_review_type
     raise NotImplementedError, 'Subclass of BaseContestableIssuesController must implement decision_review_type method'
   end
 
+  def filtered_caseflow_response(decision_review_type, caseflow_response, filter)
+    return caseflow_response unless filter
+    return caseflow_response unless decision_review_type == 'appeals' # NOD requires this filtering step, HLR does not
+    return caseflow_response if caseflow_response.body['data'].nil?
+
+    caseflow_response.body['data'].reject! do |issue|
+      issue['attributes']['ratingIssueSubjectText'].nil?
```
cilestin, line 73: Should we do this with `present?` instead - ensure that non-nil but otherwise blank subject texts still get filtered out?

cilestin: Eh, something to consider for tech debt. Let's just get this through for UAT.

cilestin (approved):

PR #7297, opened 2021-06-24
[20767] Sidekiq Queues

cbtkc:
## Description of change
Separates queues into two different processes: `critical_job` & `default_job`

**critical_job**

Handles two queues; `critical` & `tasker`. The two queues are weighted 2 & 1 respectively. Weighting will ensure that tasker jobs will be attempted even if some critical jobs remain in queue, however given their weight of 1 they will be checked less often than critical jobs.

Moving faster performing jobs to critical will ensure the frequency with which the tasker queue is checked should improve. So if queues get backed up, we will not encounter scenarios where we are unaware of what is happening, because tasker jobs aren't being reported.

**default_job**

Handles two queues; `default` & `low`. These two queues are also weighted 2 & 1 respectively. Slower jobs will remain in default and low queues. Faster jobs, especially those where the user will be sitting and waiting for something to happen (i.e. login related jobs) have been moved to the critical queue.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#20767


Procfile
```diff
@@ -1,4 +1,5 @@
 web: bundle exec puma -p 3000 -C ./config/puma.rb
-job: bundle exec sidekiq -q default -q critical -q tasker
+critical_job: bundle exec sidekiq -q critical,2 -q tasker,1
```
ericboehs, line 2: After reading up about the polling interval, I think one job would be fine if we did something like this:
```suggestion
job: bundle exec sidekiq -q critical,4 -q tasker,3 -q default, 2 -q low,1
```

Thoughts?

Also, I wonder how the Procfile and our ASG is set up. Do we need to reconfigure it to look for the two jobs rather than just "job"?

cbtkc: Either would be an improvement on what we are currently doing! Introducing more processes may not be necessary, we can try out your suggested change and adjust as needed.

cbtkc:
@ericboehs your suggestion has been implemented.

ericboehs (approved):

PR #7298, opened 2021-06-24
Add rswag-generated swagger to claims, include in all environments

hinzed1127:
Third time's the charm, including in the the higher environments as well.

## Original issue(s)
https://vajira.max.gov/browse/API-7861

This was previously reverted because it did not include `rswag` in the right group in the `Gemfile`. This has been updated accordingly


Gemfile
```diff
@@ -124,6 +124,7 @@ gem 'redis-namespace'
 gem 'request_store'
 gem 'restforce'
 gem 'rgeo-geojson'
+gem 'rswag'
```
hinzed1127, line 127: this was the change missing previously causing things to break in higher environments

jw81 (approved):

PR #7307, opened 2021-06-28
Api 8152/deprecate v1 contestable issues

yakschuss:
[API-8152](https://vajira.max.gov/browse/API-8152)

This PR adds deprecation headers to V1 HLR Contestable Issues endpoint.

### DIFF
--- a/modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/higher_level_reviews/contestable_issues_controller.rb
+++ b/modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/higher_level_reviews/contestable_issues_controller.rb
@@ -4,7 +4,7 @@ class AppealsApi::V1::DecisionReviews::HigherLevelReviews::ContestableIssuesCont
   include AppealsApi::HeaderModification
 
   def index
-    deprecate(response: response, link: AppealsApi::HeaderModification::V2_DEV_DOCS)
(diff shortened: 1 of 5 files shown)


cilestin (dismissed):

cilestin (dismissed):

nathanbwright (dismissed):

cilestin (approved):

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/higher_level_reviews/contestable_issues_controller.rb
```diff
@@ -1,9 +1,27 @@
...
+  end
+
   private
 
   def decision_review_type
     'higher_level_reviews'
   end
+
+  def hlr_v2_live?
+    Settings.modules_appeals_api.documentation.path_enabled_flag
```
nathanbwright, line 25: A little confusing that we're using this `documentation` setting in the controller, but I think it makes sense for the time being...

nathanbwright (approved):

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/higher_level_reviews/contestable_issues_controller.rb
yakschuss: I agree. I don't really like it here, but considering the value is the same for what we're using in the controller, once we go live, we can remove it.

PR #7310, opened 2021-06-28
deprecate v1 HLR

yakschuss:
[API-8154](https://vajira.max.gov/browse/API-8154)

This PR adds deprecation headers to the HLR v1 endpoints, with a link to the release notes and a sunset date
(2022-01-31)

### DIFF
--- a/modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/higher_level_reviews_controller.rb
+++ b/modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/higher_level_reviews_controller.rb
@@ -6,6 +6,7 @@ require 'appeals_api/form_schemas'
 class AppealsApi::V1::DecisionReviews::HigherLevelReviewsController < AppealsApi::ApplicationController
   include AppealsApi::JsonFormatValidation
   include AppealsApi::StatusSimulation
+  include AppealsApi::HeaderModification
 
   skip_before_action(:authenticate)
   before_action :validate_json_format, if: -> { request.post? }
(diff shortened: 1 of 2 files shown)


modules/appeals_api/app/controllers/concerns/appeals_api/header_modification.rb
```diff
@@ -7,10 +7,12 @@ module HeaderModification
     included do
       V1_DEV_DOCS = 'https://developer.va.gov/explore/appeals/docs/decision_reviews?version=1.0.0'
       V2_DEV_DOCS = 'https://developer.va.gov/explore/appeals/docs/decision_reviews?version=2.0.0'
+      RELEASE_NOTES_LINK = 'https://developer.va.gov/release-notes/appeals'
```
cilestin, line 10: I think it'd be useful to link to the specific release that deprecated an API version, if possible. Since the reasoning for deprecation will change between releases, maybe we should have the release link be a method on the controller similar to `sunset_date`. What do you think?

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/higher_level_reviews_controller.rb
```diff
@@ -22,22 +23,27 @@ class AppealsApi::V1::DecisionReviews::HigherLevelReviewsController < AppealsApi
   )['definitions']['hlrCreateParameters']['properties'].keys
 
   def create
+    deprecate(response: response, link: AppealsApi::HeaderModification::RELEASE_NOTES_LINK, sunset: sunset_date)
```
cilestin, line 26: Per our discussion, we should only be adding deprecation headers when the `path_enabled_flag` setting is true.

cilestin (changes requested):

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/higher_level_reviews_controller.rb
yakschuss: Absolutely, have to push up what I have

modules/appeals_api/app/controllers/concerns/appeals_api/header_modification.rb
yakschuss: There are no anchor tags for specific release notes. The appeals release notes are on a single page.

cilestin: Browsers have supported using any DOM's `id` attribute as an allowable fragID in the url, and will behave the same as (and actually take precedence over) an anchor tag.

So we can link to a specific release by linking to e.g. `https://developer.va.gov/release-notes/appeals#january-31-2019`

yakschuss: We'll have to update this when we have official release notes. As of this moment the release notes page is relatively small, so it's a pretty negligible difference, but I think it's a good idea!

nathanbwright (approved):

cilestin (approved):

PR #7316, opened 2021-06-29
26500 Add statsd to feature caching

dillo:
## Description of change

- Adding StatsD to ActiveSupport caching on the Flipper adapter so that we can pull data on cache hit/miss statistics for the MemoryStore cache currently used by the FeatureToggles functionality in production environments.
## Original issue(s)
department-of-veterans-affairs/va.gov-team#26500


ericboehs (approved):

darkbushido:
This isn't going to work as a way to check the number of cache hits vs SQL queries.
This will work with the current implementation since it doesn't let you know if you have a cache hit or miss

PR #7317, opened 2021-06-29
automatically shut off endpoint if sunset date is defined

yakschuss:
[API-8155](https://vajira.max.gov/browse/API-8155)

Automatically deactivate endpoint based on defined sunset date


cilestin (approved):
Nice! A lot better than my thought that it'd be part of the Settings!

PR #7318, opened 2021-06-29
API-8153 - Update v2 docs

nathanbwright:
## Description of change
This PR makes our NOD v1 docs render on the v2 path. I am curious to hear your perspective of doing it this way instead of copy and pasting everything from NOD v1 to NOD v2...

## Original issue(s)
https://vajira.max.gov/browse/API-8153


cilestin (approved):
Nice! I definitely like this approach. Wish we could do the same for some of the markdown as well, but we can explore that separately.

nathanbwright:
> Nice! I definitely like this approach. Wish we could do the same for some of the markdown as well, but we can explore that separately.

Yeah, I thought about that but the the markdown files are actually different (due to us consolidating HLR statuses)...

PR #7321, opened 2021-06-29
API-8103 - Sort contestable issues by approxDecisionDate

nathanbwright:
## Description of change
Sort contestable issues by approxDecisionDate

## Original issue(s)
https://vajira.max.gov/browse/API-8103


modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/base_contestable_issues_controller.rb
```diff
@@ -73,6 +73,12 @@ def filtered_caseflow_response(decision_review_type, caseflow_response, filter)
       issue['attributes']['ratingIssueSubjectText'].nil?
     end
 
+    caseflow_response.body['data'].sort_by! do |issue|
```
cilestin, line 76: Let me know if this feels too code-golfy, but you should be able to use `sort!` with the combined comparison operator (`<=>`) to sort descending with one iteration, rather than `sort_by!` & `reverse!`

Simple example:
```
> dates = ['2021-01-10', '2021-01-07', '2021-01-08', '2021-01-03']
> puts dates.sort! do |a, b| 
  Date.strptime(b, '%Y-%m-%d') <=>Date.strptime(a, '%Y-%m-%d')
end
```
Results in:
```
2021-01-10
2021-01-08
2021-01-07
2021-01-03
```

nathanbwright: I like code golfing, so I love your comment! 👍 

The reason I chose to use `sort_by` instead of `sort` is because `sort_by` is more efficient for "expensive" computations (I'm not sure how expensive parsing dates is because I didn't run any actual benchmarks)... I just went and looked up the details (because I couldn't remember the details off the top of my head) and `sort_by` uses a `Schwartzian Transform` and this allows each key to be computed only once time. https://ruby-doc.org/core-2.6.6/Enumerable.html#method-i-sort_by

I think that at this scale (just sorting a "small" number of objects), it probably doesn't really matter which one we use. Do you think combining everything into one `sort!` is better for readability?

cilestin: Agree that it doesn't really matter which we use, so let's just keep what you wrote :)

cilestin (approved):

PR #7329, opened 2021-06-30
Kevin.cranmer/api 3314 expired doc error

KevinCranmer:
## Description of change

This PR is to add a more detailed description for errors relating to the /downloads endpoint.
Previously, if a request with an invalid guid was sent, we would return 500 error. Now it is a 404 error explaining the guid is missing.
Previously, if a request for a file that has been deleted from S3 was sent, we would return 500 error. Now it is a 404 error explaining the file has been removed from our storage.

## Original issue(s)
https://vajira.max.gov/browse/API-3314

A new rspec test was created to test the 404 scenario. An existing rspec test was modified such that the object doubles allowed the new method calls when checking if a file exists.

The endpoint was also hit with Postman requests testing:
Valid guid
Invalid guid
Valid guid with removed file


gpbowman-git (commented):
in uploads controller, change the raise to use the same parameter as the call above it for consistency.

submission = VBADocuments::UploadSubmission.find_by(guid: params[:upload_id])
raise Common::Exceptions::**RecordNotFound, params[:upload_id]** if submission.nil?

cshupp1 (approved):

PR #7334, opened 2021-07-01
Api 7661/refactor form validations and contestable issues

yakschuss:
[API-7661](https://vajira.max.gov/browse/API-7661)

This PR refactors the validations on HLR model that deal with form sizing and input box space.

We move it out of the data access layer (model) and moved into the controller with feature parity.

Additionally, refactored some access methods for informal_conference_rep which were unnecessary.

It also adds a model for contestable issue, as an abstraction to be used independently.


modules/appeals_api/app/controllers/concerns/appeals_api/higher_level_reviews/v1_validator.rb
```diff
@@ -0,0 +1,49 @@
...
+      def informal_conference_rep
+        request_body.dig('data', 'attributes', 'informalConferenceRep')
+      end
+
+      def informal_conference_rep_name
+        informal_conference_rep&.dig('name')
+      end
+
+      def informal_conference_rep_phone
+        AppealsApi::HigherLevelReview::Phone.new(informal_conference_rep&.dig('phone'))
```
nathanbwright, line 45: What about returning early instead of using the `&` operator?  It seems like it might read a little more clearly...

modules/appeals_api/app/services/appeals_api/pdf_construction/higher_level_review/v2/form_data.rb
```diff
@@ -99,27 +99,27 @@ def rep_last_name
...
 
-          higher_level_review.rep_phone_data&.dig('areaCode') || ''
+          higher_level_review.informal_conference_rep_phone.area_code || ''
         end
 
         def rep_phone_prefix
           return if rep_country_code != '1'
 
-          higher_level_review.rep_phone_data&.dig('phoneNumber')&.first(3) || ''
+          higher_level_review.informal_conference_rep_phone.phone_number&.first(3) || ''
```
nathanbwright, line 108: What do you think about pushing this phone number logic (for several of these related methods) down at least one layer?

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/higher_level_reviews_controller.rb
```diff
@@ -65,6 +66,18 @@ def validate_json_schema_for_body
     AppealsApi::FormSchemas.new.validate!(self.class::FORM_NUMBER, @json_body)
   end
 
+  def validate_json_schema_for_pdf_fit
+    status, error = AppealsApi::HigherLevelReviews::PdfFormatValidation.new(
```
nathanbwright, line 70: When I initially saw this class name `PdfFormatValidation` I thought of validations of the pdf itself (dimensions, file size, etc) and was somewhat confused by it... Maybe it's just me, but something like `PdfFormFieldValidation` (or something completely different) would be more clear about what is being checked...

nathanbwright (commented):
Overall, I really like getting some of these validations out of the model! Nice work! I left a few inline comments about some of the details.

modules/appeals_api/app/controllers/concerns/appeals_api/higher_level_reviews/v1_validator.rb
yakschuss: I prefer to instantiate the `Phone` class with a nil value then return nil altogether. That way, when we access the phone methods, (such as ext, country_code, etc) it returns nil instead of erroring.

modules/appeals_api/app/services/appeals_api/pdf_construction/higher_level_review/v2/form_data.rb
yakschuss: down a layer, into the `Phone` class, you mean?

nathanbwright: Yeah, that's what I was thinking. Instead of `&.first(3)` it could be `area_code`, `prefix`, etc... I suppose the next question is - who's responsibility is it to return the empty string? The phone class or the `FormData#rep_phone_prefix` method, I'm not sure what I think about that...

modules/appeals_api/app/controllers/concerns/appeals_api/higher_level_reviews/pdf_format_validation.rb
```diff
@@ -0,0 +1,49 @@
...
+  module HigherLevelReviews
+    class PdfFormatValidation
+      def initialize(request_body, request_headers, version_validator:)
+        @request_body = request_body
+        @request_headers = request_headers
+        @version_validator = version_validator.new(request_body, request_headers)
+      end
+
+      def validate!
+        return error(422, veteran_phone.too_long_error_message) if veteran_phone.too_long?
```
cilestin, line 15: I don't think that HLR v2 has the same character limits for phone number as v1 did, so I don't really think this shared utility class is necessary.

I think we could remove this class and just let each version validator get called directly from each controller. Even if there's a bit of duplication (in the case that the phone number validation _was_ the same across both versions), I think that separating out what's valid for each version is fine and somewhat more clear - I won't have to remember which layer handled which validation, I'd just look at the validator for that version. Thoughts?

nathanbwright (approved):
Looks good to me!

PR #7335, opened 2021-07-01
Api 7661/contestable issue class

yakschuss:
[API-7661](https://vajira.max.gov/browse/API-7661)

This PR is pulled out of [this PR](https://github.com/department-of-veterans-affairs/vets-api/pull/7334) to allow for the 500 line limit].

It pulls out contestable issues into an abstractions, updates references and refactors validations.

### DIFF
--- a/modules/appeals_api/app/models/appeals_api/higher_level_review.rb
+++ b/modules/appeals_api/app/models/appeals_api/higher_level_review.rb
@@ -195,7 +195,7 @@ module AppealsApi
     def contestable_issues
       issues = form_data.dig('included') || []
 
-      issues.map do |issue|
+      @contestable_issues ||= issues.map do |issue|
         AppealsApi::ContestableIssue.new(issue)
       end
     end
--- a/modules/appeals_api/app/models/appeals_api/higher_level_review.rb
+++ b/modules/appeals_api/app/models/appeals_api/higher_level_review.rb
@@ -193,7 +193,11 @@ module AppealsApi
 
     # 15. YOU MUST INDICATE BELOW EACH ISSUE...
     def contestable_issues
-      form_data&.dig('included')
+      issues = form_data.dig('included') || []
+
+      issues.map do |issue|
+        AppealsApi::ContestableIssue.new(issue)
+      end
     end
 
     # 16B. DATE SIGNED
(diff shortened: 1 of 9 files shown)


modules/appeals_api/app/models/appeals_api/higher_level_review.rb
```diff
@@ -193,7 +193,11 @@ def soc_opt_in
 
     # 15. YOU MUST INDICATE BELOW EACH ISSUE...
     def contestable_issues
-      form_data&.dig('included')
+      issues = form_data.dig('included') || []
+
+      issues.map do |issue|
```
cilestin, line 198: Let's memoize the built ContestableIssues array, so that we don't have to recreate the instances each call.

modules/appeals_api/app/models/concerns/appeals_api/contestable_issue.rb
```diff
@@ -0,0 +1,63 @@
...
+
+class AppealsApi::ContestableIssue
+  def initialize(issue)
+    @issue = issue
+  end
+
+  def decision_date
+    return unless decision_date_string
+
+    parse_date(decision_date_string)
```
cilestin, line 11: I feel like we should memoize the date parsing, too.

modules/appeals_api/app/models/concerns/appeals_api/contestable_issue.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def soc_date_string
+    issue.dig('attributes', 'socDate')
+  end
+
+  def text
+    issue.dig('attributes', 'issue')
+  end
+
+  delegate :[], to: :issue
```
cilestin, line 32: 👍

modules/appeals_api/app/models/concerns/appeals_api/contestable_issue.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def text_exists?
+    text.present?
+  end
+
+  def soc_date_past?
```
cilestin, line 38: While we're at it, let's combine the various soc/text/decision_date methods together, since this is the first commit that introduces them.

cilestin (changes requested):
Overall looks good! Just some memoization and tidying to go, I think :)

modules/appeals_api/app/models/concerns/appeals_api/contestable_issue.rb
yakschuss: I don't really understand how these are supposed to combine, or the reasoning you're presenting for it. Could you elaborate a bit more?

cilestin: Ah, sorry, I can see how "combine" could mean something more than I intended. I just meant move the methods around so that all the SOC methods are grouped together, etc.

yakschuss: Ah, got it.

cilestin (approved):

PR #7339, opened 2021-07-01
API-8258 - Reduce NOD controller duplication using concerns

nathanbwright:
## Description of change
In an effort to remove duplication between NOD v1 and v2 controllers we're looking at two different possible solutions. This is one way we could do it using concerns (literally just copied everything from the controllers and put it in the concern).

## Original issue(s)
https://vajira.max.gov/browse/API-8258


cilestin (approved):
Looks good!

nathanbwright:
Closing in favor of other approach.

PR #7343, opened 2021-07-01
[API-8289] - ClaimsApi prevent burial ITF submission from Veteran

jw81:
## Description of change

Adds logic to prevent case where Veteran would be submitting an ITF for their own burial.

## Original issue(s)
[API-8289](https://vajira.max.gov/browse/API-8289)

### DIFF
--- a/modules/claims_api/app/controllers/claims_api/v1/forms/intent_to_file_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/v1/forms/intent_to_file_controller.rb
@@ -22,6 +22,7 @@ module ClaimsApi
         # @return [JSON] Response from BGS
         def submit_form_0966
           validate_json_schema
+          check_for_invalid_burial_submission! if form_type == 'burial'
 
           bgs_response = bgs_service.intent_to_file.insert_intent_to_file(intent_to_file_options)
           render json: bgs_response,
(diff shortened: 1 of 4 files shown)


modules/claims_api/app/controllers/claims_api/v1/forms/intent_to_file_controller.rb
```diff
@@ -63,10 +63,10 @@ def validate
         private
 
         def participant_claimant_id
-          return target_veteran.participant_id unless form_type == 'burial'
-          return target_veteran.participant_id unless header_request?
```
SlakrHakr, line 67: this part is important though (at least i think)
burial itf can't, currently, be submitted by either a representative or the Veteran themself
cant be the Veteran for obvious reasons, but can't be the representative because our current implementation of itf uses the Veteran's participant id for the submitter as well.

SlakrHakr (changes requested):

modules/claims_api/app/controllers/claims_api/v1/forms/intent_to_file_controller.rb
jw81: After a lot of back-and-forth between us and Mark, I think we decided to just forbid the action right now if we were not given a `participant_claimant_id`.  [This work](https://github.com/department-of-veterans-affairs/vets-api/pull/7343/commits/a421fbe04d08af8f69e872c2e2b2b19b1d1900ea) provides that logic.

SlakrHakr (approved):

PR #7350, opened 2021-07-02
API-7796 Legacy appeals endpoint with caseflow service

ElleeB:
## Description of change
Given a Veteran's SSN or file_number, retrieve (active) Legacy Appeals from Caseflow (which retrieves from VACOLS).

## Ticket
https://vajira.max.gov/browse/API-7796


modules/appeals_api/app/controllers/appeals_api/v2/decision_reviews/legacy_appeals_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+
+  def caseflow_response_from_backend_service_exception
+    Struct.new(:status, :body).new(
+      backend_service_exception.original_status,
+      backend_service_exception.original_body
+    )
+  end
+
+  def render_unusable_response_error
+    render json: UNUSABLE_RESPONSE_ERROR, status: UNUSABLE_RESPONSE_ERROR[:errors].first[:status]
```
cilestin, line 92: No need to dig into the UNUSABLE_RESPONSE_ERROR for the status code - just return a 502.

modules/appeals_api/app/controllers/appeals_api/v2/decision_reviews/legacy_appeals_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+      validation_errors << { status: 422, detail: "X-VA-SSN has an invalid format. Pattern: #{SSN_REGEX.inspect}" }
+    end
+
+    render_validation_errors(validation_errors)
+  end
+
+  def render_validation_errors(validation_errors)
+    return if validation_errors.empty?
+
+    render json: { errors: validation_errors }, status: :unprocessable_entity unless validation_errors.empty?
```
cilestin, line 64: You've already guarded against empty validation errors above, no need to do it a second time with the `unless`.

modules/appeals_api/app/controllers/appeals_api/v2/decision_reviews/legacy_appeals_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+  SSN_REGEX = /^[0-9]{9}$/.freeze
+
+  skip_before_action(:authenticate)
+  before_action :validate_headers, only: %i[index]
+
+  EXPECTED_HEADERS = %w[X-VA-SSN X-VA-File-Number].freeze
+
+  UNUSABLE_RESPONSE_ERROR = {
+    errors: [
+      {
```
cilestin, line 16: We should already have a common exception for a bad gateway. Let's use that instead. Our consumers won't necessarily have any context for what caseflow is. If _we_ want to know there was an error from caseflow (and we do), we should log it but not bubble up the service that failed to the consumer.

modules/appeals_api/config/locales/en.yml
```diff
@@ -1,6 +1,7 @@
 en:
   appeals_api:
       errors:
+        caseflow_bad_gateway: "Received an unusable response from Caseflow."
```
cilestin, line 4: Where is this being used? I don't think this needs to be present at all if we end up using the common exception I _think_ this was trying to be used with anyway.

cilestin (changes requested):
Overall lookin' pretty good! Just had a few smaller change requests.

modules/appeals_api/app/controllers/appeals_api/v2/decision_reviews/legacy_appeals_controller.rb
ElleeB: derp

modules/appeals_api/app/controllers/appeals_api/v2/decision_reviews/legacy_appeals_controller.rb
ElleeB: OK I'll look into this. We use it in contestable issues I believe.

modules/appeals_api/config/locales/en.yml
ElleeB: Looking into it.

modules/appeals_api/app/controllers/appeals_api/v2/decision_reviews/legacy_appeals_controller.rb
ElleeB: I'm going to rewire this design. I reused it and realize now that I probably shouldn't have.

ElleeB: reworked for something a little more simple.

modules/appeals_api/app/controllers/appeals_api/v2/decision_reviews/legacy_appeals_controller.rb
ElleeB: simplified this.

cilestin (approved):

modules/appeals_api/app/controllers/appeals_api/v2/decision_reviews/legacy_appeals_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+      render_unusable_response_error
+    end
+  end
+
+  private
+
+  attr_reader :caseflow_response, :caseflow_exception_response
+
+  def request_headers
+    EXPECTED_HEADERS.reduce({}) do |hash, key|
```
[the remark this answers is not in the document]
ElleeB: oh, ya!

cilestin (approved):

PR #7352, opened 2021-07-02
API-8413: Prevent user matching multiple accredited representatives from using system

SlakrHakr:
## Description of change
We noticed an issue that can occur if a user is an accredited representative (not the Veteran themself) but has the same first and last name as another representative. A while back I was discovered there is at least one instance of duplicate first and last names within our current OGC data-set.

We've discussed alternative methods for matching a user to OGC data with the maintainers of OGC but unfortunately they dont store ssn or birthdate information. (and dont have any plan to in the near future)

This code change has the following effect:
In the small instances that an accredited representative both has the same first and last name, our system will respond with unauthorized to both users with the following message:
"Ambiguous VSO Representative Results"

## Original issue(s)
https://vajira.max.gov/browse/API-8413

### DIFF
--- a/lib/bgs/power_of_attorney_verifier.rb
+++ b/lib/bgs/power_of_attorney_verifier.rb
@@ -20,14 +20,13 @@ module BGS
     end
 
     def verify(user)
-      veteran_poa_code = current_poa_code
       reps = Veteran::Service::Representative.all_for_user(first_name: user.first_name,
-                                                           last_name: user.last_name,
-                                                           poa_code: veteran_poa_code)
+                                                           last_name: user.last_name)
       raise ::Common::Exceptions::Unauthorized, detail: 'VSO Representative Not Found' if reps.blank?
       raise ::Common::Exceptions::Unauthorized, detail: 'Ambiguous VSO Representative Results' if reps.count > 1
 
       rep = reps.first
+      veteran_poa_code = current_poa_code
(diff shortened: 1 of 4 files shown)


jw81 (approved):

lib/bgs/power_of_attorney_verifier.rb
```diff
@@ -20,15 +20,16 @@ def previous_poa_code
...
-        unless matches(veteran_poa_code, rep)
-          Rails.logger.info("POA code of #{rep.poa_codes.join(', ')} not valid for veteran code #{veteran_poa_code}")
-          raise Common::Exceptions::Unauthorized, detail: "Power of Attorney code doesn't match Veteran's"
-        end
-      else
-        raise Common::Exceptions::Unauthorized, detail: 'VSO Representative Not Found'
+      reps = Veteran::Service::Representative.all_for_user(first_name: user.first_name,
+                                                           last_name: user.last_name)
+      raise ::Common::Exceptions::Unauthorized, detail: 'VSO Representative Not Found' if reps.blank?
+      raise ::Common::Exceptions::Unauthorized, detail: 'Ambiguous VSO Representative Results' if reps.count > 1
```
ericboehs, line 26: How unlucky to receive this error. Haha.

ericboehs (approved):

PR #7353, opened 2021-07-02
API-8439 Increase NOD email character limit

cilestin:
## Description of change
* Changed minLength & maxLength on schema to align with VA Profile
  * Moved length limits to the nodCreateEmail ref so it shows properly on documentation
* Adds additional page w/ Preferred Email if the email will not fit into the field (anything > 120 characters)
  * Field has "See additional page..." text in the field when length is too long

## Original issue(s)
https://vajira.max.gov/browse/API-8439


modules/appeals_api/app/services/appeals_api/pdf_construction/notice_of_disagreement/pages/hearing_type_and_additional_issues.rb
```diff
@@ -43,12 +44,22 @@ def extra_issues?
             form_data.contestable_issues.count > MAX_ISSUES_ON_FIRST_PAGE
           end
 
+          def short_preferred_email?
+            form_data.preferred_email.length <= 120
```
ElleeB, line 48: Can someone remind me why we are referring to this as `preferred` email as opposed to `veteran_email` or `email`?

ElleeB: The name in the pdf file? Or we named the field? Just wondering.

modules/appeals_api/spec/services/appeals_api/pdf_construction/generator_spec.rb
```diff
@@ -41,6 +41,31 @@
...
+          notice_of_disagreement.form_data['data']['attributes']['veteran']['emailAddressText'] = new_email
+          notice_of_disagreement.save!
+        end
+
+        it 'places long email addresses onto additional page' do
+          Timecop.freeze(Time.zone.parse('2020-01-01T08:00:00Z'))
+          generated_pdf = described_class.new(notice_of_disagreement).generate
+          reader = PDF::Reader.new(generated_pdf)
+          # Character representation of pdf can be hard to test against, since its layout is converted to line breaks,
+          # whitespace, & other special characters in non-intuitive ways. We do our best to work with it, below.
```
nathanbwright, line 59: What do you think about including a max length email in our `expected_10182_extra.pdf` test? Then we wouldn't have to do this character by character testing like this?

modules/appeals_api/app/services/appeals_api/pdf_construction/notice_of_disagreement/pages/hearing_type_and_additional_issues.rb
cilestin: Yes, the "physical" 10182 pdf file uses the term `My Preferred E-Mail Address` for the field.

modules/appeals_api/spec/services/appeals_api/pdf_construction/generator_spec.rb
cilestin: I keep hoping that we can do tests that don't need a rendered fixture file to check against - mostly because it _feels_ like I'm actually testing a specific piece of functionality instead of a mashup of "all things that may produce an extra page."

That said, I agree with hindsight that the fixture file is probably the better way to make this less piecemeal to work around `pdf -> text` craziness. I'll get that updated tomorrow :)

nathanbwright (dismissed):
👍

nathanbwright (approved):

PR #7362, opened 2021-07-06
API-8438 - Update status job error threshold in lower envs

nathanbwright:
## Description of change
This error notification was firing off more than we wanted for the lower envs. Upping the threshold to try to quiet it down somewhat.

## Original issue(s)
https://vajira.max.gov/browse/API-8438

### DIFF
--- a/modules/appeals_api/app/workers/appeals_api/higher_level_review_upload_status_updater.rb
+++ b/modules/appeals_api/app/workers/appeals_api/higher_level_review_upload_status_updater.rb
@@ -23,7 +23,7 @@ module AppealsApi
       if Settings.vsp_environment == 'production'
         [2, 4]
       else
-        [3]
+        [5]
       end
     end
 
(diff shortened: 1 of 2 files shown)


yakschuss (approved):

modules/appeals_api/app/workers/appeals_api/higher_level_review_upload_status_updater.rb
```diff
@@ -23,7 +23,7 @@ def retry_limits_for_notification
       if Settings.vsp_environment == 'production'
         [2, 4]
       else
-        [3]
+        [5]
```
cilestin, line 26: IIRC the retry_count starts at 0 for the first retry, so w/ 5 retries this would be `4` and not `5`.

yakschuss: we add 1 to the middleware when it comes in, so this would be the 5th retry, which would have a `retry_count` of 4

cilestin (approved):

PR #7368, opened 2021-07-06
update sql for InProgressForms to allow either case

annaswims:
## Description of change
it seems that InProgressForm metadata can be in either case (snake or camel), depending on how old it is, update sql to reflect this. 

## Original issue(s)


ericboehs (approved):
It would be nice if we prevented submission of invalid data here (preferring underscore to camel) instead of fixing it on retrieval.

PR #7369, opened 2021-07-07
API-8410: Return current active power of attorney for a Veteran

SlakrHakr:
## Description of change
Return current active Power of Attorney for Veteran within v2.

## Original issue(s)
https://vajira.max.gov/browse/API-8410

image


modules/veteran/app/models/veteran/user.rb
```diff
@@ -44,7 +44,7 @@ def bgs_service
       external_key = "#{@user.first_name} #{@user.last_name}"
       @bgs_service ||= BGS::Services.new(
         external_uid: @user.mpi_icn,
-        external_key: external_key
+        external_key: external_key.presence || @user.mpi_icn
```
SlakrHakr, line 47: This was required due to not having a first name or last name when retrieving the Veteran from MPI using only their ICN.

SlakrHakr: It should be fine as BGS only uses these values for logging purposes, as I understand.

jw81 (approved):

PR #7371, opened 2021-07-07
update vets-json-schema

annaswims:
## Description of change
There was a misunderstanding about when EVSS would have this change in prod so this will need to be temporarily reverted.
https://github.com/department-of-veterans-affairs/vets-json-schema/pull/615


rileyanderson (dismissed):

annaswims:
spec failed because swagger wasn't liking the json comments. That's fixed now

rileyanderson (approved):

PR #7373, opened 2021-07-07
Increse HLR v2 veteran & rep email lengths w/ additional page when needed

cilestin:
## Description of change
Use the VA Profile email schema for HLR v2's Veteran and Representative emails. 
Move to an additional page if the emails are too long for the fields, with a note in the field to check the additional page.

## Original issue(s)
https://vajira.max.gov/browse/API-8444


nathanbwright (approved):

modules/appeals_api/app/services/appeals_api/pdf_construction/higher_level_review/pages/v2/additional_issues.rb
```diff
@@ -16,6 +16,14 @@ def build!
 
               pdf.start_new_page
 
+              unless short_veteran_email?
+                pdf.text "\n<b>Veteran Email:</b>\n#{form_data.veteran_email}\n", inline_format: true
```
yakschuss, line 20: Should this be on it's own page? Or should it be with the additional issues as you have it?

cilestin: I thought about that too, but I figured it'd be easier on the intake specialist if we just let additional info on attached pages page break wherever they needed to, rather than risking an email page (which is otherwise mostly blank) and issues page(s).

PR #7385, opened 2021-07-08
Added PractitionerIds field to Create Appointment request, will be up…

grossb1:
## Description of change

This update was to include the `PractitionerIds` field to the Create Appointment request.  There will be a story added to update the field to camel case once the associated changes are made on the VAMF side.

## Original issue(s)

Service specs were updated to reflect this change.


rileyanderson (approved):

saneshark:
LGTM, but if VAMF is changing the attribute soon, maybe not worth it?

PR #7386, opened 2021-07-08
Kevin.cranmer/api 8294 s3 file remover bug

KevinCranmer:
## Description of change

A job is repeatedly ran to clear out old files from our S3 storage. If someone were to manually remove a file from S3 and they forgot to check the s3_deleted flag to 'true', then the loop to remove old files would 'break' instead of moving to the 'next' file. This PR allows the upload remover job to move past these manually removed files and still remove other old files.

## Original issue(s)
[API 8294](https://vajira.max.gov/browse/API-8294)

rspec test has been included inside ./modules/vba_documents/spec/jobs/upload_remover_spec.rb

The new rspec test fails before the change. After the change, the test passes.


gpbowman-git (approved):
this looks really good.

cshupp1 (approved):

PR #7389, opened 2021-07-08
[6165] Changing Flipper adapter to in-memory version

bosawt:
## Description of change
This PR transitions the Flipper adapter to use the in-memory adapter. This shouldn't be a problem, because although the features yaml has grown to over a hundred features, we are still only saving an extra ~400 lines in memory. Given the immense size of the application itself, I don't see how this could be a problem whatsoever

## Original issue(s)
department-of-veterans-affairs/vets-api/issues/6165

- This PR represents the backend `feature_toggles` route being optimized to about as fast as it will ever be
- I also changed the `feature_toggles` route with the `features` param to more closely  match the non-parameter version of the route. This means if you call `feature_toggles` with one or more specific `features`, it'll return an attribute array with two elements: the snake-case version of the feature, and the camelized version of the feature.

### DIFF
--- a/app/controllers/v0/feature_toggles_controller.rb
+++ b/app/controllers/v0/feature_toggles_controller.rb
@@ -7,29 +7,28 @@ module V0
     before_action :load_user
 
     def index
+      features_array = []
+
       if params[:features].present?
         features_params = params[:features].split(',')
 
-        features = features_params.collect do |feature_name|
+        features_params.each do |feature_name|
           underscored_feature_name = feature_name.underscore
           actor_type = FLIPPER_FEATURE_CONFIG['features'].dig(feature_name, 'actor_type')
-
-          { name: feature_name, value: Flipper.enabled?(underscored_feature_name, actor(actor_type)) }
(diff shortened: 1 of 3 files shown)


annaswims:
By removing the authoritative source for the value for the feature toggle values (postgresql) you'll remove the ability to share toggle values across instances.  This PR would instantly disable all features and lead to a major outage.  

I'm happy to chat more about this tomorrow, but as it stands there is no performance problem on the back end and I think the only necessary improvement is for the FE to call this endpoint less often.

bosawt:
> By removing the authoritative source for the value for the feature toggle values (postgresql) you'll remove the ability to share toggle values across instances. This PR would instantly disable all features and lead to a major outage.
> 
> I'm happy to chat more about this tomorrow, but as it stands there is no performance problem on the back end and I think the only necessary improvement is for the FE to call this endpoint less often.

Oof, yeah you're totally right. There is a file store adapter as well, but I can't imagine that being any faster than Postgres.

Since I looked into calling this by just a single feature instead of index of all features as well, and didn't see much improvement, I don't even think there's much there. So basically I agree with you that BE really doesn't have anything further to do here.

PR #7393, opened 2021-07-09
27114 ch31 report find_each bug

RudyOnRails:
## Description of change

Switches `spec/mailers/ch31_submissions_report_mailer_spec.rb` so that the `submitted_claims` helper used is not a duplicate (and outdated) implementation of the logic inside `VRE::CreateCh31SubmissionsReport#get_claims_submitted_in_range`, which is the method responsible for generating the object passed to `Ch31SubmissionsReportMailer`. 

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/27114

The spec that should have failed during https://github.com/department-of-veterans-affairs/vets-api/pull/7320's CI was altered to now rely on the preparatory class and pass the same type of object used in production. Once I got that to fail, I just fixed the bug.

### DIFF
--- a/app/mailers/views/ch31_submissions_report.html.erb
+++ b/app/mailers/views/ch31_submissions_report.html.erb
@@ -7,7 +7,7 @@
     <th style="white-space:nowrap; text-align:left; width:15%">Type of Form</th>
     <th style="white-space:nowrap; text-align:left; width:5%">Total</th>
   </tr>
-  <% @submitted_claims.find_each.with_index do |claim, index|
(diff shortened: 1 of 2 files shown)


RudyOnRails:
I didn't mention that this is a production bug, and the (mid)nightly email report (for Ch31 claims) is not running because of it. Any assistance in getting it approved today would be appreciated by not just me, but all the recipients who depend on the report.

Thanks for reading,
Kevin

rileyanderson (approved):

RudyOnRails:
@rileyanderson thank you!

PR #7395, opened 2021-07-09
[API-8394] - ClaimsApi - GET V2 Veterans Claims

jw81:
## Description of change

Fleshes out the `ClaimsApi V2 GET Veterans Claims` endpoint to use BGS as the source for retrieving claims.

### Note
We'll want to wait to use this until [this PR](https://github.com/department-of-veterans-affairs/vets-api/pull/7424) with a small schema update is merged and migrations ran.

## Original issue(s)
[API-8394](https://vajira.max.gov/browse/API-8394)


modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
```diff
@@ -1,25 +1,47 @@
...
 module ClaimsApi
   module V2
     module Veterans
       class ClaimsController < ClaimsApi::V2::ApplicationController
-        # TODO: REMOVE BEFORE IMPLEMENTATION
-        skip_before_action :authenticate, only: %i[index]
-        ICN_FOR_TEST_USER = '1012667145V762142'
-        # TODO: REMOVE BEFORE IMPLEMENTATION
-
         def index
```
SlakrHakr, line 7: i might be missing it but it looks like this doesnt validate the logged in user is either the Veteran or their POA?

modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
```diff
@@ -1,25 +1,47 @@
...
-          raise ::Common::Exceptions::Unauthorized if request.headers['Authorization'].blank?
-          unless params[:veteranId] == ICN_FOR_TEST_USER
-            raise ::Common::Exceptions::ResourceNotFound.new(detail: 'Resource not found')
-          end
+          veteran = build_veteran(veteran_id: params[:veteranId], loa: @current_user.loa)
+          detail  = 'Veteran not found'
+          raise ::Common::Exceptions::ResourceNotFound.new(detail: detail) unless veteran_found?(veteran: veteran)
```
SlakrHakr, line 10: i went the route of never giving back a 404 if the Veteran is present or not. If the Veteran doesnt exist in MPI its still a 403:
https://github.com/department-of-veterans-affairs/vets-api/pull/7369/files#diff-9e276f2ddc002e7e8d30faaf090c0613cc94fe20f05e9353804d1210966fd9a2R10

I did this so that an person who is unauthorized couldnt determine whether a Veteran "existed" or not.

I think the "only 403" approach makes sense, but the bigger thing is just being consistent, so definitely up for discussing this if you disagree.

modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
```diff
@@ -1,25 +1,47 @@
...
+
+          service   = bgs_service(veteran_participant_id: veteran.participant_id)
+          params    = { participant_id: veteran.participant_id }
+          response  = service.benefit_claims.find_claims_details_by_participant_id(params)
+          claims    = transform_response(response: response)
```
SlakrHakr, line 15: it looks like this just responds with established claims from upstream? i know the ticket AC was a little back and forth but we ended with wanting to return both established upstream claims and our pending claims here

SlakrHakr (changes requested):

modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
jw81: Wow, yeah you're right.  I totally missed that.
I've updated the PR to follow the same pattern you introduced in [your PR](https://github.com/department-of-veterans-affairs/vets-api/pull/7369#pullrequestreview-704406712).

modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
jw81: Eh, personally it feels a little odd to return a 403 when something doesn't exist, but I understand the intention of obscuring that result.  But I agree, consistency is most important, so I updated the PR to follow the same pattern you introduced in [your PR](https://github.com/department-of-veterans-affairs/vets-api/pull/7369#pullrequestreview-704406712).

modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
jw81: Ok.  It'll take me some time to work through this one.
I'll update this comment when I've implemented that additional logic.

modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
```diff
@@ -1,25 +1,29 @@
...
-        skip_before_action :authenticate, only: %i[index]
-        ICN_FOR_TEST_USER = '1012667145V762142'
-        # TODO: REMOVE BEFORE IMPLEMENTATION
-
         def index
-          raise ::Common::Exceptions::Unauthorized if request.headers['Authorization'].blank?
-          unless params[:veteranId] == ICN_FOR_TEST_USER
-            raise ::Common::Exceptions::ResourceNotFound.new(detail: 'Resource not found')
-          end
+          raise ::Common::Exceptions::Forbidden unless user_is_target_veteran? || user_is_representative?
```
SlakrHakr, line 8: This endpoint is slightly different from the active POA i was working on. This endpoint should require that the person authenticated is either the Veteran themself or the Veteran's assigned POA. `user_is_representative?` just checks if the authenticated user is any of the accredited representatives in OGC.

modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
```diff
@@ -1,25 +1,29 @@
...
+
+          service           = bgs_service(veteran_participant_id: target_veteran.participant_id)
+          service_params    = { participant_id: target_veteran.participant_id }
+          bgs_claims        = service.benefit_claims.find_claims_details_by_participant_id(service_params)
+
+          query_params      = { veteran_icn: target_veteran.mpi.icn, status: %w[pending errored] }
```
SlakrHakr, line 14: why just pending or errored here? we should get established claims as well so we can respond with our guid instead of the upstream id

modules/claims_api/app/services/claims_api/bgs_to_lighthouse_claims_mapper_service.rb
```diff
@@ -0,0 +1,69 @@
...
+    def claims_exist?
+      return true if bgs_claims.key?(:bnft_claim_detail) || lighthouse_claims.present?
+
+      false
+    end
+
+    def map_claims
+      bgs_claims[:bnft_claim_detail].map do |bgs_claim|
+        if find_bgs_claim_in_lighthouse_collection(claim: bgs_claim)
+          matching_claim = find_bgs_claim_in_lighthouse_collection(claim: bgs_claim)
```
SlakrHakr, line 31: nit: if this is moved above the `if` check the logic in this method wouldn't have to be executed twice

modules/claims_api/app/services/claims_api/bgs_to_lighthouse_claims_mapper_service.rb
```diff
@@ -0,0 +1,69 @@
...
+      # this claim was submitted via Lighthouse, so use the 'id' the user is most likely to know
+      { id: matching_claim.id, type: bgs_claim[:bnft_claim_type_nm] }
+    end
+
+    def build_unmatched_bgs_claim(bgs_claim:)
+      { id: bgs_claim[:bnft_claim_id], type: bgs_claim[:bnft_claim_type_nm] }
+    end
+
+    def build_unmatched_lighthouse_claim(lighthouse_claim:)
+      { id: lighthouse_claim.id, type: lighthouse_claim.claim_type }
```
SlakrHakr, line 58: i didnt think about this scenario... what's an example of a `bgs_claim[:bnft_claim_type_nm]` value? just want to make sure the value is the same structure (capitalized/lower case/camel case, etc.)

modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
jw81: Hmmmm, maybe I'm misinterpreting the acceptance criteria?
> "Should return both our pending/errored claims and established claims"

I read that as "return our (Lighthouse) claims that have a status of 'pending'/'errored' and return claims from BGS".

modules/claims_api/app/services/claims_api/bgs_to_lighthouse_claims_mapper_service.rb
SlakrHakr: ah, im guessing these are example values from BGS?
https://github.com/department-of-veterans-affairs/vets-api/pull/7395/files#diff-08f5bc438fee25554ed287b3f4ccd81a18e6b5f6eb4abdef35548b446a698b3fR429

if so, it looks like our claim types have the same structure at least

jw81: `bnft_claim_type_nm: 'Appeals Control'` is an example value.

modules/claims_api/spec/requests/v2/veterans/rswag_claims_request_spec.rb
```diff
@@ -71,15 +96,20 @@
         end
       end
 
-      describe 'Getting a 404 response' do
-        let(:veteranId) { 'not-the-same-id-as-tamara' } # rubocop:disable RSpec/VariableName
-
-        response '404', 'Resource Not Found' do
+      describe 'Getting a 403 response' do
+        response '403', 'Resource Not Found' do
```
SlakrHakr, line 100: should say "Forbidden" instead of "Resource Not Found"

modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
SlakrHakr: ah, gotcha. the blurb for the value in the id field is:
> id:           GUID from our system if we know about the claim, evss_id from EVSS if we do not have a record for that claim
which is for established claims as well though, any claim we know about, whatever state its in

jw81: [tweaked here](https://github.com/department-of-veterans-affairs/vets-api/pull/7395/commits/a3468432550dba6ce15206eb20b5251ed13d7112)

modules/claims_api/app/services/claims_api/bgs_to_lighthouse_claims_mapper_service.rb
jw81: Fair enough.
[Changed here](https://github.com/department-of-veterans-affairs/vets-api/pull/7395/commits/7bd5e8be206124ab757b40ae285b01e945f331a0)

modules/claims_api/spec/requests/v2/veterans/rswag_claims_request_spec.rb
jw81: [D'oh!](https://github.com/department-of-veterans-affairs/vets-api/pull/7395/commits/f3dbe6bcd7a8a8e6125ec16a33803b74c508e295)
🤦

modules/claims_api/app/services/claims_api/bgs_to_lighthouse_claims_mapper_service.rb
```diff
@@ -0,0 +1,74 @@
...
+      mapped_claims
+    end
+
+    private
+
+    def claims_exist?
+      return true if bgs_claims.key?(:bnft_claim_detail) || lighthouse_claims.present?
+
+      false
+    end
```
hinzed1127, line 26: I'm not super hip to ruby conventions, but this could also be

```suggestion
    def claims_exist?
      bgs_claims.key?(:bnft_claim_detail) || lighthouse_claims.present?
    end
```

modules/claims_api/app/services/claims_api/bgs_to_lighthouse_claims_mapper_service.rb
```diff
@@ -0,0 +1,74 @@
...
+
+    def map_claims
+      bgs_claims[:bnft_claim_detail].map do |bgs_claim|
+        matching_claim = find_bgs_claim_in_lighthouse_collection(claim: bgs_claim)
+        if matching_claim
+          remove_bgs_claim_from_lighthouse_collection(claim: matching_claim)
```
hinzed1127, line 32: Just for my own understanding, for a given veteran:

1. we're grabbing all their claims submitted through Lighthouse
2. we're grabbing all their claims that exist in BGS
3. we're removing any Lighthouse claims from the BGS results
4. we're merging the 2 groups, resulting in all unique claims found between Lighthouse and BGS

does that sound right?

jw81: That's pretty close.

1. grab all of their claims that BGS knows about
2.  grab all of their claims that Lighthouse knows about
3. iterate the BGS collection and see if there is a claim in the Lighthouse collection with a matching external id
4. if there is a match
-- remove the matching claim from the Lighthouse collection
-- use the guid of the matching Lighthouse claim as the value for the returned `id` attribute
5. if there is not a match
-- use the `bnft_claim_id` of the BGS claim as the value for the returned `id` attribute
6. after iterating the BGS collection, if there are any claims remaining in the Lighthouse collection AND they don't have a status of `established`, then also include them in the collection that is being returned.

You might look at the specs I added to see if that helps.  I tried to keep the specs really small to make it easier to follow.

modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
jw81: Added some logic [here](https://github.com/department-of-veterans-affairs/vets-api/pull/7395/commits/b9bc790f5f75711fe00eebe43385e0c4c0d19522) to try and provide a common place to handle verifying access in V2 endpoints.

modules/claims_api/app/services/claims_api/bgs_to_lighthouse_claims_mapper_service.rb
jw81: I always forget cool stuff like that!  😂 
[Updated here.](https://github.com/department-of-veterans-affairs/vets-api/pull/7395/commits/906e5fea96187810333b059fce150e92813d4a31)

modules/claims_api/app/controllers/claims_api/v2/application_controller.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'bgs/power_of_attorney_verifier'
```
SlakrHakr, line 3: this isnt needed is it?

SlakrHakr (approved):

modules/claims_api/app/controllers/claims_api/v2/application_controller.rb
jw81: Nope.  Just a missed leftover from a previous idea.
[Removed here.](https://github.com/department-of-veterans-affairs/vets-api/pull/7395/commits/4e23c0aba33dfe9d6e36a88188dc757dad04f723)

hinzed1127 (approved):

PR #7399, opened 2021-07-09
Check for limitations of consent when displaying claims

hinzed1127:
## Description of change

When a representative looks up a claim, this adds a check to makes sure the veteran represented has not selected any limitations of consent. 

### Outstanding Questions
- Does this need be done for every single claim or only claims that contain information that would be blocked by limitations of consent? **UPDATE**:  Every single claim needs to be blocked. BGS/VBMS gives us no way to differentiate which records might have information being limited, so it's unfortunately an all or nothing thing at this point in time.
- There are a few edge cases not covered in the proposed method. Are the edge cases described [here](https://github.com/department-of-veterans-affairs/vets-api/pull/7399/files#diff-19a8fecd4f8b759336ac1b107ab77aaa9887859dc6a9ae6c35d049e24028291cR101-R108) a result of bad test data between BGS and `vets-api` or an actual discrepancy in production environments?
## Original issue(s)
https://vajira.max.gov/browse/API-6774

- Depends on changes in https://github.com/department-of-veterans-affairs/bgs-ext/pull/63

### DIFF
--- a/modules/claims_api/app/controllers/claims_api/v0/forms/base.rb
+++ b/modules/claims_api/app/controllers/claims_api/v0/forms/base.rb
@@ -72,13 +72,6 @@ module ClaimsApi
                                                  tempfile: temp_file)
         end
 
-        def bgs_service
-          BGS::Services.new(
-            external_uid: target_veteran.participant_id,
-            external_key: target_veteran.participant_id
-          )
-        end
-
         def received_date
           form_attributes['received_date']
         end
--- a/modules/claims_api/app/controllers/concerns/claims_api/poa_verification.rb
+++ b/modules/claims_api/app/controllers/concerns/claims_api/poa_verification.rb
@@ -91,27 +91,24 @@ module ClaimsApi
       def verify_consent_limitations!
         poa_code = BGS::PowerOfAttorneyVerifier.new(target_veteran).current_poa_code
 
-        form_code = if poa_code_in_organization?(poa_code)
-                      '21-22'
-                    else
-                      '21-22A'
-                    end
-        representatives = bgs_service.veteran_representative.read_all_veteran_representatives(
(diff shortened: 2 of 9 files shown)


modules/claims_api/app/controllers/concerns/claims_api/poa_verification.rb
```diff
@@ -86,6 +86,33 @@ def poa_code_in_organization?(poa_code)
...
+        # - what if it returns content but none of it matches with the current POA code?
+        #     - e.g. Ralph E Lee and Greg Anderson have this situation
+        # - what if the form_code is a false positive?
+        #     - i.e., poa_code_in_organization? says it should be 21-22 but it's actually 21-22A
+        #     - Idk if this one actually happens or not
+        rep_poa_codes = representatives.pluck(:poa_code)
+
+        unless rep_poa_codes.include?(poa_code)
+          raise ::Common::Exceptions::Forbidden,
+                detail: 'Veteran has not granted access to records protected by Section 7332, Title 38, U.S.C.'
```
hinzed1127, line 113: - [ ] Update this to 404 (or something similar) with message (paraphrased) `Talk to the information office at your closest regional office`.

Basically, we want to:

1. Just show no claims without indicating that it's due to limited consent
2. Give actionable next steps (if there's a way they could theoretically access those claims)

modules/claims_api/app/controllers/claims_api/v0/forms/base.rb
```diff
@@ -72,13 +72,6 @@ def decode_document(document)
                                                  tempfile: temp_file)
         end
 
-        def bgs_service
```
hinzed1127, line 75: moving this up a level to the application controller so it's also accessible from the claims controller

PR #7401, opened 2021-07-12
remove unused preneeds/states

michelpmcdonald:
## Description of change
the endpoint GET /v0/preneeds/states is unused and can be removed

## Original issue(s)
department-of-veterans-affairs/va.gov-team#27122


annaswims:
I think spec/support/schemas/preneeds/states.json and spec/support/schemas_camelized/preneeds/states.json can probably be removed as well

rileyanderson (approved):

PR #7402, opened 2021-07-12
25945 refactoring to remove duplication

dillo:
## Description of change

- Perform a refactor that will extract the duplicate code into its own module/class within the rake file.
## Original issue(s)
department-of-veterans-affairs/va.gov-team#25945


mdewey (approved):
LGTM

PR #7408, opened 2021-07-12
Api 6021/add received notification for hlr

yakschuss:
[API-6021](https://vajira.max.gov/browse/API-6021)

This PR triggers a notification for VANotify when an HLR is received (in version 2).

### DIFF
--- a/modules/appeals_api/app/services/appeals_api/events/appeal_received.rb
+++ b/modules/appeals_api/app/services/appeals_api/events/appeal_received.rb
@@ -9,7 +9,7 @@ module AppealsApi
       end
 
       def hlr_received
-        log_error unless email
+        log_error(guid, 'HLR') unless email
 
         template_type = 'higher_level_review_received'
         template_id = template_id(template_type)
@@ -37,8 +37,12 @@ module AppealsApi
         Settings.vanotify.services.va_gov.template_id.public_send(template)
       end
 
-      def log_error
-        Rails.logger.error 'No email present for AppealsApi::AppealReceived notification'
+      def log_error(guid, type)
+        Rails.logger.error "No email present for AppealsApi::AppealReceived notification #{type} - GUID: #{guid}"
+      end
+
+      def guid
+        opts['guid']
       end
 
       def email
@@ -50,7 +54,7 @@ module AppealsApi
       end
 
       def required_keys
-        %w[email date_submitted veteran_first_name veteran_last_name]
+        %w[guid email date_submitted veteran_first_name veteran_last_name]
       end
     end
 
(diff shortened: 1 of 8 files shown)


cilestin (dismissed):

modules/appeals_api/app/services/appeals_api/events/appeal_received.rb
```diff
@@ -0,0 +1,59 @@
...
+      def vanotify_service
+        @vanotify_service ||= VaNotify::Service.new(Settings.vanotify.services.lighthouse.api_key)
+      end
+
+      def template_id(template)
+        Settings.vanotify.services.va_gov.template_id.public_send(template)
+      end
+
+      def log_error
+        Rails.logger.error 'No email present for AppealsApi::AppealReceived notification'
```
cilestin, line 41: What do you think about passing in the HLR's guid into the `opts` so that we can know which record got this far? It's almost certain not to happen with our other checks, but if I saw this error in the logs I wouldn't really know where to start looking.

cilestin: Maybe also add that this is an error from the HLR-based event handler, so we know to look for HLRs (assuming we'll expand this to NODs or SC's later)

yakschuss: I can finagle it, but technically speaking, this file is not specific to HLR, it's for any appeal. The callback is specific to hlr, being `hlr_received`.  I'll figure it out! I like the idea.

modules/appeals_api/app/services/appeals_api/events/appeal_received.rb
```diff
@@ -0,0 +1,65 @@
...
+module AppealsApi
+  module Events
+    class AppealReceived
+      def initialize(opts)
+        @opts = opts
+        raise InvalidKeys unless required_keys?
+      end
+
+      def hlr_received
+        return unless Flipper.enabled?(:decision_review_hlr_email)
```
cilestin, line 12: Did we already add this Flipper setting in another PR? Did we make sure it was dependent on the global "send emails" flag as well (assuming it exists)?

cilestin (approved):

modules/appeals_api/app/services/appeals_api/events/appeal_received.rb
```diff
@@ -0,0 +1,65 @@
...
+
+        log_error(guid, 'HLR') unless email
+
+        template_type = 'higher_level_review_received'
```
yakschuss, line 16: @cilestin this is the tentative name of the template, but we can just adjust it later

PR #7409, opened 2021-07-12
remove unused preneeds/discharge_type code

michelpmcdonald:
## Description of change
the endpoint GET /v0/preneeds/discharge_types is unused and can be removed
## Original issue(s)
department-of-veterans-affairs/va.gov-team#27122


thilton-oddball (commented):
I found some references to get DischargeTypes in `config/preneeds/wsdl/preneeds.wsdl`
Do those need to come out?

michelpmcdonald:
> I found some references to getDischargeTypes in `config/preneeds/wsdl/preneeds.wsdl`
> Do those need to come out?

Good catch, @annaswims and I already kicked that around and ultimately decided to leave it in based on the idea that the wsdl describes the upstream service and getDischargeTypes is part of the service, it's just we don't use that part of the service and wanted to get rid of unused code, so we got rid of the code but left getDischargeTypes in the wsdl service description.......

thilton-oddball (approved):

PR #7410, opened 2021-07-12
Adding Pension Eligibility Check to Dependents Application

kathleencrawford:
## Description of change

When submitting a Dependents Application (VA Form 21-686c or VA Form 21-674), there are certain use cases that are not able to be processed automatically so we set those statuses to `MANUAL_VAGOV` for manual processing. We have been asked by our partners to add a check to determine whether or not the Veteran is currently receiving pension benefits, and change the claim type code depending on the response.

This PR does the following: if the status is `MANUAL_VAGOV`, call the awards API, then set the claim type based on whether or not the Veteran is currently receiving pension benefits.  Otherwise, use the default claim type code for claims that are not being set to manual.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#26831

This is an update to existing functionality, no new logging or settings have been added.

- [ ] Local testing
- [ ] Specs updated
- [ ] Staging testing planned

### DIFF
--- a/lib/bgs/form686c.rb
+++ b/lib/bgs/form686c.rb
@@ -97,15 +97,18 @@ module BGS
       'Started'
     end
 
+    # the default claim type code is 130DPNEBNADJ (eBenefits Dependency Adjustment)
     # if we are setting the claim to be manually reviewed
     # and the Veteran is currently receiving pension benefits
     # set the claim type code to 130DAEBNPMCR (PMC eBenefits Dependency Adjustment Reject)
-    # else use 130DPNEBNADJ (eBenefits Dependency Adjustment)
+    # else use 130DPEBNAJRE (eBenefits Dependency Adjustment Reject)
     def get_claim_type(proc_state)
       if proc_state == 'MANUAL_VAGOV'
         pension_response = bid_service.get_awards_pension
         receiving_pension = pension_response.body['awards_pension']['is_in_receipt_of_pension']
-        return '130DAEBNPMCR' if receiving_pension
+
+        claim_type = receiving_pension ? '130DAEBNPMCR' : '130DPEBNAJRE'
+        return claim_type
       end
 
       '130DPNEBNADJ'
--- a/lib/bgs/form686c.rb
+++ b/lib/bgs/form686c.rb
@@ -27,8 +27,9 @@ module BGS
 
       process_relationships(proc_id, veteran, payload)
 
+      claim_type_code = get_claim_type(vnp_proc_state_type_cd)
       vnp_benefit_claim = VnpBenefitClaim.new(proc_id: proc_id, veteran: veteran, user: @user)
-      vnp_benefit_claim_record = vnp_benefit_claim.create
+      vnp_benefit_claim_record = vnp_benefit_claim.create(claim_type: claim_type_code)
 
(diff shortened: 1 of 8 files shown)


lib/bgs/form686c.rb
```diff
@@ -96,8 +97,26 @@ def get_state_type(payload)
       'Started'
     end
 
+    # if we are setting the claim to be manually reviewed
+    # and the Veteran is currently receiving pension benefits
+    # set the claim type code to 130DAEBNPMCR (PMC eBenefits Dependency Adjustment Reject)
+    # else use 130DPNEBNADJ (eBenefits Dependency Adjustment)
+    def get_claim_type(proc_state)
+      if proc_state == 'MANUAL_VAGOV'
```
RudyOnRails, line 105: Although this may be a widely used pattern in the repo (and outside the scope of this PR), should this string be a constant somewhere? Something like:
```suggestion
      if proc_state == BGS::ProcStates::MANUAL_VAGOV
```

lib/bgs/form686c.rb
```diff
@@ -96,8 +97,26 @@ def get_state_type(payload)
...
+        pension_response = bid_service.get_awards_pension
+        receiving_pension = pension_response.body['awards_pension']['is_in_receipt_of_pension']
+        return '130DAEBNPMCR' if receiving_pension
+      end
+
+      '130DPNEBNADJ'
```
RudyOnRails, line 111: Since we are doing an explicit return on L108, should we do one here also?

lib/bgs/form686c.rb
RudyOnRails, line 105: Although this may be a widely used pattern in the repo (and outside the scope of this PR), should this string be a constant somewhere? Something like:
```suggestion
      if proc_state == BGS::ProcStates::MANUAL_VAGOV
```

kathleencrawford: > Although this may be a widely used pattern in the repo (and outside the scope of this PR), should this string be a constant somewhere? Something like:

There are only 3 proc states ('Started', 'Ready', 'MANUAL_VAGOV') so not sure we need this but where would you suggest we add it?

lib/bgs/form686c.rb
kathleencrawford: > Since we are doing an explicit return on L108, should we do one here also?

Rubocop does not like us to use `return` if its not needed.

RudyOnRails: 👌

lib/bgs/form686c.rb
RudyOnRails: So I typed out a reply, but then hit resolve conversation too quickly, thinking it would comment and resolve, but it wiped my comment. 

Basically I was saying I got a little carried away. Someone I respected at my old job put a widely used string in an Enum, so i thought I'd try to apply here.

I want to try and not repeat history where I "fix" things that aren't broken. We can always refactor when we have the time/need. Sorry for the noise here :)

rileyanderson (approved):

kathleencrawford:
Deleting this PR and replacing it with the changes here: https://github.com/department-of-veterans-affairs/vets-api/pull/7437

But wanted to leave this code (rather than replacing it) because I think eventually they will also want these changes.

PR #7411, opened 2021-07-12
Key rotation

LindseySaari:
## Description of change
**The deploy needs to be coordinated with [this PR](https://github.com/department-of-veterans-affairs/devops/pull/9556) Please DO NOT merge until both are ready to go**

The vets-api `api_session` cookie gets encrypted using the apps [secret_key_base](https://github.com/department-of-veterans-affairs/vets-api/blob/master/config/settings.yml#L64) The secret key base is stored in credstash (moving to parameter store) and accessed via settings.yml. Each environment has it’s own secret_key_base, located in the devops repo. 

This solution gracefully rotates out the old keys (secret_key_base) in order to avoid logging users out that have a current session. 

Upon visiting vets-api with a session encrypted with the old key, the session will be rotated and re-encrypted with the new secret_key_base value. Given this, currently logged in users have a chance to visit vets-api, get their cookie read with the old_secret_key_base and have it rewritten with the new secret_key_base.

In the devops repo, I’ve created a [branch](https://github.com/department-of-veterans-affairs/devops/tree/lhattamer-key-rotation) that adds the settings config for the secret_key_base and old_secret_key_base values. The devops changes and this PR will need to be merged at the same time. 

## Testing done (Locally)
1. Signed into vets-api locally
2. Grabbed session cookie
3. Open up a rails console
4. Copied the decrypt_session_cookie method from [cookie_spec](https://github.com/department-of-veterans-affairs/vets-api/blob/master/spec/rakelib/support/cookies_spec.rb) and pasted into the console
5. Decrypted session cookie / contents set aside for later
6. Killed the rails server
7. Rotated the secret_key_base and old_secret_key_base values
8. Restarted the server
9. Revisited va.gov locally
10. Decrypted the session cookie (it should be re-encrypted with the new secret_key_base value)
11. Compared the decrypted sessions. 
12. Decrypted contents were the same and the keys were gracefully rotated

## Original issue(s)
department-of-veterans-affairs/va.gov-team#23867

## Deploy/Testing Strategy
This change will be extensively tested in the dev environment before being released in production. With the rolling deploy, there is a brief moment where old and new instances may be up at the same time. For example, the case where a session is encrypted with the new key but a subsequent request gets routed to an old server that's in the process of being torn down, could result in an invalid session for a user.

#### Scenario
**Rolling deploy in process, new and old instances are up**
1. User creates a session, it get routed to the NEW server and the session is encrypted with the NEW key
2. User makes a subsequent request BUT the request is routed to the OLD server that contains the OLD encryption/decryption key
3. Users session cannot be decrypted (it’s encrypted with the NEW key but using the OLD key to decrypt)
4. Users session is marked as invalid… user is logged out

- [x] Parameter store values are updated
- [x] Testing in dev and staging
- [x] Discuss session timout refresh - It was discussed that the session timeout doesn’t need to be refreshed because the session is still the exactly the same after the rotation. The only piece that is changing is the key the session is encrypted with. The user will not experience any noticeable change and will still be working with their same session after getting their key rotated. Refreshing the timeout doesn't make sense in this scenario.


app/controllers/application_controller.rb
```diff
@@ -21,6 +21,7 @@ class ApplicationController < ActionController::API
   # for more before filters
   skip_before_action :authenticate, only: %i[cors_preflight routing_error]
   skip_before_action :verify_authenticity_token, only: :routing_error
+  before_action { session.delete :value }
```
LindseySaari, line 24: "It turns out, that when the app receives a cookie encrypted using the opposite key, it somehow adds a session item named :value with a value of an encrypted "null" string. Then, each time the cookies get read by the other app using the opposite secret_key_base, the “value” session keeps getting encrypted, which gets bigger in size, thus causing the error. While I haven’t found what causes this, fortunately, there’s a simple workaround. You just need to add a before_action filter to your base controller class that removes the problematic session."  https://blog.kmkonline.co.id/rotating-ruby-on-rails-secret-key-base-9a4dcdf0d817

app/controllers/concerns/authentication_and_sso_concerns.rb
```diff
@@ -89,6 +89,7 @@ def extend_session!
   def set_api_cookie!
     return unless @session_object
 
+    @session_object.delete :value
```
LindseySaari, line 92: "It turns out, that when the app receives a cookie encrypted using the opposite key, it somehow adds a session item named :value with a value of an encrypted "null" string. Then, each time the cookies get read by the other app using the opposite secret_key_base, the “value” session keeps getting encrypted, which gets bigger in size, thus causing the error. While I haven’t found what causes this, fortunately, there’s a simple workaround. You just need to add a before_action filter to your base controller class that removes the problematic session." https://blog.kmkonline.co.id/rotating-ruby-on-rails-secret-key-base-9a4dcdf0d817

app/controllers/concerns/authentication_and_sso_concerns.rb
```diff
@@ -89,6 +89,7 @@ def extend_session!
   def set_api_cookie!
     return unless @session_object
 
+    session.delete :value
```
LindseySaari, line 92: "It turns out, that when the app receives a cookie encrypted using the opposite key, it somehow adds a session item named :value with a value of an encrypted "null" string. Then, each time the cookies get read by the other app using the opposite secret_key_base, the “value” session keeps getting encrypted, which gets bigger in size, thus causing the error. While I haven’t found what causes this, fortunately, there’s a simple workaround. You just need to add a before_action filter to your base controller class that removes the problematic session." https://blog.kmkonline.co.id/rotating-ruby-on-rails-secret-key-base-9a4dcdf0d817

LindseySaari:
Question: Do we want to tear out this code once the keys are rotated?

rileyanderson (approved):
This looks good! I just want to double check that Sidekiq Web is still working since I just fixed some session issues with that

rileyanderson:
> Question: Do we want to tear out this code once the keys are rotated?

Don't we want to periodically rotate them?

LindseySaari:
> > Question: Do we want to tear out this code once the keys are rotated?
> 
> Don't we want to periodically rotate them?

I think we want to rotate the key on a yearly basis, so eventually there won't be any sessions encrypted with the old secret_key_base

LindseySaari:
> This looks good! I just want to double check that Sidekiq Web is still working since I just fixed some session issues with that

@rileyanderson Did you test against this branch?

f1337:
> > > Question: Do we want to tear out this code once the keys are rotated?
> > 
> > 
> > Don't we want to periodically rotate them?
> 
> I think we want to rotate the key on a yearly basis, so eventually there won't be any sessions encrypted with the old secret_key_base

We are going to want to rotate the keys more frequently than yearly. Possibly monthly. @joeniquette, what are your thoughts re: rotation frequency? We have Security Council meeting today, I will bring it up for discussion there as well.

f1337:
Overall, looking good! I don't have any merge-blockers for this PR, but one thing I do want to ensure we don't lose track of:

**The workstream isn't done until the key rotation is live.**

Muppet with sunglasses saying 'Looking good'

LindseySaari:
> > > > Question: Do we want to tear out this code once the keys are rotated?
> > > 
> > > 
> > > Don't we want to periodically rotate them?
> > 
> > 
> > I think we want to rotate the key on a yearly basis, so eventually there won't be any sessions encrypted with the old secret_key_base
> 
> We are going to want to rotate the keys more frequently than yearly. Possibly monthly. @joeniquette, what are your thoughts re: rotation frequency? We have Security Council meeting today, I will bring it up for discussion there as well.

Awesome! Thank you! The plan is that after [this PR](https://github.com/department-of-veterans-affairs/devops/pull/9597) is approved and merged, [this PR](https://github.com/department-of-veterans-affairs/vets-api/pull/7457) will be deployed via git ref and tested in the dev env. We plan to pay very close attention to the rolling deploy and will test the encrypted session after deploy to ensure that it's encrypted with the new key.

LindseySaari:
- [x] Testing in dev & staging. 
Next steps: Schedule an off cycle deploy after gathering metrics for prod and set up an alerting mechanism for rotating the keys

PR #7420, opened 2021-07-13
Kevin.cranmer/api 8100 postman webhook mvp

KevinCranmer:
## Description of change

Introduction of Postman testing capabilities for Webhooks. Included is a setup guide to create Postman monitors that gives us urls to send to the vba_documents benefits intake webhook subscription. We can then review the results of messages sent from benefits API to Postman.

## Original issue(s)
https://vajira.max.gov/browse/API-8100
https://vajira.max.gov/browse/API-8120

This PR is to create tests themselves. It does not introduce any new functionality.


mviterna (approved):
Approved, provided we don't need to mask the x-consumer-id. If we do, we will need to go back and scrub it from the git history

PR #7421, opened 2021-07-13
[API-8596] Use created_at for system-generated pdf timestamps

cilestin:
## Description of change
Veterans have X days to submit claims against decisions, and so stalls in the generation and/or submission of these pdfs could have led to forms being dated after their window expired even though the veteran submitted in time.

Fixed to use the original submission time (`created_at`) when generating these timestamps, to both prevent the above issue as well as generating the same consistent date.

This was changed for both the "date signed" field(s) as well as the timestamp placed on every page.

## Original issue(s)
https://vajira.max.gov/browse/API-8596


ElleeB (approved):
👍

PR #7442, opened 2021-07-14
Handle missing location phone numbers in mobile appointments

kreek:
## Description of change
Added a guard clause for missing mobile appointment location phone numbers so that they are nil and later null when serialized as JSON.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#27339


rileyanderson (approved):

jperk51 (approved):
LGTM

PR #7455, opened 2021-07-19
Claims V2: View claim by claim ID

hinzed1127:
## Description of change

Enables lookup of individual claims for a veteran by claim ID. Integrates with BGS to retrieve claim data, although if claims can be found within our own DBs, that takes precedence and gets returned first

## Original issue(s)
https://vajira.max.gov/browse/API-8401

~The diff is going to look bloated until https://github.com/department-of-veterans-affairs/vets-api/pull/7395 goes in. I branched off of that since this builds off of many of the same files~
Merge conflicts should be resolved now


modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
```diff
@@ -20,10 +19,25 @@ def index
           render json: claims
         end
 
+        def show
+          bgs_claim = bgs_service.benefit_claims.find_claim_details_by_claim_id(claim_id: params[:id]) || {}
```
SlakrHakr, line 23: we need to be able to accept our guid's for claims still in pending state as well

modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
```diff
@@ -20,10 +19,25 @@ def index
...
+          claim_details = bgs_claim.dig(:bnft_claim_detail)
+
+          # TODO: do we need to compare BGS vs Lighthouse claims like in claims#index?
+
+          if claim_details.present?
+            claim = { id: claim_details[:bnft_claim_id], type: claim_details[:bnft_claim_type_nm] }
+            render json: claim
+          else
+            raise ::Common::Exceptions::ResourceNotFound.new(detail: 'Claim not found')
+          end
```
SlakrHakr, line 33: can we switch this up to use (the thing i can never remember the name of 😞), like this:
```
raise ::Common::Exceptions::ResourceNotFound.new(detail: 'Claim not found') if claim_details.blank?

render json: { id: claim_details[:bnft_claim_id], type: claim_details[:bnft_claim_type_nm] }
```
i just think this structure always reads a bit better than nested if/else statements like this

modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
```diff
@@ -20,10 +19,25 @@ def index
...
         end
 
+        def show
+          bgs_claim = bgs_service.benefit_claims.find_claim_details_by_claim_id(claim_id: params[:id]) || {}
+          claim_details = bgs_claim.dig(:bnft_claim_detail)
+
+          # TODO: do we need to compare BGS vs Lighthouse claims like in claims#index?
+
+          if claim_details.present?
+            claim = { id: claim_details[:bnft_claim_id], type: claim_details[:bnft_claim_type_nm] }
```
SlakrHakr, line 29: > id:           the same ID passed in (our GUID if that was what the consumer submitted, evss_id if that was what the consumer submitted instead)

modules/claims_api/spec/fixtures/v2/veterans/claims/claim_by_id_response.json
```diff
@@ -0,0 +1,64 @@
+{
```
SlakrHakr, line 1: nit: mostly just to document my opinion here:
because this is a separate gem, i think specs within this project should mock the abstracted methods and not the actual bgs service responses. testing of the bgs service responses is for the gem to deal with, and this project shouldnt need to be aware of the underlying implementation of it's dependency.

SlakrHakr:
dont forget to re-generate openapi doc when all the rswag specs are written 👍🏻

modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
jw81: > "_the thing i can never remember the name of_"

Guard clause?

modules/claims_api/app/models/claims_api/auto_established_claim.rb
```diff
@@ -75,7 +75,7 @@ def self.evss_id_by_token(token)
 
     def self.get_by_id_or_evss_id(id)
       if id.to_s.include?('-')
-        find(id)
+        find_by(id: id)
```
hinzed1127, line 78: `find` raises a `RecordNotFound` error if there's not a matching ID. We do that with our own error messaging in the 2 calls to this method, so switching to `find_by` (returns `nil` if not found) and delegating the errors to the methods

modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
hinzed1127: I updated this aspect of the code. the logic here got refactored, but there's a different `if/else` due to the potential lookup in BGS, if needed

modules/claims_api/app/swagger/claims_api/v1/swagger.json
```diff
@@ -496,7 +496,7 @@
               "application/json": {
```
jw81, line 496: Doesn't look like anything really changed in this (other than obscure metadata).
Anytime you run the rswag logic, this file is going to have metadata modifications like this.  I've just been discarding changes like this so we don't end up pushing changes that aren't really "changes". (if that makes sense)  
Just something to consider in the future.

modules/claims_api/app/swagger/claims_api/v2/swagger.json
```diff
@@ -597,6 +597,282 @@
...
+            "name": "veteranId",
+            "in": "path",
+            "required": true,
+            "description": "ID of Veteran",
+            "schema": {
+              "type": "string"
+            }
+          },
+          {
+            "name": "claimId",
```
jw81, line 636: Hmmmm, I would think this should just be `id` because it looks like the controller is looking for [`params[:id]`](https://github.com/department-of-veterans-affairs/vets-api/pull/7455/files#diff-0f0dab107f9fa58ad7bfa8892ac1a5b69f069a82aac2d926c63f12e4839a0a8dR23).

modules/claims_api/spec/requests/v2/veterans/claims_request_spec.rb
```diff
@@ -26,38 +28,20 @@
...
-                .to receive(:user_represents_veteran?).and_return(false)
-
-              get path, headers: auth_header
-              expect(response.status).to eq(403)
-            end
-          end
-        end
-      end
-    end
-
```
jw81, line 53: Do we not want this spec?

modules/claims_api/app/swagger/claims_api/v1/swagger.json
hinzed1127: that's a good point. I'll make sure to only update v2 for the other changes below

modules/claims_api/spec/requests/v2/veterans/claims_request_spec.rb
hinzed1127: whoops, I just accidentally deleted that when copy-pasting around. I'll add it back in

modules/claims_api/app/swagger/claims_api/v2/swagger.json
hinzed1127: I definitely agree with the consistency angle, and `{id}` seems pretty conventional. Updated to change it to that.

modules/claims_api/spec/fixtures/v2/veterans/claims/claims_by_participant_id_response.json
```diff
@@ -0,0 +1,187 @@
+{
```
jw81, line 1: I wonder why this file is showing up as new.  i added it as part of my work.  🤔

modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
```diff
@@ -20,10 +19,27 @@ def index
...
+          if claim.present?
+            render json: { id: claim[:id], type: claim[:claim_type] }
+          else
+            # If we don't have it, it might still be in BGS, so check there
+            bgs_claim = bgs_service.benefit_claims.find_claim_details_by_claim_id(claim_id: params[:id])
+            claim_details = bgs_claim.dig(:bnft_claim_detail)
+
+            raise ::Common::Exceptions::ResourceNotFound.new(detail: 'Claim not found') if claim_details.blank?
+
+            render json: { id: claim_details[:bnft_claim_id], type: claim_details[:bnft_claim_type_nm] }
```
SlakrHakr, line 34: this logic isnt quite right with what the index endpoint is doing around "type"
should maybe get together with @jw81  and see if there is a way to reuse his mapper class

modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
```diff
@@ -20,10 +19,27 @@ def index
           render json: claims
         end
 
+        def show
+          claim = ClaimsApi::AutoEstablishedClaim.get_by_id_or_evss_id(params[:id])
+
+          if claim.present?
+            render json: { id: claim[:id], type: claim[:claim_type] }
```
SlakrHakr, line 26: AC in the ticket is to make sure that the id we pass back is the one the consumer passed in.
easiest way to do this would be to just send back params[:id] here instead

modules/claims_api/app/models/claims_api/auto_established_claim.rb
SlakrHakr: i wonder if this breaks expectations in v0 or v1 ?

modules/claims_api/spec/requests/v2/veterans/claims_request_spec.rb
```diff
@@ -84,11 +86,55 @@
...
+            expect_any_instance_of(BGS::BenefitClaimWebServiceV1)
+              .to receive(:find_claim_details_by_claim_id).and_return({})
+
+            get claim_by_id_path, headers: auth_header
+
+            expect(response.status).to eq(404)
+          end
+        end
+      end
+    end
```
SlakrHakr, line 138: Should be 401 and 403 cases here as well i assume?

modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
jw81: Maybe we could just extract [this logic](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/app/services/claims_api/bgs_to_lighthouse_claims_mapper_service.rb#L46-#L57) from the mapper service class, into some other thing that can be consumed by the mapper service and anywhere else that we need it.  Would also be good because it'll allow us to have a single place where we build these JSON structures.  Just thoughts.

modules/claims_api/spec/fixtures/v2/veterans/claims/claims_by_participant_id_response.json
hinzed1127: my bad, for some reason I thought I renamed it, but looks like I just duplicated and forgot to delete. will do 👍

modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
jw81: In my [current issue](https://vajira.max.gov/browse/API-8399), I'm switching to consuming a different BGS operation which is going to require me to change this mapping service class a fair amount.  If it's ok with @SlakrHakr, I can go ahead and handle adapting the service class to work for you as well while I'm in there.

modules/claims_api/app/models/claims_api/auto_established_claim.rb
hinzed1127: v1 was redundantly checking/raising 404s before. [I added that in v0, as well,](https://github.com/department-of-veterans-affairs/vets-api/pull/7455/commits/6572305515fea4df12be44c9bb758d287a7b4769) so I think we're good now

modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
hinzed1127: updated 👍

modules/claims_api/spec/requests/v2/veterans/claims_request_spec.rb
hinzed1127: I omitted them first pass since they technically are the same conditions as the other requests, didn't feel like they needed to be tested twice, but I don't mind adding them in

hinzed1127: added them in 👍

modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
SlakrHakr: i think this is the last thing here and im good approving this pr and moving forward if @jw81 is cool doing this work as a part of his ticket 👍🏻

SlakrHakr (approved):

modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
hinzed1127: cool, adding that as a note on the AC list of that ticket and merging this one

PR #7461, opened 2021-07-19
Handle and log malformed mobile appointment facility phone numbers

kreek:
## Description of change
Some facility phone numbers are causing our adapter to raise an error. A previous fix checked for blank phone numbers this PR will log the phone numbers that are causing the error so we can update the adapter to handle them.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#27555


jperk51 (approved):
LGTM

rileyanderson (approved):

PR #7482, opened 2021-07-21
API-8610: Investigate active_model_serializer alternatives for response serialization

SlakrHakr:
## Description of change
Uses "blueprinter" for claims v2 serialization.

## Original issue(s)
https://vajira.max.gov/browse/API-8610

Reasons this is not using active_model_serializers:
1. ams was configured for vets-api globally in several ways that made using it differently difficult in our module
2. vets-api is using a forked version of ams due to breaking changes in 0.10.5
3. looking at the repo for ams, ams seems pretty volatile (the repo even recommends several alternatives)
4. i came across several threads online where people discussed how ams might not be a long term solution any more and recommended alternatives (blueprinter being one of those)

Blueprinter seems super flexible based on this work at least.


modules/claims_api/app/controllers/claims_api/v2/blueprints/power_of_attorney_blueprint.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module ClaimsApi
+  module V2
+    module Blueprints
+      class PowerOfAttorneyBlueprint < Blueprinter::Base
+        identifier :code
+        field :name
+        field :type
+        field :phone do |entity, _options|
```
hinzed1127, line 10: is `_options` needed, or is this just to make one aware that it exists if needed later and the prepended `_` is the way to make `rubocop` ok with an used argument?

hinzed1127 (approved):
nice, the Blueprinter stuff felt super grok-able, hardly felt the need to reference the docs on a first pass

jw81 (approved):

modules/claims_api/app/controllers/claims_api/v2/blueprints/power_of_attorney_blueprint.rb
SlakrHakr: the latter

PR #7486, opened 2021-07-21
Api 8634/supplemental claim tech spike

yakschuss:
[API-8634](https://vajira.max.gov/browse/API-8634)

This PR is the preliminary refactors that we would want before introducing Supplemental Claim as a third decision reviews type.

It makes the EvidenceSubmissionRequestValidator type agnostic, as well as adds a v2 contestable issues controller that works for decision review types (and fixes up the shared examples to not rely on the v1 namespace)


modules/appeals_api/app/controllers/appeals_api/v2/decision_reviews/contestable_issues_controller.rb
```diff
@@ -0,0 +1,165 @@
...
+require 'common/exceptions'
+
+module AppealsApi::V2
+  module DecisionReviews
+    class ContestableIssuesController < AppealsApi::ApplicationController
+      SSN_REGEX = /^[0-9]{9}$/.freeze
+
+      skip_before_action(:authenticate)
+      before_action :validate_headers, only: %i[index]
+      before_action :validate_params, only: %i[index]
```
yakschuss, line 13: this before_action is the largest difference between V1 and V2

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
```diff
@@ -14,7 +14,8 @@ class EvidenceSubmissionRequestValidatorError < StandardError; end
 
         def create
           status, error = AppealsApi::EvidenceSubmissionRequestValidator.new(params[:nod_uuid],
-                                                                             request.headers['X-VA-SSN']).call
+                                                                             request.headers['X-VA-SSN'],
+                                                                             'notice_of_disagreement').call
```
cilestin, line 18: Why pass this in as snake case only to camelize in the initializer? Can't we just pass in the proper string format since it's not being supplied from anything the consumer hands us (like a param)?

modules/appeals_api/config/routes.rb
```diff
@@ -37,9 +37,7 @@
 
   namespace :v2, defaults: { format: 'json' } do
     namespace :decision_reviews do
-      # namespace :higher_level_reviews do
-      #   get 'contestable_issues(/:benefit_type)', to: 'contestable_issues#index'
-      # end
+      get 'contestable_issues/:decision_review_type', to: 'contestable_issues#index'
```
cilestin, line 40: Am I correct that for HLR contestable issues the `benefit type` will no longer be a part of the URL, but will be a param instead?

yakschuss: correct, because the endpoint will be for all of the appeal types.

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions_controller.rb
yakschuss: fair point, this was just a first pass. I'll change it

modules/appeals_api/config/routes.rb
cilestin: Sorry for not noticing this earlier, but is there a need to separate the decision review type? Couldn't we just ask for the benefit type and return that - or are there actual differences between what HLR vs NOD contestable issues return?

yakschuss: @cilestin we filter NOD where we don't filter HLR, and there's a definite need to have separation for future change. Not to mention that the Caseflow URL requires it.

ElleeB (approved):
looks good!

cilestin (approved):
LGTM! Thanks for the updates & answering questions 😄

modules/appeals_api/app/controllers/appeals_api/v2/decision_reviews/contestable_issues_controller.rb
```diff
@@ -0,0 +1,165 @@
...
+              code: 'unprocessable_entity',
+              detail: message,
+              status: 422
+            }
+          ]
+        }, code: 422
+      end
+
+      def request_headers
+        EXPECTED_HEADERS.index_with { |key| request.headers[key] }.compact
```
nathanbwright, line 138: Nice! I hadn't seen `index_with` before.

nathanbwright (approved):

PR #7491, opened 2021-07-22
Dependents Application PDF Upload to VBMS

kathleencrawford:
## Description of change

When submitting a Dependents Application (VA Form 21-686c), the form submission is processed by a Sidekiq job.  At the end of the job, it generates a PDF and then uploads it to VBMS.  This job was failing when uploading the PDF to VBMS, so this PR moves the PDF job to a different part of the 686c submission workflow.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#27651

- [x] Local testing
- [x] Specs updated
- [x] Tested on review instance


thilton-oddball (approved):
Looks good!

PR #7495, opened 2021-07-22
27418 mobile disability rating endpoint

jperk51:
## Description of change
Add disability endpoint to mobile module

## Original issue(s)
department-of-veterans-affairs/va.gov-team#27418

Added new spec and added endpoint to docs


kreek (approved):
Please add a spec that confirms the expected value of the endpoints response body, otherwise LGTM 🚀

modules/mobile/app/controllers/mobile/v0/disability_rating_controller.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+require_dependency 'mobile/application_controller'
+
+module Mobile
+  module V0
+    class DisabilityRatingController < ApplicationController
+      def index
+        response = VeteranVerification::DisabilityRating.for_user(@current_user)
```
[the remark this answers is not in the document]
jperk51: I could rebuild it in our module if that would be better?

kreek:
Now that we have the disability rating model in our module I think there's an opportunity to make some enhancements and align it with our patterns. Some of our, admittedly undocumented, rules (I should add this to the readme):

1.  Models (i.e. resources, non database backed) are immutable data buckets and should not fetch or adapt their own data.
2. Fetch via a proxy when reusing a service so we can make calls in parallel and customize error handling.
3. Perform data transformations in adapters.
4. Return the minimum amount of data.

So for this implementation:

1. Use `Common::Resource` as the model class to inherit from (`Virtus` is [no longer maintained](https://github.com/solnic/virtus#discontinued), `ActiveModel::Serialization` doesn't do anything since we use `FastJsonapi::ObjectSerializer`)
2. Proxy the `BGS::Services` class
3. Adapt the data (might be overkill for an adapter class a method could be fine) in that proxy and return a model
4. I believe we just need the combined disability rating (can confirm in the meeting today), only return that and the id :)

modules/mobile/app/models/mobile/v0/adapters/rating.rb
```diff
@@ -0,0 +1,55 @@
...
+          elsif response[:disability_rating_record][:ratings].instance_of?(NilClass)
+            response[:disability_rating_record][:ratings] = []
+          end
+          ratings = response[:disability_rating_record][:ratings].map do |rating|
+            Mobile::V0::IndividualRating.new(
+              decision: rating[:disability_decision_type_name],
+              effective_date: get_formatted_date(rating[:begin_date]),
+              rating_percentage: rating[:diagnostic_percent].to_i,
+              diagnostic_text: rating[:diagnostic_text],
+              type: rating[:diagnostic_type_name]
```
jperk51, line 36: Currently returning both diagnostic_text and diagnostic_type_name. Given the cassette in vets-api and the logic/tests on web it's hard to determine which of these are the more appropriate choice. Once this stuff goes in I'll check a few staging users and patch this to only return the one we want to use.

modules/mobile/app/controllers/mobile/v0/disability_rating_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+    class DisabilityRatingController < ApplicationController
+      def index
+        response = rating_proxy.for_user
+        render json: Mobile::V0::DisabilityRatingSerializer.new(response)
+      end
+
+      private
+
+      def rating_proxy
+        @rating_proxy ||= Mobile::V0::DisabilityRating::Proxy.new(@current_user)
```
kreek, line 16: In Rails a new controller instance is created on each request. This means as long as a class' method isn't accessed twice we shouldn't need to memoize (@thing ||= ) it. Places where memoization helps would be when we make parallel calls.

modules/mobile/app/services/mobile/v0/disability_rating/proxy.rb
```diff
@@ -0,0 +1,38 @@
...
+
+module Mobile
+  module V0
+    module DisabilityRating
+      class Proxy
+        def initialize(user)
+          @user = user
+        end
+
+        def for_user
```
kreek, line 13: `for_user` is a naming convention left over from when we returned an instance of a class with a local user reference. Could this describe the action taken? Maybe something like `.get_disability_ratings` or `.fetch_disability_ratings`.

kreek (approved):
Two minor comments but LGTM 🚀

kreek (approved):
LGTM 🚀

modules/veteran_verification/app/models/veteran_verification/disability_rating.rb
```diff
@@ -30,7 +30,7 @@ def self.for_user(user)
     end
 
     def self.handle_errors!(response)
-      raise_error! unless response[:disability_rating_record].class.eql? Hash
+      raise_error! unless response[:disability_rating_record].instance_of?(Hash)
```
jperk51, line 33: Tried reverting these, but the linter kept being upset about them

Teesh (approved):

thilton-oddball (approved):

PR #7499, opened 2021-07-22
API-8797:     Respond with 422 from 526 submission endpoint if consumer is a representative and the Veteran does not have a particpant_id

SlakrHakr:
## Description of change
If the currently authenticated consumer is a representative and the target Veteran does not have a participant_id, responds with an error telling the consumer that the Veteran must be authenticated to submit an original claim.

## Original issue(s)
https://vajira.max.gov/browse/API-8797

### DIFF
--- a/modules/claims_api/app/controllers/claims_api/v1/application_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/v1/application_controller.rb
@@ -12,7 +12,7 @@ module ClaimsApi
 
       skip_before_action :set_tags_and_extra_context, raise: false
       before_action :validate_json_format, if: -> { request.post? }
-      before_action :verify_mpi
+      before_action :validate_veteran_identifiers
 
       # fetch_audience: defines the audience used for oauth
       # NOTE: required for oauth through claims_api to function
@@ -21,6 +21,16 @@ module ClaimsApi
       end
 
       protected
+
+      def validate_veteran_identifiers
+        return if target_veteran.participant_id.present? && target_veteran.birls_id.present?
+        return unless header_request?
+        return if target_veteran.mpi_record?
+
+        raise ::Common::Exceptions::UnprocessableEntity.new(
+          detail: 'Submitting an original claim requires Veteran to be authenticated with an ID.me account'
+        )
+      end
 
       def source_name
         user = header_request? ? @current_user : target_veteran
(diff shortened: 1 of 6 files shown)


jw81 (approved):

modules/claims_api/app/controllers/claims_api/v1/application_controller.rb
```diff
@@ -12,7 +12,7 @@ class ApplicationController < ::OpenidApplicationController
 
       skip_before_action :set_tags_and_extra_context, raise: false
       before_action :validate_json_format, if: -> { request.post? }
-      before_action :verify_mpi
```
jw81, line 15: Are we replacing `veryify_mpi()`?  Just wondering if we should delete it f it's not going to be used anymore.

SlakrHakr: verify_mpi is still used in v0

modules/claims_api/app/controllers/claims_api/v1/application_controller.rb
```diff
@@ -22,6 +22,16 @@ def fetch_aud
 
       protected
 
+      def validate_veteran_identifiers
+        return if target_veteran.participant_id.present? && target_veteran.birls_id.present?
+        return unless header_request?
+        return if target_veteran.mpi_record?
+
+        raise ::Common::Exceptions::UnprocessableEntity.new(
+          detail: 'Submitting an original claim requires Veteran to be authenticated with an ID.me account'
```
hinzed1127, line 31: *requires the Veteran

modules/claims_api/app/controllers/claims_api/v1/application_controller.rb
```diff
@@ -22,6 +22,16 @@ def fetch_aud
 
       protected
 
+      def validate_veteran_identifiers
+        return if target_veteran.participant_id.present? && target_veteran.birls_id.present?
+        return unless header_request?
```
hinzed1127, line 27: could this be an `if` condition in the `before_action`?

hinzed1127: and then could just combine the other returns into
```ruby
return if (target_veteran.participant_id.present? && target_veteran.birls_id.present?) || target_veteran.mpi_record?
```

hinzed1127 (approved):
just a grammar thing and question on some ordering of checks in the validation method

modules/claims_api/app/controllers/claims_api/v1/application_controller.rb
SlakrHakr: i see where your mind is at but this method gets more complicated in subsequent prs around this logic. what you described might work here, but i dont believe it will work down the line

PR #7501, opened 2021-07-22
API-8799: Respond with 422 from 526 submission endpoint if Veteran has a particpant_id but is missing a birls_id

SlakrHakr:
## Description of change
Catch and respond with error if the Veteran somehow is in a state of having a participant_id but not having a birls_id. This shouldn't ever really happen as I understand it, but its not a state we can resolve if it somehow does.

## Original issue(s)
https://vajira.max.gov/browse/API-8799

### DIFF
--- a/modules/claims_api/app/controllers/claims_api/v1/application_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/v1/application_controller.rb
@@ -24,6 +24,11 @@ module ClaimsApi
 
       def validate_veteran_identifiers
         return if target_veteran.participant_id.present? && target_veteran.birls_id.present?
+
+        if target_veteran.participant_id.present? && target_veteran.birls_id.blank?
+          raise ::Common::Exceptions::UnprocessableEntity.new(detail: 'No birls_id while participant_id present')
+        end
+
         return unless header_request?
         return if target_veteran.mpi_record?
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,5 @@
 # Vets API
 
-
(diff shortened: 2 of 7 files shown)


SlakrHakr:
Will rebase and set this to merge into master once the following PR has been merged:
https://github.com/department-of-veterans-affairs/vets-api/pull/7499

jw81 (approved):

jw81 (dismissed):

modules/claims_api/app/controllers/claims_api/v1/application_controller.rb
```diff
@@ -24,6 +24,11 @@ def fetch_aud
 
       def validate_veteran_identifiers
         return if target_veteran.participant_id.present? && target_veteran.birls_id.present?
+
+        if target_veteran.participant_id.present? && target_veteran.birls_id.blank?
+          raise ::Common::Exceptions::UnprocessableEntity.new(detail: 'No birls_id while participant_id present')
```
hinzed1127, line 29: The wording of this is a little weird to me. I would have an easier time interpreting this error if it was simply "birls_id is missing for the veteran" or something like that

hinzed1127 (approved):
Question on wording, but otherwise good 👍

modules/claims_api/app/controllers/claims_api/v1/application_controller.rb
SlakrHakr: fair point regarding the wording, i just stole what va.gov is responding with:
https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/controllers/v0/mpi_users_controller.rb#L21

im not sure that "birls_id is missing for the veteran" gives the full picture though. a Veteran could be missing their birls_id and their participant_id and that would be fine. its specifically the situation where the Veteran has a participant_id but is missing a birls_id

jw81 (approved):

PR #7502, opened 2021-07-22
Don't retry on VANotify 400 errors

lihanli:
## Description of change
in the `VANotifyEmailJob`, log exceptions caused by 400 bad request errors and complete the job. retries will never succeed if the parameters are wrong. 

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/26980

tested with unit tests


thilton-oddball (approved):
LGTM

PR #7505, opened 2021-07-23
Feature flag: Move HLR wizard to `/start`

Mottie:
## Description of change

Adding a feature flag that moves the Higher-Level Review wizard from the `/introduction` page to the `/start` page.

## Original issue(s)

https://github.com/department-of-veterans-affairs/va.gov-team/issues/27745


Mottie:
Decided it was easier to use a production flag.

PR #7513, opened 2021-07-23
add rspec config for spec/serializers folder, and remove meta tags

RudyOnRails:
## Description of change

I am working on another branch where I have a new spec in `spec/serializers`. I noticed that the type (`type: :serializer`) was not being set correctly on the example metadata, so I've added it here to `rails_helper.rb` which removes the need to explicitly add it each spec in the directory.

## Original issue(s)
REFACTOR

Hopefully the suite passes ok 🍏


thilton-oddball (approved):
Nice work!

RudyOnRails:
@thilton-oddball thanks! :bowtie:

PR #7520, opened 2021-07-26
upgrade statsd-instrument to 3.1.1

cbtkc:
## Description of change
Upgrades `statsd-instrument` to 3.1.1

- `StatsD.backend` is now deprecated. Backend is configured only through environment variables now.
- `Rails.logger` no longer receives messages during testing, they are by default swallowed. Capturing increments is done using the `trigger_statsd_increment` method.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#24431


ericboehs:
What env vars need to be set in which environments once this is deployed?

thilton-oddball (approved):

cbtkc:
> What env vars need to be set in which environments once this is deployed?

`STATSD_ADDR` is the primary one and that one is set already. The other ones are fine with their defaults.

PR #7531, opened 2021-07-27
[26795] adds error handling for multiple BIRLS and CORP ids

bramleyjl:
## Description of change
As part of a broader effort to systematically manage how we handle users that have multiple conflicting IDs, this change extends existing checks for conflicting mhv_ids and edipis to also cover birls_ids and corp_ids. The second part of this change will be to implement an automatic monthly report of any users that triggered one of these mismatch errors so that their case can be resolved.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#26795

Unit tests updated, manual login/logout tests performed.


thilton-oddball (approved):

bosawt (approved):
Looks good!

PR #7532, opened 2021-07-27
Change EducationForm::Process10203Submissions job frequency

rileyanderson:
## Description of change
Changes the frequency the EducationForm::Process10203Submissions job to every 4 hours to avoid having multiple of the same job running at the same time.

### DIFF
--- a/config/sidekiq_scheduler.yml
+++ b/config/sidekiq_scheduler.yml
@@ -196,7 +196,7 @@ AppsApi::FetchConnections:
   description: "Fetches and handles notifications for recent application connections and disconnections"
 
 EducationForm::Process10203Submissions:
-  cron: "0 6-18/4 * * * America/New_York"
+  cron: "0 6-18/6 * * * America/New_York"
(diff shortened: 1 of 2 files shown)


config/sidekiq_scheduler.yml
```diff
@@ -196,7 +196,7 @@ AppsApi::FetchConnections:
   description: "Fetches and handles notifications for recent application connections and disconnections"
 
 EducationForm::Process10203Submissions:
-  cron: "0 6-18 * * * America/New_York"
+  cron: "0 6-18/6 * * * America/New_York"
```
ericboehs, line 199: This will run at 6am, 12 noon, and 6pm every day including weekends.

Previously this ran 12 times a day, now it will run three.

ericboehs (approved):

config/sidekiq_scheduler.yml
rileyanderson: Yeah, was wondering if we want to get rid of the `6-18` and let it run 4 times?

PR #7533, opened 2021-07-27
fix sql for in_progress_forms.rake

annaswims:
## Description of change
there was a new safety check included in a rails upgrade
https://github.com/rails/rails/issues/32995#issuecomment-392245004

## Original issue(s)
department-of-veterans-affairs/va.gov-team#20410


rakelib/in_progress_forms.rake
```diff
@@ -45,8 +46,8 @@ namespace :form_progress do
 
   def forms_with_args(args)
     form_id = args[:form_id] || '21-526EZ'
-    start_date = args[:start_date]&.to_date || 30.days.ago.utc
-    end_date = args[:end_date]&.to_date || Time.zone.now.utc
+    start_date = args[:start_date]&.to_date || 31.days.ago.utc
```
annaswims, line 49: also, shifting the dates so that we can exclude forms actively being worked on.

thilton-oddball (approved):

PR #7534, opened 2021-07-27
[API-8399] - ClaimsApi V2 return claims status

jw81:
## Description of change

Refactors the new/unreleased V2 Claims `index` and `show` actions to consume BGS `ebenefits` services.

## Original issue(s)
[API-8399](https://vajira.max.gov/browse/API-8399)

### DIFF
--- a/modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
@@ -98,10 +98,13 @@ module ClaimsApi
           claim_id.to_s.include?('-')
         end
 
+        # the 'ebenfits' services used in the 'index' and 'show' actions return differing data structures
+        #  massage the 'show' data structure to be in a state that the BGSToLighthouseClaimsMapperService can use
         def massage_bgs_claim(bgs_claim:)
           {
             benefit_claim_id: bgs_claim[:benefit_claim_details_dto][:benefit_claim_id],
-            claim_status_type: bgs_claim[:benefit_claim_details_dto][:claim_status_type]
+            claim_status_type: bgs_claim[:benefit_claim_details_dto][:claim_status_type],
+            phase_type: bgs_claim[:benefit_claim_details_dto][:bnft_claim_lc_status][:phase_type]
           }
         end
       end
--- a/modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
@@ -34,7 +34,7 @@ module ClaimsApi
           claim = BGSToLighthouseClaimsMapperService.process(
             bgs_claims: massaged_bgs_claim,
             lighthouse_claims: lighthouse_claim
-          )
+          ).first
 
           render json: ClaimsApi::V2::Blueprints::ClaimBlueprint.render(claim, base_url: request.base_url)
         end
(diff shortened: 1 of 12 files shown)


SlakrHakr:
LGTM
i  wish it didnt have to be quite so complicated, but after this base layer, it seems pretty straightforward to add stuff for the response in the mapper class at least

SlakrHakr (approved):

modules/claims_api/app/controllers/claims_api/v2/veterans/claims_controller.rb
```diff
@@ -41,6 +42,74 @@ def bgs_service
...
+        end
+
+        def find_bgs_claim_in_lighthouse_collection(lighthouse_collection:, bgs_claim:)
+          lighthouse_collection.find { |lighthouse_claim| lighthouse_claim.evss_id == bgs_claim[:benefit_claim_id] }
+        end
+
+        def find_lighthouse_claim!(claim_id:)
+          lighthouse_claim = ClaimsApi::AutoEstablishedClaim.get_by_id_or_evss_id(claim_id)
+
+          if looking_for_lighthouse_claim?(claim_id: claim_id) && lighthouse_claim.blank?
```
hinzed1127, line 79: is there a reason we need to check whether the `claim_id` is the lighthouse id vs EVSS id?

hinzed1127: nvm, I forgot that `evss_id` is misleading and serves as the main identifier in the BGS call, as well. Maybe just add a comment explaining that bit of business logic, since it's not obvious from the variable/method names here

modules/claims_api/app/services/claims_api/bgs_to_lighthouse_claims_mapper_service.rb
```diff
@@ -2,71 +2,46 @@
...
+    def initialize(bgs_claim: nil, lighthouse_claim: nil)
+      @bgs_claim        = bgs_claim
+      @lighthouse_claim = lighthouse_claim
     end
 
     def process
-      return [] unless claims_exist?
+      return matched_claim if bgs_and_lighthouse_claims_exist?
+      return unmatched_bgs_claim if bgs_claim_only?
+      return unmatched_lighthouse_claim if lighthouse_claim_only?
```
hinzed1127, line 15: nitpick: `unmatched_bgs_claim` and `unmatched_lighthouse_claim` probably don't need `unmatched_` prepended, since the `matched_claim` and `bgs/lighthouse_claim_only` names make it clear

hinzed1127 (approved):
I pretty much agree with @SlakrHakr 's sentiments. Only other thought it is maybe keeping the controller logic slimmer and throwing `map_claims` and `find_bgs_claim_in_lighthouse_collection` into the mapper, but that probably comes down more to where y'all think it makes more semantic sense to put. 

Code looks good, just had a few nitpicks and one section to clarify the logic going on

jw81:
> ... maybe keeping the controller logic slimmer and throwing `map_claims` and `find_bgs_claim_in_lighthouse_collection` into the mapper, but that probably comes down more to where y'all think it makes more semantic sense to put.

Yeah the original design of the mapper contained all of that mapping logic.  It took in a collection of lighthouse claims and a collection of bgs claims and handled iterating them and mapping.  But then we wanted the `show` action to be able to use the mapper as well.  Which ended up getting weird because the mapper expected collections, rather than single items.  So that's why all the iteration logic was removed from the mapper and the mapper just became a simple place to build hybrid objects.  🤷

PR #7536, opened 2021-07-27
form 526 - update rated disabilities when name, diagnostic_code or ra…

annaswims:
…ted_disability_id change

## Description of change
we already had code in place to handle when the name of a rated disability was updated by an upstream system, but we now see that rated disabilities need to be updated when diagnostic_code or rated_disability_id change;

## Original issue(s)
department-of-veterans-affairs/va.gov-team#22484


thilton-oddball (approved):
Looks good

PR #7542, opened 2021-07-28
7514 add ability to download dependent benefit letter

bosawt:
## Description of change
This PR updates the evss service headers so that a user who is a dependent can theoretically sign in and access the service to download a dependent benefit letter

## Original issue(s)
department-of-veterans-affairs/vets-api/issues/7514

Here are the things necessary for this to work:
- Current user must have loa3 (SSN, name, birth_date)
- Current user must have an edipi
- Current user must have a participant ID
- Current user must have a relationship entry in BGS (or MPI) that represents a sponsor veteran
- Current user must have vet360 contact info
- Current user must have headOfHousehold defined in their DS Logon credential
- Sponsor veteran must have loa3 attributes (SSN, name, birth_date)
- Sponsor veteran must have an edipi

I still need to test this to properly be confident pushing this forward, it'll be a bit of a pain to set up a test user with all these constraints

Options for the dependent benefits letter also must be confirmed, they are there for the benefits letter but there are some slight differences.

Tested with `vets.gov.user+dependent.91@gmail.com`, who has the above requirements


bramleyjl (approved):
LGTM!

LindseySaari (approved):

PR #7546, opened 2021-07-29
API-8943 Version 1.0 Veteran Verification API

dbrown-va:
## Description of change
A change for the Veteran Verification API module is requiring a new revision of the API.

Version 1.0.0 will include mostly unchanged API endpoints, but the Disability Rating endpoint will have dates transformed to `Date` only, no longer `Datetime`'s.   This is to improve consumer experience as the addition of time and timezone components is adding unnecessary complexity, especially since the underlying data set is only a date.

Follow up PR's will be completed in the Lighthouse API Gateway and eventually Developer Portal to advertise the new V1 endpoints.  Once all Sandbox and Production consumers have been migrated to V1, the V0 endpoints and routing will be removed under https://vajira.max.gov/browse/API-8963.

## Original issue(s)
- https://vajira.max.gov/browse/API-8943

- This change is a full addition of a version 1.0.0 of Veteran Verification, existing API calls to the v0 endpoint will be unchanged until a coordinated transition for consumers is completed.
- At this time, it is only the modification of the Disability Rating endpoint and return type for dates

- Unit tests were migrated forward for all endpoints
- Additional local testing comparing responses with V0
- `receive_message_chain` was flagged by linting and had to be replaced, even though just migration


esubocz-va (dismissed):
Tested functionality of all V0 and V1 endpoints, confirmed expected behavior. LGTM.

modules/veteran_verification/spec/models/disability_rating_spec_v1.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
```
thilton-oddball, line 1: I don't think this spec is running with the others. Possibly because the filename ends with _spec_v1.rb instead of just _spec.rb. When I run it by itself it is failing.

dbrown-va: Hmm, I thought I had run that rspec locally, but then when through quite a bit of Rubocop refactoring that was being enforced on existing code due to the v0/v1 folder addition.

I have corrected this and should be passing now.

esubocz-va (approved):

PR #7550, opened 2021-07-29
Api 8467 - rswag config

nathanbwright:
## Description of change
This is a work in progress PR to make rswag play nice with how vets-api consists of separate modules.

## Original issue(s)
https://vajira.max.gov/browse/API-8467

- Overriding the `Rswag::Specs::SwaggerFormatter#stop` method was due to two reasons 
  - Adding the ability to have multiple request bodies. (example below)
multiple_request_body

  - Adding a check to only write the swagger files for a given module (even through you can restrict which rswag examples with the `PATTERN` env option it doesn't change which json files get written, which seems odd to me).

I'm not super happy with monkey patching rswag and I tried several different ways to get the desired output but didn't have any success.


jw81:
LGTM!
I admittedly get a little lost in the monkey-patching of that `Rswag::Specs::SwaggerFormatter#stop` piece, but the rest is super straight-forward to me.  👍

jw81 (dismissed):

SlakrHakr (approved):
i like it 👍🏻

PR #7552, opened 2021-07-30
Add key rotation and rake task

LindseySaari:
## Background
Our database attrs are encrypted via the attr_encrypted gem. The key used to encrypt the attribute needs to be rotated (probably on a monthly basis, similar to the session key rotation) in an effort to decrease brute force and malicious user based vulnerabilities. 

If a malicious actor obtains the value of our db encryption key, they will then have full access to decrypt all the PII and sensitive data that we store in Postgres. 

Database attributes are encrypted via the apps [db_encryption_key](https://github.com/department-of-veterans-affairs/vets-api/blob/master/config/settings.yml#L57). The `db_encryption_key` is stored in parameter store and accessed via the settings.yml. Each environment has its own `db_encryption_key`, located in the devops repo

## Solution
This PR adds a rake task that allows for the graceful key rotation. The solution in place will decrypt the encrypted attributes with the old key and upon save, they will be re-encrypted with the new key. 

## Related PRs
https://github.com/department-of-veterans-affairs/devops/pull/9660
https://github.com/department-of-veterans-affairs/devops/pull/9659

## Current issues/unknowns:

1. The rake task to re-encrypt the existing records needs to be run IMMEDIATELY after the new key is deployed because the new key will be in place but the data is encrypted with the old key. 

2. With the above said, this will probably have to be run during a time with no traffic. How can we automate this? This will touch a lot of production data.

3. All instances of attributes utilizing the attr_encrypted gem are database fields **except one**. The only instance that is not a database field is [this field](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/covid_research/app/services/covid_research/volunteer/form_crypto_service.rb#L7) on the FormCryptoService. It was introduced in [this PR](https://github.com/department-of-veterans-affairs/vets-api/pull/4745) to add security for the form data arguments when the job is pushed to the Redis queue.)
-  My concern is that there could potentially be a brief moment when a job is in the Redis queue(before we deploy the new keys), we rotate the encryption keys and when the form data gets decrypted [here](https://github.com/department-of-veterans-affairs/vets-api/blob/e6483de9c146596e5daf49c648ff6936d5bd4023/modules/covid_research/app/workers/covid_research/volunteer/genisis_delivery_job.rb#L18) there will be a decryption error because it was encrypted with the old key. It doesn’t appear that this job runs often, but **we want to be aware of any jobs that are in the queue when rotating the keys** to avoid issues

3. The active record save’s in the rake task are wrapped in a transaction. If something goes wrong, we want it ALL to rollback to avoid having misaligned data (some encrypted with new key/some encrypted with the old key) 

4. Any new fields added to the database that utilize attr encrypted will need to be added to the rake task in this PR. Todo: Update or add documentation around this. 

5. This [gem](https://github.com/ankane/lockbox#active-record) has built in key rotation, might be something we want to look at since attr_encrypted does not 

For reviewer:
1. Please check for Typos
2. Please double check my work that I’ve got all the database fields covered  

## Testing Planned:
1. Test by running the rake task in the dev env. The commit in this PR will be deployed via git ref and [these changes](https://github.com/department-of-veterans-affairs/devops/pull/9673/files) will be merged/deployed as well.

2. Perform the above in staging

## Original issue(s)
department-of-veterans-affairs/va.gov-team#27317


modules/covid_vaccine/app/models/covid_vaccine/v0/expanded_registration_submission.rb
```diff
@@ -49,9 +51,13 @@ class ExpandedRegistrationSubmission < ApplicationRecord
...
-      attr_encrypted :form_data, key: Settings.db_encryption_key, marshal: true, marshaler: JsonMarshal::Marshaller
-      attr_encrypted :raw_form_data, key: Settings.db_encryption_key, marshal: true, marshaler: JsonMarshal::Marshaller
-      attr_encrypted :eligibility_info, key: Settings.db_encryption_key, marshal: true,
+      attr_encrypted :form_data, key: proc { |r|
+                                        r.encryption_key(:form_data)
+                                      }, marshal: true, marshaler: JsonMarshal::Marshaller
+      attr_encrypted :raw_form_data, key: proc { |r|
+                                            r.encryption_key(:raw_form_data)
+                                          }, marshal: true, marshaler: JsonMarshal::Marshaller
+      attr_encrypted :eligibility_info, key: proc { |r| r.encryption_key(:eligibility_info) }, marshal: true,
```
thilton-oddball, line 60: super nitpicky but `r.encryption_key(:eligibility_info)` could drop to the next line like the other examples above it.

LindseySaari: Ah, yes! I think that one must be JUST under the rubocop line limit!

LindseySaari: Welp, rubocop didn't like [my change](https://github.com/department-of-veterans-affairs/vets-api/runs/3221980173), reverting

patrickvinograd:
Building in a process that obligates vets-api to have downtime on whatever frequency you want to rotate keys seems unfortunate. 

The threat model here is not as clear as with the session key. 
* If a malicious actor gains access to a running vets-api instance then they'll also have access to the current db encryption key, rotation doesn't help us. 
* I guess this protects against somebody gaining access to an encryption key at time `X`, then at a later time `X+n months` gaining direct access to RDS/Postgres but not vets-api? Or getting a snapshot/db dump and then later getting access to the encyption key?

LindseySaari:
@thilton-oddball @rileyanderson or @ericboehs will have to review since @cbtkc and I are both contributors

rakelib/db_key_rotation.rake
```diff
@@ -25,245 +25,20 @@ namespace :attr_encrypted do
...
-        SavedClaim::EducationBenefits,
-        SavedClaim::VeteranReadinessEmploymentClaim
-      ]
-
-      saved_claim_types.each do |claim_type|
-        claim_type.all.each do |saved_claim|
-          old_form = saved_claim.form
-          saved_claim.form = old_form
-          saved_claim.save!
+    ApplicationRecord.descendants.each do |model|
```
LindseySaari, line 28: Love this! Way cleaner!

thilton-oddball (dismissed):
This all looks good! Great work! The one comment I have is that we probably need to figure out how and where we are going to let teams know that if they add an encrypted db field that it will need to be added to the rake task.

LindseySaari:
> This all looks good! Great work! The one comment I have is that we probably need to figure out how and where we are going to let teams know that if they add an encrypted db field that it will need to be added to the rake task.

I think [Camerons solution ](https://github.com/department-of-veterans-affairs/vets-api/pull/7552/files#diff-1e7f0bc7bd1066fb5eb28c290f276a28cefe57a55caaf5d408d4493ff38a2a34R28-R44)will now just collect all models that have encrypted attrs 🚀

ericboehs:
Should we close this one?

LindseySaari:
> Should we close this one?

Yeah I think we should and there may be a few others we can close as well.

PR #7556, opened 2021-07-30
Timecop freeze instead of Timecop travel to have expected timestamps

KevinCranmer:
## Description of change

Very rarely a spec test would fail because of timing issues. This PR freezes timestamps so the timestamps will always be what we expect of them.

## Original issue(s)
https://vajira.max.gov/browse/API-8996

No to all

The spec file in question was tested multiple times

### DIFF
--- a/modules/vba_documents/spec/models/upload_submission_spec.rb
+++ b/modules/vba_documents/spec/models/upload_submission_spec.rb
@@ -203,7 +203,7 @@ describe VBADocuments::UploadSubmission, type: :model do
           Timecop.freeze(time)
           upload = VBADocuments::UploadSubmission.new
           upload.consumer_name = "consumer_#{index}"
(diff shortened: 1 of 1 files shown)


modules/vba_documents/spec/models/upload_submission_spec.rb
```diff
@@ -203,7 +203,7 @@
           Timecop.freeze(time)
           upload = VBADocuments::UploadSubmission.new
           upload.consumer_name = "consumer_#{index}"
-          Timecop.travel(time + 1.minute)
+          Timecop.freeze(time + 1.minute)
```
cshupp1, line 206: Well, this just plain makes sense... HA, I see the problem now!

cshupp1 (approved):

PR #7561, opened 2021-08-02
Access tokens in openid controller are hashed

lukemarkwordtlibertyits:
## Description of change
Access tokens in the openid_application controller will be hashed

## Original issue(s)
https://vajira.max.gov/browse/API-6281

## Tests / Regression Tests

```
 ✓ Patient Token. Valid. With Launch/Patient scp. Return 200 with appropriate fields
```

image


dbrown-va (changes requested):
There is one other session look up modification that is needed in the Veteran Verification module surrounding the mock vs real emis support.

https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/veteran_verification/app/controllers/veteran_verification/application_controller.rb#L18

beaugrantham-va:
Can the hashing implementation be moved into a method?

```
def hash_token(token) ...
```

This will let us change the implementation (e.g., SHA-2 -> SHA-3) without having to update it in so many places, risking missing something.

modules/veteran_verification/app/controllers/veteran_verification/application_controller.rb
```diff
@@ -15,7 +15,7 @@ def authenticate_token
       # Not supported for Client Credentials tokens
       return false if token.blank? || token.client_credentials_token?
 
-      @session = Session.find(token)
+      @session = Session.find(Digest::SHA256.hexdigest(token.to_s))
```
dbrown-va, line 18: This should have similar `@session = Session.find(token) if @session.nil?` fallback logic, right?

dbrown-va (approved):
Looks good.

PR #7568, opened 2021-08-03
validate claim date

SlakrHakr:
## Description of change
We discovered several validations that EVSS would kick our claims back for so we decided to validate them before submitting upstream.
One of those validations is making sure `claimDate` is not in the future.

## Original issue(s)
** Jira is down **

### DIFF
--- a/modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
@@ -139,12 +139,12 @@ module ClaimsApi
 
         private
 
-        # 
+        #
         # Any custom 526 submission validations above and beyond json schema validation
-        # 
+        #
         def validate_form_526_submission_values
           return if form_attributes['claimDate'].blank?
-          return if Date.parse(form_attributes['claimDate']) <= Date.today
+          return if Date.parse(form_attributes['claimDate']) <= Time.zone.today
 
           raise ::Common::Exceptions::InvalidFieldValue.new('claimDate', form_attributes['claimDate'])
         end
--- a/modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
@@ -28,7 +28,7 @@ module ClaimsApi
         # @return [JSON] Record in pending state
         def submit_form_526 # rubocop:disable Metrics/MethodLength
           validate_json_schema
-          validate_form_526_submission_values
+          validate_form_526_submission_values!
           validate_initial_claim
 
           auto_claim = ClaimsApi::AutoEstablishedClaim.create(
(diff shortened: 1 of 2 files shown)


modules/claims_api/spec/requests/v1/disability_compensation_request_spec.rb
```diff
@@ -473,6 +475,24 @@ def obj.class
...
+
+    context 'when submitted claim_date is in the future' do
+      let(:claim_date) { (Date.today + 1.day).to_s }
+
+      before do
+        stub_mpi
+      end
+
+
+      it 'creates the sidekick job' do
```
jw81, line 486: We'd prob want a different test name here right?
Because it shouldn't be creating a job, but rather, raising an `InvalidFieldValue` exception.

modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
```diff
@@ -138,6 +139,16 @@ def validate_form_526
...
         private
 
+        #
+        # Any custom 526 submission validations above and beyond json schema validation
+        #
+        def validate_form_526_submission_values
+          return if form_attributes['claimDate'].blank?
+          return if Date.parse(form_attributes['claimDate']) <= Time.zone.today
+
+          raise ::Common::Exceptions::InvalidFieldValue.new('claimDate', form_attributes['claimDate'])
```
jw81, line 149: Isn't necessary for this specific ticket, but I'd imagine we'd want to do something like consolidate all the logical groupings into validation methods.  

Something like ::

```
def validate_form_526_submission_values
  validate_claim_date!
  validate_some_other_random_piece!
  validate_yet_another_random_piece!
end

private

def validate_claim_date!
  return if form_attributes['claimDate'].blank?
  return if Date.parse(form_attributes['claimDate']) <= Time.zone.today
  
  raise ::Common::Exceptions::InvalidFieldValue.new('claimDate', form_attributes['claimDate'])
end
```

modules/claims_api/spec/requests/v1/disability_compensation_request_spec.rb
SlakrHakr: :(

modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
SlakrHakr: yeah, probably, i can go ahead and get that setup here anyways though

SlakrHakr: https://github.com/department-of-veterans-affairs/vets-api/pull/7568/commits/37116890a4d979f8979d8dd3a0d2b56af8166bbb

modules/claims_api/spec/requests/v1/disability_compensation_request_spec.rb
SlakrHakr: https://github.com/department-of-veterans-affairs/vets-api/pull/7568/commits/37116890a4d979f8979d8dd3a0d2b56af8166bbb

jw81 (approved):

PR #7579, opened 2021-08-04
allow vets to resume a 526 form when EVSS is not responding

annaswims:
## Description of change
http://grafana.vfs.va.gov/d/000000066/form-526-disability-compensation?orgId=1&from=now-7d&to=now
Screen Shot 2021-08-04 at 4 38 49 PM
resuming an in-progress was failing when EVSS is down (nightly). Updating rated disabilities is nice to have but not essential so they should be able to continue if evss errors

## Original issue(s)
department-of-veterans-affairs/va.gov-team#28300


LindseySaari (approved):

michelpmcdonald (approved):
Looks good

PR #7586, opened 2021-08-05
Simplify HLRv2 Schema

cilestin:
## Description of change
Compresses HLR v2 schema to include far fewer references that were then turned into schemas in the API docs.
Also adds a couple new GET schemas to the API docs for hlrShow and contestableIssues requests.

## Original issue(s)
https://vajira.max.gov/browse/API-8569


yakschuss (approved):

nathanbwright (approved):
🚀

PR #7587, opened 2021-08-05
Add EVSS validations for 526 submit requests

hinzed1127:
These are mostly small, so I'm gonna continue grouping tickets from the [EVSS validation epic](https://vajira.max.gov/browse/API-9026) but keeping things under the 500 LOC limit so it's still reasonable
## Original issue(s)
https://vajira.max.gov/browse/API-9004

### DIFF
--- a/modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
@@ -144,6 +144,7 @@ module ClaimsApi
         #
         def validate_form_526_submission_values!
           validate_form_526_submission_claim_date!
+          validate_form_526_claimant_certification!
         end
 
         def validate_form_526_submission_claim_date!
@@ -151,6 +152,10 @@ module ClaimsApi
           return if Date.parse(form_attributes['claimDate']) <= Time.zone.today
 
           raise ::Common::Exceptions::InvalidFieldValue.new('claimDate', form_attributes['claimDate'])
+        end
+
+        def validate_form_526_claimant_certification!
+          raise ::Common::Exceptions::InvalidFieldValue.new('claimantCertification', form_attributes['claimantCertification']) unless form_attributes['claimantCertification'] == true
         end
 
         def flashes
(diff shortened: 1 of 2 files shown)


LindseySaari (approved):

modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
```diff
@@ -153,6 +154,13 @@ def validate_form_526_submission_claim_date!
           raise ::Common::Exceptions::InvalidFieldValue.new('claimDate', form_attributes['claimDate'])
         end
 
+        def validate_form_526_claimant_certification!
+          unless form_attributes['claimantCertification'] == true
+            raise ::Common::Exceptions::InvalidFieldValue.new('claimantCertification',
+                                                              form_attributes['claimantCertification'])
+          end
```
SlakrHakr, line 161: nit: something ive gotten accustomed to with ruby is favoring guard clauses in these types of situations
i started doing it because rubocop's default favors them
https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/GuardClause

so this could turn into something like
```
return unless form_attributes['claimantCertification'] == true

raise ::Common::Exceptions::InvalidFieldValue.new('claimantCertification',
                                                                                  form_attributes['claimantCertification'])
```
its a minor thing but i personally think it helps readability in a lot of situations by reducing nesting

hinzed1127: that is a ruby convention that I still have to remind myself of 🙃 I've got no problem updating that

LindseySaari (approved):

modules/claims_api/spec/fixtures/form_526_json_api.json
```diff
@@ -50,7 +50,7 @@
       "standardClaim": false,
       "autoCestPDFGenerationDisabled": true,
       "claimDate": "1990-01-03",
-      "applicationExpirationDate": "2015-08-28T19:53:45+00:00",
+      "applicationExpirationDate": "2300-08-28T19:53:45+00:00",
```
hinzed1127, line 53: updated to pass validation requirement that the expiration date is in the future

SlakrHakr: can we use Timecop here instead?

SlakrHakr (approved):

modules/claims_api/spec/fixtures/form_526_json_api.json
hinzed1127: I assume you mean within the request spec? updated to that 👍

SlakrHakr (approved):

PR #7592, opened 2021-08-06
28064 address subtle bug

dillo:
- Faraday might be using the URI base instead of the additional path
- Adding the base path as config in settings
## Original issue(s)
department-of-veterans-affairs/va.gov-team#28064

- [x] RSpec passing


mdewey (approved):
LGTM

LindseySaari (approved):

PR #7595, opened 2021-08-06
LOB required in V2

KevinCranmer:
## Description of change

Benefits Intake has been accepting LOB in the file upload metadata previously. This change (for v2 only) will now make it required. If a user does not submit a LOB or an invalid LOB, they will be given and error detailing LOB is required when they try to check the upload status.

The valid LOB list has also changed. This change now reflects what is in our documentation.

## Original issue(s)
https://vajira.max.gov/browse/API-9113
https://vajira.max.gov/browse/API-7954

Only required if the v2_enabled flag in Settings.local.yml is set to true

Automated Postman tests were ran and passed for both V1 and V2 for the following test cases:
- valid
- missing LOB
- lowercase LOB
- integer LOB
- space LOB
- unknown LOB


mviterna (approved):

modules/vba_documents/app/controllers/vba_documents/v2/uploads_controller.rb
```diff
@@ -82,11 +84,12 @@ def submit
         upload_model = UploadFile.new
         begin
           upload_model.multipart.attach(io: StringIO.new(request.raw_post), filename: upload_model.guid)
+          upload_model.metadata['version'] = 2
           upload_model.save!
           parts = VBADocuments::MultipartParser.parse(StringIO.new(request.raw_post))
           inspector = VBADocuments::PDFInspector.new(pdf: parts)
           validate_parts(parts)
-          validate_metadata(parts[META_PART_NAME])
+          validate_metadata(parts[META_PART_NAME], submission_version: upload_model.metadata['version'])
```
cshupp1, line 92: For upload_model.metadata['version'] it looks like the possible values are nil,1, or 2?

If we pass in: upload_model.metadata['version'].to_i, nil  => 0, 1 => 1, 2 =>2

modules/vba_documents/lib/vba_documents/upload_validator.rb
```diff
@@ -71,10 +71,15 @@ def validate_names(first, last)
       end
     end
 
-    def validate_line_of_business(lob)
-      return if lob.to_s.empty?
+    def validate_line_of_business(lob, submission_version)
+      return if lob.to_s.empty? && !(submission_version && submission_version >= 2)
```
cshupp1, line 75: If to_i is called (comment above) the nil check can go

modules/vba_documents/lib/vba_documents/upload_validator.rb
```diff
@@ -71,10 +71,15 @@ def validate_names(first, last)
...
 
-      unless VALID_LOB.keys.include?(lob)
+      if lob.to_s.blank? && Settings.vba_documents.v2_enabled && submission_version >= 2
```
cshupp1, line 77: Why do we need the settings check here?  We couldn't have added the value of two unless we got to the v2 route or am I missing something?

modules/vba_documents/spec/jobs/upload_processor_spec.rb
```diff
@@ -503,14 +503,12 @@
...
 
-      %w[LOG MED BUR OTH DROC].each do |future_lob|
-        it "maps the future line of business #{future_lob} to CMP" do
-          @md['businessLine'] = future_lob
-          described_class.new.perform(upload.guid, test_caller)
-          updated = VBADocuments::UploadSubmission.find_by(guid: upload.guid)
-          expect(updated.uploaded_pdf['line_of_business']).to eq(future_lob)
-          expect(updated.uploaded_pdf['submitted_line_of_business']).to eq('CMP')
-        end
+      it 'maps the future line of business OTH to CMP' do
```
cshupp1, line 506: We have this expectation:
    it 'sets error status and records invalid lines of business' do

Can we get a similar one?  Such as:
    it 'requires line of business for version 2' do

cshupp1 (changes requested):

cshupp1 (approved):

modules/vba_documents/lib/vba_documents/pdf_inspector.rb
```diff
@@ -74,8 +74,8 @@ def inspect_pdf(add_file_key)
 
     def add_line_of_business(data, parts_metadata)
       if parts_metadata.key? 'businessLine'
-        data['line_of_business'] = parts_metadata['businessLine']
-        data['submitted_line_of_business'] = VALID_LOB[parts_metadata['businessLine']].to_s
+        data['line_of_business'] = parts_metadata['businessLine'].to_s.upcase
+        data['submitted_line_of_business'] = VALID_LOB[parts_metadata['businessLine'].to_s.upcase]
```
LindseySaari, line 78: I see that the to_s was on the outside of the closing bracket and now moved inside. Just checking to make sure this is what's expected

modules/vba_documents/lib/vba_documents/upload_validator.rb
```diff
@@ -99,7 +104,7 @@ def perfect_metadata(model, parts, timestamp)
         metadata["ahash#{i + 1}"] = att_info[:hash]
         metadata["numberPages#{i + 1}"] = att_info[:pages]
       end
-      metadata['businessLine'] = VALID_LOB[metadata['businessLine']].to_s if metadata.key? 'businessLine'
+      metadata['businessLine'] = VALID_LOB[metadata['businessLine'].to_s.upcase] if metadata.key? 'businessLine'
```
LindseySaari, line 107: Same comment here

modules/vba_documents/lib/vba_documents/pdf_inspector.rb
KevinCranmer: Yes that is what is expected. VALID_LOB has string keys, and one requirement was to capitalize businessLine for the users. If a user submits a nonstring, we cannot capitalize it unless we first convert to a string. This way we will throw errors for nonstring values without having to directly check them because they won't be in the VALID_LOB keys

PR #7618, opened 2021-08-10
API-9173 Added Veteran Verification Metadata Controller

dbrown-va:
## Description of change
Veteran Verification API needs to expose a metadata endpoint to advertise both the deprecated v0 and new v1 endpoints.

```
{
    "meta": {
        "versions": [
            {
                "version": "0.0.1",
                "internal_only": false,
                "status": "Previous Version",
                "path": "/services/veteran_verification/docs/v0/veteran_verification",
                "healthcheck": "/services/veteran_verification/v0/health"
            },
            {
                "version": "1.0.0",
                "internal_only": false,
                "status": "Current Version",
                "path": "/services/veteran_verification/docs/v1/veteran_verification",
                "healthcheck": "/services/veteran_verification/v1/health"
            }
        ]
    }
}
```

## Original issue(s)
- https://vajira.max.gov/browse/API-9173

* Metadata controller will be used by Developer Portal for swagger rendering of Veteran Verification API docs
* V1 is configured as the primary resource
* Route and rspec added as well


modules/veteran_verification/app/controllers/veteran_verification/metadata_controller.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+require 'appeals_api/health_checker'
```
beaugrantham-va, line 3: Should this module rely on appeals?

dbrown-va: Nope!  Was a missed dependency removal when porting over the metadata controller.  Will get that removed.

beaugrantham-va (approved):

PR #7631, opened 2021-08-12
Kevin mark/api 8673 sha256 uploads

mviterna:
## Original issue(s)


modules/vba_documents/lib/vba_documents/multipart_parser.rb
```diff
@@ -63,12 +64,13 @@ def self.create_file_from_base64(infile)
         contents = content.sub %r{data:((multipart)/.{3,}),}, ''
       end
       decoded_data = Base64.decode64(contents)
+      record_sha256(submission, 'base64', decoded_data) if submission
```
KevinCranmer, line 67: Does this mean we're recording it twice for a Base64 upload? I think line 73 sends it back up to parse and this time, it's no longer base64 which means it goes to parse_file where we're already recording the sha256

modules/vba_documents/lib/vba_documents/multipart_parser.rb
```diff
@@ -166,9 +167,22 @@ def self.consume_body_tempfile(lines, separator)
           tf.rewind
           return tf, true
         else
+          # AWS appends a new line at the end of the pdf, we must remove it to maintain the original sha256 value
+          line.chomp! if line.encoding.name == 'ASCII-8BIT' && /%%EOF\n\r\n/.match?(line)
```
KevinCranmer, line 171: If we are removing a line in the file, and then recording the sha256, but AWS is storing the file with the extra newline, then wouldn't our upstream (maybe downstream? Don't know those terms yet haha but whoever is next after us) comparisons not match? Should we be removing the newline from the file in AWS?

modules/vba_documents/lib/vba_documents/multipart_parser.rb
mviterna: We are adding a sha256 for the entire base64 payload, as well as individual content and metadata. when they submit base64, it contains both, so we do the sha on the whole base64, then metadata and content, attachment etc separately

modules/vba_documents/lib/vba_documents/multipart_parser.rb
mviterna: It shouldn't be, but I did see some tweaks I want to make on base64 payloads. We remove the newline because s3 stores it with an appended newline, so what we really want is the sha BEFORE it hits aws. But I'm going to go ahead and mock GCIO and make sure the sha matches at the end of that workflow and add another unit test, just so we can have that extra warm and fuzzy

KevinCranmer (approved):

KevinCranmer (approved):

PR #7633, opened 2021-08-12
[API-9010] - Adds EVSS validation checks for Form526 "homelessness" attributes

jw81:
## Description of change

Adds logic to validate that the `veteran.homelessness` attributes provided in a Form526 submission meet EVSS expectations.

## Original issue(s)
[API-9010](https://vajira.max.gov/browse/API-9010)


modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
```diff
@@ -188,6 +189,32 @@ def validate_form_526_location_codes!
           end
         end
 
+        def validate_form_526_veteran_homelessness!
+          if too_many_homelessness_attributes_provided? || unnecessary_homelessness_point_of_contact_provided?
```
SlakrHakr, line 193: nit: i would use go with a guard clause here personally. less nesting the easier to read imo

SlakrHakr (approved):

modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
jw81: I would too, but it's make the line too long.  😞

PR #7634, opened 2021-08-12
send ICN to vanotify if email isnt present

yakschuss:
[API-8568](https://vajira.max.gov/browse/API-8568)

This PR utilizes VaNotify's [capability](https://github.com/department-of-veterans-affairs/notification-api/blob/daba8c5b687143699e6c205a66b51c9d0c850b18/documents/openapi/openapi.yaml#L1623) to look up a Veteran's info using any one of the MPI compatible identifiers.

We hook into Claim's abstract class `Veteran` to query MPI for the ICN using the data submitted with the HLR.

Future refactors may have Claims and Appeals both inherit from the same Veteran abstraction, and remove the need for LOA for server-side (non-oauth specific) requests.


cilestin (approved):

modules/appeals_api/app/models/appeals_api/veteran.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module AppealsApi
+  class Veteran
+    delegate_missing_to :@veteran
+
+    def initialize(ssn:, first_name:, last_name:, birth_date:)
+      @veteran = ClaimsApi::Veteran.new(
```
nathanbwright, line 8: Curious to hear your perspective on the advantage of wrapping the `ClaimsApi::Veteran` object, if we're passing any and all messages onto the wrapped object. What are we gaining over just instantiating a regular `ClaimsApi::Veteran` object?

modules/appeals_api/app/services/appeals_api/events/appeal_received.rb
```diff
@@ -34,28 +27,58 @@ def vanotify_service
...
+      end
+
       def template_id(template)
         Settings.vanotify.services.lighthouse.template_id.public_send(template)
       end
 
+      def personalisation
+        {
+          'first_name' => opts['first_name'],
+          'date_submitted' => opts['date_submitted'].strftime('%B %d, %Y')
```
nathanbwright, line 60: Probably not worth it yet... But we might want extract this time formatting pattern if we end up using it more...

nathanbwright (approved):

modules/appeals_api/app/models/appeals_api/veteran.rb
yakschuss: If we control the interface, we can make changes as needed to our public API, as well as switch out the internal implementation when/if we decide to talk to MPI ourselves. This decouples our client code (such as higher_level_review.rb lookup of ICN) from the claims implementation of Veteran.

modules/appeals_api/app/services/appeals_api/events/appeal_received.rb
yakschuss: For sure, this was done very quickly to satisfy the AC! I think it can be refactored in the future.

PR #7643, opened 2021-08-13
Kmm improve ch31 logging add pid etc

RudyOnRails:
## Description of change

This improves logging for trying to figure out the Ch31 issues.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/28692


LindseySaari:
Was it determined that the pid doesn't need to be sanitized?

RudyOnRails:
> Was it determined that the pid doesn't need to be sanitized?

I was told by @jason-gcio that it is not PII, and that is why we are already emailing a nightly report of VRE submissions.

LindseySaari:
@thilton-oddball @rileyanderson or @ericboehs will have to review since @cbtkc and I are both contributors

rileyanderson (approved):

PR #7645, opened 2021-08-13
Ensure PDF generation can support accented characters

cilestin:
## Description of change
* Ensure that some romanized special characters (accented characters) are supported in our PDF generation. Allows for characters in the Windows-1252 space.
* Adds character limits to documentation

## Original issue(s)
https://vajira.max.gov/browse/API-9087
https://vajira.max.gov/browse/API-9088

* Using Windows-1252 was more a limitation of the built-in fonts when doing pdf generation. There are options to expand the character support (embedding external fonts), but since so far our production data has yet to have anything beyond what the built-in fonts support, we can cross that bridge if it's required.


cilestin:
Closing for now to ensure CentralMail can handle Windows-1252 as well, since we send the veteran's first & last name to their service as metadata.

PR #7661, opened 2021-08-16
API-9008 - Validate serviceInformation.confinements for 526 submissions

hinzed1127:
## Description of change

This adds the following validation checks to `serviceInformation.confinements`:
    - Each confinement period must fall within one of the `servicePeriod` dates.
    - All `confinementBeginDate`s must be after the earliest `servicePeriod.activeDutyBeginDate`
    - If there are more than 2 confinements, they cannot overlap with the dates of another confinement period.

I grouped the first 2 validations within the same method, since, since any period that meets the first condition automatically meets the second, as well.

## Original issue(s)
https://vajira.max.gov/browse/API-9008


modules/claims_api/spec/requests/v1/disability_compensation_request_spec.rb
```diff
@@ -15,6 +15,7 @@
 
   before do
     stub_poa_verification
+    stub_mpi
```
hinzed1127, line 18: Put this in the top-level before and removed from all the individual tests to remove some repetitive code. Note there are some cases where we stub MPI not found to override this

modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
```diff
@@ -190,6 +191,44 @@ def validate_form_526_location_codes!
...
+
+          raise ::Common::Exceptions::InvalidFieldValue.new('confinements', confinements)
+        end
+
+        def confinements_within_service_periods?(confinements, service_periods)
+          confinements.each do |confinement|
+            next if service_periods.any? do |service_period|
+              time_range = service_period['activeDutyBeginDate']..service_period['activeDutyEndDate']
+              time_range.cover?(confinement['confinementBeginDate']) &&
+              time_range.cover?(confinement['confinementEndDate'])
```
jw81, line 211: Should we maybe convert these `xBeginDate` & `xEndDate` values into actual Date objects and then build our ranges?
Right now, we'll get a range of `String` values and it will include values like `1984-00-00` and `1984-99-99`.

So something like this would be possible.
```
range1 = '1980-03-05'..'1985-01-07'
range1.include?('1984-66-99')  #will be true
```

Just wondering if we need to protect against that by making our ranges be date based.

hinzed1127: good catch. I was doing that conversion in the request spec, but I should be doing that here, instead

modules/claims_api/spec/requests/v1/disability_compensation_request_spec.rb
SlakrHakr: hmmm, im surprised this would work this way... typically an "allow_any_instance_of" run more than once will just stack responses, not override previous ones.

hinzed1127: maybe "override" was the wrong word choice. I can look into the exact reason why this doesn't cause issues and revert if explicitly mocking per test is easier to reason through

modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
hinzed1127: Addressed [here](https://github.com/department-of-veterans-affairs/vets-api/pull/7661/commits/c8daeded90a3e781f7a6022b7333f7957b11a52b

SlakrHakr (approved):

PR #7673, opened 2021-08-17
rate-limit PPMS lookup to curb abuse (REDUX)

f1337:
In response to ongoing DOS activity against the Facility Locator remote Bing Maps lookup.

Please see https://github.com/department-of-veterans-affairs/va.gov-team-sensitive/blob/master/Postmortems/2021-08-16-facility-locator-possible-DOS.md for add'l details.

Now with `remote_ip`, because we're behind an ELB.

### DIFF
--- a/spec/middleware/rack/attack_spec.rb
+++ b/spec/middleware/rack/attack_spec.rb
@@ -94,6 +94,26 @@ RSpec.describe Rack::Attack do
     end
   end
 
+  describe 'facility_locator/ip' do
+    before do
+      limit.times do
+        post endpoint, headers: headers
+        expect(last_response.status).not_to eq(429)
+      end
+
+      post endpoint, headers: headers
+    end
+
+    context 'limit PPMS abuse' do
+      let(:limit) { 3 }
+      let(:endpoint) { '/facilities_api/v1/ccp/provider' }
+
+      it 'limits requests' do
+        expect(last_response.status).to eq(429)
+      end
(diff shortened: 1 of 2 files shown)


LindseySaari (approved):
💯 🙇‍♀️

spec/middleware/rack/attack_spec.rb
```diff
@@ -94,6 +94,34 @@ def app
     end
   end
 
+  describe 'facility_locator/ip' do
+    subject { last_response.status }
+
+    let(:endpoint) { '/facilities_api/v1/ccp/provider' }
+    let(:limit) { 3 }
```
LindseySaari, line 101: Since 3 is the rate limit, should we set the test to something greater than 3, which would then throw a 429 (I think)?

LindseySaari: Ah nvm, I think line 109 is the fourth call to the endpoint. Scratch that

rileyanderson (approved):

rileyanderson (approved):

spec/middleware/rack/attack_spec.rb
```diff
@@ -101,11 +101,11 @@ def app
 
     before do
       limit.times do
-        get endpoint, params: nil, session: headers
```
LindseySaari: BAD RUBOCOP 😠 👊

PR #7674, opened 2021-08-17
Api 8568/vanotify icn lookup 2

yakschuss:
[API-8568](https://vajira.max.gov/browse/API-8568)

This PR utilizes VaNotify's capability to look up a Veteran's info using any one of the MPI compatible identifiers.

We hook into Claim's abstract class Veteran to query MPI for the ICN using the data submitted with the HLR.

Future refactors may have Claims and Appeals both inherit from the same Veteran abstraction, and remove the need for LOA for server-side (non-oauth specific) requests.


modules/appeals_api/app/services/appeals_api/events/appeal_received.rb
```diff
@@ -34,28 +27,58 @@ def vanotify_service
...
+          lookup,
+          template_id,
+          personalisation
+        ].reduce(&:merge)
+      end
+
+      def lookup
+        if opts['email_identifier'] == 'email'
+          {
+            email_address: opts['email_identifier']['id_value']
```
cilestin, line 41: If `opts['email_identifier']` is a string (when it checks for `== 'email'`, how can it also be a hash with a `id_value` key?

yakschuss: Good catch

PR #7675, opened 2021-08-17
Mobile disability rating service change

jperk51:
## Description of change
Switching which disability rating service mobile is using

## Original issue(s)
department-of-veterans-affairs/va-mobile-app#1345

Added new suite of tests to ensure our custom errors are being triggered


modules/mobile/app/services/mobile/v0/disability_rating/proxy.rb
```diff
@@ -11,26 +11,35 @@ def initialize(user)
         end
 
         def get_disability_ratings
-          response = rating_service.rating.find_rating_data(@user.ssn)
-          handle_errors!(response)
-          Mobile::V0::Adapters::Rating.new.disability_ratings(response)
+          combine_response = common_service.get_rating_info
+          individual_response = compensation_service.get_rated_disabilities
```
kreek, line 15: Looks like this request is going to wait for the first one to complete. Could they be called in parallel like appointments and claims? (This could also follow in a separate PR)

kreek (approved):
One comment but otherwise LGTM 🚀

modules/mobile/app/services/mobile/v0/disability_rating/proxy.rb
jperk51: Updated

kreek (approved):
LGTM 🚀

PR #7676, opened 2021-08-17
API-8467 - Add additional hlr rswag examples

nathanbwright:
## Description of change
Adds `/higher_level_reviews/validate` rswag spec.

## Original issue(s)
https://vajira.max.gov/browse/API-8467

### DIFF
--- a/modules/appeals_api/spec/docs/v2/hlr_spec.rb
+++ b/modules/appeals_api/spec/docs/v2/hlr_spec.rb
@@ -150,7 +150,7 @@ describe 'Higher-Level Reviews', swagger_doc: 'modules/appeals_api/app/swagger/a
       response '200', 'Info about a single Higher-Level Review' do
         schema AppealsApi::SwaggerSharedComponents.response_schemas[:hlr_response_schema]
 
-        hlr = FactoryBot.create(:higher_level_review)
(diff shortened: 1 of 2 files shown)


modules/appeals_api/spec/docs/v2/hlr_spec.rb
```diff
@@ -353,5 +353,166 @@
...
+            }
+          }
+        end
+
+        it 'Violates JSON schema' do |example|
+          assert_response_matches_metadata(example.metadata)
+        end
+      end
+
+      response '422', 'Error' do
```
cilestin, line 487: Not necessary now, but I'd like to see these more common errors that are the same across our endpoints made into support helpers so we could do things like `it_behaves_like 'Violates JSON Schema'`, etc :)

cilestin (approved):
Lookin' good!

Rswag's tests are a little hard to parse out, but isn't too bad overall. Just something to get used to.

nathanbwright:
> Lookin' good!
> 
> Rswag's tests are a little hard to parse out, but isn't too bad overall. Just something to get used to.

Yeah, I had a little trouble reading them at first too... It would be nice to look into making them more reader friendly.

PR #7679, opened 2021-08-18
[28888] makes Vista Account Numbers available on User model

bramleyjl:
## Description of change

In order to support the Healthcare Experience Team, a users's Vista Account Numbers, as derived from their GCID string, needs to be made available as a public-facing method on the User model.

These account numbers are tied to VHA Facility IDs and derived from the same string in the GCID that comes back from MPI. As such, we decided to make them available in a hash with the `vha_facility_ids` as keys and an array of vista account numbers as the value. Given that this is providing somewhat duplicate functionality to the existing `vha_facility_ids` methods in the User model class, the Identity Team plans to refactor these two parses and getter methods into one at some point in the future.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#28888

Unit tests were added to cover the parsing and User model getter methods, Rails console testing confirmed the existence of the `vha_facility_hash` on multiple test user accounts.


dillo (dismissed):
Looks good to me!

jscanish:
Looks good!

dillo (approved):

cbtkc (approved):
LGTM

jscanish (approved):

PR #7685, opened 2021-08-18
Api 9107 service history gras integration

lukemarkwordtlibertyits:
# Description

Added filtering for Title 32 Reserve periods.
https://vajira.max.gov/browse/API-9107

Why: The base get_military_service_episodes_v2 service includes potentially non-active guard service obligations like training, rather than actual service history.

As such, we need more advanced filtering to return the correct service periods for benefit eligibility.

Service History will now be a combination of the Military Service Episodes and Guard Reserve Service Periods responses, with the filtering below.
```
1. get_military_service_episodes_v2, personnel_category_type_code
  - If A, then active and should be included
  - If V, N, or Q, then guard periods and require additional filtering

2. get_guard_reserve_service_periods, trainingIndicatorCode
      - If Y, then should be filtered out
      - If N, then more filtering is required
        3. get_guard_reserve_service_periods, guardReservePeriodStatuteCode
           - If [J,N,P,Q,Z] then title 32 and should be filtered out
           - Otherwise, then active duty guard period and should be included
```
# Related PRs

https://github.com/department-of-veterans-affairs/vets-api-mockdata/pull/173 -> this PR needs to merge first


dbrown-va (approved):
Looks good.  Confirmed locally with Active duty and Guard Reserve response test scenarios.

Noting that for GRAS responses, there are fields unavailable like Pay Grade and Branch (of Guard/Reserve).  This is expected as the underlying datasource does not contain this information in the response.

PR #7686, opened 2021-08-18
[API-9023] - ClaimsApi Forms526 "servicePay" schema clean up

jw81:
## Description of change

This PR lays the groundwork necessary to add new EVSS validations around the `servicePay` attribute in a Form526 submission. 

**Removes unnecessary "requiredness" of some attributes**
 Our existing schema for the `servicePay` attribute had multiple sub-attributes that were marked as required.  However, after comparing against both the [ebenefits Form526 endpoint/schema](https://www.ebenefits.va.gov/wss-form526-services-web-v2/swagger-ui/index.html?url=https://www.ebenefits.va.gov/wss-form526-services-web-v2/rest/swagger.yaml#/form526/submit) and the translation work performed by va.gov, it was discovered that those attributes don't need to be required.  ([va.gov endpoint data](https://department-of-veterans-affairs.github.io/va-digital-services-platform-docs/api-reference/#/form_526/postSubmitFormV2) gets [translated before sending to EVSS](https://github.com/department-of-veterans-affairs/vets-api/blob/0c270a530492b43a498271e52f622035d7b4023a/lib/evss/disability_compensation_form/data_translation_all_claim.rb#L49))

**Adds many attributes to the schema that don't currently exist**
Many of the new validations that need to be added, are for attributes that don't exist in our current schema.  So this PR adds those attributes as a first-step before adding the new validations.  A new PR will be created to add all of the new validations.

## Original issue(s)
[API-9023](https://vajira.max.gov/browse/API-9023)


modules/claims_api/spec/requests/v1/disability_compensation_request_spec.rb
```diff
@@ -791,85 +789,85 @@ def obj.class
           end
         end
       end
+    end
 
-      describe "'disabilities.ratedDisabilityId' validations" do
```
jw81, line 795: I noticed that these `disabilities.ratedDisabilityId` specs I recently added accidentally got enclosed in another context/scope.  So I moved them out to the appropriate level.

SlakrHakr (approved):

PR #7694, opened 2021-08-19
28780 sanitize data before sending it to ppms

darkbushido:
## Description of change

PPMS is throwing a stack trace if we send 13 digits of accuracy on lat/long, its something we get from mapbox.
this also limits the radius to 500 miles
this is only for the facilities engine

## Original issue(s)
department-of-veterans-affairs/va.gov-team#28780

* rounds lat and long to 6 digits
* clamps radius from 1 to 500
* adds tests to verify these are working


cbtkc (approved):
LGTM

PR #7698, opened 2021-08-19
Reactivate Direct Deposit only for loa3 id.me users

lihanli:
## Description of change
Reactivate Direct Deposit only for loa3 id.me users
MFA is no longer required

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/28903

tested with unit tests


app/policies/ppiu_policy.rb
```diff
@@ -4,7 +4,7 @@
 
 PPIUPolicy = Struct.new(:user, :ppiu) do
   def access?
-    user.loa3? && user.multifactor && Flipper.enabled?(:direct_deposit_cnp, user)
+    user.loa3? && user.identity.sign_in[:service_name] == 'idme' && Flipper.enabled?(:direct_deposit_cnp, user)
```
kreek, line 7: Mobile users don't go through the same provider as .gov users. This means that users who select id.me as the sign in method will have 'oauth_IDME' returned for `user.identity.sign_in[:service_name]`.
```suggestion
    user.loa3? && ['idme', 'oauth_IDME'].include?(user.identity.sign_in[:service_name]) && Flipper.enabled?(:direct_deposit_cnp, user)
```

cbtkc (approved):

PR #7701, opened 2021-08-19
add rake task to test for old STI data

LindseySaari:
## Description of change
An issue was found after running the attr_encrypted rake task to rotate db keys in the dev environment. We had to rollback the db key rotation changes in dev.

## Issue 1:
We currently have a model HCAAttachment that inherits from FormAttachment, but there were some records in the db BEFORE we implemented inflections (so the type = HcaAttachment…lowercase) and that’s where the issue is

I ran FormAttachment.where(type: "HcaAttachment").update_all(type: "HCAAttachment")
And we’re golden now, but there may be other records like this.

## Issue 2:
See [this ticket](https://app.zenhub.com/workspaces/vsp---backend-tools-6019ab9387ae890011e05e73/issues/department-of-veterans-affairs/va.gov-team/29444)

When running the rake task in staging, this issue popped up:`ActiveRecord::SubclassNotFound (The single-table inheritance mechanism failed to locate the subclass errors for the SavedClaim records` The issue is most likely due to old/bad data hanging around where the subclass no longer exists. The type column is used for STI, but that type (child class) no longer exists. 

Having this rake task in place will help to point out bad data.

I’ve since migrated the data manually in dev, but want to run this task to surface other issues of the same nature.

Running this task WILL NOT migrate the data as it will just re-save with the current encryption key.

We will need to run another rake task to migrate any mismatched record types as a follow up to this.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#27317


cbtkc (dismissed):
looks good!

cbtkc (approved):

PR #7695, opened 2021-08-19
22440 Betamocks and Swagger

dillo:
## Description of change

- Dependent on https://github.com/department-of-veterans-affairs/vets-api-mockdata/pull/172
- Medical Copays `/v0/medical_copays` endpoint transforms keys to camelcase for clients
- Setup basic swagger docs
- Setup betamocks
- Faraday connection uses breakers
- Adding to Debt team CODEOWNERS
## Original issue(s)
department-of-veterans-affairs/va.gov-team#22440

## Things to know about this PR

- [x] Rubocop passing
- [x] RSpec passing

PR #7708, opened 2021-08-20
add placeholder url for mcp.vbs.url

annaswims:
otherwise I see
`ArgumentError: bad argument (expected URI object or URI string)
(erb):34:in <main>` when I try to start the rails server or console in development.

Caused by https://github.com/department-of-veterans-affairs/vets-api/pull/7695

cc: @dillo


annaswims:
duplicate of #7706

PR #7719, opened 2021-08-23
Add rswag setup doc

nathanbwright:
## Description of change
Creating a doc to show a minimum implementation of Rswag in vets-api (to make it play nice with rails engines).


cbtkc (approved):
👍

PR #7720, opened 2021-08-23
Add new profile create_or_update APIs

lihanli:
## Description of change
we are seeing a lot of errors with users trying to POST changes to va profile when a record already exists. this pr adds new APIs that first check if a record exists before calling the right POST or PUT request

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/28398

tested with unit tests


cbtkc (approved):
Looks good!

PR #7727, opened 2021-08-24
Validate "disabilities.diagnosticCode" values in 526 submissions

hinzed1127:
## Description of change

Adds the following checks to 526 submission validations for `disabilities.diagnosticCode`:

- Becomes required if `disabilityActionType` equals "INCREASE".
- Becomes required if `disabilityActionType` equals "NONE" AND there are secondary disabilities included with the primary.
   
There are 2 additional requirements that I did not add in here because the extra lift did not seem worth it, at least in this PR
- Must match an existing rated disability in corporate if `disabilityActiontype` equals "INCREASE".
- Must match an existing rated disability in corporate if `disabilityActiontype` equals "NONE" AND there are secondary disabilities included with the primary.

There is a BGS service for getting diagnostic codes, but we don't have access currently. Also, that's not quite the same thing as what this validation is looking for. If there's thoughts on a way to easily do what the validation is checking, I'm down to add it in, but I'm not sure that there is.

## Original issue(s)
https://vajira.max.gov/browse/API-9014


modules/claims_api/config/schemas/526.json
```diff
@@ -975,17 +975,33 @@
             }
```
SlakrHakr, line 975: I'm guessing rswag needs re-generate our doc since this file was changed?
https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/README.md#L5

SlakrHakr (approved):

SlakrHakr (changes requested):
Sorry! I didn't mean to accept it and now the only way to "unaccept" it is to request changes :(
All I'm requesting is to re-generate our doc with rswag though
https://github.com/department-of-veterans-affairs/vets-api/pull/7727/files#r695232509
Everything else looks great 👍🏻

PR #7728, opened 2021-08-24
API-9260 - Add initial NOD rswag spec

nathanbwright:
## Description of change
Initial rswag NOD spec.

## Original issue(s)
https://vajira.max.gov/browse/API-9260


modules/appeals_api/spec/docs/v2/nod_spec.rb
```diff
@@ -0,0 +1,497 @@
...
+
+        before do |example|
+          submit_request(example.metadata)
+        end
+
+        it 'all fields used' do |example|
+          assert_response_matches_metadata(example.metadata)
+        end
+
+        after do |example|
```
cilestin, line 101: Maybe I'm mixing my PRs, but I feel like I've seen this exact `after` block a number of times now. Would be another good candidate for extracting into a helper or two. Perhaps even our own block that can set up the `before`s and `afters` a little more cleanly and just have us get into the real business of `it` tests and not all this boilerplate? Food for thought.

modules/appeals_api/spec/docs/v2/nod_spec.rb
```diff
@@ -0,0 +1,497 @@
...
+        { apikey: [] }
+      ]
+      consumes 'application/json'
+      produces 'application/json'
+
+      parameter name: :nod_body, in: :body, schema: { '$ref' => '#/components/schemas/nodCreateRoot' }
+
+      parameter in: :body, examples: {
+        'minimum fields used' => {
+          value: JSON.parse(File.read(AppealsApi::Engine.root.join('spec', 'fixtures', 'valid_10182_minimum.json')))
```
cilestin, line 248: Helpers for these JSON parsing from path fixtures would be nice soon, too.

cilestin (approved):

PR #7729, opened 2021-08-24
updates populate task to overwrite existing tud accounts [29023]

bramleyjl:
## Description of change

This change fixes/updates the TUD population script to be able to run repeatedly on the same table to update existing accounts. Only fields that have values will be updated; most notable the `notes` field that is intended for users to fill out will not be affected if the account in the .csv has no notes.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#29023

Tested on empty and pre-populated test_user_dashboard_tud_accounts tables


modules/test_user_dashboard/app/models/test_user_dashboard/tud_account.rb
```diff
@@ -4,16 +4,15 @@ module TestUserDashboard
   class TudAccount < ApplicationRecord
     self.ignored_columns = %w[standard available]
 
-    ID_PROVIDERS = %w[id_me dslogon mhv].freeze
+    ID_PROVIDERS = %w[idme dslogon mhv logingov].freeze
 
     validates :first_name, :last_name, :email, :gender, presence: true
     validates :email, uniqueness: true
+    validates :id_type, presence: true
+    validates :id_type, inclusion: { in: ID_PROVIDERS }
```
jscanish, line 12: A minor thing, but I would probably make this one line - `validates :id_type, presence: true, inclusion: { in: ID_PROVIDERS }`. You also validate a bunch of attributes for presence in line 9 so it could also make sense for you to move line 11 there.

modules/test_user_dashboard/app/services/test_user_dashboard/create_test_user_account.rb
```diff
@@ -12,9 +12,15 @@ def call
       account = Account.create(idme_uuid: row.delete('idme_uuid'))
       birth_date = Date.parse(row.delete('birth_date')) if row.key?('birth_date')
       services = row.delete('services').split(',')
```
jscanish, line 14: This is fine and will definitely work, but how would you feel about moving a lot of these variables into their own private methods? It's kind of a stylistic thing, but I would argue that it's good practice and more Ruby-like/object-oriented to have smaller methods that each handle a much more narrow workload, rather than one method that gathers all the data it needs and does the heavy lifting for the whole class.

So `birth_date`, for example, could be rewritten as a method like this. Then you're ultimately left with a cleaner `call` method that has a narrower focus.
```
def birth_date
  return unless row.key?('birth_date')
  Date.parse(row.delete('birth_date'))
end
```

modules/test_user_dashboard/app/services/test_user_dashboard/create_test_user_account.rb
```diff
@@ -12,9 +12,15 @@ def call
...
-                                   services: services))
+      account_details = row.merge(account_uuid: account.uuid,
+                                  birth_date: birth_date,
+                                  services: services).compact
+
+      if (tud_account = TudAccount.where(email: row['email']).first)
+        tud_account.update!(account_details)
+      else
+        TudAccount.create!(account_details)
+      end
```
jscanish, line 23: This might be a good place for a method like [find_or_initialize_by](https://apidock.com/rails/v4.0.2/ActiveRecord/Relation/find_or_initialize_by)

jscanish (approved):
I have some minor thoughts but overall it looks good.

PR #7731, opened 2021-08-25
[Bugfix] Transliterate to acceptable centralmail characters

cilestin:
## Description of change
Fixes bug where names like `O'Reilly` were erroring when sending pdfs to Centralmail (its metadata has a small range of acceptable characters, which does NOT include `'`)

## Original issue(s)
https://vajira.max.gov/browse/API-9471


modules/appeals_api/app/workers/concerns/appeals_api/character_utilities.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module AppealsApi
+  module CharacterUtilities
+    extend ActiveSupport::Concern
+
+    included do
+      def transliterate_for_centralmail(str)
+        I18n.transliterate(str.to_s).gsub(/[^a-zA-Z\-\s]/, '').strip.first(50)
```
cilestin, line 9: Note this may not have ALL the characters centralmail allows, but it seemed a decent set to get us through the bug.

ElleeB (approved):
looks good

PR #7738, opened 2021-08-25
HLR - Create v1 service that uses lighthouse 2.0 HLR

michelpmcdonald:
## Original issue(s)


michelpmcdonald:
Can't this to pass unit test, closing

PR #7764, opened 2021-08-27
Track logins with account login stats

jscanish:
## Description of change
This populates the `account_login_stats` table as part of the existing AfterLoginJob. The goal is to gather data on how accounts are choosing to authenticate. We will save the most recent login time for each of the three methods (mhv, dslogon, id.me). The next step after this PR is to create a job that rolls up this data and makes it accessible so that we have a better understanding of how many accounts log in and how they choose to do so.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/28911

Added a series of RSpec tests to ensure correct behavior.

This pull request must be merged after https://github.com/department-of-veterans-affairs/vets-api/pull/7763


bosawt (commented):
Logged in, confirmed that `AfterLoginJob` created an `AccountLoginStat` entry with:

`#<ActiveRecord::Relation [#<AccountLoginStat id: 1, account_id: 21, idme_at: "2021-08-31 23:17:08.169392000 +0000", myhealthevet_at: nil, dslogon_at: nil, created_at: "2021-08-31 23:16:56.571099000 +0000", updated_at: "2021-08-31 23:17:08.175890000 +0000">]>`

app/controllers/concerns/accountable.rb
```diff
@@ -14,8 +14,27 @@ def create_user_account
...
+    return unless login_stats.present? && login_type.in?(AccountLoginStat::LOGIN_TYPES)
+
+    login_stats.update("#{login_type}_at" => Time.zone.now)
+  end
+
   private
 
+  def login_stats
+    @login_stats ||=
+      if @current_user.account.blank?
```
bosawt, line 27: It might be interesting to have a log or Sentry error when account is blank off a current_user. We could monitor and see if this ever happens (which it probably shouldn't), and if we can confirm it never happens, maybe we can make the assumption that a user has an account.

You could potentially do that in this PR, or a future PR, if you decided to add that, either way works IMO

app/models/account_login_stat.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+# It's important that the datetime columns on this table fit a "#{name}_at" pattern
+# where the name matches one of the service_names in the SAML::User::AUTHN_CONTEXTS hash.
+class AccountLoginStat < ApplicationRecord
+  # ['idme', 'myhealthevet', 'dslogon']
+  LOGIN_TYPES = SAML::User::AUTHN_CONTEXTS.map { |_k, v| v[:sign_in][:service_name] }.uniq.freeze
```
bosawt, line 7: This makes me think we could pull out constants in SAML::User, maybe into a file like `authn_constants.rb` or something like that, at least try to attempt to simplify/centralize these Authn definitions.

I wouldn't say that's a part of this PR, but maybe an idea for clarity in the future

app/controllers/concerns/accountable.rb
bosawt: Oh wait, I see in `AfterLoginJob`, we try to create an account right before adding a record in `AccountLoginStats`, it's a log with `cannot_create_unique_account_record` if an Account isn't created. We could look that up in Sentry and see if we ever hit it

bosawt (dismissed):
Looks good! I tested it and it appears to work great!

I just threw out ideas for maybe future work in future PRs, but nothing I think you should worry about here.

app/controllers/concerns/accountable.rb
jscanish: I looked this up. There have been 219 examples in the last 30 days, so failing to create an account is something that can definitely happen.

app/controllers/concerns/accountable.rb
```diff
@@ -14,8 +14,27 @@ def create_user_account
...
+        nil
+      else
+        AccountLoginStat.find_or_initialize_by(account_id: @current_user.account.id)
```
ericboehs, line 30: You could invert the if condition's logic to `if @current_user.account.present?` and then you wouldn't need an else block.

ericboehs (dismissed):

bosawt (approved):

app/controllers/concerns/accountable.rb
```diff
@@ -14,8 +14,23 @@ def create_user_account
...
+    return unless login_stats.present? && login_type.in?(AccountLoginStat::LOGIN_TYPES)
+
+    login_stats.update("#{login_type}_at" => Time.zone.now)
+  end
+
   private
 
+  def login_stats
+    @login_stats ||= AccountLoginStat.find_or_initialize_by(account_id: @current_user.account.id) \
+      if @current_user.account.present?
```
ericboehs, line 27: I believe RuboCop might have had too much motor oil. 🍸 
```suggestion
    @login_stats ||=
      @current_user.account.present? &&
      AccountLoginStat.find_or_initialize_by(account_id: @current_user.account.id)
```
Give this a shot. I think there's something weird with line length here, RuboCop should not be recommending inline if modifier if it makes the line too long.

ericboehs (approved):

bosawt (approved):

PR #7765, opened 2021-08-27
API-9152 static token validation support

dbrown-va:
## Description of change
This change adds support for isolated static token with the Lighthouse token validation service.

## Original issue(s)
https://vajira.max.gov/browse/API-9132

* Static tokens are isolated to test scenarios specifically defined in the OAuth Proxy
* If `Settings.oidc.issued_url` is not set, all opaque tokens are disabled

Added corresponding rspec for happy path and mismatched aud 401 responses.  A non-jwt 401 scenario was already in place.

Related configuration PR isolated to dev can be found below.
https://github.com/department-of-veterans-affairs/devops/pull/9846


cbtkc (approved):
👍

app/controllers/openid_application_controller.rb
```diff
@@ -198,6 +215,27 @@ def fetch_smart_launch_context
...
+    if response.code == 200
+      json_response = JSON.parse(response.body)
+      if json_response['scopes']
+        json_response['scp'] = json_response['scopes']
+        json_response['scopes'] = nil
+      end
+      json_response
+    end
+  rescue => e
+    raise raise error_klass('Invalid token') if e.to_s.include?('Unauthorized')
```
beaugrantham-va, line 234: Is `raise raise` correct or a typo?

beaugrantham-va, line 234: What conditions cause this rescue block to execute? Any non 200 response?

dbrown-va: Typo, should just be a single raise.  Will correct.

dbrown-va: This would be if the call succeeds to /issued endpoint, but there isn't a corresponding entry for the access token.

And then the raise below is if the call itself fails for any other reason, like misconfigured endpoint.

cbtkc (approved):

beaugrantham-va (approved):

PR #7796, opened 2021-08-31
1449 mobile log when an appointment's facility can not be found

kreek:
## Description of change
Upstream appointments include a facility or sta6aid which vets-api uses to look up location data such as address and phone number. Sometimes an appointment's facility can not be found. This PR updates the facility adapter to log the facility or sta6aid causing the issue and returns an unmodified appointment (location remains null).

## Original issue(s)
https://app.zenhub.com/workspaces/va-mobile-eng-60f1a34998bc75000f2a489f/issues/department-of-veterans-affairs/va-mobile-app/1449


kreek:
@narin does the mobile app currently handle null appointment locations?

narin (approved):

PR #7805, opened 2021-09-01
API-9020: [Form526 EVSS Validations] - "disabilities.secondaryDisabilities"

SlakrHakr:
## Description of change
Several validations around accepted values of secondary disabilities payload within a 526 submission.

## Original issue(s)
https://vajira.max.gov/browse/API-9020


modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
```diff
@@ -350,22 +350,123 @@ def validate_form_526_disabilities!
...
+                    secondary_disability['name'].length <= 255
+
+          raise ::Common::Exceptions::InvalidFieldValue.new(
+            'disabilities.secondaryDisabilities.name',
+            secondary_disability['name']
+          )
+        end
+
+        # rubocop:disable Metrics/MethodLength
+        def validate_form_526_disability_secondary_disability_approximate_begin_date!(secondary_disability)
```
jw81, line 392: We talked about it offline, but FWIW, in a couple places, I took a different approach with these "approximateDate" fields that EVSS accepts.  Rather than our API accepting the optional components of an approximate date, I've been having our API accept valid date patterns. (ie. `yyyy-mm-dd`). Then I convert the valid date into the components required by EVSS.  That approach seemed cleaner to me, rather than the approach EVSS is taking where they accept components of a date and sometimes some of those date components are required based on the existence of other date components.  (i.e. You can send in just a `year`.  Or you can send in a `year` and a `month`.  Or you can send in a `year`, `month`, and `day`.  But you can't pass in a `day` and a `year`.).  That just felt like clunky and confusing to me.  🤷

modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
```diff
@@ -350,22 +350,123 @@ def validate_form_526_disabilities!
...
           end
         end
 
+        def bgs_classification_ids
+          return @bgs_classification_ids if @bgs_classification_ids.present?
+
+          contention_classification_type_codes = bgs_service.data.get_contention_classification_type_code_list
+          @bgs_classification_ids = contention_classification_type_codes.pluck(:clsfcn_id)
+        end
+
```
jw81, line 469: Nice!  👍

modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
SlakrHakr: https://github.com/department-of-veterans-affairs/vets-api/pull/7805/commits/f4da4380e69443b9ae807c5d4c823a9ee4ca33c0

jw81 (approved):

PR #7810, opened 2021-09-01
Dev 28208 include clinic name

grossb1:
## Description of change

All appointment v2 endpoint responses have been supplemented with the clinic name (as `stationName`).  In the event that the request for the clinic name fails, the appointment will be returned without the `stationName` field.

## Original issue(s)

<!-- Please describe testing done to verify the changes or any testing planned. 
Tests were added/updated to reflect this change


modules/vaos/app/services/vaos/v2/appointments_service.rb
```diff
@@ -39,7 +39,7 @@ def post_appointment(request_object_body)
         end
       end
 
-      def update_appointment(appt_id:, status:)
+      def update_appointment(appt_id, status)
```
ajmagdub, line 42: curious why you removed the keyword parameters

modules/vaos/app/controllers/vaos/v2/appointments_controller.rb
```diff
@@ -46,21 +64,47 @@ def appointment
...
+            appt[:station_name] = cached_clinic_names[appt[:clinic]] if cached_clinic_names[appt[:clinic]]
+          end
+        end
       end
 
-      def cancellation_reason
-        params.require(:reason)
+      def get_clinic_name(location_id, clinic_id)
+        clinics = systems_service.get_facility_clinics(location_id: location_id, clinic_ids: clinic_id)
+        clinics.first[:station_name]
```
ajmagdub, line 93: so the assumption is that the only clinic that you're interested in will be the first one returned?

ajmagdub (approved):
approved with a couple of comments

modules/vaos/app/services/vaos/v2/appointments_service.rb
grossb1: For consistency atm.  I like the explicit field naming better, I'll get a ticket in to change it.

modules/vaos/app/controllers/vaos/v2/appointments_controller.rb
grossb1: That is correct, with this particular solution we will request a single clinic at a time so only a single clinic should be returned in the arry.

PR #7816, opened 2021-09-02
Lockbox db key rotation

LindseySaari:
## Description of change

After discovering the [lockbox gem](https://github.com/ankane/lockbox), it was decided to migrate away from attr_encrypted in favor of some of the Lockbox built in functionality for [rotating keys](https://github.com/ankane/lockbox#key-rotation). 

The steps in the [migration guide](https://github.com/ankane/lockbox#migrating-from-another-library) were followed and commited in this PR. Upon success of the migration (and passing tests) we can proceed to the next step in the migration guide to remove the attr_encrypted functionality. 

The devops settings PR is located [here](https://github.com/department-of-veterans-affairs/devops/pull/9882) (awaiting approval) and I've placed the new Lockbox master keys into parameter store. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#29612

- All of our encrypted attrs are strings, but from the Lockbox docs: "Note: Use a text column for the ciphertext in migrations, regardless of the type" 
- Lockbox does use an iv, but it's now stored in one field instead of two. See ["Cleaner Schema"](https://ankane.org/modern-encryption-rails) heading in the docs.


modules/covid_vaccine/app/models/covid_vaccine/v0/expanded_registration_submission.rb
```diff
@@ -49,6 +49,9 @@ class ExpandedRegistrationSubmission < ApplicationRecord
         reg.form_data&.symbolize_keys!
       end
 
+      serialize :form_data, :raw_form_data, JSON
```
stephenBarrs, line 52: Should `eligibility_info` also be serialized here?

stephenBarrs (approved):
LGTM

modules/covid_vaccine/app/models/covid_vaccine/v0/expanded_registration_submission.rb
LindseySaari: Good catch! YES! Thank you, just pushed up a change!

modules/covid_vaccine/app/models/covid_vaccine/v0/registration_submission.rb
```diff
@@ -13,6 +13,9 @@ class RegistrationSubmission < ApplicationRecord
 
       attr_encrypted :form_data, key: Settings.db_encryption_key, marshal: true, marshaler: JsonMarshal::Marshaller
       attr_encrypted :raw_form_data, key: Settings.db_encryption_key, marshal: true, marshaler: JsonMarshal::Marshaller
+
+      serialize :form_data, :raw_form_data, :eligibility_info, JSON
```
ericboehs, line 17: Have you tested the format of the JSON is the same?

app/models/async_transaction/base.rb
```diff
@@ -13,6 +15,8 @@ class Base < ApplicationRecord
     }
 
     attr_encrypted :metadata, key: Settings.db_encryption_key
+    serialize :metadata, JsonMarshal::Marshaller
```
LindseySaari, line 18: AsyncTransaction::Base record metadata is an array (verified in prod and staging. The metadata needs to be coerced into a string before saving into the db and retrieved as an array. The JsonMarshal::Marshaller class will properly save and retrieve in the expected form.

app/models/in_progress_form.rb
```diff
@@ -27,6 +29,8 @@ def cast(value)
 
   attribute :user_uuid, CleanUUID.new
   attr_encrypted :form_data, key: Settings.db_encryption_key
+  serialize :form_data, JsonMarshal::Marshaller
```
LindseySaari, line 32: InProgressForm record form_data is a Hash (verified in prod and staging). The form_data needs to be coerced into a string before saving into the db and retrieved as a Hash. The JsonMarshal::Marshaller class will properly save and retrieve in the expected form.

config/settings.yml
```diff
@@ -40,6 +40,10 @@ ssoe_eauth_cookie:
 session_cookie:
   secure: false
 
+lockbox:
+  master_key: "a798f118b400a1a7552597cc94f44c577919c76b3d26c1eed8ae0703f4240216"
+  #old_master_key: we will need this in the near future when rotating
```
LindseySaari, line 45: I plan to integrate the previous_key in a separate PR later on to avoid introducing multiple changes in a single PR

lib/json_marshal/marshaller.rb
```diff
@@ -7,7 +7,7 @@ def self.dump(obj)
     end
 
     def self.load(attribute)
-      JSON.parse(attribute)
+      JSON.parse(attribute) unless attribute.nil?
```
LindseySaari, line 10: if attribute = nil JSON.parse(nil) throws an error, causing issues when marshaling/loading encrypted attrs (hashes, arrays, etc)

For example, some AsyncTransaction::Base records have metadata as nil, throwing a`TypeError (no implicit conversion of nil into String)`

modules/covid_research/app/services/covid_research/volunteer/form_crypto_service.rb
```diff
@@ -3,28 +3,26 @@
 module CovidResearch
   module Volunteer
     class FormCryptoService
-      extend AttrEncrypted
```
LindseySaari, line 6: For the attr_encrypted approach to encrypted strings, you had to pass around the iv, but this is now not the case with lockbox. CovidResearch:: Volunteer:: FormCryptoService is not an ActiveRecord class so I followed the instructions [here](https://github.com/ankane/lockbox#strings) for string encryption

modules/covid_vaccine/app/models/covid_vaccine/v0/registration_submission.rb
LindseySaari: This has since been changed to use the JsonMarshal::Marshaller class which calls JSON.parse(some_attribute) when retrieving a field and calls .to_json when saving to the database to save as a string format

LindseySaari:
@ericboehs The migration to create the new lockbox columns was successfully run [here](http://jenkins.vfs.va.gov/job/deploys/job/vets-api-server-vagov-dev-post-deploy/104/ ) in the dev env. Next step is to merge/deploy or deploy [this change](https://github.com/department-of-veterans-affairs/vets-api/pull/7816) via git ref? to the dev env and run the rake task.

config/settings.yml
ericboehs: I'd recommend "previous_key" instead of old key (saw it in the lockbox docs).

app/models/async_transaction/base.rb
ericboehs, line 18: Do we need to add the `serialize` lines while we're migrating?

modules/covid_research/app/services/covid_research/volunteer/form_crypto_service.rb
```diff
@@ -3,28 +3,26 @@
...
-      attr_encrypted :submission, key: Settings.db_encryption_key
+      # attr_encrypted :submission, key: Settings.db_encryption_key
```
ericboehs, line 6: How does this get migrated?

modules/covid_research/lib/redis_format.rb
```diff
@@ -2,8 +2,6 @@
 
 module CovidResearch
   class RedisFormat
-    attr_reader :iv
```
ericboehs, line 5: Where was this `iv` being used? I'm only seeing the test changes in this PR. What was the iv used for here? Do we need to replicate it in lockbox?

rakelib/lockbox_migration.rake
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+namespace :lockbox do
+  desc 'connection testing'
```
ericboehs, line 4: connection testing? 🤔

modules/covid_research/app/services/covid_research/volunteer/form_crypto_service.rb
ericboehs: Looking at it again. This "submission" attribute is purely an attribute in Ruby and not in the db?

app/models/async_transaction/base.rb
LindseySaari: Yes because if new records are created during the migration process, lockbox will properly save the encrypted data into the lockbox field (as well as in the attr_encrypted field).

modules/covid_research/app/services/covid_research/volunteer/form_crypto_service.rb
LindseySaari: It doesn't necessarily get migrated. This class is used to encrypt job params when they're pushed to redis. We will have to be mindful of any jobs in the queue (which is doesn't seem that this job runs often) during the migration process.

rakelib/lockbox_migration.rake
LindseySaari: Oopsie copy/paste typo

modules/covid_research/lib/redis_format.rb
LindseySaari: No, lockbox doesn't use an iv. iv was passed to the encrypt/decrypt methods in the modules/covid_research/app/services/covid_research/volunteer/form_crypto_service.rb (see the changes made to that file in this PR)

modules/covid_research/app/services/covid_research/volunteer/form_crypto_service.rb
LindseySaari: So the submission attribute was just an attribute on a PORO purely used to encrypt job params when pushed to the queue in redis

LindseySaari:
@ericboehs If this looks ok, I think I would like to deploy via git ref to dev and run the rake task

modules/covid_research/lib/redis_format.rb
ericboehs: Reading the lockbox docs, Lockbox uses a unique key per field via CipherSweet.

LindseySaari: @ericboehs Related to persisting encrypted data in redis, the encryption logic is only intended to be used for form data when it's put into Redis (for Sidekiq), so it doesn't appear that we need to be concerned with additional encrypted data in Redis.

LindseySaari: Lockbox does use an iv, but it's now stored in one field instead of two. See ["Cleaner Schema"](https://ankane.org/modern-encryption-rails) heading in the docs.

ericboehs (dismissed):

LindseySaari:
I'll wait for a 👍  from a security perspective before I test/deploy via git ref

f1337 (commented):
For future reviewers:

Lockbox's AES GCM implementation uses OpenSSL: https://github.com/ankane/lockbox/blob/master/lib/lockbox/aes_gcm.rb. Lines 11 and 34 are currently:

```
cipher = OpenSSL::Cipher.new("aes-256-gcm")
```

That string in the constructor has a [specific format](https://ruby-doc.org/stdlib-2.4.0/libdoc/openssl/rdoc/OpenSSL/Cipher.html):

```
cipher = OpenSSL::Cipher.new('<name>-<key length>-<mode>')
```

Which means `cipher = OpenSSL::Cipher.new("aes-256-gcm")` => 256-bit AES-GCM


They also provide a XSalsa20 implementation via Libsodium. I'm satisfied with both implementations in a general sense. This is not a blanket endorsement of all future states of Lockbox, just a record of where things stand today.

f1337 (approved):
Nice! 
Will Ferrell dancing while playing cowbell as others sing and play instruments.

ericboehs:
Some more info:
- Lockbox uses per-field derivative keys via CipherSweet. (The master key can't be derived from brute forcing a single column of data.)
- IVs are stored with the encrypted data so the schema only uses a single ciphertext field.
- We are not auditing changes to any of these fields. This is out of scope for our objective here (rotate db keys).
- Lockbox is using the same cipher that attr_encrypted and Rails 7 uses (aes 256 gcm).
- Lockbox allows for easy key rotation.
  - It also supports using a Key Management Service (such as AWS KMS) (which we're interested in using in the future).
- We are using 16 byte padding to guard against data leakage.
- Encryption is using openssl via Ruby's stdlib `OpenSSL::Cipher`.
- Lockbox can use hybrid cryptography allowing for a write key which can't read data (and vice versa).

bramleyjl (approved):

stephenBarrs (approved):

ericboehs (approved):

PR #7824, opened 2021-09-02
[API-9701] - ClaimsApi - Make Form526 'applicationExpirationDate' optional

jw81:
## Description of change

Makes `applicationExpirationDate` optional in a Form526 submission.  If `applicationExpirationDate` is not provided, we will set `applicationExpirationDate` to a date string that is 1 year in the future from the current date.

## Original issue(s)
[API-9701](https://vajira.max.gov/browse/API-9701)


modules/claims_api/app/models/claims_api/auto_established_claim.rb
```diff
@@ -53,6 +53,7 @@ class AutoEstablishedClaim < ApplicationRecord
     alias token id
 
     def to_internal
+      form_data['applicationExpirationDate'] ||= build_application_expiration
```
SlakrHakr, line 56: should we be throwing away what the consumer submitted? (if they submitted of course)
im torn because it sounds like it doesnt really matter but also it seems wrong to toss aside something the consumer submitted to us 🤷🏻

SlakrHakr (dismissed):

SlakrHakr (approved):

SlakrHakr (approved):

PR #7825, opened 2021-09-02
29577 logingov UUID validations accounts table

bosawt:
## Description of change
This PR adds validations for the logingov_uuid column to the Account model, in addition to adding capability in User and UserIdentity model to properly store the logingov_uuid and to match an Accounts record based off of the logingov_uuid

## Original issue(s)
department-of-veterans-affairs/va.gov-team/issues/29577

- A user won't be able to currently sign in with a logingov_uuid on the user model, so to test, you'll have to set a breakpoint and hack it in. I recommend `app/models/user_session_form.rb`, line 25, taking the `normalized_attributes` from the `saml_response` and injecting a `logingov_uuid` before continuing
- I might have found an existing bug when looking at this. I'm not sure if we have more than a single record with a `sec_id` and null `idme_uuid`, which seems egregious and would have been around for a long time


bramleyjl (approved):
LGTM

rileyanderson (approved):

PR #7835, opened 2021-09-03
[7834] Relaxing uniqueness constraint on idme_uuid for Accounts table…

bosawt:
…, so that nil records are possible when sec_id is defined

## Description of change
This PR relaxes the idme_uuid constraint on the `Account` model so that multiple records can be nil, given that a `sec_id` is defined. Previously, there was a bug so that only a single `accounts` record could have a `nil` `idme_uuid`

## Original issue(s)
department-of-veterans-affairs/vets-api/issues/7834

- To test this, in rails console I ran:
`Account.new(sec_id: 1).save`
`Account.new(sec_id: 2).save`
And confirmed that both were properly saved in the accounts table
Also, `Account.new.save` still failed, as expected, proving that a record still needs with a `sec_id` or `idme_uuid` defined


annaswims:
I think there's a bit more to this one... I think we should have some sort of unique constraint that there can't be more than one record with a nil `idme_uuid` and a specific `sec_id`

bramleyjl:
> I think there's a bit more to this one... I think we should have some sort of unique constraint that there can't be more than one record with a nil `idme_uuid` and a specific `sec_id`
> 
> ```ruby
> Account.new(sec_id: 1).save!
> Account.new(sec_id: 1).save!  
> ```
> 
> shouldn't be allowed to create two records

I agree, shouldn't we add a reflecting `validates :idme_uuid
, uniqueness: true, allow_nil: true`

annaswims:
no, a veteran could have multiple idme uuid's, like if they'd created a ds logon  and id.me account and identity-verified in a certain order. 

how about sec_id's must be unique if idme_uuid is blank`validates :sec_id, uniqueness: true, if: -> { idme_uuid.blank? }`

rileyanderson (approved):

PR #7836, opened 2021-09-03
rough draft of contestable issues swagger block

yakschuss:
[API-9408](https://vajira.max.gov/browse/API-9408)

This PR is a rough draft of contestable-issues v2, standalone for all decision_review types.

We were originally planning on pushing all new endpoints into RSWAG, but do to OAS requirements and deadlines, we're pushing this up now and will rewrite in rswag time permitting.

### DIFF
--- a/modules/appeals_api/app/swagger/appeals_api/v2/responses_contestable_issues.json
+++ b/modules/appeals_api/app/swagger/appeals_api/v2/responses_contestable_issues.json
@@ -1,0 +1,151 @@
+{"200": {
+    "description": "JSON:API response returning all contestable issues for a specific veteran.",
+    "content": {
+      "application/vnd.api+json": {
+        "schema": {
+          "type": "object",
+          "properties": {
+            "data": {
+              "type": "array",
+              "items": {
+                "type": "object",
+                "description": "A contestable issue (to contest this, you include it as a RequestIssue when creating a HigherLevelReview, SupplementalClaim, or Appeal)",
+                "properties": {
+                  "type": { "type": "string", "enum": [ "contestableIssue" ] },
+                  "id": { "type": "string", "nullable": true },
+                  "attributes": {
+                    "type": "object",
+                    "properties": {
+                      "ratingIssueReferenceId": {
+                        "type": "string",
+                        "nullable": true,
+                        "description": "RatingIssue ID",
+                        "example": "2385"
+                      },
+                      "ratingIssueProfileDate": {
+                        "type": "string",
+                        "nullable": true,
+                        "format": "date",
+                        "description": "(yyyy-mm-dd) RatingIssue profile date",
+                        "example": "2006-05-31"
+                      },
+                      "ratingIssueDiagnosticCode": {
+                        "type": "string",
+                        "nullable": true,
+                        "description": "RatingIssue diagnostic code",
+                        "example": "5005"
+                      },
+                      "ratingDecisionReferenceId": {
+                        "type": "string",
+                        "nullable": true,
+                        "description": "The BGS ID for the contested rating decision. This may be populated while ratingIssueReferenceId is nil",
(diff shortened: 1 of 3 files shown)


modules/appeals_api/app/swagger/appeals_api/v2/responses_contestable_issues.json
```diff
@@ -0,0 +1,151 @@
...
+                        "type": "string",
+                        "nullable": true,
+                        "description": "RatingIssue diagnostic code",
+                        "example": "5005"
+                      },
+                      "ratingDecisionReferenceId": {
+                        "type": "string",
+                        "nullable": true,
+                        "description": "The BGS ID for the contested rating decision. This may be populated while ratingIssueReferenceId is nil",
+                        "example": null
```
cilestin, line 42: If there isn't an example, do you have to supply the key?

modules/appeals_api/app/swagger/appeals_api/v2/responses_contestable_issues.json
```diff
@@ -0,0 +1,151 @@
...
+              "errors": [{ "status": 422, "code": "invalid_receipt_date", "title": "Invalid Receipt Date", "detail": "\"2019-02-18\" is before AMA Activation Date (2019-02-19)." }]}},
+          "future receipt date": {
+            "value": {
+              "errors": [{ "status": 422, "code": "invalid_receipt_date", "title": "Invalid Receipt Date", "detail": "\"2020-07-02\" is in the future (today: 2020-07-01; time zone: (GMT-05:00) America/New_York)." }]}},
+          "unparsable date": {
+            "value": {
+              "errors": [{ "status": 422, "code": "invalid_receipt_date", "title": "Invalid Receipt Date", "detail": "\"Widdershins\" is not a valid date." }]}},
+          "invalid veteran ssn": {
+            "value": {
+              "errors": [{ "status": 422, "code": "invalid_veteran_ssn", "title": "Invalid Veteran SSN", "detail": "SSN regex: /^\\d{9}$/)." }]}}}}}},
```
cilestin, line 140: I don't believe this is how this error will turn out, as the DetailedSchemaError should catch this as a common schema pattern mismatch error. Is there a different validator we've written that results in this special SSN-specific code & detail?

modules/appeals_api/app/swagger/appeals_api/v2/responses_contestable_issues.json
```diff
@@ -0,0 +1,151 @@
...
+        "schema": {
+          "type": "object",
+          "properties": { "errors": { "$ref": "#/components/schemas/errorWithTitleAndDetail" } }},
+        "examples": {
+          "before AMA": {
+            "value": {
+              "errors": [{ "status": 422, "code": "invalid_receipt_date", "title": "Invalid Receipt Date", "detail": "\"2019-02-18\" is before AMA Activation Date (2019-02-19)." }]}},
+          "future receipt date": {
+            "value": {
+              "errors": [{ "status": 422, "code": "invalid_receipt_date", "title": "Invalid Receipt Date", "detail": "\"2020-07-02\" is in the future (today: 2020-07-01; time zone: (GMT-05:00) America/New_York)." }]}},
```
cilestin, line 134: Would the "today" portion be adjusted based on the request's timezone? Is that even something we want to support? I'm not sure how useful reporting the current system time is (whether adjusted for the timezone or not), but not saying it _isn't_ useful either.

Also, remember that JSON:API ErrorObject allows for a `meta` key, so an alternative that may be more programmatically useful to the consumers might look like:
```
{...
  detail: "Receipt date is in the future", 
  meta: { 
    given_date: "2020-07-02", 
    current_date: "2020-07-17", 
    timezone: { 
      offset: "GMT -5:00", 
      name: "America/New_York" } 
}
```

PR #7856, opened 2021-09-08
[VRE 👍 ] 29552 Send comm preferences in VRE summary

RudyOnRails:
## Description of change

This PR does a few things:
2. adjusts mailer to include two new columns, with test
3. removes brittle fixture from mailer spec and replaces with standard `include` matcher
4. add preview mailer

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/29552


lib/boolean_helper.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module BooleanHelper
+  def boolean_to_yesno(boo)
+    boo ? 'Yes' : 'No'
+  end
```
ericboehs, line 6: This seems a bit overkill for just two variables.

```
boolean_to_yesno(boo)
```
is longer than 
```
boo ? 'Yes' : 'No'
```
and adds a layer of abstraction so you have to go look up what it does.

I understand avoiding writing logic in a view but conditionals of this simplicity seem okay.

Could you explain more why you added this help? I think I'm missing something.

RudyOnRails: My unit test on the helper did catch a failure (I expected "Yes" but was actually returning "yes" at first). Wouldn't it be more work to write a view spec? Would I have to revert to using a fixture of markup?

I am totally open to using your suggestion, but am curious about the testing of it.

RudyOnRails: Good morning sir. If you get a second can you share your thoughts on this ☝️.

RudyOnRails: Ok this has been resolved. Helper and spec removed. Logic moved to email view.

app/mailers/ch31_submissions_report_mailer.rb
```diff
@@ -3,6 +3,8 @@
 require 'reports/uploader'
 
 class Ch31SubmissionsReportMailer < ApplicationMailer
+  include BooleanHelper
```
ericboehs, line 6: This doesn't need to be here anymore, right?

RudyOnRails: Yeah, CI just caught it, and it's gone.

ericboehs (approved):

PR #7860, opened 2021-09-08
Fix mobile appointment id in missing facility logging

kreek:
## Description of change
My additional logging for the missing facility error itself has a bug in it 😐 it now it correctly logs the `appointment.id`

## Original issue(s)
N/A


jperk51 (approved):
LGTM

PR #7867, opened 2021-09-09
[VRE 👍  ] 29025 adjust ch31 solo mailer for cmp case

RudyOnRails:
**This is a repeat of https://github.com/department-of-veterans-affairs/vets-api/pull/7851 - I had to revert it b/c the stakeholder wanted to see it before production. I am prepping this again but won't merge until they 👀 .**

**Details copied from 7851.**

## Description of change

This PR makes it so that a `SavedClaim::VeteranReadinessEmploymentClaim` is initialized with a new ivar of `@sent_to_cmp = false` and if the claim ends up routing to CMP, then that will be indicated by displaying "(routed to CMP)" in the body of the email.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/29025

- model spec
- mailer spec


ericboehs (dismissed):

RudyOnRails:
Sorry @ericboehs, I'll try to avoid wasting your time in the future, but the stakeholder had a change so I had to tweak this a wee bit. Sorry for the double review :(

app/mailers/veteran_readiness_employment_mailer.rb
```diff
@@ -1,13 +1,12 @@
...
-  def build(user, email_addr)
+  def build(user, email_addr, routed_to_cmp)
     email_addr = 'kcrawford@governmentcio.com' if FeatureFlipper.staging_email?
-
     @submission_date = Time.current.in_time_zone('America/New_York').strftime('%m/%d/%Y')
-    @pid = user.participant_id || 'Sent to CM'
-
-    template = File.read('app/mailers/views/veteran_readiness_employment.html.erb')
+    @pid = user.participant_id.to_s
+    cmp = routed_to_cmp ? '_cmp' : ''
```
ericboehs, line 8:
```suggestion
    cmp = '_cmp' if routed_to_cmp
```

app/models/saved_claim/veteran_readiness_employment_claim.rb
```diff
@@ -74,6 +76,11 @@ class SavedClaim::VeteranReadinessEmploymentClaim < SavedClaim
 
   validate :veteran_information, on: :prepare_form_data
 
+  def initialize(args)
+    @sent_to_cmp = false
```
ericboehs, line 80:
```suggestion
    self.sent_to_cmp = false
```

app/models/saved_claim/veteran_readiness_employment_claim.rb
```diff
@@ -134,7 +141,7 @@ def send_to_vre(user)
       )
     end
 
-    VeteranReadinessEmploymentMailer.build(user, email_addr).deliver_now if user.present?
+    VeteranReadinessEmploymentMailer.build(user, email_addr, @sent_to_cmp).deliver_now if user.present?
```
ericboehs, line 144:
```suggestion
    VeteranReadinessEmploymentMailer.build(user, email_addr, sent_to_cmp).deliver_now if user.present?
```

app/models/saved_claim/veteran_readiness_employment_claim.rb
```diff
@@ -172,6 +179,7 @@ def send_to_central_mail!
     update(form: form_copy.to_json)
 
     log_message_to_sentry(guid, :warn, { attachment_id: guid }, { team: 'vfs-ebenefits' })
+    @sent_to_cmp = true
```
ericboehs, line 182:
```suggestion
    self.sent_to_cmp = true
```

ericboehs (commented):
Small change: don't use instance variables when defining getter/setters methods.

app/models/saved_claim/veteran_readiness_employment_claim.rb
RudyOnRails: I never needed this to be public in the first place, so I stuck with the ivar.

app/models/saved_claim/veteran_readiness_employment_claim.rb
RudyOnRails: I never needed this to be public in the first place, so I stuck with the ivar.

app/mailers/veteran_readiness_employment_mailer.rb
```diff
@@ -1,13 +1,12 @@
...
 class VeteranReadinessEmploymentMailer < ApplicationMailer
-  def build(user, email_addr)
+  def build(user, email_addr, routed_to_cmp)
     email_addr = 'kcrawford@governmentcio.com' if FeatureFlipper.staging_email?
-
     @submission_date = Time.current.in_time_zone('America/New_York').strftime('%m/%d/%Y')
-    @pid = user.participant_id || 'Sent to CM'
-
-    template = File.read('app/mailers/views/veteran_readiness_employment.html.erb')
+    @pid = user.participant_id.to_s
```
ericboehs, line 7: Is there a reason this needs to be to_s? The mailer is using erb which should cast int to str.

app/models/saved_claim/veteran_readiness_employment_claim.rb
ericboehs: I try to pass private variables and use getters/setters if they can be public, avoiding instance variables. local variables are easy to refactor to public getters.

But this is fine. :+1:

spec/models/saved_claim/veteran_readiness_employment_claim_spec.rb
```diff
@@ -146,8 +146,10 @@
       let(:user_object) { create(:unauthorized_evss_user) }
 
       it 'PDF is sent to Central Mail and not VBMS' do
-        expect(claim).to receive(:send_to_central_mail!).once
+        expect(claim).to receive(:send_to_central_mail!).once.and_call_original
         expect(claim).not_to receive(:upload_to_vbms)
+        expect(VeteranReadinessEmploymentMailer).to receive(:build).with(user_object, 'VRE.VBAPIT@va.gov',
+                                                                         true).and_call_original
```
ericboehs, line 152:
```suggestion
        expect(VeteranReadinessEmploymentMailer).to receive(:build).with(
          user_object, 'VRE.VBAPIT@va.gov', true
        ).and_call_original
```

Optional suggestion to make it look better. ^

ericboehs (approved):

app/mailers/veteran_readiness_employment_mailer.rb
RudyOnRails: Nope. Can't remember exactly at this moment, but I think I was doing some string interpolation in a private method somewhere in a previous revision before I created another template, and rubocop suggested `.to_s` instead. 

Good catch 🎣

rileyanderson (approved):

PR #7874, opened 2021-09-10
29451 handle bdn downtime

RudyOnRails:
## Description of change

1. adds sidekiq middleware to check if dependent services are up or down, and either yields to run job, or schedules job for later when downtime has ended. Includes specs.
2. adds new DownTimeChecker class to check if current time is within a predefined downtime window; if it isn't return false; if it is, return the duration of time until the window ends. Includes specs.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/29451


lib/down_time_checker.rb
```diff
@@ -0,0 +1,29 @@
...
+    ['Friday at 2am UTC', 'Friday at 10am UTC'],
+    ['Saturday at 2am UTC', 'Saturday at 10am UTC']
+  ].freeze
+
+  def initialize(service)
+    @service = service
+  end
+
+  def down?
+    if DOWNTIME_WINDOWS.any? { |w| (parse_time(w[0])..parse_time(w[1])).cover?(Time.now.utc) }
```
ericboehs, line 17: Could you refactor this to be clearer?

Maybe extract this line to:
```suggestion
    if within_downtime_window?
```

```ruby
def within_downtime_window?
 DOWNTIME_WINDOWS.any? do |window|
    window = downtime_window window
    window.cover? Time.now.utc
  end
end

def downtime_window(begin_end_time)
  begin_time = Time.zone.parse begin_end_time.first
  end_time = Time.zone.parse begin_end_time.last

  begin_time..end_time
end
```

lib/middleware/sidekiq/only_run_while_up.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Middleware
+  module Sidekiq
+    require 'down_time_checker'
+    class OnlyRunWhileUp
+      def call(worker, job, _queue)
+        if worker.respond_to?(:downtime_checks)
+          array = worker.downtime_checks.map { |service| DownTimeChecker.new(service).down? }
+          if array.all? { |element| element == false }
```
ericboehs, line 10: Can you use `.none?` here instead of this any check.

lib/middleware/sidekiq/only_run_while_up.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Middleware
+  module Sidekiq
+    require 'down_time_checker'
+    class OnlyRunWhileUp
+      def call(worker, job, _queue)
+        if worker.respond_to?(:downtime_checks)
```
ericboehs, line 8: Can you refactor this `call` code to something succinct?

Example:
```ruby
if should_do_downtime_checks? && services_down?
  reschedule_job
else
  yield
end
```

You might have to pass some vars there.

lib/middleware/sidekiq/only_run_while_up.rb
```diff
@@ -0,0 +1,21 @@
...
+          array = worker.downtime_checks.map { |service| DownTimeChecker.new(service).down? }
+          if array.all? { |element| element == false }
+            yield
+          else
+            worker.class.perform_in(array.select(&:itself).max, job['args'])
```
ericboehs, line 13: I'm not sure I understand why `select(&:itself)` is necessary here. Could you explain please?

RudyOnRails: I sorta began to build it for future use to expand to more than just checking one service.

in that scenario, you'd possibly have `[false, 600, 1800]`. This selects all the truthy's.

ericboehs (approved):
Approving with the caveat we'll refactor in another PR.

lib/down_time_checker.rb
RudyOnRails: done

lib/middleware/sidekiq/only_run_while_up.rb
RudyOnRails: ✅

RudyOnRails: I know this PR is closed, but wanted to associate my comment.

Is this what you were thinking? `@downtime_checks_array.none? { |element| !element }`

That's what I went with in the new PR. You'll probably see it soon!

PR #7875, opened 2021-09-10
updated the query parameters on the get_patient_appointment_meta endp…

grossb1:
…oint to be camel case

The work here is to fix a small bug related to the casing of query parameters sent to the new VAOS service.

## Original issue(s)

VCR Cassettes were updated to reflect this change


ajmagdub (approved):
LGTM

PR #7893, opened 2021-09-13
Unlock and Update Flipper

ericboehs:


Gemfile.lock
```diff
@@ -846,12 +844,6 @@ GEM
       connection_pool (>= 2.2.2)
       rack (~> 2.0)
       redis (>= 4.2.0)
-    sidekiq-ent (2.2.3)
-      einhorn (>= 0.7.4)
-      sidekiq (>= 6.2.0)
```
LindseySaari, line 851: Uh oh! 😬 Looks like sidekiq & related gems were removed

ericboehs: NOOOOOO

LindseySaari: 🚨

LindseySaari (dismissed):
Looks good! Did you test flipper out locally? I know we use some custom views etc and just want to be sure nothing strange happens there with the minor version bump.

ericboehs:
> Looks good! Did you test flipper out locally? I know we use some custom views etc and just want to be sure nothing strange happens there with the minor version bump.

Ahhhhh. So I checked that UI at least loaded. And it did. But I checked again and it fails to load the styles. In fact, it loads fine the first load, but after a refresh it doesn't. AND Flipper broke it in a PATCH release. How dare they!

Anyways, I didn't dive into why it's not working right now but I'm 90%+ certain it's the CSP PR which I referenced in the Gemfile.

LindseySaari (approved):
💯

PR #7895, opened 2021-09-13
API-6342 - Update PII removal logic

nathanbwright:
## Description of change
Updates to our PII removal job allow for NOD (evidence_submission lane only) to hold onto it's PII longer (to allow for better evidence submission upload experience).

## Original issue(s)
https://vajira.max.gov/browse/API-6342


modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
```diff
@@ -5,6 +5,20 @@
 
 module AppealsApi
   class NoticeOfDisagreement < ApplicationRecord
+    PII_POLICY = lambda {
+      complete_status = COMPLETE_STATUSES.map { |status| "'#{status}'" }.join(', ')
+
+      <<~SQL.squish
```
yakschuss, line 11: Is this not possible to write in ActiveRecord?

nathanbwright: I suppose that we could remove all the logic from `AppealsApi::RemovePii#records_to_be_expunged` and delegate everything to the model and that would allow us to use standard AR queries. I wrote this as a way to keep the encrypted data (header and form) queries in `AppealsApi::RemovePii#records_to_be_expunged` and have the appeal specific rules inside the actual models...

Do you think that would be preferable, or did you have something else in mind?

ElleeB: I like this structure, but would love to know of other ideas and/or why ActiveRecord might be preferable.

nathanbwright: I just pushed up a commit that uses AR queries instead of raw sql for us to consider... Is that more of what you had it mind @yakschuss? I like to use AR queries when possible but I'm not sure I prefer it in the NOD's case...

modules/appeals_api/app/models/appeals_api/higher_level_review.rb
```diff
@@ -7,6 +7,10 @@ module AppealsApi
   class HigherLevelReview < ApplicationRecord
     include HlrStatus
 
+    scope :pii_expunge_policy, lambda {
+      where('updated_at < ? AND status IN (?)', 7.days.ago, COMPLETE_STATUSES)
```
yakschuss, line 11: I definitely prefer this iteration, I think the main idea is leveraging the tools available, and promoting readability. What about the NOD version do you dislike? I think it's a pretty readable subquery now.

modules/appeals_api/app/services/appeals_api/remove_pii.rb
```diff
@@ -40,19 +40,13 @@ def valid_form_type?
...
                  .or(
                    form_type.where.not(
                      encrypted_auth_headers: nil
                    )
-                 ).where(
-                   status: form_type::COMPLETE_STATUSES
-                 ).where(
-                   'updated_at < ?', 1.week.ago
-                 )
+                 ).pii_expunge_policy
```
yakschuss, line 49: 👍  I like this

modules/appeals_api/spec/services/appeals_api/remove_pii_spec.rb
```diff
@@ -11,23 +11,63 @@ module AppealsApi
...
+        week_old_has_pii = create :higher_level_review, :completed_a_week_ago
 
-        expect(has_pii.encrypted_form_data).not_to be_nil
+        expect(day_old_has_pii.encrypted_form_data).not_to be_nil
+        expect(week_old_has_pii.encrypted_form_data).not_to be_nil
 
         RemovePii.new(form_type: HigherLevelReview).run!
 
-        expect(has_pii.reload.encrypted_form_data).to be_nil
+        expect(day_old_has_pii.reload.encrypted_form_data).not_to be_nil
```
yakschuss, line 22: Awesome!

yakschuss (dismissed):
🚀

yakschuss (approved):
Looks good to me! Good catch on the reordering.

PR #7897, opened 2021-09-13
RuboCop: Disable pending cops

ericboehs:
RuboCop moved to a "pending on release" model for releasing new Cops. This allows for feature/patch updates without breaking CI/linting. We initially enabled all pending cops. I believe this is in error as not all pending cops will be enabled by default when RuboCop releases a new major version (e.g. 2.0). See [Versioning](https://docs.rubocop.org/rubocop/versioning).

Instead, we should create a section in our `.rubocop.yml` for pending cops and explicitly enable the cops we feel strongly about. We should put it under a "Pending until 2.0" heading so they can be removed easily (and not get mixed with the other cops).

Out of scope for this PR: we should look into cops disabled by default. On the versioning page, RuboCop's doc say that some cops are always disabled but could be useful to run periodically. We should see if any would be useful for us.


LindseySaari (approved):
👍 I think this makes sense. Should we throw a ticket on the board to review pending cops and enable as necessary?

ericboehs:
Sounds good. Added. https://github.com/department-of-veterans-affairs/va.gov-team/issues/29888

PR #7900, opened 2021-09-13
fall back to MPI for blank user attributes

annaswims:
## Description of change
There are cases where a user.gender returns nil even though the vet has a gender in MPI.  I don't know for sure, but I suspect that sometimes the identity provider might return empty string rather than nil. So, fall back to MPI when the attribute is blank.

I suspect the same thing could happen for name so I made the change there too
## Original issue(s)
department-of-veterans-affairs/va.gov-team#29859


spec/support/vcr_cassettes/mpi/find_candidate/orch_search_with_attributes.yml
```diff
@@ -1,93 +1,12 @@
 ---
 http_interactions:
 - request:
-    method: POST
-    uri: http://www.example.com/
+    method: post
+    uri: "<MPI_URL>"
     body:
       encoding: ASCII-8BIT
-      base64_string:
```
annaswims, line 8: The actual diff is that blank middle name is no longer adding a second `<given>` node
Screen Shot 2021-09-14 at 10 03 12 PM

rileyanderson (approved):

lib/evss/disability_compensation_auth_headers.rb
```diff
@@ -37,7 +37,7 @@ def gender
       when 'M'
         'MALE'
       else
-        Raven.extra_context(user_gender: @user.gender)
+        Raven.extra_context(user_gend: @user.gender)
```
bosawt, line 40: Was this intentional? Just making sure this wasn't a typo, otherwise I'll approve

bramleyjl: It's the same header as in the spec that uses it, so I'm pretty sure it's intentional.

bramleyjl (approved):

lib/evss/disability_compensation_auth_headers.rb
annaswims: yeah, maybe it's a bit sketchy of me, but I'm trying to get around the PII sanitizer.  I figure if we're logging this string and nothing else it's not a violation of anyone's privacy.

PR #7901, opened 2021-09-13
require gender for form 526

annaswims:
## Description of change
require gender to be present to begin form 526 (evss requirement)

## Original issue(s)
department-of-veterans-affairs/va.gov-team#29859


annaswims:
replaced by #7941

PR #7902, opened 2021-09-13
Fix failing time dependent test

ajmagdub:
Added timecop to jwt_wrapper_spec.rb so that the timing of the test will not cause the test to fail.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#29872

Added delay in test to get the test to fail then introduced timecop and verified that the test passed.


ericboehs (approved):

ericboehs:
Thanks AJ!!

grossb1 (commented):
LGTM

PR #7905, opened 2021-09-13
Revert "Update Sidekiq"

annaswims:
Reverts department-of-veterans-affairs/vets-api#7890


annaswims:
my bad. Not the problem

PR #7911, opened 2021-09-14
API-9730: update legacy appeals docs to Rswag

ElleeB:
_Since the Draft PR, I've modified the 422 error tests in order to have both examples show up in the docs._
---

Updates the legacy appeals dev portal docs to use rswag to generate.

Ticket: https://vajira.max.gov/browse/API-9730


modules/appeals_api/spec/docs/v2/legacy_appeals_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+            }
+          }
+        end
+
+        it 'returns a 422 response' do |example|
+          assert_response_matches_metadata(example.metadata)
+        end
+      end
+
+      # response '422', 'When X-VA-SSN and X-VA-File-Number are missing' do
```
yakschuss, line 102: Is the plan to flesh this out further or to remove it?

ElleeB: I left a comment about this in the engineering chat.

modules/appeals_api/spec/docs/v2/legacy_appeals_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+      security [{ apikey: [] }]
+      consumes 'application/json'
+      produces 'application/json'
+      description = 'Returns eligible Legacy Appeals for a Veteran. A Legacy Appeal is eligible if an SOC/SSOC ' \
+      'has been declared, and if the date of declaration is within the last 60 days. ' \
+      'SOC (Statement of Case) / SSOC (Supplemental Statement of Case)'
+      description description
+
+      parameter AppealsApi::SwaggerSharedComponents.header_params[:veteran_ssn_header]
+      parameter AppealsApi::SwaggerSharedComponents.header_params[:veteran_file_number_header]
```
nathanbwright, line 26: I think you'll want to update the description on their params...

Maybe something like this:
```
parameter AppealsApi::SwaggerSharedComponents.header_params[:veteran_ssn_header].merge({ required: false, description: 'Either X-VA-SSN or X-VA-File-Number is required' })
```
https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/appeals_api/spec/docs/v2/hlr_spec.rb#L212

modules/appeals_api/spec/docs/v2/legacy_appeals_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+
+      response '200', 'Returns eligible legacy appeals for a veteran' do
+        let(:'X-VA-SSN') { '123456789' }
+
+        schema AppealsApi::SwaggerSharedComponents.response_schemas[:legacy_appeals_response_schema]
```
nathanbwright, line 31: I'm not seeing this schema show up on the front end for some reason (it's showing the `Example Value` link but no `Schema` link)...

Maybe you could move the schema to the Rswag config and then reference it by the `ref`? Something like this: https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/appeals_api/spec/docs/v2/hlr_spec.rb#L219

nathanbwright (commented):
Looking really good! I left a few inline comments.

modules/appeals_api/spec/docs/v2/legacy_appeals_spec.rb
ElleeB: Oh! I see now :) Thanks

modules/appeals_api/spec/docs/v2/legacy_appeals_spec.rb
ElleeB: I see now!! Thank you!

modules/appeals_api/spec/support/schemas/legacy_appeal.json
```diff
@@ -0,0 +1,46 @@
...
+      "decisionDate": {
+        "type": "string",
+        "description": "Date of the Appeal's original decision.",
+        "example": "2018-09-28T00:00:00.000Z"
+      },
+      "latestSocSsocDate": {
+        "type": "string",
+        "description": "Date of the Appeal's most recent SOC/SSOC (Statement of Case/Supplemental Statement of Case)",
+        "example": "2018-12-29T00:00:00.000Z"
+      }
```
nathanbwright, line 43: I think these keys should be moved 'up' to be inside the `attributes->properties`... Right now they aren't showing up in the schema like they should.

yakschuss (dismissed):
Looks good to me!

nathanbwright (commented):
Looking good! The schema for the `200` response still needs some additional updates (left an inline comment but I also think the uuid vs vacols id may still need to be updated, I'm just comparing what the swagger blocks is outputting vs what we have for rswag)...

Are you planning to do another PR for the 5xx responses?

modules/appeals_api/spec/support/schemas/legacy_appeal.json
ElleeB: Oy. Yes. Thanks for catching that

cilestin (approved):

PR #7913, opened 2021-09-14
Mobile user table

jperk51:
## Description of change
Created table to track mobile users via icn for later use by our push notifications service. Also added a lock to protect the table. 

## Original issue(s)

Added new tests to mobile application_controller.rb as well as created new spec for new user table job.


kreek (changes requested):
The job part looks pretty good, some changes to the database schema.

kreek (approved):
LGTM 🚀

PR #7914, opened 2021-09-14
mobile user table db changes

jperk51:
## Description of change
db changes for creating a mobile users table

## Original issue(s)

db only PR


modules/mobile/db/migrate/20210913194827_create_mobile_users.rb
```diff
@@ -0,0 +1,9 @@
+class CreateMobileUsers < ActiveRecord::Migration[6.1]
+  def change
+    create_table :mobile_users do |t|
+      t.string :user_id, null: false
```
kreek, line 4: For the initial use case of SM notification filtering, ICN will be the id. Which I think this is a holder for :) but it could be good to make it explicitly that, as there will be other user ids in the future.

modules/mobile/db/migrate/20210913194827_create_mobile_users.rb
```diff
@@ -0,0 +1,9 @@
...
+      t.datetime :created_at, null: false
```
kreek, line 5: Could this use the built in Rails timestamps?
```suggestion
      t.timestamps null: false
```

modules/mobile/db/migrate/20210913194827_create_mobile_users.rb
jperk51: yeah, I used user_id since I didn't know what the other customers would be using and I was thinking we would still want them to be the same field from our perspective. Do you think them being the same field could cause issues in the future?

modules/mobile/db/migrate/20210913194827_create_mobile_users.rb
jperk51: I used the created_at by itself since I didn't think updated_at would be relevant for us and timestamps creates both, but yeah I could definitely change it to use timestamps if we'd rather do that

modules/mobile/db/migrate/20210913194827_create_mobile_users.rb
```diff
@@ -0,0 +1,9 @@
+class CreateMobileUsers < ActiveRecord::Migration[6.1]
+  def change
+    create_table :mobile_users do |t|
+      t.string :icn, null: false
+      t.timestamps null: false
+      t.index %i[:icn], name: "mobile_user_index", unique: true
```
kreek, line 6: I think you can leave the custom name part off and one using Rails conventions will be generated.

kreek (approved):
One minor comment but otherwise LGTM 🚀

modules/mobile/db/migrate/20210913194827_create_mobile_users.rb
jperk51: gotcha

thilton-oddball (approved):
LGTM

PR #7916, opened 2021-09-14
[29949] updates TUD accounts from User object on login & logout

bramleyjl:
## Description of change

This change updates TUD accounts by pulling attributes from its corresponding User object on login and logout. The update happens as part of a Sidekiq job on login and should not impact performance; on logout it must happen as part of the normal logout workflow because the User object is destroyed at the end of it and does not persist to be used - at present there is no `AfterLogoutJob` logic except what TUD is currently using - this PR will remove it entirely. The `services` update only pulls from authorized services that a user has, auth_free services are not included.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#29949

Unit tests updated, repeated login/logouts to test proper updating of TUD account values.


jscanish (dismissed):
I noted a few minor style things you can feel free to change or ignore, but otherwise it looks good.

rileyanderson (approved):

jscanish (approved):

PR #7920, opened 2021-09-15
1547 mobile appointments phone only flag

kreek:
## Description of change
Adds a `phoneOnly` boolean flag for phone based mobile appointments. Also splist the appointments request spec into two specs, one for the list endpoint and another for the cancel endpoint.

## Original issue(s)
department-of-veterans-affairs/va-mobile-app/issues/1547


jperk51 (approved):
LGTM

PR #7929, opened 2021-09-16
27858 clear session jwt

dillo:
## Description of change

- Clear the jwt key from the session upon successful check-in
- Set the jwt key in the session only if authentication is successful 
## Original issue(s)

- [x] RSpec passing


mdewey (approved):
LGTM

PR #7933, opened 2021-09-17
Adds Mock Json response from external service claimants information endpoint at v0/eli…

bradbergeron-us:
…gibility endpoint

## Description of change

This PR adds a mock endpoint for an external service endpoint. The response generated from the controller method simulates what would be coming back from an http request in the service that has not been implemented yet. The goal is to be able to have an endpoint that simulates a http request body in json format for the frontend application to consume.
## Original issue(s)


config/routes.rb
```diff
@@ -122,6 +122,8 @@
     get 'limited', to: 'example#limited', as: :limited
     get 'status', to: 'admin#status'
 
+    get 'eligibility', to: 'claimants#eligibility', as: :eligibility
```
rileyanderson, line 125: I think it might be best to wrap this in a feature flag for the time being (that way it can be turned on and off on the different environments) [here](https://depo-platform-documentation.scrollhelp.site/developer-docs/Feature-toggles.1859780873.html#Featuretoggles-Addinganewfeaturetoggle(backend)) and [here](https://depo-platform-documentation.scrollhelp.site/developer-docs/Using-feature-toggles-to-progressively-release-changes.590938137.html) is the documentation for that.

```suggestion
if Flipper.enabled?(:feature_flag)
  get 'eligibility', to: 'claimants#eligibility', as: :eligibility
end
```

PR #7942, opened 2021-09-17
Return end product code in claims v2 detail call

hinzed1127:
## Original issue(s)
https://vajira.max.gov/browse/API-9726


modules/claims_api/app/controllers/claims_api/v2/blueprints/claim_blueprint.rb
```diff
@@ -16,6 +17,10 @@ class ClaimBlueprint < Blueprinter::Base
           }
         end
 
+        view :list do
+          exclude :end_product_code
```
hinzed1127, line 21: Based this part off of the AC:
>    EP code is returned in the Claim Detail response
I'm assuming we're going to want the index list view to show less than a "claim by id" response, but this can be adjusted accordingly. I'm not sure if that affects what we return in the HATEOAS flavor of the claim detail view, but that's later work, anyways

jw81: This `view` feature of `BluePrinter` is nice!

Yeah, I never know where to draw the line when it comes to how much detail to provide in the `index` output vs the `show/detail` output.  I think just showing this in the `show/detail` like you've done makes sense though.  

If it's an attribute that can be `nil` at times (like this attribute), then maybe it should only be in the `show/detail`output.  That seems to make sense to me.  🤷

jw81 (approved):

rileyanderson (approved):

PR #7948, opened 2021-09-20
Add feature flag for Legacy Appeals

ElleeB:
## Description of change
Added feature keyword to settings config.
Added before_action to the controller to check if feature enabled.
Added settings to test and tested for feature disabled.

## Original issue(s)
https://vajira.max.gov/browse/API-10045


modules/appeals_api/app/controllers/appeals_api/v2/decision_reviews/legacy_appeals_controller.rb
```diff
@@ -80,4 +81,8 @@ def render_unusable_response_error
     render status: :bad_gateway,
            json: Common::Exceptions::BackendServiceException.new(detail: 'Unusable upstream response')
   end
+
+  def enabled?
+    true if Settings.modules_appeals_api.legacy_appeals
```
cilestin, line 86: Let's return the setting value directly, since it's already a boolean

yakschuss, line 86: I don't think `legacy_appeals` as a key denotes whether it's enabled or not - wouldn't `legacy_appeals_enabled: true/false` be more indicative of that fact?

config/settings.yml
```diff
@@ -242,6 +242,7 @@ modules_appeals_api:
     region: "region"
     bucket: "bucket"
   status_simulation_enabled: false
+  legacy_appeals: false
```
cilestin, line 245: You could set this to true so that you don't have to worry about forcing it to true in the `with_settings` of your specs. Remember there's another repo that controls the value here per env, the vets-api repo's config/settings.yml is the example for local development.

modules/appeals_api/spec/requests/v2/legacy_appeals_controller_spec.rb
```diff
@@ -6,33 +6,39 @@
   describe '#index' do
     context 'when only ssn provided' do
       it 'GETs legacy appeals from Caseflow successfully' do
-        VCR.use_cassette('caseflow/legacy_appeals_get_by_ssn') do
-          get_legacy_appeals(ssn: '502628285', file_number: nil)
-          expect(response).to have_http_status(:ok)
-          json = JSON.parse(response.body)
-          expect(json['data']).not_to be nil
-          expect(json['data'][0]['attributes']).to include 'latestSocSsocDate'
+        with_settings(Settings.modules_appeals_api, legacy_appeals: true) do
```
cilestin, line 9: Possibly unnecessary if you set the flag to `true` in config/settings.yml

cilestin (changes requested):

modules/appeals_api/app/controllers/appeals_api/v2/decision_reviews/legacy_appeals_controller.rb
ElleeB: OK!

ElleeB: oh yes

config/settings.yml
ElleeB: the ticket says set to false

modules/appeals_api/spec/requests/v2/legacy_appeals_controller_spec.rb
ElleeB: see above comment

cilestin (approved):
We'll clean this up and set it to `true` in config/settings.yml once we go live on prod

config/settings.yml
```diff
@@ -242,6 +242,7 @@ modules_appeals_api:
     region: "region"
     bucket: "bucket"
   status_simulation_enabled: false
+  legacy_appeals_enabled: false
```
cilestin, line 245: There's a `config/settings/test.yml` that we can use to set this to `true` and remove all the `with_settings` additions.

cilestin (changes requested):

config/settings.yml
ElleeB: ok sweet. thanks

config/settings.yml
```diff
@@ -1138,4 +1139,3 @@ bid:
 # Settings for Test User Dashboard modules
 test_user_dashboard:
   env: dev
-
```
ElleeB, line 1141: I have no idea why this is happening. I have 1142 lines in my local `settings.yml`, with line 1142 being a newline.

PR #7960, opened 2021-09-21
Dockerize ClamAV and Sidekiq in local development

f1337:
- Extracts ClamAV and Sidekiq processes into their own Docker containers in preperation for moving to Kubernetes.
- Logs ClamAV results via Rails.logger (for Loki).
- ClamAV config is now copied to /etc (default config location) instead of passed as an argument on each run.
- Foreman is not completely removed here as it's still used in higher environments.


lib/shrine/plugins/validate_virus_free.rb
```diff
@@ -9,7 +9,8 @@ module AttacherMethods
         def validate_virus_free(message: nil)
           cached_path = get.download.path
           result = Common::VirusScan.scan(cached_path)
-          # TODO: Log a the full result to sentry
+          # TODO: Log a the full result to ~sentry~ Loki
+          # Logger.info('Common::VirusScan result', result)
```
f1337, line 13: I'd love to go ahead and fix this TODO if we can. What's it take to log to Loki?

lib/common/virus_scan.rb
```diff
@@ -7,8 +7,7 @@ module VirusScan
     def scan(file_path)
       # `clamd` runs within service group, needs group read
       File.chmod(0o640, file_path)
-      args = ['--config', Rails.root.join('config', 'clamd.conf').to_s]
-      ClamScan::Client.scan(location: file_path, custom_args: args)
+      ClamScan::Client.scan(location: file_path)
```
f1337, line 10: Moved the custom config to the Dockerfile, because that's where it truly belongs.

docker-compose.yml
```diff
@@ -42,15 +46,16 @@ services:
       "Settings.saml.key_path": "/srv/vets-api/src/config/certs/vetsgov-localhost.key"
       "Settings.redis.app_data.url": "redis://redis:6379"
       "Settings.redis.sidekiq.url": "redis://redis:6379"
-      "Settings.binaries.clamdscan": "clamscan" # Not running a separate process within the container for clamdscan, so we use clamscan which requires no daemon
```
f1337, line 45: We will probably want to restore the ability for local dev to skip clamav. Then again, it's really fast, so :shrug:.

docker-compose.yml
```diff
@@ -1,5 +1,9 @@
 version: '3.4'
 services:
+  clamav:
+    image: mkodockx/docker-clamav
+    ports:
+      - "3310:3310"
```
f1337, line 6: The magic is mostly here. 🪄

config/initializers/clamscan.rb
```diff
@@ -2,4 +2,5 @@
 
 ClamScan.configure do |clam_config|
   clam_config.client_location = Settings.binaries.clamdscan
+  clam_config.default_scan_options = { stdout: false }
```
f1337, line 5: I'm unsure if this is really necessary. But stdout pollution is evil. So I suppressed it.

config/clamd.conf
```diff
@@ -1,5 +1,3 @@
 Foreground yes
-DatabaseDirectory /srv/vets-api/clamav/database
-LocalSocket /srv/vets-api/clamav/clamd.ctl
 TCPSocket 3310
-TCPAddr 127.0.0.1
\ No newline at end of file
+TCPAddr clamav
```
f1337, line 3: The hostname of the container. If we name the pods the same, this might work unchanged in k8s?

Procfile
```diff
@@ -1,4 +1,2 @@
 web: bundle exec puma -p 3000 -C ./config/puma.rb
 job: bundle exec sidekiq -q critical,4 -q tasker,3 -q default,2 -q low,1
-freshclam: /usr/bin/freshclam -d --config-file=config/freshclam.conf
-clamd: /usr/sbin/clamd -c config/clamd.conf
```
f1337, line 4: 2 down, 1 to go (job).

Dockerfile
```diff
@@ -12,13 +12,14 @@ SHELL ["/bin/bash", "-c"]
 RUN groupadd -g $userid -r vets-api && \
     useradd -u $userid -r -m -d /srv/vets-api -g vets-api vets-api
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
-    dumb-init clamav clamdscan clamav-daemon imagemagick pdftk curl poppler-utils libpq5 vim
+    dumb-init clamdscan imagemagick pdftk curl poppler-utils libpq5 vim
```
f1337, line 15: Don't need clam or the daemon, just the client.

Dockerfile
```diff
@@ -12,13 +12,14 @@ SHELL ["/bin/bash", "-c"]
...
 # The pki work below is for parity with the non-docker BRD deploys to mount certs into
 # the container, we need to get rid of it and refactor the configuration bits into
 # something more continer friendly in a later bunch of work
 RUN mkdir -p /srv/vets-api/{clamav/database,pki/tls,secure,src} && \
     chown -R vets-api:vets-api /srv/vets-api && \
     ln -s /srv/vets-api/pki /etc/pki
+COPY config/clamd.conf /etc/clamav/clamd.conf
```
f1337, line 22: because overriding the AV config at runtime feels wrong.

docker-compose.yml
```diff
@@ -42,17 +47,51 @@ services:
...
       POSTGRES_PASSWORD: "${POSTGRES_PASSWORD:-password}"
     depends_on:
+      - clamav
+      - postgres
+      - redis
+    links:
+      - clamav
+      - postgres
+      - redis
+  worker:
```
f1337, line 62: Sidekiq worker extracted too 🤘🏻 .

lib/shrine/plugins/validate_virus_free.rb
dginther: Set the docker logging driver to loki and point it at the loki server. an example: https://github.com/department-of-veterans-affairs/devops/blob/a04b11a7208379360672ae4dbb24049affd1df1b/ansible/deployment/config/vets-api-server-vagov-dev.yml#L78

docker-compose.yml
dginther: likely we do not want to pull this from Dockerhub, as we may run into quota issues. Recommend setting up a repo that mirrors the clamav container to ECR, similar to the way we do here: https://github.com/department-of-veterans-affairs/vsp-infra-sentry/blob/main/.github/workflows/mirror-images.yaml and then using the container from there.

docker-compose.yml
```diff
@@ -1,5 +1,9 @@
 version: '3.4'
 services:
+  clamav:
+    image: clamav/clamav
+    ports:
+      - "3310:3310"
   redis:
     image: redis:5.0-alpine
```
dginther, line 8: Same comment as about the clamav container.

docker-compose.yml
f1337: right on, so is the idea that i would create such a yaml in vets-api/.github/workflows/mirror-images.yml?

f1337: curious: is that an ansible file? or is that argo?

dginther: that is a github actions workflow

Makefile
```diff
@@ -148,5 +148,5 @@ else
 endif
 
 .PHONY: up
-up: db  ## Starts the server and associated services with docker-compose, use `clam=1 make up` to run ClamAV
-	@$(BASH_DEV) "rm -f tmp/pids/server.pid && foreman start -m ${FOREMAN_ARG}"
+up:
```
rileyanderson, line 151: Do we need to keep `db` here to set up the databasae?

f1337: Doh! Good catch!

f1337: fixed.

lib/shrine/plugins/validate_virus_free.rb
f1337: i think since Loki is live now, as long as I use `Rails.logger`, this should go to Loki. i hope? 🤞🏻

ericboehs (approved):

PR #7963, opened 2021-09-22
sidekiq: add DeleteOldUploadsJob to scheduler

jchilders:
## Description of change
This adds the job `Form1010cg::DeleteOldUploadsJob` to the scheduler, to be run every night at midnight ET. It is intended to remove orphaned uploads, i.e. uploads that were received but never further acted upon or processed, and which serve no further use.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#26252

I do not have access to S3, so was unable to do local verification that the job does what it is supposed to do. The specs for the job pass, and appear to cover the functionality needed. Without manual verification, though, I'm not entirely confident in this PR.


jchilders:
For the specs, locally I'm seeing different behavior. When the specs are run inside docker with `make spec_parallel_setup spec_parallel` there are failures. When they are run locally, with either `parallel_rspec` or `rspec`, they all pass.

rileyanderson (approved):

PR #7966, opened 2021-09-22
add evss id to rake task

annaswims:
## Description of change
participant id may be PII and we don't use it often, so remove it from the report. EVSS transaction id is useful though.

## Original issue(s)
helps with 
department-of-veterans-affairs/va.gov-team#11352


michelpmcdonald (approved):
Looks good

rileyanderson (approved):

PR #7967, opened 2021-09-22
Rubocop thread safety class and module attributes

thilton-oddball:
## Description of change
This PR corrects multiple ThreadSafety/ClassAndModuleAttributes offenses caught by Rubocop. Some are inline disabled where the use of class attributes was legitimate.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/29091


rileyanderson (approved):

lib/saml/health_status.rb
```diff
@@ -14,8 +14,6 @@ class << self
       include SentryLogging
       include StatusMessages
 
-      attr_accessor :consecutive_signature_failures
```
thilton-oddball, line 17: This was removed since it is not used anywhere.

bramleyjl (approved):
LGTM

PR #7971, opened 2021-09-23
Fix to parse old date format on email jobs until queue is cleared

cilestin:
## Description of change
Old jobs don't magically disappear after you change the format (derp). This fix parses both date_submitted formats so that we can clear out the old queue.

## Original issue(s)
https://vajira.max.gov/browse/API-10165
https://dsva.slack.com/archives/CSKKUL36K/p1632411192021500


modules/appeals_api/app/services/appeals_api/events/appeal_received.rb
```diff
@@ -65,7 +65,12 @@ def guid
       end
 
       def date_submitted
-        @date_submitted ||= DateTime.iso8601(opts['date_submitted']).strftime('%B %d, %Y')
+        # TODO: Remove "%m/%d/%Y" format parsing after all old jobs using that format have cleared the queue
+        @date_submitted ||= if opts['date_submitted'] =~ %r{^\d{2}/\d{2}/\d{4}}
```
yakschuss, line 69: yeccchhhhh, but also _cool_!

yakschuss (approved):

Issue #7972, opened 2021-09-23
Add log when multiple sec_ids are detected with a user's information from a SAML Response


bosawt:
Found at least one user on prod with a saved Accounts record with `sec_id='<sec_id1>,<sec_id2>'`, this has the potential to create problems on the backend, so we should at least log to see how often this happens, and potentially consider denying login when this happens

jscanish:
https://github.com/department-of-veterans-affairs/vets-api/pull/7987

nsoutouras:
Should close tomorrow when Josh returns

PR #7987, opened 2021-09-24
Log instances of multiple sec ids

jscanish:
## Description of change
We've seen a number of cases where a user's `sec_id` is a string of values and not a singular id (e.g. '1234567890,0987654321`). This pull request adds additional logging so we can better understand how often this is happening. It also cleans up a little bit of the repetition surrounding logging unexpected ID results.

## Original issue(s)
https://github.com/department-of-veterans-affairs/vets-api/issues/7972

I've added some RSpec testing to ensure we write to the logs only in the correct situations.


lib/saml/user_attributes/ssoe.rb
```diff
@@ -253,26 +262,31 @@ def mhv_id_mismatch?
...
-        corp_ids = safe_attr('vba_corp_id')&.split(',') || []
-        corp_ids.reject(&:nil?).uniq.size > 1
+        attribute_has_multiple_values?('vba_corp_id')
+      end
+
+      def sec_id_mismatch?
+        attribute_has_multiple_values?('va_eauth_secid')
+      end
+
+      def attribute_has_multiple_values?(attribute)
```
bosawt, line 287: Nice refactor!

bosawt (approved):
Looks good!

I logged in with a user that has sec_id: ["1234567,123555"], and confirmed that log: `User attributes contains multiple sec_id values` was output, then logged in with a user that has sec_id: ["1234567"], and confirmed no log

bramleyjl (approved):
LGTM, thanks for cleaning up those duplicate IDs into one method!

PR #7991, opened 2021-09-24
Update source of IAMUser id

kreek:
## Description of change
Updates the IAM User to use the user identity uuid as its id. Previously the id was generated from a SHA1 hash of last name and email. But changes to the email object returned from the IAM introspect endpoint mean this field can no be nil which would cause an error.

The user id is part of the response as JSON API spec requires it. At the application level it's used for session length caching of upstream service responses and debugging.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va-mobile-app/issues/1640#event-5357896522


jperk51 (approved):
LGTM

rileyanderson (approved):

PR #7998, opened 2021-09-27
[API-9828] - ClaimsApi - Add v2 GET Intent-to-File Endpoint

jw81:
## Description of change

Adds the `/v2/veterans/{veteranId}/intent-to-files/{type}` endpoint to ClaimsApi.

## Original issue(s)
[API-9828](https://vajira.max.gov/browse/API-9828)


modules/claims_api/app/models/claims_api/intent_to_file.rb
```diff
@@ -2,12 +2,29 @@
...
 module ClaimsApi
   class IntentToFile
+    attr_reader :id, :creation_date, :expiration_date, :status, :type
+
     ITF_TYPES = {
       'compensation' => 'C',
       'burial' => 'S',
-      'pension' => 'P'
+      'pension' => 'P',
+      'C' => 'compensation',
```
hinzed1127, line 11: why add the ITF types to go both ways? wouldn't we want to just have a one-way conversion to enforce less values?

hinzed1127: ie, having be either
```ruby
ITF_TYPES = {
 'compensation' => 'C',
 'burial' => 'S',
 'pension' => 'P'
}
```
or 
```ruby
ITF_TYPES = {
 'C' => 'compensation',
 'S' => 'burial',
 'P' => 'pension'
}
```

jw81: Eh, I mainly just did this to support line#23 of the initializer.
(`@type = ITF_TYPES[type]`)

It just made sense to me to go ahead and add the ability to map the other direction in that constant, rather than adding logic in the initializer to do that conversion.  But I'm down for alternative ideas if we don't like this.

modules/claims_api/app/controllers/claims_api/v2/veterans/intent_to_files_controller.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module ClaimsApi
+  module V2
+    module Veterans
+      class IntentToFilesController < ClaimsApi::V2::ApplicationController
```
hinzed1127, line 6: this is a nitpick out of curiosity, but any reason for opting for plural over singular on `File(s)`? I.e., `IntentToFilesController` over `IntentToFileController`

jw81: Lee and I had talked about this at one point and we felt that pluralizing the route (and therefore the controller name) felt right.  Doesn't mean that it can't be changed of course.

modules/claims_api/app/models/claims_api/intent_to_file.rb
hinzed1127: Oh I think I see now. BGS provides that code in letters and we want to provide it in full strings. I would maybe just refactor it into separate hashes to make the division more explicit. E.g. keep renaming the existing hash to `ITF_TYPES_TO_BGS` and then the new one `BGS_TO_ITF_TYPES`. or something similar. any feelings about that modification?

modules/claims_api/app/controllers/claims_api/v2/veterans/intent_to_files_controller.rb
hinzed1127: ahhh, ok. thinking about in the context of multiple ITFs for the same veteran makes more sense. I don't mind that, then

modules/claims_api/app/models/claims_api/intent_to_file.rb
jw81: I'm totally down for that.  I was trying to avoid having inverse constants, but if it introduces confusion, then yeah I'm totally cool with having 2 separate mappings.  I'll make that change and update the PR.

hinzed1127: yeah, I think as long as they're appropriately named constants, inverses of each other is easier to grok than having everything in the same hash

jw81: That change has been made [here](https://github.com/department-of-veterans-affairs/vets-api/pull/7998/commits/b8927734b5ad34373b0bae5894d51a653a7cd15c).

hinzed1127 (approved):
LGTM

PR #8007, opened 2021-09-28
 Mobile appointments fix hostname check in non-prod toggle

kreek:
## Description of change
Two facilities used for testing staging have different ids than in production. When reverse engineering the web client code the hostname check was not changed from `www.va.gov` to `api.va.gov`.

## Original issue(s)
https://app.zenhub.com/workspaces/va-mobile-eng-60f1a34998bc75000f2a489f/issues/department-of-veterans-affairs/va-mobile-app/1655


jperk51 (approved):
LGTM

PR #8013, opened 2021-09-28
[30408] ISAM/Login.gov LoA Integration

bramleyjl:
## Original issue(s)
department-of-veterans-affairs/va.gov-team#30408


thilton-oddball (approved):

lib/saml/url_service.rb
```diff
@@ -91,6 +91,14 @@ def idme_url
       build_sso_url(build_authn_context(LOA::IDME_LOA1_VETS))
     end
 
+    def logingov_url
+      @type = 'logingov'
+      build_sso_url(build_authn_context(
+                      ['http://idmanagement.gov/ns/assurance/ial/1', 'http://idmanagement.gov/ns/assurance/aal/2'],
+                      'https://eauth.va.gov/csp?Select=logingov3'
```
bosawt, line 98: We should add an entry in `Settings.yml` for this, similar to the existing `idme_authn_context`

lib/saml/url_service.rb
```diff
@@ -91,6 +91,14 @@ def idme_url
       build_sso_url(build_authn_context(LOA::IDME_LOA1_VETS))
     end
 
+    def logingov_url
+      @type = 'logingov'
+      build_sso_url(build_authn_context(
+                      ['http://idmanagement.gov/ns/assurance/ial/1', 'http://idmanagement.gov/ns/assurance/aal/2'],
```
bosawt, line 97: Can we set these as constants, similar to the LOA module? `config/initializers/loa.rb`, but just an `IAL` module definition?

lib/saml/url_service.rb
```diff
@@ -168,7 +176,8 @@ def build_sso_url(link_authn_context)
 
     def build_authn_context(assurance_level_url, identity_provider = Settings.saml_ssoe.idme_authn_context)
       if identity_provider
-        [assurance_level_url, identity_provider]
+        assurance_level_url = [assurance_level_url] unless assurance_level_url.is_a?(Array)
```
bosawt, line 179: For now, I think maybe we shouldn't touch `build_authn_context`, and just add a separate function, `build_logingov_authn_context`, that we can refactor later. Since we're looking to merge these changes without being able to properly test, I'd prefer if we lessened the chance to negatively impact existing login methods

bosawt (changes requested):
Some quick comments, just trying to think of how to make this as safe as possible given the circumstances. 

Also, we should add specs for this. Even if we can't do acceptance testing, we can still unit test these changes

PR #8017, opened 2021-09-28
[8015] Converting AfterLoginJob to synchronous AfterLoginActions

bosawt:
## Description of change
Everything in AfterLoginJob is able to be done synchronously, and even the one call we make to another service (EVSS account creation) is done with a separate job, so there's really no reason to have an AfterLoginJob here. This PR converts this into a class, AfterLoginActions, which is the same as AfterLoginJob, except that it is synchronous

## Original issue(s)
department-of-veterans-affairs/vets-api/issues/8015

- Shouldn't have any difference in production, except lack of extra sidekiq jobs
- To test this, log in with a user, and confirm that an account has been created. Also, confirm that an AccountLoginStat entry has been created


jscanish (approved):
LGTM

thilton-oddball (approved):

PR #8018, opened 2021-09-29
DependentsSerializer handle nil dependents

RudyOnRails:
## Description of change

When `get_dependents` returns nil, the `DependentsSerializer` breaks, so in place of nil, return a mock object with no persons in it.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/30642

I tried to find a cassette inside the `bgs-ext` gem but they didn't have one. We should try and make a SoapUI call for one of the EBN users with no dependents so I can create a VCR cassette and finish my request/controller spec using a BGS response that results in nil. I will create a new issue for this, but for now this is enough to handle nil.


thilton-oddball (approved):

RudyOnRails:
BEFORE:
Screen Shot 2021-10-01 at 4 05 10 PM


AFTER:
Screen Shot 2021-10-01 at 4 03 03 PM

PR #8024, opened 2021-09-29
Claims V2: Add endpoint to appoint individual POA

hinzed1127:
## Description of change

This adds the ability to make only _individual_ POA updates. Appointing an organizational POA is a forthcoming endpoint.

## Original issue(s)
https://vajira.max.gov/browse/API-9826

- This is the first endpoint in v2 that actually involves making an upstream submission. As such, there were a handful of helper methods present in v0 and v1 that needed to be added in here. Assubsequent refactor would probably be appropriate to consolidate some of these methods, whether it's in the concerns (idk Rails conventions well enough to say if that's the best way to do this), or as @jw81 suggested, moving what we can into the appropriate models. If you have any thoughts/suggestions on that aspect of future development, please don't hesitate calling it out here.
  - **Update**: Ticket for that work added [here](https://vajira.max.gov/browse/API-10870).

- Along those similar lines of thought, this will probably get refactored a bit when we implement [organizational POA](https://vajira.max.gov/browse/API-9825) after this, which is extremely similar save for some validation logic and generating a slightly different form (21-22 vs 21-22a)


modules/claims_api/app/controllers/claims_api/v2/veterans/power_of_attorney_controller.rb
```diff
@@ -17,6 +21,45 @@ def show
           )
         end
 
+        def appoint_individual
```
jw81, line 24: Unless I'm completely missing something, I think there has to be some level of authorization check here to make sure the person making the request is actually authorized to do so.  It seems like we would only want to allow the Veteran themself to actually use this endpoint.  (again, I could totally be missing something)

modules/claims_api/app/controllers/claims_api/v2/veterans/power_of_attorney_controller.rb
```diff
@@ -17,6 +21,45 @@ def show
...
+          validate_json_schema
+
+          poa_code = form_attributes.dig('serviceOrganization', 'poaCode')
+          validate_poa_code!(poa_code)
+          if !poa_code_in_organization?(poa_code)
```
jw81, line 29: Is the negation correct here?
It looks like it's saying, "_if the poa code is NOT in an organization, then raise exception_".

hinzed1127: ope, you are correct. [fixed](https://github.com/department-of-veterans-affairs/vets-api/pull/8024/commits/45e7876309e49ddcd399fbe4c15badd8c3d6bcf6)

modules/claims_api/app/controllers/claims_api/v2/veterans/power_of_attorney_controller.rb
hinzed1127: Yup, forgot to wire that in. [Added](https://github.com/department-of-veterans-affairs/vets-api/pull/8024/commits/f48bb955aaf2b76b17b980fbb68fee1c841656aa). Will probably make it a `before-` action validation once there's more of them.

hinzed1127: actually, realized that `show` is just re-implementing `verify_access!`, so I will move this to a top level controller thing

app/models/mpi_data.rb
```diff
@@ -137,8 +137,8 @@ def error
   end
 
   # @return [MPI::Responses::FindProfileResponse] the response returned from MVI
-  def mvi_response
-    @mvi_response ||= response_from_redis_or_service
+  def mvi_response(user_key: user_identity.uuid)
```
jw81, line 140: Are we sure there's no one else in `vets-api` currently calling this method and that this won't have an impact on them?

jw81 (approved):

PR #8027, opened 2021-09-29
30643 V2 Check-in authentication

dillo:
## Description of change

- V2 is behind a feature flip and is currently turned off in production
- Implement user authentication on V2 Check-in
## Original issue(s)

- [x] RSpec passing


modules/check_in/spec/request/v2/sessions_request_spec.rb
```diff
@@ -79,10 +76,105 @@
...
+
+        post '/check_in/v2/sessions', session_params
+
+        expect(response.status).to eq(200)
+        expect(JSON.parse(response.body)).to eq(resp)
+      end
+    end
 
-      expect(response.status).to eq(501)
+    context 'when unauthorized' do
```
randomsync, line 142: Is this "unauthorized" flow, since the response both from this service and LoROTA are success responses? Or is it just worded incorrectly?

dillo: Good spot. Something definitely went wrong there. I'll take a look 👍

modules/check_in/app/models/check_in/v2/session.rb
```diff
@@ -109,9 +109,7 @@ def success_message
       # @return [Faraday::Response]
       #
       def client_error
-        body = { error: true, message: 'Invalid uuid, last4 or last name!' }
-
-        Faraday::Response.new(body: body, status: 400)
+        { error: true, message: 'Invalid uuid, last4 or last name!' }
```
stephenBarrs, line 112: If this error passes through to be displayed to the user we should not mention UUID.

stephenBarrs (approved):
LGTM!

modules/check_in/app/models/check_in/v2/session.rb
dillo: Good idea. Will remove it in the next PR! 👍

thilton-oddball (approved):

PR #8029, opened 2021-09-29
[API-10127] - ClaimsApi - Handle Invalid Disability Names on Form526 submissions

jw81:
## Description of change

Adds logic to handle invalid `disability.name` values on a Form526 submission, rather than raising an exception to the end-user.
- If the `disability.name` is over 255 characters long, then truncate it to 255 characters.
- If the `disability.name` includes special characters, then remove the special characters.

## Original issue(s)
[API-10127](https://vajira.max.gov/browse/API-10127)


modules/claims_api/app/models/claims_api/auto_established_claim.rb
```diff
@@ -297,5 +298,29 @@ def breakout_date_components(date:)
...
+    # EVSS requires disability names to be less than 255 characters and cannot contain special characters.
+    # Rather than raise an exception to the user, massage the name into a valid state that EVSS will accept.
+    def massage_invalid_disability_names
+      disabilities = form_data.dig('disabilities')
+      invalid_characters = %r{[^a-zA-Z0-9\\\-'\.,\/\(\) ]}
+
+      disabilities.map do |disability|
+        name = disability['name']
+        name = truncate_disability_name(name: name) if name.length > 255
+        name = sanitize_disablity_name(name: name, regex: invalid_characters) if name.match?(invalid_characters)
```
hinzed1127, line 311: stylistic nitpick: I think the 2 helper methods are fine as inline calls (more readable and they're short calls that are clear as to what they're doing), but I'll leave that up to you.

hinzed1127 (approved):
Small thing, but good either way

PR #8036, opened 2021-09-30
Revert populating mobile_users table

kreek:
## Description of change
Reverts the application code from #7913 which assumes the `mobile_users` is correctly migrated. But keeps [the appending of local engine migrations](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/mobile/lib/mobile/engine.rb#L10-L17) in `modules/mobile/lib/engine.rb` so that migrations can be run.

## Original issue(s)
https://app.zenhub.com/workspaces/va-mobile-eng-60f1a34998bc75000f2a489f/issues/department-of-veterans-affairs/va-mobile-app/1666


jperk51 (approved):
LGTM

patrickvinograd (approved):

PR #8037, opened 2021-09-30
add source apps to tags for statsd and sentry

annaswims:
## Original issue(s)


dginther:
> make a grafana graph that uses these stats/tags

I'd say it would be better to make a DataDog graph that uses these stats/tags. The statsd metrics are already being piped over to DD, and the prometheus statistics in Grafana will be gone once we sunset Prometheus. Let me know if you need an invite to DD!

thilton-oddball (approved):

PR #8038, opened 2021-09-30
1660 mobile user to include signin service

jperk51:
## Description of change
Added sign-in service to mobile user object so we can react to it on the FE

## Original issue(s)
department-of-veterans-affairs/va-mobile-app#1660

Added a new test to make sure it's in the user response and updated the docs accordingly


modules/mobile/docs/schemas/User.yml
```diff
@@ -54,6 +55,9 @@ properties:
             $ref: "./PhoneUpdate.yml"
           faxPhoneNumber:
             $ref: "./PhoneUpdate.yml"
+          signinService:
+            type: string
```
kreek, line 59: Could this be an enum with the values?

kreek (approved):
One minor doc comment, but LGTM 🚀

modules/mobile/docs/schemas/User.yml
jperk51: Yup, good call

PR #8050, opened 2021-10-01
Rake login task

jscanish:
## Description of change
This is a simple rake task designed to be run after logging in with a particular LOA1 ID.me user (yet to be decided). It will validate that the data we have access to is what we would expect to see for an LOA1 user.

Note, the task is designed with a specific user account in mind, hence the hardcoded values. We'll have to decide what that user account looks like.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/30827

Ran the task to confirm that it works correctly and catches important errors.


rakelib/login.rake
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+namespace :login do
+  desc 'Login as <some test user>, then run this task to validate that LOA1 credentials are set properly'
+  task idme_loa_1: [:environment] do
```
bramleyjl, line 5: Great start! I have two main thoughts so far.

I'm wondering if it would be a lot more flexible if we let the user pass in the account identifier as a param, that way we could log in and test on whatever account we want? Ideally we'd want to look them up by email but we could start with `uuid` as a simpler & safer alternative.

Also thinking we should break up the tests into blocks based on IDP, then we can run an IDP's suite of tests all together when we're logged in on the right account for it. If it gets big enough we might even think about breaking out each IDP's test into their own file.

Thinking something like this:
```
task :idme, [:uuid] => [:environment] do
  user = User.find(args[:uuid])
  //run your basic 'did it find a user' validation here
  Rake::Task["login:idme_loa_1"].invoke
  Rake::Task["login:idme_loa_3"].invoke

Comment (jscanish) [reply rakelib/login.rake]: I made the change to accept `uuid` as an argument, but I feel like we should hold off on trying to group tests at this point. It's a good idea, and we definitely should group tests in that way at some point, but as it stands now the manual process this task relies on makes running LOA1 and LOA3 tests together as a package kind of impossible. It's difficult to simulate logging in as more than one user at a time and you would expect opposing results in a lot of cases. I think as we write additional and more complex tasks we'll be able to more naturally judge how they should be grouped later.

Is there anything else you think we should be testing for an ID.me LOA1 user? I'm not sure what else would be valuable to check.

**Review by bosawt — APPROVED** [review]
- Tested this with an LOA3 user, got the expected error that it's not an LOA1 user
- Tested with an LOA1, user, got the expected passing validations:

> Running login validations...
> User ID.me UUID matches user account ID.me UUID
> User has an LOA level of 1
> User does not have SSN
> User account login stats incremented correctly
> All LOA1 validations pass!

File: rakelib/login.rake (hunk 1 above)
Comment (bramleyjl) [reply rakelib/login.rake]: Right, makes sense to keep it more straightforward right now. As for other things to test for LoA1 I can't think of anything super important except maybe also checking the AuthN context that comes back to us?

```
validate(
      user.authn_context == LOA::IDME_LOA1_VETS,
      'User has an ID.me LoA1_Vets Authn Context',
      'User does not have an ID.me LoA1_Vets Authn Context'
    )
```

bosawt:
Some ideas of where to go from here:
- We should probably create a class to hold most of this, and have the rake task call the class. This lets us define many tasks in this Rakefile and not have to clutter up the file with the implementation of each task
- We could automate the aspect of this requiring us to find the user_uuid we just logged in with, by just inspecting the latest entry updated in the `AccountLoginStat` table, then we just grab the `account_id` from that, and the `idme_uuid` from that Account table entry. There will be a few exceptions here (which is why letting us specify the user_uuid is probably still good) but that should mostly work and would be a step further towards completely automating this
- If we're trying to exhaustively test everything we can about an LOA1 user, here's all of the points I can think of (these are in addition to what you already have): first name exists, last name exists, email exists, credential identifier exists (mhv_id, edipi, idme_uuid, depending on which credential is logged in with), sec_id exists, sec_id matches accounts table sec_id

PR #8052, opened 2021-10-01
API-10053 - Update swagger json and add dev swagger json

nathanbwright:
## Description of change
This PR adds  `swagger_dev.json` for use in the lower envs and removes the standalone contestable issues and legacy appeals endpoint docs from the main `swagger.json` file for use in prod (there is also quite a bit of noise because the schemas are now ordered alphabetically).

## Original issue(s)
https://vajira.max.gov/browse/API-10053


ElleeB (dismissed):
Looks good to me!

yakschuss (approved):
🚀

PR #8053, opened 2021-10-01
Mobile user table

jperk51:
## Description of change
Adding back logic for mobile user table now that the db migration has been confirmed. 

## Original issue(s)

Tests were added for new logic


jperk51:
Moved to draft since this was part of an architecture for push notifications that we may not use any more

PR #8057, opened 2021-10-04
[26521] add 60s statement timeout for postgres

cbtkc:
## Description of change
Adds a 60 second timeout to database queries, which will prevent long running queries from running indefinitely.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#26521


ericboehs:
This will affect migrations. How should we handle this, especially as we move to CD with auto migration runs?

[Kane's timeout guide suggests](https://github.com/ankane/the-ultimate-guide-to-ruby-timeouts#postgresql) adding a `STATEMENT_TIMEOUT` env var. Maybe when we auto-run migrations we set the timeout to 5 minutes or something?

cbtkc:
> This will affect migrations. How should we handle this, especially as we move to CD with auto migration runs?
> 
> [Kane's timeout guide suggests](https://github.com/ankane/the-ultimate-guide-to-ruby-timeouts#postgresql) adding a `STATEMENT_TIMEOUT` env var. Maybe when we auto-run migrations we set the timeout to 5 minutes or something?
> 
> Should we just wait adding the ENV var until this becomes a problem?

Yes, good call on that - I will add it.

thilton-oddball (approved):

PR #8065, opened 2021-10-04
combine spec/jobs and spec/workers

annaswims:
## Description of change
there's no reason to have both `spec/jobs` and `spec/workers`.   I think `spec/workers` would be the preferable one to keep, but I went with `spec/jobs` because it had many more files and I wanted to minimize potential merge conflicts.


annaswims:
The failing test is a flaky one unrelated to this change

thilton-oddball (approved):

PR #8070, opened 2021-10-04
Making PR7102 smaller by removing this code

jperk51:
## Description of change
Removing code that was removed with #7102 so that PR will be smaller

## Original issue(s)
department-of-veterans-affairs/va.gov-team#24909
department-of-veterans-affairs/va-mobile-app#1070

Only removed code that wasn't being used and was removed as a part of #7102


jperk51:
Confirmed with Patrick S that this endpoint was not being used by the app currently

lexicalninja (approved):

Teesh (approved):

PR #8095, opened 2021-10-06
Remove attr_encrypted from model

LindseySaari:
## Description of change
As part of the migration to Lockbox, this PR removes attr_encrypted from the model layer in vets-api. It also removes the attr_encrypted gem from the Gemfile. This is PR number 2 in the three step process to remove attr_encrypted. Next up will be the removal of the attr_encrypted db attributes.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#29612

- [x] Updated specs accordingly


modules/covid_vaccine/app/models/covid_vaccine/v0/expanded_registration_submission.rb
```diff
@@ -47,19 +47,13 @@ class ExpandedRegistrationSubmission < ApplicationRecord
...
-                                        marshaler: JsonMarshal::Marshaller
-
       serialize :eligibility_info, JsonMarshal::Marshaller
       serialize :form_data, JsonMarshal::Marshaller
       serialize :raw_form_data, JsonMarshal::Marshaller
 
-      encrypts :form_data, migrating: true, **lockbox_options
-      encrypts :eligibility_info, migrating: true, **lockbox_options
-      encrypts :raw_form_data, migrating: true, **lockbox_options
+      encrypts :form_data, **lockbox_options
```
thilton-oddball, line 54: Super nitpicky but could these three lines just be on one line?

encrypts :form_data, :eligibility_info, :raw_form_data, **lockbox_options

thilton-oddball (dismissed):
One comment but otherwise looks great!

PR #8099, opened 2021-10-06
Remove Layout/BeginEndAlignment from rubocop todo

rileyanderson:
## Description of change
RuboCop provides many safety checks and best practices for use in vets-api. We're currently ignoring ~3000 RuboCop violations. Some of these are legitimate and should be ignored, but the bulk of them should not be and should be properly addressed.

**Layout/BeginEndAlignment**

These changes are consitent to what is in the rails codebase. See example [here](https://github.com/rails/rails/blob/6-1-stable/actionview/test/template/template_error_test.rb#L7-L11)

## Original issue(s)
department-of-veterans-affairs/va.gov-team#31093


thilton-oddball (approved):

michelpmcdonald (approved):
Looks much better now, thank you

PR #8025, opened 2021-09-29
API-8918: Supplemental claim form data and structure

ElleeB:
Ticket: https://vajira.max.gov/browse/API-8918

PR #8110, opened 2021-10-06
API-8918 Split 1 Supplemental claim model updates

ElleeB:
Breaking down PR: https://github.com/department-of-veterans-affairs/vets-api/pull/8025

Adds model-related files and updates.


modules/appeals_api/app/models/appeals_api/supplemental_claim.rb
```diff
@@ -28,6 +28,120 @@ def self.date_from_string(string)
       if: proc { |a| a.form_data.present? }
     )
 
+    def pdf_structure(version)
```
yakschuss, line 31: I would add this method together with the `Structure` object

yakschuss: Our `PdfConstruction::Generator` model calls `pdf_structure` on the appeal, in order to get a reference to the Structure instance. However, in this specific case, there's no PdfConstruction::SupplementalClaim::Structure` in our code base. I'm saying that logically, you should add the reference to the structure with the pdf_structure interface in the same Pull Request. Does that make sense?

modules/appeals_api/app/models/appeals_api/supplemental_claim.rb
```diff
@@ -31,6 +31,114 @@ def self.date_from_string(string)
...
+
+    def mailing_address_box
+      veteran.dig('address', 'addressLine3') || ''
+    end
+
+    def mailing_address_city
+      veteran.dig('address', 'city') || ''
+    end
+
+    def mailing_address_city_and_box
```
cilestin, line 94: Shouldn't the addressline3 be a part of the address and not the city?  Doesn't seem correct to have `Cincinnati PO Box 9` in the city field.

yakschuss (approved):

modules/appeals_api/app/models/appeals_api/supplemental_claim.rb
```diff
@@ -31,6 +31,114 @@ def self.date_from_string(string)
...
+
+    def mailing_address_number_and_street
+      veteran.dig('address', 'addressLine1') || ''
+    end
+
+    def mailing_address_apartment_or_unit_number
+      veteran.dig('address', 'addressLine2') || ''
+    end
+
+    def mailing_address_box
```
cilestin, line 86: I don't think we should assume what these address lines represent. It's not guaranteed that every PO Box will come in on `addressLine3` - if the address is only their PO Box, they'd likely put it in `addressLine1`. I think it's fine to keep these method names tied to their schema fields.

modules/appeals_api/app/models/appeals_api/supplemental_claim.rb
```diff
@@ -31,6 +31,114 @@ def self.date_from_string(string)
...
+    def mailing_address_state
+      veteran.dig('address', 'stateCode') || ''
+    end
+
+    def mailing_address_country
+      veteran.dig('address', 'countryCodeISO2') || ''
+    end
+
+    def zip_code
+      if zip_code_5 == '00000'
```
cilestin, line 107: A tiny optimization in this case, but something to keep in mind in general - `zip_code_5` is being called multiple times here, and it's not memoized nor (alternatively) is the return value cached in this method to avoid repeating potentially expensive computations. 

If `zip_code_5` took 1 second to perform, then there's a potential for this method to take 2 seconds to return simply because it's calling that same method twice. If it'd been memoized or had its return value cached in this method, the worst case would be a 1 second call.

cilestin (commented):
Comments after merge. Y'all were too fast! 😆

PR #8155, opened 2021-10-12
fix linting error after ruby upgrade

annaswims:
it seems that ruby 2.7 has different opinions on style so master has a linting error


annaswims:
fixed in #8158

PR #8219, opened 2021-10-20
Add disagreementArea data to HLR unit test

michelpmcdonald:
## Description of change
Higher Level Reviews added a new field, area of disagreement.  New field was added to vets-json-schema repo.  This job adds area of disagreement test data to make sure incoming HRLs JSON with area of disagreement data validates against the JSON schema.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#31143

HLRs v2 is not live yet, behind a feature flag, only enabled in staging at this point.


michelpmcdonald:
> I see you've added data to the yml file, but do you need to update the unit test to check for this field too?

Good catch, in this case, the vets-api is the middle man between the website and the lighthouse HLR endpoint, we just kind of pass along the json data along....receive it from the website, validate the data against the HLR JSON Schema, and make a lighthouse call with the to lighthouse with the incoming JSON if it's valid.

The library we use to validate the incoming JSON schema against the HLR JSON schema does access the new field, it's just not an explicit test.  The new test field data I added to to the VCR is used in existing unit tests, but the test did not require any specific modification.  At first, I forgot to to bring in the newest HLR JSON Schema(via the latest version of vets-json-schema gem) and the unit test failed because the validation detected the new field in the data, but not in the JSON Schema.  As soon as I updated the gem to the latest version with the Updated JSON schema(that included the new field), the test started passing.  Accidental TDD, I guess....

PR #8320, opened 2021-10-28
[8318] Fixing slow VBADocuments::UploadSubmission query by adding app…

bosawt:
…ropriate indexes and rearranging the query

## Description of change
Looking at PGHero stats on production, I noticed that a query for `vba_documents_upload_submissions` was taking roughly 50% of the database pressure of the entire backend. Part of the query uses fields that don't have indexes, so adding indexes to these should speed up the query considerably. 

## Original issue(s)
department-of-veterans-affairs/vets-api/issues/8318

- Separating this query into components and analyzing the query shows that sequential scans are used on the created_at and s3_deleted 'where' conditions

- I tested this with a migration and a rollback, to make sure it can go either way
- I also analyzed the query after the indexes were in place and confirmed that the sequential scan is now replaced with index scans


LindseySaari (approved):
👏 👏

PR #8321, opened 2021-10-28
API-10583: Remove old date format from AppealReceived

cilestin:
## Description of change
Removes old job date format parsing now that those jobs have been properly cleared from the retry & dead sets.

## Original issue(s)
https://vajira.max.gov/browse/API-10583


cilestin:
No, it's required - removing the iso8601 format would put it back into needing the old date format (calling `to_s` on a Date object puts it into MM/DD/YYYY format)

ElleeB (approved):

PR #8324, opened 2021-10-28
Update Form526JobStatus Error Tracking

cnemeth:
## Description of change

Form 526 submission error history was lost because only the error from the last run was preserved. This PR adds a jsonb attribute to Form526JobStatus where the errors of successive runs (submission retries) are aggregated.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#31489

- Prerequisites PR #8353 and PR #8338
- Legacy attributes of Form526JobStatus, error_class and error_message, are preserved for now.
- Future tasks:
  - add rake task to move historical data into new jsonb attribute
  - add migration to remove legacy attributes error_class and error_message


db/migrate/20211022224127_add_errors_to_form526_job_statuses.rb
```diff
@@ -0,0 +1,11 @@
+class AddErrorsToForm526JobStatuses < ActiveRecord::Migration[6.1]
+  def change
+    safety_assured do
+      change_table :form526_job_statuses do |t|
+        t.jsonb :bgjob_errors, default: {}
```
michelpmcdonald, line 5: Why are we using a column type(jsonb) designed to store-query unstructured data to store structured data that, as far as I know, could be stored using our current relational table pattern ? 

I thought the use case for storing data in a json column is that the struct(fields) of the data varies from record to record(unstructured).  Our error data is structured the same, we don't have different fields per error record.

After we discussed this with Dillo, my understanding was that we decided to just keep it relational unless we end up with a specific error or use case that our pure relational DB approach could not reasonably address.
Did you hit an issue that forced you to store the structured error data in a jsonb column ?

cnemeth: > Why are we using a column type(jsonb) designed to store-query unstructured data to store structured data that, as far as I know, could be stored using our current relational table pattern ?

`jsonb` can store a strings, arrays, hashes, etc. There is no hard rule about this. 

We could just store the content of `error/data` as is in the `jsonb` attribute but I added the `error_class` and `error_message` keys to follow the semantics of the existing code. 

> After we discussed this with Dillo
I'm aware what we discussed earlier. However, I chose this solution because it was the path of least resistance. This solution requires virtually no modification of the existing models, model relations and code unlike the other suggested solution.

LindseySaari:
The /db changes will need to be broken out into a separate PR for backwards compatibility reasons.

LindseySaari (changes requested):
The /db changes will need to be broken out into a separate PR for backwards compatibility reasons.

LindseySaari (approved):

PR #8332, opened 2021-10-29
[8331] Renames references columns in user_verifications and deprecate…

bosawt:
…d_user_accounts tables

## Description of change
This PR updates foreign key references columns on user_verifications and deprecated_user_accounts tables. They need to be singular for Rails to perform it's 'magic' with column references

## Original issue(s)
department-of-veterans-affairs/vets-api/issues/8331

This runs drop_table on user_verifications, deprecated_user_accounts, and user_accounts tables. This is done because it's just way easier to drop the tables and start from scratch, rather than doing everything necessary for rails to rename the columns and recreate the foreign_key indexes
These tables are not in use and not in code yet, so shouldn't be an issue to do this


LindseySaari (approved):

LindseySaari:
I can run the migrations when you're ready! Just ping in the platform support channel

PR #8333, opened 2021-11-01
Create submit mock endpoint for claim submission

cesaresantiago:
## Description of change
- Updated a couple of the get endpoint mocks to match what they will get from DGI services.
- Created mock submit endpoint so client can send a request and continue integration work.

## Original issue(s)


modules/meb_api/app/controllers/meb_api/v0/base_controller.rb
```diff
@@ -5,6 +5,7 @@ module V0
     class BaseController < ::ApplicationController
       before_action :check_flipper
       skip_before_action :authenticate
+      #       skip_before_action :verify_authenticity_token
```
LindseySaari, line 8: Is this meant to be commented out?

cesaresantiago: @LindseySaari : Yes, but I've taken out the comment now on new push. to address your comment. Thanks.

LindseySaari: Ok great! Sounds good, it looks like you'll have to update this branch with the lastest master. There was a fix to bump the brakeman version.

bradbergeron-us:
After updating to latest master it looks that triggered some additional test failures in appeals api. Is there another commit expected to address these failures?

modules/meb_api/app/controllers/meb_api/v0/education_benefits_controller.rb
```diff
@@ -3,58 +3,84 @@
 module MebApi
   module V0
     class EducationBenefitsController < MebApi::V0::BaseController
+      # disabling checks while we serve big mock JSON objects. Check will be reinstated when we integrate with DGIB
+      # rubocop:disable all
```
LindseySaari, line 7: Can you inline disable the specific rubocop error here instead of a disable all?

modules/meb_api/config/routes.rb
```diff
@@ -6,5 +6,6 @@
     get 'service_history', to: 'education_benefits#service_history'
     get 'eligibility', to: 'education_benefits#eligibility'
     get 'claim_status', to: 'education_benefits#claim_status'
+    post 'submit_claim', to: 'education_benefits#submit_claim'
```
LindseySaari, line 9: Is there updated swaqgger for this new endpoint?

modules/meb_api/app/controllers/meb_api/v0/education_benefits_controller.rb
cesaresantiago: Will address.

modules/meb_api/config/routes.rb
cesaresantiago: Actually, @LindseySaari , I was just talking to @bradbergeron-us and he tells me we have not added any swagger on this API as the services that this API will talk to are not yet finalized and it is not settled yet how the request/responses will look. At this time we are just adding mocks that allow the client to ping the service to allow us to make progress. Is it ok to add swagger on a later PR once we reach a final determination on the API?

Thanks

Cesar

modules/meb_api/app/controllers/meb_api/v0/base_controller.rb
cesaresantiago: I've just merged master into branch and pushed.

modules/meb_api/config/routes.rb
LindseySaari: Yes definitely! Thanks for the thorough explanation!

LindseySaari (commented):
My only request is the inline rubocop disable, other than that, this looks good!

modules/meb_api/app/controllers/meb_api/v0/education_benefits_controller.rb
cesaresantiago: I've added   # rubocop:disable Metrics/MethodLength instead of disable all. I think we are good now.

LindseySaari (approved):

PR #8334, opened 2021-11-01
31760 chip service layer refactor phase one

dillo:
## Description of change

- Adding CHIP services client for making HTTP requests.
## Original issue(s)
department-of-veterans-affairs/va.gov-team#31760

- [x] RSpec passing
- [x] Rubocop passing


modules/check_in/app/services/v2/chip/client.rb
```diff
@@ -0,0 +1,109 @@
...
+
+      private
+
+      ##
+      # Create a Faraday connection object that glues the attributes
+      # and the middleware stack for making our HTTP requests to Chip
+      #
+      # @return [Faraday::Connection]
+      #
+      def connection
```
LindseySaari, line 88: Can this move to a configuration class? I know most of the service objects follow that pattern throughout the code base

dillo: I think that's a great idea. The intent is to eventually pull everything into a singular location once the refactoring phases are complete 👍

LindseySaari (approved):

PR #8337, opened 2021-11-01
1757 mobile vaccine facilities

jperk51:
## Description of change
Added health location endpoint for vaccine record info

## Original issue(s)
department-of-veterans-affairs/va-mobile-app#1757

Added tests and updated docs. Additionally updated docs for the vaccine record endpoint itself

Location model update required updating some of the existing appointments logic. I could rework it to be a separate model if that seems like a better call. The model was updated to include an id field


modules/mobile/app/controllers/mobile/v0/immunizations_controller.rb
```diff
@@ -9,6 +9,17 @@ def index
         render json: Mobile::V0::ImmunizationSerializer.new(immunizations_adapter.parse(service.get_immunizations))
       end
 
+      def location
+        lh_location = service.get_location(params[:id])
+        id_string = lh_location[:identifier][0][:value]
```
kreek, line 14: `[0]` is 100% correct but you can also do `.first` which can be easier to read.

modules/mobile/app/controllers/mobile/v0/immunizations_controller.rb
```diff
@@ -9,6 +9,17 @@ def index
...
+        id = id_string.slice(4, id_string.length)
```
kreek, line 15: You could also do this without passing in the length `id_string[4..-1]` or `lh_location[:identifier].first[:value][4..-1]`

modules/mobile/app/models/mobile/v0/adapters/immunizations.rb
```diff
@@ -24,6 +24,19 @@ def parse(immunizations)
           end
         end
 
+        def parse_location(facility, lh_id)
```
kreek, line 27: Could this be in its own class? There might be a time we fetch a location outside of immunizations.

modules/mobile/app/serializers/mobile/v0/immunization_location_serializer.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Mobile
+  module V0
+    class ImmunizationLocationSerializer
```
kreek, line 5: Same as above Locations aren't always tied to Immunizations

modules/mobile/config/routes.rb
```diff
@@ -14,6 +14,7 @@
     post '/claim/:id/request-decision', to: 'claims_and_appeals#request_decision'
     get '/disability-rating', to: 'disability_rating#index'
     get '/health/immunizations', to: 'immunizations#index'
+    get '/health/location/:id', to: 'immunizations#location'
```
kreek, line 17: URL looks good, I'd give it its own controller and following Rails conventions it would be the #show action.

modules/mobile/docs/openapi.yaml
```diff
@@ -368,6 +368,149 @@ paths:
             application/json:
               schema:
                 $ref: "./schemas/Errors.yml"
+  /v0/health/immunizations:
```
kreek, line 371: Thanks for adding this! 🏆

modules/mobile/app/controllers/mobile/v0/immunizations_controller.rb
```diff
@@ -9,6 +9,17 @@ def index
         render json: Mobile::V0::ImmunizationSerializer.new(immunizations_adapter.parse(service.get_immunizations))
       end
 
+      def location
+        lh_location = service.get_location(params[:id])
```
kreek, line 13: I think we'll never get `nil` for a location as not finding one should return a 404. Any chance walking the data (`[:identifier][0][:value]`) would end up with a nil value?

modules/mobile/app/helpers/mobile/facilities_helper.rb
```diff
@@ -16,5 +16,41 @@ def get_facility_names(facility_ids)
     def facilities_service
       Lighthouse::Facilities::Client.new
     end
+
+    def address_from_facility(facility)
+      address = facility.address['physical']
+      return nil unless address
+
+      Mobile::V0::AppointmentAddress.new(
```
kreek, line 24: Maybe we rename this `Mobile::V0::Address` now that we're using it across products? We determine the address shape for our models so I think it's safe, and we could always add fields if another product needs it (as you did with id).

kreek (changes requested):
Thanks for getting this together so quickly! Main feedback is Location being tied to Immunization when I think it can be its own controller/model/adapter.

modules/mobile/app/controllers/mobile/v0/immunizations_controller.rb
jperk51: My guess would be no since not found is a 404

modules/mobile/app/controllers/mobile/v0/immunizations_controller.rb
```diff
@@ -15,6 +15,10 @@ def immunizations_adapter
         Mobile::V0::Adapters::Immunizations.new
       end
 
+      def locations_adapter
```
kreek, line 18: Think this got left behind

kreek (approved):
One minor comment, but LGTM 🚀

kreek (approved):
LGTM 🚀

PR #8354, opened 2021-11-02
Chaning how we handle PPMS errors

darkbushido:
this should stop the error_handler from raising a nil on body
I'm also going to try passing the `innererror` message as the source
to help with debugging

## Description of change

The logic around what the PPMS parser was doing had an issue when the message wasn't one of two cases
This should stop raising errors when we try to access the body in the error_handler.
This tries to pass the inner error message as the source to help with some debugging

## Original issue(s)
department-of-veterans-affairs/va.gov-team#31587


modules/facilities_api/app/services/facilities_api/v1/lighthouse/middleware/errors.rb
```diff
@@ -28,5 +28,3 @@ def parse_body(env)
     end
   end
 end
-
```
LindseySaari, line 31: Is there a specific reason for removing the Faraday::Response.register_middleware and explicitly calling  FacilitiesApi::V1::Lighthouse::Middleware::Errors instead?

modules/facilities_api/app/services/facilities_api/v1/ppms/configuration.rb
```diff
@@ -2,8 +2,8 @@
 
 require 'common/client/configuration/rest'
 require 'common/client/middleware/response/raise_error'
-require 'common/client/middleware/response/ppms_parser'
-require 'facilities/ppms/do_not_encoder'
+# require 'common/client/middleware/response/ppms_parser'
```
LindseySaari, line 5: Can this commented out require be removed?

LindseySaari (commented):
Just a few comments/questions mentioned inline

modules/facilities_api/app/services/facilities_api/v1/lighthouse/middleware/errors.rb
darkbushido: Namespace collision, the lighthouse client was copied from the client in lib. It looks like it wasn't using the correct one even though they are currently identical

LindseySaari (dismissed):

modules/facilities_api/app/services/facilities_api/v1/lighthouse/middleware/errors.rb
LindseySaari: I see you made a change here. Did something go wrong?

LindseySaari (approved):

mpelzsherman (approved):

PR #8356, opened 2021-11-02
[32323] adds handling to add postLogin param to redirect url

bramleyjl:
## Description of change

As part of MHV switching over to va.gov's unified sign-in we need to include the `postLogin` param in the redirect url we send the user to after authentication with vets-api. This change updates `vets-api` to look for the `postLogin` param and append it to the end of a redirect if it is found to be true.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#32323

Manual login testing of redirect urls with and without `postLogin` param, unit specs updated.


bosawt (approved):
Looks good! I tested with `v1/sessions/idme/new?redirect=url-to-redirect&postLogin=true` and can confirm that these were added to the redirected URL after the `saml_callback` was processed

PR #8360, opened 2021-11-03
8323 create model definitions for user accounts tables

bosawt:
## Description of change
This PR introduces model definitions for UserAccount, UserVerification, DeprecatedUserAccount, letting the three tables interact with each other in the complete ways they were intended to work together

## Original issue(s)
department-of-veterans-affairs/vets-api/issues/8323

- This is quite well covered in specs
- Tested this by creating instances of each model and confirming just the basics of the models, each model association can be referred to from the appropriate models, dependent models are destroyed when appropriate, etc


spec/models/deprecated_user_account_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+  let(:user_verification) { create(:user_verification) }
+  let(:user_account) { create(:user_account) }
+
+  describe 'validations' do
+    describe '#user_account' do
+      subject { deprecated_user_account.user_account }
+
+      context 'when user_account is nil' do
+        let(:user_account) { nil }
+        let(:expected_error_message) { "Validation failed: User account must exist, User account can't be blank" }
```
joeniquette, line 18: Im always nervous when I see `'` in my error messages as I cant always guarantee some other system wont be vulnerable to an attack based on it. I could be a little tin hat here, but if its all the same, can we change `can't` to `cannot`?

spec/models/deprecated_user_account_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+      end
+    end
+
+    describe '#user_verification' do
+      subject { deprecated_user_account.user_verification }
+
+      context 'when user_verification is nil' do
+        let(:user_verification) { nil }
+        let(:expected_error_message) do
+          "Validation failed: User verification must exist, User verification can't be blank"
```
joeniquette, line 40: Im always nervous when I see `'` in my error messages as I cant always guarantee some other system wont be vulnerable to an attack based on it. I could be a little tin hat here, but if its all the same, can we change `can't` to `cannot`?

joeniquette (approved):
Looks good, modifying the cant to cannot is optional.

LindseySaari (approved):
👍 Also, if the newly added models here are something that identity will mostly be working with, feel free to add a codeowners entry

spec/models/deprecated_user_account_spec.rb
bosawt: Ha, unfortunately, you'll have to take it up with the `Rails` project itself. These errors are validation failures from Rail's `ActiveRecord`

bosawt: Although these are model validations and won't be serialized directly to pass back as an api response if that's any consolation

PR #8362, opened 2021-11-03
[API-10808] Add in check for CCG token in claims v1 authorization

hinzed1127:
## Description of change

This adds in support for the Client Credentials Grant authorization flow for Claims v1 endpoints

## Original issue(s)
https://vajira.max.gov/browse/API-10808

### Testing

- Tested and verified in local development using the [test client described here](https://lighthouseva.slack.com/archives/CTN4N4PL2/p1635882730052800).


modules/claims_api/app/controllers/claims_api/v1/application_controller.rb
```diff
@@ -91,6 +93,18 @@ def veteran_from_headers(with_gender: false)
         vet
       end
 
+      def validate_ccg_token!
+        client = TokenValidation::V2::Client.new(api_key: Settings.token_validation.api_key)
+        request_method_to_scope = {
+          'GET' => 'claim.read',
+          'PUT' => 'claim.write',
+          'POST' => 'claim.write'
+        }
```
hinzed1127, line 102: to my knowledge, these are the only 2 OAuth scopes for the claims endpoints. Since `token_valid` currently checks against a single scope, I figured this is a good way to generalize what scopes different request types get

modules/claims_api/app/controllers/claims_api/v1/application_controller.rb
```diff
@@ -91,6 +93,19 @@ def veteran_from_headers(with_gender: false)
...
+        raise ::Common::Exceptions::Forbidden unless
+          client.token_valid?(audience: "#{request.base_url}/services/claims",
```
hinzed1127, line 104: `request.base_url` probably needs to adjust for local requests, right? localhost won't work for this since it's hosted separately

jw81: Yeah.  If you're running this locally, you'd need to replace `request.base_url` with the url of whatever environment you're trying to talk to.

jw81:
I think this looks pretty simple and straightforward. 👍 

We'll have to wait to merge this though until we request Team Pivot setup the necessary Prod resources for us.
There's a ticket for us to go make that request [here](https://vajira.max.gov/browse/API-11179).
I'm also currently trying to figure out if there's a way for us to use the deployed review instance so we can functionally try it out.  Would be awesome if we could figure that out and try it before merging into Master.

modules/claims_api/app/controllers/claims_api/v1/application_controller.rb
```diff
@@ -91,6 +93,19 @@ def veteran_from_headers(with_gender: false)
         vet
       end
 
+      def validate_ccg_token!
+        client = TokenValidation::V2::Client.new(api_key: Settings.token_validation.api_key)
```
jw81, line 97: The setting will have to be changed to :: `Settings.claims_api.token_validation.api_key` to match what has been stored in the DEVOPS repo.

modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
```diff
@@ -10,9 +10,8 @@ class PowerOfAttorneyController < ClaimsApi::V1::Forms::Base
         include ClaimsApi::EndpointDeprecation
         include ClaimsApi::PoaVerification
 
-        before_action except: %i[schema] do
```
hinzed1127, line 13: schema got moved up to `ClaimsApi::V1::Forms::Base`, so this wasn't needed and forgotten to be removed, as is

modules/claims_api/app/controllers/claims_api/v1/application_controller.rb
```diff
@@ -91,6 +93,22 @@ def veteran_from_headers(with_gender: false)
         vet
       end
 
+      def validate_ccg_token!
+        client = TokenValidation::V2::Client.new(api_key: Settings.claims_api.token_validation.api_key)
+        root_url = request.base_url == 'http://localhost:3000' ? 'https://sandbox-api.va.gov' : request.base_url
```
hinzed1127, line 98: this seems clean enough to me, but let me know if there's a better way to accomplish this

modules/claims_api/app/controllers/claims_api/v1/application_controller.rb
```diff
@@ -91,6 +93,22 @@ def veteran_from_headers(with_gender: false)
...
+        claims_audience = "#{root_url}/services/claims"
+        request_method_to_scope = {
+          'GET' => 'claim.read',
+          'PUT' => 'claim.write',
+          'POST' => 'claim.write'
+        }
+
+        @is_token_valid ||= client.token_valid?(audience: claims_audience,
```
hinzed1127, line 106: I'm not good enough at ruby to know if this is a good case for an instance variable, but it seems like it?

my thought: since this gets called via `target_veteran` multiple times, there's no need to call the token service after the first time, so cache it in an instance variable

jw81: I think this is fine.
If we didn't immediately raise for a `false` value, then there'd be some concern because on the 2nd/3rd time through, the `||=` would see a "falsey" value in the ivar and try to make the service call again.  But we are immediately raising in the case that the token is invalid, so this should be fine.

jw81:
Looks like the test failures are actually related to this code.  (not random failures from other modules)
Hopefully it'll be as simple as tweaking/adding a mock or two.  🤷

modules/claims_api/spec/requests/v1/claims_request_spec.rb
```diff
@@ -22,6 +24,36 @@
...
+          allow_any_instance_of(TokenValidation::V2::Client).to receive(:token_valid?).and_return(false)
+
+          with_settings(Settings.claims_api.token_validation, api_key: 'some_value') do
+            get '/services/claims/v1/claims', params: nil, headers: request_headers.merge(auth_header)
+            expect(response.status).to eq(403)
+          end
+        end
+      end
+    end
+  end
```
jw81, line 55: Any chance we could move these 2 specs over into the GET POA context?  
Just feels weird that we did this work to directly support GET POA, but the specs are in the GET Claims context.

hinzed1127: I think I did that at first just because the majority of the logic was at the application controller level and so those checks will happen in all of the instances, but that makes sense to me, will do 👍

jw81: Yeah, I'm totally with ya.  I was originally thinking the exact same thing.  It just looked a little out of place once we got to the end.

hinzed1127: Moved [here](https://github.com/department-of-veterans-affairs/vets-api/pull/8362/commits/502f778856d422ff14ba733d272e11335721cf52).

jw81 (approved):

PR #8366, opened 2021-11-03
[30408] Login.gov IAL1 Implementation

bramleyjl:
## Description of change

This change allows users to authenticate with Login.gov as an IAL1 user. Further updates will be made to allow for upcerting, either through a second request or automatically on Login.gov's end. It updates the SAML headers sent to ISAM to optionally include a "minimum" `authn_context_comparison` value that is used for Login.gov requests. SAML header generation is further expanded to take multiple assurance_level urls, a feature that is also used for Login.gov requests. The `logingov_uuid` attribute has been added to the SAML User model and therefore is available on the User model. This change also allows users without an `idme_uuid` to authenticate successfully as long as they possess a `logingov_uuid`.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#30408

Manual signins with Login.gov & non-Login.gov credentials, updated unit tests to cover SAML header changes & incoming Login.gov-backed user attributes.


spec/lib/saml/ssoe_user_spec.rb
```diff
@@ -952,6 +1005,7 @@
           uuid: '85ba80dba1b93ed3bf080b2989cde313',
           email: nil,
           idme_uuid: nil,
+          logingov_uuid: nil,
           loa: { current: 3, highest: 3 },
           sign_in: {
             service_name: 'dslogon',
```
bosawt, line 1011: Can we first change the context description of the test near this: `with missing ID.me UUID`, to `with missing ID.me UUID and missing login.gov UUID`
And then add a test: `with missing ID.me UUID but a defined login.gov uuid`, just to prove that this works with a login.gov uuid even if ID.me uuid is nil

bosawt (commented):
-I confirmed it worked by going to `sessions/v1/logingov/new` and signing in with an existing user
- I also confirmed that the user can be logged out successfully
- Additionally, I tested with an ID.me user, and confirmed that they can successfully log in and log out (proving this has not broken existing functionality)

- My only request is that we add a single spec to account for when the user has a login.gov uuid and not an ID.me uuid, just to prove that it is an acceptable flow now

LindseySaari (approved):

bosawt (approved):
Looks good!

PR #8368, opened 2021-11-03
fix (Jenkinsfile): 'when' -> if

jsgarmon:
## Description of change
See failure at bottom: http://jenkins.vfs.va.gov/job/testing/job/vets-api/job/master/6227/console
The keyword `when` is allowed only inside a stage directive.


LindseySaari:
Since the builds have now passed after the issue (believed to be a GH issue), is this necessary?

jsgarmon:
> Since the builds have now passed after the issue (believed to be a GH issue), is this necessary?

Yes, whenever the build or deploy fails, you'll see the same error when trying to execute the `failure` post block. Likely you haven't seen the error message before because the build rarely fails.

LindseySaari (approved):

PR #8373, opened 2021-11-04
31499 deprecate check_in v1 api

randomsync:
## Description of change

This PR deprecates the check_in v1 api, which is no longer used. It also removes the unused feature flag.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#31499

- [x] rspec
- [x] rubocop


LindseySaari (commented):
What about the v0/apidocs_controller? Do you want to remove the API docs as well or will that come in a separate PR?

modules/check_in/app/controllers/check_in/v1/patient_check_ins_controller.rb
```diff
@@ -4,40 +4,11 @@ module CheckIn
   module V1
     class PatientCheckInsController < CheckIn::ApplicationController
       def show
-        check_in = CheckIn::PatientCheckIn.build(uuid: params[:id])
```
LindseySaari, line 7: I might be a bit of a sticker here too, but wondering if you can remove v1 incrementally to avoid the dangerbot failure. I'm thinking the first iteration can be the changes to the controllers + corresponding specs, then probably the services classes (broken up into separate PRs)

LindseySaari, line 7: I might be a bit of a sticker here too, but wondering if you can remove v1 incrementally to avoid the dangerbot failure. I'm thinking the first iteration can be the changes to the controllers + corresponding specs, then probably the services classes (broken up into separate PRs). I think this one will be pretty straight forward to break apart

randomsync: I agree that it can be done that way but was hoping to avoid having to create 4 or 5 separate PRs, one of which might still go above the 500 line limit. Is there a way to get an exception for this, and if not I'll go ahead and break it up.

randomsync:
> What about the v0/apidocs_controller? Are these the only api docs for all the check_in module versions?

I know we have another engineer working on swagger docs as part of #29594 so we'll be adding more documentation and might be removing this one from `v0`.

modules/check_in/app/controllers/check_in/v1/patient_check_ins_controller.rb
LindseySaari: Yeah lets go ahead and do it... this one makes sense for it to all go out together. Did you see my question about the apidocs_controller?

LindseySaari (approved):

dillo:
@randomsync can we also remove the v1 configuration from the settings.yml and test.yml files?

config/settings.yml
```diff
@@ -1109,29 +1109,6 @@ check_in:
     key_path: fake_api_path
     redis_session_prefix: check_in
     timeout: 30
-  chip_api_v1:
```
kanchanasuriya, line 1112: Wondering removing these keys will cause any build issue since we have the keys defined in [devops repository](https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/deployment/config/vets-api/prod-settings.local.yml.j2#L709-L731).  Please check if it is safe to remove here and remove it devops repository as a fast followup

kanchanasuriya (commented):
Except one comment with key remove in settings.yml, other changes LGTM

config/settings.yml
randomsync: It shouldn't be an issue since those keys aren't being used in code anywhere, but yes they can be removed from devops repo in a separate PR

LindseySaari: +1 to this issue. I don't _think_ there should be an issue since the keys aren't referenced in the code anywhere, but I would remove from the devops repo to be safe

randomsync: Added a PR in devops repo: https://github.com/department-of-veterans-affairs/devops/pull/10236

LindseySaari (approved):

dillo (approved):
LGTM

config/features.yml
```diff
@@ -54,10 +54,6 @@ features:
     actor_type: user
     description: Enables the health care check-in experiences to show demographics page.
     enable_in_development: true
-  check_in_experience_low_authentication_enabled:
-    actor_type: user
-    description: Enables the health care check-in experiences to use the lowest tier of authentication, i.e. last 4 and last name
-    enable_in_development: true
```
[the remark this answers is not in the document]
dillo: Checked with FE and this is safe to remove!

PR #8378, opened 2021-11-04
Api 11017 - SC upload evidence indicated

nathanbwright:
## Description of change
Save if the SC has indicated it will have additional evidence (on the model) and print information on the pdf to indicate additional evidence will be sent.

## Original issue(s)
https://vajira.max.gov/browse/API-11017


ElleeB:
Two items:

1. Neither of the pdfs has marked 'YES' for `noticeAcknowledgement`
2. What's concerning is that everything is passing, but shouldn't the generator test fail when making the pdf comparison?

ElleeB (changes requested):
Two items:

1. Neither of the pdfs has marked 'YES' for noticeAcknowledgement (I'm looking into this since that was a change I made)
2. What's concerning is that everything is passing, but shouldn't the generator test fail when making the pdf comparison?

ElleeB:
Ok, this was my bad. https://github.com/department-of-veterans-affairs/vets-api/blob/e6d18b3c83eb048ca8fed725e864d9f9ce5a4ac5/modules/appeals_api/app/models/appeals_api/supplemental_claim.rb#L180

I'm going to create a bug ticket and take care of this.

ElleeB (approved):
This looks good to me!

modules/appeals_api/app/models/appeals_api/supplemental_claim.rb
```diff
@@ -7,6 +7,16 @@ module AppealsApi
   class SupplementalClaim < ApplicationRecord
     include ScStatus
 
+    EVIDENCE = Struct.new(:type, :attributes) do
```
cilestin, line 10: I think this should live in our module's `lib` folder as a PORO instead of as a constant. Doesn't feel like a constant if you're calling `new` on it, IMO.

Not a blocker, though. We can play with it later.

cilestin (approved):
One non-blocking comment. LGTM!

modules/appeals_api/app/models/appeals_api/supplemental_claim.rb
nathanbwright: Yeah, I thought about that too, I just wasn't sure that it was mature enough for it's own file... I tend to think that if we end up doing anything else to it (adding functionality or otherwise), it should be upgraded to its own file.

ElleeB (approved):
Go go green!

PR #8382, opened 2021-11-04
[32399] splits signup routes for idme & logingov

bramleyjl:
## Description of change

This change splits the `signup` sessions_controller route into two separate routes to accommodate the addition of Login.gov as a CSP alongside ID.me. The ID.me route has been renamed but is unchanged; both routes are intended to ultimately resolve to the same CSP page as their signin route - this page also serves to allow users to create a new account with said CSP.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#32399

Unit tests updated to reflect split in routes, tested ID.me signup route & new Login.gov signup route through account creation & authorization.


AGarcia-Clarity (approved):
LGTM.

Do both the Login.gov & ID.me require the `&op=signup` or just ID.me?

bosawt:
Has the frontend been changed to not call `signup` yet? Otherwise won't this break that until the frontend has changed? Just wondering if we should keep `signup` for now, merge this, get the frontend changes, then remove `signup`?

bosawt:
> LGTM.
> 
> Do both the Login.gov & ID.me require the `&op=signup` or just ID.me?

I believe it should just be ID.me, it sounded like Login.gov didn't have it, I think we can remove it from the login.gov one?

AGarcia-Clarity:
> > LGTM.
> > Do both the Login.gov & ID.me require the `&op=signup` or just ID.me?
> 
> I believe it should just be ID.me, it sounded like Login.gov didn't have it, I think we can remove it from the login.gov one?

I have made some updates on the FE (https://github.com/department-of-veterans-affairs/vets-website/pull/19306) and just need to add that additional query parameter for `?op=signup`.  This merge might be able to wait until FE changes are made.

bramleyjl:
> Has the frontend been changed to not call `signup` yet? Otherwise won't this break that until the frontend has changed? Just wondering if we should keep `signup` for now, merge this, get the frontend changes, then remove `signup`?

I was going to wait 

> LGTM.
> 
> Do both the Login.gov & ID.me require the `&op=signup` or just ID.me?

Just ID.me; fixed this.

lib/saml/url_service.rb
```diff
@@ -92,6 +92,18 @@ def idme_url
       build_sso_url(build_authn_context(LOA::IDME_LOA1_VETS))
     end
 
+    def signup_url
+      @type = 'signup'
+      @query_params[:op] = 'signup'
```
LindseySaari, line 97: The signup_url and idme_signup_url methods look identical. Can they be combined somehow?

bramleyjl: They are, the idme one is replacing the original but we need to preserve both until the frontend actually switches over, then remove `signup_url`.

LindseySaari: Ah ok, sounds good1

bosawt (approved):
Looks good! I agree with Lindsey's comment about the copied function (in fact this whole file should probably be refactored), but given that we'll expect a follow-on PR to remove `signup` imminently, I think that's not an issue

I tested by running through the current idme signup, confirmed that it still worked as expected!

PR #8383, opened 2021-11-04
Mobile location bug fixes

jperk51:
## Description of change
Update location logic to catch missing identifier object from LH and strip facility id down to just the middle 3 digits. 

## Original issue(s)
department-of-veterans-affairs/va-mobile-app#1928
department-of-veterans-affairs/va-mobile-app#1757

Added test to check for 400 response if identifier is missing from LH response. Also updated cassettes to expect 3 digit facility number instead of 7 digit one use previously.


modules/mobile/app/controllers/mobile/v0/locations_controller.rb
```diff
@@ -7,10 +7,16 @@ module V0
...
         Rails.logger.info('Mobile Location', param_id: params[:id])
+
         lh_location = service.get_location(params[:id])
         Rails.logger.info('Mobile Location, Lighthouse Info', lh_location: lh_location)
-        id = lh_location[:identifier].first[:value][4..]
+        if lh_location[:identifier].nil?
+          raise Common::Exceptions::BackendServiceException, 'validation_errors_bad_request'
+        end
+
+        id = lh_location[:identifier].first[:value][4..][0..2]
```
kreek, line 17: Don't know if you could use one range here? e.g. `lh_location[:identifier].first[:value][4..6]` (assuming the id is always in the same spot).

kreek (approved):
One minor comment, but LGTM 🚀

modules/mobile/app/controllers/mobile/v0/locations_controller.rb
jperk51: Yup, that makes more sense

kreek (approved):
LGTM 🚀

PR #8391, opened 2021-11-05
API-11210: BUG/Add default value to notice acknowledgement

ElleeB:
Bug ticket: https://vajira.max.gov/browse/API-11210

This was a miss. We'll always check 'YES' for notice acknowledgement since those with a falsey value will have already been filtered out.


nathanbwright (commented):
It doesn't feel right to me to make this always true in the model, because it's a required field and the model isn't valid unless it's true. I know that we are checking at the controller level, but I'm still uneasy about doing it here. Shouldn't we be doing this at the seed data level (for our specs)?

nathanbwright:
I went back and looked at the seed data (https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/appeals_api/spec/fixtures/valid_200995.json#L21) and it just looks like a name mismatch `data_attributes` vs `dataAttributes`!

ElleeB:
> It doesn't feel right to me to make this always true in the model, because it's a required field and the model isn't valid unless it's true. I know that we are checking at the controller level, but I'm still uneasy about doing it here. Shouldn't we be doing this at the seed data level (for our specs)?

Yeah I went back and forth. I'll go ahead and make it feel better!

ElleeB:
> I went back and looked at the seed data (https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/appeals_api/spec/fixtures/valid_200995.json#L21) and it just looks like a name mismatch `data_attributes` vs `dataAttributes`!

I'll take a look! Thank you :)

modules/appeals_api/app/models/appeals_api/supplemental_claim.rb
```diff
@@ -176,7 +176,7 @@ def new_evidence_dates
     end
 
     def notice_acknowledgement
-      data_attributes['notice_acknowledgement']
```
nathanbwright, line 179: I'm probably missing something here... Is there a reason we need `? 1 : 'Off'` here? Can't it just be true/false?

modules/appeals_api/app/models/appeals_api/supplemental_claim.rb
```diff
@@ -176,7 +176,7 @@ def new_evidence_dates
...
+      data_attributes['noticeAcknowledgement'] ? 1 : 'Off'
```
nathanbwright, line 179: I'm probably missing something here... Is there a reason we need `? 1 : 'Off'` here? Can't it just be true/false?

ElleeB: The value consumed by the structure needs to be 1 or 'Off' if I'm not mistaken.

cilestin (dismissed):

cilestin (approved):

PR #8400, opened 2021-11-08
[30116] updates ID.me auth flow to use minimum authn_context_comparison

bramleyjl:
## Description of change

This changes updates the SAML headers that we send for ID.me logins to use a "minimum" `authn_context_comparison`, which allows us to make an LoA1 request and receive either an LoA1 or LoA3 response depending on the user's maximum LoA3, removing the need to make two separate requests for LoA3 users.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#30116

The impact to `vets-api` from this change is very minimal, with the 'minimum' headers the LoA3 response that comes back functions identically to the one generated by 'verify' auth path that we currently use to upcert LoA1 users to LoA3. Once our Login.gov implementation is also successfully using "minimum" authentication headers to have one-trip IAL2 logins it is likely that some of the upcerting logic can safely be refactored and simplified. 

Unit tests updated to check for 'minimum' authn context comparison header. Manual login/logout flow for ID.me and non-ID.me users tested.


LindseySaari (dismissed):

lib/saml/url_service.rb
```diff
@@ -89,7 +89,7 @@ def dslogon_url
 
     def idme_url
       @type = 'idme'
-      build_sso_url(build_authn_context(LOA::IDME_LOA1_VETS))
+      build_sso_url(build_authn_context(LOA::IDME_LOA1_VETS), 'minimum')
```
bosawt, line 92: Can we set `minimum`, `exact` as constants?

bosawt (dismissed):
Confirmed this worked with ID.me standard login (a single redirect loop)
Confirmed the original functionality was preserved when `vsp_environment` is set to `production` (two redirect loops)

bosawt (approved):

LindseySaari (approved):

PR #8415, opened 2021-11-09
Memoize constants in concerns

ericboehs:
These constants probably shouldn't be defined in a concern as they'll be included multiple times. This fixes the warning that is thrown from doing so.


LindseySaari (approved):
Nice!

PR #8421, opened 2021-11-10
[8420] Adding Temporary middleware environment log for dev environmen…

bosawt:
…t only

## Description of change
This adds a log to inspect the `env` in `SOAPHeaders`, there is some failure around there on dev.va.gov stack only, for some reason. 

## Original issue(s)
department-of-veterans-affairs/vets-api/#8420

- This should not impact prod
- This should only be around until we are able to figure out what's preventing dev.va.gov logins


lib/common/client/middleware/request/soap_headers.rb
```diff
@@ -8,6 +8,7 @@ class SOAPHeaders < Faraday::Middleware
           def call(env)
             env.request_headers['Date'] = Time.now.utc.strftime('%a, %d %b %Y %H:%M:%S GMT')
             env.request_headers['Content-Length'] = env.body.bytesize.to_s
+            Rails.logger "[SOAPHeaders] Environment: #{env}" if Settings.test_user_dashboard.env == 'dev'
```
LindseySaari, line 11: I think you may be able to use the [Settings.vsp_environment yaml setting](https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/deployment/config/vets-api/dev-settings.local.yml.j2#L2) here, but `== 'dev'` will need to be `== 'development'` It doesn't appear that the test_user_dashboard setting is defined in [the dev settings in the devops repo](https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/deployment/config/vets-api/dev-settings.local.yml.j2)

LindseySaari: Just confirmed, TUD is only available in staging and only has test_user_dashboard.env set in the [staging yml](https://github.com/department-of-veterans-affairs/devops/blob/f7d2dd14c05d85c6538c748c4953a9517267c6c3/ansible/deployment/config/vets-api/staging-settings.local.yml.j2#L519)

LindseySaari (changes requested):
Just a couple comments inline

lib/common/client/middleware/request/soap_headers.rb
bosawt: ah yes, you're right. I should have checked. I made this update

bosawt:
This was resolved so the log is not necessary

PR #8423, opened 2021-11-10
Fix services config

bradbergeron-us:
## Description of change

Removing the betamocks setting entirely fixes the problem locally for Lighthouse SLA environment.

## Original issue(s)


dbrown-va (approved):
This appears to fix the problem running locally and fetching Betamocks configured responses.

PR #8425, opened 2021-11-10
API-9841 Single source of truth for rswag schemas

ElleeB:
Attempt to use our existing config schemas for our rswag generation - removing repetition and noise in the rswag_config file.

Also included for comment are updates to 200995 schema and fixtures. Upon generating the schema via this work, it was discovered that neither the `noticeAcknowledgement values`, nor the `issues`, were generating for supplemental claims.

_this pr will eventually be broken into smaller prs_


ElleeB:
closing draft in favor of three separate PRs to address PR length.

PR #8426, opened 2021-11-10
Add Sidekiq job for db record rotation

LindseySaari:
## Description of change
This PR adds a Sidekiq job for rotating/re-encrypting db records. With out current calculations, it would take nearly 5.5 straight days to do this work via ssm/rails console manually on top of experiencing ssm timeouts as well. Sidekiq weighed out to be the best option and if we're able to run 5 jobs, the rotation should take ~1day. While running the jobs, it'll be best practice to monitor Grafana charts for Sidekiq and RDS CPU spikes.

Benchmarking to split up the jobs evenly across 5 jobs will be the next step

## Original issue(s)
department-of-veterans-affairs/va.gov-team#31630


app/workers/kms_record_rotation_job.rb
```diff
@@ -0,0 +1,24 @@
...
+      model.descendants.empty? && model.try(:lockbox_attributes) && !model.lockbox_attributes.empty?
+    end
+
+    models.map(&:constantize).each do |m|
+      encrypted_attributes = m.lockbox_attributes.keys
+      Rails.logger.debug "re-encrypting model..... #{m}
+                          Total unencrypted records: #{m.where(encrypted_kms_key: nil).count}"
+
+      Lockbox.rotate(m.where(encrypted_kms_key: nil), attributes: encrypted_attributes)
+    end
```
ericboehs, line 20:
```suggestion
  def perform(models = ApplicationRecord.descendants.map(&:name))
    HealthQuest::QuestionnaireResponse.skip_callback(:save, :before, :set_user_demographics, raise: false)

    models
      .map(&:constantize)
      .select { |model| model.descendants.empty? && model.try(:lockbox_attributes) && !model.lockbox_attributes.empty? }
      .each do |m|
        encrypted_attributes = m.lockbox_attributes.keys

        Rails.logger.debug "Re-encrypting model: #{m}"
        Rails.logger.debug "Total unencrypted records: #{m.where(encrypted_kms_key: nil).count}"
        Lockbox.rotate(m.where(encrypted_kms_key: nil), attributes: encrypted_attributes)
      end
```

ericboehs (approved):

PR #8429, opened 2021-11-10
API-11252 Make match_pdf work with checkboxes

cilestin:
## Description of change
Updates the `match_pdf` rspec matcher to validate checkboxes being checked/unchecked as part of its matching. 

Before this change, the matcher would only work against the text of the pages and wouldn't recognize the state of a checkbox.

## Original issue(s)
https://vajira.max.gov/browse/API-11252

* I had to update the supplemental claims expected pdfs since a previous change had checked a checkbox that we had formerly been unable to catch.


modules/appeals_api/spec/support/pdf_matcher.rb
```diff
@@ -4,13 +4,18 @@
 
 RSpec::Matchers.define :match_pdf do |expected|
   match do |actual|
+    # Write-to-write, there may be small whitespace differences that still render exactly the same
+    normalize_page_content = lambda do |pages|
+      pages.map(&:raw_content).join("\n").lines.map(&:strip).map { |l| l.gsub(/\s+/, ' ') }.join("\n")
```
nathanbwright, line 9: Curious why you're  not just smushing everything together and gsubing the resulting string to deal with changing whitespace...? Something like:

```
pages.map(&:raw_content).join.gsub(/\s+/, '')
```

Or am I missing something?

cilestin: Probably me just iterating over the different issues and not thinking about how to consolidate. I ended up going a different direction with it regardless, because the CI generates the raw_content differently to where all the gsubs in the world probably won't be worth it :(

nathanbwright (approved):

PR #8432, opened 2021-11-11
[32757] splits saml_callback verify_url logic

bramleyjl:
## Description of change

This change splits the logic for generating verification SAML headers based on whether the call is coming from a user that wishes to upcert their account or the automatic upcert process that occurs with Login.gov and ID.me logins. `saml_callback`-derived calls now properly use the `type` attribute that vets-api sets and returns to itself to identify which CSP the headers should be targeting. It is a prerequisite for the Login.gov implementation by removing a blocker on IAM's end to generate an artificial authn_context attribute to match our parsing, as is currently done with ID.me. After Login.gov is implemented we plan to return to this method to further remove the dependency on the authn_context attribute.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#32757

This PR also contains a small change to the `build_authn_context` method: the `identity_provider` argument is defaulted to ID.me and the method is never called with it set to a falsey value, so the conditional for it can safely be removed.

Unit tests updated, login with LoA1/3 ID.me, DSLogon, and verification for LoA 1 ID.me users performed.


bosawt (approved):
I think this is good. I did a deeper dive on this since it basically touches all logins.
- Appears to return the expected authn_contexts for idme/mhv/dslogon upleveling
- I did not test inbound, it would be worth testing on staging and watching metrics when it goes love on production

bosawt:
Also, once we can confirm this is working as expected, we should remove some of the original link_authn_context stuff in `verify_url`

PR #8443, opened 2021-11-15
1930 vaccine group name

kpethtel:
## Description of change
This is only a draft PR. When this is code complete, the migration will be separated into its own PR.

## Original issue(s)

No yet, just a draft.


db/migrate/20211112213456_create_vaccines.rb
```diff
@@ -0,0 +1,10 @@
+class CreateVaccines < ActiveRecord::Migration[6.1]
```
kpethtel, line 1: If I understand this project correctly, we could also include the manufacturer as an attribute on Vaccine. Chika tells me that we're only including manufacturer for covid-19 vaccines at this time, so there's no concern about multiple manufacturers.

modules/mobile/app/services/mobile/v0/vaccinations/service.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
```
kpethtel, line 1: If we added manufacturer info, it could also be added here by pulling it from the other xml file.

modules/mobile/app/services/mobile/v0/vaccinations/service.rb
```diff
@@ -0,0 +1,34 @@
...
+    module Vaccinations
+      # Service that connects to CDC's vaccinations XML table
+      # https://www2.cdc.gov/vaccines/iis/iisstandards/XML.asp?rpt=vax2vg
+      #
+      class Service
+        doc = Nokogiri::XML(URI.open('https://www2.cdc.gov/vaccines/iis/iisstandards/XML.asp?rpt=vax2vg')) do |config|
+          config.strict.noblanks
+        end
+
+        doc.root.children.each do |node|
```
kpethtel, line 14: I wasn't able to figure out how to fetch this data purely by xpath.

app/models/vaccine.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+class Vaccine < ApplicationRecord
```
kreek, line 3: Model looks great, could we move it and the migration into our mobile module? Apologies if I misled you on this, Jayson had a PR go in recently that added the code needed for migrations to work from there. 

New paths would be:
`modules/mobile/app/models/mobile/v0/vaccine.rb` and
`modules/mobile/db/migrate/20211112213456_create_vaccines.rb`

db/migrate/20211112213456_create_vaccines.rb
kreek: As above this can move to the mobile module dir.

RE: potentially adding a manufacturer column. I think it's safe to add manufacturer here for now as we control the timing of the update and could migrate data later if we needed a `has_many` relationship (and a second table).

modules/mobile/app/services/mobile/v0/vaccinations/service.rb
kreek: I think this way fine, it's easy to read 👍

modules/mobile/app/services/mobile/v0/vaccinations/service.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Mobile
+  module V0
+    module Vaccinations
+      # Service that connects to CDC's vaccinations XML table
+      # https://www2.cdc.gov/vaccines/iis/iisstandards/XML.asp?rpt=vax2vg
+      #
+      class Service
```
kreek, line 9: Could this logic either move to a background job or be called from a background job? That way it can be scheduled to run once a day. Schedules are set in the [sidekiq_scheduler.yml](https://github.com/department-of-veterans-affairs/vets-api/blob/00f79f734be15295546d877f495335c647b16690/config/sidekiq_scheduler.yml#L226-L229) file. And then the job would be in a similar location (but in our module) to this [batch job](https://github.com/department-of-veterans-affairs/vets-api/blob/5b880ce76c1e33d6923cf6b6eaf23617fc571980/modules/covid_vaccine/app/jobs/covid_vaccine/scheduled_batch_job.rb).

kreek (commented):
A couple comments but looks good so far. If @cadibemma's manufacturer work can use the same table/migration maybe we pull that out into a separate PR.

app/models/vaccine.rb
kpethtel: Yes, and thank you for clearing this up. It wasn't clear to me if this should go in the base level or not. Given that there didn't seem to be any database backed models in the mobile module, I left this at the base level as a starting point.

modules/mobile/app/services/mobile/v0/vaccinations/service.rb
kpethtel: That's a good point. I was planning to put it on the scheduler, but I didn't think about the fact that scheduled jobs should be async.

PR #8447, opened 2021-11-15
API-9841: Add sc create schema handling

ElleeB:
Refactored rswag_config:

1. Updated `#sc_create_schemas`
2. Modified `200995.json` and fixtures after discovering missing schema on dev portal

Ticket: https://vajira.max.gov/browse/API-9841

Will fail until this ticket merged: #8444


modules/appeals_api/config/schemas/v2/200995.json
```diff
@@ -123,25 +123,15 @@
...
-                    },
-                    { "required": ["retrieveFrom"] }
-                  ]
-                },
-                "noticeAcknowledgement": {
-                  "enum": [true, false]
+                  "if": { "properties": { "evidenceType": { "const": "retrieval" } } },
+                  "then": { "required": ["retrieveFrom"] }
                 },
+                "noticeAcknowledgement": { "type": "string", "enum": ["yes", "no"] },
```
cilestin, line 129: I don't think this should have been an enum to begin with - probably just an oversight. Should just be a `type: boolean` instead.

`"noticeAcknowledgement":  { "type": "boolean" }`

modules/appeals_api/config/schemas/v2/200995.json
```diff
@@ -123,25 +123,15 @@
...
                 "socOptIn": { "type": "boolean" }
               },
               "required": ["veteran", "evidenceSubmission", "socOptIn", "noticeAcknowledgement"],
               "if": { "properties": { "benefitType": { "const": "compensation" } } },
-              "then": { "properties": { "noticeAcknowledgement": { "const": true } } }
+              "then": { "properties": { "noticeAcknowledgement": { "const": "yes" } } }
```
cilestin, line 134: Does the const stuff fail when it's set to `false` after you update the acknowledgement to type boolean?

modules/appeals_api/config/schemas/v2/200995.json
```diff
@@ -155,7 +145,7 @@
               "attributes": {
                 "type": "object",
                 "properties": {
-                  "issue": { "allOf": [ { "$ref": "#/definitions/nonBlankString" }, { "maxLength": 140 } ] },
+                  "issue": { "$ref": "#/definitions/nonBlankString", "maxLength": 140 },
```
cilestin, line 148: I'm not sure this is a valid way to combine the requirements. Did you generate the docs and view it with the swagger ui? I feel like I tried something similar a bit ago and it wasn't happy about it 💥

modules/appeals_api/spec/requests/v2/supplemental_claims_controller_spec.rb
```diff
@@ -70,9 +70,9 @@ def base_path(path)
     end
 
     context 'noticeAcknowledgement' do
-      it 'benefitType = compensation and noticeAcknowledgement = false' do
+      it "benefitType = compensation and noticeAcknowledgement = 'no'" do
```
cilestin, line 73: This ripple effect of changing the type really isn't feeling like the right solution, and goes against the fact that there's already a type (boolean) that takes care of this for us.

If this change was necessary to make rswag happy, then it should be done at the rswag conversion level - we shouldn't be breaking schema conventions just to appease our testing tool.

cilestin (changes requested):

modules/appeals_api/config/schemas/v2/200995.json
ElleeB: yes, I believe that was the problem and part of the reason I changed it back after trying to set it to a boolean. I'll revisit.

modules/appeals_api/config/schemas/v2/200995.json
ElleeB: I believe combining the requirements like this is fine, but I could def be wrong. In this instance, however, the issues weren't generating so I copied and pasted what was in NOD since it works over in those parts!

modules/appeals_api/spec/requests/v2/supplemental_claims_controller_spec.rb
ElleeB: The answer is yes, but it wasn't a boolean before, either.

I'll go back and try again with a boolean, since it really feels like it should work. Also, just a note, not that this changed my opinion about the boolean value, but the form is two distinct checkboxes - one for yes, and one for no

modules/appeals_api/config/schemas/v2/200995.json
cilestin: Hmm well since everything seems fine over in NOD land, I guess whatever explosions happened to me were for a different reason. Let's keep this as-is - it's nicer anyway 👍

cilestin: Thinking on it a bit more, maybe what I'm remembering was actually sending in more than the `maxLength` number of characters and it not catching it as invalid. I'll confirm.

modules/appeals_api/config/schemas/v2/200995.json
ElleeB: Resolved!

modules/appeals_api/config/schemas/v2/200995.json
ElleeB: Ah, ok. I can look into it!

modules/appeals_api/spec/requests/v2/supplemental_claims_controller_spec.rb
ElleeB: Resolved!

modules/appeals_api/config/schemas/v2/200995.json
ElleeB: Ok, at first glance it _does not_ appear to be catching it. I'll fix.

cilestin (approved):

PR #8448, opened 2021-11-15
API-1311 Add totals to AppealsAPI weekly report

cilestin:
## Description of change
Adds total sum of various appeals to our weekly report.

## Original issue(s)
https://vajira.max.gov/browse/API-11311


modules/appeals_api/app/views/appeals_api/decision_review_mailer/mailer.html.erb
```diff
@@ -42,26 +42,26 @@
   <body>
     <h1><%= @friendly_duration %> <%= @friendly_env %> Decision Review Report from <%= @report.from %> -- <%= @report.to %> </h1>
 
-    <%= render partial: 'appeal_table', locals: { appeal_name: 'HLR', appeal_status_and_count: @report.hlr_by_status_and_count, faulty_appeals: @report.faulty_hlr } %>
+    <%= render partial: 'appeal_table', locals: { appeal_name: 'HLR', appeal_status_and_count: @report.hlr_by_status_and_count, faulty_appeals: @report.faulty_hlr, total_count: @friendly_duration == 'Weekly' ? @report.total_hlr_successes : 0 } %>
```
nathanbwright, line 45: Not really a fan of inline conditionals in the views, but I suppose we can let it slide in this time... Can we just set `total_count` to nil instead of 0 and then we could simplify the logic in the appeal_table partial?

modules/appeals_api/spec/lib/decision_review_report_spec.rb
```diff
@@ -49,6 +49,14 @@
     end
   end
 
+  describe '#total_hlr_successes' do
+    it 'shows correct count of all successful HLRs regardless of timeframe' do
+      create_list :higher_level_review, 5, created_at: 3.weeks.ago
```
nathanbwright, line 54: TIL I leaned about `create_list`! Cool! 🎉

modules/appeals_api/app/views/appeals_api/decision_review_mailer/mailer.html.erb
cilestin: Agreed on the inline conditionals :( Updated for nil checks. Thanks!

nathanbwright (approved):
🚀

PR #8451, opened 2021-11-16
Mpm/legacy appeals ep

michelpmcdonald:
## Description of change
Add new Lighthouse Legacy Appeals endpoint to Contestable Issues index.  Since Legacy Appeals  and Contestable Issues are closely related, FE requested that the new Legacy Appeal endpoint data is merged into the existing Contestable Issues route. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#31149

Hidden behind the hlr_2 feature flag.  Testing will be done as part of the overall plan to move HLR 2 into production


app/controllers/v1/higher_level_reviews/contestable_issues_controller.rb
```diff
@@ -15,6 +16,31 @@ def index
...
+          ci_la = { "data": contestable_issues['data'] + la['data'] }
+        rescue => e
+          # If LA fails keep going Legacy Appeals are not critical, return original contestable_issues
+          log_exception_to_personal_information_log(
+            e,
+            error_class: "#{self.class.name}#index exception #{e.class} (HLR_V1_LEGACY_APPEALS)",
+            benefit_type: params[:benefit_type]
+          )
+          contestable_issues
+        else
```
michelpmcdonald, line 40: Eat-log any exceptions.  The idea here is that since this is a new endpoint, I don't want to prevent users from creating HLRs if the endpoint OR the consuming code has an issue.  

Currently, the Legacy Appeals that the new endpoint provide are not critical to creating HLRs.  Users with no Legacy Appeals can skip the Legacy Appeal Opt in question on the HLR form wizard in the website, so Legacy Appeals are not absolutely required to create an HLR.  The FE team is on board with this and can deal with a missing "legacyAppeals" section in the rendered JSON

ericboehs (approved):

PR #8469, opened 2021-11-17
[33057] uses ial over credentialassurancelevel for Login.gov

bramleyjl:
## Description of change

This change fixes the broken Login.gov logins that occurred after ISAM dropped the `va_eauth_credentialassurancelevel` from their Login.gov SAML responses, vets-api will now look for the `va_eauth_ial` attribute for Login.gov logins and use that to set a user's LoA within vets-api.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#33057

## Things to know about this PR

Unit spec factories update to no longer provide `credentialassurancelevel`, unit specs pass. Manual Login.gov login and ID.me logins function correctly and users have proper internal LoAs.

PR #8471, opened 2021-11-18
[33050] removes attributes request from Login.gov SAML request headers

bramleyjl:
## Description of change

Removes unneeded attribute request SAML generation for Login.gov - the `email` and `verifiedAt` headers are now always provided in ISAM's response to vets-api.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#33050

Login.gov IAL1 logins (the only affected) tested & function properly (with the yet-unmerged IAL fix: https://github.com/department-of-veterans-affairs/vets-api/pull/8469) - this change is prompted by an external data change and therefore cannot be properly unit tested.


thilton-oddball (approved):

bosawt (approved):
Confirmed that email and verified_at come back even without them being specifically requested.

PR #8472, opened 2021-11-18
32943 Precheck-in controller routes

kanchanasuriya:
## Description of change

Adding routes and controller for PreCheck-in

## Original issue(s)
department-of-veterans-affairs/va.gov-team#32943

## Testing
- [X] rspec
- [X] rubocop


modules/check_in/app/controllers/check_in/v2/pre_check_in_controller.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module CheckIn
+  module V2
+    class PreCheckInController < CheckIn::ApplicationController
```
randomsync, line 5: Rails controller names are plural by convention: https://guides.rubyonrails.org/action_controller_overview.html#controller-naming-convention

modules/check_in/app/controllers/check_in/v2/pre_check_in_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+        head :not_implemented
+      end
+
+      def create
+        head :not_implemented
+      end
+
+      private
+
+      def session_params
```
randomsync, line 16: Suggest changing this to pre_check_in_params to align with controller name

modules/check_in/config/routes.rb
```diff
@@ -19,5 +19,6 @@
   namespace :v2, defaults: { format: :json } do
     resources :patient_check_ins, only: %i[show create]
     resources :sessions, only: %i[show create]
+    resources :pre_check_in, only: %i[show create]
```
randomsync, line 22: same as above, resource name by convention would be `:pre_check_ins`

modules/check_in/spec/request/v2/pre_check_in_request_spec.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'V2::PreCheckInController', type: :request do
+  let(:id) { '2f90aa72-deba-4bd7-9bd6-5d17c70bb1b3' }
+  let(:memory_store) { ActiveSupport::Cache.lookup_store(:memory_store) }
```
randomsync, line 7: minor: probably not needed at this point.

randomsync (changes requested):

modules/check_in/app/controllers/check_in/v2/pre_check_in_controller.rb
kanchanasuriya: Interesting to know about the rails convention.  Thanks for the reference link.

modules/check_in/app/controllers/check_in/v2/pre_check_in_controller.rb
kanchanasuriya: Sure, modified the method name.

modules/check_in/config/routes.rb
kanchanasuriya: 👍

modules/check_in/spec/request/v2/pre_check_in_request_spec.rb
kanchanasuriya: Sure Gaurav, we might need when we add specs for verifying funcitonality.  I will leave it as it is for now.

randomsync (approved):
lgtm

PR #8473, opened 2021-11-18
[30116] Enable ID.me 'minimum' authn context comparison on vets-api production

bramleyjl:
--DO NOT MERGE UNTIL 0700 ET 19NOV21--

## Description of change

This change removes the guard clause preventing the `minimum` authn_context_comparison headers from being enabled on production. When this is merged into `master` the single-trip LoA3 login for ID.me users will go live.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#30166

Unit tests were updated in the main ID.me minimum authn context comparison headers PR: https://github.com/department-of-veterans-affairs/vets-api/pull/8400
The removed tests were not replaced because there is no longer any LoA upleveling as of this PR. IAL upleveling exists, and identical tests for Login.gov/IAL can be found in the same spec file.
ID.me logins were performed: on anything lower than production there is no noticeable change.


lib/saml/post_url_service.rb
```diff
@@ -41,7 +41,7 @@ def initialize(saml_settings, session: nil, user: nil, params: {}, loa3_context:
     def should_uplevel?
       # handles uplevel checks for ID.me & Login.gov 'two-trip' logins
       # verified_at is a Login.gov atttribute indicating the presence of IAL2 status
-      user.verified_at.present? ? (user.loa[:current] < LOA::THREE) : (user.loa[:current] < user.loa[:highest])
+      user.verified_at.present? && user.loa[:current] < LOA::THREE
```
bosawt, line 44: Can we change the comment here as well, to reflect that this is no longer relevant with ID.me?

joeniquette (dismissed):
please do not merge this before 0700 19NOV21. This also is pending a change being made by IAM tonight during their deploy. If that fails or is aborted, this change cannot be merged. Check with me before merging.

bosawt (approved):
I just had a comment to update the comment in should_uplevel to reflect the new behavior. It seems scary to remove upleveling code but I honestly couldn't think of a use case where it'd still be relevant, so I think this is good to go

joeniquette (approved):

ericboehs (approved):

PR #8474, opened 2021-11-18
[33058] switches verify_url Login.gov check to correct LOGIN_GOV_IAL1 value

bramleyjl:
## Description of change

This change will enable user-initiated verification calls to properly function on non-localhost environments by correctly setting the expected Login.gov `authn_context` value to that returned by IAM. Conversely, manual verification will be broken on localhost until ISAM is updated to send the correct `authn_context` there as well.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#33058

We have independently confirmed that the new value is being sent by ISAM to non-localhost environments, localhost was tested and no longer works (as expected).


bosawt (changes requested):
I want to wait before we merge this, until localhost SAML Response from IAM changes behavior. Otherwise we just get stuck on a loop and can't log in with localhost

bramleyjl:
> I want to wait before we merge this, until localhost SAML Response from IAM changes behavior. Otherwise we just get stuck on a loop and can't log in with localhost

This shouldn't break upleveling on login, since that logic gets abstracted out in the `callback_verify_url` above. It *would* break manual upleveling of an IAL1 Login.gov account on localhost but that doesn't like a major issue? Regardless, you're correct that the best solution would be to wait for IAM to get their `authn_context` on localhost to align with the values on the environments before changing, I'm just not sure how long that might take.

bosawt:
> > I want to wait before we merge this, until localhost SAML Response from IAM changes behavior. Otherwise we just get stuck on a loop and can't log in with localhost
> 
> This shouldn't break upleveling on login, since that logic gets abstracted out in the `callback_verify_url` above. It _would_ break manual upleveling of an IAL1 Login.gov account on localhost but that doesn't like a major issue? Regardless, you're correct that the best solution would be to wait for IAM to get their `authn_context` on localhost to align with the values on the environments before changing, I'm just not sure how long that might take.

I was logging in with this PR and seeing an infinite loop. But I'll test again and see, there were a lot of PRs in flight yesterday, very possible I had the wrong branch or some change that shouldn't have been there

bosawt (approved):
Weird, I tested again and didn't see the infinite loop on localhost, so perhaps something else was going on. In that case, looks good, haha

PR #8480, opened 2021-11-18
Update mobile immunizations adapter to handle blank notes

kreek:
## Description of change
Immunization note fields that are null or an empty array are causing an error in production. Log line was found with the [following query](http://grafana.vfs.va.gov/explore?orgId=1&left=%5B%22now-24h%22,%22now%22,%22Loki%20(Prod)%22,%7B%22expr%22:%22%7B%20app%3D~%5C%22vets-api-server%5C%22,%20filename%3D~%5C%22.%2Bjson.log%5C%22%20%7D%20%7C~%20%5C%22adapters%2Fimmunizations.rb%5C%22%22%7D%5D):

```{ app=~"vets-api-server", filename=~".+json.log" } |~ "adapters/immunizations.rb"```

The fix catches blank note arrays (either nil or empty) and returns nil in that case. 

## Original issue(s)
department-of-veterans-affairs/va-mobile-app/2038


jperk51 (approved):
LGTM

PR #8481, opened 2021-11-18
WIP Add/Update Primary key for veteran_representatives table

LindseySaari:
DO NOT MERGE: Breaking out DB changes to a separate PR [found here](https://github.com/department-of-veterans-affairs/vets-api/pull/8485)

This change adds a primary key to the veteran_represenatives table. It appears that the current implementation allowed for a null primary key. It was noticed that the primary key (representative_id) was blank in a handful of records across all env databases. This caused issues during the key rotation wrt KMS because the primary key is used to find a record and update its respective encrypted_kms_key field. Errors were thrown on records that had a nil representative_id. 

This change still requires the representative_id to be unique but uses ‘id’ for it’s new primary key, which will never be null. Throughout the code, records are created using find_or_initalize_by (not plain find/create), so this change doesn’t break the current implementation. 

The spec had to be updated because the `Veteran::Service::Representative` model is now ordered by id rather than representative_id. Before `Veteran::Service::Representative.last` returned Kevin Allen (as seen in the VCR cassette because his representative_id was the highest. 40393 vs 42535). Since James Alston is created last ([see cassette](https://github.com/department-of-veterans-affairs/vets-api/blob/master/spec/support/vcr_cassettes/veteran/ogc_claim_agent_data.yml#L43-L56)), the test now reflects the correct ordering by id.

### DIFF
--- a/db/migrate/20211118154042_add_primary_id_to_veteran_representatives_table.rb
+++ b/db/migrate/20211118154042_add_primary_id_to_veteran_representatives_table.rb
@@ -1,0 +1,5 @@
+class AddPrimaryIdToVeteranRepresentativesTable < ActiveRecord::Migration[6.1]
+  def change
+    add_column :veteran_representatives, :id, :primary_key
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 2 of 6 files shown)


LindseySaari:
Closing in favor of just editing the data + adding a not null constraint to the rep_id after speaking with @SlakrHakr

PR #8483, opened 2021-11-18
Feature dgi claimant info

bradbergeron-us:
## Description of change

This PR is still incomplete in terms of what this endpoint is attempting to accomplish but we are at a point where we need to get some guidance upon review. The goal is to transition our mocked claimant information service to be dynamic based on sending the users ssn in the body of the the request. Right now the the third party service that is being called only has a single user in the local database with the { "ssn": "539139735" }. This is being passed in the body as a hardcoded JSON parameter. We are looking for guidance on how we can pass in the user.ssn parameter value dynamically. We have made several attempts in development but our attempts have been failing because of errors regarding nil values relating to the user. Additionally we need some guidance on the response class which has some rubocop issues relating to method length and complexity. 
## Original issue(s)

Swagger Docs will be updated in a follow up PR.
Settings.yml was altered from our initial development testing the staging server to the docker url running our service in development.
PII questions see description surrounding sending the user's ssn dynamically n the post request to our 3rd party endpoint.


modules/meb_api/lib/dgi/automation/service.rb
```diff
@@ -27,9 +31,7 @@ def end_point
         end
 
         def json
-          # Passes Back User SSN in the Re body
-          # { "ssn": '539139735' }
-
+          # { "ssn": "539139735" }
```
bradbergeron-us, line 34: Instead of passing this param in the controller ideally we would pass in the user's ssn dynamically from this service class. Any feedback on how we might be able to accomplish is welcomed.

modules/meb_api/lib/dgi/automation/claimant_response.rb
```diff
@@ -0,0 +1,54 @@
...
+        attribute :city, String
+        attribute :zipcode, String
+        attribute :address_type, String
+        attribute :email_address, String
+        attribute :mobile_phone_number, String
+        attribute :home_phone_number, String
+        attribute :country_code, String
+        attribute :state_code, String
+
+        def initialize(status, response = nil)
```
bradbergeron-us, line 28: This method is causing Rubocop errors regarding complexity and length. Some thoughts we've had are that because the data is already being massaged in the third party service that we are calling and the response has so many attributes that it would be ideal if we didn't have to fetch each value and we could just dump the entire json response. Is there an existing pattern we can follow to adopt this approach or what is recommended as far as best practice here?

ericboehs: I think `dig` would help with complexity here. And for method length, would it work to extract that hash to an ` attributes ` method?

ericboehs (dismissed):

rileyanderson (approved):

PR #8485, opened 2021-11-18
break out migration and schema to separate PR

LindseySaari:
## Description of change
This change adds a primary key to the veteran_represenatives table. It appears that the current implementation allowed for a null primary key. It was noticed that the primary key (representative_id) was blank in a handful of records across all env databases. This caused issues during the key rotation wrt KMS because the primary key is used to find a record and update its respective encrypted_kms_key field. Errors were thrown on records that had a nil representative_id.

The subsequent PR can be found [here](https://github.com/department-of-veterans-affairs/vets-api/pull/8481) but the db changes were broken apart from the first PR to ensure app code be backwards compatible with the db.

## Original issue(s)


LindseySaari:
Closing in favor of just editing the data + adding a not null constraint to the rep_id after speaking with @SlakrHakr

PR #8487, opened 2021-11-18
2043 mobile handle empty cvx and dates immunizations

jperk51:
## Description of change
The mobile API should return 200s if cvx or vaccine date are missing

## Original issue(s)
department-of-veterans-affairs/va-mobile-app#2043

Tests added for handling potentially missing data


kreek (commented):
This LGTM. Before merging I want to double check if we want this functionality in the initial MVP vaccine release or the 2.0 release. @narin is the version of the app for the 23rd able to handle null `groupName` or `date` fields?

kreek:
@jperk51 got word from Ashely that we can wait for the design and the front-end work to be done for this before merging.

PR #8492, opened 2021-11-19
Add null check constraint

LindseySaari:
## Description of change
IMPORTANT: #8493  and #8497 need to be merged first. /db and app code changes cannot go out in the same PR. Specs will fail until I merge #8493 and #8497 into master and rebase

Tried to use `change_column_null :veteran_representatives, :representative_id, false`
in this migration, but this is considered a dangerous operation

=== Dangerous operation detected #strong_migrations ===

Setting NOT NULL on an existing column blocks reads and writes while every row is checked.
Instead, add a check constraint and validate it in a separate migration.

```
class ValidateAddNotNullConstraintToVetRepTable < ActiveRecord::Migration[6.1]
  def change
    add_check_constraint :veteran_representatives, "representative_id IS NOT NULL", name: "veteran_representatives_representative_id_null", validate: false
  end
end
```

```
class ValidateValidateAddNotNullConstraintToVetRepTable < ActiveRecord::Migration[6.1]
  def change
    validate_check_constraint :veteran_representatives, name: "veteran_representatives_representative_id_null"
  end
end
```

This change adds a primary key to the veteran_represenatives table. It appears that the current implementation allowed for a null primary key and it was noticed that the primary key (representative_id) was blank in a handful of records across all env databases. This caused issues during the key rotation wrt KMS because the primary key is used to find a record and update its respective encrypted_kms_key field. Errors were thrown on records that had a nil representative_id.

@SlakrHakr can you update the existing (bad) representative_id values, otherwise I think that the records with the null rep_id's will fail

## Original issue(s)


SlakrHakr (dismissed):

SlakrHakr (dismissed):

SlakrHakr:
> can you update the existing (bad) representative_id values, otherwise I think that the records with the null rep_id's will fail

i'll go update those in deployed environments now and update here when complete

db/migrate/20211119153306_validate_add_not_null_constraint_to_vet_rep_table.rb
```diff
@@ -0,0 +1,5 @@
+class ValidateAddNotNullConstraintToVetRepTable < ActiveRecord::Migration[6.1]
+  def change
+    add_check_constraint :veteran_representatives, "representative_id IS NOT NULL", name: "veteran_representatives_representative_id_null", validate: false
+  end
+end
```
ericboehs, line 5: Let's get new line chars here pls.

ericboehs (approved):

PR #8500, opened 2021-11-19
Reworking how facilities_api handles PPMS pagination

darkbushido:
## Description of change

PPMS gave us some new endpoints that support pagination
this pr removes all of the pagination hacks we have used in the past.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#32812


config/initializers/statsd.rb
```diff
@@ -203,8 +203,17 @@
...
 
       if params['radius']
+        count = case payload.body
+                when Hash
+                  payload.dig(:body, :value) || 0
+                when Array
+                  payload.body&.count || 0
+                else
+                  0
+                end
```
ericboehs, line 213:
```suggestion
        count =
          case payload.body
          when Hash
            payload.body[:value]
          when Array
            payload.body.count
          end
```

config/initializers/statsd.rb
```diff
@@ -203,8 +203,17 @@
...
+
         tags << "facilities.ppms.radius:#{params['radius']}"
-        tags << "facilities.ppms.results:#{payload[:body].is_a?(Array) ? payload[:body]&.count : 0}"
+        tags << "facilities.ppms.results:#{count}"
```
ericboehs, line 216:
```suggestion
        tags << "facilities.ppms.results:#{count || 0}"
```

config/initializers/statsd.rb
ericboehs: What do you think of moving the multiple `|| 0` down to where it's accessed?

ericboehs (approved):
Small refactor suggestion.

config/initializers/statsd.rb
darkbushido: I'm worried about the & case

config/initializers/statsd.rb
darkbushido: missed this, I'll give it a shot

rileyanderson (approved):

PR #8504, opened 2021-11-19
[33130] adds logingov_uuid parsing from GCIDs

bramleyjl:
## Original issue(s)
department-of-veterans-affairs/va.gov-team#33130


ericboehs (approved):

lib/identity/parsers/gc_ids_constants.rb
```diff
@@ -34,6 +34,9 @@ module GCIdsConstants
       # IDME_ID_REGEX, ex. 54e78de6140d473f87960f211be49c08^PN^200VIDM^USDVA^A
       IDME_ID_REGEX = /^\w+\^PN\^200VIDM\^USDVA\^A$/.freeze
 
+      # LOGINGOV_ID_REGEX, ex. aa478abc-e494-4af1-9f87-d002f8fe1cda^PN^200VLGN^USDVA^A
+      LOGINGOV_ID_REGEX = /^[\w-]+\^PN\^200VLGN\^USDVA\^A$/.freeze
```
bosawt, line 38: Is logingov the only field in the MPI Response that has dashes?

bosawt: I just looked, I guess it is. It looks consistent with the eauth header so I suppose that's what we want

bosawt (approved):
Looks good! I originally had a comment but it looks like it's not an issue

PR #8505, opened 2021-11-22
1930 add vaccines

kpethtel:
## Description of change
Adds a vaccine model with group name and manufacturer. Also adds a cron script that fetches vaccine data from the CDC and creates vaccine models from it. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#1929

This is a draft PR and the data migration still needs to be separated into its own PR.

Testing: this has been tested by running the VaccinesUpdaterJob locally. It has not yet been tested on QA.


modules/mobile/app/models/mobile/v0/vaccine.rb
```diff
@@ -0,0 +1,27 @@
...
+    # Stores vaccine data from the CDC for use in Immunization records
+    # @example create a new instance
+    #   Mobile::V0::Vaccine.create(cvx_code: 1, group_name: 'FLU', manufacturer: 'Moderna')
+    #
+    class Vaccine < ApplicationRecord
+      validates :cvx_code, presence: true, uniqueness: true
+      validates :group_name, presence: true
+
+      # Adds incoming name to group name unless it's already included
+      # does not save record
```
kpethtel, line 14: The fact that this doesn't update the record is a bit atypical, but it helps us to prevent saving twice if we also need to update the manufacturer.

modules/mobile/app/workers/mobile/v0/vaccines_updater_job.rb
```diff
@@ -0,0 +1,87 @@
+# frozen_string_literal: true
+
+module Mobile
+  module V0
+    # This job is run daily and pulls data from the CDC to create vaccine records
+    class VaccinesUpdaterJob
```
kpethtel, line 6: TBH, I think this is a bit of a naive implementation. It simply goes through the first xml file (the one that has group names) and creates or updates records from those. If it encounters a new group name for a vaccine, it adds it to the vaccine record. The holes in this process are that 1) if there's ever a vaccine that's removed from the CDC xml, we don't delete it 2) if a group name is changed on the CDC xml, it doesn't remove the old group name, just adds the new one. 

These are both fixable issues, but it wasn't exactly clear how we want this to behave.

modules/mobile/spec/request/immunizations_request_spec.rb
```diff
@@ -422,5 +422,50 @@
         )
       end
     end
+
+    describe 'vaccines' do
```
kreek, line 426: Could context blocks be used here?

```ruby
describe 'vaccine group name and manufacturer population' do
  let(:immunizations_request) do
    VCR.use_cassette('lighthouse_health/get_immunizations', match_requests_on: %i[method uri]) do
      get '/mobile/v0/health/immunizations', headers: iam_headers, params: nil
    end
  end
  
  context 'when the vaccine record exists' do
    let!(:vaccine) { create(:vaccine, cvx_code: 140, group_name: 'COVID-19', manufacturer: 'Moderna') }

    it 'uses the vaccine group name and manufacturer in the response' do
      immunizations_request
      expect(response.parsed_body['data'][2]['attributes']).to eq(...)
    end
  end
  
  context 'when the vaccine record does not exist' do
    it 'sets group name and manufacturer to nil' do
      immunizations_request
      expect(response.parsed_body['data'][2]['attributes']).to eq(...)
    end
  end
end
```

modules/mobile/spec/workers/vaccines_updater_job_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+        expect(no_manufacturer.group_name).to eq('DTAP')
+        covid_vaccine = Mobile::V0::Vaccine.find_by(cvx_code: 207)
+        expect(covid_vaccine.group_name).to eq('COVID-19')
+        non_covid_vaccine = Mobile::V0::Vaccine.find_by(cvx_code: 2)
+        expect(non_covid_vaccine.group_name).to eq('POLIO')
+      end
+    end
+  end
+
+  it 'sets manufacturer when manufacturer data is available and group name is COVID-19' do
```
kreek, line 31: Same for this spec, every time `when` makes an appearance that's the context :)

modules/mobile/app/models/mobile/v0/adapters/immunizations.rb
```diff
@@ -9,16 +9,17 @@ def parse(immunizations)
             immunization = i[:resource]
             vaccine_code = immunization[:vaccine_code]
             cvx_code = vaccine_code[:coding].first[:code].to_i
+            vaccine = Mobile::V0::Vaccine.find_by(cvx_code: cvx_code)
```
kreek, line 12: This is happening inside a map so it's a SQL call for each iteration. Maybe there's a way to extract the CVX codes we need to find first, make one ActiveRecord find call and then update the model from that list? It would mean mapping over the date twice but I think that would be more performant than multiple SQL calls.

kreek (commented):
A few comments but this is looking good, I like how you track (and log) creation vs persisted vs update in the job 👍. I think you're right that it's a couple PRs, maybe even three? One for the migration, one for the background job, and then finally the last one for the update to the adapter/request.

modules/mobile/app/workers/mobile/v0/vaccines_updater_job.rb
```diff
@@ -0,0 +1,87 @@
...
+        logger.info('BEGIN --- Updating vaccine records from CDC')
+        results = { created: 0, updated: 0, persisted: 0 }
+
+        group_name_xml.root.children.each do |node|
+          result = process_vaccine(node)
+          results[result] += 1
+        end
+
+        if (results[:created] + results[:updated] + results[:persisted]).zero?
+          raise VaccinesUpdaterError, 'No records processed'
```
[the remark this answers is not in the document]
kpethtel: This is a canary in the coalmine error. I'm not sure that it's even possible. I tried testing the case where there are no child nodes but I wasn't able to make that happen because Nokogiri seems to add an empty child, which results in reaching the "malformed xml" error instead. But my concern is that this is a background job and there's always risk of background jobs failing silently and us not realizing it for months. I'm open to removing this, but that was my thought process in adding it.

modules/mobile/spec/request/immunizations_request_spec.rb
kpethtel: Definitely. I may need to do some style adjustment. My last company had a rule against using `let!` and a strong preference for local variables in rspec over `let`. I have... mixed feelings about it.

modules/mobile/spec/workers/vaccines_updater_job_spec.rb
kpethtel: I hear you. I usually avoid contexts unless there are multiple tests within it and perhaps data setup required for all tests within in. But I'm happy to conform to the shared style.

modules/mobile/app/models/mobile/v0/adapters/immunizations.rb
kpethtel: I'll look into that.

modules/mobile/app/workers/mobile/v0/vaccines_updater_job.rb
```diff
@@ -0,0 +1,87 @@
...
+      include Sidekiq::Worker
+
+      GROUP_NAME_URL = 'https://www2.cdc.gov/vaccines/iis/iisstandards/XML.asp?rpt=vax2vg'
+      MANUFACTURER_URL = 'https://www2a.cdc.gov/vaccines/iis/iisstandards/XML.asp?rpt=tradename'
+
+      class VaccinesUpdaterError < StandardError; end
+
+      # fetches group name and manufacturer data from the CDC and stores them in the vaccines table
+      def perform
+        logger.info('BEGIN --- Updating vaccine records from CDC')
```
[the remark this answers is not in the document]
kpethtel: Good to know. I'll make that change.

modules/mobile/app/workers/mobile/v0/vaccines_updater_job.rb
```diff
@@ -0,0 +1,87 @@
...
+
+      private
+
+      def process_vaccine(node)
+        cvx_code = find_value(node, 'CVXCode')
+        group_name = find_value(node, 'Vaccine Group Name')
+        manufacturer = group_name == 'COVID-19' ? find_manufacturer(cvx_code) : nil
+
+        vaccine = Mobile::V0::Vaccine.find_by(cvx_code: cvx_code)
+        if vaccine
```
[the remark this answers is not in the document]
kpethtel: Done.

PR #8507, opened 2021-11-22
Add rotation verification job and spec

LindseySaari:
## Description of change
This PR creates an encryption verification job to ensure that the rotation/re-encryption from Lockbox to KMS was successful. On Friday, we were made aware of some decryption issues from a vfs team member referencing two `AppealsApi::HigherLevelReview` records. These records were still encrypted via Lockbox and were not rotated properly. The records had an encrypted_kms_key, but were throwing decryption errors. Our theory is that the job to rotate may have been killed/restarted after the deploy and the records were in the middle of being rotated, causing incomplete rotation. After reverting the [PR to remove previous_version lockbox defaults ](https://github.com/department-of-veterans-affairs/vets-api/pull/8499), the records decrypted via Lockbox properly and further recovery efforts were not needed.  This job will ensure that all records are rotated properly.

Upon success of this job, the Lockbox previous_version options/keys can be removed. 

The master encryption key was added back to the initializer for the time being because it is need for the Lockbox decryption via the Lockbox::Box.

## Original issue(s)

1. The Lockbox::DecryptionError is the good scenario - If we are seeing a decryption error, this is GOOD and means that the record is not encrypted via Lockbox
2. The 'else' path is bad... meaning that the decryption via Lockbox was successful and shouldn't be since we've completed rotating the records.


app/workers/kms_encryption_verification_job.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+class KmsEncryptionVerificationJob
+  include Sidekiq::Worker
+  include SentryLogging
+
+  def perform(models = ApplicationRecord.descendants_using_encryption.map(&:name))
+    models.map(&:constantize).each do |m|
+      keys = m.lockbox_attributes.keys
```
ericboehs, line 9: Could we move the definition of keys into the lockbox_decrypt method?

LindseySaari:
Closing this PR in favor of a different approach to verify decryption via the kms key

PR #8511, opened 2021-11-23
API-11488 Add "stuck" appeals to error reporting

cilestin:
## Description of change
Adds to the Daily Error Report appeals that have not moved into a "completed" status for an extended period of time.

Screen Shot 2021-11-23 at 14 45 51

## Original issue(s)
https://vajira.max.gov/browse/API-11488


modules/appeals_api/spec/lib/decision_review_report_spec.rb
```diff
@@ -5,6 +5,25 @@
 
 describe AppealsApi::DecisionReviewReport do
   # rubocop:disable Layout/FirstHashElementIndentation
+  shared_examples 'stuck appeals' do |opts|
+    it 'retrieves records stuck in incomplete statuses' do
+      stuck_appeal_no_updates = stuck_appeal_with_updates = unstuck_appeal = nil
```
nathanbwright, line 10: I like the fact that this shortens the line count... But I think I would lean towards more readability (and more lines) in this case...

nathanbwright (approved):
Looks good! Left one non-blocking comment.

nathanbwright (approved):
🚀

PR #8514, opened 2021-11-23
KMS decryption/verification job

LindseySaari:
## Description of change
This PR creates a decryption verification job to ensure that the rotation/re-encryption from Lockbox to KMS was successful. On Friday, we were made aware of some decryption issues from a vfs team member referencing two `AppealsApi::HigherLevelReview` records. These records were still encrypted via Lockbox and were not rotated properly. The records had an encrypted_kms_key, but were throwing decryption errors. Our theory is that the job to rotate may have been killed/restarted after the deploy and the records were in the middle of being rotated, causing incomplete rotation. After reverting the [PR to remove previous_version lockbox defaults ](https://github.com/department-of-veterans-affairs/vets-api/pull/8499), the records decrypted via Lockbox properly and further recovery efforts were not needed.  This job will ensure that all records are rotated properly.

Upon success of this job, the Lockbox previous_version options/keys can be removed. 

## Original issue(s)


app/workers/kms_encryption_verification_job.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class KmsEncryptionVerificationJob
+  include Sidekiq::Worker
+
+  def perform(models = ApplicationRecord.descendants_using_encryption.map(&:name))
+    models.map(&:constantize).each do |model|
+      keys = model.lockbox_attributes.keys
+      model.where('encrypted_kms_key IS NOT NULL').each do |record|
```
ericboehs, line 9:
```suggestion
      model.where.not(encrypted_kms_key: nil).each do |record|
```

app/workers/kms_encryption_verification_job.rb
```diff
@@ -0,0 +1,25 @@
...
+        keys.each do |k|
+          box = KmsEncrypted::Box.new
+          decrypted_data_key = box.decrypt(record.encrypted_kms_key,
+                                           context: { model_name: model.to_s, model_id: record.id })
```
ericboehs, line 13:
```suggestion
          decrypted_kms_key = box.decrypt(
            record.encrypted_kms_key, context: { model_name: model.to_s, model_id: record.id }
          )
```

Two suggestions here, rename the variable, and (nit) don't do a super indent on line 13.

app/workers/kms_encryption_verification_job.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class KmsEncryptionVerificationJob
+  include Sidekiq::Worker
+
+  def perform(models = ApplicationRecord.descendants_using_encryption.map(&:name))
+    models.map(&:constantize).each do |model|
+      keys = model.lockbox_attributes.keys
```
ericboehs, line 8:
```suggestion
      encrypted_attributes = model.lockbox_attributes.keys
```

These aren't the encryption keys, right? These are the db fields/model attributes?

app/workers/kms_encryption_verification_job.rb
```diff
@@ -0,0 +1,25 @@
...
+      model.where('encrypted_kms_key IS NOT NULL').each do |record|
+        keys.each do |k|
+          box = KmsEncrypted::Box.new
+          decrypted_data_key = box.decrypt(record.encrypted_kms_key,
+                                           context: { model_name: model.to_s, model_id: record.id })
+          lockbox = Lockbox.new(key: decrypted_data_key, encode: true)
+          begin
+            lockbox.decrypt(record.send("#{k}_ciphertext"))
```
ericboehs, line 16:
```suggestion
            lockbox.decrypt(record.public_send("#{k}_ciphertext"))
```

app/workers/kms_encryption_verification_job.rb
LindseySaari: Yeah they're the model attributes

app/workers/kms_encryption_verification_job.rb
LindseySaari: So this isn't the actual encrypted kms key itself, but rather the data key

app/workers/kms_encryption_verification_job.rb
```diff
@@ -0,0 +1,25 @@
...
+      keys = model.lockbox_attributes.keys
+      model.where.not(encrypted_kms_key: nil).each do |record|
+        keys.each do |k|
+          box = KmsEncrypted::Box.new
+          decrypted_data_key = box.decrypt(record.encrypted_kms_key,
+                                           context: { model_name: model.to_s, model_id: record.id })
+          lockbox = Lockbox.new(key: decrypted_data_key, encode: true)
+          begin
+            lockbox.decrypt(record.send("#{k}_ciphertext"))
+          rescue
```
ericboehs, line 17: Can we rescue a specific exception here instead of all exceptions?

app/workers/kms_encryption_verification_job.rb
```diff
@@ -0,0 +1,40 @@
...
+  rescue Lockbox::DecryptionError
+    Rails.logger.error("Record re-encryption unsuccessful model:
+                        #{record.class.name} - id: #{record.id} - attribute: #{attribute}")
+    false
+  else
+    true
+  end
+
+  def log_corrupted_records(corrupted_records)
+    log_message_to_sentry('Lockbox Decryption Error: Lockbox::DecryptionError', :error,
```
ericboehs, line 36: I think just raising is enough to send it to sentry, right?

app/workers/kms_encryption_verification_job.rb
```diff
@@ -0,0 +1,40 @@
...
+          record unless decrypted?(record, attribute)
+        end
+      end
+    end.compact
+    log_corrupted_records(corrupted_records) if corrupted_records.any?
+  end
+
+  private
+
+  def decrypted?(record, attribute)
```
ericboehs, line 21: I'd name this `can_decrypt?` or `decryptable?` as you're not asking if the data is already decrypted, but if it's capable of being decrypted.

app/workers/kms_encryption_verification_job.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+class KmsEncryptionVerificationJob
+  include Sidekiq::Worker
+  include SentryLogging
+
+  def perform(models = ApplicationRecord.descendants_using_encryption.map(&:name))
+    corrupted_records = models.map(&:constantize).flat_map do |model|
+      attributes = model.lockbox_attributes.keys
+      model.where('encrypted_kms_key IS NOT NULL').flat_map do |record|
```
ericboehs, line 10:
```suggestion
      model.where.not(encrypted_kms_key: nil).flat_map do |record|
```

app/workers/kms_encryption_verification_job.rb
LindseySaari: Ah yes, I think you're right. The exception will bubble up to the application_controller (I think) and be logged to Sentry via [this concern ](https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/controllers/concerns/exception_handling.rb)

thilton-oddball (approved):

PR #8519, opened 2021-11-23
32875 precheck in get implementation

kanchanasuriya:
## Description of change

Implementation for Pre-Checkin/GET implementation to get demographics & appointments from LoROTA GET endpoint

## Original issue(s)
department-of-veterans-affairs/va.gov-team#32875

- [X] rspec
- [X] rubocop


modules/check_in/spec/request/v2/pre_check_ins_request_spec.rb
```diff
@@ -3,7 +3,7 @@
 require 'rails_helper'
 
 RSpec.describe 'V2::PreCheckInsController', type: :request do
-  let(:id) { '2f90aa72-deba-4bd7-9bd6-5d17c70bb1b3' }
+  let(:id) { Faker::Internet.uuid }
   let(:memory_store) { ActiveSupport::Cache.lookup_store(:memory_store) }
 
   before do
```
randomsync, line 9: Would suggest doing `Rails.cache.clear` so tests are order independent

randomsync (approved):
one minor comment but otherwise looks good

modules/check_in/spec/request/v2/pre_check_ins_request_spec.rb
randomsync: sorry, didn't see that it was already included previously.

PR #8516, opened 2021-11-23
1930: create vaccines table

kpethtel:
## Description of change
Add vaccines table with cvx_code and manufacturer for use in immunizations records. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#1930
https://app.zenhub.com/workspaces/va-mobile-eng-60f1a34998bc75000f2a489f/issues/department-of-veterans-affairs/va-mobile-app/1930

## Things to know about this PR
Requires data migration. This is only the data change. More will come in subsequent PRs. 

This code has been tested locally by running a script that fetches all vaccine data from the CDC and creates vaccines for each.

PR #8522, opened 2021-11-24
1930: create vaccines from CDC XML data

kpethtel:
## Description of change
Adds a new job to the scheduler that will create vaccine records based on data from the CDC. This improves upon the current code by ensuring that any changes added by the CDC will automatically be pulled into the VA mobile app.

## Original issue(s)
https://app.zenhub.com/workspaces/va-mobile-eng-60f1a34998bc75000f2a489f/issues/department-of-veterans-affairs/va-mobile-app/1930

Data migration was performed here: https://github.com/department-of-veterans-affairs/vets-api/pull/8516

After this is merged, we will release another PR to change the immunizations model to use the new vaccine model.

This has been tested locally by running the sidekiq job and creating the vaccine records.


modules/mobile/app/workers/mobile/v0/vaccines_updater_job.rb
```diff
@@ -0,0 +1,99 @@
+# frozen_string_literal: true
+
+module Mobile
+  module V0
+    # This job is run daily and pulls data from the CDC to create vaccine records
+    class VaccinesUpdaterJob
+      include Sidekiq::Worker
+
```
kreek, line 8: The default number of sidekiq retry attempts is 25 (which with its [exponential back-off algorithm](https://github.com/mperham/sidekiq/wiki/Error-Handling#automatic-job-retry) ends up being 21 days) since this is a daily job I think 3-10 would keep it within one day, but we probably only need the lower end.
```ruby
sidekiq_options retry: 3
```

kreek (approved):
One minor comment about retrying, but LGTM 🚀

modules/mobile/app/workers/mobile/v0/vaccines_updater_job.rb
kpethtel: That's a good point. Added.

modules/mobile/spec/workers/vaccines_updater_job_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+      end
+    end
+  end
+
+  context 'when vaccine record exists' do
+    let!(:covid_no_manufacturer) { create(:vaccine, cvx_code: 503, group_name: 'FLU', manufacturer: 'Bayer') }
+    let!(:covid_with_manufacturer) { create(:vaccine, cvx_code: 207, group_name: 'FLU', manufacturer: 'Bayer') }
+    let!(:non_covid_multiple_groups) { create(:vaccine, cvx_code: 110, group_name: 'FLU', manufacturer: 'Bayer') }
+
+    it 'updates the record' do
```
kreek, line 52: When you have multiple expectations in one block you can [aggregate failures](https://justincypret.com/blog/aggregate-failures-in-rspec) to avoid one expectation causing the whole block to fail. This matters more when you're working/TDD'ing through a pass-fail cycle locally, but it also makes the merge check report more accurate should it fail.

kreek (approved):
One comment, but LGTM, feel free to merge once that's in 🚀

modules/mobile/spec/workers/vaccines_updater_job_spec.rb
kpethtel: Done. Apparently I need another review after adding a commit.

kreek (approved):
LGTM 🚀

holdenhinkle:
Hey @kpethtel,

I just merged master into a pr and saw your update to `config/sidekiq_scheduler.yml`. I noticed that you indented line 277:

```
  Mobile::V0::VaccinesUpdaterJob:

```

I guessing that job won't run because of that indentation.

Anyway, just wanted to let you know. I noticed because I'm trying to added another job to the file and I had to accept your incoming changes.

Holden

PR #8525, opened 2021-11-24
API-10869: Fix POA test user data

hinzed1127:
## Description of change

This updates POA sample data to conform to 2 things:
1. the null check constraint added in #8492 
2. Tamara Ellis and John Doe have different unique individual POA codes. This both passes v2 POA validations and is also more accurately reflective of real-world data

## Original issue(s)
https://vajira.max.gov/browse/API-10869

- No tests were specifically added, since this was fixing test data. The scenario this addresses is already tested [here](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/spec/requests/v2/veterans/power_of_attorney_request_spec.rb#L68).


rileyanderson (approved):

modules/veteran/lib/tasks/veteran_tasks.rake
```diff
@@ -16,13 +16,15 @@ namespace :veteran do
     Veteran::Service::Organization.create(poa: '095', name: '095 - ITALIAN AMERICAN WAR VETERANS OF THE US, INC.')
     Veteran::Service::Organization.create(poa: '1NY', name: '1NY - SAMANTHA Y WARSHAUER')
     Veteran::Service::Representative.create(
+      representative_id: '98765',
```
jw81, line 19: FWIW, when @SlakrHakr and I were working on the cleaning up the existing POA data so that it would conform to the new constraints, we set Tamara's `representative_id` to `999999999999`.

modules/veteran/lib/tasks/veteran_tasks.rake
```diff
@@ -16,13 +16,15 @@ namespace :veteran do
...
       poa_codes: %w[A1Q 095 074 083 1NY],
       first_name: 'Tamara',
       last_name: 'Ellis',
       email: 'va.api.user+idme.001@gmail.com'
     )
     Veteran::Service::Representative.create(
-      poa_codes: %w[A1Q 095 074 083 1NY],
+      representative_id: '12345',
```
jw81, line 26: FWIW, when @SlakrHakr and I were working on the cleaning up the existing POA data so that it would conform to the new constraints, we set John Doe's `representative_id` to `9999999999999`. (one more `9` than Tamara's id)

jw81 (approved):

PR #8527, opened 2021-11-25
32876 Implement POST for Pre Check-in

randomsync:
## Description of change

This PR implements the POST endpoint for PreCheckIns Controller which users can use to confirm that they have up to date information for their demographics and NOK data.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#32876

- [x] rspec
- [x] rubocop
- [x] local testing


rileyanderson (approved):

modules/check_in/spec/services/v2/chip/service_spec.rb
```diff
@@ -176,6 +176,50 @@
...
+      end
+
+      it 'returns unauthorized status' do
+        expect(subject.build(check_in: valid_check_in, params: params)
+                      .pre_check_in.status).to eq(401)
+      end
+    end
+  end
+
   describe '#token' do
```
kanchanasuriya, line 223: Can we please add tests for ``demographic_confirmations`` method?

modules/check_in/spec/request/v2/pre_check_ins_request_spec.rb
```diff
@@ -161,4 +153,48 @@
...
+
+      before do
+        allow_any_instance_of(::V2::Chip::Service).to receive(:pre_check_in).and_return(unauth_response)
+      end
+
+      it 'returns unauthorized response' do
+        post '/check_in/v2/pre_check_ins', params: post_params
+
+        expect(response.body).to eq(unauth_response.body.to_json)
+        expect(response.status).to eq(401)
```
kanchanasuriya, line 196: nit: may be return unauth_response.status instead of 401 directly

modules/check_in/app/services/v2/chip/service.rb
```diff
@@ -90,6 +98,19 @@ def identifier_params
...
 
+      def demographic_confirmations
+        demographics_needs_update = check_in_body[:demographics_up_to_date] ? false : true
+        next_of_kin_needs_update = check_in_body[:next_of_kin_up_to_date] ? false : true
+        {
+          demographicConfirmations: {
+            demographicsNeedsUpdate: demographics_needs_update,
+            demographicsConfirmedAt: Time.zone.now.iso8601,
+            nextOfKinNeedsUpdate: next_of_kin_needs_update,
+            nextOfConfirmedAt: Time.zone.now.iso8601
```
kanchanasuriya, line 109: nit: Wondering if the payload is `nextOfConfirmedAt` instead of `nextOfKinConfirmedAt`

modules/check_in/spec/request/v2/pre_check_ins_request_spec.rb
```diff
@@ -161,4 +153,48 @@
...
   end
+
+  describe 'POST `create`' do
+    let(:post_params) do
+      {
+        pre_check_in: {
+          uuid: id,
+          demographics_up_to_date: true,
+          next_of_kin_up_to_date: true,
+          check_in_type: :pre_check_in
```
kanchanasuriya, line 164: nit: "preCheckIn" instead of : pre_check_in

modules/check_in/app/docs/check_in/v2/check_in.yaml
```diff
@@ -284,6 +284,56 @@ paths:
...
+          description: Unauthorized
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Unauthorized'
+              example:
+                status: 401
+                data:
+                  error: true
+                  message: 'Token verification failed: error message'
```
kanchanasuriya, line 316: nit: `unauthorized_message` from `Session` object seem to be returning `{ permissions: 'read.none', status: 'success', uuid: uuid }`

modules/check_in/app/services/v2/chip/service.rb
acrollet: As of now `nextOfConfirmedAt` is correct: https://github.com/department-of-veterans-affairs/clinician_workflow/blob/b8dde1a4d6d89d44f8b03eceda12c448dabc3d6f/openapi/v1/openapi.yaml#L1202

If you want you can try to convince @charleystran to change it :-D

kanchanasuriya: :-D thanks for chiming in.  May be this field is common for all nextOf__ :-D.  I will resolve this.

modules/check_in/app/services/v2/chip/service.rb
```diff
@@ -90,6 +98,19 @@ def identifier_params
         }
       end
 
+      def demographic_confirmations
+        demographics_needs_update = check_in_body[:demographics_up_to_date] ? false : true
```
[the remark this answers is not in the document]
randomsync: no, it specifically has to be reversed since the FE sends us the flag whether the data is up to date or not, and CHIP expects us to send whether the data needs to be updated or not. so if 'up to date' = true, 'update needed' = false

modules/check_in/spec/request/v2/pre_check_ins_request_spec.rb
randomsync: updated in 2f84e5f50a810f003c289662a79e79a1d922b0ed

modules/check_in/spec/request/v2/pre_check_ins_request_spec.rb
randomsync: updated in 2f84e5f50a810f003c289662a79e79a1d922b0ed

modules/check_in/app/docs/check_in/v2/check_in.yaml
randomsync: updated in 2f84e5f50a810f003c289662a79e79a1d922b0ed

modules/check_in/spec/services/v2/chip/service_spec.rb
randomsync: As part of rpsec, I feel that this method is being called within service spec and there's not much else to test. It probably makes more sense to test it in some kind of integration test (which we don't have currently) to make sure that what we get from FE through the controller POST params is translated correctly to the CHIP service. 

Do you still feel rspec for this are needed, and if so what would you suggest testing in this method?

kanchanasuriya: Since we have a ternary condition check, thought is good to have tests like,
- `demographics_needs_update` set with true when FE sends `check_in_body[:demographics_up_to_date]` with false 
- `demographics_needs_update` set with false when FE sends `check_in_body[:demographics_up_to_date]` with true 
- `next_of_kin_needs_update` set with true when FE sends `check_in_body[:next_of_kin_up_to_date]` with false
- `next_of_kin_needs_update` set with false when FE sends `check_in_body[:next_of_kin_up_to_date]` with true

randomsync: Are we testing here that a ternary condition works?

kanchanasuriya: Yes 👍

kanchanasuriya: So in future this rspec will make sure that this contract between FE/vets-api/CHIP doesn't change in check-in module

randomsync: I honestly feel that it's superfluous and unnecessary to test that ternary conditions in Ruby works. Like I said earlier, it makes more sense for an integration test to validate the input and output and since we don't have an automated integration test approach currently, it will be a part of integration testing we'll do as soon as FE code is ready.

kanchanasuriya: Synced offline, above tests will be covered in integration test.

dillo (approved):
LGTM

Issue #6000, opened 2021-02-24
Consolidate MPI::Service and MPI::AttrService


bosawt:
It looks like we use both of these as ways to call MPI, but they are set up largely the same. Here's the prod settings: [Prod settings.yml](https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/deployment/config/vets-api/prod-settings.local.yml.j2)
Inspect the `mvi` and `mvi_hca` sections.

The only difference I could find was the `service_name` is different for both of these in the `mpi/configuration` and `mpi/attr_configuration`. 

I think we could consolidate both of these service entry points to MPI and make it a unified service

nsoutouras:
Nick: Tech debt cleanup.  We have multiple ways of calling MPI that are redundant.

PR #8531, opened 2021-11-29
Remove attr_service and replace with mpi-service

jscanish:
## Description of change
We use both `MPI::Service` and `MPI::AttrService` to call `MPI`. The goal here is to remove `MPI::AttrService` and reduce our `MPI` calls to just the `MPI::Service` class.

This pull request will remove `MPI::AttrService` and replace it, but there will need to be a separate pull request later to remove and/or update some of the configurations. That will likely require changes to the devops repo as well.

## Original issue(s)
https://github.com/department-of-veterans-affairs/vets-api/issues/6000


bramleyjl (approved):
LGTM!

lib/mpi/service.rb
```diff
@@ -166,13 +168,21 @@ def create_add_message(user_identity)
...
-      return message_edipi(user_identity, search_type) if user_identity.edipi.present? && Settings.mvi.edipi_search
+      return message_icn(user_identity, search_type) if mhv_icn_present?(user_identity)
+      return message_edipi(user_identity, search_type) if edipi_present?(user_identity)
       raise Common::Exceptions::ValidationErrors, user_identity unless user_identity.valid?
 
       message_user_attributes(user_identity, search_type)
     end
 
+    def mhv_icn_present?(user_identity)
+      user_identity.respond_to?(:mhv_icn) && user_identity.mhv_icn.present?
```
bosawt, line 179: Instead of `respond_to?` for all of these, we can just do the safe access operator: `user_identity&.mhv_icn.present?`

modules/veteran_confirmation/app/services/veteran_confirmation/service.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+require 'mpi/service'
```
bosawt, line 3: I don't think this file is being used anywhere?

bosawt (commented):
Left a couple of comments, just small stuff to change, nothing huge.

I noticed that Prod has the same settings for `mvi` and `mvi_hca` (which is good), but there are different settings for the other stacks. It'd be good to communicate to other teams via Slack and just make sure there aren't expectations with the health care applications service and this calling specific MPI stacks on different test stacks

lib/mpi/service.rb
jscanish: Ordinarily yes, but in this case `user_identity` is a struct object and they don't function with the safe operator. It might be a good idea for me to put a comment here clarifying that though so someone doesn't try to refactor it later.

modules/veteran_confirmation/app/services/veteran_confirmation/service.rb
jscanish: Yeah good catch. I don't think the configuration file is needed either actually.

lib/mpi/service.rb
bosawt: Ahhh ok, makes sense

modules/veteran_confirmation/app/services/veteran_confirmation/service.rb
bosawt: 👍

bosawt (approved):
I think this is good now, also, we can remove the `mvi_hca` block in `settings.yml` but that can be in this PR or another one (also should be removed in the `DevOps` repo for different stacks). 

And of course when this is ready to go, should be communicated somewhere in case anyone is relying on this being different stacks than our other MPI Service

app/models/health_care_application.rb
```diff
@@ -117,7 +117,7 @@ def self.enrollment_status(icn, loa3)
 
   def self.user_icn(user_attributes)
     HCA::RateLimitedSearch.create_rate_limited_searches(user_attributes) unless Settings.mvi_hca.skip_rate_limit
-    MPI::AttrService.new.find_profile(user_attributes)&.profile&.icn
+    MPI::Service.new.find_profile(user_attributes)&.profile&.icn
```
jchilders, line 120: Shouldn't this be `VeteranConfirmation::Service` instead? `MPI::AttrService` had the overridden `create_profile_message` and `measure_info` methods, which have been moved into `VC::Service`.

bosawt: Oh yeah, good catch. Actually I think we should just remove the `VeteranConfirmation::Service` file. The validation that was failing before in `lib/mpi/service.rb` should work now with these changes.

I think `measure_info` will have issues as , but it can be updated to work fine with this use case (it probably was failing because it calls `.uuid` and `.authn_context` on the input `identity` object but since it's just a log we can contextually drop those fields if they don't exist in the input object

jscanish: Those methods were moved to a `VeteranConfirmation` support file since I was under the impression that they were most relevant to `VeteranConfirmation::AttrService.` That's where most of the issues around validations where originating. The `VeteranConfirmation::Service` file was being used in `veteran_confirmation/status_service` to avoid those validation issues.

It does make the whole thing messy and confusing though so I removed `VeteranConfirmation::Service` entirely. The validation still fails however as `MPI::Service` is generally operating with a complex `User` object and VeteranConfirmation is building a very simple OpenStruct.

It seems hacky to check the status of the `user_identity` object and just skip the validations if it's an OpenStruct, but I can't think of another way to handle it that doesn't involve dramatically rewriting a lot of validations. Since the validations were being skipped entirely before anyway I figure there's no reason to do that work now.

Does that make sense?

lib/mpi/service.rb
```diff
@@ -166,13 +168,21 @@ def create_add_message(user_identity)
     end
 
     def create_profile_message(user_identity, search_type: MPI::Constants::CORRELATION_WITH_RELATIONSHIP_DATA)
-      return message_icn(user_identity, search_type) if user_identity.mhv_icn.present?
-      return message_edipi(user_identity, search_type) if user_identity.edipi.present? && Settings.mvi.edipi_search
+      return message_icn(user_identity, search_type) if mhv_icn_present?(user_identity)
```
bosawt, line 171: I don't think we need the changes in this file anymore, do we?

jscanish: We do because of the health care application. Specifically [these lines](https://github.com/department-of-veterans-affairs/vets-api/blob/d8f083a5a1c66f480f2c57573d0f3d845579ff10/app/models/health_care_application.rb#L118-L139) which are using the `UserAttributes` class to format data and call `MPI`.

We could take a similar approach and refactor it to use `UserIdentity` like the other instances, but it's a little more involved in this case.

lib/hca/user_attributes.rb
```diff
@@ -12,12 +12,15 @@ class HCA::UserAttributes
...
+  attr_reader :mhv_icn, :edipi, :gender, :authn_context
+
   def ssn=(new_ssn)
     super(new_ssn&.gsub(/\D/, ''))
   end
 
-  def gender
-    # MVI message_user_attributes expects a gender value but it's not asked on the HCA form
-    nil
+  def uuid
```
bosawt, line 23: I think this will need LOA defined as well, won't it?

jscanish: It depends. When [this validation](https://github.com/department-of-veterans-affairs/vets-api/blob/master/lib/mpi/service.rb#L171) is called `user_identity` is actually an `HCA::UserAttributes` object. That object has its own validations and doesn't currently require `loa`.

So adding `loa` is not necessary to make the tests pass and it's not an attribute the HCA form seems to care about, but I'm not sure if it's needed elsewhere in the application.

bosawt: Ah yes you're right, I forgot that was just wrapped up in the validation part

bosawt (approved):
Looks good!

LindseySaari (approved):

bramleyjl (approved):

PR #8532, opened 2021-11-29
1985 mobile direct deposit mfa flags

jperk51:
## Description of change
Checking multifactor to decide if a user has access to direct deposit

## Original issue(s)
department-of-veterans-affairs/va-mobile-app#1985

Updated user spec and docs


modules/mobile/docs/index.html
```diff
@@ -1078,7 +1078,7 @@
...
...onses/408/content/application~1json/schema"}}}},"502":{"description":"An upstream service the API depends on returned an error.","content":{"application/json":{"schema":{"$ref":"#/paths/~1/post/responses/408/content/application~1json/schema"}}}},"503":{"description":"An upstream service is unavailable, or its circuit breaker may have been tripped.","content":{"application/json":{"schema":{"$ref":"#/paths/~1/post/responses/408/content/application~1json/schema"}}}},"504":{"description":"A response from an upstream service timed out.","content":{"application/json":{"schema":{"$ref":"#/paths/~1/post/responses/408/content/application~1json/schema"}}}}}},"put":{"summary":"/v0/push/prefs/{endpointSid}","description":"Set the user's push notification preferences","operationId":"putPrefs","security":[{"Bearer":[]}],"parameters":[{"in":"path","name":"endpointSid","required":true,"schema":{"type":"string"},"description":"device endpointSid provided by the register endpoint"}],"requestBody":{"description":"Push notification preferences","required":true,"content":{"application/json":{"schema":{"type":"object","required":["preference","enabled"],"properties":{"preference":{"type":"string","examp
```
kpethtel, line 1081: I might have to take your word on this one. :)

kpethtel (commented):
The changes in the serializer make sense, but it looks as though all the spec changes are just adding a response from the payment service. Do we need any new tests more specifically testing the authorized services within the user endpoint response?

modules/mobile/app/serializers/mobile/v0/user_serializer.rb
```diff
@@ -42,7 +42,7 @@ class UserSerializer
         appeals: :appeals,
         appointments: :vaos,
         claims: :evss,
-        directDepositBenefits: %i[evss ppiu],
+        directDepositBenefitsView: %i[evss ppiu],
```
kreek, line 45: Have you checked with @narin and the FE crew that it's ok to change an authorized service name? I think adding one is fine (following the [API evolution](https://apisyouwonthate.com/blog/api-evolution-for-rest-http-apis) strategy) but otherwise we might have to version the endpoint.

jperk51: That makes sense. I'll check with the FE

modules/mobile/docs/index.html
jperk51: Yeah, the html is massive. The real changes happen in the docs yml files. In this case it was User.yml

modules/mobile/app/serializers/mobile/v0/user_serializer.rb
jperk51: FE let me know they use this so I reverted it to the original. directDepositBenefits in authorized_services indicates view access only

kreek (approved):
LGTM 🚀

rileyanderson (approved):

PR #8545, opened 2021-11-30
[API-11613] - ClaimsApi - Restore "Not authorized" responses for 401 errors

jw81:
## Description of change

A consumer of the `Benefits Claims Api` has built a workflow for handling expired tokens.  In order for their workflow to function, they have built it to depend on the verbiage contained in the error response `title`.  Specifically, their workflow depends on the `title` value being `Not authorized` (this of course is not ideal).  Back on [May 7th, 2021](https://github.com/department-of-veterans-affairs/vets-api/commit/f9887a0d5a30d30a99f7270979d4de9a18a11978), a change was made to the global exception mapping logic in `vets-api`.  That change made it such that the `title` value became `Token Validation Error`.  (which breaks the consumer workflow).

In an effort to keep the consumer from having to make a code change, this PR looks to catch `::Common::Exceptions::TokenValidationError` exceptions and instead raise `::Common::Exceptions::Unauthorized` exceptions.  This should restore the `title` value of `Not authorized` that the consumer depends on.

## Original issue(s)
[API-11613](https://vajira.max.gov/browse/API-11613)


dbrown-va (approved):
This looks good.

In the future on the token validation side we can work on defining an enum of specific responses so that we don't have to blindly include the `e.detail` if it ever ends up a stack trace.

hinzed1127 (approved):

PR #8554, opened 2021-12-01
1930: use vaccine model in immunizations

kpethtel:
## Description of change
Changes the way we form immunizations records to use data from the newly created vaccine model.

## Original issue(s)

This changes the way vaccine group names are fetched and adds the vaccine manufacturer to some covid-19 immunization records. It does not change the shape of the data. 

This has been tested locally but will also be tested by curling the immunizations endpoint on the deploy environment before merging. I'm unable to do that currently because I'm still waiting on MFA access.


jperk51 (approved):
LGTM

PR #8555, opened 2021-12-01
[33436] updates multifactor method to parse AAL for Login.gov logins

bramleyjl:
## Description of change

This change updates the SSOE User Attribute `multifactor` method to parse the AAL from ISAM's response in the case of Login.gov logins. An AAL of 2 or greater must include 2FA, this covers all Login.gov logins. Other CSPs will continue to rely on the `va_eauth_multifactor` attribute passed back by ISAM instead of AAL.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#33436

Unit specs updated, manual logins with Login.gov confirmed `multifactor = true` for IAL1 & IAL2, manual logins with other CSPs to confirm they are unaffected.


jscanish (dismissed):
Seems reasonable to me.

AGarcia-Clarity (approved):

lib/saml/user_attributes/ssoe.rb
```diff
@@ -184,7 +184,11 @@ def loa_highest
       end
 
       def multifactor
-        safe_attr('va_eauth_multifactor')&.downcase == 'true'
+        if csid == SAML::User::LOGINGOV_CSID
+          safe_attr('va_eauth_aal') >= '2'
```
bosawt, line 188: Can we make the `2` a constant? I believe we could add it in the `aal.rb` initializer

bosawt: Also, can we make it `==`? I just suggest that because anything greater than `2` would be unexpected behavior, so we may as well prevent it

bosawt (approved):
Looks good!

rileyanderson (approved):

PR #8563, opened 2021-12-02
Natashajibrahim/hypertension pilot refactor

NatashaJIbrahim:
## Description of change
This moves the Fast Track work into it's own module and moves all of the helper classes into their own files. It also re-records the cassettes and reorganizes the specs so that their directory structure matches that of the module they're testing.

This work doesn't introduce any functional changes to the code. Also, the code is still behind a feature flag.


app/models/form526_submission.rb
```diff
@@ -55,7 +55,7 @@ def start
         'submission_id' => id
       )
       jids = workflow_batch.jobs do
-        DisabilityCompensationFastTrackJob.perform_async(id, get_full_name)
```
julialeague, line 58: Should this be renamed to FastTrack::DisabilityCompensationJob, or should that module be renamed? I think I prefer this name to the other.

NatashaJIbrahim: eek - It should be `FastTrack::DisabilityCompensationJob` and I should look into why the spec didn't hit that code. 

I like the old name too. Do you have any ideas for naming the module?

julialeague: I commented on the wrong side of the diff here, sorry - I meant that I prefer `DisabilityCompensationFastTrack::Job`, but that's not a passionate preference.

NatashaJIbrahim: It should be fixed now .. I mean it the test is now passing etc.

tadhg-ohiggins (approved):

PR #8570, opened 2021-12-03
Add kms verification migration and schema changes

LindseySaari:
## Description of change
This PR adds a migration to add a date column to db tables with encrypted fields. 

### Background
The `KmsEncryptionVerificationJob` needs to run to verify that ALL records can be successfully decrypted with the new kms key. Records were previously encrypted via the Lockbox master key and the rotation task re-encrypted attributes via AWS KMS. 

The `KmsEncryptionVerificationJob` will record a date in the `verified_decryptable_at` db column if decryption via KMS was successful. This is needed because as the `KmsEncryptionVerificationJob` runs, we are seeing quite a few `Seahorse::Client::NetworkingError`'s come through in [Sentry](http://sentry.vfs.va.gov/organizations/vsp/issues/57920/?environment=production&project=3&query=is%3Aunresolved+KmsEncryptionVerificationJob&statsPeriod=24h). This is a timeout error related to the AWS call to KMS.  The `Seahorse::Client::NetworkingError` causes the job `KmsEncryptionVerificationJob` to restart. With this change in place, the next run of the job will skip over records that have a `verified_decryptable_at` date

This PR will be broken into two pieces (/db and app changes). The second PR that makes changes to the `KmsEncryptionVerificationJob` [can be found here](https://github.com/department-of-veterans-affairs/vets-api/pull/8571).

## Original issue(s)


db/schema.rb
```diff
@@ -765,16 +785,6 @@
...
 
-  create_table "test_user_dashboard_tud_account_checkouts", force: :cascade do |t|
-    t.string "account_uuid"
-    t.datetime "checkout_time"
-    t.datetime "checkin_time"
-    t.boolean "has_checkin_error"
-    t.boolean "is_manual_checkin"
-    t.datetime "created_at", precision: 6, null: false
-    t.datetime "updated_at", precision: 6, null: false
-  end
```
rileyanderson, line 776: Should this table be removed?

LindseySaari: Oh boy, definitely not. I think I may have something weird locally going on.

rileyanderson: Yeah, I've definitely had that happen before too 😄

LindseySaari: Ok, the schema oopsie accident is updated 🤣

rileyanderson (approved):

PR #8576, opened 2021-12-03
Add permissions to GHA workflows

rileyanderson:
## Description of change
This adds permissions to the GitHub actions so that they don't have global write access to everything. 

This is also fixes actions started by Dependabot. By default the `GITHUB_TOKEN` in these actions is read only. Adding these permissions to the workflow allows the `GITHUB_TOKEN` to be generated with the necessary read/write access to properly run.  [GitHub Changelog](https://github.blog/changelog/2021-10-06-github-actions-workflows-triggered-by-dependabot-prs-will-respect-permissions-key-in-workflows/)


ericboehs (approved):
👨‍🍳 💋 🤌

Looks amazing

PR #8585, opened 2021-12-03
[32270] Mpi attributes

bramleyjl:
## Original issue(s)
department-of-veterans-affairs/va.gov-team#32270


rileyanderson (dismissed):

app/models/user.rb
```diff
@@ -153,7 +153,7 @@ def active_mhv_ids
   end
 
   def address
-    address = mpi_profile&.address
+    address = identity&.address || mpi_profile&.address
```
bosawt, line 156: It looks like you'll have to change the structure of the `identity.address` object for this to work. Since it's a hash, you can't to `address.street`, etc, you'd need to change it to `address[:street]`, or, alternatively make it an `OpenStruct` in the definition in ssoe.rb

lib/saml/user_attributes/ssoe.rb
```diff
@@ -54,6 +58,16 @@ def icn
         mvi_ids[:icn]
       end
 
+      def address
+        {
```
bosawt, line 62: Prob need to change this to: `OpenStruct.new({
           street: safe_attr('va_eauth_street'),
           city: safe_attr('va_eauth_city'),
           state: safe_attr('va_eauth_state'),
           country: safe_attr('va_eauth_country'),
           postal_code: zip
         })`

bosawt (changes requested):
Mostly looks good, I ran into an issue testing the address though, described in the comments

spec/services/medical_copays/vbs/service_spec.rb
```diff
@@ -5,7 +5,7 @@
 RSpec.describe MedicalCopays::VBS::Service do
   subject { described_class.build(user: user) }
 
-  let(:user) { build(:user, :loa3) }
+  let(:user) { build(:user, :loa3, :no_vha_facilities) }
```
ScottyRJames, line 8: The service that this user is being stubbed for generally expects a user to have an existing `vha_facility_hash` and while the `RequestData` result is being mocked it still might make since to keep this vha facility object intact for the user

bramleyjl: Thanks for the heads up, I changed that spec's user to have a `vha_facility_hash` that lines up with the `get_facilities_empty.yml` VCR cassette and the unit tests are all passing. Let me know if it needs any further tweaking.

app/models/user.rb
```diff
@@ -153,13 +153,14 @@ def active_mhv_ids
...
     {
-      street: address&.street,
-      city: address&.city,
-      state: address&.state,
-      country: address&.country,
-      zip: address&.postal_code
+      street: address[:street],
+      street2: address.instance_of?(Hash) ? address[:street2] : nil,
```
bosawt, line 159: I don't think we should use object introspection to determine the behavior of this field, it's not a very clear way to differentiate the fact that the `identity` version of this object has the field and the `mpi` version doesn't. 

You could add a more clear conditional in this function, or, one idea I had, we could just add `street2` to `mvi_profile_address.rb`. We won't currently write to it but it'd be fine just always being `nil` for now and maybe in the future we could investigate if there are multiline streets in MPI

bosawt (changes requested):
Just one comment to help improve code clarity that shouldn't be difficult to resolve, otherwise looks good!

bosawt (dismissed):
Looks good! Tested the address change and it appears to work both with address object sourced from identity, as well as mpi profile

rileyanderson (approved):

bramleyjl:
> lgtm. lmk when you want this merged and I will disable danger and merge it for you.

This is good to be merged now.

dillo (approved):
Lgtm

bosawt (approved):
👍

PR #8604, opened 2021-12-07
API-11753: SC notice attribute name change

ElleeB:
It was discussed that our `noticeAcknowledgement` attribute could do with a name upgrade. This PR presents an option, changing the attribute to `5103NoticeAcknowledged`.

Ticket: https://vajira.max.gov/browse/API-11753


cilestin (approved):
LGTM!

PR #8612, opened 2021-12-07
[27111] Log vista account parameters for debugging

ScottyRJames:
## Description of change
Running into negative argument error in prod with trying to build vista account id so we need to inspect what facility id looks like and length of vista account id
- Add log statement with following info
    - User uuid to separate issue by user
    - Facility id to assert vha facilities
    - Length of vista id to see if it matches requirement (only length due to sensitivity)

## Original issue(s)
[department-of-veterans-affairs/va.gov-team#27111](https://github.com/department-of-veterans-affairs/va.gov-team/issues/27111)

- PII concern with vista id but we're only logging the value length and not the actual id


LinnJS (approved):
🚀

PR #8623, opened 2021-12-08
API-9825: Add appoint_organization to claims v2 POA

hinzed1127:
## Description of change

This adds the ability to appoint an organization as POA in claims v2

## Original issue(s)
https://vajira.max.gov/browse/API-9825

- I changed the previous formatting of the route names ([explanation](https://github.com/department-of-veterans-affairs/vets-api/pull/8623#discussion_r765129007))
- [I also changed our POA verification logic](https://github.com/department-of-veterans-affairs/vets-api/pull/8623#discussion_r765111430). This one I want to really make sure I'm correct on, since this is getting utilized outside of just v2, and we have production consumers submitting to POA v1.
- I'll run the swagger updates right after getting approval. making sure everything else is in order first

### DIFF
--- a/modules/claims_api/app/controllers/claims_api/v1/application_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/v1/application_controller.rb
@@ -46,6 +46,10 @@ module ClaimsApi
         "#{user.first_name} #{user.last_name}"
       end
 
+      def user_is_target_veteran?
+        !header_request?
+      end
+
       private
 
       def claims_service
--- a/modules/claims_api/app/controllers/claims_api/v2/veterans/power_of_attorney_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/v2/veterans/power_of_attorney_controller.rb
@@ -19,16 +19,25 @@ module ClaimsApi
           )
         end
 
-        def appoint_individual # rubocop:disable Metrics/MethodLength
-          validate_request!(ClaimsApi::V2::ParamsValidation::PowerOfAttorney)
+        def appoint_organization
+          poa_code = parse_and_verify_poa_code
+          unless poa_code_in_organization?(poa_code)
+            raise ::Common::Exceptions::UnprocessableEntity.new(detail: 'POA Code must belong to an organization.')
+          end
 
-          poa_code = params.dig('serviceOrganization', 'poaCode')
-          validate_poa_code!(poa_code)
+          submit_power_of_attorney(poa_code)
+        end
 
-          if poa_code_in_organization?(poa_code)
+        def appoint_individual
+          poa_code = parse_and_verify_poa_code
+          unless poa_code_is_individual?(poa_code)
             raise ::Common::Exceptions::UnprocessableEntity.new(detail: 'POA Code must belong to an individual.')
           end
 
+          submit_power_of_attorney(poa_code)
+        end
+
+        def submit_power_of_attorney(poa_code) # rubocop:disable Metrics/MethodLength
           power_of_attorney = ClaimsApi::PowerOfAttorney.find_using_identifier_and_source(header_md5: header_md5,
                                                                                           source_name: source_name)
           unless power_of_attorney&.status&.in?(%w[submitted pending])
@@ -99,6 +108,15 @@ module ClaimsApi
                                                                     'va_eauth_service_transaction_id',
                                                                     'va_eauth_issueinstant',
                                                                     'Authorization').to_json)
+        end
+
+        def parse_and_verify_poa_code
+          validate_request!(ClaimsApi::V2::ParamsValidation::PowerOfAttorney)
+
+          poa_code = params.dig('serviceOrganization', 'poaCode')
+          validate_poa_code!(poa_code)
+
+          poa_code
         end
 
         def source_data
--- a/modules/claims_api/app/controllers/concerns/claims_api/poa_verification.rb
+++ b/modules/claims_api/app/controllers/concerns/claims_api/poa_verification.rb
@@ -97,9 +97,8 @@ module ClaimsApi
 
         representative = ::Veteran::Service::Representative.where('? = ANY(poa_codes)', poa_code).first
 
-        return false if representative.user_types.blank?
-
-        representative.user_types.include?('attorney') || representative.user_types.include?('claim_agents')
+        # As long as the representative has some kind of user_type, they're valid
+        representative.user_types.count > 0
       end
     end
   end
--- a/modules/claims_api/app/controllers/concerns/claims_api/poa_verification.rb
+++ b/modules/claims_api/app/controllers/concerns/claims_api/poa_verification.rb
@@ -83,12 +83,13 @@ module ClaimsApi
       def poa_code_in_organization?(poa_code)
         organization = ::Veteran::Service::Organization.find_by(poa: poa_code)
         representative = ::Veteran::Service::Representative.where('? = ANY(poa_codes)', poa_code).first
-        
+
         organization.present? && representative.user_types.include?('veteran_service_officer')
       end
 
       def poa_code_is_individual?(poa_code)
         return false if ::Veteran::Service::Organization.find_by(poa: poa_code).present?
+
         representative = ::Veteran::Service::Representative.where('? = ANY(poa_codes)', poa_code).first
 
         return false if representative.user_types.blank?
--- a/modules/claims_api/app/controllers/concerns/claims_api/poa_verification.rb
+++ b/modules/claims_api/app/controllers/concerns/claims_api/poa_verification.rb
@@ -82,9 +82,14 @@ module ClaimsApi
 
       def poa_code_in_organization?(poa_code)
         organization = ::Veteran::Service::Organization.find_by(poa: poa_code)
-        representative = ::Veteran::Service::Representative.where('? = ANY(poa_codes)', poa_code).first
 
-        organization.present? && representative.user_types.include?('veteran_service_officer')
+        if user_is_target_veteran?
+          organization.present?
+        else
+          representative = ::Veteran::Service::Representative.find_by(first_name: @current_user.first_name,
+                                                                      last_name: @current_user.last_name)
+          organization.present? && representative.poa_codes.include?(poa_code)
+        end
       end
 
       def poa_code_is_individual?(poa_code)
--- a/modules/claims_api/app/controllers/concerns/claims_api/poa_verification.rb
+++ b/modules/claims_api/app/controllers/concerns/claims_api/poa_verification.rb
@@ -98,7 +98,7 @@ module ClaimsApi
         representative = ::Veteran::Service::Representative.where('? = ANY(poa_codes)', poa_code).first
 
         # As long as the representative has some kind of user_type, they're valid
-        representative.user_types.count > 0
+        representative.user_types.count.positive?
       end
     end
   end
--- a/modules/claims_api/app/controllers/concerns/claims_api/poa_verification.rb
+++ b/modules/claims_api/app/controllers/concerns/claims_api/poa_verification.rb
@@ -81,11 +81,19 @@ module ClaimsApi
       end
 
       def poa_code_in_organization?(poa_code)
+        organization = ::Veteran::Service::Organization.find_by(poa: poa_code)
         representative = ::Veteran::Service::Representative.where('? = ANY(poa_codes)', poa_code).first
-        raise 'Power of Attorney not found' if representative.blank?
+        
+        organization.present? && representative.user_types.include?('veteran_service_officer')
+      end
+
+      def poa_code_is_individual?(poa_code)
+        return false if ::Veteran::Service::Organization.find_by(poa: poa_code).present?
+        representative = ::Veteran::Service::Representative.where('? = ANY(poa_codes)', poa_code).first
+
         return false if representative.user_types.blank?
 
-        representative.user_types.include?('veteran_service_officer')
+        representative.user_types.include?('attorney') || representative.user_types.include?('claim_agents')
       end
     end
   end
--- a/modules/claims_api/config/routes.rb
+++ b/modules/claims_api/config/routes.rb
@@ -66,10 +66,10 @@ ClaimsApi::Engine.routes.draw do
     namespace :veterans do
       get '/:veteranId/claims', to: 'claims#index'
       get '/:veteranId/claims/:id', to: 'claims#show'
-      get '/:veteranId/power_of_attorney', to: 'power_of_attorney#show'
-      put '/:veteranId/power_of_attorney/appoint_individual', to: 'power_of_attorney#appoint_individual'
-      put '/:veteranId/power_of_attorney/appoint_organization', to: 'power_of_attorney#appoint_organization'
-      get '/:veteranId/intent_to_files/:type',
+      get '/:veteranId/power-of-attorney', to: 'power_of_attorney#show'
+      put '/:veteranId/power-of-attorney/appoint-individual', to: 'power_of_attorney#appoint_individual'
+      put '/:veteranId/power-of-attorney/appoint-organization', to: 'power_of_attorney#appoint_organization'
+      get '/:veteranId/intent-to-files/:type',
           to: 'intent_to_files#type',
           constraints: (lambda do |request|
             ClaimsApi::V2::Veterans::IntentToFilesController::ITF_TYPES.include?(request.path_parameters[:type])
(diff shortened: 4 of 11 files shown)


modules/claims_api/app/controllers/concerns/claims_api/poa_verification.rb
```diff
@@ -81,11 +81,20 @@ def verify_representative_and_veteran(logged_in_representative_user, target_vete
...
 
       def poa_code_in_organization?(poa_code)
+        organization = ::Veteran::Service::Organization.find_by(poa: poa_code)
         representative = ::Veteran::Service::Representative.where('? = ANY(poa_codes)', poa_code).first
-        raise 'Power of Attorney not found' if representative.blank?
+
+        organization.present? && representative.user_types.include?('veteran_service_officer')
+      end
+
+      def poa_code_is_individual?(poa_code)
```
hinzed1127, line 90: it felt weird to not have this method as a contrast to `poa_code_in_organization?`, feels more logically readable in instances where we would call it.

Also, the logic is subtlety different between the 2. For individuals, we just need to check it _doesn_'t have a matching org but _does_ have a matching individual. By contrast, for orgs, we need the org code to exist _and_ have an individual with said code.

This check is more future proofed against anyone who might be both individually accredited _and_ acting for  an org. While that doesn't seem to occur, given the existing OGC, I think it's possible, and it's also how Tamara Ellis and John Doe are constructed.

modules/claims_api/config/routes.rb
```diff
@@ -66,9 +66,10 @@
     namespace :veterans do
       get '/:veteranId/claims', to: 'claims#index'
       get '/:veteranId/claims/:id', to: 'claims#show'
-      get '/:veteranId/power-of-attorney', to: 'power_of_attorney#show'
-      put '/:veteranId/power-of-attorney:appointIndividual', to: 'power_of_attorney#appoint_individual'
-      get '/:veteranId/intent-to-files/:type',
+      get '/:veteranId/power_of_attorney', to: 'power_of_attorney#show'
+      put '/:veteranId/power_of_attorney/appoint_individual', to: 'power_of_attorney#appoint_individual'
+      put '/:veteranId/power_of_attorney/appoint_organization', to: 'power_of_attorney#appoint_organization'
```
hinzed1127, line 71: using the previous format of  `'/:veteranId/power-of-attorney:appointIndividual'` was fine on its own, but something about this (I'm guessing the colon) breaks when adding in `'/:veteranId/power-of-attorney:appointOrganization'`, as well. The result was both of these request URLs just getting mapped to `appointIndividual`. 

I switched it to underscores to be more consistent with Ruby conventions, but I'm not married to that. It seems like there's already inconsistencies across Lighthouse, between us and the health APIs using camel case, and appeals stuff using underscores, so I'm fine with whatever as long as we're internally consistent

hinzed1127:
Final question: I know this is retreading old ground, but is [this](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/app/controllers/claims_api/v2/application_controller.rb#L93) _for sure_ the right logic for verifying access when changing POA?

I thought that another accredited representative can change the POA for a veteran (with their consent), but it looks like that's only true _if they are currently representing the veteran_. Just validating that that's indeed the right validation to enforce.

modules/claims_api/app/controllers/concerns/claims_api/poa_verification.rb
```diff
@@ -81,11 +81,20 @@ def verify_representative_and_veteran(logged_in_representative_user, target_vete
       end
 
       def poa_code_in_organization?(poa_code)
+        organization = ::Veteran::Service::Organization.find_by(poa: poa_code)
         representative = ::Veteran::Service::Representative.where('? = ANY(poa_codes)', poa_code).first
```
hinzed1127, line 85: I'm not sure if this is completely what we want here, at least according to the M21-1 manual. I think what we might actually want to check is this: **if the current user is a representative, verify that _they_ have a POA code for the submitted organization.**

### (A Lot More) Context

Under Section [I.i.2.E.1.e. What Constitutes a Complete VA Form 21-22 or VA Form 21-22a](https://www.knowva.ebenefits.va.gov/system/templates/selfservice/va_ssnew/help/customer/locale/en-US/portal/554400000001018/content/554400000181478/M21-1,-Part-I,-Subpart-i,-Chapter-2,-Section-E---Systems-Updates-for-Power-of-Attorney-(POA)-Appointments#1e), it lists the following as "Required Information:"

> 21-22: signature of the representative acting on behalf of the service organization with a corresponding date of signature.

However, it _also_ says this:

> Important: Completion of Blocks 16A and 16B of VA Form 21-22 is optional. If completed, there is no need to verify that the specific representative listed in Block 16A is accredited. This form documents the appointment of the entire service organization listed in Block 15 and does not limit action to a specific representative within the organization.

The field for block 16A is this:

>NAME OF OFFICIAL REPRESENTATIVE ACTING ON BEHALF OF THE
ORGANIZATION NAMED IN ITEM 15 (This is an appointment of the entire organization
and does not indicate the designation of only this specific individual to act on behalf of the
organization)

Block 16B that person's job title.

hinzed1127: i.e., I think this might be more along the lines of what we want to be checking instead:

```ruby
def poa_code_in_organization?(poa_code)
    organization = ::Veteran::Service::Organization.find_by(poa: poa_code)

    if user_is_target_veteran?
      organization.present?
    else
      representative = ::Veteran::Service::Representative.find_by(first_name: @current_user.first_name,
                                                                  last_name: @current_user.last_name)
      organization.present? && representative.poa_codes.include?(poa_code)
    end
end
```

modules/claims_api/app/controllers/claims_api/v2/veterans/power_of_attorney_controller.rb
```diff
@@ -101,6 +110,15 @@ def header_md5
...
         end
 
+        def parse_and_verify_poa_code
+          validate_request!(ClaimsApi::V2::ParamsValidation::PowerOfAttorney)
+
+          poa_code = params.dig('serviceOrganization', 'poaCode')
+          validate_poa_code!(poa_code)
+
+          poa_code
+        end
```
jw81, line 120: Disclaimer :: I'm still sick, so it's very possible that I'm not thinking clearly enough yet.

But I think this has uncovered a problem.  I don't think we're going to want the "appoint individual" and "appoint organization" endpoints to expect the same payload structure.  i.e. I don't think we should have the "appoint individual" endpoint looking for a poa_code that is nested in a `service_organization` attribute.  I think the 2 endpoints should expect different payload structures and that they should also have different validators.  Maybe I'm mistaken, or maybe this was already decided back during the original planning and I'm just forgetting it?  Maybe @SlakrHakr remembers?

modules/claims_api/config/routes.rb
jw81: I don't have a problem switching to underscores.  Consistency within the app is the most important thing I think.  👍

modules/claims_api/app/controllers/claims_api/v2/veterans/power_of_attorney_controller.rb
jw81: Ugh, actually now that I look at the form again and rethink this, maybe it is valid to allow them to share the same payload structure?  I mean, the actual section on the form is called, `SECTION III: SERVICE ORGANIZATION INFORMATION`, even though you can specify an individual or an organization in there.   🤦

modules/claims_api/spec/requests/v2/veterans/rswag_power_of_attorney_spec.rb
```diff
@@ -314,4 +316,172 @@
       end
     end
   end
+
+  path '/veterans/{veteranId}/power_of_attorney/appoint_organization' do
+    put 'Appoint an organization Power of Attorney for a Veteran.' do
+      tags 'Power of Attorney'
+      operationId 'appointIndividualPowerOfAttorney'
```
jw81, line 323: Prob want this to be `appointOrganizationPowerOfAttorney`.

modules/claims_api/app/controllers/claims_api/v2/veterans/power_of_attorney_controller.rb
jw81: Double-Ugh....I forgot that there are 2 different forms.  
- the [21-22A](https://www.vba.va.gov/pubs/forms/vba-21-22a-are.pdf) for appointing an individual
- the [21-22](https://www.vba.va.gov/pubs/forms/vba-21-22-are.pdf) for appointing an organization

And the 2 forms do have different fields for the "service organization" section.  So maybe it does make sense to have the 2 endpoints have different payload structures (and different validators).

jw81:
> 



> Related question: I know this is retreading old ground, but is [this](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/app/controllers/claims_api/v2/application_controller.rb#L93) _for sure_ the right logic for verifying access when changing POA?
> 
> I thought that another accredited representative can change the POA for a veteran (with their consent), but it looks like that's only true _if they are currently representing the veteran_. Just validating that that's indeed the right validation to enforce.

That seems correct to me.  It doesn't seem like we'd want to allow just any accredited person to do stuff on behalf of just any Veteran.

modules/claims_api/app/controllers/claims_api/v2/veterans/power_of_attorney_controller.rb
SlakrHakr: these 2 endpoints should definitely have different payload structures. thats part of the reason why they are designed to be 2 different endpoints as opposed to 1 single endpoint.

modules/claims_api/config/routes.rb
hinzed1127: > Nothing in building of an api should be done because of the standard of the language its being implemented in

That was poorly worded. I assumed that things weren't working for some `{Rails Magic}` reason, which it seems was the constraints bit. The lighthouse convention, at least on the the `vets-api/modules` end, was underscores, so it seemed more consistent. Between that and underscores, I'm mostly indifferent. I do kinda lean towards 
```
/:veteranId/power-of-attorney/appoint-individual
/:veteranId/power-of-attorney:appoint-organization
```
just because that format seems more common and doesn't require the constraints. But this was mainly to get things working and not meant to be a larger naming question, so I'm happy to go with y'all's consensus.

jw81: So this works as expected and doesn't require you to add the constraints argument?
```
/:veteranId/power-of-attorney/appoint-individual
/:veteranId/power-of-attorney:appoint-organization
```

If that's true, then I think this is fine.  I actually like the custom verb part not being snake-cased and instead being hyphenized like this.

modules/claims_api/app/controllers/claims_api/v2/veterans/power_of_attorney_controller.rb
jw81: I keep going back and forth on this.  One of the reasons I initially backed off of thinking they should have different payload structures, was because both forms have the same section headings.  So it kind of makes sense to allow the payload top-level attributes to map to the form section headings.  But maybe that's not actually a good idea.  Because these forms can change over time.  Form sections could be removed/added/renamed and then our payload structure isn't going to map to the form sections anymore.  Which could be confusing for consumers in the future.

Maybe we should use V2 as a chance to breakaway from tightly mapping our payload structures to the forms they represent?  Maybe the top-level attributes in our structures should be something more like, `veteranInfo`, `claimantInfo`, `representativeInfo`?

modules/claims_api/config/routes.rb
hinzed1127: @jw81 I just realized I made a typo. second link would just be
```
/:veteranId/power-of-attorney/appoint-organization
```
but yeah, I think it reads more cleanly too, so I'm down to just keep it as such.

hinzed1127:
>  It doesn't seem like we'd want to allow just any accredited person to do stuff on behalf of just any Veteran.

Yeah. I guess II was thinking of just one very specific scenario:
```
- Veteran has a POA
- They want to change it another individual or org
- New receiver of POA privileges makes that change for them with the Veteran's consent
```
But there's not really a good way to programmatically to know the consensual aspect, so the current constraints are probably the best to go by.

modules/claims_api/app/controllers/concerns/claims_api/poa_verification.rb
```diff
@@ -81,11 +81,24 @@ def verify_representative_and_veteran(logged_in_representative_user, target_vete
...
       def poa_code_in_organization?(poa_code)
+        organization = ::Veteran::Service::Organization.find_by(poa: poa_code)
+
+        if user_is_target_veteran?
+          organization.present?
+        else
+          representative = ::Veteran::Service::Representative.find_by(first_name: @current_user.first_name,
+                                                                      last_name: @current_user.last_name)
+          organization.present? && representative.poa_codes.include?(poa_code)
+        end
```
jw81, line 92: I'm not sure if it's because this PR is too big, or if there's too many different pieces being changed at once, but this logic concerns me.  It feels like a method named, `poa_code_in_organization?` should only be checking to see if the given POA code belongs to an organization and that's it.  It feels like if we need to also make sure that the authenticated person is allowed to use that POA code, then that should be an entirely separate check outside of this logic.

jw81:
For some reason, I'm having a very hard time parsing a lot of what's happening in this PR.  Which really concerns me because there's a couple parts where we're making changes to stuff that's shared across multiple versions of the API.

Prob not want you're gonna want to hear, but I'm wondering if it would be a good idea to maybe introduce this piece-by-piece?  For example, 1 PR that just gets a "HelloWorld" version of the new endpoint added.  i.e., it includes the routing and everything needed to simply hit the endpoint and get back a simple response.   Then another PR that introduces whatever changes are going to need to be made to existing logic that is shared across versions of the api.  Then another PR that introduces true logic into the new endpoint which will consume the shared logic that was changed.

Definitely not ideal, but like I said, there's enough different stuff happening here that I'm having a very hard time understanding it all and I'm afraid we're going to break some stuff.

modules/claims_api/app/controllers/concerns/claims_api/poa_verification.rb
hinzed1127: >`poa_code_in_organization?` should only be checking to see if the given POA code belongs to an organization and that's it.

I agree. I think this is more of a poorly named method, since the [existing method](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/app/controllers/concerns/claims_api/poa_verification.rb#L83) does more than that, as well. Or rather, it does it in the roundabout way, i.e., "is there a representative with this POA code _and_ do they have `veteran_service_officer` user type? 

I would propose instead that we refactor the existing method to be much smaller and move the rest of the logic downstream into the controller, which would be clearer. E.g.
```ruby
def poa_code_in_organization?(poa_code)
   ::Veteran::Service::Organization.find_by(poa: poa_code).present?
end
```
then, we're really addressing 2 scenarios, which right now is the `if-else` clause above:
1. If the request is being made by a veteran, is the POA code for an organization?
2. If the request is being made by a representative, is the POA for an organization **AND** is the rep authorized to act on behalf of that organization

This gets rid of some of the version-specific gaps I had to bridge with adding `user_is_target_veteran?` to v1. now that could just be a header_request?` check in the v1 controller and a `user_is_target_veteran?` check in v2. 

What are your thoughts around that approach?

modules/claims_api/config/routes.rb
jw81: Ah, hmmmmm.  I was thinking we were still going to use the custom verb syntax that Lee established.  I just thought you were going to switch the custom verb portion to hyphen-case instead of snakeCase.  If it's possible to keep the custom verb format and allowing the urls to continue to being hyphen-case, then I think we should do that.

modules/claims_api/app/controllers/claims_api/v1/application_controller.rb
```diff
@@ -46,6 +46,10 @@ def source_name
         "#{user.first_name} #{user.last_name}"
       end
 
+      def user_is_target_veteran?
+        !header_request?
+      end
+
```
jw81, line 52: This method bothers me.  It feels like it implies it does something more than simply looking for the existence of headers.  I would rather just call `!header_request?`, instead of calling this method.

modules/claims_api/app/controllers/concerns/claims_api/poa_verification.rb
jw81: Yeah, I'm all for making things simpler.  So if we need to go back to the existing method and clean it up so that we can make everything make more sense, then let's definitely do that.

hinzed1127:
Closing this, as I've split it out into 3 separate PRs:

- #8690 simplifies the `poa_in_organization?` logic
- #8694 adds the new endpoint
- #8695 Updates the swagger docs

PR #8637, opened 2021-12-09
1987 mobile map logingov for old versions

jperk51:
## Description of change
In order to logingov users to have access to everything they should in older versions of the app the LOGINGOV sign in method needs to be mapped to IDME

## Original issue(s)
department-of-veterans-affairs/va-mobile-app#1987

Added a new application controller test that checks the mapping occured correctly


modules/mobile/app/controllers/mobile/application_controller.rb
```diff
@@ -45,6 +45,7 @@ def authenticate
 
       session_manager = IAMSSOeOAuth::SessionManager.new(access_token)
       @current_user = session_manager.find_or_create_user
+      map_logingov_to_idme # check method comment below for more info
```
kpethtel, line 48: Style preference: I think you could leave this comment out. If anyone wants to know what's going on, they'll look for the method declaration.

kpethtel (approved):
👍

kpethtel (approved):

kreek (approved):
Normally I'd say do this in the IAM user identity model itself, but as it's temporary and only for mobile, the controller is the right spot 👍 and this LGTM 🚀

thilton-oddball (approved):

PR #8646, opened 2021-12-10
Update settings.yml

steveweinberger:
## Original issue(s)

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -863,6 +863,8 @@ flipper:
     - hocine.halli@accenturefederal.com
     - jacob.finnern@accenturefederal.com
     - brandon.clapp@accenturefederal.com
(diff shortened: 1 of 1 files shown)


omgitsbillryan (dismissed):
LGTM - but you need approval from the BE tools team

thilton-oddball:
Looks good but you may need to remove that empty line that was added or Rubocop is going to complain.

thilton-oddball (dismissed):

thilton-oddball:
I could be wrong but I think the problem might be that it looks like you are trying to merge in a change from a fork of vets-api, which I don't think it will let you do. You just want to create a branch off department-of-veterans-affairs/vets-api, make your change, and then merge your branch.

PR #8647, opened 2021-12-10
WIP: Add deps binstub and remove docker-compose-deps.yml

rileyanderson:
## Description of change
This removes the `docker-compose-deps.yml` and creates a binstub to set up dependencies. This is part of our work of moving towards using one `docker-compose` file and k8s.


bin/deps
```diff
@@ -0,0 +1 @@
+docker-compose up postgres redis
```
ericboehs, line 1: NEW LINE

docs/setup/hybrid.md
```diff
@@ -29,6 +29,6 @@ redis:
 
 ## Running
 
-1. To start Postgres and Redis: `docker-compose -f docker-compose-deps.yml up`
+1. To start Postgres and Redis: `bin/deps`
```
ericboehs, line 32: Another option is we just have them run the docker command here instead of creating bin/deps. 🤔

PR #8649, opened 2021-12-10
Filter multiple edipis

jscanish:
## Description of change
The goal here is to change how we filter `edipi` values. Instead of relying on `dodedipnid` we will instead fetch the value or values from `mvi`.

## Original issue(s)
[department-of-veterans-affairs/va.gov-team#0000](https://github.com/department-of-veterans-affairs/va.gov-team/issues/33748)

Confirmed change works through RSpec


bosawt (approved):
Looks good! I set a breakpoint on auth in `ssoe.rb` and confirmed that `edipi_ids[:edipi]` returned an `edipi`, and `edipi_ids[:edipis]` returned an array containing an `edipi`

LindseySaari (approved):

PR #8655, opened 2021-12-13
API-11876: make directDeposit.accountType a case insensitive enum

hinzed1127:
## Description of change

JSON schema doesn't support case insensitive enums, so we need to transform submissions _before_ schema validation. This transforms the `directDeposit.accountType` enum into title case before validating the value.

## Original issue(s)
https://vajira.max.gov/browse/API-11876


modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
```diff
@@ -143,6 +144,10 @@ def validate_form_526
 
         private
 
+        def sanitize_account_type
+          form_attributes['directDeposit']['accountType'] = form_attributes['directDeposit']['accountType'].titleize
```
jw81, line 148: We want this to be all uppercase right?

> Acceptance Criteria:
   Update v1 service for 526 to be case insensitive for DirectDeposit.AccountType enum
   _**Continue to convert to all caps behind the scenes to send to EVSS**_
   Update documentation and release notes to account for these changes

hinzed1127: 🤦 yup. Renamed to change the _enum_ back to title case, but not the submission

hinzed1127: [fixed](https://github.com/department-of-veterans-affairs/vets-api/pull/8655/commits/219f0888de5a621933a35cf3560d53dd53f72e3b)

jw81 (approved):

PR #8665, opened 2021-12-13
API-11285 Only try to send emails when flipper enabled as well

cilestin:
## Description of change
We were getting errors when trying to send emails on sandbox due to the fact that the delivery method changes depending on a FeatureFlipper setting, and the default delivery method errors when the FeatureFlipper is `false`. This change updates the various jobs we have to check the FeatureFlipper setting before trying to send emails to avoid throwing the errors.

## Original issue(s)
https://vajira.max.gov/browse/API-11285


ElleeB (approved):
Nice!

PR #8666, opened 2021-12-13
Covid vaccine trials -  Serializer to support update workflow

stephenBarrs:
This PR changes the Covid Vaccine Trials app to properly serialize and then send form data for both an initial form and an update form workflow.  

This PR is still using a local schema file as the actual `update` form is still in development.  Once that is complete, we will migrate the schema to `vets-json-schema`

statsD failure counts are updated to differentiate between the intake and update workflows.  I am not sure if there is additional work needed outside this module to support that, and will investigate further tomorrow. 

rspec tests added to cover the update workflow


modules/covid_research/app/controllers/covid_research/volunteer/submissions_controller.rb
```diff
@@ -10,8 +10,11 @@ class SubmissionsController < BaseController
       INTAKE_EMAIL_TEMPLATE_NAME = 'signup_confirmation.html.erb'
       UPDATE_EMAIL_TEMPLATE_NAME = 'update_confirmation.html.erb'
 
+      INTAKE_SCHEMA = 'COVID-VACCINE-TRIAL'
+      UPDATE_SCHEMA = 'COVID-VACCINE-TRIAL-UPDATE'
```
batemapf, line 14: just being nitpicky here, but looks like you are only using this vars once. why not just drop the strings into the FormService methods directly?

batemapf (approved):

modules/covid_research/app/controllers/covid_research/volunteer/submissions_controller.rb
stephenBarrs: I like to put string constants like these into constant variables at the top of the file so if/when we need to update the names it is very clear where that update should occur.  In this case it is a pretty simple file so probably not needed.

LindseySaari (approved):

PR #8667, opened 2021-12-13
[API-9829] - ClaimsApi - Adds ITF submission endpoint for V2

jw81:
## Description of change

Adds an ITF submission endpoint to V2 of `claims_api`.

## Original issue(s)
[API-9829](https://vajira.max.gov/browse/API-9829)

- This only adds the implementation/testing.  The documentation will come in another PR.  I'm doing this so we don't end up with a PR that exceeds the Lines of Code (LOC) limitation imposed by `dangerbot`.


modules/claims_api/spec/requests/v2/veterans/intent_to_files_request_spec.rb
```diff
@@ -7,6 +7,13 @@
 
   describe 'IntentToFiles' do
     describe 'type' do
+      before do
+        allow_any_instance_of(BGS::IntentToFileWebService)
+          .to receive(:find_intent_to_file_by_ptcpnt_id_itf_type_cd).and_return(
+            stub_response
+          )
+      end
+
```
hinzed1127, line 16: nice cleanup 🧹

modules/claims_api/lib/claims_api/v2/params_validation/intent_to_file/intent_to_file_validator.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+module ClaimsApi
+  module V2
+    module ParamsValidation
+      module IntentToFile
+        class IntentToFileValidator < ActiveModel::Validator
+          def validate(record)
```
hinzed1127, line 8: I think we're gonna want to also add validation logic around burial ITFs like [this](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/app/controllers/claims_api/v1/forms/intent_to_file_controller.rb#L110-L124).

hinzed1127 (changes requested):
Looks good, just need to add one more thing to the validator

modules/claims_api/lib/claims_api/v2/params_validation/intent_to_file/intent_to_file_validator.rb
jw81: Ah, if you check the AC on the ticket, you'll see that it was decided that we're not going to support `burial` ITF's in V2 because we don't have the capability of supporting Claimants who aren't Veterans.
> AC:

Allow submission of the following types of intent to files
compensation
pension
NOT burial (can't do this without supporting claimants who aren't Veterans)

hinzed1127: OPE

hinzed1127 (approved):
I stand corrected :homerdisappear:

hinzed1127 (approved):

Issue #8650, opened 2021-12-10
Simplify SAML::User class by removing non-SSOe attribute classes


bosawt:
`lib/saml/user.rb` uses a bit of a polymorphic pattern to store the `user_attributes`, but we only use one case for this now since we're 100% SSOe. It'd simplify the code dramatically to remove the unused `user_attribute` classes that aren't SSOe

nsoutouras:
Nick: Post SSOe cleanup work.  No user impact, just cleaning up dead code.  Tech debt.

PR #8669, opened 2021-12-14
[8650] Removing dead code with alternate login user attribute models …

bosawt:
…that predate SSOe

## Description of change
This PR removes currently unused user attribute models that were used when a user logged in with different credential providers. However, these predate SSOe, and are not relevant now that we are at 100% SSOe enablement

## Original issue(s)
department-of-veterans-affairs/vets-api#8650

- Should have no change in prod
- Testing on MHV account, DSLogon Account, ID.me account should be performed to confirm that everything still works as expected


bramleyjl (approved):
LGTM, tested locally with ID.me, Login.gov, & DSLogon accounts.

thilton-oddball (approved):

PR #8671, opened 2021-12-14
34059/34060 handle two more scenarios in coe_status

RudyOnRails:
## Description of change

- Adds two more scenarios and unit specs

## Original issue(s)

- https://github.com/department-of-veterans-affairs/va.gov-team/issues/34059
- https://github.com/department-of-veterans-affairs/va.gov-team/issues/34060


thilton-oddball (approved):
Looks good. One note, in the future you may want to avoid using quotes in a branch name. I was unable to pull the branch for testing until I finally realized it had quotes in it and they would need to be escaped.

RudyOnRails:
> Looks good. One note, in the future you may want to avoid using quotes in a branch name. I was unable to pull the branch for testing until I finally realized it had quotes in it and they would need to be escaped.

Ahh good to know. When creating a new branch I typically paste in the issue title into VS Code and it converts it to a git safe branch name, but I never thought about quotes. Haven't used them until this issue, so I'll make sure to avoid quotes.

Thanks Travis.

PR #8672, opened 2021-12-14
33936 use helper method to do address serialization

randomsync:
## Description of change

Use a helper method in appointment serializer for address serialization

## Original issue(s)
department-of-veterans-affairs/va.gov-team#33936

- [x] rspec
- [x] rubocop


modules/check_in/app/serializers/check_in/v2/appointment_data_serializer.rb
```diff
@@ -93,7 +54,20 @@ class AppointmentDataSerializer
...
+      def self.address_helper(address)
+        {
+          street1: address[:street1],
+          street2: address[:street2],
+          street3: address[:street3],
+          city: address[:city],
+          county: address[:county],
+          state: address[:state],
+          zip: address[:zip],
+          zip4: address[:zip4],
```
kanchanasuriya, line 67: Could you please confirm if the payload from downstream sends `mailing_address` and `home_address` with `nil` in `zip4`?  If there is no `zip4` key in the payload, this change might break the existing use-case.

randomsync: downstream (LoROTA) in this case doesn't send zip4 in some addresses (demographics). In that case, this would be nil and we'll pass it to the front-end with `"zip4": null,` in json.

randomsync:
@dillo @kanchanasuriya do you think this needs to be wrapped around a feature flag?

kanchanasuriya (approved):
LGTM

modules/check_in/app/serializers/check_in/v2/appointment_data_serializer.rb
dillo: This is correct. LGTM!

PR #8676, opened 2021-12-14
Limit External Key for BGS Service

kathleencrawford:
## Description of change

When instantiating the BGS service, we pass in an "external key" that is unique to the user.  This is either an email address or a common name (I think common name is used for SSO users).  Pretty regularly in Sentry, we are seeing: `The length of the EXTERNAL_UID or EXTERNAL_KEY exceeds the maximum 39 characters allowed`.  This PR limits the external key to 39 characters and adds some temporary logging so we can gather some data around which field is typically longer than 39 characters.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#28044

No new settings have been added, but some temporary logging was added to Sentry.

- [x] Local testing
- [x] Specs added


thilton-oddball (approved):
Looks good! This might resolve an issue I was working on yesterday. :crossed_fingers:

PR #8679, opened 2021-12-14
34057 Removing checkin v0 controller code

kanchanasuriya:
## Description of change

As first step of deprecating v0 endpoint, removing controller code, specs and api docs for check_in/v0.  Planning to open multiple PRs for removing services code, models and removing chip_api endpoint from settings.yml/test.yml

## Original issue(s)
department-of-veterans-affairs/va.gov-team#34057

## Things to know about this PR

- [X] rspec
- [X] rubocop

PR #8682, opened 2021-12-15
Update param name to match clinician_workflow

acrollet:
## Description of change

This is a follow-on to #8527 that updates the `nextOfConfirmedAt` param to match what's now in clinician_workflow. (cf. https://github.com/department-of-veterans-affairs/clinician_workflow/commit/0a940217ae97320e25fe68da9180c947ad203cf9)

⚠️ The merge for this PR will need to be co-ordinated with a similar change in CHIP ⚠️ 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#32876


randomsync:
looks good. You'll have to pull the latest from master into this branch to pass the lint check, as mentioned here: https://github.com/department-of-veterans-affairs/vets-api/pull/8679#issuecomment-994960099

randomsync (approved):

config/features.yml
```diff
@@ -528,7 +532,7 @@ features:
   va_online_scheduling_poc_health_apt:
     actor_type: user
     enable_in_development: true
-    description: Toggle for health apartment proof of concept   
+    description: Toggle for health apartment proof of concept
```
thilton-oddball, line 535: Just want to double check if this should be 'health appointment' as opposed to 'health apartment'

thilton-oddball (approved):
One possible typo that may need to be changed but otherwise looks good.

modules/check_in/spec/services/v2/chip/service_spec.rb
```diff
@@ -160,4 +160,56 @@
       end
     end
   end
+
+  describe '#demographic_confirmations' do
+    Timecop.freeze(Time.zone.now)
```
kanchanasuriya, line 165: Thanks for covering specs for the changes.  Came across this [doc](https://www.rubydoc.info/gems/liangzan-timecop/Timecop.freeze) earlier on `Timecop.freeze` changing system time.  

Could you please use code block while using Timecop as in [this spec](https://github.com/department-of-veterans-affairs/vets-api/blob/643c77ebd2f069f3915896cc93beccda09da374c/modules/check_in/spec/services/v2/lorota/redis_client_spec.rb#L80-L82)?  This will prevent system time change when the spec runs.

config/features.yml
dillo: That makes more sense. Looks like my editor just removed the white space. Other than that, I did not make any other modifications to the line.

modules/check_in/app/services/v2/chip/service.rb
```diff
@@ -77,14 +77,24 @@ def identifier_params
...
+              nextOfKinNeedsUpdate: next_of_kin_needs_update,
+              nextOfConfirmedAt: confirmed_at
+            }
           }
-        }
+
+        result.tap do |hash|
+          if Flipper.enabled?(:check_in_experience_chip_service_nok_confirmation_update_enabled)
+            hash[:demographicConfirmations].delete(:nextOfConfirmedAt)
+            hash[:demographicConfirmations].store(:nextOfKinConfirmedAt, confirmed_at)
```
kanchanasuriya, line 95: Just a thought, wondering if we can add `nextOfConfirmedAt` or `nextOfKinConfirmedAt` based on the feature flag instead of adding `nextOfConfirmedAt` & then deleting and adding `nextOfKinConfirmedAt` key.

modules/check_in/spec/services/v2/chip/service_spec.rb
dillo: I will go ahead and make the change so that we're using the block instead since this makes it clearer for someone reading the test. However, if you look toward the bottom, you can see that I'm returning `Timecop.return` which will put the time back.

modules/check_in/app/services/v2/chip/service.rb
dillo: Good thought! I think this would be the approach if we had implemented this key at the beginning with certain requirements in mind. However, the idea is not to break/touch the existing production hash. If we add the key now based on the condition of the feature toggle, then we would have modified the existing production hash as it's meant to be returned today.

modules/check_in/spec/services/v2/chip/client_spec.rb
```diff
@@ -103,7 +103,7 @@
           demographicsNeedsUpdate: true,
           demographicsConfirmedAt: '2021-11-30T20:45:03.779Z',
           nextOfKinNeedsUpdate: true,
-          nextOfConfirmedAt: '2021-11-30T20:45:03.779Z',
+          nextOfKinConfirmedAt: '2021-11-30T20:45:03.779Z',
```
randomsync, line 106: Do we need to explicitly enable the `check_in_experience_chip_service_nok_confirmation_update_enabled` feature flag in a `before` block for this?

randomsync (approved):
lgtm, one minor comment.

modules/check_in/spec/services/v2/chip/service_spec.rb
kanchanasuriya: Thanks for updating Timecop to use blocks @dillo

modules/check_in/app/services/v2/chip/service.rb
kanchanasuriya: Synced offline on the change

kanchanasuriya (approved):
LGTM

modules/check_in/spec/services/v2/chip/client_spec.rb
dillo: I think this is just an example demographics structure that we're passing into the Client instances pre_check_in method. In both test cases, I don't think we're making use of the actual keys or values within. The entire hash just gets turned into a JSON and we're testing a couple of outcomes. We could technically pass in a blank hash and I think the test cases will pass without any issues. However, I think it would be nice if we can pass in a fully qualified structure that illustrates to the user what the contents may actually be.

dillo (approved):
Everyone says my code LGTM

randomsync (approved):

thilton-oddball (approved):

PR #8684, opened 2021-12-15
Filter multiple corpids

jscanish:
## Description of change
Similar to https://github.com/department-of-veterans-affairs/vets-api/pull/8649. Using parsed `MVI` value to determine if there are multiple `corp_ids`.

## Original issue(s)
[department-of-veterans-affairs/va.gov-team#0000](https://github.com/department-of-veterans-affairs/va.gov-team/issues/33830)

RSpec to confirm no changes in behavior.


bosawt (approved):
Looks good! I tested with a pry on localhost auth in `ssoe.rb`, `mvi_ids[:vba_corp_ids]` returns an array of `corp_id`, and `corp_id_mismatch?` is `false` if there are zero, or one corp_ids, and true if there are multiple

LindseySaari (approved):

PR #8687, opened 2021-12-15
API-11692 Add comments to generator spec re: radio buttons

ElleeB:
Our pdf generator spec cannot match, and thus cannot be relied upon to check, radio buttons. This PR adds comments to the hlr v2 portion of the spec to alert future devs of this.


cilestin (changes requested):
On the right track, but I don't think we need to specify which radio buttons need to be manually checked, since they could change as the forms change. Just a "Be sure to check radio button differences manually" or something, instead :)

Let's also add some comments to the `match_pdf` matcher itself about _why_ it can't check against radio buttons (we flatten the pdf early in the process for upstream submission/security reasons).

cilestin (approved):

Issue #8691, opened 2021-12-16
Add verified_at and query protections for UserVerification/UserAccount after login actions


bosawt:
### Description
This is part of an effort to create a more persistent user layer strategy, ultimately giving users a more consistent experience after they have logged in.

PR #8692, opened 2021-12-16
[8691] Adding verified_at functionality and query protections for use…

bosawt:
…r_verifications and user_actions

## Description of change
This PR adds a verified_at timestamp to `UserVerification` models that occurs when an associated `UserAccount` is added or updated with an `ICN` (each time it is changed, `verified_at` will update)

## Original issue(s)
department-of-veterans-affairs/vets-api#8691

- This should be covered in specs
- This doesn't currently do anything on prod, so to test you'll have to sign in with a user and inspect the `UserVerification` and `UserAccount` records for that user
- Confirm `verified_at` is on the `user_verification` after signing in with a verified account


thilton-oddball (approved):
Looks good. You'll just need to merge master in to get the Security and Linting Check to pass.

bosawt:
@thilton-oddball Can I get a re-review? Had to merge master in as you asked

thilton-oddball (approved):

PR #8699, opened 2021-12-16
API-9829 - ClaimsApi - Adds docs for V2 ITF POST endpoint

jw81:
## Description of change

- Adds docs for new V2 ITF POST endpoint
- Versions the existing `SwaggerSharedComponents` logic.  This will help protect us from making changes to doc generation that would accidentally permeate across multiple versions of the API.

## Original issue(s)
[API-9829](https://vajira.max.gov/browse/API-9829)


hinzed1127 (approved):
Good call on versioning the shared components

PR #8703, opened 2021-12-16
Hypertension Fast Track JSON key fix

tadhg-ohiggins:
## Description of change
Missing keys broke the original code here, we're using `.dig` and `|| []` to account for those. This PR also adds a test for the prior broken case.

-----

Don't just reach blindly;
Instead, dig into the keys
In case we lack some.


NatashaJIbrahim (approved):

app/services/fast_track/hypertension_medication_request_data.rb
```diff
@@ -32,8 +32,8 @@ def get_notes_from_note(verbose_notes)
     end
 
     def get_text_from_dosage_instruction(dosage_instructions)
-      toplevel_texts = dosage_instructions.map { |instr| instr['text'] }
-      code_texts = dosage_instructions.map { |instr| instr['timing']['code']['text'] }
+      toplevel_texts = dosage_instructions.map { |instr| instr['text'] || [] }
```
julialeague, line 35: Should we dig here too?

tadhg-ohiggins: No need, if `'text'` doesn't exist here it will return `[]`.

thilton-oddball (approved):

PR #8705, opened 2021-12-16
HTN Fast Track: account for logingov_uuid issue in ICN lookup

tadhg-ohiggins:
## Description of change
See https://dsva.slack.com/archives/C0229UT3HSR/p1638223547049000 for background. We need to support looking up `Accounts` using alternatives to `idme_uuid`, which is what this adds.

-----

No I D dot me
Means try login dot gov, then
E D I P I.


app/workers/fast_track/disability_compensation_job.rb
```diff
@@ -56,7 +56,10 @@ def send_fast_track_engineer_email_for_testing(form526_submission_id)
     end
 
     def get_icn(form526_submission)
-      Account.where(idme_uuid: form526_submission.user_uuid).first.icn
+      account = Account.where(idme_uuid: form526_submission.user_uuid).first
+      account = Account.where(logingov_uuid: form526_submission.user_uuid).first if account.blank?
+      account = Account.where(edipi: form526_submission.auth_headers['va_eauth_dodedipnid']).first if account.blank?
+      account.icn if account.present? && account.icn.present?
```
julialeague, line 62: Does anyone know what the `sec_id` attribute is in Account? It looks like right now the uuid could live under that attribute when it's not found in `logingov_uuid` or `idme_uuid`.

thilton-oddball (approved):

PR #8558, opened 2021-12-01
Add hypertension fast track flow to Form 526 submission

tadhg-ohiggins:
## Description of change
This PR alters the beginning of the existing Form 526 submission process in order to add a PDF of health data to hypertension single-issue claims for increase. The PDF contains blood pressure readings and medication history.

## Original issue(s)
This is a new feature, not covered by an issue.

## Things to know about this PR
The entire feature is behind a Flipper flag, `disability_hypertension_compensation_fast_track`.

The altered flow checks whether the submission is a single-issue hypertension claim for increase. 

If it is,  the claim will follow what we'll refer to as the “hypertension path,” detailed below.
If it is not, then the prior Form 526 submission flow is kicked off and no further changes are made.

If there are errors, failures, or sufficient timeouts on the hypertension path, it will kick off the prior Form 526 submission flow without further changes.

If the hypertension path succeeds, it will kick off the prior Form 526 submission flow. In this case the submitted form JSON will been altered in two ways:

-   The `disabilities` field will have been altered to add `RRD`to the list of special issues for disabilities that have a `diagnosticCode` of 7101 and a `disabilityActionType` of “increase”.
-   The Form 526 uploaded files will have had a newly-generated PDF added to them.

The hypertension path does the following:

-   Looks up the ICN of the person who submitted the claim.
-   Uses that ICN to look up the person’s blood pressure and medication data in the Lighthouse Veterans Health API.
-   Checks to see if there are recent blood pressure readings. If not, it exits and returns to the prior Form 526 submission flow.
-   Filters and sorts the blood pressure readings and medications to make them suitable for summarization in a PDF.
-   Generates a PDF with that data using the Prawn library.
-   Stores that PDF in S3 via the existing `SupportingEvidenceAttachment` class.
-   Adds the information about that PDF to the submission's `form_json` field.
-   Adds the special issue `RRD` to the submission's `form_json` field. (The RRD special issue is used by EVSS and NWQ further in the claims process to route the claim to the appropriate regional office)
-   Saves the submission.
-   Returns the submission to the prior Form 526 submission flow.

The hypertension path is started in `app/models/form526_submission.rb`, in the `start` method. Success or failure results in calling the `start_evss_submission` method, which is the prior Form 526 submission flow.

The work of the hypertension path is in a worker, and starts in `app/workers/disability_compensation_fast_track_job.rb` in the `DisabilityCompensionFastTrackJob` class's `perform` method.

[...]

PR #8708, opened 2021-12-16
Change Form526Submission full name retrieval to fall back to auth headers values

julialeague:
## Description of change
This PR is another follow up to #8558.
While doing end-to-end testing, we saw some full names being registered as nil when we rely solely on the User object to retrieve the full name. We saw at least one submission in production that had name values in the Form526Submission auth headers while it did NOT have an existing User associated with the submission.

We decided to update the full_name method to account for this (hopefully) edge case by falling back to the auth header name values if they're available when the user full name isn't.

Added RSpec tests to account for the changes.


spec/models/form526_submission_spec.rb
```diff
@@ -614,8 +614,47 @@
...
+        end
+      end
+    end
+
+    context 'when the User is NOT found' do
+      before { allow(User).to receive(:find).and_return nil }
+
+      it 'returns the first and last name of the user from the auth headers' do
+        expect(subject.full_name).to eql(full_name_hash.merge(middle: nil, suffix: nil))
+      end
```
NatashaJIbrahim, line 657: Great specs 👍

NatashaJIbrahim (approved):

thilton-oddball (approved):

PR #8711, opened 2021-12-17
2004 paginate vaccines

kpethtel:
## Original issue(s)


modules/mobile/app/controllers/mobile/v0/claims_and_appeals_controller.rb
```diff
@@ -20,7 +20,7 @@ class ClaimsAndAppealsController < ApplicationController
       end
 
       def index
-        validated_params = validate_params(params)
```
kpethtel, line 23: In addition to the pagination changes, I changed some of the params handling, because params is already defined in the controller and I found passing them around under the arg name "params" confusing.

modules/mobile/app/controllers/mobile/v1/immunizations_controller.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
```
kpethtel, line 1: Not a part of the initial PR to separate out pagination changes.

modules/mobile/app/models/mobile/v0/contracts/get_paginated_list.rb
```diff
@@ -7,10 +7,11 @@ class GetPaginatedList < Base
         params do
           optional(:start_date).maybe(:date_time, :filled?)
           optional(:end_date).maybe(:date_time, :filled?)
```
kpethtel, line 9: Unclear how we we want to handle this, but this is one approach. AFAICT, we're currently treating page number and page size as required because the existing pagination approach would break if they were missing. We can handle this a few ways. We can do some combination of 1) change the pagination code to work without them 2) raise an error in the paginator if they're missing 3) require them here on the assumption that we'll always use this for the paginator. I'm open to suggestion here.

modules/mobile/app/helpers/mobile/pagination_helper.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+
+    def link_url(display_page_number)
+      "#{url_without_page_number}&page[number]=#{display_page_number}"
+    end
+
+    def url_without_page_number
+      @url_without_page_number ||= begin
+        query_strings = []
+        %w(start_date end_date use_cache show_completed).each do |key|
```
kpethtel, line 58: Note that reverse_sort is not included because the current implementation did not handle it.

modules/mobile/app/serializers/mobile/v1/immunization_serializer.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
```
kpethtel, line 1: Not a part of the initial PR to separate out pagination changes.

modules/mobile/config/routes.rb
```diff
@@ -64,4 +64,8 @@
       end
```
kpethtel, line 64: Not a part of the initial PR to separate out pagination changes.

modules/mobile/spec/request/v1/immunizations_request_spec.rb
```diff
@@ -0,0 +1,493 @@
+# frozen_string_literal: true
```
kpethtel, line 1: Not a part of the initial PR to separate out pagination changes.

modules/mobile/spec/request/v1/immunizations_request_spec.rb
```diff
@@ -0,0 +1,493 @@
...
+
```
kpethtel, line 2: The existing specs are not namespaced under v0. Do we want to add that namespace or come up with a different scheme?

modules/mobile/app/helpers/mobile/pagination_helper.rb
```diff
@@ -0,0 +1,71 @@
...
+          }
+        },
+        links: links(pages.size)
+      }
+      return [[], page_meta_data] if page_number > pages.size
+
+      [pages[page_number - 1], page_meta_data]
+    end
+
+    def links(number_of_pages)
```
kpethtel, line 38: I extracted the links helper into this because the two seem tightly coupled.

modules/mobile/app/helpers/mobile/pagination_helper.rb
```diff
@@ -0,0 +1,71 @@
+module Mobile
```
kpethtel, line 1: Might be worth adding explicit tests for this rather than just relying on request specs.

modules/mobile/app/controllers/mobile/v1/immunizations_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+
+      def service
+        Mobile::V0::LighthouseHealth::Service.new(@current_user)
+      end
+
+      def pagination_params
+        @pagination_params ||= Mobile::V0::Contracts::GetPaginatedList.new.call(
+          page_number: params.dig(:page, :number) || 1,
+          page_size: params.dig(:page, :size) || 10,
+          # use_cache: params[:use_cache] || true
```
kpethtel, line 30: The one thing I haven't figured out yet is the caching. I added redis to the immunization model and followed the pattern in the other controllers, but I kept getting an error. I'll try again today, but if it continues to be a challenge, I may just break it out into a separate ticket, since it's not a requirement of this ticket.

modules/mobile/spec/request/appointments_list_request_spec.rb
```diff
@@ -132,11 +132,11 @@
           it 'has the correct links with no prev' do
```
kpethtel, line 132: These changes are all because the pagination helper orders the params differently.

modules/mobile/app/controllers/mobile/v0/appointments_controller.rb
```diff
@@ -13,21 +13,23 @@ def index
...
         end_date = params[:endDate] || one_year_from_now.iso8601
-        page = params[:page] || { number: 1, size: 10 }
         reverse_sort = !(params[:sort] =~ /-startDateUtc/).nil?
 
         validated_params = Mobile::V0::Contracts::GetPaginatedList.new.call(
           start_date: start_date,
           end_date: end_date,
-          page_number: page[:number],
-          page_size: page[:size],
+          page_number: params.dig(:page, :number) || 1,
```
kpethtel, line 21: I think it could be worth moving these defaults into the pagination helper. Currently, these are required here because I made them required in `GetPaginatedList`. This might be evidence against making them required.

modules/mobile/app/models/mobile/v0/contracts/get_paginated_list.rb
kreek: I think these should remain optional and that we should do 1) change the pagination code to work without them. Otherwise I think we'd have to version appointments and claims too because the contract changed (even though we're fixing the code to not to break if they're missing).

modules/mobile/app/controllers/mobile/v0/appointments_controller.rb
kreek: As mentioned in the comment below, I think moving these defaults into the pagination helper is the approach to take.

PR #8719, opened 2021-12-17
add tracer config for datadog

jsgarmon:
Tell app where to send traces to


LindseySaari (approved):

f1337 (approved):
Ship it! :shipit:
Guinea pigs let out of a pen and running like lemmings

PR #8728, opened 2021-12-20
Add statsd metric to track missing date and cvx code from immunizatio…

aherzberg:
…ns as well as corresponding tests

## Description of change

Adds statsd metric to count how many times a date or cvx code is missing from immunization data. 

## Original issue(s)

https://github.com/department-of-veterans-affairs/va-mobile-app/issues/2085

Borrowed most of the code from @cadibemma branch: 2085-missing-date-cvx-code-metrics but created my own as to not step all over their branch.


modules/mobile/app/models/mobile/v0/adapters/immunizations.rb
```diff
@@ -31,6 +31,16 @@ def parse(immunizations)
 
         private
 
+        def date(immunization)
+          date = immunization[:occurrence_date_time]
+          date.presence ? date : StatsD.increment('mobile.immunizations.date_missing') && nil
```
kpethtel, line 36: I'd use `.present?` instead of `.presence`. Using the `&&` in a ternary also seems strange to me, but I don't think there's anything wrong with it. I tend to be hesitant about using && for control flow because I've been caught before by cases where the first clause returned a falsey value so the second clause doesn't execute. But I don't think that should be an issue here.

kpethtel (dismissed):
I left one small comment. I also like what you've done with the specs and the vcr cassette to slim it down and make it more readable. I can't really read the diff for `modules/mobile/docs/index.html`, so I'm going to have to take your word on it. I assume you've tested that locally and it is what we want.

modules/mobile/app/models/mobile/v0/adapters/immunizations.rb
aherzberg: I typically go for `.present?` as well but the linter was correcting me on it and wanted to do ` date.presence || StatsD.increment('mobile.immunizations.date_missing') && nil ` instead so maybe it is just easier to avoid the ternary all together. For the index.html, yes, I locally reviewed that and it looks good. Got guidance from Jayson on how to represent the nullable field as well to ensure we're consistent.

aherzberg: decided to just take out the ternary and put a normal if else statement

kpethtel: I guess I can see the linter's point of view. Its suggestion should achieve the same thing, but it does feel less obvious when paired with the && for control flow. I think the fourth option we can add to the list of ways we can handle this is with a guard: 
```
return date if date.present?

StatsD.increment('mobile.immunizations.date_missing')
nil
```

So many options. This might be why some people don't like ruby. :)

kpethtel (approved):

modules/mobile/app/models/mobile/v0/adapters/immunizations.rb
aherzberg: That's much cleaner. Updated to use that styling but did if `date.blank?` rather then `unless date.present?`. Also did the same for cvx_code.

I updated the documentation to make the new attribute reaction nullable as well to stay consistent with the other attributes.

modules/mobile/spec/request/immunizations_request_spec.rb
```diff
@@ -471,6 +471,78 @@
...
+        end
+
+        it 'returns a 200' do
+          immunizations_request_missing_cvx
+          expect(response).to have_http_status(:ok)
+        end
+
+        it 'sets cvxCode to nil' do
+          immunizations_request_missing_cvx
+          expect(response.parsed_body['data'][0]['attributes']).to eq(
```
kpethtel, line 495: This test and the matching one below are susceptible to churn if the logic behind any other attributes change. For example, if we add or remove an attribute, these tests will need to be adjusted. It might be better to just test the specific attribute we care about like:
`expect(response.dig('data', 0, 'attributes', 'cvxCode').to be_nil`

kpethtel (approved):
One small comment, but it looks good.

modules/mobile/spec/request/immunizations_request_spec.rb
aherzberg: Good point. I made this change but for cvxcode I kept the manufacturer and group name check for nil since those are derived from the cvxcode.

kpethtel (approved):
👍

PR #8736, opened 2021-12-20
[34382] Add zero routing number check to ch33 call

bosawt:
## Description of change
This PR adds a check for ch33 bank accounts implementation so that when the routing number for a find_ch33_dd_eft call is `0` or `nil`, we don't make a follow up call to get the bank name from routing number.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#34382

- I tested this with a user that doesn't have bank account information, was able to see that `ch33_bank_accounts_controller` rendered response did not cause errors, although the bank information was set to `nil`, as expected
- Additionally, I no longer see a stack trace logged since the second call to get bank name from routing number is no longer made in this case


tpharrison:
@bosawt - thanks for keeping me in the loop.  Changes look fine and spec passes, but I’m very green.  Is there anything else I should check? Clicking "view deployment" gives me a 502 bad gateway and my local setup isn’t quite finished.  Not sure if the intent was for me to approve or just keep eyes on it.

bosawt:
> @bosawt - thanks for keeping me in the loop. Changes look fine and spec passes, but I’m very green. Is there anything else I should check? Clicking "view deployment" gives me a 502 bad gateway and my local setup isn’t quite finished. Not sure if the intent was for me to approve or just keep eyes on it.

Ah yes, to check with a PR instance deployment (the instance that pops up when you click `View Deployment`), you need SOCKS proxy access. The behavior you're seeing is the case for PRs when you're attempting to access it without the SOCKS proxy

bosawt:
> @bosawt - thanks for keeping me in the loop. Changes look fine and spec passes, but I’m very green. Is there anything else I should check? Clicking "view deployment" gives me a 502 bad gateway and my local setup isn’t quite finished. Not sure if the intent was for me to approve or just keep eyes on it.

To be honest, as long as you recognize this is a change that has an effect on the `Ch33BankAccountsController`, that's probably good enough for now. It's a pretty small change so I'm not too particularly worried about it having some unexpected regression or anything

lib/bgs/service.rb
```diff
@@ -192,7 +196,7 @@ def get_ch33_dd_eft_info
     end
 
     def find_bank_name_by_routng_trnsit_nbr(routing_number)
-      return if routing_number.blank?
+      return if routing_number.blank? || routing_number == EMPTY_ROUTING_NUMBER
```
RudyOnRails, line 199: I was thinking you could also do this
```suggestion
      return if routing_number.to_i.zero?
```

but the constant leaves an opportunity to separate business related logic and make a great supporting comment like you did.

lib/bgs/service.rb
```diff
@@ -16,6 +16,10 @@ class Service
     # (I = Input, U = Update, D = Delete)
     JOURNAL_STATUS_TYPE_CODE = 'U'
 
+    # It appears that a find_ch33_dd_eft that returns empty bank account information
+    # will set the routing number field to '0' instead of 'nil', at least in certain cases
+    EMPTY_ROUTING_NUMBER = '0'
```
RudyOnRails, line 21: Would we want to freeze this string?
```suggestion
    EMPTY_ROUTING_NUMBER = '0'.freeze
```

RudyOnRails (approved):
Nice work @bosawt - this looks like it will remove a lot of extra calls to BGS.

lib/bgs/service.rb
bosawt: Not a bad idea, I'm gonna leave as-is for now but there might be an opportunity to go further with this check (figuring out the entire range of possibilities for the routing number and watching out for those)

thilton-oddball (approved):

tpharrison (approved):

lib/bgs/service.rb
RudyOnRails: Ahh, right, must be the `frozen_string_literal: true` kicking in. Sorry about that.

PR #8738, opened 2021-12-20
2004 implement immunizations pagination

kpethtel:
## Description of change
Adds a new v1 immunizations route with pagination. 

## Original issue(s)
department-of-veterans-affairs/va-mobile-app/issues/2004

This is not yet utilized by the front end. 
This does not include documentation or caching. Those will be broken into separate tickets.


jperk51 (dismissed):
LGTM

modules/mobile/spec/request/v1/immunizations_request_spec.rb
```diff
@@ -0,0 +1,538 @@
+# frozen_string_literal: true
```
kpethtel, line 1: This is mostly just a copy of the v0 version of the tests. Because v0 is being deprecated, those tests shouldn't be changing in the future, so I didn't want to create any shared tests.

modules/mobile/spec/request/v1/immunizations_request_spec.rb
```diff
@@ -0,0 +1,538 @@
...
+            },
+            'links' => {
+              'self' => 'http://www.example.com/mobile/v1/health/immunizations?page[size]=100&page[number]=1',
+              'first' => 'http://www.example.com/mobile/v1/health/immunizations?page[size]=100&page[number]=1',
+              'prev' => nil,
+              'next' => nil,
+              'last' => 'http://www.example.com/mobile/v1/health/immunizations?page[size]=100&page[number]=1'
+            } }
+        )
+      end
```
jperk51, line 347: if you still have LOC issues you could just pull this check out then make a specific PR for it after the rest goes in

jperk51 (approved):
LGTM

PR #8741, opened 2021-12-21
API-11754: Update claims mailer to include metrics for POA endpoints

hinzed1127:
## Description of change

localhost_3000_rails_mailers_claims_api_unsuccessful_report_mailer_build (1)

- Adds in POA metrics.
- ~ITF forthcoming, just wanted to validate my approach first~ It turns out we don't actually create any kind of `ActiveRecord` storage, so we can't add that here at this time. TBD if that's something we're interested in or if Kibana and other logging is sufficient there

## Original issue(s)
https://vajira.max.gov/browse/API-11754

This looks like it changes more than it really does. I did a lot of variable renaming to make things explicit between `claims` and `poa` things. This will also help when `ITF` is added into the mix


modules/claims_api/app/views/claims_api/unsuccessful_report_mailer/unsuccessful_report.html.erb
```diff
@@ -30,81 +30,77 @@
...
-    <h2>
-      <%= @unsuccessful_submissions.count %>
-      <span class="title">Errorred Submissions</span>
-    </h2>
-    <%= render partial: 'submission_table', locals: { claims: @unsuccessful_submissions } unless @unsuccessful_submissions.count.zero? %>
+    <h3>
+      <%= @unsuccessful_poa_submissions.count %>
+      <span class="title">Errored Submissions</span>
+    </h3>
+    <%= render partial: 'poa_errors_table', locals: { poa_errors: @unsuccessful_poa_submissions } unless @unsuccessful_poa_submissions.count.zero? %>
```
hinzed1127, line 104: this is weird formatting by the git diff. This is new for `poa_errors`. _Claims_ submission errors got renamed in lines 84-88 above

modules/claims_api/lib/claims_api/vbms_sidekiq.rb
```diff
@@ -6,7 +6,7 @@ module ClaimsApi
   module VBMSSidekiq
     def rescue_file_not_found(power_of_attorney)
       power_of_attorney.update(
-        status: 'failed',
+        status: ClaimsApi::PowerOfAttorney::ERRORED,
```
hinzed1127, line 9: this seemed like a one-off use of `"failed"` as the status value. I refactored this to the `ERRORED` constant to consolidate all POA errors together

jw81 (approved):

PR #8742, opened 2021-12-21
2004: additional immunizations specs

kpethtel:
## Description of change
These are simply additional specs that would not fit under 500 LoC in the previous PR for this ticket. 

## Original issue(s)
department-of-veterans-affairs/va.mobile-team#2004

Nothing. These are just additional tests.


jperk51 (approved):
LGTM

PR #8779, opened 2021-12-29
2242 cache immunizations

kpethtel:
## Description of change
Adds caching to v1 immunizations

## Original issue(s)
department-of-veterans-affairs/va-mobile-app/2242

This endpoint is not yet in use.


modules/mobile/app/controllers/mobile/v1/immunizations_controller.rb
```diff
@@ -29,9 +27,30 @@ def service
...
+        else
+          immunizations = immunizations_adapter.parse(service.get_immunizations)
+          Mobile::V0::Immunization.set_cached(@current_user, immunizations)
+          Rails.logger.info('mobile immunizations service fetch', user_uuid: @current_user.uuid)
+        end
+
+        immunizations.reverse
+      end
+
+      def use_cache?
```
kpethtel, line 49: It wasn't clear if we should add this param or not, but we do use it elsewhere and I thought it could be helpful for testing purposes to be able to bust the cache. I'm definitely open to removing the param.

modules/mobile/app/controllers/mobile/v1/immunizations_controller.rb
```diff
@@ -27,11 +25,26 @@ def service
       end
 
       def pagination_params
-        Mobile::V0::Contracts::GetPaginatedList.new.call(
+        @pagination_params ||= Mobile::V0::Contracts::GetPaginatedList.new.call(
           page_number: params.dig(:page, :number),
-          page_size: params.dig(:page, :size)
+          page_size: params.dig(:page, :size),
+          use_cache: params[:useCache] || true
```
kpethtel, line 31: Unclear if we really need to add this param rather than just always using the cache, but it might be useful for debugging.

modules/mobile/app/controllers/mobile/v1/immunizations_controller.rb
jperk51: I think adding it is good for the purpose you mentioned.

modules/mobile/app/controllers/mobile/v1/immunizations_controller.rb
jperk51: Same as above.

LindseySaari (approved):

jperk51 (approved):
LGTM

PR #8781, opened 2021-12-30
Feature submit claim

s-caso:
## Description of change
This PR is the initial iteration of submitting  a claim to vets-service.  There will need to be changes made eventually once the front end is updated, namely updating the params structure and removing hard coded values.

## Original issue(s)
[department-of-veterans-affairs/va.gov-team#0000](https://jira.np.afsp.io/browse/SA-839)

Swagger Docs will be updated in a follow up PR.

Request spec was added and will be manually tested via postman.


modules/meb_api/app/controllers/meb_api/v0/education_benefits_controller.rb
```diff
@@ -26,10 +27,14 @@ def claim_status
...
-        render json:
-               { data: {
-                 'status': 'received'
-               } }
+        response = submission_service.submit_claim(params)
+
+        # @NOTE: Need front end to send in_progress_form_id as well as claimant_id
+        render json: {
+          data: {
+            'status': response.status
```
LindseySaari, line 35: Do you need to define the serializer: here?

s-caso: For the time being this is just sending a 200 back so i wasn't sure if it was needed.  I can add one though if it's the convention.

LindseySaari: I could go either way but if you will need it in the future, you may want to explicitly define it or make some sort of comment for future use

s-caso: Sounds good, I added to the note to mention the serializer.

LindseySaari (approved):

PR #8803, opened 2022-01-05
API-12124: Part I: Add claimant form fields hlr v2

ElleeB:
## Description of change
Updates HLR form fields to include claimant related methods and removes leftover methods for fields that are stamped and auto-sized.

## Original issue(s)
https://vajira.max.gov/browse/API-12124


modules/appeals_api/app/services/appeals_api/pdf_construction/higher_level_review/v2/form_fields.rb
```diff
@@ -44,18 +44,6 @@ def insurance_policy_number
           'form1[0].#subform[2].ClaimantsLastName[1]'
         end
 
-        def mailing_address_street
```
cilestin, line 47: Were these methods removed intentionally? What replaces them if so?

ElleeB: They're stamped and auto-sized fields. We don't call these methods...they were just hanging around.

cilestin (approved):

PR #8807, opened 2022-01-05
change claim_status end point to use ssn

s-caso:
## Description of change
Update claim status end point to use current user's ssn to retrieve claimant id.  This end point will be hit by multiple apps not all of which will have the claimant id available to it.

## Original issue(s)


s-caso:
This code will be implemented in a separate PR by @bradbergeron-us

PR #8809, opened 2022-01-06
Hypertension Fast Track: update account/ICN retrieval and add tests

julialeague:
## Description of change
This is a minor update to the Hypertension Fast Track Sidekiq job. Account retrieval is refactored to avoid multiple database queries, and updates/tests around setting the ICN. Mostly adds specs testing for nil/blank cases.

Associated Notion Card: https://enchanting-fossa-083.notion.site/Unit-test-for-veteran-who-doesn-t-have-an-ICN-daf870ee7433410d9185d08a02be5aea


ericboehs (approved):

tadhg-ohiggins (approved):

spec/jobs/fast_track/disability_compensation_job_spec.rb
```diff
@@ -85,5 +85,28 @@
...
     end
+
+    context 'when an account for the user is NOT found' do
+      before do
+        allow(Account).to receive(:where).and_return Account.none
+      end
+
+      it 'raises ActiveRecord::RecordNotFound exception' do
+        expect { subject.new.perform(submission.id, user_full_name) }.to raise_error(ActiveRecord::RecordNotFound)
+      end
```
NatashaJIbrahim, line 96: Is this spec testing anything in our code? It feels like it's just testing Active Record.

app/workers/fast_track/disability_compensation_job.rb
```diff
@@ -64,10 +64,14 @@ def send_fast_track_engineer_email_for_testing(form526_submission_id, error_mess
     end
 
     def get_icn(form526_submission)
-      account = Account.where(idme_uuid: form526_submission.user_uuid).first
-      account = Account.where(logingov_uuid: form526_submission.user_uuid).first if account.blank?
-      account = Account.where(edipi: form526_submission.auth_headers['va_eauth_dodedipnid']).first if account.blank?
-      account.icn if account.present? && account.icn.present?
+      account(form526_submission).icn.presence
```
NatashaJIbrahim, line 67: Do we need to call [`#presence`](https://apidock.com/rails/Object/presence) here? I think this is just returning the or `nil` which is the same behavior as `account(form526_submission).icn` right?

app/workers/fast_track/disability_compensation_job.rb
```diff
@@ -64,10 +64,14 @@ def send_fast_track_engineer_email_for_testing(form526_submission_id, error_mess
...
+    end
+
+    def account(form526_submission)
+      user_uuid = form526_submission.user_uuid
+      @account ||= Account.where(idme_uuid: user_uuid)
+                          .or(Account.where(logingov_uuid: user_uuid))
+                          .or(Account.where(edipi: form526_submission.auth_headers['va_eauthdodedipnid'])).first!
```
NatashaJIbrahim, line 74: I was a little nervous about the use of `or` here because of the assignment but checking it in the console it seems solid. 🙌

```
irb(main):001:0> bird = false or true
=> true
irb(main):002:0> bird
=> false
irb(main):003:0> bird = false || true
=> true
irb(main):004:0> bird
=> true
irb(main):005:0> bird = nil || true
=> true
irb(main):006:0> bird
=> true
irb(main):007:0> bird = nil or true
=> true
```

NatashaJIbrahim (approved):
This looks good to me. I think one of the specs might be a little redundant to Rails internal tests and there's a method chain that could be simplified but I leave that to you whether you want to remove it or not; neither is blocking.

spec/jobs/fast_track/disability_compensation_job_spec.rb
julialeague: I'm more testing that I'm properly raising the error so that we can exit out of the job at that point. Like, if that line was changed again for some reason, I want to make sure we don't forget that.

julialeague: To be specific, the error is only being raised because we're using `.first!` instead of `.first`

app/workers/fast_track/disability_compensation_job.rb
julialeague: I used `presence` because we don't want a blank icn - and there's no validation for ICNs on the Account model.

julialeague: It is arguably overly-cautious, I grant you that, since we are handling blank/nil case when initializing the client. 🤷

NatashaJIbrahim: Hm - so I guess the concern here is that it would return something other than nil? If there's no ICN in the account model it would return nil right or does it return an empty string?

NatashaJIbrahim: Totally valid - I'm just making sure I understand.

julialeague: It would return nil with no icn, but the ICN could theoretically be set to a blank string also

 => #<Account id: 1, uuid: "c6e57917-b44f-41c6-bbff-ed24db9e547c", idme_uuid: "6400bbf301eb4e6e95ccea7693eced6f", icn: "", edipi: "1013590059", created_at: "2021-10-26 00:14:24.202283000 +0000", updated_at: "2022-01-06 20:17:41.543335000 +0000", sec_id: "0001740097,0000027647,0000027648", logingov_uuid: nil> 

2.7.4 :008 > a.icn => "" 

2.7.4 :009 > a.valid?
2022-01-06 12:20:56.071340 D [57849:67420 log_subscriber.rb:126] ActiveRecord::Base --   Account Exists? (0.8ms)  SELECT 1 AS one FROM "accounts" WHERE "accounts"."uuid" = $1 AND "accounts"."id" != $2 LIMIT $3  [["uuid", "c6e57917-b44f-41c6-bbff-ed24db9e547c"], ["id", 1], ["LIMIT", 1]]
2022-01-06 12:20:56.075892 D [57849:67420 log_subscriber.rb:126] ActiveRecord::Base --   Account Exists? (1.1ms)  SELECT 1 AS one FROM "accounts" WHERE "accounts"."idme_uuid" = $1 AND "accounts"."id" != $2 LIMIT $3  [["idme_uuid", "6400bbf301eb4e6e95ccea7693eced6f"], ["id", 1], ["LIMIT", 1]]
 => true

PR #8819, opened 2022-01-06
API-12124: Part II : Add claimant model methods hlr v2

ElleeB:
Part I PR: https://github.com/department-of-veterans-affairs/vets-api/pull/8803

## Original issue(s)
https://vajira.max.gov/browse/API-12124


modules/appeals_api/app/models/appeals_api/claimant.rb
```diff
@@ -0,0 +1,117 @@
+class Claimant
+  def initialize(auth_headers, form_data)
+    @auth_headers = auth_headers
+    @form_data = form_data
+  end
+
+  def first_name
+    first_name_header
```
cilestin, line 8: Not really sure why this method layering needs to happen. Why not use `auth_headers['X-Claimant-First-Name']` here?

modules/appeals_api/app/models/appeals_api/claimant.rb
```diff
@@ -0,0 +1,117 @@
+class Claimant
+  def initialize(auth_headers, form_data)
+    @auth_headers = auth_headers
+    @form_data = form_data
```
cilestin, line 4: I'm not sure this needs the full form data since it's only concerned about the claimant fields. I think passing in `form_data['data']['attributes']['claimant']` to the initializer would be fine, unless I'm missing something?

ElleeB: totes right

modules/appeals_api/app/models/appeals_api/claimant.rb
ElleeB: I was trying something out it's being reverted today

modules/appeals_api/app/models/appeals_api/claimant.rb
```diff
@@ -0,0 +1,117 @@
...
+
+  def ssn
+    ssn_header
+  end
+
+  def full_name
+    "#{first_name} #{middle_initial} #{last_name}".squeeze(' ').strip
+  end
+
+  def birth_mm
```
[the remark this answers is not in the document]
ElleeB: I'm going to update both these and the HLR model veteran methods as well then

modules/appeals_api/app/models/appeals_api/higher_level_review.rb
```diff
@@ -48,7 +48,13 @@ def pdf_structure(version)
       ).new(self)
     end
 
-    # 1. VETERAN'S NAME
+    def claimant
+      # TODO: Consider how we want to handle incomplete claimant headers and form data - this is not necessarily ideal
```
ElleeB, line 52: Please consider and provide feedback here :)

nathanbwright: Looking good, nice work!

Hmm, what do you think about using the null object pattern here? That way we wouldn't have to litter the form data and structure classes with conditional logic...

ElleeB: Yes! I'll work on it now

modules/appeals_api/app/models/appeals_api/non_veteran_claimant.rb
```diff
@@ -0,0 +1,155 @@
...
+      @auth_headers = auth_headers
+      @form_data = form_data
+    end
+
+    def first_name
+      auth_headers['X-VA-Claimant-First-Name']
+    end
+
+    def middle_initial
+      auth_headers['X-VA-Claimant-Middle-Initial'] || ''
```
cilestin, line 15: Why all the blank string returns? I thought pdftk did the translation from nil values into a blank string for us?

modules/appeals_api/app/models/appeals_api/non_veteran_claimant.rb
```diff
@@ -0,0 +1,155 @@
...
+    def phone_string
+      phone_formatted.to_s
+    end
+
+    def area_code
+      phone_data&.dig('areaCode') || ''
+    end
+
+    def phone_prefix
+      return unless domestic_country_code_present?
```
cilestin, line 87: So here you return nil but if the phone is not domestic but also not filled out, you return a blank string. Let's at least keep this consistent with either returning nil or a blank string but not both.

modules/appeals_api/app/models/appeals_api/non_veteran_claimant.rb
```diff
@@ -0,0 +1,155 @@
...
+
+      phone_data&.dig('phoneNumber')&.first(3) || ''
+    end
+
+    def phone_line_number
+      return unless domestic_country_code_present?
```
cilestin, line 93: Same here w/ consistency in "blank" return values.

modules/appeals_api/app/models/appeals_api/non_veteran_claimant.rb
```diff
@@ -0,0 +1,155 @@
...
+    def ssn_last_four
+      ssn.last(4) || ''
+    end
+
+    private
+
+    attr_accessor :auth_headers, :form_data
+
+    def birth_date
+      Date.parse(birth_date_string)
```
cilestin, line 131: Should be memoized so as to not re-translate the string to a date every time birth_date is called.

cilestin (changes requested):

modules/appeals_api/app/models/appeals_api/non_veteran_claimant.rb
ElleeB: I thought it didn't! I'm looking into this!

cilestin (approved):

PR #8820, opened 2022-01-06
AL 34768 add email to list

alsleaders:
This PR is for [#34768](https://github.com/department-of-veterans-affairs/va.gov-team/issues/34768) to add Amanda Leaders to the email list for the Chapter 31 daily report as well as the "silent failure" email notification list for 686/674 applications.

Furthermore, this PR removes the functions that were sending out staging email notifications.

Also, this PR updates the specs for the ruby files that were changed, so the tests continue to pass.


ericboehs (approved):

config/settings.yml
```diff
@@ -1145,6 +1145,7 @@ veteran_readiness_and_employment:
       - Jason.Wolf@va.gov
       - Kathleen.Crawford@va.gov
       - Kevin.Musiorski@va.gov
+      - Amanda.Leaders@va.gov
     staging_emails:
```
RudyOnRails, line 1149: One thought (or two), take it or leave it...

You can also probably take out this `staging_emails:` key and the single element array now that we don't use it. Or keep it and maybe only comment out the other line:

```
# Settings.veteran_readiness_and_employment.daily_report.staging_emails.dup
return
```
...in case we want to turn it on one day.

alsleaders: I think I'm going to leave the array there (just in case it has downstream effects I didn't see) but I like the idea of leaving that line but commenting it out, in case we do want to come back to it later.

app/mailers/ch31_submissions_report_mailer.rb
```diff
@@ -10,7 +10,8 @@ def build(submitted_claims)
 
     opt[:to] =
       if FeatureFlipper.staging_email?
-        Settings.veteran_readiness_and_employment.daily_report.staging_emails.dup
+        # Settings.veteran_readiness_and_employment.daily_report.staging_emails.dup
+        return
```
ericboehs, line 14: Does this prevent the email from sending all together? Or just return nil. If the latter, why not just delete the `if` block and change to `unless`?

RudyOnRails: Thanks @ericboehs 

This code stems from my suggestion when Amanda and I paired. My thought was to sort of save/comment the staging part in case we ever want to troubleshoot something again. But we can just forget it all together b/c this feature is in prod and all rough stuff should be cleaned up.

ericboehs (approved):

spec/jobs/vre/create_ch31_submissions_report_job_spec.rb
```diff
@@ -53,6 +64,14 @@
...
         subject
       end
+
+      it 'does not send if FeatureFlipper.staging_email?  is true' do
+        RSpec::Mocks.space.proxy_for(Ch31SubmissionsReportMailer).reset
+        RSpec::Mocks.space.proxy_for(FeatureFlipper).reset
+        expect(FeatureFlipper).to receive(:staging_email?).once.and_return(true)
+        expect(Ch31SubmissionsReportMailer).not_to receive(:build).with(submitted_claims)
+        subject
+      end
```
alsleaders, line 74: @RudyOnRails Is this test the same as the one you added around line 50?

RudyOnRails: @alsleaders It is similar but just within a different context.

PR #8826, opened 2022-01-07
removes original signup route & saml generation logic [32633]

bramleyjl:
## Description of change

This is a tech debt cleanup, removing the original `signup` route that assumed only ID.me signups were available. vets-api now has CSP specific `idme_signup` and `logingov_signup` routes, rendering the original unnecessary. There should be no functional difference as a result of this change.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#32633

Tested both Login.gov & ID.me signup routes to ensure they were unaffected.


bosawt (approved):
Looks good! Tested that signup still works for idme and logingov and they do.

PR #8839, opened 2022-01-10
2224 cancel appointment error noise

kpethtel:
## Description of change
Prevents logging to sentry when a facility does not permit appointment cancellation. 

## Original issue(s)
department-of-veterans-affairs/va-mobile-team#2224

After some consideration, I chose to only do this for the specific error case I saw in sentry, but it could be worth quieting other errors found in the exception file, like "disability rating not found". I'm not sure how frequent those errors are.


jperk51 (approved):
LGTM

PR #8841, opened 2022-01-10
[API-12388] - ClaimsApi - Remove "source" functionality when submitting/retrieving 526 submissions

jw81:
## Description of change

- Removes the logic that stored `source` on a 526 submission
- Removes the logic that searched on `source` when retrieving a 526 submission
- This should now allow Claims to be found by `id` and not dependent on whether it was the Veteran or the Representative that submitted the claim

## Original issue(s)
[API-12388](https://vajira.max.gov/browse/API-12388)


modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
```diff
@@ -40,11 +40,10 @@ def submit_form_526 # rubocop:disable Metrics/MethodLength
             auth_headers: auth_headers,
             form_data: form_attributes,
             flashes: flashes,
-            special_issues: special_issues_per_disability,
-            source: source_name
```
SlakrHakr, line 44: Would it be worth keeping this in for now? In case the finds need to go back to checking source at some point down the road?

hinzed1127:
It looks like we're doing [something similar for POA](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/app/models/claims_api/power_of_attorney.rb#L25). Would another ticket for removing that make sense, too?

SlakrHakr (approved):

modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
hinzed1127: We're using it for [Sentry](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/app/controllers/claims_api/v0/application_controller.rb#L83) and the [submissions report](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/app/workers/claims_api/report_unsuccessful_submissions.rb#L13).

Those instances seem useful, so maybe we keep it on the model and just omit it in the `find_by` calls?

jw81:
> It looks like we're doing [something similar for POA](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/app/models/claims_api/power_of_attorney.rb#L25). Assuming we're comfortable removing it for claims, would another ticket for removing it elsewhere make sense, too?

Yeah.  I considered trying to remove it from both 526 and 2122 in one ticket/PR, but the 2122 stuff is a little bit more involved.  So I figured I'd tackle that piece in another ticket.

modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
jw81: Fair points. Although that would introduce a new problem.  In the case of CCG, when it goes to build the `source_name` value, there's no `@current_user`.  So it's going to try to call `.first_name` and `.last_name` on `nil`.  So we'd have to add some logic to the `source_name` method to try to account for this.  Do y'all think that's worth doing? @hinzed1127 @SlakrHakr 

Maybe something like ::
```
def source_name
  if header_request?
    return 'CCG' if token.client_credentials_token?
    return "#{@current_user.first_name} #{@current_user.last_name}"
  else
    return "#{target_veteran.first_name} #{target_veteran.last_name}"
  end
end

SlakrHakr: personally, id just return what v0 is returning for that value when ccg
ccg v1 is identical to v0 in terms of functionality and what is available to the system

jw81: Talked with Lee and Dan offline and everyone agreed with this approach.
[This commit](https://github.com/department-of-veterans-affairs/vets-api/pull/8841/commits/567e482b2c1712dbd55202b8ef2a4df2e66dd44b) makes the suggested change.

hinzed1127 (approved):

PR #8856, opened 2022-01-11
Move job status module

NatashaJIbrahim:
## Description of change
 - moves the code that was in `EVSS::DisabilityCompensationForm::JobStatus` out of the `EVSS` module so that it can be used by other classes. 
 - This is in preparation for using this module to track the [`FastTrack::DisabilityCompensationJob`](https://github.com/department-of-veterans-affairs/vets-api/blob/move-job-status-module/app/workers/fast_track/disability_compensation_job.rb). 

## Original issue(s)
[department-of-veterans-affairs/va.gov-team#0000](https://www.notion.so/64997f0e6c4c40df816b7c4b305d18d4?v=fe1d45ccd3014f6484a2f70842c3df19&p=fb8be589c41c462182089b69cbbaaab6)

## Notes for the reviewer
- I chose to move the tracking code to `lib/sidekiq/job_tracker` but I'm open to suggestions if there are better places for it to live. 
- I'm not sure how to make this pr smaller 😿 since it's simply moving a big chunk of code. I'm very open to suggestions <3 Thank you in advance!


tadhg-ohiggins:
One concern I have is that the job tracking module and classes are named very generically, but the methods are doing things that look specific to Form 526; it would probably make sense to rename the module and/or classes to refer to Form 526.

ericboehs (approved):

app/workers/evss/disability_compensation_form/job_status.rb
```diff
@@ -1,198 +1,15 @@
 # frozen_string_literal: true
 
+require 'sidekiq/form526_job_status_tracker/job_tracker'
```
julialeague, line 3: Could we remove this and the metrics class under this directory entirely if we just included those modules in EVSS::DisabilityCompensationForm::Job?

app/workers/evss/disability_compensation_form/job_status.rb
```diff
@@ -1,198 +1,15 @@
...
-      #
-      def retryable_error_handler(error)
-        upsert_job_status(Form526JobStatus::STATUS[:retryable_error], error)
-        log_error('retryable_error', error)
-        metrics.increment_retryable(error, @is_bdd)
-      end
-
-      # Metrics and logging for any non-retryable errors that occurred.
-      # Non-retryable errors will always fail, e.g. an ArgumentError in the job class
+      include Sidekiq::Form526JobStatusTracker::JobTracker
```
julialeague, line 9: Could we remove this evss-specific module and the evss metrics module under this directory entirely if we just included those modules in EVSS::DisabilityCompensationForm::Job?

julialeague (dismissed):
Looks good outside of that question I had above!

app/workers/evss/disability_compensation_form/job_status.rb
NatashaJIbrahim: I _think_ I tried that and things broke elsewhere that I wasn't prepared to address. Let me try once more just to be sure though because that would definitely be cleaner.

ericboehs (approved):

PR #8860, opened 2022-01-12
Sort medication by two fields

tadhg-ohiggins:
## Description of change
The LH API test data has multiple entries with identical timestamps, so sorting by timestamp (in this case `authoredOn`) can produce inconsistent results across platforms, making testing trickier.

This PR adds sorting by a second field, `description`, to make it consistent, and reorders the expected results in the spec to match.

-----

Simultaneous,
Yet we must sort anyway—
So use `description`.


julialeague:
Woo, thanks for fixing the tests with this change! Looks good to me

PR #8863, opened 2022-01-12
Hypertension: Add support for pagination in Lighthouse Veterans Health client

julialeague:
## Description of change
We need to add pagination support to our Lighthouse Vets Health API client so that we'll return all relevant results in our resulting PDF. This adds that support and updates/re-runs VCR specs for the same.

## Testing

Specs for hypertension MedicationRequest processing were updated to use a test staging ICN that is known to have paginated results. A VCR test was also added to test the client directly, which I decided to namespace under `lib/lighthouse/veterans_health/integration/client_spec.rb` to delineate the existing client specs as unit tests that don't rely on external data.


NatashaJIbrahim (approved):
This looks solid to me.

ericboehs (dismissed):

ericboehs (approved):

PR #8866, opened 2022-01-13
API-12125 Claimant form data and structure hlr v2

ElleeB:
Updates hlr v2 form_data and structure to handle claimant info.
**Stamp and autosize will come in another PR.**

[Form fields and model updates PR](https://github.com/department-of-veterans-affairs/vets-api/pull/8819)

[Ticket](https://vajira.max.gov/browse/API-12125)


modules/appeals_api/app/services/appeals_api/pdf_construction/higher_level_review/v2/form_data.rb
```diff
@@ -157,6 +157,96 @@ def stamp_text
           "#{last_name.truncate(35)} - #{ssn.last(4)}"
         end
 
+        def claimant_first_name
```
cilestin, line 160: Why not use the claimant object directly instead of using this layer of methods? This isn't much different than just putting all the claimant data retrieval on the model, which also felt redundant and is why we wanted the Claimant object in the first place :)

modules/appeals_api/app/services/appeals_api/pdf_construction/higher_level_review/v2/structure.rb
```diff
@@ -37,10 +38,23 @@ def form_fill
             form_fields.veteran_phone_line_number => form_data.veteran_phone_line_number,
             form_fields.veteran_phone_international_number => form_data.veteran_phone_international_number,
 
-            # Section II: Claimant's ID
-            # NOT YET SUPPORTED
-
-            # Section III: Benefit Type
+            # Claimant
+            # Claimant name is filled out through autosize text box, not pdf fields
+            form_fields.claimant_middle_initial => form_data.claimant_middle_initial,
```
cilestin, line 43: `form_data.claimant.middle_initial` would be preferable here, IMO. No need to hide the claimant data behind a bunch of facade methods.

cilestin (changes requested):

modules/appeals_api/app/services/appeals_api/pdf_construction/higher_level_review/v2/structure.rb
cilestin: We can use the null object pattern for the claimant object - basically you can still instantiate an instance even if it doesn't have any/all the data, and anything it can't respond to just returns a nil.

`myClaimant = Claimant.new(nil_headers, nil_claimant_data)
myClaimant.last_name # => nil`

This way, the claimant object will always be present and we won't have to do to any nil checks from the structure at all :)

ElleeB: Ohhhh. Nathan mentioned this but I thought it was far more complicated than this! I will work on it.

modules/appeals_api/app/models/appeals_api/non_veteran_claimant.rb
```diff
@@ -48,21 +48,33 @@ def number_and_street
...
+      form_data.dig('address', 'city')
     end
 
     def state_code
-      form_data['address']['stateCode']
+      form_data.dig('address', 'stateCode')
     end
 
     def country_code
+      return if form_data.blank?
```
cilestin, line 59: You should be checking that the specific data you're looking for (e.g. `form_data['address']['countryCodeISO2']`) is missing, not against all the form data being blank.

modules/appeals_api/app/models/appeals_api/non_veteran_claimant.rb
```diff
@@ -48,21 +48,33 @@ def number_and_street
...
     end
 
     def zip_code_5
+      return if form_data.blank?
+
       form_data['address']['zipCode5'] || '00000'
     end
 
+    def zip_code
+      if zip_code_5 == '00000'
```
cilestin, line 71: I think this should work instead of this kind of convoluted `if` structure:
```ruby
form_data.dig('address', 'internationalPostalCode') || zip_code_5
```

cilestin (changes requested):

modules/appeals_api/app/models/appeals_api/non_veteran_claimant.rb
ElleeB: I don't want it to print 'US' though, if there is no form data.

modules/appeals_api/app/models/appeals_api/non_veteran_claimant.rb
ElleeB: I don't disagree that it's ugly, but we want to replace the internationalPostalCode with '00000' if it's been supplied for zip_code5, but print the international postal code if it is present, I believe.

ElleeB: Now I'm replaying this thing in my head...

ElleeB: So if there is no international postal code, and zip_code_5 is '48323', then '48323' will print for zip code and international postal code, right?

modules/appeals_api/app/models/appeals_api/non_veteran_claimant.rb
ElleeB: i'll see if there's a nicer way to do this.

cilestin: As it stands now, you'd still be printing `US` even if an unrelated form_data key existed, like `form_data['email']`. You only want to return nil if the specific data you're trying to otherwise print is not present.

modules/appeals_api/app/models/appeals_api/non_veteran_claimant.rb
nathanbwright: If the zip_code is `00000` then we want to print the `internationalPostalCode` on the pdf - but we have to send `00000` to CMP, unless I'm misremembering something.

nathanbwright:
It's looking good! It looks the claimant first and last name aren't totally wired up for printing on the pdf... Just the middle name is printed on the `expected_200996_v2_extra.pdf` file.

modules/appeals_api/app/models/appeals_api/non_veteran_claimant.rb
cilestin: We don't differentiate on the form between zipcode & international postal code - the same field is in use for both.

So, if they've supplied an international postal code we will use it, otherwise we'll use the provided zip_code_5, which itself includes a fallback to print '00000' if it is missing. I think the logic still works out for my shortened version.

cilestin: @nathanbwright Correct, but we [use zip_code_5](https://github.com/department-of-veterans-affairs/vets-api/blob/74294c7c0590905e4d9254e2c03acc66274c62e2/modules/appeals_api/lib/appeals_api/hlr_pdf_submit_wrapper.rb#L13) when talking to centralmail to default to 00000

modules/appeals_api/app/models/appeals_api/non_veteran_claimant.rb
ElleeB: RIGHT!

modules/appeals_api/app/models/appeals_api/non_veteran_claimant.rb
ElleeB: > It's looking good! It looks the claimant first and last name aren't totally wired up for printing on the pdf... Just the middle name is printed on the `expected_200996_v2_extra.pdf` file.

The stamped and auto-sized fields are coming in the next PR!

ElleeB:
> It's looking good! It looks the claimant first and last name aren't totally wired up for printing on the pdf... Just the middle name is printed on the `expected_200996_v2_extra.pdf` file.

Coming in the next PR :)

modules/appeals_api/app/models/appeals_api/non_veteran_claimant.rb
ElleeB: I'm still confused on this one. I thought the zeros were only to indicate outside of US.

```
"zipCode5": {
      "type": "string",
      "description": "5-digit zipcode. Use \"00000\" if Veteran is outside the United States",
      "pattern": "^[0-9]{5}$"
},
 "internationalPostalCode": { "type": "string", "maxLength": 16 }
 ```

Comment (ElleeB) [reply modules/appeals_api/app/models/appeals_api/non_veteran_claimant.rb:71]: ok. i've confirmed that we do in fact assign `00000` if there is no zip code. so do we prefer this? printing potentially nothing under claimant and then randomly a '00000' for zip code?

Comment (ElleeB) [reply modules/appeals_api/app/models/appeals_api/non_veteran_claimant.rb:71]: Final report! We assign the zeros for a  _Veterant if_ they are experiencing homelessness. This will not work the same way for a claimant. If they provide any address info, the zipCode5 field is required, and it will come in as a domestic zip, or '00000', indicating an internationalPostalCode. So, it should be as simple as, zipCode5 || InternationalPostalCode.

Comment (cilestin) [reply modules/appeals_api/app/models/appeals_api/non_veteran_claimant.rb:71]: I think asking for 0's like we do for veterans is fine :)

Comment (ElleeB) [reply modules/appeals_api/app/models/appeals_api/non_veteran_claimant.rb:71]: but why ask claimants for zeros? why not supply one or the other?

File: modules/appeals_api/app/models/appeals_api/non_veteran_claimant.rb [hunk 5]
```diff
@@ -4,7 +4,7 @@ module AppealsApi
   class NonVeteranClaimant
     def initialize(auth_headers:, form_data:)
       @auth_headers = auth_headers
```
cilestin, line 6: I think you'll have a decent number of failures if you pass in `nil` for auth_headers, since there isn't an `|| {}` fallback to make methods like `first_name` work if they're `nil`.

I think this should still be `@auth_headers = auth_headers || {}`

cilestin (changes requested):
One super small change request but otherwise lookin' good!

cilestin (approved):

PR #8892, opened 2022-01-19
Api 12606 - NOD claimant data

nathanbwright:
## Description of change
Updates to our NOD form classes and pdf generation to handle claimants. I had to re-generate the 10182_min pdf and I'm not sure why because it looked exactly like the old one, maybe a false positive from our pdf matcher?

## Original issue(s)
https://vajira.max.gov/browse/API-12606


modules/appeals_api/app/services/appeals_api/pdf_construction/notice_of_disagreement/v1/form_data.rb
```diff
@@ -87,14 +100,14 @@ def representatives_name
 
         attr_accessor :notice_of_disagreement
 
-        def name
-          initial = middle_initial
+        def formatted_full_name(first, middle, last)
+          initial = middle
           initial = "#{initial}." if initial.size.positive?
```
ElleeB, line 105: While we're here...does this have to be written this way?

ElleeB (approved):
Looks great :)

cilestin (approved):

PR #8901, opened 2022-01-19
Bug: EVSS auth header key access typo introduced in a previous PR

julialeague:
## Description of change
A typo was introduced in #8809, where an underscore was removed from the key name of the EVSS auth header we're attempting to access in our code. This PR fixes that typo!


tadhg-ohiggins (dismissed):

NatashaJIbrahim:
@julialeague I don't think we want to merge this quite yet. This will lead to us loosing the weird visibility we have atm via the "Insufficient data" emails. Let's hold this until we chat with Emily on Monday about what she wants to be tracking.

tadhg-ohiggins (approved):

app/workers/fast_track/disability_compensation_job.rb
```diff
@@ -68,10 +69,16 @@ def get_icn(form526_submission)
...
+      user_uuid = form526_submission.user_uuid.presence
+      edipi = form526_submission.auth_headers['va_eauth_dodedipnid'].presence
+      account = nil
+
+      account = Account.where(idme_uuid: user_uuid).first if user_uuid
+      account = Account.where(logingov_uuid: user_uuid).first if account.blank? && user_uuid
+
+      account = Account.where(edipi: edipi).first if account.blank? && edipi.present?
+
+      account
```
rileyanderson, line 81: Could you make this an if/elsif/else statement? It would be cleaner and more efficient. 
```suggestion
      if user_uuid
        account = Account.where(idme_uuid: user_uuid).first
      elsif account.blank? && user_uuid
        account = Account.where(logingov_uuid: user_uuid).first
      elsif account.blank? && edipi.present?
        account = Account.where(edipi: edipi).first
      else
        account = nil
      end

      account
```

julialeague: I don't think that will work with what we're trying to do, unfortunately... For instance, a user_uuid could exist, but we can't be certain we'll find an Account that matches the user_uuid with an idme or logingov_uuid (we've encountered this case in production). Therefore, we need to ensure that we check for each case instead of only route through one branch. Does that make sense?

michelpmcdonald: not sure if this is any cleaner....how about something like the original code had it(||=) ?
```
account =   user_uuid ? Account.where(idme_uuid: user_uuid).first : nil
account ||= user_uuid ? Account.where(logingov_uuid: user_uuid).first : nil
account ||= edipi.present? ? Account.where(edipi: edipi).first : nil
```
Or:

```
account =   Account.where(idme_uuid: user_uuid).first if user_uuid
account ||= Account.where(logingov_uuid: user_uuid).first if user_uuid
account ||= Account.where(edipi: edipi).first if edipi.present?
```

julialeague: I love your second sugestion, @michelpmcdonald, I'm going to update the code for that! Thank you!!

julialeague: Updated that here https://github.com/department-of-veterans-affairs/vets-api/pull/8901#:~:text=method%20%2D%20much%20cleaner-,9712568,-Merge%20state

How's that look to you, @rileyanderson ?

rileyanderson (approved):

PR #8903, opened 2022-01-19
Api 11843 token validation 500

crolarlibertyva:
- Addresses a defect described in https://vajira.max.gov/browse/API-11843 for handling a badly formed request.
- No changes in functionality


rileyanderson (dismissed):

dbrown-va (approved):
This error specifically occurs on bad user input (malformed json) so a 400 errors seems appropriate.  Does not raise any inner details of the exception outside of sentry.

https://www.rfc-editor.org/rfc/rfc7231#section-6.5

PR #8915, opened 2022-01-20
API-11577: rescue ArgumentError in claims validations

hinzed1127:
## Description of change

A user was seeing an error that seems to be related to bad identifier request headers. We should rescue the error they're triggering and relay it back to them.
## Original issue(s)
[department-of-veterans-affairs/va.gov-team#0000](https://vajira.max.gov/browse/API-11577)


hinzed1127:
Random aside, but I noticed that [this test](https://github.com/department-of-veterans-affairs/vets-api/blob/8f6ed597a527cb905b83af0784158a3e55463a92/modules/claims_api/spec/requests/v1/power_of_attorney_request_spec.rb#L108-L111) seems to be checking for the wrong thing. Is this bad wording, or is the actual test wrong?

modules/claims_api/spec/requests/v1/power_of_attorney_request_spec.rb
```diff
@@ -142,6 +142,30 @@
...
+        end
+
+        it 'returns an unprocessible entity status' do
+          allow(MPI::Messages::FindProfileMessageFields).to receive(:new).and_return(
+            OpenStruct.new({
+                             given_names: ['abraham'],
+                             last_name: 'lincoln',
+                             birth_date: nil,
+                             ssn: '796111863',
+                             missing_values: [:birth_date]
```
hinzed1127, line 158: there was a lot of things getting mocked through, so this was my quick way to get the error I wanted. let me know if there's a better, cleaner way to get to that same error in a test

jw81 (dismissed):
LGTM.  Would be awesome if you wanted to update that spec name while you were in here, but I'll leave it up to ya.  😉

rileyanderson (approved):

jw81 (approved):

PR #8922, opened 2022-01-21
35259 Additional logging

dillo:
## Description of change

- Add easy to understand logging around Check-In controllers
## Original issue(s)
department-of-veterans-affairs/va.gov-team#35259


modules/check_in/app/controllers/check_in/application_controller.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+# require CheckIn::Engine.root.join('lib/logger')
```
randomsync, line 3: needed?

kanchanasuriya (dismissed):
LGTM

modules/check_in/lib/check_in/utils/logger.rb
```diff
@@ -0,0 +1,66 @@
...
+      def ctrl_action
+        ctrl.action_name
+      end
+
+      def api_status
+        body = ctrl.response.body
+
+        return 'none' if body.blank?
+
+        body.match(API_STATUS_REGEXP) { |g| g[1] if g }
```
randomsync, line 48: just noting our discussion for posterity: we discussed the concern about parsing response body on every action call, and that it would be mitigated once we update the controllers to return an appropriate HTTP response code for each request (which will require working with FE team and Product team on defining what response code should be returned based on downstream responses).

randomsync (dismissed):
LGTM, minor comments.

modules/check_in/lib/check_in/utils/logger.rb
dillo: Thanks!

modules/check_in/app/controllers/check_in/application_controller.rb
dillo: No. I'll have it removed! 👍

randomsync (approved):

rileyanderson (approved):

PR #8930, opened 2022-01-21
API-12217 Connected Apps optionally lookup by icn

dbrown-va:
## Description of change
This PR addresses an issue discovered with the updates surrounding the Login.gov CSP addition for VA.gov & Lighthouse.

Since that diverges from ID.me uuid's as the consistent way of looking up Lighthouse 3rd party application grants, this method will eventually update those lookups to be by ICN, a consistent trait across IDP's.

## Original issue(s)
https://vajira.max.gov/browse/API-12217

This PR includes a new settings.yml entry to enable the Okta lookup by ICN.  By default it will continue to use the uuid until the different environments are promoted.

`Settings.oidc.base_api_profile_key_icn` is the new setting/flag.

Added an rspec along with some local testing to verify the new behavior.  However will also default to the existing uuid behavior until dev.va.gov testing can be completed as well before promoting to any higher environments.

A number of existing Connected Apps rspec's were set up with `uuid == okta uid`, this was incorrect.  These instances have been updated to use the csp uuid and then corresponding vcr's updated to match since lookup by login (csp uuid) was actually how it worked.


dbrown-va:
Corresponding Devops PR enabling in Dev only first.
https://github.com/department-of-veterans-affairs/devops/pull/10613

crolarlibertyva (approved):

PR #8947, opened 2022-01-25
2361 Immunization group name and manufacturer from LH data

aherzberg:
## Description of change

- Group name is now set from LH data
- Removed group name column from vaccine model in related PR https://github.com/department-of-veterans-affairs/vets-api/pull/8969 
- Update vaccine job to only use manufacturer xml.
- Added metric to track if any covid vaccines come in without manufacturers. 
- Updated immunization and vaccine job tests. This includes replacing all VCR cassettes with the new LH data structure. 

## Original issue(s)
https://github.com/department-of-veterans-affairs/va-mobile-app/issues/2361

- Note that tests will not pass on this PR since the group name column is removed in the PR listed above.


modules/mobile/app/models/mobile/v0/adapters/immunizations.rb
```diff
@@ -4,22 +4,24 @@ module Mobile
...
         def parse(immunizations)
           vaccines = vaccines(immunizations)
 
           immunizations[:entry].map do |i|
             immunization = i[:resource]
             vaccine_code = immunization[:vaccine_code]
             cvx_code = cvx_code(vaccine_code)
-            vaccine = vaccines&.find_by(cvx_code: cvx_code)
+            group_name = group_name(vaccine_code)
+            vaccine = vaccines&.find_by(cvx_code: cvx_code) if group_name == 'COVID-19'
```
aherzberg, line 16: Think the vaccine variable should be initialized to nil first?

modules/mobile/app/models/mobile/v0/adapters/immunizations.rb
```diff
@@ -4,22 +4,24 @@ module Mobile
   module V0
     module Adapters
       class Immunizations
+        # rubocop: disable Metrics/MethodLength
```
aherzberg, line 7: Don't see a way to break this up intuitively so I disabled the method length rule

modules/mobile/app/workers/mobile/v0/vaccines_updater_job.rb
```diff
@@ -41,27 +40,26 @@ def perform
...
-        group_name = find_value(node, 'Vaccine Group Name')
+        manufacturer = find_value(node, 'Manufacturer')
+
+        return nil if cvx_code.blank? || manufacturer.blank?
 
-        aggregate[cvx_code] = { group_names: [], manufacturer: nil } unless aggregate[cvx_code]
-        aggregate[cvx_code][:group_names] << group_name
-        aggregate[cvx_code][:manufacturer] = find_manufacturer(cvx_code) if group_name == 'COVID-19'
+        aggregate[cvx_code] = [] unless aggregate[cvx_code]
+        aggregate[cvx_code] << manufacturer
```
aherzberg, line 48: capturing all manufacturers even though we currently only need the covid-19 ones

db/migrate/20220125155705_remove_mobile_vaccine_group_name.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveMobileVaccineGroupName < ActiveRecord::Migration[6.1]
+  def change
+    safety_assured {  remove_column :mobile_vaccines, :group_name }
```
aherzberg, line 3: Added migration but omitted the updated schema file. That'll be generated when they run migration in prod.

aherzberg:
@kreek I have yet to update the vaccine job tests or documentation but wanted to ensure I was going in the right direction with the vaccine job changes. My main thing I wanted to verify was that we want to keep track of all manufacturer data even though we currently only use the ones for covid-19. See above for justifications.

modules/mobile/app/models/mobile/v0/adapters/immunizations.rb
```diff
@@ -39,7 +48,7 @@ def date(immunization)
         end
 
         def cvx_code(vaccine_code)
-          code = vaccine_code[:coding].first[:code]
+          code = vaccine_code.dig(:coding, 0, :code)
```
kreek, line 51: I did not know you could `dig` into arrays like this! 👍

modules/mobile/app/workers/mobile/v0/vaccines_updater_job.rb
kreek: Unless @kpethtel or @cadibemma have other ideas, I think that's fine. Might want to also double check with front-end to see if that would cause an issue.

db/migrate/20220125155705_remove_mobile_vaccine_group_name.rb
kreek: 🚨 Note that migrations do not run automatically, you'll need to open a #vfs-platform-support ticket via slack when you merge this in 🚨

kreek (commented):
I think this is on the right track, looking good so far 👍

modules/mobile/app/workers/mobile/v0/vaccines_updater_job.rb
kpethtel: I agree that including manufacturer for all records is fundamentally better. It shouldn't be a problem here because 1) the adapter weeds out non-covid vaccines 2) all covid vaccines only have one manufacturer. 

If any covid vaccines ever have multiple manufacturers, it could cause some formatting issues. I don't think that will happen anytime soon, but it could be worth logging an error to sentry in the adapter if that ever occurs.

db/migrate/20220125155705_remove_mobile_vaccine_group_name.rb
kpethtel: My memory here may be backwards here, but I think @kreek told me to put migrations in the mobile module. If so, I obviously messed it up (probably while breaking my PR into several smaller PRs) and put it in rails root. It might be best to leave this one here for consistency, but it would be good to get a reminder of where mobile migrations should go.

modules/mobile/app/workers/mobile/v0/vaccines_updater_job.rb
```diff
@@ -18,7 +17,7 @@ def perform
         aggregate = {}
         results = { created: 0, updated: 0, persisted: 0 }
 
-        group_name_xml.root.children.each do |node|
+        manufacturer_xml.root.children.each do |node|
```
kpethtel, line 20: Does this result in any changes in the data? It would be good to run this script with both sources and confirm that they produce identical outputs.

kpethtel (commented):
I left some thoughts, but it looks good so far. 👍

modules/mobile/app/workers/mobile/v0/vaccines_updater_job.rb
aherzberg: Good point. There were some vaccine records before that only had a group name and cvx code but no manufacturer so those will no longer be tracked. Comparing outputs will be a good idea to ensure the group names are the same coming from LH compared to this group name xml as well as the manufacturers but the covid manufacturers that are used theoretically should be unchanged. I'll check it out.

modules/mobile/app/workers/mobile/v0/vaccines_updater_job.rb
```diff
@@ -41,27 +40,26 @@ def perform
 
       def aggregate_source_data(aggregate, node)
         cvx_code = find_value(node, 'CVXCode')
-        group_name = find_value(node, 'Vaccine Group Name')
+        manufacturer = find_value(node, 'Manufacturer')
+
+        return nil if cvx_code.blank? || manufacturer.blank?
```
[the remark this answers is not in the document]
aherzberg: there's some expected blank manufacturers but looks like on the manufacturers xml the cvx code is the primary unique identifier and doesn't have any blanks so I'll add an error for that.

modules/mobile/app/workers/mobile/v0/vaccines_updater_job.rb
kpethtel: It sounds like we may end up with fewer vaccines using this approach. However, we aren't destroying old vaccine records, so they'll still be around in production. It might be worth some sentry logging in the adapter if the vaccine isn't found, since that may be a sign that we're not getting all the data we need from the manufacturers' table. We probably should've done it before, just to have an indicator of how complete our data is.

modules/mobile/app/workers/mobile/v0/vaccines_updater_job.rb
aherzberg: It'd be worth logging an error if there's multiple covid manufacturers but that may get messy since some single covid manufacturers already have a comma in their name so in the adapter there's no fool proof way of knowing if the manufacturer string was one or more manufactures originally.

db/migrate/20220125155705_remove_mobile_vaccine_group_name.rb
kreek: yes, good catch @kpethtel, we have our own migrations for mobile and this file should move there.

modules/mobile/app/workers/mobile/v0/vaccines_updater_job.rb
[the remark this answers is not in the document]
kpethtel: Thinking about this more... if we add logging or error handling, it'll happen every day, since the source data is mostly static. I'd run this locally first and confirm that all nodes have cvx codes. If they do, then we could add some error logging for the future. If not, maybe we just consider it normal behavior and keep the code as it is.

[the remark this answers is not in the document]
kpethtel: I keep thinking about this. Honestly, I'm not sure we care. It's the CDC's data. It's really not up to us to verify their data integrity. If a record doesn't have all the info, skipping seems reasonable. Probably best to just leave it as is.

db/migrate/20220125155705_remove_mobile_vaccine_group_name.rb
kreek: Not a biggie - PRs are a team sport 😄 - the reason for putting migrations in our module is that we have code ownership over it and don't need VFS approval to merge it in.

modules/mobile/app/workers/mobile/v0/vaccines_updater_job.rb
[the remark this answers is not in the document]
aherzberg: That's fair. Could be worse off if the sentry error flies under the radar and we kill the job over again and never get any valid new vaccine record unknowingly just due to a missing cvx code of one of them.

modules/mobile/app/workers/mobile/v0/vaccines_updater_job.rb
aherzberg: If we did get a miss on the covid vaccine lookup, would it be worth interrupting flow to raise a sentry error? the result of a miss would be no manufacturer shown which is preferable to no information about the immunization at all. May be better to add a datadog metric to the vaccine dashboard and log the issue instead. Thoughts?

kpethtel: You can log to sentry without raising an error. There's a SentryLogging module that might give you some ideas. I haven't looked closely at it, but it looks like we do it via the rails logger. In the past, I've used Sentry directly to log errors without disrupting the code processes, but I don't see any evidence of us doing that so I would just follow the pattern that's been established. 

That's assuming we want to do this at all. It might be useful to know when we don't find a match, but obviously the code works fine without it.

db/migrate/20220125155705_remove_mobile_vaccine_group_name.rb
aherzberg: I've moved the migration over to the mobile module but Github is giving me the error 'Modified files in db/ should be the only files checked into this PR.'. Should I breakout the migration into it's own PR?

kreek: > I've moved the migration over to the mobile module but Github is giving me the error 'Modified files in db/ should be the only files checked into this PR.'. Should I breakout the migration into it's own PR?

Yes, we can merge them back to back.

aherzberg: Changed the migration to be in mobile module in it's own PR. (https://github.com/department-of-veterans-affairs/vets-api/pull/8969) but running the migration locally it seems like the mobile_vaccines table is out of scope. running `rails db:migrate` gives `PG::UndefinedColumn: ERROR:  column "group_name" of relation "mobile_vaccines" does not exist`. 

It looks like we have a task `mobile:install:migrations` but that just copies the the migrations into the base application and then `db:migrate` gives error due to multiple migrations with the same name.

aherzberg: Nevermind... It would probably help if I did a rollback before trying again.

modules/mobile/app/workers/mobile/v0/vaccines_updater_job.rb
aherzberg: That's good to know about sentry logging but I agree, if no ones using sentry for that, it may add some confusion on whats showing up on sentry. I ended up adding a metric to give a count of covid manufactures missing just so we can get an indicator to check out the latest manufacturer xml.

aherzberg:
Too many changed lines. Closing PR and splitting them up into separate PRs.

PR #8956, opened 2022-01-25
Lockbox Patch: Bring back the padding (but also don't)

ericboehs:
These were the records we ensured worked in prod:
```ruby
puts InProgressForm.find_by(id: 4436292).form_data
puts AppealsApi::HigherLevelReview.find_by(id: 'af224bb3-ec21-409a-80a0-45d179b90681').form_data
puts Form526Submission.lockbox_attributes.keys
puts Form526Submission.find(1065395)
puts Form526Submission.find(1065395).auth_headers_json
puts Form526Submission.find(1065395).birls_ids_tried
puts Form526Submission.find(1065395).form_json
```


app/models/application_record.rb
```diff
@@ -4,7 +4,12 @@ class ApplicationRecord < ActiveRecord::Base
   self.abstract_class = true
 
   def self.lockbox_options
-    { previous_versions: [{ padding: true, master_key: Settings.lockbox.master_key }, { padding: true }] }
+    {
+      previous_versions: [
+        { padding: false },
+        { padding: false, master_key: Settings.lockbox.master_key }
```
ericboehs, line 10: I inverted these rows in case it fallsback in the order of this array.

LindseySaari: Yes, smart idea!

LindseySaari (approved):

LindseySaari:
This is great. Almost all records (except the ones from yesterday and part of today). I think padding: false in the previous_versions makes way more sense 🙇‍♀️

PR #8957, opened 2022-01-26
Handle & alert individual record errors when updating in bulk

cilestin:
## Description of change
Ensures that individual record errors do not affect the entire batch of updates by catching exceptions as we update and notifying Sentry & Slack before continuing with the rest of the updates.

## Original issue(s)
https://vajira.max.gov/browse/API-13007


ElleeB (approved):
Looks good!

PR #8964, opened 2022-01-26
API-12689 - Add NOD email notifications

nathanbwright:
## Description of change
Adds notification email for successful NOD submission.

## Original issue(s)
https://vajira.max.gov/browse/API-12689


modules/appeals_api/spec/models/notice_of_disagreement_spec.rb
```diff
@@ -178,7 +180,50 @@
...
+                                              }
+                                            }
+                                          }
+                                        })
+      )
+
+      params = { event_type: :nod_received, opts: {
+        email_identifier: { id_value: '1013062086V794840', id_type: 'ICN' },
+        first_name: notice_of_disagreement.veteran_first_name,
+        date_submitted: notice_of_disagreement.date_signed,
```
cilestin, line 213: Whelp - this exposes that we've had a slightly incorrect test setup for a while across our HLR & SC notifiers. The `date_submitted` should be in ISO8601 format when sending to the event. Since the test only checks that we create a new Events::Handler, it doesn't expose that this date format isn't correct and may actually end up being a failure, since we use `Date.iso8601(date_submitted)` to parse the date out.

I think you can safely remove the `date_signed` method from the model, and then send in the iso8601 date here. I'll make a tech debt ticket to fix up the tests elsewhere.

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
```diff
@@ -151,13 +155,50 @@ def update_status!(status:, code: nil, detail: nil)
...
+
       update!(status: status, code: code, detail: detail)
 
       handler.handle!
+      email_handler.handle! if status == 'submitted' && email_identifier.present?
     end
 
     private
 
+    def mpi_veteran
```
cilestin, line 173: Not a blocker, but if these added methods are all the same for getting the email_identifier, maybe we should pull them out into a concern. A ticket for later, I'd assume :)

cilestin (changes requested):
We'll also need to add the `nod_received` subscription to the `AppealsApi::DataMigrations::EventSubscriptionSubscriber.run` method for the new event to trigger.

modules/appeals_api/spec/models/notice_of_disagreement_spec.rb
```diff
@@ -178,7 +180,50 @@
...
+    end
+
+    it 'successfully gets the ICN when email isn\'t present' do
+      notice_of_disagreement = described_class.create!(
+        auth_headers: auth_headers,
+        form_data: form_data.deep_merge({
+                                          'data' => {
+                                            'attributes' => {
+                                              'veteran' => {
+                                                'emailAddressText' => nil
```
cilestin, line 203: We'll need a test for NOD v2 as well, at some point, since we renamed this key to `email` there.

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
```diff
@@ -151,13 +155,50 @@ def update_status!(status:, code: nil, detail: nil)
...
+      AppealsApi::Veteran.new(
+        ssn: ssn,
+        first_name: veteran_first_name,
+        last_name: veteran_last_name,
+        birth_date: birth_date.iso8601
+      )
+    end
+
+    def email_identifier
+      return { id_type: 'email', id_value: email } if email.present?
```
cilestin, line 183: The `email` method will need to work for both NOD v1 and v2 since we changed the key from `emailAddressText` to `email`.

I'll add a ticket to circle back around to make sure v1 and v2 work.

cilestin (approved):

PR #8972, opened 2022-01-27
Add more notification email addresses for Hypertension RRD

yoomlam:
## Description of change

Add more notification email addresses for Hypertension RRD with a few new folks who've recently joined the Benefits Delivery team.

Also removed unnecessary `require prawn` from `DisabilityCompensationJob`.

## Original issue(s)
[Relevant Slack post](https://dsva.slack.com/archives/C01BVF2A3V0/p1643236446328519)


app/workers/fast_track/disability_compensation_job.rb
```diff
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require 'prawn'
-require 'prawn/table'
```
yoomlam, line 4: I didn't notice uses of Prawn in this class. I'm assuming Rspecs will fail if this removal is bad.

julialeague: Oh yes, I think we missed that during the refactor to pull out our pdf generator into another class. Good catch!

julialeague: I do wonder if we need to add this explicit require to `hypertension_pdf_generator.rb` - I imagine it would have run into test failures in its own spec if it was needed, though. 🤔

julialeague (dismissed):

app/workers/fast_track/disability_compensation_job.rb
yoomlam: https://github.com/department-of-veterans-affairs/vets-api/pull/8558#discussion_r793896226

rileyanderson (approved):

PR #8974, opened 2022-01-27
update sidekiq to 6.4.0

rileyanderson:
## Description of change
In api.rb in Sidekiq before 6.4.0, there is no limit on the number of days when requesting stats for the graph. This overloads the system, affecting the Web UI, and makes it unavailable to users.

https://github.com/advisories/GHSA-jrfj-98qg-qjgv


Gemfile
```diff
@@ -196,7 +196,7 @@ group :development, :test do
   gem 'rubocop-rails'
   gem 'rubocop-rspec'
   gem 'rubocop-thread_safety'
-  gem 'sidekiq'
+  gem "sidekiq", '>= 6.4.0'
```
LindseySaari, line 199: Rubocop 👮

rileyanderson: 🚔

rileyanderson: thats what i get for copy 🍝

LindseySaari (approved):

PR #8979, opened 2022-01-28
AL 32455 document upload

alsleaders:
## Description of change
This change adds the ability to upload supporting documents to an in-progress 28-1880 file so they can be retrieved when the entire form is submitted to LGY

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/32455


alsleaders:
This is in draft because all the tests pass but I have not been able to test this locally yet.

rileyanderson (dismissed):

alsleaders:
The code and specs for this change are finished.  I have put the PR back into draft until I can log in locally (which is failing because of Id.me) and check that the document upload is successfully.

alsleaders:
Using user 71 I was able to log into my RI and see a pending application and upload a document, so the upload is successful.  I'm making a note on the PUT to LGY to make sure that when the entire form is submitted that we make sure the supporting documents are sent as well.

Gemfile.lock
```diff
@@ -791,7 +791,7 @@ GEM
     rswag-ui (2.4.0)
       actionpack (>= 3.1, < 7.0)
       railties (>= 3.1, < 7.0)
-    rubocop (1.25.1)
+    rubocop (1.25.0)
```
thilton-oddball, line 794: We don't want to roll this back.

thilton-oddball (changes requested):
One change needed so Rubocop won't be rolled back but otherwise looks good.

Gemfile.lock
alsleaders: Thanks Travis for catching that!

thilton-oddball (approved):

PR #8987, opened 2022-01-31
Refactor HypertensionSpecialIssueManager to be more rubyist

yoomlam:
## Description of change

This PR refactors `HypertensionSpecialIssueManager` to make it more idiomatic of Ruby styling, which makes it easier to read. Existing tests ensure the code behavior remains the same.

## Original issue(s)
N/A. (As a new team member, this is a good exercise to become familiar with the code.)

None, just a refactor.


app/services/fast_track/hypertension_special_issue_manager.rb
```diff
@@ -8,27 +8,31 @@ def initialize(submission)
       @submission = submission
     end
 
-    def add_special_issue
-      data = JSON.parse(submission.form_json, symbolize_names: true)
-      disabilities = data[:form526][:form526][:disabilities]
-      data[:form526][:form526][:disabilities] = add_rrd_to_disabilities(disabilities)
-      submission.update!(form_json: JSON.dump(data))
```
yoomlam, line 15: This is simplified and now on line 19.

app/services/fast_track/hypertension_special_issue_manager.rb
```diff
@@ -8,27 +8,31 @@ def initialize(submission)
...
-      rrd_code = 'RRD'
-      if disability[:specialIssues].blank?
-        disability[:specialIssues] = [rrd_code]
-      elsif !disability[:specialIssues].include? rrd_code
-        disability[:specialIssues].append(rrd_code)
+    def add_special_issue
+      disabilities.each do |disability|
+        add_rrd_code(disability) if hypertension_increase?(disability)
       end
+      submission.update!(form_json: JSON.dump(submission_data))
```
yoomlam, line 23: `disabilities` and `submission_data` are now lazily-initialized, private attributes.

app/services/fast_track/hypertension_special_issue_manager.rb
```diff
@@ -8,27 +8,31 @@ def initialize(submission)
...
+    end
+
+    def hypertension_increase?(disability)
+      disability[:diagnosticCode] == 7101 && disability[:disabilityActionType].downcase == 'increase'
+    end
+
+    RRD_CODE = 'RRD'
```
yoomlam, line 30: Extracted from method to be a constant.

app/services/fast_track/hypertension_special_issue_manager.rb
```diff
@@ -8,27 +8,31 @@ def initialize(submission)
...
+    def add_special_issue
+      disabilities.each do |disability|
+        add_rrd_code(disability) if hypertension_increase?(disability)
       end
+      submission.update!(form_json: JSON.dump(submission_data))
+    end
+
+    def hypertension_increase?(disability)
+      disability[:diagnosticCode] == 7101 && disability[:disabilityActionType].downcase == 'increase'
+    end
```
yoomlam, line 28: Extracted into a method so we know the intent of `diagnosticCode 7101`.

app/services/fast_track/hypertension_special_issue_manager.rb
```diff
@@ -8,27 +8,31 @@ def initialize(submission)
       @submission = submission
     end
 
-    def add_special_issue
-      data = JSON.parse(submission.form_json, symbolize_names: true)
-      disabilities = data[:form526][:form526][:disabilities]
-      data[:form526][:form526][:disabilities] = add_rrd_to_disabilities(disabilities)
-      submission.update!(form_json: JSON.dump(data))
+    def submission_data
+      @submission_data ||= JSON.parse(submission.form_json, symbolize_names: true)
```
NatashaJIbrahim, line 12: Is it an option to just move the JSON parse logic into the initializer and get rid of this method completely?

```
    def initialize(submission)
      @submission = submission
      @submission_data ||= JSON.parse(submission.form_json, symbolize_names: true)
    end

File: app/services/fast_track/hypertension_special_issue_manager.rb [hunk 6]
```diff
@@ -8,27 +8,31 @@ def initialize(submission)
...
     end
 
-    def add_rrd_to_disabilities(disabilities)
-      disabilities.each do |da|
-        add_rrd(da) if da[:diagnosticCode] == 7101 && da[:disabilityActionType].downcase == 'increase'
-      end
-      disabilities
+    def disabilities
```
NatashaJIbrahim, line 15: Does this need to be a public method?

app/services/fast_track/hypertension_special_issue_manager.rb
NatashaJIbrahim: I'm not sure we need to do this quite yet since "RRD" is only being invoked once and we don't have other codes. Also, this makes this available outside the class and that seems unnecessary. 

Stylistically I'm not a huge fan of adding this as a constant atm but it's not blocking.

NatashaJIbrahim (dismissed):
This looks great - much easier to read. We can take it one step further and make `#add_special_issue` the only public method.  I left a couple small comments but this is definitely a great improvement.

app/services/fast_track/hypertension_special_issue_manager.rb
yoomlam: Made methods private.

app/services/fast_track/hypertension_special_issue_manager.rb
yoomlam: My reasoning: since this is a special issue code known and used by external software, it should be a constant.
It is now a private constant since no code within our codebase uses it.

app/services/fast_track/hypertension_special_issue_manager.rb
```diff
@@ -9,26 +9,32 @@ def initialize(submission)
...
+
+    def hypertension_increase?(disability)
+      disability[:diagnosticCode] == 7101 && disability[:disabilityActionType].downcase == 'increase'
+    end
+
+    RRD_CODE = 'RRD'
+
+    # Must return an array containing special string codes for EVSS
+    def add_rrd_code(disability)
+      disability[:specialIssues] = [] if disability[:specialIssues].nil?
```
nanotone, line 36: This could be simplified to `disability[:specialIssues] ||= []`, I think.

NatashaJIbrahim (approved):

thilton-oddball (approved):

PR #8991, opened 2022-01-31
Cris.shupp api 10899

cshupp1:
## Description of change
[
](https://vajira.max.gov/browse/API-10899)
## Original issue(s)


NCHLiberty:
This fix is pending vets-api move to k8s, at which time it will be more feasible to  get a newer version of Poppler working which solves this problem. https://dsva.slack.com/archives/CBU0KDSB1/p1639169773357800

NCHLiberty:
poppler upgrade issue was resolved another way. no longer needed.

PR #8993, opened 2022-01-31
Refactor `DisabilityCompensationJob` to encapsulate Lighthouse interactions

yoomlam:
## Description of change

Refactors `DisabilityCompensationJob` to encapsulate Lighthouse interactions, which also reduces the number of arguments passed to methods.

## Original issue(s)
N/A. (As a new team member, this is a good exercise to become familiar with the code.)

None, just a refactor.

Existing tests ensure the code behavior remains the same.


app/workers/fast_track/disability_compensation_job.rb
```diff
@@ -52,16 +51,47 @@ def perform(form526_submission_id, full_name)
...
+      end
+
+      def bp_readings
+        @bp_readings ||= client.list_resource('observations')
+      end
 
-    def medications(client)
-      @medications ||= client.list_resource('medications')
-      @medications.present? ? FastTrack::HypertensionMedicationRequestData.new(@medications).transform : []
+      def observation_data
```
NatashaJIbrahim, line 82: Lighthouse uses this "observations" string but we're retrieving blood pressure readings specifically. I think this class bit easier to read when we are clear that we're looking at blood pressure readings and to avoid this "observations" key word that's really just something the client is passing on.

app/workers/fast_track/disability_compensation_job.rb
```diff
@@ -34,14 +34,13 @@ class DisabilityCompensationJob
 
     def perform(form526_submission_id, full_name)
       form526_submission = Form526Submission.find(form526_submission_id)
+      lh_data = LighthouseVetsHealth.new(form526_submission)
```
NatashaJIbrahim, line 37: I'd be more comfortable calling this "lighthouse_data" or simply "data". "lh_data" is a little confusing to read since you have to mentally keep track of abbreviations at that point.

app/workers/fast_track/disability_compensation_job.rb
```diff
@@ -52,16 +51,47 @@ def perform(form526_submission_id, full_name)
...
 
+    # class AccountNotFoundError < RuntimeError; end
+
     private
 
-    def bp_readings(client)
-      @bp_readings ||= client.list_resource('observations')
-      @bp_readings.present? ? FastTrack::HypertensionObservationData.new(@bp_readings).transform : []
-    end
+    class LighthouseVetsHealth
```
NatashaJIbrahim, line 58: I do like the idea of this work staying in the job file but I don't think this class inside a class is going to fly with pr review. If you want a nice reusable object I could see this being a struct that gets passed the submission and the account.

That way it's just a small object that returns formatted data. 🤷  maybe? Thoughts?

app/workers/fast_track/disability_compensation_job.rb
```diff
@@ -52,16 +51,47 @@ def perform(form526_submission_id, full_name)
...
+          bp_readings.present? ? FastTrack::HypertensionObservationData.new(bp_readings).transform : []
+      end
+
+      def medications
+        @medications ||= client.list_resource('medications')
+      end
+
+      def medication_request_data
+        @medication_request_data ||=
+          medications.present? ? FastTrack::HypertensionMedicationRequestData.new(medications).transform : []
```
NatashaJIbrahim, line 93: Same here - I don't see any reason to memoize this.

app/workers/fast_track/disability_compensation_job.rb
```diff
@@ -52,16 +51,47 @@ def perform(form526_submission_id, full_name)
...
+
+      def bp_readings
+        @bp_readings ||= client.list_resource('observations')
+      end
 
-    def medications(client)
-      @medications ||= client.list_resource('medications')
-      @medications.present? ? FastTrack::HypertensionMedicationRequestData.new(@medications).transform : []
+      def observation_data
+        @observation_data ||=
```
NatashaJIbrahim, line 83: I'm not sure why we're memoizing this. I think it just adds complexity. Am I missing a reason?

NatashaJIbrahim (commented):
Interesting thoughts here. I left some comments and questions.

app/workers/fast_track/disability_compensation_job.rb
yoomlam: I didn't pull the class out into a separate file because no other code uses it. The class exists to encapsulate Lighthouse-related methods and clearly distinguish the boundary between it and other methods that send emails, upload pdf, and set special issues. I don't see it as a struct, as in it's not simply a data object -- it's a Lighthouse client that uses an `Account` DB record and parses Lighthouse results.

app/workers/fast_track/disability_compensation_job.rb
yoomlam: I used `observation_data` because of the `HypertensionObservationData` class that is being used to `transform` the results. Since we are feeding in `bp_readings`, I'll rename the method -- suggestions?

app/workers/fast_track/disability_compensation_job.rb
yoomlam: `observation_data` is called twice and as a result, `FastTrack::HypertensionObservationData.new(bp_readings).transform` will be called twice. Since `bp_readings` hasn't changed between the 2 calls, the result from `observation_data` will be the same. Memoizing will avoid reinstantiating `HypertensionObservationData` and re-calling `transform` to get the same result.

app/workers/fast_track/disability_compensation_job.rb
yoomlam: Currently `medication_request_data` is called once, but there's nothing preventing us from calling it multiple times. I can remove the memoization for this and `observation_data` if it feels complex.

app/workers/fast_track/disability_compensation_job.rb
```diff
@@ -52,16 +51,47 @@ def perform(form526_submission_id, full_name)
...
+      end
+
+      def account
+        user_uuid = @form526_submission.user_uuid.presence
+        edipi = @form526_submission.auth_headers['va_eauth_dodedipnid'].presence
+        account = Account.find_by(idme_uuid: user_uuid) if user_uuid
+        account ||= Account.find_by(logingov_uuid: user_uuid) if user_uuid
+        account ||= Account.find_by(edipi: edipi) if edipi
+        # Should we raise and error instead of expecting a NoMethodError?
+        # raise AccountNotFoundError.new("for user_uuid: #{user_uuid}") unless account&.icn
```
yoomlam, line 74: Opinions on this question? See the RSpec that expect `NoMethodError`. If we know the cause for the error (i.e., `AccountNotFoundError`), we should make it clear in the error message.

julialeague: Yeah, I think that would be great - we could perhaps change that in a separate PR and track the change over in our Notion dashboard. Sound good?

spec/jobs/fast_track/disability_compensation_job_spec.rb
```diff
@@ -150,9 +150,11 @@
     context 'when an account for the user is NOT found' do
       before do
         allow(Account).to receive(:where).and_return Account.none
+        allow(Account).to receive(:find_by).and_return nil
       end
 
       it 'raises ActiveRecord::RecordNotFound exception' do
+        # NoMethodError is inconsistent with RecordNotFound description
```
julialeague, line 157: This could be fixed in that same PR where we raise a custom error 👍

app/workers/fast_track/disability_compensation_job.rb
NatashaJIbrahim: I'd probably call it "blood_pressure_readings" or "bp_readings".

app/workers/fast_track/disability_compensation_job.rb
NatashaJIbrahim: I totes hear you and I don't disagree BUT - I know stylistically the reviewers that need to approve this (not our team) wont let it get merged like that. I think it's a very rails-style thing that they like to see all classes in their own files with their own unit specs around them.

app/workers/fast_track/disability_compensation_job.rb
NatashaJIbrahim: I hear you but these data parsing classes `FastTrack::HypertensionMedicationRequestData` and `FastTrack::HypertensionObservationData` are idempotent, not very memory intensive, and aren't calling external services. There's actually no harm in them getting called again if they were to be called again. I think memoizing here is just adding unnecessary complexity.

app/workers/fast_track/disability_compensation_job.rb
NatashaJIbrahim: Ah - I see. I'm neutral here. I think I'd rather it be pulled out into it's own method and only called once. Again, just because I feel like memoizing here is a little extra. All stylistic though so 🤷‍♀️

app/workers/fast_track/disability_compensation_job.rb
NatashaJIbrahim: Second what @julialeague said here 👏  It'd be great to get an improved test in.

PR #8995, opened 2022-02-01
API-9832: Bring "get claims" v2 endpoint to parity with v1

hinzed1127:
## Description of change

Adds the following properties to each result in the `claims#index` method of v2 for displaying all claims for a given user:
- `date_filed`
- `end_product_code`
- `documents_needed`
- `requested_decision`
- `development_letter_sent`
- `decision_letter_sent`

## Original issue(s)
https://vajira.max.gov/browse/API-9832

This only updates the response for `claims#index`. The individual endpoint, `claims#show` still needs to be updated to work similarly. In the meantime, I had to skip one of the rswag tests, since it gets broken by the new mapping. This should get fixed as part of [API-9833](https://vajira.max.gov/browse/API-9833).


jw81:
Looks like Rubocop is cranky.

modules/claims_api/app/services/claims_api/bgs_to_lighthouse_claims_mapper_service.rb
```diff
@@ -10,6 +10,7 @@ def initialize(bgs_claim: nil, lighthouse_claim: nil)
     end
 
     def process
+      # binding.pry
```
jw81, line 13: Can we delete this?

jw81 (approved):

PR #8999, opened 2022-02-01
API-13104 Draft add appellant all the things

ElleeB:
## Description of change
This will be broken into multiple PRs

RFC!

Tests to be bulked up once we are in agreement on this work.

This PR creates an appellant class to manage much of the shared logic between Veteran and Non-Veteran claimant. In addition, it attempts to remove the responsibility of the Appeal (hlr, nod, sc) model to manage the appellant logic - this is only wired up for HLR V2 in this PR.

This update was in response to messiness encountered when preparing to stamp and auto-fill with our recently added claimant data.

 ## Original issue(s)
[Ticket](https://vajira.max.gov/browse/API-13104)


modules/appeals_api/app/models/appeals_api/higher_level_review.rb
```diff
@@ -110,49 +109,49 @@ def number_and_street
     end
 
     def city
-      veteran.dig('address', 'city') || ''
+      veteran_data.dig('address', 'city') || ''
```
ElleeB, line 114: renamed from veteran to veteran_data so no conflict with new veteran method above, which creates an Appellant.

modules/appeals_api/app/models/appeals_api/profile_veteran.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module AppealsApi
+  class ProfileVeteran
```
ElleeB, line 4: This was an unused class. I renamed so that I could move ahead with my intended veteran class.

cilestin: I beleive it's used when determining a veteran's profile data for building their VANotify data to send them emails.

cilestin:
So overall I like the direction but I think it can be simplified by removing the Veteran & NonVeteranClaimant classes, since the only point of them at this point is to hold the header prefix. Appellant can then take in the headers, veteran/claimant form_data attribute, and the header prefix in the initializer.

I'd double check the move from Veteran to ProfileVeteran (aside: why not VeteranProfile?) isn't being used in one of our jobs or on the model somehow. As I commented above, I think it's being used for the `email_identifier` methods on our models for eventually producing the VANotify event info.

ElleeB:
> And since you won't need the NonVeteranClaimant class, you can remove it as part of a different PR to keep your lines out of the danger zone ;)

I wasn't entirely sure that I was done with the veteran and non veteran claimant classes, but you may be right. For one, I was going to move the profile veteran stuff into the veteran class. On that note, I did update the use of the profile veteran stuff in the code that talks to Claims::Veteran. I'll look elsewhere to be sure, but again, I intend to use the functionality, but move it into the veteran class...it was just a backwards-ish way of me organizing my changes.

modules/appeals_api/app/models/appeals_api/profile_veteran.rb
ElleeB: You were totes right. I've refactored this into the current PR, which is now 5-bagillion lines long.

modules/appeals_api/app/models/appeals_api/higher_level_review.rb
```diff
@@ -101,49 +115,49 @@ def number_and_street
...
     def veteran_homeless?
       form_data&.dig('data', 'attributes', 'veteran', 'homeless')
     end
 
     def email
-      veteran&.dig('emailAddressText').to_s.strip
+      veteran_data&.dig('emailAddressText').to_s.strip
     end
 
     def email_v2
```
ElleeB, line 159: This will be removed in PR part II. Removing here will cause form data test to fail and then domino to our generator spec.

modules/appeals_api/spec/models/appellant_spec.rb
```diff
@@ -0,0 +1,179 @@
...
+    it { expect(veteran_appellant.country_code).to eq 'US' }
+    it { expect(claimant_appellant.country_code).to eq 'US' }
+  end
+
+  describe '#zip_code_5' do
+    it { expect(veteran_appellant.zip_code_5).to eq '30012' }
+    it { expect(claimant_appellant.zip_code_5).to eq '48070' }
+  end
+
+  # describe '#international_postal_code' do
```
ElleeB, line 94: May update claimant to an international address to test this.

modules/appeals_api/spec/models/appellant_spec.rb
```diff
@@ -0,0 +1,179 @@
...
+    it { expect(veteran_appellant.phone_line_number).to eq '1111' }
+    it { expect(claimant_appellant.phone_line_number).to eq '1100' }
+  end
+
+  describe '#phone_ext' do
+    it { expect(veteran_appellant.phone_ext).to eq 'x2' }
+    it { expect(claimant_appellant.phone_ext).to eq 'x4' }
+  end
+
+  # describe '#international_number' do #### need to see an example of this printing correctly (not sure how this works)
```
ElleeB, line 140: Something is off about this, even in our current HLRs. Need to dig in.

PR #9001, opened 2022-02-01
2452: only retry specs in CI environment

kpethtel:
## Description of change
Someone recently added a retry to the spec helper. This is not desirable locally and can cause confusion in the debugging process. It should only happen in CI. 

## Original issue(s)
department-of-veterans-affairs/va-mobile-team#2452


kreek (approved):
LGTM 🚀

rileyanderson (approved):

PR #9016, opened 2022-02-02
[33352] sets authentication SAML `force_authn` to false

bramleyjl:
## Description of change

This PR changes `vets-api`'s default `force_authn` argument that is passed to to ISAM. Due to [some recent changes](https://dsva.slack.com/archives/CSFV4QTKN/p1643739020294789) made by IAM setting the `force_authn` argument to true is causing an infinite loop on login for development and local environments. This change fixes that loop; the `force_authn` argument that is passed is mostly unused, meaning users attempting to authenticate will still need to enter credentials unless it is an inbound SSO login.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#33352

Unit specs updated, localhost logins with ID.me, Login.gov, & DSLogon are unaffected.


bosawt (approved):
- Confirmed ID.me auth works
- Confirmed Login.gov auth works, but requires two force authentications (expected for now)

PR #9018, opened 2022-02-02
34690 bypass ssl verification in staging for test purposes

dillo:
## Description of change

- We're bypassing SSL verification for the MuleSoft integration in staging in order to conduct testing and meet some business deadlines.
- We will be opening a ticket with operations to handle the SSL verification failure in staging against the MuleSoft staging environment. In the meantime, testing can still continue with this workaround.
- The code will not be promoted to production(feature will not be toggled on) until Ops has fixed the problem in staging and also only when we can make sure that it won't be happening in production as well.
- The code to bypass the SSL verification will be removed before we start testing in production.
## Original issue(s)
department-of-veterans-affairs/va.gov-team#34690


thilton-oddball:
I don't see where you are limiting this to just the staging environment. Any changes made here would get deployed to prod.

dillo:
> I don't see where you are limiting this to just the staging environment. Any changes made here would get deployed to prod.

@thilton-oddball The feature flag in this line determines if we're using the existing salesforce-carma client or the new mulesoft client: https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/services/form1010cg/service.rb#L50

thilton-oddball (approved):

PR #9022, opened 2022-02-03
2365 payment history api endpoint

cadibemma:
## Description of change

The mobile app is adding a payment history feature to the app. This is like [payment history on va.gov](https://staging.va.gov/va-payment-history/). We completed API design/documentation last sprint. The endpoint is ready to be wired up with a route and a controller in the mobile module that fronts the existing payment service in vets-api.

## Original issue(s)
department-of-veterans-affairs/va-mobile-app#2365


app/models/adapters/payment_shared_adapter.rb
```diff
@@ -0,0 +1,33 @@
...
+      }
+    end
+
+    def mask_account_number(address_eft, all_but = 4, char = '*')
+      return if address_eft.blank?
+
+      account_number = address_eft[:account_number]
+      return if account_number.blank?
+
+      account_number&.gsub(/.(?=.{#{all_but}})/, char)
```
kpethtel, line 22: This safety operator shouldn't be necessary due to the guard above.

modules/mobile/app/controllers/mobile/v0/payment_history_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+
+require_dependency 'mobile/application_controller'
+require 'adapters/payment_shared_adapter'
+
+module Mobile
+  module V0
+    class PaymentHistoryController < ApplicationController
+      def index
+        start_date = params[:startDate] || (DateTime.now.utc.beginning_of_day - 1.year).iso8601
+        end_date = params[:endDate] || DateTime.now.utc.beginning_of_day.iso8601
```
kpethtel, line 11: Should we use end of day here instead of beginning? Idk how the payments system works (maybe the timestamp doesn't matter), but it seems possible there could be payments that were made later in the day. If I had to guess, I'd speculate the start and end dates coming in from the params are probably dates and GetPaginatedList is coercing them into datetimes. In which case, maybe those get coerced to beginning of day too.

modules/mobile/app/controllers/mobile/v0/payment_history_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+        payments = adapt_payments(response.dig(:payments, :payment))
+        list, meta = paginate(payments, validated_params)
+        render json: Mobile::V0::PaymentHistorySerializer.new(list, meta)
+      end
+
+      private
+
+      def adapt_payments(payments)
+        payments.map do |payment|
+          if payment.dig(:return_payment, :check_trace_number).blank?
```
kpethtel, line 33: Having a conditional within a map like this will leave nils when the condition isn't met. You may need to change how you're doing this or use `compact` to get rid of the nils.

modules/mobile/app/controllers/mobile/v0/payment_history_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+              payment_method: ::Adapters::PaymentSharedAdapter.new.get_payment_method(payment),
+              bank: payment.dig(:address_eft, :bank_name),
+              account: ::Adapters::PaymentSharedAdapter.new.mask_account_number(payment[:address_eft])
+            )
+          end
+        end
+      end
+
+      def paginate(payments, validated_params)
+        # payments = payments.filter do |payment|
```
kpethtel, line 47: What's up with this?

modules/mobile/spec/request/payment_history_request_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+
+      it 'returns 200' do
+        expect(response).to have_http_status(:ok)
+      end
+
+      it 'matches expected schema' do
+        expect(response.body).to match_json_schema('payment_history')
+      end
+
+      it 'includes the expected properties for payment history' do
```
kpethtel, line 35: It might also be good to test the case where `mask_account_number` returns nil.

kpethtel (dismissed):
A few minor comments, but overall this looks good. 👍

modules/mobile/app/controllers/mobile/v0/payment_history_controller.rb
cadibemma: That is for the pagination functionality (which I'll be working on this sprint)

app/models/adapters/payment_shared_adapter.rb
cadibemma: This part is from the va.gov api which is why I didn't tamper with it outside of moving it to a shared adapter to use it within the mobile api (to reduce duplicate code).

modules/mobile/app/controllers/mobile/v0/payment_history_controller.rb
cadibemma: End of day would probable make more considering the later payment within the same day. I will look into the payment systems more to determine why beginning of day way used for va.gov. The date params is used for the pagination (which is still WIP and will be finishing it this sprint).

modules/mobile/app/controllers/mobile/v0/payment_history_controller.rb
cadibemma: Thanks! - made the correction

modules/mobile/spec/request/payment_history_request_spec.rb
cadibemma: Cool! I updated specs

kreek (approved):
I helped with the refactor so I'd like to get @kpethtel, @aherzberg, or @jperk51 approval too, but this LGTM 🚀

jperk51 (approved):
LGTM

cadibemma (commented):
LGTM

cadibemma:
> A few minor comments, but overall this looks good. 👍

All is fixed

PR #9024, opened 2022-02-03
Hypertension RRD: improve error handling: add AccountNotFoundError

yoomlam:
## Description of change

In `DisabilityCompensationJob`, if we know the cause of the error is due to an account not found in the DB, then raise the explicitly-named `AccountNotFoundError`. This will aid in debugging job failures.

## Original issue(s)
None.

Updated test to use `AccountNotFoundError` instead of `NoMethodError` due to a `nil` account.


spec/jobs/fast_track/disability_compensation_job_spec.rb
```diff
@@ -130,8 +130,8 @@
       it 'raises an error' do
         Sidekiq::Testing.inline! do
           expect do
-            subject.new.perform_async(submission_without_account_or_edpid.id, user_full_name)
```
yoomlam, line 133: I saw weird failures when using `subject.new`, so I used `FastTrack::DisabilityCompensationJob` similar to other tests.

allthesignals: Curious why `subject.new` was used here. Seems using `subject` is fine in other places because it calls a Sidekiq class method `subject.perform_async`?

allthesignals: maybe it raised an error and then passed because the test invokes `subject.new` (rather than for the intended simulated error scenario)

app/workers/fast_track/disability_compensation_job.rb
```diff
@@ -79,7 +81,10 @@ def send_fast_track_engineer_email_for_testing(form526_submission_id, error_mess
     end
 
     def get_icn(form526_submission)
-      account(form526_submission).icn.presence
+      account = account(form526_submission)
+      raise AccountNotFoundError, "for user_uuid: #{form526_submission.user_uuid} or their edipi" unless account
+
+      account.icn.presence
```
allthesignals, line 87: Looks good to me!

spec/jobs/fast_track/disability_compensation_job_spec.rb
allthesignals: Curious if `subject.perform_async` works too but I don't really care either way

allthesignals (approved):

allthesignals:
Nice catch! I bet that old test was passing because it was failing for "meta" reasons `subject.new.perform_async` literally raises nomethod error and the original test writer thought this was the correct error. Just a theory.

app/workers/fast_track/disability_compensation_job.rb
```diff
@@ -52,6 +52,8 @@ def perform(form526_submission_id, full_name)
       end
     end
 
+    class AccountNotFoundError < RuntimeError; end
```
nanotone, line 55: It looks like the vast majority of custom exceptions in this repo subclass `StandardError` rather than `RuntimeError`. I don't have a strong preference.

app/workers/fast_track/disability_compensation_job.rb
```diff
@@ -79,7 +81,10 @@ def send_fast_track_engineer_email_for_testing(form526_submission_id, error_mess
     end
 
     def get_icn(form526_submission)
-      account(form526_submission).icn.presence
+      account = account(form526_submission)
```
nanotone, line 84: I think this method would be more readable if the local variable and the method had different names.

nanotone (commented):
I have two recommendations, but neither are hard-blocking. Looks good!

app/workers/fast_track/disability_compensation_job.rb
yoomlam: Given what you said and reading https://www.honeybadger.io/blog/ruby-exception-vs-standarderror-whats-the-difference/ and https://www.honeybadger.io/blog/ruby-custom-exceptions/, I'll change it to `StandardError`.

app/workers/fast_track/disability_compensation_job.rb
```diff
@@ -79,16 +81,19 @@ def send_fast_track_engineer_email_for_testing(form526_submission_id, error_mess
...
     def account(form526_submission)
       user_uuid = form526_submission.user_uuid.presence
       edipi = form526_submission.auth_headers['va_eauth_dodedipnid'].presence
       # rubocop:disable Lint/UselessAssignment
-      account = Account.where(idme_uuid: user_uuid).first if user_uuid
-      account ||= Account.where(logingov_uuid: user_uuid).first if user_uuid
-      account ||= Account.where(edipi: edipi).first if edipi
+      account = Account.find_by(idme_uuid: user_uuid) if user_uuid
+      account ||= Account.find_by(logingov_uuid: user_uuid) if user_uuid
+      account ||= Account.find_by(edipi: edipi) if edipi
```
yoomlam, line 96: I got inconsistent failures with this change but it eventually passed. `find_by` should call `where`, which is set up to return `Account.none` in [the test](https://github.com/department-of-veterans-affairs/vets-api/pull/9024/files#diff-45106a8a44ee0d5f717fd239680d2e9bd7853c549f03ec4ae26f094d207c63e2R159).

spec/jobs/fast_track/disability_compensation_job_spec.rb
yoomlam: I played around with it and don't understand why some variations work and others don't (in the back of my mind, I blaming `spring`). Regardless, it would be nice to have all the tests call `DisabilityCompensationJob` the same way.

app/workers/fast_track/disability_compensation_job.rb
yoomlam: I ended up adding `allow(Account).to receive(:find_by).and_return nil` to the test setup due to [this test failure](https://github.com/department-of-veterans-affairs/vets-api/pull/9024/checks?check_run_id=5058823500)

yoomlam: One difference is `where` orders the results but this is irrelevant for `idme_uuid` and `logingov_uuid` because they have DB uniqueness constraints. But there is an argument to use `where` for `edipi` but is ordering by `id` appropriate?

```ruby
Account.find_by(idme_uuid: "asdf")
2022-02-03 15:57:34.147248 D [18204:67400 log_subscriber.rb:126] ActiveRecord::Base --   Account Load (13.6ms)  SELECT "accounts".* FROM "accounts" WHERE "accounts"."idme_uuid" = $1 LIMIT $2  [["idme_uuid", "asdf"], ["LIMIT", 1]]
=> nil

Account.where(idme_uuid: "asdf").first
2022-02-03 15:57:47.214873 D [18204:67400 log_subscriber.rb:126] ActiveRecord::Base --   Account Load (7.1ms)  SELECT "accounts".* FROM "accounts" WHERE "accounts"."idme_uuid" = $1 ORDER BY "accounts"."id" ASC LIMIT $2  [["idme_uuid", "asdf"], ["LIMIT", 1]]
```

nanotone (approved):
Looks good to me!

NatashaJIbrahim (approved):

thilton-oddball (approved):

PR #9026, opened 2022-02-03
Api-13104 Part 1: Add appellant class and spec

ElleeB:
**[Parent PR](https://github.com/department-of-veterans-affairs/vets-api/pull/8999)**

## Description of change
Adds an appellant class and the corresponding spec. Removes non veteran claimant and corresponding spec.

Follow-up tickets will include updates to the hlr model, hlr_v2 form_data and structure

## Original issue(s)
[Ticket](https://vajira.max.gov/browse/API-13104


modules/appeals_api/app/models/appeals_api/appellant.rb
```diff
@@ -149,11 +186,23 @@ def address_combined
...
+      # TODO: revisit - in what ways are we differentiating between our handling of international vs domestic
       phone_country_code == '1'
     end
 
     def address
-      form_data['address'].presence || {}
+      form_data['address'] || ''
+    end
+
+    def mpi_veteran
```
cilestin, line 197: This should guard against being used with a `Claimant` type, and return nil if the appellant is a claimant.

modules/appeals_api/app/models/appeals_api/appellant.rb
```diff
@@ -129,10 +142,34 @@ def ssn_last_four
...
     end
 
+    def timezone
+      form_data&.dig('timezone').presence&.strip
+    end
+
+    def submission_email_identifier
+      email_id = "#{type.downcase}_email"
+
+      return { id_type: email_id, id_value: email } if email.present?
```
cilestin, line 152: Could you double check how `id_type` is being used in the email notification handling and ensure that having a an `id_type: "claimant_email"` will be ok w/ VANotify (assuming we even send this data to them)?

modules/appeals_api/app/models/appeals_api/appellant.rb
```diff
@@ -129,10 +142,34 @@ def ssn_last_four
...
+
+      type == 'Veteran' && auth_headers.exclude?('X-VA-Claimant-Last-Name')
+    end
+
     private
 
     attr_accessor :auth_headers, :form_data
 
+    def header_prefix
+      type == 'Veteran' ? '' : '-Claimant'
```
cilestin, line 170: You may as well memoize this since it's called in so many other methods we're likely to use.

Also, any reason the `type` needs to be a string? Not a blocker, but I don't see anything that indicates this couldn't be passed into the initializer as a symbol.

cilestin (changes requested):

modules/appeals_api/app/models/appeals_api/appellant.rb
ElleeB: Yes I'll look into that now! I thought the id_type was just for us, but I'm going to make sure.

modules/appeals_api/app/models/appeals_api/appellant.rb
ElleeB: I never even thought to pass it as a symbol. Going to memoize and look into benefits of passing a symbol.

cilestin (approved):

modules/appeals_api/app/models/appeals_api/appellant.rb
```diff
@@ -0,0 +1,214 @@
...
+         address['addressLine3']].compact.map(&:strip).join(' ')
+    end
+
+    def domestic_phone?
+      # TODO: revisit - in what ways are we differentiating between our handling of international vs domestic
+      phone_country_code == '1'
+    end
+
+    def address
+      form_data['address'] || ''
```
cilestin, line 198: Whoops missed this. This will result in an error if there's no address supplied but we call `appellant.city`, since we'll be trying to pull a key from a string with `address['city']`.

cilestin (changes requested):

modules/appeals_api/app/models/appeals_api/appellant.rb
ElleeB: My bad. It was an empty hash - supposed to be anyway.

ElleeB: Fixed and added a test.

PR #9027, opened 2022-02-03
Refactor HypertensionPdfGenerator

allthesignals:
## Description of change

This PR refactors the HypertensionPdfGenerator class by taking a few cues from other similar pdf-generator classes:

1. [Use an instance variable for the Prawn pdf rather than passing reference as arguments to methods](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/appeals_api/app/services/appeals_api/pdf_construction/notice_of_disagreement/pages/hearing_type_and_additional_issues.rb#L11)
2. [Encapsulate certain logical flags in semantic helper methods](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/appeals_api/app/services/appeals_api/pdf_construction/notice_of_disagreement/pages/hearing_type_and_additional_issues.rb#L47)
3. Move static prose into constants
4. "Flattens" method calls out so that methods are not calling other methods, move from tree-like to linear.
5. Remove two unused methods: #add_medications_table and #add_blood_pressure_table — happy to add them back but they are preserved in commit history....
6. Sets certain methods to private
7. [Generally follow code style in other Prawn-related classes](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/appeals_api/app/services/appeals_api/pdf_construction/higher_level_review/pages/v2/additional_issues.rb)

Not married to any of this and happy to break this up into smaller commits!

One future improvement might be to separate text content from direct calls to Prawn (to take another cue from other similar classes). So, for example, `generate` might simply be a string of calls to `pdf.text` with methods simply generating text content. This is limited by necessary calls to `pdf.table`.

Another future improvement might be to avoid mixing of the "html-like" DSL in text content with the Prawn API options (for example, `pdf.text "<font size="11">Hi</font>"` vs `pdf.text "Hi", size: 11`). There isn't really a great templating option for this. Multiline text interpolation syntax highlighting looks kinda bad.

Lastly, another option might be to use ERB templates and convert that into a PDF, similar to some of the mailer classes. Most of the app does indeed use Prawn PDF generation, however, so it'd be an idiosyncratic approach, but might be easier to maintain _if_ layout and composition of this PDF doesn't require anything beyond basic word processing and tables.


app/services/fast_track/hypertension_pdf_generator.rb
```diff
@@ -2,31 +2,43 @@
...
+    private
+
+    def blood_pressure_data?
+      @blood_pressure_data.length.positive?
     end
 
-    def add_intro(pdf)
-      patient_name = stringify_patient
+    def medications?
+      @medications.length > 1
```
NatashaJIbrahim, line 35: This is interesting and raises a _slightly out of scope question_ for me about the logic here. @tadhg-ohiggins do we want this to be `>` or `=>`? Right now `medications?` only returns true if there's more than one medication but not if there's only one.

app/services/fast_track/hypertension_pdf_generator.rb
```diff
@@ -2,31 +2,43 @@
 
 module FastTrack
   class HypertensionPdfGenerator
-    attr_accessor :patient, :bp_data, :medications
+    attr_accessor :patient, :blood_pressure_data, :medications, :pdf
 
-    def initialize(patient, bp_data, medications)
+    def initialize(patient, blood_pressure_data, medications)
+      @pdf = Prawn::Document.new
       @patient = patient
```
NatashaJIbrahim, line 9: I think this `patient` object is a name hash right? Would it help to rename it something like `patient_name_hash`?

NatashaJIbrahim (approved):
I think this is a solid improvement.

app/services/fast_track/hypertension_pdf_generator.rb
NatashaJIbrahim: Cool - @allthesignals do you mind updating that in a follow up commit with it's own message?

allthesignals: @NatashaJIbrahim yes will do!

allthesignals:
I made a few changes to how we reference instance variables. There isn't current use for exposing certain instance variables... the accessor feature of ruby classes aren't being used so I went ahead and removed them: https://stackoverflow.com/questions/25571642/ruby-private-and-public-accessors

app/services/fast_track/hypertension_pdf_generator.rb
```diff
@@ -2,31 +2,41 @@
 
 module FastTrack
   class HypertensionPdfGenerator
-    attr_accessor :patient, :bp_data, :medications
```
allthesignals, line 5: I looked around and can't find any reason to make these accessible outside the class. I think until we need it we should hold off from making them accessible.

app/services/fast_track/hypertension_pdf_generator.rb
```diff
@@ -2,31 +2,41 @@
...
+    private
+
+    def blood_pressure_data?
+      @blood_pressure_data.length.positive?
     end
 
-    def add_intro(pdf)
-      patient_name = stringify_patient
+    def medications?
+      @medications.any?
```
allthesignals, line 33: Before, it checked if there were more than 1 medication, but instead it should be checking if there are any medications to show.

app/services/fast_track/hypertension_pdf_generator.rb
```diff
@@ -171,72 +125,78 @@ def add_medications_intro(pdf)
...
+      '<link href="https://www.ecfr.gov/current/title-38/chapter-I/part-3/' \
+      'subpart-A/subject-group-ECFR7629a1b1e9bf6f8/section-3.159"><color rgb="0000ff">' \
+      'operationalize existing statutory rules</color><link> in 38 U.S.C § 5103a(d).',
+      "\n",
+      'Not included in this document:',
+      ' •  Private medical records',
+      ' •  VAMC data for clinics using CERNER Electronic Health Record system ' \
+      '(Replacing VistA, but currently only used at Mann-Grandstaff VA Medical Center in Spokane, Washington)',
+      ' •  JLV/Department of Defense medical records'
+    ].freeze
```
yoomlam, line 176: You could get rid of quotes and backslashes using [heredoc syntax](https://riptutorial.com/ruby/example/2933/multiline-strings) for multiline strings.

allthesignals: I'll give that a shot and see if it helps with readability. Thanks!

app/services/fast_track/hypertension_pdf_generator.rb
```diff
@@ -171,72 +125,78 @@ def add_medications_intro(pdf)
...
         if medication['dosageInstructions'].present?
-          pdf.text "Dosage instructions: #{medication['dosageInstructions'].join('; ')}"
+          @pdf.text "Dosage instructions: #{medication['dosageInstructions'].join('; ')}"
         end
-        pdf.text "\n", size: 8
+        @pdf.text "\n", size: 8
       end
 
-      pdf
+      @pdf
```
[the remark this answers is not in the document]
allthesignals: Ahh yes you're right. Thanks! Will fix

app/services/fast_track/hypertension_pdf_generator.rb
allthesignals: > You could get rid of quotes and backslashes using [heredoc syntax](https://riptutorial.com/ruby/example/2933/multiline-strings) for multiline strings.

I'm very interested in makign this easier to work with. It looks like you have to do some [weird stuff](https://www.ruby-forum.com/t/array-of-here-documents/128579/5) to get it to work in this specific context:

```ruby
stuff = [ <<DOC1, <<DOC2, <<DOC3 ]
Hello World
DOC1
How’s it going?
DOC2
It’s just crazy, thanks.
DOC3

p stuff
#=> [“Hello World\n”, “How’s it going?\n”, “It’s just crazy, thanks.
\n”]
```

I'll toy around some more but it would be nice to remove this mix of string concatenation

app/services/fast_track/hypertension_pdf_generator.rb
```diff
@@ -69,98 +68,46 @@ def add_blood_pressure_intro(pdf)
...
           size: 10, border_width: 0, background_color: 'f3f3f3'
         }
       )
 
-      pdf.text "\n"
-      pdf.text rating_schedule_link,
-               inline_format: true, color: '0000ff', size: 11
-      pdf
+      @pdf.text "\n"
+      @pdf.text RAGING_SCHEDULE_LINK,
```
thilton-oddball, line 102: Guessing this should be RATING_SCHEDULE_LINK?

app/services/fast_track/hypertension_pdf_generator.rb
```diff
@@ -171,72 +118,72 @@ def add_medications_intro(pdf)
...
+        '40%', 'Diastolic pressure 120 or more'
+      ],
+      [
+        '60%', 'Diastolic pressure 130 or more'
+      ]
+    ].freeze
 
-      pdf
-    end
+    RAGING_SCHEDULE_LINK =  "<link href='" \
```
thilton-oddball, line 185: Guessing this should be RATING_SCHEDULE_LINK?

thilton-oddball (approved):
A couple of suspected typos but otherwise looks good.

allthesignals:
@thilton-oddball yes, thank you! Fixed it.

thilton-oddball (approved):

PR #9029, opened 2022-02-03
API-13104 Part 2: HLR appellant methods

ElleeB:
This PR: Part 2 Updates the HLR model to handle the new appellant objects - updates for V2 only

Waiting for Part 1: https://github.com/department-of-veterans-affairs/vets-api/pull/9026
Part 1 added the appellant class and related specs

**this pr will fail until part 1 is merged**


modules/appeals_api/app/models/appeals_api/higher_level_review.rb
```diff
@@ -274,7 +327,7 @@ def data_attributes
       form_data&.dig('data', 'attributes')
     end
 
-    def veteran
+    def veteran_data
```
ElleeB, line 330: Renamed so not to conflict with the new veteran method that creates an Appellant

modules/appeals_api/spec/models/higher_level_review_spec.rb
```diff
@@ -341,8 +341,102 @@
...
+      it { expect(subject.class).to eq AppealsApi::Appellant }
+    end
+
+    describe '#veteran' do
+      subject { higher_level_review_v2.veteran }
+
+      it { expect(subject.class).to eq AppealsApi::Appellant }
+    end
+
+    describe '#signing_appellant' do
```
ElleeB, line 353: This makes the assumption that we want to sign with the Claimant's name if a claimant is present. If not, let's discuss and update.

modules/appeals_api/app/models/appeals_api/higher_level_review.rb
```diff
@@ -101,49 +115,49 @@ def number_and_street
     end
 
     def city
-      veteran.dig('address', 'city') || ''
+      veteran_data.dig('address', 'city') || ''
```
cilestin, line 118: We've hooked up the veteran appellant object - let's use it for as many of these `veteran_data` calls as we can. It's exactly why Appellant was written, yeah? :)

cilestin (changes requested):

modules/appeals_api/app/models/appeals_api/higher_level_review.rb
```diff
@@ -199,6 +213,32 @@ def date_signed_yyyy
...
+
+    def signing_appellant
+      claimant.signing_appellant? ? claimant : veteran
+    end
+
+    def signature_v2
+      "#{signing_appellant.full_name[0...180]}\n- Signed by digital authentication to api.va.gov"
+    end
+
+    def date_signed_v2_mm
```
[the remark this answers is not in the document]
ElleeB: Do you mean the actual formatting bit? I feel like having the hlr in charge of date_signed makes sense. Do you mean the mm, dd, yyyy? But yes, let's make a note.

modules/appeals_api/app/models/appeals_api/higher_level_review.rb
ElleeB: If you go to the "all the things" draft pr, you can see how v2 data and structure work with the new setup.

cilestin (approved):

PR #9034, opened 2022-02-04
[API-13066] - ClaimsApi - Add CCG support for V2

jw81:
## Description of change

Adds CCG support for V2

## Original issue(s)
[API-13066](https://vajira.max.gov/browse/API-13066)


modules/claims_api/app/controllers/claims_api/v2/application_controller.rb
```diff
@@ -115,6 +119,40 @@ def user_represents_veteran?
 
         rep.poa_codes.include?(veteran_poa_code)
       end
+
+      private
+
+      def validate_ccg_token!
```
hinzed1127, line 125: seems like overkill to abstract/consolidate with a single duplication, but just noting that this is a duplication of the same `validate_ccg_token` method in the v1 `ApplicationController`

modules/claims_api/app/controllers/claims_api/v2/application_controller.rb
```diff
@@ -115,6 +119,40 @@ def user_represents_veteran?
...
+          'GET' => 'claim.read',
+          'PUT' => 'claim.write',
+          'POST' => 'claim.write'
+        }
+
+        @is_valid_ccg_flow ||= client.token_valid?(audience: claims_audience,
+                                                   scope: request_method_to_scope[request.method],
+                                                   token: token)
+        raise ::Common::Exceptions::Forbidden unless @is_valid_ccg_flow
+      end
```
hinzed1127, line 139: seems like overkill to abstract/consolidate with a single duplication, but just noting that this is a duplication of the same `validate_ccg_token` method in the v1 `ApplicationController`

hinzed1127 (approved):
Looks good to me. The only other thing I thought about when looking over all the tests is something that I've passively noticed before, which is all of the boilerplate mocking that we have to do in certain places. I don't think it's worth the effort in this PR, and it's already in other places, but there's some potential for some stubbing/mocking functions to take care of the repetitive things we're doing here for CCG (and probably other things in other request specs). I'm thinking something similar to what's done by [stub_poa_verification](https://github.com/department-of-veterans-affairs/vets-api/blob/master/spec/support/poa_stub.rb):

```ruby
def stub_ccg_verification(is_valid: true)
	allow(JWT).to receive(:decode).and_return(nil)
	allow(Token).to receive(:new).and_return(ccg_token)
	allow_any_instance_of(TokenValidation::V2::Client).to receive(:token_valid?).and_return(is_valid)
end

# E.g. specify validity, or allow it to default to true
stub_ccg_verification(is_valid: true)
stub_ccg_verification(is_valid: false)
stub_ccg_verification # defaults to true
```

PR #9037, opened 2022-02-04
2365 payment history refactor

kreek:
## Description of change
This PR is to reduce the size of the [mobile api's PR](https://github.com/department-of-veterans-affairs/vets-api/pull/9022) for adding payment history. We are adding an adapter to sit between the BGS payment service's response and the serializer. The mobile api then [subclasses the adapter](https://github.com/department-of-veterans-affairs/vets-api/blob/7a91422e28295aa539b63efb1f662e01b4940fc8/modules/mobile/app/models/mobile/v0/adapters/payment_history_adapter.rb) as we include fields (id and date) that the core API does not. The mobile API will also only display payments while the core API displays payments and return payments.

I added a use case to the controller specs (plus a vcr cassette) for when there are both payments and return payments.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va-mobile-app/issues/2365


spec/controllers/v0/profile/payment_history_controller_spec.rb
```diff
@@ -6,18 +6,32 @@
   let(:user) { create(:evss_user) }
 
   describe '#index' do
-    context 'with a valid bgs response' do
-      it 'returns true if a logged-in user has a valid va file number' do
+    context 'with only regular payments' do
+      it 'returns only payments and no return payments' do
```
kpethtel, line 10: This is probably so edge-casey to not merit a test, but what about the case when there are only return payments?

kpethtel (approved):
Looks good 👍

thilton-oddball (approved):

spec/controllers/v0/profile/payment_history_controller_spec.rb
kreek: 👍 That would be good to add. This PR is blocking Chika's mobile api PR (which is blocking FE) so I think I'll get this in and include that to the upcoming PRs for date filtering.

PR #9039, opened 2022-02-04
DisabilityCompensationJob: reformat failure email using HTML

yoomlam:
## Description of change

When a `DisabilityCompensationJob` fails and sends an email, the email is now formatted using HTML so that new lines are apparent.

Also updated the email subject to use "Rapid Ready for Decision (RRD)".

## Original issue(s)
Follow up to PR #9012.

The email is being sent with metadata:
```
Content-Type: text/html; charset=US-ASCII
```
which means the new lines in the email body are not apparent when viewing the email.

Rather than figure out how to change the `Content-Type` to `plain/text`, let's keep it as `text/html` so that we have the option of making it look nicer down the road.

Existing tests updated to match the new subject line and body.


NatashaJIbrahim (approved):
Good catch! 🐼

app/workers/rapid_ready_for_decision/disability_compensation_job.rb
```diff
@@ -59,13 +59,16 @@ def medications(client)
     def send_fast_track_engineer_email_for_testing(form526_submission_id, error_message, backtrace)
       # TODO: This should be removed once we have basic metrics
       # on this feature and the visibility is imporved.
-      body = "A claim just errored on the #{Rails.env} environment " \
-             "with submission id: #{form526_submission_id} and job_id #{jid}." \
-             "The error was: #{error_message}. The backtrace was:\n #{backtrace.join(",\n ")}"
+      body = <<~BODY
+        A claim errored in the #{Rails.env} environment \
```
thilton-oddball, line 63: If you want to get an accurate response for the environment, you will need to use Settings.vsp_environment instead of Rails.env. I believe our dev, staging, and sandbox environments all have Rails.env set to 'production'.

NatashaJIbrahim: Thanks for catching this. I _think_ we probably want to catch errors in both Staging and Production. Dev wont send an email so 🤷  but it would probably be best to explicitly exclude it.

yoomlam: Updated.

thilton-oddball (approved):

PR #9040, opened 2022-02-04
API-13104 Part3: Update appellant hlr data and structure

ElleeB:
[Part 1:](https://github.com/department-of-veterans-affairs/vets-api/pull/9026)
Part 1 added the appellant class and related specs

[Part 2:](https://github.com/department-of-veterans-affairs/vets-api/pull/9029)
Part 2 Updated the HLR model to handle the new appellant objects - updates for V2 only

This PR updates the data and structure files to use the new appellant objects and related methods in order to print on the pdf.


modules/appeals_api/app/services/appeals_api/pdf_construction/higher_level_review/v2/structure.rb
```diff
@@ -259,15 +282,18 @@ def whiteout(pdf, at:, width:, height: 15)
           pdf.fill_color '000000'
         end
 
-        def whiteout_line(pdf, attr, text_override: nil, length_for_override: SHORT_EMAIL_LENGTH)
+        def whiteout_line(pdf, attr, text:, text_override: nil, length_for_override: SHORT_EMAIL_LENGTH)
```
cilestin, line 285: Ugh, sorry I didn't notice this before but this looks like this method has become a catch-all that's harder and harder to use & follow.

Adding the `text`: keyword argument and removing the `text = form_data.send(attr)`  makes the `attr` argument superfluous except to define bounding boxes, and the added `fill_text` variable & presence check in addition to the `nil?` guard clauses feels more like changes were made more as a shoehorn to work, rather than following the intent of the method.

Some of this is my own fault when asking for changes in other PRs without recognizing some of the usage in other places that could have been affected. This method is feeling the brunt of that, for sure. 

Let's pair up on it and get it over the finish line :)

cilestin (changes requested):
Before we pair, let's remove NonVeteranClaimant as its own PR so we have room to focus on some refactoring.

modules/appeals_api/app/services/appeals_api/pdf_construction/higher_level_review/v2/structure.rb
ElleeB: yeah. it's wasn't fun to rework, and it's not pretty :( I'll work on it this morning and then let's jump on a call.

ElleeB: I would argue though that the fact that `attr` pulls in the coordinates, and in a way that maps logically and uses naming to make clear what each field is, makes it absolutely essential.

modules/appeals_api/app/services/appeals_api/pdf_construction/higher_level_review/v2/structure.rb
```diff
@@ -259,14 +279,16 @@ def whiteout(pdf, at:, width:, height: 15)
           pdf.fill_color '000000'
         end
 
-        def whiteout_line(pdf, attr, text_override: nil, length_for_override: SHORT_EMAIL_LENGTH)
```
ElleeB, line 262: renamed so it reads when called: `fill_text pdf, :veteran_first_name`

ElleeB: and then in the `fill_text` method we call `whiteout`

ElleeB: makes more sense I think

ElleeB:
> Before we pair, let's remove NonVeteranClaimant as its own PR so we have room to focus on some refactoring.

Had to remove the non veteran claimant spec - it was failing due to updated json fixtures

cilestin (approved):
Lookin' good! Thanks for all the rounds of change requests :)

PR #9046, opened 2022-02-07
2327 remove duplicate vaccines

jperk51:
## Description of change
Catching and removing duplicate vaccine records. Right now the logic just checks the existing list for duplicate datetimes and vaccine names and only adds the first. This seems to be what Apple Health is doing as there aren't any obvious identifiers for why the removed one was removed. The linked ticket contains json objects from the health API which are nearly identical. 

## Original issue(s)
department-of-veterans-affairs/va-mobile-app#2327

Added a duplicate record to v1 immunizations cassette. This way if the tests still passed as the were we know it's filtering that record correctly.


modules/mobile/app/models/mobile/v0/adapters/immunizations.rb
```diff
@@ -6,31 +6,39 @@ module Adapters
       class Immunizations
         def parse(immunizations)
           vaccines = vaccines(immunizations)
-
+          date_text_list = []
```
kpethtel, line 9: I think this could be done a bit more simply without creating the date_text_list. 
```
immunizations[:entry].each_with_object([]) do |i, acc|
  cvx_code = cvx_code(immunization[:vaccine_code])
  vaccine = vaccines&.find_by(cvx_code: cvx_code)
  date = date(immunization)
  text = immunization[:vaccine_code][:text]

  next if acc.any? {|immunization| immunization.date == date && immunization.short_description == text }
  
  # create immunization record
  
  acc << immunization
end
```

kpethtel (changes requested):
It looks like your general approach should work, but I think we need a test case for it.

modules/mobile/app/models/mobile/v0/adapters/immunizations.rb
jperk51: Interesting, yeah, that's a good idea. I'll give that a try, or possibly something simpler

modules/mobile/spec/request/v1/immunizations_request_spec.rb
```diff
@@ -286,39 +286,39 @@
...
         end
 
         it 'does not use cache when useCache is false' do
           expect_any_instance_of(Mobile::V0::LighthouseHealth::Service).to receive(:get_immunizations)
 
-          VCR.use_cassette('lighthouse_health/get_immunizations', match_requests_on: %i[method uri]) do
+          VCR.use_cassette('lighthouse_health/get_immunizations_v1', match_requests_on: %i[method uri]) do
             get '/mobile/v1/health/immunizations', headers: iam_headers, params: { useCache: false }
           end
         end
```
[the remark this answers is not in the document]
jperk51: I added a duplicate to the cassette so the original test case for success now also tests that by ensuring there are 15 entries and not 16. I can add a more deliberate test if we think that's worth.

modules/mobile/app/models/mobile/v0/adapters/immunizations.rb
jperk51: End up using array.uniq. Seemed like the simplest solution, must have looked over it when I was looking at the array methods.

modules/mobile/spec/request/v1/immunizations_request_spec.rb
```diff
@@ -28,7 +28,7 @@
   describe 'GET /mobile/v1/health/immunizations' do
     context 'when the expected fields have data' do
       before do
-        VCR.use_cassette('lighthouse_health/get_immunizations', match_requests_on: %i[method uri]) do
```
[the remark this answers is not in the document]
jperk51: Removed versioned cassette and added duplicate to the original.

aherzberg:
We'll get some conflicts with https://github.com/department-of-veterans-affairs/vets-api/pull/9052. When this gets merged in I'll resolve the conflict on my side.

modules/mobile/spec/request/v1/immunizations_request_spec.rb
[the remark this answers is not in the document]
kpethtel: No, that makes sense. I didn't look closely enough.

kpethtel (approved):
Looks good. 👍

modules/mobile/app/models/mobile/v0/adapters/immunizations.rb
```diff
@@ -24,9 +22,10 @@ def parse(immunizations)
               manufacturer: vaccine&.manufacturer,
               note: note(immunization[:note]),
               reaction: reaction(immunization[:reaction]),
-              short_description: vaccine_code[:text]
+              short_description: immunization[:vaccine_code][:text]
             )
           end
+          vaccine_map.uniq { |immunization| [immunization[:date], immunization[:short_description]] }
```
kreek, line 28: Does this keep the most recent duplicate?

kreek: I guess they're the same if the date matches, does date include time? Can an entry be on the same day but the correction a couple minutes apart?

jperk51: The date does include time. No idea about the corrections. Raf's examples show the exact same time for both

kreek: Ok, I think is the most conservative de-duping either way. I wasn't sure if the date was time received or time entered. Sounds like entered based on Raf's data.

kreek (approved):
LGTM 🚀

PR #9047, opened 2022-02-07
Rapid Ready for Decision pilot: Update email alerts to use recipients from Settings config

julialeague:
## Description of change
Changing the error and tracking alerts for the RRD pilot to store recipient email addresses in the settings file.

## Original issue(s)

There are additions to the `settings.yml` file adding recipients for our error and tracking alerts. They don't currently vary by environment, but they may in the future. It's best that we store the actual email addresses in the private devops repo, and that is the main impetus for this change. Environments where the email recipients aren't set should still work as long as the update to `settings.yml` here is present.

Update to the settings files for staging and prod here: https://github.com/department-of-veterans-affairs/devops/pull/10685


NatashaJIbrahim:
This is really great. A much better solution until we're able to get rid of these emails all together.

NatashaJIbrahim (approved):

yoomlam (approved):

thilton-oddball (approved):

PR #9055, opened 2022-02-07
Cancel appointment makes a call to new_appointment, should call updated_appointment

ajmagdub:
## Description of change

the V2 appointments controller update method was calling new_appointment, changed the new_appointment call to updated_appointment.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#36521


grossb1 (approved):
LGTM

PR #9059, opened 2022-02-08
[33352] sets force_authn to false in all environments

bramleyjl:
## Description of change

As part of the Login.gov rollout we are changing the vets-api SAML headers to ISAM to set `force_authn` to false. The change is happening on a per-environment basis; this PR is the last and will add the changes to production/make them universal to all environments and SAML calls made.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#33352

This should not be merged until a date coordinated with IAM in order to prevent an outage of service on production. The release is planned for 3/10.


bosawt (approved):
Looks good!

PR #9068, opened 2022-02-08
2364 pending appointments

kpethtel:
This is a draft PR for: 
https://app.zenhub.com/workspaces/va-mobile---backend-61d75d03a7505500147ec7f6/issues/department-of-veterans-affairs/va-mobile-app/2440

Not yet included:
- update schema
- write adapter specs


modules/mobile/app/controllers/mobile/v0/appointments_controller.rb
```diff
@@ -74,18 +75,26 @@ def fetch_cached_or_service(validated_params)
...
         appointments.reverse! if validated_params[:reverse_sort]
+
         appointments
       end
 
       def paginate(appointments, validated_params)
         appointments = appointments.filter do |appointment|
-          appointment.start_date_utc.between? validated_params[:start_date], validated_params[:end_date]
+          appointment.start_date_utc.between?(
+            validated_params[:start_date].beginning_of_day, validated_params[:end_date].end_of_day
```
kpethtel, line 94: I changed these because at one point I was testing with a very limited date range and experiencing issues.

modules/mobile/app/models/mobile/v0/adapters/appointment_requests.rb
```diff
@@ -0,0 +1,208 @@
...
+            minutes_duration: nil,
+            phone_only: phone_only?(request),
+            start_date_local: start_date,
+            start_date_utc: start_date.utc,
+            status: status(request),
+            status_detail: nil,
+            time_zone: time_zone(request),
+            vetext_id: nil,
+            reason: request[:purpose_of_visit],
+            is_covid_vaccine: nil, # unable to find or create test data for this
```
kpethtel, line 62: None of the fixture data contains covid vaccinations. I tried creating covid vaccination appointment requests in the web app but they never appeared in the list data. I tried adding a "useCache: false" param, but it didn't change anything. I checked again the next day but again nothing changed. We almost surely can know this from "visit type" but I need to figure out how to get better data.

modules/mobile/app/models/mobile/v0/adapters/appointment_requests.rb
```diff
@@ -0,0 +1,208 @@
...
+          hour = request[:option_time1] == 'AM' ? 8 : 12
+
+          DateTime.new(year, month, day, hour, 0).in_time_zone(time_zone)
+        end
+
+        def status(request)
+          request[:status].upcase
+        end
+
+        # this is not correct for cc appointment requests.
```
kpethtel, line 103: For the time being, I'm just using the facility data to determine time zone for all appointments. That should be mostly correct and is unlikely to be off by more than an hour. We could try finding a ruby library that converts zip codes to timezones.

modules/mobile/app/models/mobile/v0/adapters/va_appointments.rb
```diff
@@ -65,25 +65,18 @@ class VAAppointments
         # @return Hash the adapted list
         #
         def parse(appointments)
-          facilities = Set.new
```
kpethtel, line 68: I moved the facility handling to the proxy to at least keep all of that logic in one place.

modules/mobile/app/services/mobile/v0/appointments/proxy.rb
```diff
@@ -63,15 +49,83 @@ def put_cancel_appointment(params)
...
-          return [] if appointments.nil?
+        def normalize_appointments(responses, start_date, end_date)
+          va_appointments = va_appointments_adapter.parse(responses[:va][:response].body)
+          cc_appointments = cc_appointments_adapter.parse(responses[:cc][:response].body)
+          va_appointment_requests, cc_appointment_requests =
+            requests_adapter.parse(responses[:requests][:response].body[:appointment_requests])
+
+          # There's currently a bug in the underlying Community Care service
+          # where date ranges are not being respected
+          cc_appointments.select! do |appointment|
```
kpethtel, line 60: This should probably be done in the adapter IMO. I just haven't changed it yet.

modules/mobile/app/services/mobile/v0/appointments/appointment_requests_service.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module Mobile
+  module V0
+    module Appointments
+      # connects to appointment request service
+      class AppointmentRequestsService < VAOS::SessionService
+        # fetches the past 90 days of appointment requests
```
kpethtel, line 8: We always fetch 90 days. Due to the way we handle caching in the controller, we always request records from at least a year ago through at least a year from now, so the past 90 days are always required.

modules/mobile/app/services/mobile/v0/appointments/appointment_requests_service.rb
```diff
@@ -0,0 +1,37 @@
...
+        def get_appointment_requests
+          response = nil
+          error = nil
+          end_date = Time.zone.today
+          start_date = end_date - 90.days
+          date_params = { startDate: date_format(start_date), endDate: date_format(end_date) }
+
+          with_monitoring do
```
kpethtel, line 16: This is directly cribbed from the vaos code.

modules/mobile/app/models/mobile/v0/adapters/appointment_requests.rb
jperk51: Yeah, that's about as good as we can get right now. Gets even more complicated when the parent facility and the clinic don't share a timezone. We have a blocked ticket that's wrapped up in all this stuff. Waiting on info from Lighthouse that might help us figure out how to proceed.

modules/mobile/app/controllers/mobile/v0/appointments_controller.rb
```diff
@@ -74,18 +75,26 @@ def fetch_cached_or_service(validated_params)
             start_date: [validated_params[:start_date], one_year_ago].min,
             end_date: [validated_params[:end_date], one_year_from_now].max
           )
+
           Mobile::V0::Appointment.set_cached(@current_user, appointments)
           Rails.logger.info('mobile appointments service fetch', user_uuid: @current_user.uuid)
         end
 
+        unless validated_params[:included]&.include?('pending')
```
kpethtel, line 83: Fetching the requests is feature flagged in the proxy, but this should help the front end devs to test on deployed environments. Once the feature flag is turned on, we need to always fetch the appointment requests to maintain cache integrity.

modules/mobile/app/models/mobile/v0/adapters/appointment_requests.rb
kreek: I don't think zip codes are enough to tell you which time zone you're in. Lat Long can and we could pull down all the clinics and create a lookup table using the [Google maps API](https://developers.google.com/maps/documentation/timezone/get-started).

modules/mobile/app/models/mobile/v0/adapters/appointment_requests.rb
kpethtel: judy

modules/mobile/app/models/mobile/v0/appointment.rb
```diff
@@ -20,9 +20,11 @@ class Appointment < Common::Resource
         'VA',
         'VA_VIDEO_CONNECT_ATLAS',
         'VA_VIDEO_CONNECT_GFE',
-        'VA_VIDEO_CONNECT_HOME'
+        'VA_VIDEO_CONNECT_HOME',
+        'COMMUNITY_CARE_REQUEST',
+        'VA_REQUEST'
```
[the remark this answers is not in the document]
kreek: I know it adds a field but I think an appointment's pending-ness is a `state` rather than a `type`.

modules/mobile/app/services/mobile/v0/appointments/appointment_requests_service.rb
kreek: Do we need our own service because of the date range requirements on mobile? We have our own service for the appointments list because we parallelize it. For canceling appointments, we re-use their service. It's cool if we really do need our own service, but we should lean towards reusing services as much as possible.

modules/mobile/app/services/mobile/v0/appointments/appointment_requests_service.rb
kreek: This adds some metrics automatically behind the scenes. Not sure that we need it as long as we [register our service with the breakers gem](https://github.com/department-of-veterans-affairs/vets-api/blob/1ce184eae8c63b9ff8142ad1ba8c6350b1e2e5cd/config/initializers/breakers.rb#L80). That enables metrics automatically for the service and it shows up in Datadog dashboards (assuming we keep the service rather than re-use VAOS's).

kreek (commented):
I think this is on the right track with adding fields rather than having a separate model 👍

modules/mobile/app/services/mobile/v0/appointments/appointment_requests_service.rb
kpethtel: I'll look into that. I'd also prefer to re-use code.

modules/mobile/app/services/mobile/v0/appointments/appointment_requests_service.rb
kpethtel: I'll look into that. I'd also prefer to re-use code.

modules/mobile/app/models/mobile/v0/appointment.rb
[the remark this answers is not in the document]
kpethtel: That's an easy change to make. I can try to just use the existing appointment types. From the data I've been able to access from the web app, it's unclear if a patient can request a phone or video CC appointment. I can't find any examples of it and it's not clear to me whether you can even create CC requests in the web app (i haven't been able to yet). Any idea if express visits mean video?

modules/mobile/app/models/mobile/v0/adapters/appointment_requests.rb
kpethtel: Sounds like this is good enough for now then, because that definitely sounds like a separate ticket.

modules/mobile/app/models/mobile/v0/adapters/appointment_requests.rb
kpethtel: Interesting. So when you make a covid vaccine request, it bypasses pending and becomes an actual appointment?

modules/mobile/app/services/mobile/v0/appointments/appointment_requests_service.rb
kpethtel: I've done some playing around with it, and I believe I can use their service. It's a little funky because their service handles errors differently, but I think it should work. Thanks for the suggestion.

modules/mobile/app/models/mobile/v0/adapters/appointment_requests.rb
```diff
@@ -0,0 +1,238 @@
...
+            type_of_care: request[:appointment_type],
+            patient_phone_number: request[:phone_number],
+            patient_email: request[:email],
+            best_time_to_call: request[:best_timeto_call]
+          )
+        end
+        # rubcop:enable Metrics/MethodLength
+
+        def appointment_type(request, klass)
+          # this is temporary because test data does not include video type
```
kpethtel, line 66: I will work with Ashley to figure out who to talk to to get a more exhaustive list of options. The one we really care about is video, which is not in the available test data.

modules/mobile/app/models/mobile/v0/appointment.rb
```diff
@@ -22,11 +22,12 @@ class Appointment < Common::Resource
         'VA_VIDEO_CONNECT_GFE',
         'VA_VIDEO_CONNECT_HOME'
       )
-      STATUS_TYPE = Types::String.enum('BOOKED', 'CANCELLED', 'HIDDEN')
+      STATUS_TYPE = Types::String.enum('BOOKED', 'CANCELLED', 'HIDDEN', 'SUBMITTED')
```
kpethtel, line 25: Leaving this status as "submitted" to avoid yet another translation from api to api. It doesn't make a lot of sense by itself, but I think it makes sense when paired with is_pending.

modules/mobile/app/models/mobile/v0/adapters/appointment_requests.rb
```diff
@@ -0,0 +1,239 @@
...
+            vetext_id: nil,
+            reason: request[:purpose_of_visit],
+            is_covid_vaccine: nil,
+            is_pending: true,
+            proposed_times: proposed_times(request),
+            type_of_care: request[:appointment_type],
+            patient_phone_number: request[:phone_number],
+            patient_email: request[:email],
+            best_time_to_call: request[:best_timeto_call],
+            friendly_location_name: request[:friendly_location_name]
```
kpethtel, line 61: We want both the facility name and the friendly name. I didn't add this to the location data because that's used in other places. It's also not embedded in `location` in the source data.

PR #9074, opened 2022-02-09
API-12654 Add new v2 methods to nod model

ElleeB:
## Description of change
Adds handling of extension request, extension reason, and whether or not the Veteran is appealing a VHA claim.

Fixtures added for v2 minimum, standard, and extra.

Tests added for v2 methods in model spec file.

## Original issue(s)
[Ticket](https://vajira.max.gov/browse/API-12654)


modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
```diff
@@ -94,6 +94,10 @@ def email
       veteran_contact_info['emailAddressText']
     end
 
+    def email_v2
```
cilestin, line 97: I think i'd prefer if we update the `email` method to:
```ruby
veteran_contact_info['email'].presence || veteran_contact_info['emailAddressText']
```

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
```diff
@@ -164,6 +168,18 @@ def update_status!(status:, code: nil, detail: nil)
       email_handler.handle! if status == 'submitted' && email_identifier.present?
     end
 
+    def extension_request?
+      form_data&.dig('data', 'attributes', 'extensionRequest')
```
cilestin, line 172: We already have `data_attributes` to get to the form_data's attributes :)

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
```diff
@@ -177,6 +193,7 @@ def mpi_veteran
 
     def email_identifier
       return { id_type: 'email', id_value: email } if email.present?
+      return { id_type: 'email', id_value: email_v2 } if email_v2.present?
```
cilestin, line 196: Updating `email` as suggested above would get rid of the need for this check, too.

cilestin (changes requested):
Some small things to change, but we should also wait for Nathan's fixture refactor PR to get merged before adding more files that'll just need to be moved anyway.

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
ElleeB: V1 form data will expect `email` and V2 will expect `email_v2`. The attribute name has changed from `emailAddressText` to `email`. This is how it is handled throughout. It will change once we upgrade our email handling.

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
ElleeB: 😑

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
ElleeB: oh. derp. just saw your message above.

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
ElleeB: agreed

cilestin (approved):

PR #9085, opened 2022-02-09
36610 Fix check_in statsd metric

randomsync:
## Description of change

Fix for the parameter that is retrieved from session parameter, in case the session parameter is nil

## Original issue(s)
department-of-veterans-affairs/va.gov-team#36610


kanchanasuriya (approved):
LGTM

PR #9097, opened 2022-02-10
2066 fix whitespace removal for message drafts

jperk51:
## Description of change
When a user saves a draft message all whitespace should be preserved. This is especially apparent when a user creates a draft that just has their signature. Currently, the message is returned with the proceeding whitespace (4 returns) removed by the MessageDraft model. This PR preserves the original draft body. 

## Original issue(s)
department-of-veterans-affairs/va-mobile-app#2066

Tests were added to check the whitespace is preserved. Making the fix in the client should fix this issue for both mobile and web. If it would be safer to just fix it for mobile I can move the fix into the mobile module.


lib/sm/client.rb
```diff
@@ -163,7 +163,9 @@ def post_create_message_draft(args = {})
       validate_draft(args)
 
       json = perform(:post, 'message/draft', args, token_headers).body
-      MessageDraft.new(json)
+      draft = MessageDraft.new(json)
+      draft.body = draft.instance_variable_get(:@original_attributes)[:body]
```
kreek, line 167: I think we could expose the body through a method or an attr_reader. That would avoid using `instance_variable_get` which is usually reserved for working around access to attributes in code you can't modify (e.g. inside a gem).

lib/sm/client.rb
```diff
@@ -181,7 +183,9 @@ def post_create_message_draft_reply(id, args = {})
       json = perform(:post, "message/#{id}/replydraft", args, token_headers).body
       json[:data][:has_message] = true
 
-      MessageDraft.new(json).as_reply
+      draft = MessageDraft.new(json)
+      draft.body = draft.instance_variable_get(:@original_attributes)[:body]
```
kreek, line 187: Same here.

lib/sm/client.rb
jperk51: gotcha, will do

modules/mobile/spec/request/message_drafts_request_spec.rb
```diff
@@ -15,6 +15,7 @@
   let(:draft) { attributes_for(:message, body: 'Body 1', subject: 'Subject 1') }
   let(:params) { draft.slice(:category, :subject, :body, :recipient_id) }
   let(:va_patient) { true }
+  let(:draft_signature_only) { attributes_for(:message, body: '\n\n\n\nSignature\nExample', subject: 'Subject 1') }
```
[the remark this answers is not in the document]
jperk51: That's good to know. As is it's a little easier to read and it's more accurate to what exactly we get from the front

aherzberg (approved):
Looks good 👍

thilton-oddball (approved):

PR #9102, opened 2022-02-11
API-9833: Bring "get claim by ID" v2 endpoint to parity with v1

hinzed1127:
## Description of change
Adds in missing properties for retrieving a claim by ID in v2

## Original issue(s)
https://vajira.max.gov/browse/API-9833

There are a few discrepancies right now between the v1 claim detail and this endpoint. This stems from 2 different things:
1. The v1 endpoint was calling EVSS' 526 service, while this is calling BGS directly. As such, while there are a lot of similarities in the responses between the 2 services, there are some differences. I wanted to use this as a first pass to tie as much of the properties in the BGS response to what we provide in v1.
2. There's a handful of properties that seem to be pretty strongly tied to the VA.gov claims tooling. The same logic doesn't necessarily need to apply to what we return, so we should have a little more leeway in what we return and how we format it (e.g. the different phases not getting reduced from 8 to 5 different values).

I'm still working on fixing a few tests that got broken by adding in some of the new properties, but the logic in the controller/blueprint itself should be ok.


modules/claims_api/app/services/claims_api/bgs_to_lighthouse_claims_mapper_service.rb
```diff
@@ -33,26 +33,20 @@ def lighthouse_claim_only?
...
     def unmatched_bgs_claim
+      { id: bgs_claim[:benefit_claim_id] }.merge(shared_claim_traits)
+    end
+
+    def shared_claim_traits
       {
-        id: bgs_claim[:benefit_claim_id],
         type: bgs_claim[:claim_status_type],
-        date_filed: bgs_claim[:claim_dt],
+        date_filed: bgs_claim[:claim_dt].strftime('%D'),
```
hinzed1127, line 46: Something about this check is tripping up some tests, still working on figuring out why. Trying to include the safe navigation operator like so didn't seem to fix it:
```ruby
date_filed: bgs_claim[:claim_dt]&.strftime('%D'),
```

modules/claims_api/app/services/claims_api/bgs_to_lighthouse_claims_mapper_service.rb
```diff
@@ -33,26 +33,20 @@ def lighthouse_claim_only?
...
     def unmatched_bgs_claim
+      { id: bgs_claim[:benefit_claim_id] }.merge(shared_claim_traits)
+    end
+
+    def shared_claim_traits
       {
-        id: bgs_claim[:benefit_claim_id],
         type: bgs_claim[:claim_status_type],
-        date_filed: bgs_claim[:claim_dt],
+        date_filed: bgs_claim[:claim_dt].present? ? (Date.parse(bgs_claim[:claim_dt])).strftime('%D') : nil,
```
hinzed1127, line 46: I don't think this will ever be `nil`, but it helped fix some tests, and I don't think it hurts to have a fallback case handling this

modules/claims_api/spec/requests/v2/veterans/claims_request_spec.rb
```diff
@@ -114,7 +114,7 @@
                       claim_status_type: 'Compensation',
                       claim_dt: '2017-05-02',
                       phase_type: 'Pending',
-                      attention_needed: true,
+                      attention_needed: 'Yes',
```
hinzed1127, line 117: This is how the response from BGS actual looks ('Yes' or 'No'), so reflecting that here

spec/support/schemas/claims_api/v2/veterans/claims/claim.json
```diff
@@ -32,6 +17,78 @@
...
+      "example": ["abnormal heart (New)", "abscess kidney (New)"]
+    },
+    "dateFiled": {
+      "type": "string",
+      "description": "The date a claim was filed",
+      "example": "5/17/20"
+    },
+    "decisionLetterSent": {
+      "type": "boolean",
+      "description": "Indicates if a Development Letter has been sent to the Claimant regarding a benefit claim"
```
jw81, line 33: Probably mean Decision Letter here.

spec/support/schemas/claims_api/v2/veterans/claims/claim.json
```diff
@@ -32,6 +17,78 @@
...
+    },
+    "endProductCode": {
+      "type": "string",
+      "description": "End product code of claim",
+      "example": "930"
+    },
+    "requestedDecision": {
+      "type": "boolean",
+      "nullable": true,
+      "description": "Indicates if a decision has been request and/or Waiver 5103 submitted"
```
jw81, line 51: Probably mean "requested" here.

spec/support/schemas/claims_api/v2/veterans/claims/claim.json
```diff
@@ -32,6 +17,78 @@
...
+      }
+    },
+    "type": {
+      "type": "string",
+      "description": "Name of claim type",
+      "example": "Appeals Control"
+    },
+    "vaRepresentative": {
+      "type": "string",
+      "description": "The current organization of individual representing the Veteran on a claim",
```
jw81, line 90: Probably mean "The current organization OR individual ....".

spec/support/schemas/claims_api/v2/veterans/claims/claim.json
```diff
@@ -2,28 +2,26 @@
...
+    "dateFiled",
+    "decisionLetterSent",
+    "developmentLetterSent",
+    "documentsNeeded",
+    "endProductCode",
+    "status",
+    "supportingDocuments",
+    "type",
+    "vaRepresentative"
+  ],
```
hinzed1127, line 18: the semantics of "required" are a little odd here, because it's for a GET response. I'm interpreting that as "we guarantee these values will be present in the response," which in this case meant I added everything outside of `requestedDecision`, which could have a null value

hinzed1127: If we're ok with that interpretation, then we should update the [claims list schema required values](https://github.com/department-of-veterans-affairs/vets-api/blob/master/spec/support/schemas/claims_api/v2/veterans/claims/claims.json#L5) in a separate PR

thilton-oddball (dismissed):

spec/support/schemas/claims_api/v2/veterans/claims/claim.json
jw81: It's late in the day and my brain is probably just tired, but I want to make sure that we're not ever optionally including attributes in the response, right?  Like, even if an attribute could have a null/blank value, we're still going to include that attribute even when it's null/blank aren't we?

hinzed1127: > even if an attribute could have a null/blank value, we're still going to include that attribute even when it's null/blank aren't we?

Yeah. [According to the JSON schema docs](https://json-schema.org/understanding-json-schema/reference/object.html?highlight=required#required-properties), it defaults to properties not being required (which I don’t think we need to care too much about anyway in the context of a GET request. this is just how we build out the responses w/ rswag, so it's more just a means to an end):

> By default, the properties defined by the properties keyword are not required. However, one can provide a list of required properties using the required keyword.

hinzed1127: Removed required [here](https://github.com/department-of-veterans-affairs/vets-api/pull/9102/commits/24d5a62175c3ebed7a41a3ee321a7e15569971b8)

jw81: Probably want RSWAG to regenerate the V2 docs after removing the "required" stuff.

thilton-oddball (approved):

jw81 (approved):

PR #9105, opened 2022-02-11
API-12125 HLR v2 fix up claimant requirements and handling

ElleeB:
Fixes issue with our hlr v2 claimant header requirements
Updates the #signing_appellant method in Appellant class


cilestin (approved):
Approved, but let's wait to merge until next week just in case the frontend folks have any questions for us on Monday.

PR #9110, opened 2022-02-14
RRD PDF: Create an “empty state” for medications heading and query information

allthesignals:
## Description of change
Addresses [this ticket](https://www.notion.so/64997f0e6c4c40df816b7c4b305d18d4?v=fe1d45ccd3014f6484a2f70842c3df19&p=05540a5211f34aa68adda31e270ab003). Ensures "Active medications" header is always present and, if no meds are present, shows message.

image

## Original issue(s)
https://www.notion.so/64997f0e6c4c40df816b7c4b305d18d4?v=fe1d45ccd3014f6484a2f70842c3df19&p=05540a5211f34aa68adda31e270ab003

In the spec, it generates the RRD PDF with different arguments.


spec/services/rapid_ready_for_decision/hypertension_pdf_generator_spec.rb
```diff
@@ -33,10 +33,18 @@
     RapidReadyForDecision::HypertensionPdfGenerator.new(patient_name, parsed_bp_data, parsed_medications_data)
   end
 
+  let(:pdf_generator_no_meds) do
+    RapidReadyForDecision::HypertensionPdfGenerator.new(patient_name, parsed_bp_data, [])
+  end
+
```
allthesignals, line 39: @nanotone i doubt this is the best way to do this but couldn't infer from the way it was setup

nanotone: Okay yeah, I see how this is set up. My suggestion is to add a new

```ruby
context 'when no medications are present'
```

block _inside_ the `describe '#generate'` block, and within that context do the override

```ruby
let(:parsed_medications_data) { [] }
```

That way you can reuse the existing `pdf` variable within your two new `it` blocks.

(I'm not that familiar with VCR magic yet, but I think it should work although the cassette might need a little rejiggering.)

nanotone: Also, in the spirit of leaving things nicer than they were, I don't think the `pdf` variable needs to be `let!`. It's referenced in every `it` block, and should really just be a `subject { }`.

allthesignals: Nice! Thanks, much cleaner. I think VCR functions at the HTTP request step so should be okay but will regenerate just to make sure nothing weird is going on.

allthesignals: Works beautifully. I'm guessing `context` here re-runs 

> Also, in the spirit of leaving things nicer than they were, I don't think the `pdf` variable needs to be `let!`. It's referenced in every `it` block, and should really just be a `subject { }`.

Just saw this, will clean that up.

app/services/rapid_ready_for_decision/hypertension_pdf_generator.rb
```diff
@@ -123,15 +123,17 @@ def add_medications_intro
...
+          @pdf.text medication['description'], size: 11, style: :bold
+          @pdf.text "Prescribed on: #{medication['authoredOn'][0, 10].to_date.strftime('%m/%d/%Y')}"
+          if medication['dosageInstructions'].present?
+            @pdf.text "Dosage instructions: #{medication['dosageInstructions'].join('; ')}"
+          end
+          @pdf.text "\n", size: 8
         end
-        @pdf.text "\n", size: 8
+      else
+        @pdf.text 'No active medications were found in the last year', size: 8
```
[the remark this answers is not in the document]
allthesignals: Note that `return` no longer returns the `@pdf` because these methods should not be returning anything anyway (most do not).

nanotone (approved):
Splendid!

thilton-oddball (approved):

PR #9112, opened 2022-02-14
API-13230: Align validation endpoints with submission validations for claims v1

hinzed1127:
## Description of change

This refactors the `validate`endpoints for Claims v1 services to use the same logic as the validations present in the actual submission endpoints. This didn't require deleting or modifying any of the existing validations, simply adding in validations that had been tacked on to the submission endpoints, but not the validate endpoints (probably because the `validate` endpoints were deprecated, but because some consumers are still using them, we should maintain parity for the time being).

## Original issue(s)
https://vajira.max.gov/browse/API-13230


jw81 (approved):

jw81:
@hinzed1127 Kayla said that we should hold onto this code until we have approved Release Notes that are also ready to go with the code.

PR #9114, opened 2022-02-15
2500 mobile payment history date filter

cadibemma:
## Description of change
The UX flow for payment history is to first show the most recent payments within one calendar year. e.g. If the most recent payments are from 2020, then the first set payments returned should only be for 2020. We should still paginate them 10 at a time, but pagination would end in 2020. It would not roll over into 2021 or 2019. The date parameter (year may be a better name) will not be like appointments with a start and end date. It will have one value like 2020 or 2017.

## Original issue(s)
department-of-veterans-affairs/va-mobile-app#2500


modules/mobile/app/controllers/mobile/v0/payment_history_controller.rb
```diff
@@ -43,8 +42,26 @@ def bgs_service_response
       end
 
       def paginate(payments, validated_params)
+        payment_years = Set[]
```
kpethtel, line 45: I'm not sure I understand the purpose of initializing as a set only to convert to an array. Why not just start with an array?

modules/mobile/spec/request/payment_history_request_spec.rb
```diff
@@ -115,5 +140,41 @@
         )
       end
     end
+
+    context 'with a valid date params' do
```
kpethtel, line 144: We could drop the word "a" here.

modules/mobile/spec/request/payment_history_request_spec.rb
```diff
@@ -115,5 +140,41 @@
...
+
+      it 'returns a 200' do
+        expect(response).to have_http_status(:ok)
+      end
+
+      it 'matches expected schema' do
+        expect(response.body).to match_json_schema('payment_history')
+      end
+
+      it 'only paginates and returns payments for that year' do
```
kpethtel, line 166: Is it really about the year or is it about being in the requested date range?

kpethtel (approved):
Looks good. 👍

modules/mobile/app/controllers/mobile/v0/payment_history_controller.rb
kpethtel: I see. Outside of our codebase, I'm not used to seeing set a lot. Thanks for explaining.

kreek (changes requested):
It's not in the changes for this PR but I think we need to update how we're setting default date ranges. Currently, we're using the appointments way which is today +- one year. But it should be the start and end of the first year available (which also means we need to know that first before we set dates if they're not in the params).

kreek (approved):
LGTM 🚀 but I have some commits on the branch so please also wait for @kpethtel to take a look again 😄

modules/mobile/app/controllers/mobile/v0/payment_history_controller.rb
kreek: We ended up changing it to `uniq` anyway because we now need to know the dates before we filter so we don't have that loop.

modules/mobile/app/controllers/mobile/v0/payment_history_controller.rb
```diff
@@ -43,8 +39,26 @@ def bgs_service_response
       end
 
       def paginate(payments, validated_params)
+        available_years = payments.map { |p| p.date.year }.uniq.sort { |a, b| b <=> a }
+        most_recent_year = available_years.first
+        start_date = validated_params[:start_date]
+        end_date = validated_params[:end_date]
+
+        unless start_date && end_date
+          start_date =  DateTime.new(most_recent_year).beginning_of_year.utc.iso8601
```
kpethtel: This has an extra space in it.

modules/mobile/app/controllers/mobile/v0/payment_history_controller.rb
```diff
@@ -43,8 +39,26 @@ def bgs_service_response
       end
 
       def paginate(payments, validated_params)
+        available_years = payments.map { |p| p.date.year }.uniq.sort { |a, b| b <=> a }
+        start_date = validated_params[:start_date]
+        end_date = validated_params[:end_date]
+
+        unless start_date && end_date
+          most_recent_year = available_years.first
```
kpethtel, line 47: Doesn't have any impact, but I think you could sort in ascending order above (i.e., without a block) and do .last here.

modules/mobile/spec/request/payment_history_request_spec.rb
```diff
@@ -48,11 +48,36 @@
...
+      it 'returns meta data that includes the years in which the user has payments' do
+        expect(response.parsed_body['meta']).to include(
+          {
+            'availableYears' => [2019, 2018, 2017, 2016, 2015]
+          }
+        )
+      end
+
+      it 'only paginates and returns the number of records for the latest year' do
+        expect(response.parsed_body['meta']).to include(
```
kpethtel, line 61: I think this test would be more valuable if we did something like map all the ids or dates from the returned payment data and confirm that they're the correct ones. Otherwise, we're just going by "i expect 7 entries and i got 7" but there's nothing to indicate that they're the correct 7. The same comment applies below.

kpethtel (approved):
👍

PR #9116, opened 2022-02-15
API-12125 Fixup appellant and hlr class methods

ElleeB:
Clean up appellant class methods and hlr class methods, especially beef up specs.


ElleeB:
Strange files sticking around - just going to push up fresh branch.

PR #9118, opened 2022-02-15
RRD Hypertension PDF: Display effective date instead of issued date

allthesignals:
## Description of change
Displays `effectiveDateTime` in the Hypertension PDF instead of `issued`. This required a change to the HypertensionObservationData to provide the `effectiveDateTime` for the PDF generator.

## Original issue(s)
Addresses https://www.notion.so/64997f0e6c4c40df816b7c4b305d18d4?v=fe1d45ccd3014f6484a2f70842c3df19&p=b82a61b2e11544f18e8b7f361e9022d5

Some concerns:

 - This is still set up to filter for readings from the last year using the `issued` date. Should this matter? From what I understand, the `issued` date and `effectiveDateTime` can differ by a significant amount of time. Should I update that as well? Will ask researchers...
 - Is `effectiveDateTime` ever an optional field? If it comes back null from Lighthouse, it will break PDF generation.
 - Testing/refactoring note: Not sure what the best testing strategy is. There's a bit of "inappropriate intimacy" going on between HypertensionObservationData class and HypertensionPdfGenerator class. Some possible refactor work there...
   - There was some mocked data in DisabilityCompensationJob spec that caused the tests there to fail. I updated that mocked data.

image


app/services/rapid_ready_for_decision/hypertension_observation_data.rb
```diff
@@ -30,12 +31,13 @@ def filtered_entries(bp_readings)
...
     def transform_entry(raw_entry)
-      entry = raw_entry['resource'].slice('issued', 'component', 'performer')
+      entry = raw_entry['resource'].slice('issued', 'component', 'performer', 'effectiveDateTime')
       practitioner_hash = get_display_hash_from_performer('Practitioner', entry)
       organization_hash = get_display_hash_from_performer('Organization', entry)
       bp_hash = get_bp_readings_from_entry(entry)
 
-      { issued: entry['issued'] }.merge(practitioner_hash, organization_hash, bp_hash)
+      { issued: entry['issued'], effectiveDateTime: entry['effectiveDateTime'] }.merge(practitioner_hash,
+                                                                                       organization_hash, bp_hash)
```
allthesignals, line 40: Maybe this could be refactored to be configurable through a constant if more data is needed:

```ruby
KEYS = [''issued', 'component', 'performer', 'effectiveDateTime']
...
entry = raw_entry['resource'].slice(KEYS)
```

Then we could tidy up this object literal and merge...

spec/services/rapid_ready_for_decision/hypertension_observation_data_spec.rb
```diff
@@ -16,6 +16,7 @@
   let(:response_with_recent_bp) do
     original_first_bp_reading = response.body['entry'].first
     original_first_bp_reading['resource']['issued'] = (Time.zone.today - 2.weeks).to_s
+    original_first_bp_reading['resource']['effectiveDateTime'] = (Time.zone.today - 2.weeks).to_s
```
allthesignals, line 19: I... don't love this but it gets the job done. I know in some frameworks there's a more robust system for generating mock data, and I'm sure that exists in Rails land somewhere.

tadhg-ohiggins:
Do we want to remove references to `issued` entirely, if we're not going to use that field?

allthesignals:
@tadhg-ohiggins 

> Do we want to remove references to `issued` entirely, if we're not going to use that field?

Yeah! I should remove that stuff from the HypertensionPdfGenerator spec... but separately, and maybe this is obvious, the HypertensionObservationData class _still_ uses `issued` to deal with filtering and sorting. I'm not sure what the consequences for changing that might be or if I'm overthinking that.

tadhg-ohiggins:
I think if we're displaying `effectiveDateTime` we should use it in all the places we were using `issued`.

PR #9119, opened 2022-02-15
API-12125 Fixup appellant and HLR class

ElleeB:
Fixup items discovered after having stepped away for a minute.

References: https://vajira.max.gov/browse/API-12125 | https://vajira.max.gov/browse/API-13104

**Questions still on the table**: Do we want to white out and print 'USE PHONE ON FILE' or 'USE EMAIL ON FILE' ?

V1 was easy since the fields are text boxes, while V2 is individual cels per character, so we'd need to do some extra whiting-out and auto fill etc.


modules/appeals_api/app/models/appeals_api/appellant.rb
```diff
@@ -101,18 +101,26 @@ def phone_string
     end
 
     def area_code
+      return unless domestic_phone?
```
cilestin, line 104: These are still valid parts of international number. We shouldn't be gating them to only domestic numbers.

modules/appeals_api/app/models/appeals_api/higher_level_review.rb
```diff
@@ -153,11 +162,7 @@ def veteran_homeless?
     end
 
     def email
-      veteran_data&.dig('emailAddressText').to_s.strip
-    end
-
-    def email_v2
-      veteran_data&.dig('email').to_s.strip
+      veteran.email.presence || veteran_data&.dig('emailAddressText').to_s.strip
```
cilestin, line 165: Let's add a comment for why we're doing it this way.

cilestin (changes requested):

modules/appeals_api/app/models/appeals_api/appellant.rb
ElleeB: K! Going to rework this into form_data so the area code doesn't print an extra area code on the form.

cilestin (approved):

PR #9121, opened 2022-02-15
2440: modify appointments, appointments proxy, and adapters in prepar…

kpethtel:
## Description of change
This is all groundwork for adding appointment requests to the appointments index. It adds new fields to the appointments model and adds handling for the new included query param.

For reference, here is the larger diff from which this was pulled due to LoC restrictions: https://github.com/department-of-veterans-affairs/vets-api/pull/9068/files

## Original issue(s)
department-of-veterans-affairs/va-mobile-app#2440

The only impact this should have is adding some new attributes to appointments. Those should not break the data contract. 

Tested on the review branch by curling for judy's appointments. No errors.


modules/mobile/app/models/mobile/v0/appointment.rb
```diff
@@ -22,11 +22,12 @@ class Appointment < Common::Resource
         'VA_VIDEO_CONNECT_GFE',
         'VA_VIDEO_CONNECT_HOME'
       )
-      STATUS_TYPE = Types::String.enum('BOOKED', 'CANCELLED', 'HIDDEN')
+      STATUS_TYPE = Types::String.enum('BOOKED', 'CANCELLED', 'HIDDEN', 'SUBMITTED')
```
kpethtel, line 25: Appointment requests can be either cancelled or submitted. "Submitted" isn't quite as clear as "proposed" but I stuck with the name because I thought it was better to stick with established names in other parts of the system. I think it makes sense when combined with `isPending`.

modules/mobile/app/helpers/mobile/pagination_helper.rb
```diff
@@ -69,6 +69,7 @@ def url_without_page_number
           query_strings << "#{camelized}=#{validated_params[key]}"
         end
 
+        query_strings << 'included[]=pending' if validated_params[:included]&.include?('pending')
```
kreek, line 72: Maybe we could subclass the pagination helper for appointments and override this method? Claims, payments, etc. don't know about 'pending' things :)

kreek: Or `link_url` could be simpler to override, just tack on the include if necessary.

jperk51 (commented):
+1 to Alastair's comment but otherwise it LGTM

modules/mobile/app/helpers/mobile/pagination_helper.rb
kpethtel: I'll remove this part of the PR and deal with it in the next PRs.

kpethtel: Removed.

kreek (approved):
If getting the included param for the links data is going in another PR then this LGTM 🚀

PR #9131, opened 2022-02-16
Allow at home video appointments to be cached when no facility

aherzberg:
## Description of change

Allows At Home video appointments to be cached when facility id given in appointment is null. 

Front end confirmed returned appointment only needs a non-null name to be processed. UX is unchanged as location information is hidden for at home appointment types. wording "No Location provided" was given by UX but will not currently show. Used the given wordage anyways in case this ever changes. 

facility lookup will now only happen if there are facilities given in appointment. 

## Original issue(s)
https://github.com/department-of-veterans-affairs/va-mobile-app/issues/2470

VCR Cassette was created manually as I have been unable to reproduce in staging but was derived with following error logged: `{"host":"12345","application":"vets-api-worker","environment":"production","timestamp":"2022-02-16T15:54:58.320871Z","level":"warn","level_index":3,"pid":10,"thread":"processor","named_tags":{"class":"Mobile::V0::PreCacheAppointmentsJob","jid":"143564ab4848f5f4065d22ca87","request_id":"N/A","remote_ip":"N/A","user_agent":"N/A","user_uuid":"N/A","source":"mobile"},"name":"Rails","message":"Could not find matching facility for mobile appointment","payload":{"appointment_id":"1234567","facility_id":null,"sta6aid":null,"id_for_address":null,"type":"VA_VIDEO_CONNECT_HOME","location":{"id":null,"name":null,"address":{"street":null,"city":null,"state":null,"zip_code":null},"lat":null,"long":null,"phone":{"area_code":null,"number":null,"extension":null},"url":"https://care.va.gov","code":"1234567#"},"status":"BOOKED","healthcare_service":"HAMPTON VA MEDICAL CENTER"}}`


modules/mobile/app/models/mobile/v0/adapters/va_facilities.rb
```diff
@@ -5,23 +5,24 @@ module V0
...
       class VAFacilities
         def map_appointments_to_facilities(appointments, facilities)
-          facilities_by_id = facilities.index_by(&:id)
+          facilities_by_id = facilities&.index_by(&:id)
 
           appointments.map do |appointment|
-            facility = facilities_by_id["vha_#{appointment.id_for_address}"]
-            if facility.nil?
+            facility = facilities_by_id&.dig("vha_#{appointment.id_for_address}")
+            if facility.nil? &&
```
kpethtel, line 12: So this condition seems to be the heart of the ticket. In the case of a video connect appointment without a facility, we still want to have a location name. And location requires an address, which is why we're changing the facilities helper. That's leaving us in an odd place where we have a bunch of conditional logic (see all the safety operators here and in the facilities helper). It might be better to be more explicit in our conditions. 
 ```
if facility
  # create appointment as normal
elsif appointment.appointment_type == VAAppointments::APPOINTMENT_TYPES[:va_video_connect_home]
  # create appointment with name 'No location provided' and a bunch of nils
else
  # raise error
end
 ```

modules/mobile/app/helpers/mobile/facilities_helper.rb
```diff
@@ -18,19 +18,18 @@ def facilities_service
     end
 
     def address_from_facility(facility)
-      address = facility.address['physical']
-      return nil unless address
```
kpethtel, line 22: This nil return actually seems problematic because we only use this function for filling out locations, which require an address object.

kpethtel (commented):
I think you're in a similar place as I am dealing with ambiguous data. I added a suggestion for how we can maybe keep this a bit simpler and more explicit. I'm not sure that my suggestion is any better, but I think it requires less conditional code.

modules/mobile/app/models/mobile/v0/adapters/va_facilities.rb
aherzberg: yes, I agree. I'll update this.

modules/mobile/app/helpers/mobile/facilities_helper.rb
aherzberg: yes, I think this needed to be changed outside this issue. I bet the address has just never been nil before now.

kreek (dismissed):
LGTM 🚀 - reminder that code merged in before 2 EST/11 PST will go out at 3 EST/12 PST. So if you want 24hrs to test on staging merge it after that.

modules/mobile/app/helpers/mobile/facilities_helper.rb
kpethtel: So I tested this in the command line and indeed you can't create a location without an address. So this should be safe to remove.

kpethtel (dismissed):
Nice work.

kpethtel (approved):
👍

modules/mobile/app/services/mobile/v0/appointments/proxy.rb
```diff
@@ -78,7 +78,7 @@ def fetch_facilities(appointments)
           facility_ids.each do |facility_id|
             Rails.logger.info('metric.mobile.appointment.facility', facility_id: facility_id)
           end
-          Mobile::FacilitiesHelper.get_facilities(facility_ids)
+          Mobile::FacilitiesHelper.get_facilities(facility_ids) if facility_ids.any?
```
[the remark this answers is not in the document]
aherzberg: Good point. updated to this use this guard instead. Just wanted to ensure that the facility locator wasn't needlessly called cause it was previously just pinging for facility `vha_`

kpethtel (approved):

PR #9140, opened 2022-02-16
33582 profile bad address

tpharrison:
## Description of change
VA Profile has an indicator in their data source that can notify us if an address has been flagged as bad because mail has been returned.  

This PR adds the bad address indicator (bad_address) to the backend response (/v0/user) so that the frontend can use it.  

## Original issue(s)
department-of-veterans-affairs/va.gov-team/issues/33582

- The bad_address field is read-only.

Tested a few users in the review instance to make sure that the `bad_address` field was included in the `v0/user` response.


spec/lib/va_profile/contact_information/service_spec.rb
```diff
@@ -32,6 +32,15 @@
           expect(response.person.addresses[0].province).to eq('province')
         end
       end
+
+      it 'has a bad address' do
+        VCR.use_cassette('va_profile/contact_information/person_full', VCR::MATCH_EVERYTHING) do
+          response = subject.get_person
+
+          puts response.person.addresses[1].inspect
```
lihanli, line 40: should delete this

lihanli (dismissed):
LGTM except for the debug line

spec/lib/va_profile/contact_information/service_spec.rb
LindseySaari: ➕ Looks like a debug line that was forgotten

LindseySaari (approved):

PR #9144, opened 2022-02-16
2440: add appointment requests adapter

kpethtel:
## Description of change
Adds an adapter for adapting appointment requests into appointments. 

## Original issue(s)
department-of-veterans-affairs/va-mobile-app#2440

This is new code that's not yet in use. Testing will be done in a subsequent PR. Request and unit tests for this will be added in future PRs as well. Adding unit tests here would likely put me over 500 LoC.


modules/mobile/app/models/mobile/v0/adapters/appointment_requests.rb
```diff
@@ -0,0 +1,239 @@
...
+            patient_email: request[:email],
+            best_time_to_call: request[:best_timeto_call],
+            friendly_location_name: request[:friendly_location_name]
+          )
+        end
+        # rubcop:enable Metrics/MethodLength
+
+        def appointment_type(request, klass)
+          # this is temporary because test data does not include video type
+          unless request[:visit_type].in?(['Office Visit', 'Express Care', 'Phone Call'])
```
kpethtel, line 68: We're talking to the VAOS team tomorrow, so hopefully we can clear this up then.

modules/mobile/app/models/mobile/v0/adapters/appointment_requests.rb
```diff
@@ -0,0 +1,239 @@
...
+        def status(request)
+          request[:status].upcase
+        end
+
+        def status_detail(request)
+          return nil unless request[:status] == 'Cancelled'
+
+          first_detail = request.dig(:appointment_request_detail_code, 0)
+          cancellation_code = first_detail&.dig(:detail_code, :code)
+          return nil unless cancellation_code
```
kpethtel, line 119: Depending on how the front end uses this data, it might be better to set this to 'Cancelled - other'.

modules/mobile/app/models/mobile/v0/adapters/appointment_requests.rb
```diff
@@ -0,0 +1,239 @@
...
+
+          case cancellation_code
+          when 'DETCODE22', 'DETCODE8'
+            'CANCELLED BY PATIENT'
+          when 'DETCODE19'
+            'CANCELLED BY CLINIC'
+          when 'DETCODE24'
+            'CANCELLED - OTHER'
+          else
+            log_message_to_sentry('Unknown appointment request cancellation code', :error, { detail: first_detail })
```
kpethtel, line 129: Logging here as a way of finding out what other cases we need to account for.

modules/mobile/app/models/mobile/v0/adapters/appointment_requests.rb
```diff
@@ -0,0 +1,239 @@
...
+                number: phone_captures[2].presence,
+                extension: phone_captures[3].presence
+              },
+              url: nil,
+              code: nil
+            }
+          end
+          # rubocop:enable Metrics/MethodLength
+
+          def self.phone_captures(request)
```
kpethtel, line 230: I cribbed this from one of our other adapters. If we find ourselves using it a third time, we should probably create a helper.

modules/mobile/app/models/mobile/v0/adapters/appointment_requests.rb
```diff
@@ -0,0 +1,237 @@
...
+          klass::APPOINTMENT_TYPE
+        end
+
+        def phone_only?(request)
+          request[:visit_type] == 'Phone Call'
+        end
+
+        def proposed_times(request)
+          {
+            option_date1: request[:option_date1] == 'No Date Selected' ? nil : request[:option_date1],
```
kreek, line 80: I know the VAMF service returns it like this but maybe we could have proposed times be an array of hashes? Is each time a specific time or like 'morning/afternoon', or 'from 2-4pm'?

modules/mobile/app/models/mobile/v0/adapters/appointment_requests.rb
aherzberg: Looks like we do it a third time in the modules/mobile/app/helpers/mobile/facilities_helper.rb:38. Regex is slightly different but I believe it's essentially equivalent. May be worth standardizing it with a helper. Given the scope of all these appointment additions maybe we break it out to a separate issue?

kpethtel: Yeah, good idea.

modules/mobile/app/models/mobile/v0/adapters/appointment_requests.rb
kpethtel: I'm happy to change the format. I'll let Ken know that's coming. All the data I've seen just says "AM" or "PM".

kreek: 👍 I think that future proofs it, and the FE can also just iterate over the list. I've also seen some option_17 option_18 etc in other government APIs 😬

kpethtel: If it's anything other than AM/PM it'll cause issues because we're using proposed times for sorting purposes. But we'll deal with those errors if they arise.

modules/mobile/app/models/mobile/v0/adapters/appointment_requests.rb
kpethtel: https://app.zenhub.com/workspaces/va-mobile---backend-61d75d03a7505500147ec7f6/issues/department-of-veterans-affairs/va-mobile-app/2580

modules/mobile/app/models/mobile/v0/adapters/appointment_requests.rb
kpethtel: Good news, I finally figured out what a video conference type is called. I'll update this in an upcoming PR.

aherzberg (approved):
LGTM

PR #9145, opened 2022-02-16
Unified sign in force ial2 loa3

jscanish:
## Description of change
We only want to allow use of the mobile app to verified users. The frontend is going to send the backend a param in the URL that indicates the authn call should only use verified authns. IAL1/LOA1 accounts should not be able to log in.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/37031

Frontend component of this change is [here](https://github.com/department-of-veterans-affairs/vets-website/pull/20249).

Tested using LOA1 account to ensure user could not sign in. LOA1 user is redirected to "Verify Your Identity" page.


joeniquette (approved):

lib/saml/url_service.rb
```diff
@@ -82,22 +82,43 @@ def mhv_url
       build_sso_url(build_authn_context('myhealthevet'))
     end
 
+    def mhv_verified_url
```
bosawt, line 85: I wonder if this is a good opportunity to reduce the number of these functions, we could pass `type` and `authn_context` from `sessions_controller` and reduce a bunch of these to a single function

bosawt (changes requested):
Can you try testing this real fast? When I test it with `idme_verified` (and I create a new account), I get a 'server error' in the eauth redirect back. This could definitely be something on my end, or something with IAM, but want to make sure this isn't happening as a result of something we're doing.

Otherwise, it does appear that it connects to ID.me and expects an LOA3 account (it forces me to verify instead of accepting an LOA1 account), so besides that 'server error' which might not even be us, it appears to be working!

app/controllers/v1/sessions_controller.rb
```diff
@@ -168,16 +170,28 @@ def login_params(type)
       case type
       when 'mhv'
         url_service.mhv_url
+      when 'mhv_verified'
+        url_service.mhv_verified_url
       when 'dslogon'
         url_service.dslogon_url
+      when 'dslogon_verified'
+        url_service.dslogon_verified_url
```
bosawt, line 178: We'll need specs for all this too

jscanish:
I tested this locally using `/v1/sessions/idme_verified/new`. When signing in with my LOA1 test account I was redirected and asked to verify my information. I also attempted to create a new ID.me account using the same URL. After confirming my email I was again redirected and asked to verify my information. So it seems to be working, as far as I can tell.

bosawt:
Testing:
* Unverified idme account, `idme` param for sessions: logs in as an unverified account
* Unverified idme account, `idme_verified` param for sessions: asks for verification during ID.me authentication
* Verified idme account, `idme` param for sessions: logs in as a verified account
* Verified idme account, `idme_verified` param for sessions: error (this appears to be due to us using `LOA::IDME_LOA3_VETS` authn instead of `LOA::IDME_LOA3`
* `idme_signup` param for sessions: appears to be same behavior as `idme` param
* `idme_signup_verified` param for sessions: appears to be same behavior as `idme_verified` param
* Unverified login.gov account, `logingov` param for sessions: logs in as an unverified account
* Unverified login.gov account, `logingov_verified` param for sessions: asks for verification during login.gov authentication
* Verified login.gov account, `logingov` param for sessions: logs in as a verified account (two redirect loops through login.gov)
* Verified login.gov account, `logingov_verified` param for sessions: logs in as a verified account (single redirect loop through login.gov)
* `logingov_signup` param for sessions: appears to be same behavior as `logingov` param
* `logingov_signup_verified` param for sessions: appears to be same behavior as `logingov_verified` param
* `mhv`/`mhv_verified` param for sessions: redirects to MyHealtheVet auth form (not tested due to localhost)
* DS Logon account, `dslogon` param for sessions: logs in successfully
*  DS Logon account, `dslogon_verified` param for sessions: Logs in and asks for verification 
I believe all of these are expected, except the `idme_verified` and `idme_signup_verified`, which need to use `LOA::IDME_LOA3` instead of `LOA::IDME_LOA3_VETS`

spec/lib/saml/post_url_service_spec.rb
```diff
@@ -40,14 +40,15 @@
             )
           expect_any_instance_of(OneLogin::RubySaml::Settings)
             .to receive(:authn_context_comparison=).with('minimum')
-          url, params = subject.logingov_url
+          url, params = subject.login_url('logingov', [IAL::LOGIN_GOV_IAL1, AAL::LOGIN_GOV_AAL2],
```
bosawt, line 43: Also, can we add specs for the new `*_verified` types that were added?

bosawt:
Changes are looking pretty good! Can we add tests to `v1/sessions_controller_spec.rb` for the added type params? This would additional specs for the contexts:
`routes not requiring auth` (would add to the iterated types here)
`routes /sessions/idme_signup/new to SessionsController#new` (would copy these contexts and add the new signup types)

bosawt (approved):
Looks good! Thanks for making those changes!

PR #9147, opened 2022-02-16
Module for new and existing VaNotify functionality

nathanbwright:
## Description of change
The VA.gov forms strike team will be adding email notifications to a number of existing forms on VA.gov. We would like to migrate the existing `VaNotify` code at the top level app to our own module so that we can keep everything more modular and encapsulated. 
Creating a new module is the first step towards this goal. Right now the module is called `Notify` to avoid naming collisions with the existing `VaNotify` code. Once the new module has taken over responsibility of the existing code we can rename the module to better represent itself.


modules/notify/notify.gemspec
```diff
@@ -0,0 +1,20 @@
...
+
+# Describe your gem and declare its dependencies:
+Gem::Specification.new do |spec|
+  spec.name        = 'notify'
+  spec.version     = Notify::VERSION
+  spec.authors     = ['Nathan Wright']
+  spec.email       = ['nathan.wright@oddball.io']
+  spec.homepage    = 'https://api.va.gov'
+  spec.summary     = 'An api.va.gov module'
+  spec.description = 'This module was auto-generated please update this description'
```
LindseySaari, line 16: You may also want to update line 16 with a short desc as well

LindseySaari:
Looks good! Just a couple of comments to update the README and gemspec

LindseySaari (approved):

PR #9151, opened 2022-02-17
Mobile empty payment history bug

kreek:
## Description of change
Fixes a bug when a user's payment history is an empty list. Also splits up payment date filtering and pagination.

## Original issue(s)
department-of-veterans-affairs/va-mobile-app/issues/2564


kpethtel (approved):
Looks good, assuming it's not possible for the service to return a nil value.

kreek:
> Looks good, assuming it's not possible for the service to return a nil value.

@kpethtel correct, BGS returns an empty list for not found (the spec data is from test user Judy Morrison who has no payments)

PR #9153, opened 2022-02-17
Update sidekiq

LindseySaari:
## Description of change
This PR upgrades Sidekiq to 6.4.1 to avoid some Redis related warnings. Upgrading the gem fixes the warnings

```
* Listening on http://0.0.0.0:3000/
16:49:55 web.1    | Use Ctrl-C to stop
16:49:55 job.1    | Pipelining commands on a Redis instance is deprecated and will be removed in Redis 5.0.0.
16:49:55 job.1    | 
16:49:55 job.1    | redis.pipelined do
16:49:55 job.1    |  redis.get("key")
16:49:55 job.1    | end
16:49:55 job.1    | 
16:49:55 job.1    | should be replaced by
16:49:55 job.1    | 
16:49:55 job.1    | redis.pipelined do |pipeline|
16:49:55 job.1    |  pipeline.get("key")
16:49:55 job.1    | end
16:49:55 job.1    | 
16:49:55 job.1    | (called from /usr/local/bundle/gems/sidekiq-scheduler-3.1.1/lib/sidekiq-scheduler/redis_manager.rb:140:in `block in register_job_instance'}
16:49:55 job.1    | Pipelining commands on a Redis instance is deprecated and will be removed in Redis 5.0.0.
```
## Original issue(s)


lib/sidekiq/semantic_logging.rb
```diff
@@ -14,6 +14,7 @@ def call(_worker, item, queue)
       user_uuid: item['user_uuid'] || 'N/A',
       source: item['source']
     }
+    Thread.current[:sidekiq_context] = {}
```
ericboehs, line 17: @RachalCassity Could you provide some feedback on why this is neccessary and how you came about this solution?

RachalCassity: @ericboehs In the new version of Sidekiq, a perform action was added before the call.

`    
    def call(item, queue)
      start = ::Process.clock_gettime(::Process::CLOCK_MONOTONIC)
      @logger.info("start")

      yield

      Sidekiq::Context.add(:elapsed, elapsed(start))
      @logger.info("done")
    rescue Exception
      Sidekiq::Context.add(:elapsed, elapsed(start))
      @logger.info("fail")

      raise
    end

def prepare(job_hash, &block)
      # If we're using a wrapper class, like ActiveJob, use the "wrapped"
      # attribute to expose the underlying thing.
      h = {
        class: job_hash["display_class"] || job_hash["wrapped"] || job_hash["class"],
        jid: job_hash["jid"]
      }
      h[:bid] = job_hash["bid"] if job_hash.has_key?("bid")
      h[:tags] = job_hash["tags"] if job_hash.has_key?("tags")

      Thread.current[:sidekiq_context] = h
      level = job_hash["log_level"]
      if level
        @logger.log_at(level, &block)
      else
        yield
      end
    ensure
      Thread.current[:sidekiq_context] = nil
    end`

In this method,  Thread.current[:sidekiq_context] is being set to a hash.

In the Sidekiq::Context.add(:elapsed, elapsed(start)) method,  Thread.current[:sidekiq_context]["#{elasped(start)}"] is called. The job will error out if the Thread.current[:sidekiq_context] is not set to a hash.

thilton-oddball (approved):

PR #9160, opened 2022-02-18
[API-13174] - ClaimsApi - Handle failure to find Veteran 'FileNumber'

jw81:
## Description of change

- Adds logic to the Benefits ClaimsApi `PUT /forms/2122/:id` endpoint to find the Veteran's `FileNumber` in BGS.  If the `FileNumber` cannot be found, the call will return a 422 with a detail for how to contact the National Help Desk.  This is in an effort to fail-fast when we know we won't be able to upload a file to VBMS for the Veteran because they don't have a known `FileNumber`.

## Original issue(s)
[API-13174](https://vajira.max.gov/browse/API-13174)


modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
```diff
@@ -223,6 +224,25 @@ def build_representative_info(poa_code)
...
           end
         end
+
+        def check_file_number_exists!
+          ssn = target_veteran.ssn
+
+          begin
+            response = bgs_service.people.find_by_ssn(ssn) # rubocop:disable Rails/DynamicFindBy
+            unless response && response[:file_nbr].present?
+              error_message = 'Unable to locate Veteran FileNumber. '\
```
hinzed1127, line 234:
```suggestion
              error_message = 'Unable to locate Veteran file number. '\
```
assuming this messaging was discussed with Kayla and or whoever else would need to copy edit. looks fine to me other than this small tweak

modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
```diff
@@ -223,6 +224,25 @@ def build_representative_info(poa_code)
...
+                              'We apologize for the inconvenience. '\
+                              'Teams are actively working on getting this fixed. '\
```
hinzed1127, line 236: Do we know this is the case? I would be fine just shortening the error message to "Unable to locate Veteran file number. Please contact the National Help Desk at 1-800-983-0937 to have a ticket created."

hinzed1127 (dismissed):
Business logic/code looks fine. I just have a few questions around the error messaging.

hinzed1127 (approved):

PR #9162, opened 2022-02-18
Copy existing VaNotify service from top level into module

nathanbwright:
## Description of change
This PR copies the existing `VaNotify` service and config classes into the new `notify` module (to be renamed `va_notify` once the move is complete) as part of migrating all VaNotify functionality into its own module.

Instead of having the top level service in the `rubocop_todo.yml` https://github.com/department-of-veterans-affairs/vets-api/blob/master/.rubocop_todo.yml#L137 - I just made it inline... Should I have just put it in the todo file instead?


modules/notify/spec/lib/service_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+require 'notify/service'
+
+describe Notify::Service do
+  before(:example, test_service: false) do
+    @test_api_key = 'test-aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa-bbbbbbbb-bbbb-bbbb-bbbb-bbbbbbbbbbbb'
+    @test_base_url = 'http://fakeapi.com'
+    allow_any_instance_of(Notify::Configuration).to receive(:base_path).and_return(@test_base_url)
+  end
+
+  send_email_parameters = {
```
LindseySaari, line 13: I'm wondering if you should wrap the hash into a before(:each)? Prob nitpicky ha

modules/notify/lib/notify/service.rb
```diff
@@ -0,0 +1,78 @@
...
+  class Service < Common::Client::Base
+    include Common::Client::Concerns::Monitoring
+
+    STATSD_KEY_PREFIX = 'api.vanotify'
+
+    configuration VaNotify::Configuration
+
+    attr_reader :notify_client
+
+    # rubocop:disable Lint/MissingSuper
```
LindseySaari, line 17: Is there a way around this without disabling inline? For example, I see that [this service class ](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/meb_api/lib/dgi/service.rb)in a module didn't have the same call to super linting issue

LindseySaari: Nvm ignore me! It looks like we're adding the cop disabling to the .rubocop_todo.yml

modules/notify/spec/lib/service_spec.rb
nathanbwright: Yeah, I was just copying over what already existed, but that is probably a good call.

modules/notify/lib/notify/service.rb
LindseySaari: @nathanbwright You'll just have to update the rubocop_todo! Thanks!

PR #9164, opened 2022-02-18
2275 update mobile user for logingov

jperk51:
## Description of change
Versioned mobile user endpoint to return logingov as a login service. Updated old user endpoint to map logingov to idme in order for older versions of the app to treat it as multifactor correctly. 

## Original issue(s)
department-of-veterans-affairs/va-mobile-app#2275

Added new tests to both v0 and v1 user request specs to make sure signin service is correct. Also updated the docs with the new user endpoint


LindseySaari (approved):

kpethtel (dismissed):
I don't really have enough knowledge of the underlying issue, but the code looks good. 👍

LindseySaari (approved):

PR #9165, opened 2022-02-18
Skip RRD logic when restarting EVSS submission of Form 526

nanotone:
## Description of change
This PR fixes a potential bug in Rapid Ready for Decision -related logic, that could result in multiple RRD PDFs being uploaded and submitted to EVSS. The underlying cause is that `Form526Submission#start` is not idempotent with respect to preparing RRD PDFs for submission.

The bug was most likely introduced in #8558 when the previous `#start` functionality was [moved into a new method](https://github.com/department-of-veterans-affairs/vets-api/commit/4f49fd64b6e8200b702b3b30004cb3da94a56040#diff-ada3116cb66cda03a293e8c804592ccb3bc181f3e64689c66fe27d19b0d3cad1R76) called `#start_evss_submission`. However, this rename should have also been reflected in [start_but_use_a_birls_id_that_hasnt_been_tried_yet!](https://github.com/department-of-veterans-affairs/vets-api/blob/6d614e4326895badc7288b11bcac676373dc068e/app/models/form526_submission.rb#L111) which is a retry mechanism that assumes the submission was already in order for a previous EVSS attempt.

So in addition to updating the implementation of that method, I've gone ahead and given it the slightly more informative name `submit_with_birls_id_that_hasnt_been_tried_yet!` -- and tweaked the comment.

The [long-term resolution](https://www.notion.so/Can-we-make-the-PDF-generation-idempotent-f87a21da16a740d687a8a2d5c81033b9) for this is to make `#start` actually idempotent, but that's a heavier lift and to my knowledge there is no other place that `#start` should've been replaced with `#start_evss_submission`.

## Original issue
[Notion ticket](https://www.notion.so/Don-t-Call-start-here-because-our-start-method-is-NOT-idempotent-2a980844af7c456bae25f054c4ec1a66)
Props to @michelpmcdonald for spotting this and alerting the RRD team.


yoomlam (dismissed):
👍

nanotone:
Security lint is failing for now, I'll rebase once https://nokogiri.org/CHANGELOG.html#1132-2022-02-21 gets folded into `master`.

michelpmcdonald (approved):
Looks good

thilton-oddball (approved):

PR #9166, opened 2022-02-18
2440: add appointment requests to appointments index

kpethtel:
## Description of change
This utilizes the new appointment request adapter. 

## Original issue(s)
department-of-veterans-affairs/va-mobile-app/issues/2440

The actual fetching of the appointment requests is behind a feature flag. Inclusion of the appointment requests in the response to the user is controlled via a query param. The reason it's gated this way is because the appointment requests are cached with appointments. Once we've turned on the flag, appointment requests must be fetched to ensure that the cache is correct in case a developer uses the query param to include appointment requests. 

This will be tested using curl on a deploy environment.


modules/mobile/app/controllers/mobile/v0/appointments_controller.rb
```diff
@@ -29,6 +29,9 @@ def index
 
         appointments = fetch_cached_or_service(validated_params)
         page_appointments, page_meta_data = paginate(appointments, validated_params)
+        if validated_params[:included].include?('pending')
```
kpethtel, line 32: This is the light-weight alternative to handling this in the pagination helper.

modules/mobile/app/controllers/mobile/v0/appointments_controller.rb
```diff
@@ -81,19 +84,26 @@ def fetch_cached_or_service(validated_params)
...
         appointments.reverse! if validated_params[:reverse_sort]
+
         appointments
       end
 
       def paginate(appointments, validated_params)
         appointments = appointments.filter do |appointment|
           appointment.start_date_utc.between?(
-            validated_params[:start_date], validated_params[:end_date]
+            validated_params[:start_date].beginning_of_day, validated_params[:end_date].end_of_day
```
kpethtel, line 94: I added this because I was testing with a one day date range and experiencing issues.

kpethtel:
Closing temporarily while I fix something.

modules/mobile/app/controllers/mobile/v0/appointments_controller.rb
aherzberg: do the start and end dates always formatted with just dates, no times? if so, could this also be utilized in the primary appointments call in `fetch_cached_or_service`?

kpethtel: It's a good question, but `fetch_cached_or_service` has a totally different behavior. It always sets the begin date to at least a year ago and always sets the end date to at least a year from now: 
```
          appointments = appointments_proxy.get_appointments(
            start_date: [validated_params[:start_date], one_year_ago].min,
            end_date: [validated_params[:end_date], one_year_from_now].max
          )
```
It's combined with this method:
```
      def use_cache?(validated_params)
        # use cache if date range is within +/- 1 year and use_cache is true
        validated_params[:start_date] >= one_year_ago.iso8601 &&
          validated_params[:end_date] <= one_year_from_now.iso8601 && validated_params[:use_cache]
      end
```
to ensure that the cache is always accurate when requesting between those dates. Which is undoubtedly the large majority of requests. If we're outside of those dates, we don't use the cache. There might also be some slight edge cases there, but they're a lot less noticeable because of the wider date range. The issue that caused me to experience this bug is that I was using very narrow date ranges. Which meant that the controller was requesting two years of data, but then filtering out everything.

config/features.yml
```diff
@@ -340,6 +340,10 @@ features:
   mobile_api:
     actor_type: user
     description: API endpoints consumed by the VA Mobile App (iOS/Android)
+  mobile_appointment_requests:
```
kpethtel, line 343: Not sure if I truly need to add this, but it makes it easier to flip the flag.

kreek (dismissed):
I think this is looking good, I took a stab a cleaning up the model generation i.e. moving the VA and CC embedded classes to other files in this branch `2440-add-appointment-requests-to-appointments-templates`. Feel free to use or improve.

config/features.yml
kpethtel: I ended up later removing this because it was mysteriously breaking specs. But now I've re-added it and realize the spec failures are because flipper.rb turns all features on in the test environment. I didn't expect that behavior, but it's clearly written in the code.

kpethtel:
> I think this is looking good, I took a stab a cleaning up the model generation i.e. moving the VA and CC embedded classes to other files in this branch `2440-add-appointment-requests-to-appointments-templates`. Feel free to use or improve.

I figured the use of sub-classes would eventually come up.

modules/mobile/app/controllers/mobile/v0/appointments_controller.rb
```diff
@@ -81,19 +84,26 @@ def fetch_cached_or_service(validated_params)
...
           )
         end
         url = request.base_url + request.path
         Mobile::PaginationHelper.paginate(list: appointments, validated_params: validated_params, url: url)
       end
 
+      def include_pending_in_links(links)
```
kpethtel, line 101: I believe this should be sufficient because there should always be other query params. If I'm wrong, I can make this more robust with the URI library.

kreek (dismissed):
LGTM 🚀

modules/mobile/app/controllers/mobile/v0/appointments_controller.rb
aherzberg: All good points. I also missed that the `one_year_from_now` and `one_year_ago` methods were already using the `beginning_of_day`

thilton-oddball (approved):

aherzberg (approved):
LGTM

PR #9175, opened 2022-02-21
Fix error handling in PoA signatures

acovrig:
## Description of change

When there's a `StampSignatureError`, set the `status` to `ERRORED` and store `signature_errors`.

## Original issue(s)
[API-13337](https://vajira.max.gov/browse/API-13337)


hinzed1127 (changes requested):
Just need to add a test or two for this update and then should be good to go!

modules/claims_api/spec/workers/poa_form_builder_job_spec.rb
```diff
@@ -90,5 +91,29 @@
...
+            signatures: {
+              veteran: bad_b64_image,
+              representative: bad_b64_image
+            }
+          }
+        ))
+      end
+
+      it 'generates the pdf to match example' do
+        allow_any_instance_of(ClaimsApi::PoaPdfConstructor::Organization).to receive(:construct)
```
jw81, line 110: Can you try something for me?
Can you change this to `expect_any_instance_of`?

If I'm thinking correctly, this check at this moment isn't really doing anything.  I think it's just saying, "_if an instance receives this message and raises this exception, then cool, else, don't worry about it._".  
Whereas, in this test, you're actually feeding it bad data and you are expecting it to raise an exception.

modules/claims_api/app/workers/claims_api/poa_form_builder_job.rb
```diff
@@ -22,7 +22,8 @@ def perform(power_of_attorney_id)
     rescue Errno::ENOENT
       rescue_file_not_found(power_of_attorney)
     rescue ClaimsApi::StampSignatureError => e
-      power_of_attorney.update(signature_errors: e.detail)
+      signature_errors = (power_of_attorney.signature_errors || []).push(e.detail)
```
hinzed1127, line 25: in the case that `power_of_attorney.signature_errors` is already defined, wouldn't we still want to add push `e.detail` and not just in the latter case where it's getting added to an empty array?

acovrig: I'm not sure I understand the question - the code it supposed to take the current `signature_errors` and add the current error to it. the `|| []` bit should make it an empty array if it's `nil`, I presume it will never be `nil` but figure it couldn't hurt to put that bit in just in case.

hinzed1127: oh my bad, I missed the parentheses. I was reading it as
```ruby 
signature_errors = power_of_attorney.signature_errors || [].push(e.detail)
```
your logic makes total sense, disregard my question.

hinzed1127 (approved):

jw81 (approved):

PR #9180, opened 2022-02-22
Refactor Form526Submission to separate concerns

yoomlam:
## Description of change

In preparation for adding more logic to handle additional contention types, this PR refactors 1 model class by extracting code from `Form526Submission` into new concerns `Form526Workflow` and `Form526BirlsIds` to make the code more manageable and to separate concerns.
* `Form526Workflow` has code relating to Sidekiq jobs and steps in the Form 526 workflow
* `Form526BirlsIds` has code relating to BIRLS ids

This results in a `Form526Submission` that is a much smaller model class with 2 entry points (`start` and `start_but_use_a_birls_id_that_hasnt_been_tried_yet!`), along with methods to access the model's attributes.

## Original issue(s)
`Form526Submission` is becoming large with code relating to different concerns or responsibilities.

No change in behavior. Only a code refactor.

The test `spec/models/form526_submission_spec.rb` can be refactored in a separate PR to make it clear that this PR makes minimal changes to the test, ensuring that existing behavior is retained.


app/models/concerns/form526_workflow.rb
```diff
@@ -0,0 +1,173 @@
...
+  #
+  # @return [String] the job id of the first job in the batch, i.e the 526 submit job
+  #
+  def start_evss_submission(_status, options)
+    submission = Form526Submission.find(options['submission_id'])
+    id = submission.id
+    workflow_batch = Sidekiq::Batch.new
+    workflow_batch.on(
+      :success,
+      'BatchDoneHandlers#perform_ancillary_jobs_handler',
```
yoomlam, line 26: This might need to change to
```suggestion
      'Form526Workflow::BatchDoneHandlers#perform_ancillary_jobs_handler',
```
but I'm not sure how to test it because "Batches do not support the inline or fake testing modes" -https://github.com/mperham/sidekiq/wiki/Batches#testing.

I might create callback classes instead: https://github.com/mperham/sidekiq/wiki/Batches#callbacks

app/models/concerns/form526_workflow.rb
```diff
@@ -0,0 +1,172 @@
...
+
+  FORM_526 = 'form526'
+  FORM_526_UPLOADS = 'form526_uploads'
+  FORM_4142 = 'form4142'
+  FORM_0781 = 'form0781'
+  FORM_8940 = 'form8940'
+  FLASHES = 'flashes'
+  SUBMIT_FORM_526_JOB_CLASSES = %w[SubmitForm526AllClaim SubmitForm526].freeze
+
+  class EvssSubmission
```
yoomlam, line 14: The 3 classes in this file could be extracted into separate files in `app/services` directory, if desired.

app/models/concerns/form526_workflow.rb
```diff
@@ -0,0 +1,168 @@
...
+    # Callback method called after RRD Sidekiq::Batch job succeeds in Form526Submission
+    def on_success(_status, options)
```
nanotone, line 16: Even though you have this nice informative comment here, it seems a little weird to just call this method `EvssSubmission#on_success` which makes it sound like the EVSS submission was already successful. Is there a generic enough class name with the suffix `Handler` that could be used here, to match the other two class names?

app/models/concerns/form526_workflow.rb
nanotone: To rephrase [my other comment](https://github.com/department-of-veterans-affairs/vets-api/pull/9180#discussion_r813003500) in this context, isn't `EvssSubmission` also a Sidekiq::Batch callback class? Given how it's being used in [Form526Submission#start](https://github.com/department-of-veterans-affairs/vets-api/pull/9180/files#diff-ada3116cb66cda03a293e8c804592ccb3bc181f3e64689c66fe27d19b0d3cad1R47).

app/models/concerns/form526_workflow.rb
yoomlam: How about renaming `EvssSubmission` to `PrepEvssSubmissionJobHandler`?

app/models/concerns/form526_workflow.rb
yoomlam: Yes, it is. Unlike the other 2 `*JobHandler`, its `start_evss_submission` class method is called directly in various locations, as you are aware.

yoomlam:
Closing this PR in favor of #9206.

PR #9181, opened 2022-02-22
RRD refactoring: extract new ProcessorSelector class from Form526Submission

yoomlam:
## Description of change

In preparation for adding more logic to handle additional contention types to RapidReadyForDecision (RRD), this PR extracts most RRD-specific code from `Form526Submission` into a new `RapidReadyForDecision::ProcessorSelector` service class.

Also creates `RapidReadyForDecision::DiagnosticCodes` for constants that are used in multiple locations and makes the code more readable.

## Original issue(s)

Existing code is not conducive to add more contention types for RRD. RRD code should be separated into its own class to avoid add more complexity to `Form526Submission`.

No change in behavior. Only a code refactor.

Adds `trait :hypertension_claim_for_increase` to `spec/factories/form526_submissions.rb` to DRY up some of the code and for reuse in future tests.


app/models/form526_submission.rb
```diff
@@ -47,17 +47,18 @@ class Form526Submission < ApplicationRecord
   SUBMIT_FORM_526_JOB_CLASSES = %w[SubmitForm526AllClaim SubmitForm526].freeze
 
   def start
-    if single_issue_hypertension_claim? && Flipper.enabled?(:disability_hypertension_compensation_fast_track)
+    rrd_processor_class = rrd_process_selector.processor_class
```
yoomlam, line 50: Move this hypertension-check logic into `ProcessorSelector`.

app/models/form526_submission.rb
```diff
@@ -337,13 +342,6 @@ def bdd?
     form.dig('form526', 'form526', 'bddQualified') || false
   end
 
-  def single_issue_hypertension_claim?
-    disabilities = form.dig('form526', 'form526', 'disabilities')
-    disabilities.count == 1 &&
-      disabilities.first['disabilityActionType'] == 'INCREASE' &&
-      disabilities.first['diagnosticCode'] == 7101
-  end
-
```
yoomlam, line 346: Move into `ProcessorSelector`

app/workers/evss/disability_compensation_form/submit_form526.rb
```diff
@@ -55,10 +55,7 @@ def perform(submission_id)
       private
 
       def send_hypertension_fast_track_pilot_email(submission)
-        if submission.single_issue_hypertension_claim? && Flipper
-           .enabled?(:disability_hypertension_compensation_fast_track)
-          HypertensionFastTrackPilotMailer.build(submission).deliver_now
-        end
+        HypertensionFastTrackPilotMailer.build(submission).deliver_now
```
yoomlam, line 58: Move condition check outside this method for readability -- see line 42.

app/workers/rapid_ready_for_decision/disability_compensation_job.rb
```diff
@@ -31,7 +31,7 @@ def perform(form526_submission_id, full_name)
 
           return if bp_readings(client).blank?
 
-          pdf = pdf(full_name, bp_readings(client), medications(client))
+          pdf = pdf(form526_submission.full_name, bp_readings(client), medications(client))
```
yoomlam, line 34: Use the `full_name` from the `form526_submission` instead of passing it in.

app/models/form526_submission.rb
```diff
@@ -47,17 +47,18 @@ class Form526Submission < ApplicationRecord
...
+    rrd_processor_class = rrd_process_selector.processor_class
+    if rrd_processor_class
       workflow_batch = Sidekiq::Batch.new
       workflow_batch.on(
         :success,
         'Form526Submission#start_evss_submission',
         'submission_id' => id
       )
-      jids = workflow_batch.jobs do
-        RapidReadyForDecision::DisabilityCompensationJob.perform_async(id, full_name)
```
yoomlam, line 58: Instead of passing in the result of calling `full_name`, within the job's `perform` method, call `full_name` on the `Form526Submission` record identified by `id` . The user's full name is not expected to change between calling `full_name` here versus calling it within the job.

spec/factories/form526_submissions.rb
```diff
@@ -20,6 +20,12 @@
     end
   end
 
+  trait :hypertension_claim_for_increase do
+    form_json do
+      File.read("#{::Rails.root}/spec/support/disability_compensation_form/submissions/only_526_hypertension.json")
+    end
+  end
+
```
yoomlam, line 28: Creating `hypertension_claim_for_increase` submissions is now easier and more readable in tests.

spec/services/rapid_ready_for_decision/hypertension_special_issue_manager_spec.rb
```diff
@@ -27,7 +27,9 @@ def form526_hash(form)
   end
 
   def filter_disabilities(form)
-    form[:disabilities].filter { |item| item[:diagnosticCode] == 7101 }
+    form[:disabilities].filter do |item|
+      item[:diagnosticCode] == RapidReadyForDecision::DiagnosticCodes::HYPERTENSION
+    end
```
yoomlam, line 32: Now we know what `7101` means.

app/services/rapid_ready_for_decision/hypertension_special_issue_manager.rb
```diff
@@ -17,24 +19,17 @@ def add_special_issue
 
     private
 
-    def submission_data
-      @submission_data ||= JSON.parse(submission.form_json, symbolize_names: true)
-    end
```
yoomlam, line 22: The `add_special_issue` method parses the submission form and then writes it back out to JSON, so there's not much value in `symbolize_names: true`. Also, make it consistent with how `Form526Submission` parses the form [here](https://github.com/department-of-veterans-affairs/vets-api/blob/3c48ee89440de8898b011079bffe547871a61096/app/models/form526_submission.rb#L154) so that we can maybe DRY it up later.

spec/factories/form526_submissions.rb
allthesignals: I didn't realize we had factories setup — great!

app/services/rapid_ready_for_decision/diagnostic_codes.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module RapidReadyForDecision
+  module DiagnosticCodes
+    HYPERTENSION = 7101
+  end
+end
```
allthesignals, line 7: Out of curiosity what was the reasoning for making this its own module? Is this a pattern for dealing with constants?

app/services/rapid_ready_for_decision/processor_selector.rb
```diff
@@ -0,0 +1,48 @@
...
+module RapidReadyForDecision
+  class ProcessorSelector
+    def initialize(form526_submission)
+      @form526_submission = form526_submission
+    end
+
+    def processor_class
+      return nil unless rrd_enabled?
+
+      if hypertension_enabled? && single_issue_claim_applicable?(DiagnosticCodes::HYPERTENSION)
```
allthesignals, line 12: I wonder if this condition can be wrapped up more neatly in a named method, if there's a good term...

app/services/rapid_ready_for_decision/diagnostic_codes.rb
yoomlam: There wasn't an existing class AFAIK that made sense where I could put a constant like:
```ruby
DIAGNOSTIC_CODES = {
  hypertension: 7101
}
```
It's used for various classes in RRD, so I ended up putting the constant in its own file. I'm open to suggestions.

app/services/rapid_ready_for_decision/processor_selector.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    def hypertension_enabled?
+      Flipper.enabled?(:disability_hypertension_compensation_fast_track)
+    end
+
+    def form_disabilities
+      @form_disabilities ||= @form526_submission.form.dig('form526', 'form526', 'disabilities')
+    end
+
+    def single_issue_claim_applicable?(diagnostic_code)
```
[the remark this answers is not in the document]
yoomlam: I presume there will be different codes for asthma and sleep apnea.

app/services/rapid_ready_for_decision/processor_selector.rb
```diff
@@ -0,0 +1,48 @@
...
+      true
+    end
+
+    def hypertension_enabled?
+      Flipper.enabled?(:disability_hypertension_compensation_fast_track)
+    end
+
+    def form_disabilities
+      @form_disabilities ||= @form526_submission.form.dig('form526', 'form526', 'disabilities')
+    end
```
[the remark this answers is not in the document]
yoomlam: > Seems memoizing should only happen in their own methods.

Could you clarify your feedback?

`form_disabilities` is called twice in `single_issue_claim_applicable?`, so I created `form_disabilities` so that we can easily retrieve `form.dig('form526', 'form526', 'disabilities')`. I could remove the memoization since `@form526_submission.form` is already memoized.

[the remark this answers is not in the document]
allthesignals: I meant that as a question. I'm mostly just checking that there's real benefit to memoizing here and it seems there is. Thanks!

allthesignals (approved):

app/services/rapid_ready_for_decision/processor_selector.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    private
+
+    def rrd_enabled?
+      # In next PR, change this to be a Flipper configuration to disable RRD completely
+      true
+    end
+
+    def hypertension_enabled?
```
julialeague, line 35: It could be premature to make this method more generalized and applicable to any contention's Flipper check, but the upside is that it might enforce some kind of naming standard for the Flipper feature. Just a thought!

yoomlam: Are you suggesting a change that looks something like this:
```suggestion
    def hypertension_enabled?(disability)
      Flipper.enabled?("disability_#{disability}_compensation_fast_track".to_sym)
```
If so, I like it. We may want to rename the Flipper key to include `rrd` or something to that effect.

LindseySaari (approved):

app/services/rapid_ready_for_decision/processor_selector.rb
julialeague: Yeah, that looks like what I was thinking of! Maybe the name could be `rrd_#{disability}_compensation` or something like that?

spec/services/rapid_ready_for_decision/processor_selector_spec.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe RapidReadyForDecision::ProcessorSelector do
+  subject { described_class.new(submission) }
+
+  let(:submission) { create(:form526_submission) }
```
julialeague, line 8: Is it possible to just `build` the form526 submission factory instead of `create`? I always ask, since there's a significant performance slowdown when using `create`.

julialeague (approved):

PR #9183, opened 2022-02-22
VaNotify rename module

nathanbwright:
## Description of change
This PR renames the `notify` module to be `va_notify`. It removes the existing classes in `/lib/va_notify`. Because we are using the same namespace it should be a drop in replacement for all existing code that relies on the `VaNotify::Service` and `VaNotify::Configuration` classes.


thilton-oddball (dismissed):

spec/spec_helper.rb
```diff
@@ -60,7 +60,7 @@
     add_group 'Identity', 'modules/identity/'
     add_group 'MebApi', 'modules/meb_api/'
     add_group 'Mobile', 'modules/mobile/'
-    add_group 'Notify', 'modules/notify/'
+    add_group 'VaNotify', 'modules/va_notify/'
```
LindseySaari, line 63: One nitpick: Would you be able to alphabetize this entry? I believe the module generator does the alphabetization (which was originally correct when the module was named 'Notify')

spec/simplecov_helper.rb
```diff
@@ -68,7 +68,7 @@ def self.add_modules
     add_group 'HealthQuest', 'modules/health_quest'
     add_group 'Identity', 'modules/identity/'
     add_group 'MebApi', 'modules/meb_api/'
-    add_group 'Notify', 'modules/notify/'
+    add_group 'VaNotify', 'modules/va_notify/'
```
LindseySaari, line 71: Here too please!

LindseySaari (approved):

PR #9193, opened 2022-02-22
API-13467: Fix POA logic for PDF generation and updating POA codes

hinzed1127:
## Description of change

This fixes 2 bugs with the POA 21-22 submission process in Claims v1:
1. Veterans and Representatives were being treated differently. 21-22 requires a signed form regardless of who is submitting, so we should enforce that regardless of who is submitted
2. Since updating POA codes in BGS is asynchronous, use the POA code as part of the submission request body and not the POA code returned by BGS (since this could be the previous POA code and the code being used to update).

## Original issue(s)
https://vajira.max.gov/browse/API-13467


modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
```diff
@@ -47,14 +47,17 @@ def submit_form_2122 # rubocop:disable Metrics/MethodLength
...
           if enable_vbms_access?
-            ClaimsApi::VBMSUpdater.perform_async(power_of_attorney.id,
-                                                 target_veteran.participant_id)
+            ClaimsApi::VBMSUpdater.perform_async(power_of_attorney.id, target_veteran.participant_id)
           end
+
           data = power_of_attorney.form_data
-          ClaimsApi::PoaFormBuilderJob.perform_async(power_of_attorney.id) if data['signatures'].present?
+
+          if data['signatures'].present?
```
hinzed1127, line 56: one thought I had around validating this: is this enough, or should we go one level deeper and make sure that `data['signatures']` contains `data['signatures']['veteran']` and `data['signatures']['representative']`? I'm leaning towards adding in the stricter validation

jw81: I don't think there's any harm in adding that stricter validation.  I think the `validate_json_schema` call will blow up if `signatures` is included, but doesn't include both the nested `veteran` and `representative` keys.  But it also isn't going to hurt anything to be extra thorough here as well.

modules/claims_api/lib/claims_api/vbms_sidekiq.rb
```diff
@@ -6,6 +6,20 @@ module ClaimsApi
   module VBMSSidekiq
     include SentryLogging
```
jw81, line 8: Just making doubly-sure, but all this change did was move the existing logic closer to the top of the class, right?
That's what it looks like and I love the change BTW.  Just want to make sure I'm not overlooking some additional change.

modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
```diff
@@ -47,14 +47,17 @@ def submit_form_2122 # rubocop:disable Metrics/MethodLength
...
+            ClaimsApi::VBMSUpdater.perform_async(power_of_attorney.id, target_veteran.participant_id)
           end
+
           data = power_of_attorney.form_data
-          ClaimsApi::PoaFormBuilderJob.perform_async(power_of_attorney.id) if data['signatures'].present?
+
+          if data.dig('signatures', 'veteran').present? && data.dig('signatures', 'representative').present?
+            # Autogenerate a 21-22 form from the request body and upload it to VBMS.
+            # If upload is successful, then the PoaUpater job is also called to update the code in BGS.
+            ClaimsApi::PoaFormBuilderJob.perform_async(power_of_attorney.id, poa_code)
```
jw81, line 59: I'm surprised this isn't blowing up because it looks like `poa_code` is being passed as a 2nd arg, but the `perform` method signature is defined to only accept 1 arg.  🤔

modules/claims_api/app/workers/claims_api/vbms_upload_job.rb
```diff
@@ -9,12 +9,17 @@ class VBMSUploadJob
     include Sidekiq::Worker
     include ClaimsApi::VBMSSidekiq
 
+    # Uploads a 21-22 or 21-22a form for a given POA request to VBMS.
+    # If successfully uploaded, it queues a job to update the POA code in BGS, as.
```
jw81, line 13: I think there's maybe a word or two missing at the end of this sentence?

modules/claims_api/app/workers/claims_api/poa_form_builder_job.rb
```diff
@@ -11,12 +11,19 @@ class PoaFormBuilderJob
...
+    # Uploads the generated form to VBMS. If successfully uploaded,
+    # it queues a job to update the POA code in BGS, as well.
+    #
+    # @param power_of_attorney_id [String] Unique identifier of the submitted POA
     def perform(power_of_attorney_id)
       power_of_attorney = ClaimsApi::PowerOfAttorney.find(power_of_attorney_id)
+      poa_code = power_of_attorney.form_data['serviceOrganization']['poaCode']
 
-      output_path = pdf_constructor(power_of_attorney).construct(data(power_of_attorney), id: power_of_attorney.id)
+      output_path = pdf_constructor(poa_code).construct(data(power_of_attorney))
```
jw81, line 23: Do we not need to include that `id:` arg as a 2nd argument to the `construct()` method anymore?
It looks like it'll default to a `SecureRandom.uuid` value if not defined.  But maybe there's value in continuing to allow it to use the id of the POA record as the identifier?  🤷

modules/claims_api/app/workers/claims_api/vbms_upload_job.rb
hinzed1127: some copy-paste stuff getting lost between files/commits. fixing in this next commit

modules/claims_api/app/workers/claims_api/poa_form_builder_job.rb
hinzed1127: oops, that's a good catch. I think that was an unnecessary removal as I was changing the first call to just have the `poa_code`. I'll revert that in this next commit, too.

modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
hinzed1127: good catch, removing it in this next commit. I'm also surprised it didn't cause tests to fail

jw81 (approved):

PR #9199, opened 2022-02-23
2322 payment history logging

aherzberg:
## Description of change

Adding additional logging for payment history. looks like the upstream service will return nil when they rescue errors in some cases so handling this as well. Some sentry errors have been thrown due to nil being returned. 

error was `undefined method '[]' for nil:NilClass on app/models/adapters/payment_history_adapter.rb in initialize at line 8`
## Original issue(s)
https://github.com/department-of-veterans-affairs/va-mobile-app/issues/2322


jperk51 (approved):
LGTM

LindseySaari (approved):

PR #9202, opened 2022-02-23
[API-13567] - ClaimsApi - Retrieve Veteran file number and provide to vbms

jw81:
## Description of change

- Updates the background logic for uploading POA docs to VBMS so that it can retrieve and pass the Veteran's `FileNumber` to VBMS when making the call.

## Original issue(s)
[API-13567](https://vajira.max.gov/browse/API-13567)


modules/claims_api/lib/claims_api/vbms_sidekiq.rb
```diff
@@ -33,17 +47,24 @@ def rescue_vbms_file_number_not_found(power_of_attorney)
...
-      power_of_attorney.update(
-        status: ClaimsApi::PowerOfAttorney::UPLOADED,
-        vbms_new_document_version_ref_id: upload_response[:vbms_new_document_version_ref_id],
-        vbms_document_series_ref_id: upload_response[:vbms_document_series_ref_id]
+    private
+
+    def retrieve_veteran_file_number(power_of_attorney:)
+      ssn = power_of_attorney.auth_headers['va_eauth_pnid']
+
+      begin
```
hinzed1127, line 55: more general ruby question: isn't this `begin` not strictly needed and the rescue would still work as intended within the method block?

hinzed1127 (approved):

modules/claims_api/lib/claims_api/vbms_sidekiq.rb
jw81: Yeah. I usually write it like this and then byebug gripes at me.  But it seems like sometimes byebug doesn't enforce it.  🤷

hinzed1127 (approved):

PR #9203, opened 2022-02-23
2440: appointment request adapter specs

kpethtel:
## Description of change
Adds specs for recent adapter changes. After this is merged, I will put out a separate PR for breaking the appointment requests adapter into templates. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#2440

It's just specs. There weren't enough LoC to include these in previous PRs.


modules/mobile/spec/models/adapters/appointment_requests_adapter_spec.rb
```diff
@@ -0,0 +1,339 @@
...
+    subject.parse(data)
+  end
+  let(:va_appointment_requests) { adapted_appointment_requests[0] }
+  let(:cc_appointment_requests) { adapted_appointment_requests[1] }
+  let(:all_appointment_requests) { [va_appointment_requests + cc_appointment_requests].flatten }
+  let(:booked_request_id) { '8a48dea06c84a667016c866de87c000b' }
+  let(:resolved_request_id) { '8a48e8db6d7682c3016d88dc21650024' }
+  let(:va_request_id) { '8a48e8db6d70a38a016d72b354240002' }
+  let(:cc_request_id) { '8a48912a6d02b0fc016d20b4ccb9001a' }
+  let(:va_request_data) do
```
aherzberg, line 23: This is a good way to track what each request is and describe how it's going to be used with the variable name instead of picking them out of an array as needed for each test. We should do something similar with the appointments list request spec sometime.

aherzberg (approved):
LGTM

modules/mobile/spec/models/adapters/appointment_requests_adapter_spec.rb
kpethtel: Yeah, I wasn't about to refactor 1k lines of specs, but I agree in principle. There are multiple valid ways of structuring your specs. At my last place, the principal developer preferred to avoid using `let` much at all and preferred to define most things as local variables within tests. There are some real advantages to that approach. As test files grow, it becomes increasingly difficult to tell what something is defined as in a certain scope. And as teams grow, it becomes increasingly likely that new people will add new tests without looking up to see what's available to them and end up re-creating things. It's all just trade-offs.

aherzberg: oh yeah, definitely not within scope of this issue. If a spec does get too large where that's a problem, that'll have some code smell and something else is likely going on.

PR #9206, opened 2022-02-24
RRD: prep for new contention type - part 1

yoomlam:
## Description of change

In preparation for adding more logic to handle additional contention types for RRD (RapidReadyForDecision), this PR cleans up some code:

* uses `build` instead of `create` for FactoryBot calls to speed up tests
* uses FactoryBot `trait :hypertension_claim_for_increase` to DRY (don't repeat yourself) up code
* clarifies use of `Sidekiq::Batch` by adding comments and renaming methods to consistently use `*_handler` and `*_job`
* removes unused `SubmitForm526JobStatusesError`
* DRY up `Form526ConfirmationEmailJob` and `Form526SubmissionFailedEmailJob`
* fixes typos

Note the reduction in lines of code, which means less code to maintain!

## Original issue(s)
RRD (RapidReadyForDecision) will likely need to tweak the Form526Submission workflow (which uses `Sidekiq::Batch`). This PR does some clean up to prepare for those tweaks. 

## Things to know about this PR

No change in behavior. Only a code refactor and clean up.

PR #9207, opened 2022-02-24
RRD: prep for new contention type - part 2: add refactored hypertension job

yoomlam:
This is part 2 of "RRD: prep for new contention type", part 1 is PR #9206.

## Description of change

In preparation for adding more logic to handle additional contention types for RRD (RapidReadyForDecision), this PR adds refactored classes called `Form526HypertensionJob` and `Form526BaseJob`, which mimics `DisabilityCompensationJob` but is designed to be more extensible for new contention types.

## Original issue(s)
`DisabilityCompensationJob` was written as a prototype and supports a single contention type. We need to support additional contention types.

New classes are added in the PR, but they are not yet used. To keep this PR small, they will be used/enabled in a separate PR.


app/workers/rapid_ready_for_decision/form526_base_job.rb
```diff
@@ -0,0 +1,107 @@
...
+      # rubocop:enable Lint/UselessAssignment
+    end
+
+    def upload_pdf(form526_submission, pdf)
+      RapidReadyForDecision::HypertensionUploadManager.new(form526_submission).handle_attachment(pdf.render)
+    end
+
+    def set_special_issue(form526_submission)
+      RapidReadyForDecision::HypertensionSpecialIssueManager.new(form526_submission).add_special_issue
+    end
```
yoomlam, line 105: These lines currently refer to `Hypertension` classes to maintain parity with `DisabilityCompensationJob`. A future  PR will generalize and rename these classes.

app/workers/rapid_ready_for_decision/form526_hypertension_job.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+require 'lighthouse/veterans_health/client'
+
+module RapidReadyForDecision
+  class Form526HypertensionJob < Form526BaseJob
```
yoomlam, line 6: Contention-specific RRD jobs should now be simpler, like this one.

app/workers/rapid_ready_for_decision/form526_hypertension_job.rb
```diff
@@ -0,0 +1,51 @@
...
+      assessed_data
+    end
+
+    private
+
+    def query_and_assess_lighthouse(form526_submission)
+      client = lighthouse_client(form526_submission)
+      bp_readings = assess_bp_readings(client.list_resource('observations'))
+      # stop querying if bp_readings.blank?
+      medications = assess_medications(client.list_resource('medications')) if bp_readings.present?
```
yoomlam, line 25: Other RRD jobs can split this into separate `query_lighthouse` and `assess_lighthouse_data` methods. For this class, since we want to stop querying if the assessed bp_readings is blank, we do the querying and assessing in a single method.

spec/jobs/rapid_ready_for_decision/form526_hypertension_job_spec.rb
```diff
@@ -0,0 +1,186 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require 'sidekiq/testing'
+
+RSpec.describe RapidReadyForDecision::Form526HypertensionJob, type: :worker do
```
yoomlam, line 6: This file is a copy of `DisabilityCompensationJob`'s rspec with updates only to the class name.
Running `diff spec/jobs/rapid_ready_for_decision/disability_compensation_job_spec.rb spec/jobs/rapid_ready_for_decision/form526_hypertension_job_spec.rb` will show only name changes.

app/workers/rapid_ready_for_decision/form526_base_job.rb
```diff
@@ -0,0 +1,107 @@
+# frozen_string_literal: true
+
+require 'lighthouse/veterans_health/client'
+require 'sidekiq/form526_job_status_tracker/job_tracker'
+require 'sidekiq/form526_job_status_tracker/metrics'
+
+module RapidReadyForDecision
+  class Form526BaseJob
```
yoomlam, line 8: Running `diff app/workers/rapid_ready_for_decision/form526_base_job.rb app/workers/rapid_ready_for_decision/disability_compensation_job.rb` shows that most of the code in this new `Form526BaseJob` class is copied from `DisabilityCompensationJob`.

spec/models/form526_submission_spec.rb
```diff
@@ -608,7 +608,10 @@
     context 'when the first name is NOT populated on the User' do
       before do
         # Ensure `subject` is called before stubbing `first_name` so that the auth headers are populated correctly
-        allow(User.find(subject.user_uuid)).to receive(:first_name).and_return nil
+        subject
+        user_with_nil_first_name = User.create(user)
+        allow(user_with_nil_first_name).to receive(:first_name).and_return nil
+        allow(User).to receive(:find).with(subject.user_uuid).and_return user_with_nil_first_name
```
yoomlam, line 614: When a `User` for the `subject.user_uuid` is requested, provide our stubbed `user_with_nil_first_name` User record.
The previous code only stubbed on a specific instance resulting from the `User.find(subject.user_uuid)` call, which can be different than the instance returned from the `User.find(subject.user_uuid)` in the `get_first_name` method, even though the instances refer to the same `User` record.

app/workers/rapid_ready_for_decision/form526_hypertension_job.rb
```diff
@@ -0,0 +1,51 @@
...
+
+      {
+        bp_readings: bp_readings,
+        medications: medications
+      }
+    end
+
+    def assess_bp_readings(bp_readings)
+      return RapidReadyForDecision::HypertensionObservationData.new(bp_readings).transform if bp_readings.present?
```
nanotone, line 34: This is only my personal preference, but because this line is pretty long on the eyes, and contains most of the moving parts, I find it more readable to split it up, short-circuit style

```ruby
return [] unless bp_readings.present?

RapidReadyForDecision::HypertensionObservationData.new(bp_readings).transform
```

YMMV and I'm fine with the code in its current form too.

nanotone (dismissed):
LGTM! I just had that one tiny suggestion.

app/workers/rapid_ready_for_decision/form526_hypertension_job.rb
yoomlam: That's much better! Thanks!

nanotone (approved):
👍

thilton-oddball (approved):

PR #9208, opened 2022-02-24
2402 get messages

kpethtel:
## Description of change
Adds new mobile endpoint for fetching appointment request messages by appointment request id.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va-mobile-app/issues/2402

This will be used by the appointment request details view. It does not implement caching at this time. Adding caching would require creating a model. Given the very limited scope of the data and the short life of most of our caches, it probably makes more sense to cache this in redux on the user's device. 

Does not yet include schema. That will be in a subsequent PR.


modules/mobile/config/routes.rb
```diff
@@ -40,7 +41,6 @@
     post '/user/phones', to: 'phones#create'
     put '/user/phones', to: 'phones#update'
     delete '/user/phones', to: 'phones#destroy'
-    get 'messaging/health/messages/signature', to: 'messages#signature'
```
kpethtel, line 43: This was redundant. The same route exists above.

modules/mobile/spec/request/appointment_request_messages_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    allow_any_instance_of(IAMUser).to receive(:icn).and_return('1012845331V153043')
+    iam_sign_in(build(:iam_user))
+    allow_any_instance_of(VAOS::UserService).to receive(:session).and_return('stubbed_token')
+  end
+
+  describe '#index' do
+    let(:request_id) { '8a4886886e4c8e22016e5bee49c30007' }
+
+    it 'returns 200 with message details upon success' do
+      VCR.use_cassette('vaos/messages/get_messages', match_requests_on: %i[method uri]) do
```
kpethtel, line 17: AFAICT, we always make copies of vcr cassettes from other modules. I haven't done that here because I wouldn't have changed anything about the cassettes. In other cases, I've made copies because I needed them to have different params or return a more limited data set. Given that the cassettes are directly tied to the vaos service we use for this endpoint, it makes sense to continue using their cassettes. I can change this if we would rather make our own copies as a policy.

modules/mobile/app/controllers/mobile/v0/appointment_request_messages_controller.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+require_dependency 'mobile/application_controller'
+
+module Mobile
+  module V0
+    class AppointmentRequestMessagesController < ApplicationController
```
kpethtel, line 7: I named it this because we already have a messages controller that does something else.

modules/mobile/spec/request/appointment_request_messages_spec.rb
kreek: I would make a copy as well in this case just so we're in control of the resources for our specs.

modules/mobile/spec/request/appointment_request_messages_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+              'attributes' => {
+                'messageText' => 'Testing',
+                'messageDateTime' => '11/11/2019 12:26:13',
+                'appointmentRequestId' => '8a4886886e4c8e22016e5bee49c30007',
+                'date' => '2019-11-11T12:26:13.931+0000'
+              }
+            }
+          ]
+        }
+        expect(response.status).to eq(200)
```
kreek, line 34: Could we add a JSON schema check here too? It's one way we check that the shared service hasn't changed out from under us.

kpethtel: Yup, done.

kreek (approved):
LGTM 🚀

modules/mobile/spec/request/appointment_request_messages_spec.rb
kpethtel: I wasn't sure if we do this by policy or not. Sounds like it's policy. Thanks for clarifying.

kreek (approved):

PR #9213, opened 2022-02-24
LGY Update COE Application

kathleencrawford:
## Description of change

This PR is an update to the LGY COE application (Loan Guaranty Certificate of Eligibility).  There is a mismatch between the form data being passed in from our front end and the values expected by LGY for military branch and service type.  

For military branch, LGY expects `[ARMY, NAVY, MARINES, AIR_FORCE, COAST_GUARD, OTHER]` and then has another field for service type which expects `[ACTIVE_DUTY, RESERVE_NATIONAL_GUARD]`.

On our form, we have the following values: `[Air Force, Air Force Reserve, Air National Guard, Army, Army National Guard, Army Reserve, Coast Guard, Coast Guard Reserve, Marine Corps, Marine Corps Reserve, Navy, Navy Reserve]`.

This PR just adds some logic to split the military branch that is being passed in and sending it as the appropriate military branch and service type.  

In addition, it removes a check for certain error codes so that an error is displayed on the screen if the form submission was not successful.  LGY often sends us 404 responses back which are valid responses, but in this case we want to inform the user if the application submission failed.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#36896

This feature is not launched in production yet and is only being tested in staging.  This is just an update to the form submission so no new settings/logging have been added.

- [x] Local testing
- [x] Staging testing planned


app/models/saved_claim/coe_claim.rb
```diff
@@ -95,10 +95,26 @@ def relevant_prior_loans(form_copy)
...
     parsed_form['periodsOfService'].each do |service_info|
+      # values from the FE for military_branch are:
+      # ["Air Force", "Air Force Reserve", "Air National Guard", "Army", "Army National Guard", "Army Reserve",
+      # "Coast Guard", "Coast Guard Reserve", "Marine Corps", "Marine Corps Reserve", "Navy", "Navy Reserve"]
+      # these need to be formatted because LGY only accepts [ARMY, NAVY, MARINES, AIR_FORCE, COAST_GUARD, OTHER]
+      # and then we have to pass in ACTIVE_DUTY or RESERVE_NATIONAL_GUARD for service_type
+      military_branch = service_info['militaryBranch'].parameterize.underscore.upcase
+      service_type = 'ACTIVE_DUTY'
+
+      if military_branch.include?('NATIONAL_GUARD') || military_branch.include?('RESERVE')
```
LindseySaari, line 106:
```suggestion
      if ['RESERVE', 'NATIONAL_GUARD'].include?(military_branch)
```

This may be. a cleaner way to write the conditional

kathleencrawford: Thanks Lindsey!  I made the suggested change but then rubocop suggested I use %w instead of string literals.  Thank you for noticing - that does look cleaner.

kathleencrawford: Actually Lindsey, I had to change the logic again... the military branch is the longer string (ex. AIR_NATIONAL_GUARD) so the .include? fails when we had it that way initially.  I made some changes - let me know what you think.  Thanks!

kathleencrawford:
One more change @LindseySaari, our FE changed the schema so instead of `militaryBranch` it is now `serviceBranch`.

ericboehs (approved):

PR #9214, opened 2022-02-25
Update appointment caching to allow caching back to beginning of last…

aherzberg:
… year

## Description of change

Update appointment caching to allow caching back to beginning of last year. This is to be consistent with front end date filter which filter by calendar year. before this change, if user was to filter by all of last calendar year, none of those appointment would have been cached. 

Kept default start date to still be one_year_ago. 

## Original issue(s)
https://github.com/department-of-veterans-affairs/va-mobile-app/issues/2581


modules/mobile/spec/request/appointments_list_request_spec.rb
```diff
@@ -91,35 +91,45 @@
...
+      va_appointments = Mobile::V0::Adapters::VAAppointments.new.parse(
+        JSON.parse(va_json, symbolize_names: true)
+      )
+      cc_appointments = Mobile::V0::Adapters::CommunityCareAppointments.new.parse(
+        JSON.parse(cc_json, symbolize_names: true)
+      )
+
+      appointments = (va_appointments + cc_appointments).sort_by(&:start_date_utc)
+
+      let(:start_date) { (Time.now.utc.beginning_of_year - 1.year).iso8601 }
```
aherzberg, line 107: Update caching specs to be at beginning of previous year but this is just implicitly testing the start date limit. Let me know if you think this should be explicitly tested.

modules/mobile/app/controllers/mobile/v0/appointments_controller.rb
```diff
@@ -49,14 +49,18 @@ def cancel
 
       def use_cache?(validated_params)
         # use cache if date range is within +/- 1 year and use_cache is true
-        validated_params[:start_date] >= one_year_ago.iso8601 &&
+        validated_params[:start_date] >= beginning_of_last_year.iso8601 &&
```
kpethtel, line 52: I think this PR needs to get rid of one_year_ago entirely. All uses of it should be replaced with beginning_of_last_year. As it's written now, it will fetch from the cache if the start date is >= beginning of last year, but it's not actually fetching records from that time, so the cache won't include records starting at the beginning of last year. The other two places it's used are: 1) setting the default start date value when none is provided by the client 2) indicating the topmost range of the start date we allow to be passed into the proxy to ensure that the cache remains valid for the default range.

modules/mobile/spec/request/appointments_list_request_spec.rb
kpethtel: I asked myself the same question when I was working on appointment requests and I came to the same conclusion that the start date is being implicitly tested by VCR. However, given that the change here is specific to the date ranges and (AFAICT from a quick search through this file) we don't seem to explicitly test the default date setting behavior, it might be worth a more explicit testing approach. Maybe something like:
```
context 'when start date is after the beginning of the previous year' do
  it 'defaults start date to the beginning of the previous year' do
    expect(Faraday).to have_received(:get).with(url, beginning_of_year)
  end
end
```
That's an oversimplification, but you get the idea. A slightly easier way would be to do the same with the proxy, but this is a less valuable test because these are integration tests that also test the proxy, which does not have its own unit tests. 
```
expect_any_instance_of(Mobile::V0::Appointments::Proxy).to receive(:get_appointments).with(start_date: beginning_of_last_year)
```
I'm happy to pair with you on this.

modules/mobile/spec/support/fixtures/cc_appointments.json
```diff
@@ -23,7 +23,7 @@
             "zip_code": "32826"
           },
           "instructions_to_veteran": "Please arrive 15 minutes ahead of appointment.",
-          "appointment_time": "04/25/2020 13:30:00",
```
kpethtel, line 26: I like the boundary testing here.

kpethtel (changes requested):
Unfortunately, this will be a little more involved. Once you've made the changes I'm suggesting in the controller, you'll also have to change the vcr URIs to match the new date search range. I'm happy to pair with you on this if you want some help, but I think it's mostly just going to be tedious spec fixes. VCR is a super handy tool, but this is where it becomes brittle. This was a pain for me while working on appointment requests as well.

modules/mobile/app/controllers/mobile/v0/appointments_controller.rb
aherzberg: I was thinking to avoid changing default behavior as to not make the default more expensive then is likely needed to be for the user going on the appointment page without date filters. As it is now, as long as the start and end date are set in params, they can use the cache as early as beginning of last year. Thinking about it more, even if we do get more appointments by broadening the default window, it should only be marginally more expensive since we do paginate. Make the code cleaner and more consistent too.

kpethtel: You're right that it'll make the call slightly more expensive. And it's possible that for a power appointments user like judy it might even start causing timeouts. In which case we'd have to revisit the caching strategy. But it's also worth remembering that the app pre-caches appointments when a user logs in, so it shouldn't impact the user most of the time. Speaking of which... I just realized that you also need to update the start date in `modules/mobile/app/workers/mobile/v0/pre_cache_appointments_job.rb`.

But the point I'm making above is that the cache will not be accurate if you don't change all instances of the start date.

aherzberg: I see your point now. I updated the default start date to match as well as the start dates of the VCR cassettes. Still working on additional specs to explicitly test the default start date

modules/mobile/spec/request/appointments_list_request_spec.rb
```diff
@@ -39,7 +39,18 @@
         end
       end
 
-      it 'defaults to a range of -1 year and +1 year' do
+      it 'date range defaults to the beginning of the previous year to one year from now' do
+        expect_any_instance_of(Mobile::V0::Appointments::Service).to receive(:fetch_va_appointments).with(
```
kpethtel, line 43: I think we could get more out of these tests if we combine this context and the `with an invalid date in params` context below into a single describe block. 
```
describe 'start and end date' do
  context 'when omitted from query params'
    it 'defaults to beginning of the previous year to one year from now'
    
  context 'when provided start date is after beginning of the previous year'
    it 'defaults to the beginning of the the previous year'
    
  context 'when provided start date is before the beginning of the previous year'
    it 'uses the provided start date'
    
  context 'when provided end date is before one year from now'
    it 'defaults to one year from now'
    
  context 'when provided end date is after one year from now'
    it 'uses the provided end date'
  
  context 'when provided dates are invalid'
    it 'returns a bad request error'
   
end
```

modules/mobile/spec/request/appointments_list_request_spec.rb
```diff
@@ -296,31 +317,48 @@
         let(:cc_appointment_with_provider) { response.parsed_body['data'][2]['attributes'] }
         let(:covid_appointment) { response.parsed_body['data'][6]['attributes'] }
 
+        it 'date range defaults to the beginning of the previous year to one year from now' do
```
kpethtel, line 320: If you choose to follow my suggestion above to move these tests into a describe, this can go away.

kpethtel (changes requested):
This looks good, but I added a few comments about the specs. The only one that's a hard blocker is the test that is now broken due to moving the action to the before block. The rest are just suggestions.

modules/mobile/spec/request/appointments_list_request_spec.rb
```diff
@@ -91,35 +102,45 @@
...
-        appointments = (va_appointments + cc_appointments).sort_by(&:start_date_utc)
         Mobile::V0::Appointment.set_cached(user, appointments)
+        get '/mobile/v0/appointments', headers: iam_headers, params: params
       end
 
       after { Timecop.return }
 
       it 'retrieves the cached appointments rather than hitting the service' do
         expect_any_instance_of(VAOS::AppointmentService).not_to receive(:get_appointments)
-        get '/mobile/v0/appointments', headers: iam_headers, params: params
```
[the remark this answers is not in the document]
aherzberg: Good catch. Now fixed.

modules/mobile/spec/request/appointments_list_request_spec.rb
aherzberg: You're right. This is much cleaner. I've made this change.

modules/mobile/spec/request/appointments_list_request_spec.rb
```diff
@@ -28,6 +28,161 @@
...
     after { Timecop.return }
 
+    describe 'start and end date' do
+      let(:beginning_of_last_year) { (DateTime.now.utc.beginning_of_year - 1.year) }
+      let(:one_year_from_now) { (DateTime.now.utc.beginning_of_day + 1.year) }
+
+      context 'when omitted from query params' do
+        let(:params) { { page: { number: 1, size: 10 }, useCache: false } }
+        let(:request_appointments) do
+          VCR.use_cassette('appointments/get_facilities', match_requests_on: %i[method uri]) do
```
kpethtel, line 38: FYI, you don't need to do the VCR setup here because the code will stop executing at the watched method and never get to the VCR. Also, we're not testing anything in the response. You can confirm this by adding a test for response.status == 200. It will fail. If you add `.and_call_original` to the end of your `expect_any_instance_of` lines, then you'd get a status of 200 because the rest of your code will execute and use the VCRs.

kpethtel (approved):
Looks good. Thanks for improving out testing around start and end date. I added one comment about some unnecessary test setup. It's not really a blocker but it's a good thing to be aware of.

modules/mobile/spec/request/appointments_list_request_spec.rb
aherzberg: Thanks for catching that. Force of habit. I've removed the VCRs in those specs.

kpethtel (approved):
🚀

PR #9217, opened 2022-02-25
API-13369: Only update VBMS eFolder access once a POA form is successfully submitted

hinzed1127:
## Description of change

Similarly to #9193, this moves a sidekiq job _within_ the jobs uploading 21-22/21-22a forms to VBMS, ensuring that eFolder access is not granted until the POA code and form calls have successfully been made

## Original issue(s)
https://vajira.max.gov/browse/API-13369

Note that this PR is currently set to merge into [dh-api-13467](https://github.com/department-of-veterans-affairs/vets-api/tree/dh-api-13467) and **not** master. This is to simplify the diff to show the actual changes being made here, as well as giving you the option to merge it all together as a single change, if desired (cc @jw81 )


modules/claims_api/app/workers/claims_api/poa_form_builder_job.rb
```diff
@@ -16,14 +16,20 @@ class PoaFormBuilderJob
     # it queues a job to update the POA code in BGS, as well.
     #
     # @param power_of_attorney_id [String] Unique identifier of the submitted POA
-    def perform(power_of_attorney_id)
+    def perform(power_of_attorney_id, participant_id)
```
jw81, line 19: Unfortunately, this job is referenced in other places and those references would also need to be updated to account for this new arg.  But I think there might be an easier way of getting access to this value without needing to pass a new arg.  Since you're OOO, I'm going to try my hand at it.

modules/claims_api/app/workers/claims_api/vbms_upload_job.rb
```diff
@@ -13,13 +13,21 @@ class VBMSUploadJob
     # If successfully uploaded, it queues a job to update the POA code in BGS, as well.
     #
     # @param power_of_attorney_id [String] Unique identifier of the submitted POA
-    def perform(power_of_attorney_id)
+    def perform(power_of_attorney_id, participant_id)
```
jw81, line 16: Unfortunately, this job is referenced in other places and those references would also need to be updated to account for this new arg.  But I think there might be an easier way of getting access to this value without needing to pass a new arg.  Since you're OOO, I'm going to try my hand at it.

jw81 (changes requested):
Just doing this so it doesn't accidentally get merged too soon.

modules/claims_api/app/workers/claims_api/poa_form_builder_job.rb
jw81: [Here is the PR](https://github.com/department-of-veterans-affairs/vets-api/pull/9232) that I've proposed.

modules/claims_api/app/workers/claims_api/vbms_upload_job.rb
jw81: [Here is the PR](https://github.com/department-of-veterans-affairs/vets-api/pull/9232) that I've proposed.

hinzed1127:
closing in favor of https://github.com/department-of-veterans-affairs/vets-api/pull/9239

hinzed1127:
closing in favor of https://github.com/department-of-veterans-affairs/vets-api/pull/9239

PR #9225, opened 2022-02-25
2646 mobile update maintenance window service map

kreek:
## Description of change
Adds an entry for bgs as a dependency of payment_history in the maintenance window service map.

## Original issue(s)
department-of-veterans-affairs/va-mobile-app/issues/2646


kpethtel (approved):
👍

PR #9226, opened 2022-02-25
2440 appointment request schema

kpethtel:
## Description of change
Updates schema documentation for appointments to reflect recent appointment request work. 

## Original issue(s)
department-of-veterans-affairs/va-mobile-team#2440

Updates docs.


modules/mobile/docs/examples/appointments/cc_appointment.yml
```diff
@@ -5,8 +5,6 @@ data:
       appointmentType: COMMUNITY_CARE
       cancelId: null
       comment: Please arrive 20 minutes before the start of your appointment
-      clinicId: ~
```
kpethtel, line 8: These do not exist in the serializer, so I'm not sure why they're here.

modules/mobile/docs/examples/appointments/cc_appointment_request.yml
```diff
@@ -0,0 +1,54 @@
...
+      minutesDuration: null
+      phoneOnly: false
+      startDateLocal: 2020-11-02T01:00:00.000-07:00
+      startDateUtc: 2020-11-02T08:00:00.000Z
+      status: SUBMITTED
+      statusDetail: null
+      timeZone: America/Denver
+      vetextId: null
+      reason: 'New Issue'
+      is_covid_vaccine: null
```
kpethtel, line 35: Working on the docs made me realize this should really be false. I initially set it to nil because I thought there would be cases where it would be true, but I've since learned that's probably not so. It doesn't seem to be a problem for the front end, but I can change this to false in a separate PR.

modules/mobile/docs/schemas/Appointment.yml
```diff
@@ -34,7 +34,7 @@ properties:
     properties:
       appointmentType:
         type: string
-        enum: [ COMMUNITY_CARE, VA, VA_VIDEO_CONNECT_ATLAS, VA_VIDEO_CONNECT_HOME, VA_VIDEO_CONNECT_GFE, VA_VIDEO_CONNECT_ONSITE ]
```
kpethtel, line 37: This is no longer an option in the model.

jperk51 (approved):
LGTM

PR #9230, opened 2022-02-25
Form 526: Handle nil User Redis record when calling get_first_name

yoomlam:
## Description of change

Handle nil `User` Redis record when calling `get_first_name`. Resort to using the `auth_headers`, similar to the `full_name` method.

Adds code comments to alert developers of attempting to use a `User` record that has been deleted from Redis.

## Original issue(s)
As a result of [this Slack thread](https://dsva.slack.com/archives/C01BVF2A3V0/p1645818861329309?thread_ts=1645807137.696329&cid=C01BVF2A3V0), there's a potential bug for RedisStore records like `User` if the record is queried after the record is expired and deleted (after the TTL (time to live)).


app/models/form526_submission.rb
```diff
@@ -373,6 +374,8 @@ def submit_form_8940
 
   def submit_flashes
     user = User.find(user_uuid)
+    # Note that the User record is stored in Redis temporarily, for 30 minutes: User.redis_namespace_ttl => 1800
+    # If this method runs after the TTL, then there is a possibility the flashes will not be applied
     BGS::FlashUpdater.perform_async(id) if user && Flipper.enabled?(:disability_compensation_flashes, user)
```
yoomlam, line 379: @michelpmcdonald Do you agree with this code comment?

app/models/form526_submission.rb
```diff
@@ -126,6 +126,7 @@ def submit_with_birls_id_that_hasnt_been_tried_yet!(
     log_exception_to_sentry e, extra_content_for_sentry
   end
 
+  # Note that the User record is stored in Redis temporarily, for 30 minutes: User.redis_namespace_ttl => 1800
   def get_first_name
     user = User.find(user_uuid)
     user&.first_name&.upcase
```
yoomlam, line 132: Can we update this method to use the auth_headers like the `full_name` method?

app/models/form526_submission.rb
```diff
@@ -126,9 +126,14 @@ def submit_with_birls_id_that_hasnt_been_tried_yet!(
...
 
+  # Note that the User record is stored in Redis temporarily, for 30 minutes: User.redis_namespace_ttl => 1800
+  # Checks against the User record first, and then resorts to checking the auth_headers
+  # for the name attribute if the User record doesn't exist or doesn't contain the first name
   def get_first_name
     user = User.find(user_uuid)
-    user&.first_name&.upcase
+    return user&.first_name&.upcase if user&.first_name.present?
+
+    auth_headers&.dig('va_eauth_firstName')&.upcase
```
yoomlam, line 136: Mimicking the `full_name` method, use the `auth_headers` in case `user` or `user.first_name` is nil

julialeague (approved):
Looks good to me!

app/models/form526_submission.rb
yoomlam: @annaswims (since you added this method in #7732) When the job retries several times, there's a possibility that this method is executed but the `user` record is no longer available in Redis. Do we still need to check the Flipper flag for particular users?

app/models/form526_submission.rb
allthesignals: Why not just use auth_headers since it's more reliable? Is it because it's not the canonical source?

yoomlam: Not sure, but I don't want to make a change that conflicts with current behaviors that may expect info from the `user` record, which is created upon [user session creation](https://dsva.slack.com/archives/C01BVF2A3V0/p1645818462607279?thread_ts=1645807137.696329&cid=C01BVF2A3V0). Just playing it safe.

michelpmcdonald: I have no idea why we don't use the name from the auth_headers all the time......I do know we have issues with names when submitting claims.

Originally, get_first_name was called within the scope of the web user's submit 526 call, at this point the code is synchronous, the user has to be logged in to make the submit call, so the calling user has a current session, and User.find should work.

However, after the fasttrack changes, get_first_name is called in a method in a place that is outside of the users initial 526 submit call.  Most of the time, it's well within the redis 30 minute eject time, but it is not guaranteed since now the call is behind the asynchronous fasttrack sidekiq job.

All that being said, this new version of get_first_name is a good idea since it is now possible that this method get called after the user session has expired.

app/models/form526_submission.rb
michelpmcdonald: @annaswims sadly is long gone....

Since you asked(I'm not personally too picky about comments), not in love with the comment, I think the code without the comment already clearly shows that flashes may not be applied if user is nil, and following through to the User object, it's pretty obvious that it's a reddis backed model.  Also, since the ttl is a setting, the actual value(1800) should not, IMHO, be enshrined in a comment...

I do think this code might be able to be improved, I'm unsure of the implications of not sending a flash, but since this code is now behind(time wise) the sidekiq fasttrack job, we now have more exposure to the Redis User model being ejected from the cache.  To harden this up, I'd look to see what info "flash" is using from user, and see if that info can be pulled from the auth_headers instead, but that is not part of the scope of this ticket...

yoomlam: Since this looked like a bug, I wanted to make the behavior explicit in the comments. Also it took me a good amount of time to connect the dots, so a reminder comment seemed useful. I can reword them so the configuration values are not hard-coded.

The implication of not sending a flash is the claim is not prioritized like it should be.

michelpmcdonald (approved):
looks good

app/models/form526_submission.rb
```diff
@@ -126,9 +126,12 @@ def submit_with_birls_id_that_hasnt_been_tried_yet!(
     log_exception_to_sentry e, extra_content_for_sentry
   end
 
+  # Note that the User record is stored in Redis temporarily -- `User.redis_namespace_ttl`
```
ericboehs, line 129: This comment makes it sounds like we plan to change this. I don't believe we do. Maybe the word "cache" would be better here.

app/models/form526_submission.rb
```diff
@@ -126,9 +126,12 @@ def submit_with_birls_id_that_hasnt_been_tried_yet!(
...
   def get_first_name
     user = User.find(user_uuid)
-    user&.first_name&.upcase
+    return user&.first_name&.upcase if user&.first_name.present?
+
+    auth_headers&.dig('va_eauth_firstName')&.upcase
```
ericboehs, line 134: Another way to write this:

```suggestion
    user&.first_name&.upcase.presence ||
      auth_headers&.dig('va_eauth_firstName')&.upcase
```

app/models/form526_submission.rb
```diff
@@ -373,6 +376,8 @@ def submit_form_8940
 
   def submit_flashes
     user = User.find(user_uuid)
+    # Note that the User record is stored in Redis temporarily -- `User.redis_namespace_ttl`
+    # If this method runs after the TTL, then the flashes will not be applied -- a possible bug.
```
ericboehs, line 380: Does this happen? Seems like if a user is active on the site, they'd have a user in Redis and if they weren't on the site, they wouldn't need a flash message.

ericboehs (approved):

app/models/form526_submission.rb
yoomlam: Updated.

app/models/form526_submission.rb
yoomlam: If the job fails and retries, it could take days for the job to eventually succeed. By that time, the `User` record could be expired.

ericboehs (approved):

app/models/form526_submission.rb
yoomlam: AFAIK, it doesn't currently happen, but it is possible since the current TTL is 30 minutes and Sidekiq jobs can retry 8 times with longer delays with each subsequent retry.

PR #9231, opened 2022-02-25
Refactor Hypertension PDF generator to move view logic into an ERB template

allthesignals:
## Description of change
This PR moves the Hypertension PDF generator "view logic" into an ERB template. Instead of calling Prawn methods directly, this branch uses a [gem](https://github.com/puzzle/prawn-markup) that supports rudimentary HTML features. Some basic business logic is kept in the main class instead.

This change reads in the ERB template and calls the Prawn method "markup" enabled by the newly-added gem:
https://github.com/department-of-veterans-affairs/vets-api/blob/5f5aa2bc449092b9eaccb53dc4b7b1972e226db8/app/services/rapid_ready_for_decision/hypertension_pdf_generator.rb#L30-L36

There are definitely some tradeoffs with this approach: while it's (in my opinion) a bit easier to reason about/read/maintain, there is less flexibility in some ways ([cannot change text color](https://github.com/puzzle/prawn-markup/issues/15), table configuration options are limited). It might be worth contributing to prawn-markup if we want to get more out of it, but I'm not sure how important nuanced styling is for these documents :) 

 - Adds a new Prawn gem to enable basic HTML markup.
 - This is a not a 1:1 identical conversation in terms of the PDF output. There are some spacing issues and some text color cannot be changed.
 - Some spacing hacks are needed like `<p> <br /> </p>` — this does not behave exactly like an HTML document but comes close.


app/services/rapid_ready_for_decision/hypertension_pdf_generator.rb
```diff
@@ -8,17 +8,29 @@ def initialize(patient, blood_pressure_data, medications)
...
+        },
+        table: {
+          width: 150,
+          cell: {
+            size: 10,
+            border_width: 0,
+            background_color: 'f3f3f3'
+          }
+        }
+      }
```
allthesignals, line 27: Some style globals — this is like a heavy-handed CSS file

yoomlam:
> There are definitely some tradeoffs with this approach: while it's (in my opinion) a bit easier to reason about/read/maintain, there is less flexibility in some ways (https://github.com/puzzle/prawn-markup/issues/15, table configuration options are limited).

Check with designers to see if any of the limitations would even be used by us. 

> Some spacing hacks are needed like <p> <br /> </p> — this does not behave exactly like an HTML document but comes close.

We'll definitely want to make spacing adjustments in the pdf. Try using [some css](https://www.w3schools.com/css/css_text_spacing.asp) to adjust the spacing instead.

app/services/rapid_ready_for_decision/hypertension_pdf_generator.rb
```diff
@@ -8,17 +8,29 @@ def initialize(patient, blood_pressure_data, medications)
...
-      add_intro
-      add_blood_pressure_intro
-      add_blood_pressure_list
-      add_blood_pressure_outro
-      add_medications_intro
-      add_medications_list
-      add_about
+      template = File.read('app/services/rapid_ready_for_decision/views/hypertension_pdf.erb')
+
+      @pdf.markup ERB.new(template).result(binding)
```
yoomlam, line 33: What Rails magic 🪄 is this? 
* `ERB.new(template)` loads the file as an ERB object
* What does `result(binding)` do?

app/services/rapid_ready_for_decision/hypertension_pdf_generator.rb
```diff
@@ -33,174 +45,25 @@ def medications?
       @medications.any?
     end
 
-    def add_intro
+    def patient_name
       names = @patient.with_indifferent_access
```
yoomlam, line 49: :TIL: https://apidock.com/rails/ActiveSupport/HashWithIndifferentAccess

app/services/rapid_ready_for_decision/hypertension_pdf_generator.rb
allthesignals: > What Rails magic 🪄 is this?
> 
> * `ERB.new(template)` loads the file as an ERB object
> * What does `result(binding)` do? I presume it binds the instance variables (e.g., `@blood_pressure_data`) to the context of the new ERB object. Is this how the `prawn-markup` gem should be used?

Yeah, I was surprised, too. This actually takes a cue from how mailer templates are handled ([for example](https://github.com/department-of-veterans-affairs/vets-api/blob/53db8ed5d7a8b2eab2552f1edd7081ade7ae5d36/app/mailers/hypertension_fast_track_pilot_mailer.rb#L19)). That's mostly correct: it's not just instance variables, but the entire class context... @nanotone and I were looking through it the other day. Think of it as `this` in JavaScript... 

All `#result` does is generate a plain string that is passed to the #markup method. As for whether this is how `prawn-markup` _should_ be used, I would argue that it seems valid: we can think of ERB as a more advanced and feature-rich version of multiline strings. 

That said, I'm not married to this — there are definite trade-offs. I'd be happy to investigate another approach OR see if I can minimize those trade-offs even more.

allthesignals:
Original:
[rrd-hypertension-original.pdf](https://github.com/department-of-veterans-affairs/vets-api/files/8155072/rrd-hypertension-original.pdf)

Proposed:
[rrd-hypertension-proposed.pdf](https://github.com/department-of-veterans-affairs/vets-api/files/8155073/rrd-hypertension-proposed.pdf)

app/services/rapid_ready_for_decision/hypertension_pdf_generator.rb
allthesignals: Any theories as to why? Seems it's always keyed with symbols...

yoomlam: It's possible that `patient` is keyed by strings (or maybe a mix of strings and symbols), and this class uses symbols exclusively.

app/services/rapid_ready_for_decision/views/hypertension_pdf.erb
```diff
@@ -0,0 +1,139 @@
+Hypertension Rapid Ready for Decision | Claim for Increase <br />
```
yoomlam, line 1: Perhaps add a comment to remind developers that the html won't be interpreted/rendered strictly as html.

app/services/rapid_ready_for_decision/views/hypertension_pdf.erb
```diff
@@ -0,0 +1,139 @@
...
+<h1>
+  VHA Hypertension Data Summary for <br/>
```
yoomlam, line 3: Do you want to start generalizing this template, e.g.:
```suggestion
  VHA <%= disability_title %> Data Summary for <br/>
```
Also rename the file to something like `rrd_data_summary_pdf.erb` or something that matches the [`VAMC_Hypertension_Rapid_Decision_Evidence.pdf` filename](https://github.com/department-of-veterans-affairs/vets-api/blob/113697926ea68cbf3c8c638cd36064c1e05a20c0/app/services/rapid_ready_for_decision/hypertension_upload_manager.rb#L15).

app/services/rapid_ready_for_decision/views/hypertension_pdf.erb
```diff
@@ -0,0 +1,139 @@
...
+
+  <p>
+    <b>Blood pressure: <%= systolic %>/<%= diastolic %></b> <br />
+    Taken on: <%= bp[:effectiveDateTime].to_date.strftime('%m/%d/%Y') %>
+      at <%= Time.iso8601(bp[:effectiveDateTime]).strftime('%H:%M %Z') %> <br />
+    Location: <%= bp[:organization] || 'Unknown' %>
+  </p>
+<% end %>
+
+<h4>Hypertension Rating Schedule</h4>
```
yoomlam, line 39: Remember to wrap this in an `if` block if you decide to generalize this template.

PR #9233, opened 2022-02-25
API-12854 Add timeliness exemption reason page to pdf

cilestin:
## Description of change
Adds an additional page for the veteran to indicate their timeliness exemption reason in the NODv2 pdf

## Original issue(s)
https://vajira.max.gov/browse/API-12854


ElleeB:
This makes me wonder if we should add something near the 'additional sheets' indicating the number of additional sheets, or number the additional sheets 1/2, 2/2? Maybe overkill, but just a thought.

ElleeB (approved):

PR #9235, opened 2022-02-25
[37201] Adding ability to create, encrypt, and decrypt opaque refresh…

bosawt:
… token for sign in

## Description of change
This PR adds a non-persistent model, `RefreshToken`, and an encryption/decryption functionality for this token

## Original issue(s)
department-of-veterans-affairs/va.gov-team#37201

- There are no interactions with prod for this PR
- To test, you'll have to create a `RefreshToken` in `Rails Console` -> `SignIn::RefreshToken.new`, defining at least `session_handle`, `user_uuid`, and `anti_csrf_token`
- Then, you can encrypt this token with `SignIn::RefreshTokenEncryptor.new(refresh_token: refresh_token).perform`
- You can decrypt with `SignIn::RefreshTokenDecryptor.new(encrypted_refresh_token: encrypted_refresh_token).perform`


app/services/sign_in/refresh_token_decryptor.rb
```diff
@@ -0,0 +1,59 @@
...
+      version = get_version_part(split_refresh_token)
+
+      validate_token!(decrypted_part, nonce, version)
+
+      SignIn::RefreshToken.new(session_handle: decrypted_part.session_handle,
+                               user_uuid: decrypted_part.user_uuid,
+                               parent_refresh_token_hash: decrypted_part.parent_refresh_token_hash,
+                               anti_csrf_token: decrypted_part.anti_csrf_token,
+                               nonce: decrypted_part.nonce,
+                               version: decrypted_part.version)
```
ericboehs, line 25: Nit:

```suggestion
      SignIn::RefreshToken.new(
        session_handle: decrypted_part.session_handle,
        user_uuid: decrypted_part.user_uuid,
        parent_refresh_token_hash: decrypted_part.parent_refresh_token_hash,
        anti_csrf_token: decrypted_part.anti_csrf_token,
        nonce: decrypted_part.nonce,
        version: decrypted_part.version
      )
```

app/services/sign_in/refresh_token_decryptor.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    private
+
+    def validate_token!(decrypted_token, nonce, version)
+      raise SignIn::Errors::RefreshVersionMismatchError if decrypted_token.version != version
```
ericboehs, line 31:
```suggestion
      raise SignIn::Errors::RefreshVersionMismatchError unless decrypted_token.version == version
```
?

app/services/sign_in/refresh_token_decryptor.rb
```diff
@@ -0,0 +1,59 @@
...
+    def initialize(encrypted_refresh_token:)
+      @encrypted_refresh_token = encrypted_refresh_token
+    end
+
+    def perform
+      split_refresh_token = split_encrypted_refresh_token(encrypted_refresh_token)
+
+      decrypted_part = get_decrypted_part(split_refresh_token)
+      nonce = get_nonce_part(split_refresh_token)
+      version = get_version_part(split_refresh_token)
```
ericboehs, line 16: Thoughts on making `split_refresh_token` a private method that `get_decrypted_part` calls directly (instead of being passed)?

`get_` could be removed from the methods and these locals would go away and `validate_token!` would call the private methods instead.

app/services/sign_in/refresh_token_decryptor.rb
```diff
@@ -0,0 +1,59 @@
...
+    def get_nonce_part(split_refresh_token)
+      split_refresh_token[SignIn::Constants::REFRESH_TOKEN_NONCE_POSITION]
+    end
+
+    def get_version_part(split_refresh_token)
+      split_refresh_token[SignIn::Constants::REFRESH_TOKEN_VERSION_POSITION]
+    end
+
+    def split_encrypted_refresh_token(encrypted_refresh_token)
+      encrypted_refresh_token.split('.')
```
ericboehs, line 48:
```suggestion
      encrypted_refresh_token.split('.', 3)
```

Should splitting be limited to three parts?

image

app/services/sign_in/refresh_token_decryptor.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    def decrypt_refresh_token(encrypted_refresh_token)
+      message_encryptor.decrypt_and_verify(encrypted_refresh_token)
+    end
+
+    def message_encryptor
+      ActiveSupport::MessageEncryptor.new(Rails.application.secrets.secret_key_base[0..31])
```
ericboehs, line 56: Should we be using Lockbox for encryption here so that the key gets rotated appropriately?

app/services/sign_in/refresh_token_decryptor.rb
```diff
@@ -0,0 +1,59 @@
...
+    def get_decrypted_part(split_refresh_token)
+      decrypt_refresh_token(split_refresh_token[SignIn::Constants::REFRESH_TOKEN_ENCRYPTED_POSITION])
+    end
+
+    def get_nonce_part(split_refresh_token)
+      split_refresh_token[SignIn::Constants::REFRESH_TOKEN_NONCE_POSITION]
+    end
+
+    def get_version_part(split_refresh_token)
+      split_refresh_token[SignIn::Constants::REFRESH_TOKEN_VERSION_POSITION]
```
ericboehs, line 44: Could you describe the benefit of making these consants?

For example, these methods could be removed in favor of assuming their position:

```ruby
token, nonce, version = encrypted_refresh_token.split '.'
```

And likewise for building:

```ruby
encrypted_refresh_token = [token, nonce, version].join '.'
```

app/services/sign_in/refresh_token_decryptor.rb
bosawt: 👍

app/services/sign_in/refresh_token_decryptor.rb
bosawt: I made these constants because the positioning of the components is kinda subtle, and since there are two objects that have to separately infer the locations of these components, I was trying to store that in a shared location (array locations as constants).

For example, if we decide that we want to change the order of these components or add a field for a V1 version of this, we just have to change the constants, or we can have constants that describe the makeup of a V0 vs V1 refresh token, instead of that info being hidden in the encryptor/decryptor logic

app/services/sign_in/refresh_token_decryptor.rb
bosawt: 👍

app/services/sign_in/refresh_token_decryptor.rb
bosawt: Only reason I hesitate is to keep the ability to split versatile, with the ability to change the number of fields if necessary with different versions

bosawt: You know, this gives me an idea. Maybe it's appropriate to create another model: `EncryptedRefreshToken` that incapsulates this logic. Would probably make this more straightforward

bosawt: I formatted the constants a bit differently to give me an array length of these elements, so I can use that as a way to limiting the splitting and remain versatile

app/services/sign_in/refresh_token_decryptor.rb
bosawt: I reorganized this a bit to hopefully make this appear simpler

app/services/sign_in/refresh_token_decryptor.rb
bosawt: Updated this to use KMSEncrypted for encrypting/decrypting

bramleyjl (approved):
LGTM, token creation, encryption, and decryption all worked without issue.

bosawt:
@ericboehs Should be reviewable again, just had to merge in master to fix an unrelated broken spec

app/services/sign_in/refresh_token_decryptor.rb
```diff
@@ -0,0 +1,62 @@
...
+module SignIn
+  class RefreshTokenDecryptor
+    attr_reader :split_token_array
+
+    def initialize(encrypted_refresh_token:)
+      @split_token_array = split_encrypted_refresh_token(encrypted_refresh_token)
+    end
+
+    def perform
+      decrypted_component = get_decrypted_component
```
ericboehs, line 12: Why not just name the method `decrypted_component`?

app/services/sign_in/refresh_token_decryptor.rb
```diff
@@ -0,0 +1,62 @@
...
+    def decrypt_refresh_token(encrypted_part)
+      message_encryptor.decrypt(encrypted_part)
+    end
+
+    def deserialize_token(decrypted_string)
+      JSON.parse(decrypted_string, object_class: OpenStruct)
+    end
+
+    def message_encryptor
+      KmsEncrypted::Box.new(previous_versions: [{ key_id: Settings.lockbox.master_key }])
```
ericboehs, line 59: You shouldn't need the previous_versions here unless you already have encrypted records in production.

ericboehs (approved):
I'd prefer you remove the "previous_versions" stuff but the code looks good. I can reapprove in the morning if you want to make those changes.

bosawt:
> I'd prefer you remove the "previous_versions" stuff but the code looks good. I can reapprove in the morning if you want to make those changes.

Sounds good, I removed `previous_versions` from this

app/services/sign_in/refresh_token_encryptor.rb
```diff
@@ -0,0 +1,42 @@
...
+      string_array.join('.')
+    end
+
+    def serialize_and_encrypt_refresh_token
+      serialized_refresh_token = refresh_token.to_json
+      message_encryptor.encrypt(serialized_refresh_token)
+    end
+
+    def message_encryptor
+      KmsEncrypted::Box.new(previous_versions: [{ key_id: Settings.lockbox.master_key }])
```
ericboehs, line 39: This one is also not needed.

ericboehs (approved):

PR #9238, opened 2022-02-28
37350 oauth services scaffolding

bramleyjl:
## Description of change

This PR is the beginning framework for version two of the vets-api user sessions controller, implementing an OAuth authentication interface with Login.gov.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#37350
department-of-veterans-affairs/va.gov-team#37358

In order to test the new controller, navigate to `localhost:3000/sign_in/logingov/new` and complete the authentication process, you should be able to successfully login and be forwarded to the `vets-website` post-login endpoint.

The purpose of this PR is to establish a base of development in the vets-api `master` branch for Identity team members to iteratively develop upon. It allows for localhost authentication with Login.gov only; the necessary certificates and configuration settings have not been added to enable functionality on dev, staging, or production. This PR only covers login, logout still runs through the v1 controller, which results in the session properly ending on va.gov & vets-api but continuing on Login.gov; attempting to login again with Login.gov will allow the user to continue their session without re-authentication.

The new sessions controller has unit specs to cover new authentication and callback methods for Login.gov, as well as some specs for the Login.gov auth service itself. Manual logins through the v1 controller have been tested and are unaffected by this change.


bosawt (commented):
This is looking good!

We will need specs for `logingov/service.rb` too, the public functions in there, otherwise is looking good!

spec/controllers/sign_in_controller_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+              expect(get(:new, params: { type: type }))
+                .to have_http_status(:ok)
+              expect_oauth_post_form(response.body, "#{type}-form", url)
+            end
+          end
+        end
+      end
+    end
+
+    describe 'POST callback' do
```
bosawt, line 31: I think this is `GET` now

lib/sign_in/sign_in.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+module SignIn
```
bosawt, line 3: Oh yeah, are we using anything for this file? Probably can get rid of it

spec/fixtures/sign_in/logingov.crt
```diff
@@ -0,0 +1,16 @@
+-----BEGIN CERTIFICATE-----
+MIICljCCAX4CCQD6AhBNd7f/2zANBgkqhkiG9w0BAQsFADANMQswCQYDVQQGEwJV
```
bosawt, line 2: Now that this is here as a fixture, why don't we just set this path in `test.yml` settings file? Then we wouldn't even need to stub the cert/key in the test specs, it'd just pick these up

Also, could we just go ahead and set the localhost paths to these fixture files as well?

thilton-oddball (approved):

PR #9239, opened 2022-02-28
[API-13369] - ClaimsApi - Handle eFolder Access Failures

jw81:
## Description of change

Only invoke the eFolder access grant logic (`ClaimsApi::VBMSUpdater`) after the POA record has been successfully updated in BGS.  This should prevent situations where we're trying to grant eFolder access before the POA has been changed in BGS.

## Original issue(s)
[API-13369](https://vajira.max.gov/browse/API-13369)


hinzed1127 (approved):
Nice, this is much cleaner. LGTM

PR #9243, opened 2022-03-01
API-13151 Part 1: NOD v2 create endpoint and evidence submission

ElleeB:
This PR does the following:

1. Wires up the _basic_ NOD v2 create (doesn't yet submit the actual pdf)
2. Adds an NOD v2 evidence submission controller and spec, which includes the addition of `X-VA-File-Number` to our ev sub validator
3. Rswag config and spec file added for NOD V2 (some tests marked as pending until next PR)
4. A couple fix-ups in general (comments in diff)

---
_Note: I'm a little perplexed by the reach of the NODs versioning we did when creating our V2 API. Please look closely and let me know if I've modified or removed something essential._


modules/appeals_api/app/services/appeals_api/evidence_submission_request_validator.rb
```diff
@@ -7,9 +7,9 @@ class EvidenceSubmissionRequestValidator
       SupplementalClaim
     ].freeze
 
-    def initialize(appeal_uuid, request_ssn, appeal_type)
+    def initialize(appeal_uuid, veteran_identifier, appeal_type)
       @appeal_uuid = appeal_uuid
-      @request_ssn = request_ssn
+      @veteran_identifier = veteran_identifier
```
ElleeB, line 12: renaming to represent newly added `ssn || file number`

modules/appeals_api/config/schemas/v2/10182.json
```diff
@@ -80,19 +80,19 @@
...
-                      "properties": {
-                        "name": { "type": "string", "maxLength": 120 }
-                      }
                     }
                   },
                   "required": ["homeless",  "phone", "email"],
                   "if": { "properties": { "homeless": { "const": false } } },
                   "then": { "required": ["address"] }
                 },
+                 "representative": {
```
ElleeB, line 89: This had been left within `veteran`. Probably not caught since it's not being tested?

modules/appeals_api/spec/docs/v2/nod_spec.rb
```diff
@@ -116,8 +115,8 @@
         schema '$ref' => '#/components/schemas/errorModel'
 
         let(:nod_body) do
-          request_body = JSON.parse(File.read(AppealsApi::Engine.root.join('spec', 'fixtures', 'v1', 'valid_10182.json')))
-          request_body['data']['attributes'].delete('socOptIn')
```
ElleeB, line 120: soc removed for v2

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements_controller.rb
```diff
@@ -83,20 +83,11 @@ def new_notice_of_disagreement
       auth_headers: request_headers,
       form_data: @json_body,
       source: request_headers['X-Consumer-Username'],
-      board_review_option: @json_body['data']['attributes']['boardReviewOption'],
-      api_version: api_version
```
cilestin, line 87: We definitely want to keep the api version saved on the model, whether it's v1 or v2.

modules/appeals_api/app/services/appeals_api/evidence_submission_request_validator.rb
```diff
@@ -72,8 +73,8 @@ def invalid_review_option_error
     def invalid_veteran_id_error
       {
         title: 'unprocessable_entity',
-        detail: I18n.t('appeals_api.errors.mismatched_ssns'),
-        code: 'DecisionReviewMismatchedSSN',
+        detail: I18n.t('appeals_api.errors.mismatched_veteran_identifier'),
+        code: 'DecisionReviewMismatchedVeteranIdentifier',
```
cilestin, line 77: We don't want to do this yet, as this change affects all NOD versions, and we have a contract to keep for production even if we don't have any clients using it.

We should either make this respond with this new error only for V2, or wait until we get confirmation that we can remove NOD v1.

modules/appeals_api/config/routes.rb
```diff
@@ -51,18 +51,11 @@
       end
 
       namespace :notice_of_disagreements do
-        resources :evidence_submissions,
-                  only: %i[create show],
-                  controller: '/appeals_api/v1/decision_reviews/notice_of_disagreements/evidence_submissions'
+        get 'contestable_issues', to: 'contestable_issues#index'
```
cilestin, line 54: We don't want contestable issues inside the NOD v2 namespace. They should use the standalone contestable issues endpoint.

modules/appeals_api/spec/requests/v1/notice_of_disagreements_controller_spec.rb
```diff
@@ -91,26 +91,6 @@ def base_path(path)
         expect(nod.board_review_option).to eq('hearing')
       end
     end
-
-    context 'keeps track of which version of the api it is serving' do
```
cilestin, line 95: We still want this - just because the version may be going away doesn't mean we don't want to track which version is being used.

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements_controller.rb
```diff
@@ -27,7 +27,7 @@ class AppealsApi::V1::DecisionReviews::NoticeOfDisagreementsController < Appeals
 
   def create
     @notice_of_disagreement.save
-    AppealsApi::PdfSubmitJob.perform_async(@notice_of_disagreement.id, 'AppealsApi::NoticeOfDisagreement', api_version)
+    AppealsApi::PdfSubmitJob.perform_async(@notice_of_disagreement.id, 'AppealsApi::NoticeOfDisagreement')
```
cilestin, line 30: I think being explicit about the api version argument is better in this case, even if it's just to pass `v1` as a string into it here.

modules/appeals_api/app/services/appeals_api/pdf_construction/notice_of_disagreement/v2/form_data.rb
```diff
@@ -29,8 +29,7 @@ def mailing_address
             signing_appellant.city,
             signing_appellant.state_code,
             signing_appellant.zip_code,
-            signing_appellant.country_code,
-            signing_appellant.international_postal_code
+            signing_appellant.country_code
```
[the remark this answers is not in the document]
cilestin: Then what happens if they supply an iternationalpostalcode w/ 00000s in the zipcode?

modules/appeals_api/spec/services/appeals_api/evidence_submission_request_validator_spec.rb
```diff
@@ -61,43 +61,44 @@ module AppealsApi
...
+            notice_of_disagreement.id, 'fake_file_number', 'NoticeOfDisagreement'
           ).call
 
           expect(result).to eq([
                                  :error,
                                  {
                                    title: 'unprocessable_entity',
-                                   detail: "Request header 'X-VA-SSN' does not match the associated appeal's SSN",
-                                   code: 'DecisionReviewMismatchedSSN',
+                                   detail: "Request header 'X-VA-SSN' or 'X-VA-File-Number' does not match the "\
```
cilestin, line 88: Again, this will change for both NOD v1 and v2, and we don't want to change responses on v1.

modules/appeals_api/spec/services/appeals_api/evidence_submission_request_validator_spec.rb
```diff
@@ -9,7 +9,7 @@ module AppealsApi
       context 'Notice of Disagreement' do
         it 'returns an error if the NOD does not exist' do
           result = described_class.new(
-            'fake_id', 'irrelevant_ssn', 'NoticeOfDisagreement'
+            'fake_id', 'irrelevant_file_number', 'NoticeOfDisagreement'
```
cilestin, line 12: I appreciate testing the change from SSN to SSN or FileNumber, but also moving all the created NODs to v2 seems premature at the moment.

modules/appeals_api/spec/services/appeals_api/evidence_submission_request_validator_spec.rb
```diff
@@ -160,8 +161,9 @@ module AppealsApi
                                  :error,
                                  {
                                    title: 'unprocessable_entity',
-                                   detail: "Request header 'X-VA-SSN' does not match the associated appeal's SSN",
-                                   code: 'DecisionReviewMismatchedSSN',
+                                   detail: "Request header 'X-VA-SSN' or 'X-VA-File-Number' does not match the "\
```
cilestin, line 164: Same. This changes v1 and v2 when it should only change for v2.

modules/appeals_api/spec/support/rswag_config.rb
```diff
@@ -365,11 +365,11 @@ def hlr_v2_response_schemas(ref_root)
     }
   end
 
-  def nod_create_schemas
```
cilestin, line 368: Don't we want to keep the v1 schemas around for production documentation?

cilestin (changes requested):

modules/appeals_api/app/services/appeals_api/pdf_construction/notice_of_disagreement/v2/form_data.rb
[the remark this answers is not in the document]
ElleeB:
```
    def zip_code
      international_postal_code || zip_code_5
    end
```

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements_controller.rb
ElleeB: Ok. But, we were only doing this for NOD because it checked if it was 'v1' or 'v2'. We don't currently have a pattern for holding onto the API version. We just pass it in, or it defaults to 'v1'.

modules/appeals_api/app/services/appeals_api/evidence_submission_request_validator.rb
ElleeB: Right. Ok. Things get murky in our handling of the ev sub validation, but I'll make it work.

modules/appeals_api/config/routes.rb
ElleeB: Oh, right! Omg.

modules/appeals_api/spec/requests/v1/notice_of_disagreements_controller_spec.rb
ElleeB: We were doing this because 'v2' was rerouted to 'v1'. You still want this around?

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements_controller.rb
ElleeB: The pdf submit job defaults to 'v1'

modules/appeals_api/spec/services/appeals_api/evidence_submission_request_validator_spec.rb
ElleeB: K.

modules/appeals_api/spec/services/appeals_api/evidence_submission_request_validator_spec.rb
ElleeB: Ok. I figured since the ssn is tested in SC, that I'd just test file number, but I can return some of them to v1

modules/appeals_api/spec/support/rswag_config.rb
ElleeB: I'm confused about this. We didn't for hlr...but wait, hlr had the old schema set up. But also, don't we hand pick what we change?

Ugh. The rswag stuff is so annoying.

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements_controller.rb
cilestin: I believe we have an example pattern with HLRs. Since we're adding the v2 controller, we should update both it and this v1 controller to store the correct api version.

modules/appeals_api/spec/requests/v1/notice_of_disagreements_controller_spec.rb
cilestin: It's still a core part of the controller, yes. Every appeal record should store which api version was used to create it (and which pdf version we rendered it with when that job happens).

modules/appeals_api/app/controllers/appeals_api/v1/decision_reviews/notice_of_disagreements_controller.rb
cilestin: It does, but I still think being explicit on the version we're using makes things a little more clear.

cilestin (approved):

PR #9244, opened 2022-03-01
Coe prefill changes

alsleaders:
## Original issue(s)


spec/models/form_profile_spec.rb
```diff
@@ -67,31 +67,6 @@
     }
   end
 
-  let(:v40_10007_expected) do
```
thilton-oddball, line 70: Since there are still models for these two form profiles, these tests should probably be left in. If the forms are no longer used at all, it might be better to remove the models, tests and anything else relating to them in a separate PR.

spec/models/form_profile_spec.rb
```diff
@@ -1035,17 +1032,6 @@ def stub_methods_for_emis_data
         end
       end
 
-      context 'with emis and vet360 prefill for 0873' do
```
thilton-oddball, line 1038: Same as previous, since there is still a model for form_profile va_0873, this should probably be left in.

spec/models/form_profile_spec.rb
```diff
@@ -1114,13 +1100,13 @@ def stub_methods_for_emis_data
           22-1995
           22-5490
           22-5495
-          40-10007
```
thilton-oddball, line 1117: Same as previous, since there is still a model for form_profile va_4010007, this should probably be left in.

thilton-oddball (changes requested):

spec/models/form_profile_spec.rb
alsleaders: Hi @thilton-oddball, I pulled these tests out based on this [slack thread](https://dsva.slack.com/archives/CBU0KDSB1/p1642802354171800) to get around a `RSpec/MultipleMemoizedHelpers: Max: 37` error.  

What I would prefer to do is bump up the rubocop max and leave these tests alone.  Would you be alright with that?

thilton-oddball (approved):
I think for now at least, bumping up the Max to 40 will be fine.

alsleaders:
> I think for now at least, bumping up the Max to 40 will be fine.

Thank you @thilton-oddball!

PR #9246, opened 2022-03-01
Handle a nil response from the BGS payments service

kreek:
## Description of change
A recent refactor is causing 500s when the BGS service in vets-api catches [SOAP errors](http://sentry.vfs.va.gov/organizations/vsp/issues/82645/?environment=staging&project=3&query=is%3Aunresolved) and returns a nil response. ~~This restores the previous functionality which was returning an empty list.~~ It now returns a 502 unless the error includes a message of 'No Data Found' in which case it returns an empty list.

## Original issue(s)
department-of-veterans-affairs/va-mobile-app/issues/2664


jperk51 (approved):
LGTM

ericboehs (approved):

app/models/adapters/payment_history_adapter.rb
```diff
@@ -5,13 +5,18 @@ class PaymentHistoryAdapter
...
     def initialize(input)
-      @input_payments = [input[:payments][:payment]].flatten
-      payments, return_payments = @input_payments.partition do |payment|
-        payment.dig(:return_payment, :check_trace_number).blank?
-      end
+      @payments = []
+      @return_payments = []
+
+      if input
+        @input_payments = [input[:payments][:payment]].flatten
```
kpethtel, line 12: Is there any reason to think input might not have payments? That could lead to an error here.

kpethtel (dismissed):
Looks good. I left one comment about another potential error we could preemptively avoid, but it might be better to let it happen so that we can explicitly handle it.

kpethtel (approved):
🚢

PR #9251, opened 2022-03-02
[37884] adds CORP ID to mhv-outbound-redirect duplicate id validation exceptions

bramleyjl:
## Description of change

This change updates the login validation logic to allow logins for users with multiple CORP IDs or multiple unique ICNs who are using MHV's unified signin authentication flow. MHV unified signin users now only logs warning for multiple:

* MHV IENs
* CORP IDs
* ICNs

The behavior of these exceptions has also been modified to more closely resemble the normal authentication flow; the logging generated by each is now identical.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#37884
department-of-veterans-affairs/va.gov-team#38552

Unit tests updated to cover MHV unified signin for the affected IDs, manual logins through normal and MHV unified signin flows with duplicate & non-duplicate ids performed.


lib/saml/user_attributes/ssoe.rb
```diff
@@ -269,9 +269,9 @@ def validate!
       def multiple_id_validations
         # EDIPI, ICN, and CORP ID all trigger errors if multiple unique IDs are found
         check_mhv_icn_mismatch
-        check_id_mismatch(mvi_ids[:vba_corp_ids], :multiple_corp_ids)
+        check_id_mismatch(mvi_ids[:vba_corp_ids], :multiple_corp_ids, mhv_outbound_redirect: mhv_outbound_redirect)
```
bosawt, line 272: It looks like we're no longer logging message to sentry when we are `mhv_outbound_redirect` and hit the multiple ids issue. In that case, it might be simpler at this point to refactor these lines to this format:
`check_id_mismatch(mvi_ids[:vba_corp_ids], :multiple_corp_ids) unless mhv_outbound_redirect`
Then we can clean up the `check_id_mismatch` function as well

thilton-oddball (approved):

lib/saml/user_attributes/ssoe.rb
```diff
@@ -323,41 +321,23 @@ def mhv_iens
         mhv_iens.append(safe_attr('va_eauth_mhvuuid')).reject(&:nil?).uniq
       end
 
-      def mhv_inbound_outbound
+      def mhv_outbound_redirect
```
jscanish, line 324: This might be worth memoizing. We call this multiple times and I don't see any reason why we would need to evaluate it each time.
```suggestion
      def mhv_outbound_redirect
        @mhv_outbound_redirect ||= begin
          tracker = SAMLRequestTracker.find(@tracker_uuid)
          tracker&.payload_attr(:skip_dupe_id_validation) == 'true'
        end
      end
```

lib/saml/user_attributes/ssoe.rb
```diff
@@ -323,41 +321,23 @@ def mhv_iens
...
       def birls_id_mismatch?
         attribute_has_multiple_values?('va_eauth_birlsfilenumber')
       end
 
-      def check_corp_id_mismatch
-        check_id_mismatch(mvi_ids[:vba_corp_ids], :multiple_corp_ids)
-      end
-
-      def check_id_mismatch(ids, multiple_ids_error_type, mhv_inbound_outbound: false)
+      def check_id_mismatch(ids, multiple_ids_error_type, mhv_outbound_redirect: false)
```
jscanish, line 333: `mhv_outbound_redirect` doesn't need to be an argument. It's already a method in the class so you have access to it.

jscanish (commented):
Looks good overall. Just a few minor suggestions.

lib/saml/user_attributes/ssoe.rb
bramleyjl: Pretty sure we need it as an argument so that we can force it to `false` in the case of the EDIPI check, even for mhv_outbound_redirect we still want the normal behavior: throwing an error & preventing login. Or am I misinterpreting what you're saying?

jscanish (approved):

PR #9254, opened 2022-03-02
2565 remove gender fax mobile user v1

jperk51:
## Description of change
Since profile is being updated to no longer include gender and fax we are versioning and updating our user endpoint to also no longer include those fields. 

## Original issue(s)
department-of-veterans-affairs/va-mobile-app#2565

Updated v1 tests, schemas, and docs to reflect these changes


modules/mobile/app/serializers/mobile/v1/user_serializer.rb
```diff
@@ -0,0 +1,142 @@
...
+        super(resource, options)
+      end
+
+      private
+
+      def filter_keys(value, keys)
+        value&.to_h&.slice(*keys)
+      end
+
+      def profile
```
kreek, line 75: Are we able to have this serializer extend the v0 serializer and just override this profile method to remove gender and fax?

jperk51: I'll give a try. Good call

PR #9255, opened 2022-03-02
Api 13516 consolidate schemas

cilestin:
## Description of change
Since we tried to align on a few schema definitions across NOD and SC v2, this work extracts those shared schemas into the common definition space within each file. This is a first step to getting to a shared schema endpoint that we can reference to at some point.

## Original issue(s)
https://vajira.max.gov/browse/API-13516

* Swagger docs have not been updated in this PR


ElleeB (approved):
👏

ElleeB (approved):
Still looks awesome!

PR #9256, opened 2022-03-03
Allows complete bypass of login for local dev

patrickvinograd:
## Description of change
Allows complete mocking/bypass of SAML sign-in process, suitable for local development. Instantiates a UserIdentity/User/Session from a YAML file and immediately redirects to the frontend post-login callback page. 

An example of a mock file is shown here:
https://github.com/department-of-veterans-affairs/vets-api-mockdata/pull/193
To complete this, I would write a script to parse the existing mocked MPI data in vets-api-mockdata and write it out in the YAML format suitable for this PR. 

## Original issue(s)

* New settings `saml_ssoe.mock` and `saml_ssoe.mock_user_file`. `mock` should be false in all deployed environments.

Looking for feedback about whether developers would find this valuable for local use, and if so we can populate additional mock data files for all the existing test users.


patrickvinograd:
Screen cap showing immediate login without any external redirects. 

mock_sign_in

app/controllers/v1/sessions_controller.rb
```diff
@@ -31,8 +31,9 @@ class SessionsController < ApplicationController
     # For more details see SAML::SSOeSettingsService and SAML::URLService
     def new
       type = params[:type]
-
-      if type == 'slo'
+      if Settings.saml_ssoe.mock
```
bosawt, line 34: I like the overall idea but I'm not sure if I like this being embedded in `sessions_controller`. What do you think about making a separate controller entirely for this, like `mock_auth_controller`? I think it'd have the following benefits:
* Accidentally enabling `Settings.saml_ssoe.mock` on any stack wouldn't impact the standard auth flow
* It would make mocked authentication a very intentional action by forcing devs to type in `localhost:3000/mock_auth`
* We could enable the route itself only on non-prod environments, preventing misuse of mocked auth, even if the settings were improperly set

thilton-oddball (approved):

app/controllers/v1/sessions_controller.rb
joeniquette: I agree with this approach from Trevor. This is too high of a security risk to be put in the same controller used for our current auth flows. Creating a separate controller entirely should reduce the risk and provide engineers with the ability to mock auth. @patrickvinograd what are your thoughts on this approach? Our team cannot approve this change as is, so we should collab on getting something that meets your requirements but also is safe.

PR #9258, opened 2022-03-03
2677 appointment request adapter error

kpethtel:
## Description of change
Fixes a bug that's causing issues with Judy in staging. I ssh'ed to a server, ran the code manually, and found this error:
```
undefined method `[]' for nil:NilClass
/srv/vets-api/src/modules/mobile/app/models/mobile/v0/templates/community_care_appointment.rb:45:in `location'
```
This PR fixes the nil error in location and changes how the adapter handles errors. The previous approach was to rescue any error and return nil, which caused other issues because the proxy expects to receive an array, not nil. I've changed that behavior to rescue the bug, log it, and skip the record. That means we may sometimes get missing appointment requests in responses, but this is feature flagged and this should help us debug any issues we see while also not blocking dev work. 

## Original issue(s)
department-of-veterans-affairs/va-mobile-app#2677

Tested on the review instance. This is also feature flagged and shouldn't disrupt production.


modules/mobile/app/models/mobile/v0/templates/community_care_appointment.rb
```diff
@@ -41,20 +41,25 @@ def location
             },
             lat: nil,
             long: nil,
-            phone: {
-              area_code: phone_captures[1].presence,
```
kpethtel, line 45: This was the real problem. It was bombing when there was was no phone number.

modules/mobile/app/models/mobile/v0/adapters/appointment_requests.rb
```diff
@@ -19,12 +19,15 @@ def parse(requests)
...
+          rescue => e
+            Rails.logger.error(
+              'Error adapting appointment request',
+              appointment_request: request, error: e.message, backtrace: e.backtrace
+            )
+            next
           end
 
           [va_appointments, cc_appointments]
-        rescue => e
```
kpethtel, line 25: This was the other problem. If we encountered any adapter issues, we rescued and logged but didn't re-raise. Which meant we returned nil instead of an array of two arrays.

modules/mobile/spec/models/adapters/appointment_requests_adapter_spec.rb
```diff
@@ -328,12 +328,49 @@
...
           long: nil,
-          phone: { area_code: '703', number: '652-0000', extension: nil },
+          phone: { area_code: '703', number: '652-0000', extension: 'x1000' },
           url: nil,
           code: nil
         }
         expect(adapted_cc_appt_request.location.to_h).to eq(expected_location)
       end
+
+      it 'handles nil phone numbers' do
```
kpethtel, line 338: My next ticket is to extract the phone number parser into a helper method. It will involve detailed testing.

aherzberg (approved):
LGTM

jperk51 (approved):

PR #9260, opened 2022-03-03
RRD: Fix flakey get_first_name test

yoomlam:
## Description of change

Fix flakey test for `get_first_name`.

When a `User` for the `subject.user_uuid` is requested, provide our stubbed `user_with_nil_first_name` User record.
The previous code only stubbed on a specific instance resulting from the `User.find(subject.user_uuid)` call, which can be different than the instance returned from the `User.find(subject.user_uuid)` in the `get_first_name` method, even though the instances refer to the same `User` record.

## Original issue(s)
The flakey test failure is blocking PR merges.
Notified by @joeniquette about failure.


app/services/rapid_ready_for_decision/processor_selector.rb
```diff
@@ -22,7 +22,7 @@ def rrd_applicable?
 
     def self.disability_increase?(disability, diagnostic_code)
       disability['diagnosticCode'] == diagnostic_code &&
-        disability['disabilityActionType'] == 'INCREASE'
+        disability['disabilityActionType']&.upcase == 'INCREASE'
```
yoomlam, line 25: Not a current problem, but just in case, let's `upcase`.

bramleyjl (approved):

thilton-oddball (approved):

PR #9261, opened 2022-03-03
API-13151 Part 2: show and validate endpoints

ElleeB:
This PR:

1. Adds pdf submit job to create endpoint
2. Adds show, validate, and schema endpoints
3. Adds model validation: error if extension request false but reason given
4. Turns on the rswag specs for show, validate, and schema endpoints

[Ticket](https://vajira.max.gov/browse/API-13151)


modules/appeals_api/app/controllers/appeals_api/v2/decision_reviews/notice_of_disagreements_controller.rb
```diff
@@ -28,9 +28,29 @@ class AppealsApi::V2::DecisionReviews::NoticeOfDisagreementsController < Appeals
 
   def create
     @notice_of_disagreement.save
+    AppealsApi::PdfSubmitJob.perform_async(
+      @notice_of_disagreement.id,
+      'AppealsApi::NoticeOfDisagreement',
+      API_VERSION
```
cilestin, line 34: I'll preface this by saying I recognize it sounds like I'm splitting hairs, but I think it's important that we distinguish between our api versions and the pdf generator versions. So far, we've had them in sync with each other (api v2 == pdf v2), but that's not always guaranteed. 

It could be that there's a v3 pdf we'll need to make where none of the form_data changes but the fields move around - in that case using API_VERSION would be incorrect. I think we should just pass the version string directly into the job argument in this case.

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
```diff
@@ -255,6 +256,20 @@ def validate_hearing_type_selection
       end
     end
 
+    # v2 specific validation
+    def validate_extension_request
```
cilestin, line 260: I'm torn between whether we should do this or just document that we'll ignore the extensionReason if extensionRequest is false. I'll have to look over some of our other validations to see if anything sways me. Not a blocker for now, but wanted to mention where my head's at so it doesn't come as a surprise if we remove it later.

cilestin (changes requested):
Lookin' good! Just a small change and a note.

modules/appeals_api/app/controllers/appeals_api/v2/decision_reviews/notice_of_disagreements_controller.rb
ElleeB: Not splitting. I had the same thought and went in the lamer direction. I'll get it fixed up.

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
ElleeB: I asked in our bp private channel and kelly responded that we should error. I'll bring it up during stand up next week.

cilestin (approved):
Looks great!

PR #9263, opened 2022-03-03
Add safe navigation for clinics in v2 appt controller

ajmagdub:
## Description of change

Added safe navigation in v2 AppointmentsController to handle cases where a nil clinic object is returned

Staging data had a case failing when clinic was returning nil. Made changes to the v2 Appointments controller and verified that the nil clinic case did not throw an exception.


grossb1 (approved):
LGTM

PR #9265, opened 2022-03-04
2438 cancel pending appointment

aherzberg:
## Description of change

Add logic to allow canceling of pending appointment through already existing cancel appointment endpoint. This requires getting the single appointment through VAOS due to the params required for the cancel have fields that we do no retain in our appointment model. 

Endpoint determine if cancelling a pending appointment by if the cancel id is a uuid appointment id. If it is not, then it is assumed to be an encoded string of fields used for canceling normal confirmed appointments. 
 
## Original issue(s)
 https://github.com/department-of-veterans-affairs/va-mobile-app/issues/2438


modules/mobile/app/controllers/mobile/v0/appointments_controller.rb
```diff
@@ -37,16 +37,39 @@ def index
...
-        contract = Mobile::V0::Contracts::CancelAppointment.new.call(decoded_cancel_params)
-        raise Mobile::V0::Exceptions::ValidationErrors, contract if contract.failure?
+        id = params[:id]
+        # appointment request cancel ids is appointment id while other appointments will be encoded string
+        if uuid?(id)
+          appointment_request = appointments_proxy.get_appointment_request(id)
+          raise Common::Exceptions::BackendServiceException, 'MOBL_502_upstream_error' if appointment_request.nil?
+
+          appointment_status = appointment_request['status']
+          if appointment_status == 'Submitted'
```
aherzberg, line 47: Technically the VAOs endpoint will accept already cancelled requests and just give back the same thing but wanted to avoid hiding the issue.

modules/mobile/app/services/mobile/v0/appointments/proxy.rb
```diff
@@ -47,6 +47,20 @@ def put_cancel_appointment(params)
...
         end
 
+        def get_appointment_request(id)
+          vaos_appointment_requests_service.get_request(id)[:data]
+        end
+
+        def put_cancel_appointment_request(id, request)
+          parsed_request = request.as_json['table']
+          parsed_request['status'] = 'Cancelled'
+          parsed_request['appointment_request_detail_code'] = ['DETCODE8']
```
kpethtel, line 57: Did you get DETCODE8 from the VAOS code somewhere? How'd you determine this is the one to use?

modules/mobile/app/services/mobile/v0/appointments/proxy.rb
```diff
@@ -148,12 +162,11 @@ def handle_cancel_error(e, params)
           if e.original_status == 409
             Rails.logger.info(
               'mobile cancel appointment facility not supported',
-              clinic_id: params[:clinicId],
-              facility_id: params[:facilityId]
+              clinic_id: params[:clinicId] || nil,
```
kpethtel, line 165: The `|| nil` is redundant here.

kpethtel (commented):
Looks good, just a couple questions we should consider before approval.

modules/mobile/app/controllers/mobile/v0/appointments_controller.rb
aherzberg: That's fair. Probably want to mirror the logic for cancelling confirmed appointments which doesn't do any checks on status before attempting to cancel. I'll confirm with FE but will assume for now they will want it consistent and update that.

modules/mobile/app/services/mobile/v0/appointments/proxy.rb
aherzberg: from the VAOS spec appointment request factory. When making a cancellation request it updates the status and detail code to DETCODE8. I believe that code essentially means, "cancelled by user"

modules/mobile/app/controllers/mobile/v0/appointments_controller.rb
kpethtel: I'm not sure that there is a clear right or wrong here. Logging an error if it's already been cancelled could reveal issues, but it will also almost surely sometimes be the cause of a race condition and cause us to think there's an issue when there isn't. Not checking first has multiple advantages though: the request will be faster, the code will be simpler, and we don't have to mess around another VCR cassette in specs. And my question about 404s becomes irrelevant.

modules/mobile/app/services/mobile/v0/appointments/proxy.rb
kpethtel: It is, but according to code I wrote here: modules/mobile/app/models/mobile/v0/templates/base_appointment.rb
there's another cancelled by user case. DETCODE22. I believe they all have a status detail that gives more info, but I don't recall what they were. If this is what they use in VAOS, I'm sure it's fine.

aherzberg: oh yeah, I didn't see that. The returned status detail for DETCODE8 is `"provider_message":"Cancelled by Veteran - Pending",
                  "veteran_message":"Your request to cancel this appointment has been sent.",` which seems correct.

modules/mobile/app/services/mobile/v0/appointments/proxy.rb
aherzberg: fixed

modules/mobile/app/controllers/mobile/v0/appointments_controller.rb
```diff
@@ -37,16 +37,39 @@ def index
...
+            appointments_proxy.put_cancel_appointment_request(id, appointment_request)
+          else
+            Rails.logger.info("mobile appointment status is not 'SUBMITTED'", user_uuid: @current_user.uuid,
+                                                                              status: appointment_status)
+            raise Common::Exceptions::ValidationErrors, 'MOBL_500_wrong_status'
+          end
+
```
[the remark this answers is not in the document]
aherzberg: fixed

modules/mobile/app/controllers/mobile/v0/appointments_controller.rb
aherzberg: Thats a good argument to pass return the 200 without any additional error logging. I've updated it to do that.

kpethtel (approved):
This looks good. My one concern is that we're now naming some cassettes `pending_appointments` and some `appointment_requests`. I think that will cause some confusion. This is something I thought about while adding appointment requests. The word "request" is overloaded and carries some inherent confusion. But I ended up choosing it because that's what we seem to call it in most other places (VAOS, VAMF) and I thought it would be better to be consistent. Regardless, I think we should pick a naming convention and stick with it.

modules/mobile/app/controllers/mobile/v0/appointments_controller.rb
```diff
@@ -37,18 +37,31 @@ def index
       end
 
       def cancel
-        decoded_cancel_params = Mobile::V0::Appointment.decode_cancel_id(params[:id])
-        contract = Mobile::V0::Contracts::CancelAppointment.new.call(decoded_cancel_params)
-        raise Mobile::V0::Exceptions::ValidationErrors, contract if contract.failure?
+        id = params[:id]
+        # appointment request cancel ids is appointment id while other appointments will be encoded string
+        if uuid?(id)
+          appointment_request = appointments_proxy.get_appointment_request(id)
```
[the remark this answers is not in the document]
aherzberg: Yeah a cancel endpoint would have been ideal. Referring to the vcr naming, good catch. I updated the names to all have `pending`.

kpethtel (approved):
Looks good. 🚀

kreek (approved):
LGTM 🚀

PR #9273, opened 2022-03-05
[37200] Sign In Access token creation, encoding, decoding

bosawt:
## Description of change
This PR adds access token model with compatibility with the Sign In methods, in addition to JWT encoding and decoding of the access token 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#37200

* This PR has no current interactions with prod
* To test, need to get on rails console, create an access token with `SignIn::AccessToken.new(session_handle: 'some-session-handle', user_uuid: 'some-user-uuid', refresh_token_hash: 'some-refresh-token-hash', anti_csrf_token: 'some-anti-csrf-token', last_regeneration_time: Time.now)`
* Test encoding with `SignIn::AccessTokenJwtEncoder.new(access_token: access_token).perform`
* Test decoding with `SignIn::AccessTokenJwtDecoder.new(access_token_jwt: access_token_jwt).perform(with_validation: with_validation)` (`with_validation` set to `true` for JWT validation, `false` to just decode without validation)


jscanish (approved):
Tested in the console. Seems to work well, though your instructions have two minor issues:
1) The initial creation of the token is missing the parent module, so `AccessToken.new` needs to be `SignIn::AccessToken.new`.
2) The decoding step has a `with_validation` variable that is never set up, so it should probably just be `true`. `.perform(with_validation: with_validation)` -> `.perform(with_validation: true)`

bramleyjl (approved):
LGTM! Agreed with updating the instructions for when they go into documentation, it took me a minute to remember that the `AccessToken` class had to be accessed through `SignIn` module.

thilton-oddball (approved):

PR #9277, opened 2022-03-07
API-13911 Finalize NOD v2 pdf printing placement

ElleeB:
The new form went live and is now official. This PR saves the new form in our config files and ensures that all data is printed in the proper position.

[Ticket](https://vajira.max.gov/browse/API-13911)


modules/appeals_api/spec/services/appeals_api/pdf_construction/generator_spec.rb
```diff
@@ -20,7 +20,7 @@
 
         it 'generates the expected pdf' do
           generated_pdf = described_class.new(notice_of_disagreement).generate
-          expected_pdf = fixture_filepath('expected_10182_minimum.pdf', version: 'v1')
```
ElleeB, line 23: making this cap to create consistency with the other uses in the spec for slc and hlr - made a ticket to take a look at all the places where we have discrepancies in case we want to fix.

cilestin (approved):

PR #9291, opened 2022-03-08
API-13787 Add deprecate key to swagger blocks for nod v1

ElleeB:
Learned that a deprecated key is necessary for each endpoint in swagger blocks to complete the process.

[Ticket](https://vajira.max.gov/browse/API-13787)


ElleeB:
Will bring this back up when nod v2 goes to production

PR #9295, opened 2022-03-08
RRD: invalidate memoized form_hash variable after it is updated

yoomlam:
## Description of change

Add `invalidate_form_hash` to invalidate the memoized `@form_hash` variable to be called after updating `form_json`.

Add code comment to remind folks to call `invalidate_form_hash` after updating `form_json`.

## Original issue(s)
`form_json` is memoized when `submission.form` is called so any updates to `form_json` will not be reflected if the memoization occurs before the update happened. This is not a current problem because `submission.form` happens to not be called until after `form_json` is updated but there is nothing enforcing that, so this is a latent bug.

No change in software behavior.


app/services/rapid_ready_for_decision/hypertension_upload_manager.rb
```diff
@@ -18,7 +18,8 @@ def add_upload(confirmation_code)
       }
       uploads.append(new_upload)
       data['form526_uploads'] = uploads
-      submission.update(form_json: JSON.dump(data))
+      submission.update!(form_json: JSON.dump(data))
```
yoomlam, line 21: https://marcgg.com/blog/2020/04/08/rails-update-bang/

app/models/form526_submission.rb
```diff
@@ -155,12 +155,20 @@ def full_name
...
   def form
     @form_hash ||= JSON.parse(form_json)
   end
 
+  # Call this method to invalidate the memoized @form_hash variable after updating form_json.
+  # A hook that calls this method could be added so that we don't have to call this manually,
+  # see https://stackoverflow.com/questions/24314584/run-a-callback-only-if-an-attribute-has-changed-in-rails
+  def invalidate_form_hash
+    remove_instance_variable(:@form_hash) if instance_variable_defined?(:@form_hash)
+  end
```
yoomlam, line 170: https://makandracards.com/makandra/21031-careful-with-it-s-not-memoize

app/services/rapid_ready_for_decision/hypertension_special_issue_manager.rb
```diff
@@ -15,6 +15,8 @@ def add_special_issue
         add_rrd_code(disability) if hypertension_increase?(disability)
       end
       submission.update!(form_json: JSON.dump(submission_data))
+      submission.invalidate_form_hash
+      submission
```
yoomlam, line 19: It's not clear that this method wants/needs to return a `submission` record, but let's do that for consistency with `hypertension_upload_manager.rb`.

nanotone (approved):
👍

thilton-oddball (approved):

PR #9296, opened 2022-03-08
RRD Hypertension: Append date range to PDF filename

allthesignals:
## Description of change
Addresses https://www.notion.so/Change-PDF-document-attributes-cafd6028c47749ccbccfb29deb9c5023 by appending search dates to the filename.

The PR includes two interrelated changes:

1. Pulls the strategy for checking whether an RRD Hypertension claim was processed into the form526_base_job class.
2. Appends relevant dates to the PDF filename.

Currently, the code checks for RRD claim status by checking (by filename) if the PDF is referenced in form526_submission's `form_json`. Because attaching (dynamic) search date information to that filename would break this logic, this PR modifies the logic slightly to check for the general document name.

Future updates should consider more sophisticated approaches to grabbing search date information. This PR assumes that search date is the date on which the PDF is uploaded. Another improvement would be to infer RRD claim submission status through more explicit means (via job status, etc).


spec/services/rapid_ready_for_decision/hypertension_upload_manager_spec.rb
```diff
@@ -26,13 +26,24 @@
...
 
+  let(:time_freeze_time) { '10-10-2021' }
+
+  before do
+    Timecop.freeze(time_freeze_time)
+  end
+
+  after do
+    Timecop.return
+  end
```
allthesignals, line 37: Borrowed from in order to have a stable date to assert with: https://github.com/department-of-veterans-affairs/vets-api/blob/master/spec/requests/backend_status_request_spec.rb

app/mailers/hypertension_fast_track_pilot_mailer.rb
```diff
@@ -5,7 +5,7 @@ def build(submission)
     @submission = submission
 
     subject =
-      if submission.form_json.include? 'VAMC_Hypertension_Rapid_Decision_Evidence.pdf'
+      if submission.form_json.include? 'VAMC_Hypertension_Rapid_Decision_Evidence'
```
allthesignals, line 8: I wonder if this could be made a constant just so that name itself isn't floating around

app/mailers/hypertension_fast_track_pilot_mailer.rb
```diff
@@ -5,7 +5,7 @@ def build(submission)
     @submission = submission
 
     subject =
-      if submission.form_json.include? 'VAMC_Hypertension_Rapid_Decision_Evidence.pdf'
+      if submission.form_json.include? RapidReadyForDecision::HypertensionUploadManager::HYPERTENSION_PDF_NAME
```
allthesignals, line 8: Next step: new form526_submission method to detect this

app/mailers/hypertension_fast_track_pilot_mailer.rb
```diff
@@ -5,7 +5,7 @@ def build(submission)
     @submission = submission
 
     subject =
-      if submission.form_json.include? 'VAMC_Hypertension_Rapid_Decision_Evidence.pdf'
+      if submission.rrd_hypertension_claim_processed
```
yoomlam, line 8: Add a `?` and make it more general:
```suggestion
      if submission.rrd_claim_processed?
```

app/models/form526_submission.rb
```diff
@@ -336,6 +336,13 @@ def bdd?
     form.dig('form526', 'form526', 'bddQualified') || false
   end
 
+  # Infers that an HTN RRD Claim Processed
```
yoomlam, line 339:
```suggestion
  # @return if this claim submission was processed and fast-tracked by RRD
```

app/models/form526_submission.rb
```diff
@@ -336,6 +336,13 @@ def bdd?
...
+  def rrd_hypertension_claim_processed
+    form_json.include? RapidReadyForDecision::HypertensionUploadManager::HYPERTENSION_PDF_NAME
+  end
```
yoomlam, line 342: I recommend moving this to `Form526BaseJob` as `self.rrd_claim_processed?`. As much as possible, let's keep RRD-related code separate from the `Form526Submission` ActiveRecord  (to follow separation of concerns practice).

app/services/rapid_ready_for_decision/hypertension_upload_manager.rb
```diff
@@ -4,15 +4,25 @@ module RapidReadyForDecision
...
     def initialize(submission)
       @submission = submission
+      @date = Time.zone.today
+    end
+
+    def file_upload_name
+      end_date = @date.strftime('%m/%d/%Y')
+      start_date = (@date - 1.year).strftime('%m/%d/%Y')
+
+      "#{HYPERTENSION_PDF_NAME}[#{start_date}-#{end_date}].pdf"
```
yoomlam, line 18: `@date` is only used within one method, so:
```suggestion
    end

    def file_upload_name
      today = Time.zone.today
      end_date = today.strftime('%m/%d/%Y')
      start_date = (today - 1.year).strftime('%m/%d/%Y')

      "#{HYPERTENSION_PDF_NAME}[#{start_date}-#{end_date}].pdf"
```
Why is `start_date` 1 year ago (FYI: you could use `1.year.ago` instead)? Should it be the submission date?

spec/services/rapid_ready_for_decision/hypertension_upload_manager_spec.rb
```diff
@@ -26,13 +26,24 @@
...
+
   describe '#add_upload(confirmation_code)' do
     context 'success' do
       it 'appends the new upload and saves the expected JSON' do
         RapidReadyForDecision::HypertensionUploadManager.new(form526_submission).add_upload('fake_confirmation_code')
         parsed_json = JSON.parse(form526_submission.form_json)['form526_uploads']
+
         expect(parsed_json).to match original_form_json_uploads + [
-          { 'name' => 'VAMC_Hypertension_Rapid_Decision_Evidence.pdf',
+          { 'name' => 'VAMC_Hypertension_Rapid_Decision_Evidence[10/10/2020-10/10/2021].pdf',
```
yoomlam, line 46: Are we sure we want to use square brackets in filenames? Brackets are treated specially in command prompt shells. How about a simple dash or double dash like: `VAMC_Hypertension_Rapid_Decision_Evidence-10/10/2020-10/10/2021` or `VAMC_Hypertension_Rapid_Decision_Evidence--10/10/2020-10/10/2021`

yoomlam (changes requested):

spec/services/rapid_ready_for_decision/hypertension_upload_manager_spec.rb
yoomlam: Yup. That too! I overlooked the slashes as I was copying and pasting.

app/models/form526_submission.rb
allthesignals: Sorry, not sure I'm following understand, @yoomlam — Form526Submission is an ActiveRecord class, is there a base job class I'm not seeing? I can see that Form526Submission has some related records (`SavedClaim::DisabilityCompensation` and `Form526JobStatus`) but these aren't RRD-specific.

Maybe you mean the RapidReadyForDecision processor selector? That would make sense as it gets the `form526_submission`, but I'm not sure if the Sidekiq memory context would be an issue...

Separately, `self.rrd_claim_processed?` would be a class method and wouldn't have access to form submission context, unless there's something about Ruby I'm misunderstanding...

Thanks!

yoomlam: `Form526BaseJob` was added yesterday in #9207. You could put move it to `ProcessorSelector` but that's not where I'd first look for a method that checks if the RRD job successfully processed the claim.

I was thinking the method signature would be `self.rrd_claim_processed?(submission)`.

app/services/rapid_ready_for_decision/hypertension_upload_manager.rb
allthesignals: @yoomlam So... yes, you're right. _But_ I had some weird reasoning based on early naive thinking about jobs and asynchrony that might be interesting: the actual date/time when processing starts is when the class is initialized. I assumed this was the "most accurate" because that date/time could be referenced later by a job. But upon further examination of the actual flow, this doesn't make any sense...

Anyway, I wrote this out mostly for myself but now I have a better understanding of how things fit together. (Seeing some relevant comments downstream, too).

yoomlam: Hmmm. I was considering suggesting the `submission.updated_at` date but there is a chance the job may retry over several days. Ideally, we should use the time that the lighthouse query is performed. We could somehow either record that timestamp or send it to this class, but I don't know if that additional complexity is worth the increased precision. I'd revisit the ticket motivating this PR to see how precise we need to be.

allthesignals: > Hmmm. I was considering suggesting the `submission.updated_at` date but there is a chance the job may retry over several days. Ideally, we should use the time that the lighthouse query is performed. We could somehow either record that timestamp or send it to this class, but I don't know if that additional complexity is worth the increased precision. I'd revisit the ticket motivating this PR to see how precise we need to be.

Yes, I was thinking exactly the same thing. I did some digging yesterday and the spirit of this change is to add some level of uniqueness to the PDF filename. I'm pinging researchers/designers on it to make sure! Still, we'd ideally grab the time the query was performed.

app/mailers/views/hypertension_fast_track_pilot_mailer.erb.html
```diff
@@ -5,7 +5,7 @@
   </head>
   <body>
     <p>
-      <% if @submission.form_json.include? "VAMC_Hypertension_Rapid_Decision_Evidence.pdf" %>
+      <% if @has_rrd_claim_processed %>
```
yoomlam, line 8: This reads better as:
```suggestion
      <% if @rrd_claim_processed %>
```

spec/services/rapid_ready_for_decision/hypertension_upload_manager_spec.rb
```diff
@@ -26,13 +26,24 @@
...
+
   describe '#add_upload(confirmation_code)' do
     context 'success' do
       it 'appends the new upload and saves the expected JSON' do
         RapidReadyForDecision::HypertensionUploadManager.new(form526_submission).add_upload('fake_confirmation_code')
         parsed_json = JSON.parse(form526_submission.form_json)['form526_uploads']
+
         expect(parsed_json).to match original_form_json_uploads + [
-          { 'name' => 'VAMC_Hypertension_Rapid_Decision_Evidence.pdf',
+          { 'name' => 'VAMC_Hypertension_Rapid_Decision_Evidence--10/10/2020-10/10/2021.pdf',
```
yoomlam, line 46: Let's get rid of the slashes and make it sortable:
```suggestion
          { 'name' => 'VAMC_Hypertension_Rapid_Decision_Evidence--20201010-20211010.pdf',
```
or
```ruby
          { 'name' => 'VAMC_Hypertension_Rapid_Decision_Evidence--2020-10-10--2021-10-10.pdf',
```
Also, do we need to include both the start and end dates?

app/services/rapid_ready_for_decision/hypertension_upload_manager.rb
```diff
@@ -4,15 +4,25 @@ module RapidReadyForDecision
...
 
+    DOCUMENT_TITLE = 'VAMC_Hypertension_Rapid_Decision_Evidence'
+
     def initialize(submission)
       @submission = submission
     end
 
+    def file_upload_name
+      today = Time.zone.today
+      end_date = today.strftime('%m/%d/%Y')
```
yoomlam, line 15:
```suggestion
      end_date = Time.zone.today.strftime('%m/%d/%Y')
```

app/services/rapid_ready_for_decision/hypertension_upload_manager.rb
```diff
@@ -4,15 +4,25 @@ module RapidReadyForDecision
   class HypertensionUploadManager
     attr_accessor :submission
 
+    DOCUMENT_TITLE = 'VAMC_Hypertension_Rapid_Decision_Evidence'
+
     def initialize(submission)
       @submission = submission
     end
 
+    def file_upload_name
```
yoomlam, line 13: It looks like you're calling `file_upload_name` multiple times. There's a minuscule chance that you'll get a different value if the job is running at midnight when the date changes. A solution is to memoize the return value.

yoomlam (changes requested):

spec/services/rapid_ready_for_decision/hypertension_upload_manager_spec.rb
allthesignals: > Let's get rid of the slashes and make it sortable:
> 
> or
> 
> ```ruby
>           { 'name' => 'VAMC_Hypertension_Rapid_Decision_Evidence--2020-10-10--2021-10-10.pdf',
> ```
> 
> Also, do we need to include both the start and end dates?

It's what the ticket asks for as far as representing the span of time the search was made. What is your concern?

app/services/rapid_ready_for_decision/hypertension_upload_manager.rb
allthesignals: Thanks!

allthesignals: That is one reason why I had `date` as an instance variable.

spec/services/rapid_ready_for_decision/hypertension_upload_manager_spec.rb
yoomlam: Concerns:
1. Slashes in filenames can cause problems in old OSes and can cause bugs if not treated properly.
2. Having both dates is a possible source of confusion.
3. Opinion: making the date string sortable is generally a good idea, so `YYYY-MM-DD`.

allthesignals: @yoomlam Oh, I see — I agree on all those points. I was asking about your concern specifically over having both dates.

spec/services/rapid_ready_for_decision/hypertension_upload_manager_spec.rb
allthesignals: @yoomlam do you have any issue with this approach to testing this? Is it overkill? I know testing with dates can be a source of flakiness and wanted to avoid it!

spec/services/rapid_ready_for_decision/hypertension_upload_manager_spec.rb
allthesignals: Following up with design to make sure!

app/services/rapid_ready_for_decision/hypertension_upload_manager.rb
```diff
@@ -4,15 +4,24 @@ module RapidReadyForDecision
...
+      end_date = Time.zone.today.strftime('%m%d%Y')
+      start_date = 1.year.ago.strftime('%m%d%Y')
+
+      @file_upload_name ||= "#{DOCUMENT_TITLE}--#{start_date}-#{end_date}.pdf"
```
yoomlam, line 17:
```suggestion
      @file_upload_name ||= begin
        end_date = Time.zone.today.strftime('%m%d%Y')
        start_date = 1.year.ago.strftime('%m%d%Y')
        "#{DOCUMENT_TITLE}--#{start_date}-#{end_date}.pdf"
      end
```

spec/services/rapid_ready_for_decision/hypertension_upload_manager_spec.rb
yoomlam: You should use a time zone whenever creating a timestamp (details: https://thoughtbot.com/blog/its-about-time-zones) to ensure your test setup is using the same zone as your implementation.
Also all the [examples](https://github.com/travisjeffery/timecop) I've seen for `Timecop.freeze` passes in a Time object rather than a string.

yoomlam (changes requested):

app/services/rapid_ready_for_decision/hypertension_upload_manager.rb
allthesignals: @yoomlam very cool, thanks — didn't know about this Ruby feature. What is the difference as far as behavior, though? Or is it just clearer? My understanding is that `@file_upload_name ||= "#{DOCUMENT_TITLE}--#{start_date}-#{end_date}.pdf"` would still use the memoized value, although it's true that the datetime values would still have been recomputed on subsequent calls.... thanks!

yoomlam: > although it's true that the datetime values would still have been recomputed on subsequent calls

Yea, moving that code in the `begin` block saves a few cpu cycles and needless string computation. Search for `||= begin` in other parts of the code for other examples.

spec/services/rapid_ready_for_decision/hypertension_upload_manager_spec.rb
allthesignals: Yep, this is lifted from other places in the code but we should improve the approach. I'll update!

yoomlam (approved):
👍 Presuming we want to have both start and end dates.

spec/services/rapid_ready_for_decision/hypertension_upload_manager_spec.rb
```diff
@@ -26,13 +26,24 @@
...
+
   describe '#add_upload(confirmation_code)' do
     context 'success' do
       it 'appends the new upload and saves the expected JSON' do
         RapidReadyForDecision::HypertensionUploadManager.new(form526_submission).add_upload('fake_confirmation_code')
         parsed_json = JSON.parse(form526_submission.form_json)['form526_uploads']
+
         expect(parsed_json).to match original_form_json_uploads + [
-          { 'name' => 'VAMC_Hypertension_Rapid_Decision_Evidence.pdf',
+          { 'name' => 'VAMC_Hypertension_Rapid_Decision_Evidence--20211010.pdf',
```
yoomlam, line 46: Sorry, one thing: Since there's now only one date, do you want to use a single hyphen?

yoomlam (approved):
One question

spec/services/rapid_ready_for_decision/hypertension_upload_manager_spec.rb
allthesignals: Seems reasonable!

yoomlam (approved):
🎉

ericboehs (approved):

PR #9297, opened 2022-03-08
2580 phone parser

kpethtel:
## Description of change
This ticket was originally for extracting the three instances of phone number regex parsing into a helper method. I did that work, but as I was working I was looking at real data from QA and realized that our VCR cassettes for appointment requests didn't quite match. The phone numbers coming back from QA are simple numerical strings, like "1112223333". They don't have any dashes, spaces, or parentheses. So I believe the VAOS VCRs are out of date. Which doesn't matter for their purposes because they just pass the phone number straight through without modification. That caused me to look more closely and I realized that we didn't need to add any new phone number parsing for appointment requests. I had mistakenly put the patient's phone number into the facility data. So now I'm fixing that mistake. 

## Original issue(s)
department-of-veterans-affairs/va-mobile-app#2580

Removes unnecessary phone parsing.


jperk51 (approved):
LGTM

modules/mobile/app/models/mobile/v0/templates/community_care_appointment.rb
```diff
@@ -41,26 +41,11 @@ def location
             },
             lat: nil,
             long: nil,
-            phone: phone_captures,
```
kpethtel, line 44: This was done in error. The phone number is the patient's, not the facility's. The patient's phone number is also captured in these responses as `patient_phone_number` but that is simply passed through and not parsed. I did some testing on staging and found that the front end does its own phone number validation. It only allows 10 digit numbers (typical US phone numbers) and returns them as numerical strings (i.e., with any dashes or other characters removed). The web app also displays the patient phone numbers as a plain numerical string, so the way we're handling it should be consistent.

aherzberg (approved):
Looks good!

PR #9302, opened 2022-03-08
Refactor common Account.lookup_by_user_uuid code

yoomlam:
## Description of change

Refactors common `Account` lookup logic so that changes only need to be made in one place.

## Original issue(s)

There is common logic [here](https://github.com/department-of-veterans-affairs/vets-api/blob/1dca6957adc8fd4cea63696fa3839e8e2462e982/app/workers/rapid_ready_for_decision/disability_compensation_job.rb#L87-L89) and [here](https://github.com/department-of-veterans-affairs/vets-api/blob/05ef844b76be5a3d4272b62a60bd4d3f1f07fc2f/app/workers/education_form/process10203_submissions.rb#L61-L64) that looks for the `Account` associated with the user. Let's refactor so that changes to `Account` lookup only need to be made in one place.

Slack discussion: initiated [here](https://dsva.slack.com/archives/CE4304QPK/p1646687999288969?thread_ts=1644946422.907099&cid=CE4304QPK) and answered [here](https://dsva.slack.com/archives/CSFV4QTKN/p1646777828230039)

Initially tried to use `find_by_user_uuid` (as suggested) but the linter wants to change it to `find_by(user_uuid: ...)`, so the method name is now `lookup_by_user_uuid`.


bosawt (approved):
Looks good to me, thanks for the specs too!

bosawt (approved):

thilton-oddball (approved):

PR #9305, opened 2022-03-09
Fix TypeError bug caused by #9206

nanotone:
## Description of change
This PR fixes a bug introduced by #9206 that causes uncaught TypeError exceptions to occur in `Form526Submission#workflow_complete_handler` and fill up the Sidekiq retry queue.

The bug is caused by the fact that `#workflow_complete_handler` is an instance method, but as a Sidekiq callback it gets called without being bound to an ActiveRecord model. Therefore, calling `#personalization_parameters` and therefore `#form` on the self-object will result in ActiveRecord attributes like `#form_json` to be nil.

The immediate fix is to call `#personalization_parameters` on the ActiveRecord model, `submission`. A long-term fix would be to structure the code so that the Sidekiq callbacks (there are several in this region) live in separate classes that make this subtle distinction obvious -- this was @yoomlam's intention as well.


app/models/form526_submission.rb
```diff
@@ -338,6 +337,8 @@ def personalization_parameters(first_name)
     }
   end
 
+  private
```
nanotone, line 340: I made `#personalization_parameters` public so that the bugfix actually works. It's a small and uneventfully read-only method. We will refactor this in a later PR.

yoomlam (approved):
👍

yoomlam (approved):

rileyanderson (approved):

PR #9308, opened 2022-03-09
2688 add payment history to user service dictionary

cadibemma:
## Description of change
User Story
As a veteran I'd like to know if I'm authorized to view payment history ahead of time before I try to access the screen.

Description
The payment history feature requires a [user to have BGS access](https://github.com/department-of-veterans-affairs/vets-api/blob/278d922d3edb92a846cb2c4ac269408cf9967da5/app/controllers/v0/profile/payment_history_controller.rb#L6). The way we communicate user auth status for features in the mobile app is through the [service dictionary/hash](https://github.com/department-of-veterans-affairs/vets-api/blob/800cbe7a67287fa38f1d0ec03237894b06c91de4/modules/mobile/app/serializers/mobile/v0/user_serializer.rb#L41-L51). Add an entry for payment_history protected by the :bgs flag.

## Original issue(s)
department-of-veterans-affairs/va-mobile-app#2688


kpethtel (commented):
Looking at these specs, it seems like we sometimes have specific tests for "when user doesn't have access to this service". Should we have a test for "when user doesn't have access to bgs"?

cadibemma:
> Looking at these specs, it seems like we sometimes have specific tests for "when user doesn't have access to this service". Should we have a test for "when user doesn't have access to bgs"?

Yup! I added a test to test access to bgs

thilton-oddball (approved):

PR #9309, opened 2022-03-09
fix v1 user linking

jperk51:
## Description of change
Update id in docs so v1/user links to its details correctly

## Original issue(s)
N/A

Docs only change


kpethtel (approved):
👍

PR #9312, opened 2022-03-10
38160 Removing NoK confirmation feature flag

kanchanasuriya:
## Description of change

Removing `check_in_experience_chip_service_nok_confirmation_update_enabled feature` flag and cleaned up the code behind the flag
## Original issue(s)
department-of-veterans-affairs/va.gov-team#38160

- [x] rspec
- [x] rubocop


randomsync (dismissed):
lgtm

thilton-oddball (changes requested):
It looks like there is a reference to the feature flag in modules/check_in/spec/request/v2/pre_check_ins_request_spec.rb on line 224 that needs to be removed as well.

kanchanasuriya:
> It looks like there is a reference to the feature flag in modules/check_in/spec/request/v2/pre_check_ins_request_spec.rb on line 224 that needs to be removed as well.

Thanks for pointing it out.  Updated in the latest commit

thilton-oddball (approved):

PR #9322, opened 2022-03-11
API-13844: Update validations and arguments passed to BGS update_birls_record

hinzed1127:
## Description of change

- We updated the `update_birls_record` method in the `bgs_ext` gem
- This makes the necessary changes to that method call in `vets-api`
- Also fixes the bugs described [here](https://github.com/department-of-veterans-affairs/vets-api/issues/9341)

## Original issue(s)
https://vajira.max.gov/browse/API-13844

- Dependent on the following PR to the BGS gem: department-of-veterans-affairs/bgs-ext#73
   - Updated tests are passing when pulling in the gem with my PR branch


modules/claims_api/app/workers/claims_api/poa_updater.rb
```diff
@@ -7,24 +7,28 @@ module ClaimsApi
...
-        ssn: poa_form.auth_headers['va_eauth_pnid'],
+        file_number: file_number,
+        ssn: ssn,
         poa_code: poa_form.form_data['serviceOrganization']['poaCode']
       )
 
       if response[:return_code] == 'BMOD0001'
         poa_form.status = ClaimsApi::PowerOfAttorney::UPDATED
+        # Clear out the error message if there were previous failures
+        poa_form.vbms_error_message = nil if poa_form.vbms_error_message.present?
```
hinzed1127, line 29: When debugging some of the jobs that appeared to have initially failed but then succeeded.  This change clears out the error message so that we don't see outdated errors.

In addition to the error message, it looked like a bunch of POAs still at a status of errorred. This makes less sense to me, since we're already changing the status to updated.

My only theory is that by calling [`poa_form.save` instead of `poa_form.save!` below](https://github.com/department-of-veterans-affairs/vets-api/blob/master/modules/claims_api/app/workers/claims_api/poa_updater.rb#L33), it's failing silently instead of raising the issue, but that seems half-baked. Thoughts?

jw81: The `.save!` is possible but does seem unlikely.
The only thing I can think of is that those that are still errored never made it past this part.  Or maybe they failed at a different point further down then road.

modules/claims_api/app/workers/claims_api/poa_updater.rb
```diff
@@ -7,24 +7,28 @@ module ClaimsApi
...
 
-        if enable_vbms_access?(poa_form: poa_form)
-          veteran_participant_id = poa_form.auth_headers['va_eauth_pid']
-          ClaimsApi::VBMSUpdater.perform_async(poa_form.id, veteran_participant_id)
-        end
+        ClaimsApi::VBMSUpdater.perform_async(poa_form.id) if enable_vbms_access?(poa_form: poa_form)
```
jw81, line 31: This change loses the 2nd arg (`veteran_participant_id`) to the `VBMSUpdater` job.
I believe any calls to `ClaimsApi::VBMSUpdater.perform_async()` in the specs should be blowing up because of this.

jw81 (changes requested):
This looks like this should be broken. ([ref](https://github.com/department-of-veterans-affairs/vets-api/pull/9322/files#r824818510))

modules/claims_api/app/workers/claims_api/poa_updater.rb
hinzed1127: I'm trying to understand the reasoning to avoid `.save!` outlined [here](https://stackoverflow.com/questions/5056385/when-to-use-save-vs-save-in-model), but in the meantime, I think we'll see if it continues to be an issue and tackle it accordingly

modules/claims_api/app/workers/claims_api/poa_updater.rb
hinzed1127: [fixed](https://github.com/department-of-veterans-affairs/vets-api/pull/9322/commits/d944c17824f019fa4d0dea8252974f1325c3331f)

modules/claims_api/app/workers/claims_api/poa_updater.rb
```diff
@@ -7,19 +7,26 @@ module ClaimsApi
...
-    def perform(power_of_attorney_id)
+    def perform(power_of_attorney_id) # rubocop:disable Metrics/MethodLength
       poa_form = ClaimsApi::PowerOfAttorney.find(power_of_attorney_id)
       service = BGS::Services.new(
         external_uid: poa_form.external_uid,
         external_key: poa_form.external_key
       )
+
+      ssn = poa_form.auth_headers['va_eauth_pnid']
+      file_number = service.people.find_by_ssn(ssn)[:file_nbr] # rubocop:disable Rails/DynamicFindBy
```
jw81, line 18: Just jotting this down as a reminder to us.
But we might find that we'll have to come back and add some kind of logic to handle the case where the `FileNumber` returned is `nil`.
It shouldn't really be a problem going forward since you've added the `check_file_number_exists!` to the POST.  So now the POST and the PUT have that protection and no new submissions should even be able to make it here if a `FileNumber` can't be found.  But if we retry some of the past submissions and they still fail, then it's likely because a `FileNumber` wasn't able to be found here.

jw81 (dismissed):

jw81 (approved):

ericboehs (approved):

PR #9333, opened 2022-03-11
API-13787 Add tests for nod v2 evidence submission

ElleeB:
Discovered we'd never created tests for NOD v2 Evidence Submissions!


cilestin (approved):
Woah! Excellent catch!

PR #9344, opened 2022-03-14
Monitor for and offramp RRD Asthma claims

nanotone:
## Description of change
This PR adds a Form526AsthmaJob into the RRD flow, to monitor for incoming single-issue asthma claims for increase. The new job fires off an email to the RRD team's `event_tracking`, and immediately offramps the claim from RRD logic so it goes through the usual non-RRD flow.

## Original issue(s)
https://www.notion.so/Dashboard-RRD-Health-Evidence-24d77d9988a546c3b19a3c73e7ced29b?p=cabaada76e8742768b3101659b0522fa

* No data related to the claim itself is emailed out, just that it was detected and offramped
* Feature flag is `disability_asthma_compensation_fast_track`
* New test and test setup added, to mirror those of the existing `Form526HypertensionJob`


app/workers/rapid_ready_for_decision/form526_base_job.rb
```diff
@@ -26,7 +26,7 @@ def perform(form526_submission_id)
       begin
         with_tracking(self.class.name, form526_submission.saved_claim_id, form526_submission_id) do
           assessed_data = assess_data(form526_submission)
-          return if assessed_data.nil?
+          next if assessed_data.nil?
```
nanotone, line 29: This was a great opportunity to add a test, for checking that lack of data properly offramps the claim.

The test revealed that `return if assessed_data.nil?` would skip the after-yield code in `with_tracking`, thereby leaving the job in a `try` state rather than moving it over to `success`. In Ruby, `next` is the way to go if you just want to skip to the end of the block.

app/workers/rapid_ready_for_decision/form526_asthma_job.rb
```diff
@@ -0,0 +1,32 @@
...
+    sidekiq_options retry: 2
+
+    def assess_data(form526_submission)
+      client = lighthouse_client(form526_submission)
+      client.list_resource('medication_requests')
+
+      ActionMailer::Base.mail(
+        from: ApplicationMailer.default[:from],
+        to: Settings.rrd.event_tracking.recipients,
+        subject: 'Asthma RRD Claim - Offramped',
```
yoomlam, line 18: For consistency with [recent PR](https://github.com/department-of-veterans-affairs/vets-api/pull/9349/files#diff-ed1e5d9b851dfdcbdc20273d00ded834be82a60b2df94d4dcc24482c858ac968R11)
```suggestion
        subject: 'RRD claim - Offramped',
```

app/workers/rapid_ready_for_decision/form526_asthma_job.rb
```diff
@@ -0,0 +1,32 @@
...
+
+module RapidReadyForDecision
+  class Form526AsthmaJob < Form526BaseJob
+    STATSD_KEY_PREFIX = 'worker.fast_track.form526_asthma_job'
+
+    sidekiq_options retry: 2
+
+    def assess_data(form526_submission)
+      client = lighthouse_client(form526_submission)
+      client.list_resource('medication_requests')
```
yoomlam, line 13: Do we want to include some aspect of `medication_requests` result in the body of the message? If not, why are we making a query here?

app/workers/rapid_ready_for_decision/form526_base_job.rb
yoomlam: Nice catch!

> leaving the job in a `try` state 

How did you determine the state of the job?

config/features.yml
```diff
@@ -185,7 +185,11 @@ features:
     description: Send direct deposit emails using VaNotify service
   disability_hypertension_compensation_fast_track:
     actor_type: user
-    description: Fast tracks 526 disability compensation claims by submitting additional health data
+    description: Fast tracks 526 disability compensation hypertension claims by submitting additional health data
+    enable_in_development: true
+  disability_asthma_compensation_fast_track:
```
yoomlam, line 190: I'd like to start renaming these Flipper keys to `rrd_#{disability}_compensation`
(as noted [here](https://github.com/department-of-veterans-affairs/vets-api/blob/d9e3a05a2e03a5089b18aba1bf9119f4d1466397/app/services/rapid_ready_for_decision/processor_selector.rb#L41)), so we can then add other RRD-specific Flipper keys in one contiguous section, where all entries begin with prefix `rrd_`.
Should we start doing it in this PR? 
I can add `rrd_hypertension_compensation` in a separate PR and enable it in prod (along with code changes to look for both Flipper flags), and then follow-up with another PR to remove `disability_hypertension_compensation_fast_track`.

spec/jobs/rapid_ready_for_decision/form526_asthma_job_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    EVSS::DisabilityCompensationAuthHeaders.new(user).add_headers(EVSS::AuthHeaders.new(user).to_h)
+  end
+  let(:saved_claim) { FactoryBot.create(:va526ez) }
+  let(:submission) do
+    create(:form526_submission, :with_uploads,
+           user_uuid: user.uuid,
+           auth_headers_json: auth_headers.to_json,
+           saved_claim_id: saved_claim.id,
+           submitted_claim_id: '600130094')
+  end
```
yoomlam, line 24: Try with this simpler setup:
```suggestion
  let(:submission) do
    create(:form526_submission, :with_uploads,
           saved_claim_id: saved_claim.id,
           submitted_claim_id: '600130094')
  end
```
or
```suggestion
  let(:submission) do
    create(:form526_submission, :with_uploads)
  end
```

app/workers/rapid_ready_for_decision/form526_asthma_job.rb
nanotone: Good call. I was going to just let it make the roundtrip to Lighthouse, but I'll update this to at least report back the number of returned resources.

config/features.yml
nanotone: Sure thing. Since it's already programmatically constructed in ProcessorSelector, I'll just change the asthma-related instances where it's "hard-coded" and let #9360 handle the rest.

(Unless #9360 merges first, in which case I'll rebase and update everything here to the New Way)

app/workers/rapid_ready_for_decision/form526_base_job.rb
nanotone: [This line](https://github.com/department-of-veterans-affairs/vets-api/blob/d72cc85/spec/jobs/rapid_ready_for_decision/disability_compensation_job_spec.rb#L85) in the spec was failing with observed value `try` rather than `success`.

config/features.yml
```diff
@@ -185,7 +185,11 @@ features:
     description: Send direct deposit emails using VaNotify service
   disability_hypertension_compensation_fast_track:
     actor_type: user
-    description: Fast tracks 526 disability compensation claims by submitting additional health data
+    description: Fast tracks 526 disability compensation hypertension claims by submitting additional health data
+    enable_in_development: true
+  rrd_asthma_compensation:
+    actor_type: user
+    description: Fast tracks 526 disability compensation asthma claims by submitting additional health data
```
yoomlam, line 192: FYI: I'll alphabetize this in #9360.

yoomlam (approved):
Looks good! You'll need to update `rrd_enabled_disability?` with something like https://github.com/department-of-veterans-affairs/vets-api/pull/9360/files#r827434482

yoomlam (approved):
🚀

rileyanderson (approved):

PR #9352, opened 2022-03-15
RRD: Rename DiagnosticCode to Constants

yoomlam:
## Description of change

Instead of operating on the JSON structure itself, map Form526Submission’s disabilities into a list of constants to be passed around in our code and used in conditionals. As a result, our code becomes less hardcoded and less dependent on external Form526Submission form’s JSON structure.

## Original issue(s)
The original `DiagnosticCode` class mapped a disability type to an associated diagnostics code. We want to be able to associate other constants (e.g., Flipper flags, disability text/label used in PDFs) to the disability type.


app/services/rapid_ready_for_decision/constants.rb
```diff
@@ -0,0 +1,25 @@
...
+        flipper_name: 'hypertension'
+      },
+      asthma: {
+        code: 6602,
+        label: 'asthma',
+        flipper_name: 'asthma'
+      }
+    }.freeze
+
+    DISABILITIES_BY_CODE = DISABILITIES.map { |_k, v| [v[:code], v] }.to_h
```
yoomlam, line 18: @allthesignals This is what I ended up doing to rekey the `DISABILITIES` hash by diagnostic code.

app/services/rapid_ready_for_decision/processor_selector.rb
```diff
@@ -7,26 +7,38 @@ def initialize(form526_submission)
...
+
+      # the RRD disability must be enabled
+      return unless rrd_enabled_disability?(disability_type)
 
-      nil
+      form_disability = form_disabilities.first
+      # the submitted disability must be for a claim for increase
+      return unless self.class.disability_increase?(form_disability, disability_type)
+
+      # TODO: use a constant hash instead
```
nanotone, line 26: Even better, why not just make this a value in each disability type? So, just

```ruby
return disability_type[:processor_class]
```

Maybe that's what you meant. Anyway, 👌  to do this later.

nanotone (approved):
Looks great! Just dropped in some thoughts, not really even suggestions.

app/services/rapid_ready_for_decision/processor_selector.rb
yoomlam: I considered that, but I prefer not to put references to actual classes in the `Constants` -- only have primitive types -- to avoid circular dependencies. 
For the `TODO`, I meant something like the following defined in this class:
```ruby
DISABILITY_TO_PROCESSOR_CLASS = {
  RapidReadyForDecision::Constants::DISABILITIES[:hypertension] =>
    RapidReadyForDecision::DisabilityCompensationJob, 
  RapidReadyForDecision::Constants::DISABILITIES[:asthma] =>
    RapidReadyForDecision::Form526AsthmaJob
}.freeze
```
but it looks a bit messy.

app/services/rapid_ready_for_decision/processor_selector.rb
```diff
@@ -7,26 +7,38 @@ def initialize(form526_submission)
...
-      end
+      # single-issue claims only
+      return unless disability_types.size == 1
 
-      if rrd_enabled_disability?('asthma') && single_issue_claim_applicable?(DiagnosticCodes::ASTHMA)
-        return RapidReadyForDecision::Form526AsthmaJob
-      end
+      disability_type = disability_types.first
+      # must be a disability supported by RRD
+      return unless disability_type
```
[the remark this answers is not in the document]
yoomlam: I can see the value in that. Hmmm. There's enough occurrences of the non-explicit version in other parts of the code, so I'll keep it as is for now.

app/mailers/hypertension_fast_track_pilot_mailer.rb
```diff
@@ -3,7 +3,7 @@
 class HypertensionFastTrackPilotMailer < ApplicationMailer
   def build(submission)
     @submission = submission
-    @disability = 'hypertension'
+    @disability = RapidReadyForDecision::Constants::DISABILITIES[:hypertension]
```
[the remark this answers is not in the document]
yoomlam: The label is resolved in the associated `erb.html` file: https://github.com/department-of-veterans-affairs/vets-api/pull/9352/files#diff-abd8c62d02d226d4468607485e43284477eea8b399e13a0d32adab77d645a230R9

app/services/rapid_ready_for_decision/processor_selector.rb
nanotone: Yeah I did think about circular dependencies with the `Constants` route. Luckily, there's a way to do it with primitive types, so you get the best of both worlds:

```ruby
> 'RapidReadyForDecision::Form526AsthmaJob'.constantize
=> RapidReadyForDecision::Form526AsthmaJob
```

yoomlam:
@LindseySaari Is there anything blocking approval?

app/services/rapid_ready_for_decision/processor_selector.rb
yoomlam: Here's what I ended up doing: #9403

yoomlam:
@rileyanderson Could I get a review?

yoomlam:
Closing this in favor of #9403.

PR #9356, opened 2022-03-15
Fix flake in Lighthouse Health API client tests

nanotone:
This PR fixes a flakey test introduced in #9327, which renamed the `medications` resource type to the more accurate `medication_requests`.

The original code randomly chose a resource type to run the test, causing the stale code to fail only 50% of the time. The new code parametrizes the test to run on all given resource types.

Pro-tip: the diff is quite small and readable when hiding whitespace edits.


yoomlam (approved):
Thanks for catching this!

ericboehs (approved):

PR #9359, opened 2022-03-15
API-14220: Improve error handling for VBMSUpdater job

hinzed1127:
## Description of change

- Simplifies the argument to `VBMSUpdater` by passing just a POA id and doing the participant ID lookup in the job
- checks the BGS response and adds/removes and errors accordingly

## Original issue(s)
https://vajira.max.gov/browse/API-14220


modules/claims_api/app/workers/claims_api/vbms_updater.rb
```diff
@@ -7,18 +7,30 @@ module ClaimsApi
...
-        participant_id: participant_id,
+
+      response = service.corporate_update.update_poa_access(
+        participant_id: poa_form.auth_headers['va_eauth_pid'],
         poa_code: poa_form.form_data.dig('serviceOrganization', 'poaCode'),
         allow_poa_access: 'y',
         allow_poa_c_add: allow_address_change?(poa_form) ? 'y' : 'n'
       )
+
+      if response[:return_code] == 'GUIE50000'
```
jw81, line 24: Any chance we've confirmed with BGS that this is the appropriate value to look for?

hinzed1127: this was based off of what we're getting back from current [successful call we test against](https://github.com/department-of-veterans-affairs/bgs-ext/blob/master/spec/bgs/services/corporate_update_spec.rb#L36-L37), but I will double check with the BGS team to validate

jw81 (approved):

PR #9362, opened 2022-03-15
[38588] Creating code container object to map code to user account uuid

bosawt:
## Description of change
This PR creates an object called Code Container that maps a code generated by vets-api with a user_account_uuid that will be used to log the user in

## Original issue(s)
department-of-veterans-affairs/va.gov-team#38588

- Should just be able to test by creating `SignIn::CodeContainer.new(code_challenge: 'code-challenge', code: 'code', user_account_uuid: 'user-account-uuid')`
- You can then look it up in the Redis with `SignIn::CodeContainer.find('code')`


bramleyjl (approved):

app/models/sign_in/code_container.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module SignIn
+  class CodeContainer < Common::RedisStore
+    redis_store REDIS_CONFIG[:sign_in_code_container][:namespace]
```
ericboehs, line 5: Should we dig here in case `REDIS_CONFIG[:sign_in_code_container]` is nil?

ericboehs (approved):

PR #9372, opened 2022-03-16
API-14203 charactervalidation expansion

cilestin:
## Description of change
Removes the CharacterValidation controller concern and instead accepts any characters in the request, and delegates character encoding normalization as close to the problem as possible - the PDF generation.

## Original issue(s)
https://vajira.max.gov/browse/API-14203

* Adds `attr_readonly` to all appeal models to ensure we cannot override any data that might be modified during pdf generation
* Cleans up some files that were not used (10182 character fixtures)
* Removes character encoding section of the dev documentation


modules/appeals_api/spec/services/appeals_api/pdf_construction/generator_spec.rb
```diff
@@ -145,6 +145,28 @@
...
           end
         end
+
+        context 'special character verification' do
+          it 'allows certain typography characters into Windows-1252' do
+            hlr = build(:minimal_higher_level_review)
+            hlr.form_data['included'][0]['attributes']['issue'] = 'Smartquotes: “”‘’'
+            hlr.save!
+            path = described_class.new(hlr, version: 'V2').generate
+            generated_pdf = PDF::Reader.new(path)
```
ElleeB, line 155: I wonder if naming this something different would help differentiate between the generated_pdf that we compare with expected_pdf. It seems like this plays a different role.

modules/appeals_api/spec/services/appeals_api/pdf_construction/generator_spec.rb
```diff
@@ -145,6 +145,28 @@
...
+            expect(generated_pdf.pages[1].text).to include 'Smartquotes: “”‘’'
+            File.delete(path) if File.exist?(path)
+          end
+
+          it 'removes characters that fall outsize Windows-1252 charset that cannot be downgraded' do
+            hlr = build(:minimal_higher_level_review)
+            hlr.form_data['included'][0]['attributes']['issue'] = '∑mer allergies'
+            hlr.save!
+            path = described_class.new(hlr, version: 'V2').generate
+            generated_pdf = PDF::Reader.new(path)
```
ElleeB, line 165: Same as above.

ElleeB (commented):
Love this! I have one wondering about naming in the comments :)

ElleeB (changes requested):
I have one wondering about naming in the comments :)

stiehlrod (approved):
LGTM

ElleeB (approved):
Looks good!!

PR #9379, opened 2022-03-16
Fix mirror account availability logs in big query job

holdenhinkle:
## Description of change
Fix a couple of minor things to push TUD accounts data to BigQuery.


zachclarity:
@rileyanderson  Would you review this.  Thanks.

app/workers/test_user_dashboard/mirror_accounts_in_big_query.rb
```diff
@@ -19,9 +19,13 @@ def mirror_tud_accounts_in_bigquery
...
     def accounts
-      TestUserDashboard::TudAccount.all.each.with_object([]) do |account, rows|
-        rows << account.attributes.reject { |attr, _| %w[id password].include?(attr) }
+      tud_accounts = []
+
+      TestUserDashboard::TudAccount.all.each do |tud_account|
+        tud_accounts << tud_account.attributes.reject { |attr, _| %w[id password].include?(attr) }
       end
+
+      tud_accounts.each { |row| row['services'] = [''] if row['services'].nil? }
```
rileyanderson, line 28: Would something like this work? It would avoid having to loop through all the accounts twice. Thoughts?

```suggestion
      TestUserDashboard::TudAccount.all.as_json(except: %i[id password]).each do |account|
        account['services'] = '' if account['services'].nil?
      end
```

holdenhinkle: I just pushed that. I didn't know about `as_json` and `except`. Nice.

holdenhinkle:
I think we're good to go now @rileyanderson ...

rileyanderson (approved):

bramleyjl (approved):

PR #9380, opened 2022-03-16
API-12495 Update CMPortal submission log to include all metadata

ElleeB:
It's been reported that our appeal submission metadata does not include the line of business as we expect.

In hopes of gaining more info about the state of the metadata once submitted to central mail, and to add additional info to our logs, this PR modifies the central mail submission log to included the **entire appeal's metadata**.


ElleeB:
> How are the logs used? Are reports generated from them? I am curious if removing the UUID, or the way UUID is accessed, might have adverse effects. Another example, if someone has a query running against the logs (in AWS) and it searches for UUID, this change might prevent these logs from showing up.
> 
> Maybe we could leave the original line 33 `'uuid' => metadata['uuid'],`

Nice thought. The other thing I can do is just add each metadata key individually, that way, if anyone else is actually using these logs, we'd maintain the ability to query for anything that was there before.

lib/central_mail/utilities.rb
```diff
@@ -29,10 +29,9 @@ def log_submission(uploaded_object, metadata)
       pdf_total = number_pages.count
 
       title = uploaded_object.class.to_s.split('::').first
+
       Rails.logger.info("#{title}: Submission success",
-                        'uuid' => metadata['uuid'],
-                        'source' => metadata['source'],
-                        'docType' => metadata['docType'],
+                        'metadata' => metadata,
```
cilestin, line 34: we could just pass metadata.merge({consumer_id: uploaded_object.consumer_id, ...}) into Rails.logger.info. That way all the keys will still be there and the extra data can be inserted at the same time, rather than specifically outputting the keys & values individually.

ElleeB: yeah, not going to do it individually. i just broke my pr. hold please.

PR #9385, opened 2022-03-16
API-12495 Update submission log to include all metadata

ElleeB:
Unintentionally closed original here: https://github.com/department-of-veterans-affairs/vets-api/pull/9380


lib/central_mail/utilities.rb
```diff
@@ -29,14 +29,17 @@ def log_submission(uploaded_object, metadata)
...
-                        'pdfCount' => pdf_total)
+      log_details = {
+        'uuid' => metadata['uuid'],
+        'source' => metadata['source'],
+        'docType' => metadata['docType'],
+        'consumer_id' => uploaded_object.consumer_id,
+        'consumer_username' => uploaded_object.consumer_name,
+        'pageCount' => page_total,
+        'pdfCount' => pdf_total
+      }.merge(metadata)
```
cilestin, line 40: As we discussed elsewhere, let's make sure to slice the metadata we want to keep, since there's potentially PII we shouldn't be logging by a full merge.

cilestin (changes requested):

lib/central_mail/utilities.rb
ElleeB: I ultimately went with `slice`, but considered `except` for a minute. Slice seems safer, but we'll have to make a note somewhere if we add to our metadata in the future and expect it in our logs.

stiehlrod (dismissed):
LGTM

cilestin (approved):

PR #9391, opened 2022-03-17
Api 13721 create pdf examples

stiehlrod:
[API-13721](https://vajira.max.gov/browse/API-13721)

## Description of change
Generates example PDF that utilizes the max character length for every data field so we can ensure there is no overlap.

* Adds [JSON_Schemer](https://github.com/davishmcclurg/json_schemer)
* Adds `config.filter_run :focus` to spec_helper to run one test at a time.


modules/appeals_api/spec/services/appeals_api/pdf_construction/generator_spec.rb
```diff
@@ -161,11 +161,20 @@
...
         it 'generates the expected pdf' do
-          generated_pdf = described_class.new(extra_supplemental_claim, version: 'V2').generate
-          expected_pdf = fixture_filepath('expected_200995_extra.pdf', version: 'v2')
+          sc = build(:extra_supplemental_claim)
+          sc.form_data = SchemaHelpers.supplemental_claim_w_fill(sc.form_data, max_lengths)
+          sc.save!
+          generated_pdf = described_class.new(sc, version: 'V2').generate
+          expected_pdf = fixture_filepath('expected_200995_maxlength.pdf', version: 'v2')
+
+          `cp #{generated_pdf} #{expected_pdf}`
```
ElleeB, line 174: We should remove both the copy and open here.

modules/appeals_api/spec/services/appeals_api/pdf_construction/generator_spec.rb
```diff
@@ -161,11 +161,20 @@
...
+        let(:max_lengths) { find_max_lengths '../../schemas/v2/200995.json', 'maxLength' }
 
         it 'generates the expected pdf' do
-          generated_pdf = described_class.new(extra_supplemental_claim, version: 'V2').generate
-          expected_pdf = fixture_filepath('expected_200995_extra.pdf', version: 'v2')
+          sc = build(:extra_supplemental_claim)
+          sc.form_data = SchemaHelpers.supplemental_claim_w_fill(sc.form_data, max_lengths)
+          sc.save!
+          generated_pdf = described_class.new(sc, version: 'V2').generate
+          expected_pdf = fixture_filepath('expected_200995_maxlength.pdf', version: 'v2')
```
ElleeB, line 172: The Ws in the telephone field are cut off, but I think it has something to do with printing the extension...which I don't think we actually do for domestic phone numbers. We'll have to touch base about this one.

modules/appeals_api/spec/services/appeals_api/pdf_construction/generator_spec.rb
```diff
@@ -161,11 +161,20 @@
       end
 
       context 'pdf extra content verification' do
```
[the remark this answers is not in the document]
stiehlrod: yep, I did not mean to overwrite the existing test. thanks!

modules/appeals_api/spec/support/schema_helpers.rb
```diff
@@ -15,4 +17,58 @@ def read_schema(filename, schema_version = 'v1')
       )
     )
   end
+  # rubocop:disable Layout/LineLength
+
+  def find_max_lengths(filename, _string_to_find, schema_version = 'V2')
```
ElleeB, line 22: This is awesome. I do wonder if this might become difficult to maintain as version changes occur, but probably not more work than writing new specs etc for new versions.

modules/appeals_api/spec/services/appeals_api/pdf_construction/generator_spec.rb
ElleeB: Update on my comment: I was wrong about the extension. We leave out extension for HLR because it just won't fit, but we include it for supplemental claim, so we'll definitely need to figure out how to deal with that field.

stiehlrod: Would the telephone ext fix be a separate ticket? or should I try to fix it in this one?

ElleeB: Good call. Let's make a ticket so we can make a decision without feeling rushed.

modules/appeals_api/spec/services/appeals_api/pdf_construction/generator_spec.rb
```diff
@@ -170,6 +170,22 @@
           File.delete(generated_pdf) if File.exist?(generated_pdf)
         end
       end
+
+      context 'pdf max length content verification' do
+        include SchemaHelpers
+        let(:max_lengths) { find_max_lengths '../../schemas/v2/200995.json', 'maxLength' }
```
ElleeB, line 176: Can you use `read_schema(filename, schema_version = 'v1')` here?

stiehlrod: i do use it inside of the 'find_max_lengths' method

modules/appeals_api/spec/services/appeals_api/pdf_construction/generator_spec.rb
stiehlrod: ticket created

modules/appeals_api/spec/services/appeals_api/pdf_construction/generator_spec.rb
ElleeB: My thought was that you could define the schema and pass it to find max lengths like:
```
let(:schema) { read_schema '200995.json', 'v2'}
let(:max_lengths) { find_max_lengths schema, 'maxLength' }
```
...and then you already have your schema passed to find_max_lengths. Just thinking it would be nice to remove the long path shenanigans, since that's what the helper method is for :)

modules/appeals_api/spec/support/schema_helpers.rb
```diff
@@ -15,4 +17,58 @@ def read_schema(filename, schema_version = 'v1')
...
+    schema = read_schema(filename, schema_version)
+    max_lengths = {}
+    schema['definitions'].each do |key|
+      if !key[1]['properties'].nil?
+        key[1]['properties'].each do |prop|
+          case key[0]
+          when 'timezone'
+            max_lengths[key[0]] = 'N/A'
+          when 'scCreate'
+            max_lengths['email'] =
```
cilestin, line 32: I feel like there should be a more recursive way we could extract the max lengths and their hash paths given _any_ schema. I'll have to mess around with it a bit, but won't block this PR from going through for it :)

cilestin (changes requested):

cilestin (commented):
We also need to be sure test the headers' maxLengths, though let's do that in a separate PR/ticket since the SC header schema needs some changes anyway.

modules/appeals_api/spec/support/schema_helpers.rb
```diff
@@ -15,4 +17,58 @@ def read_schema(filename, schema_version = 'v1')
       )
     )
   end
+  # rubocop:disable Layout/LineLength
+
+  def find_max_lengths(filename, _string_to_find, schema_version = 'V2')
+    schema = read_schema(filename, schema_version)
+    max_lengths = {}
+    schema['definitions'].each do |key|
+      if !key[1]['properties'].nil?
```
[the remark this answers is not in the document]
stiehlrod: I am not sure how it got there :/

modules/appeals_api/spec/support/schema_helpers.rb
```diff
@@ -15,4 +17,52 @@ def read_schema(filename, schema_version = 'v1')
       )
     )
   end
+  # rubocop:disable Layout/LineLength
+
+  def schema_max_lengths(_filename, schema, _schema_version = 'V2')
```
ElleeB, line 22: Why is the filename param still expected?

modules/appeals_api/spec/services/appeals_api/pdf_construction/generator_spec.rb
```diff
@@ -192,6 +193,22 @@
           File.delete(generated_pdf) if File.exist?(generated_pdf)
         end
       end
+
+      context 'pdf max length content verification' do
+        let(:schema) { read_schema('../../schemas/v2/200995.json', 'V2') }
+        let(:max_lengths) { schema_max_lengths '../../schemas/v2/200995.json', schema }
```
ElleeB, line 199: Can you just pass the schema?

modules/appeals_api/spec/services/appeals_api/pdf_construction/generator_spec.rb
```diff
@@ -192,6 +193,22 @@
           File.delete(generated_pdf) if File.exist?(generated_pdf)
         end
       end
+
+      context 'pdf max length content verification' do
+        let(:schema) { read_schema('../../schemas/v2/200995.json', 'V2') }
```
ElleeB, line 198: i think you can just pass `200995.json`

ElleeB (dismissed):
A few more tweaks and I think this is good, though I think we should come back to it and consider how we can refactor the helper methods to be more flexible.

modules/appeals_api/spec/support/schema_helpers.rb
```diff
@@ -15,4 +15,12 @@ def read_schema(filename, schema_version = 'v1')
       )
     )
   end
+
+  def self.schema_max_lengths(file, schema, _schema_version = 'V2')
```
cilestin, line 19: 1. Since we're actually sending in an `appeal` and not a file, let's rename `file` to `appeal`.
1. Let's also remove the `_schema_version` argument, since the method doesn't use it and we're relying on `read_schema` outside the method to load the correct schema version.
1. I'll let you choose whether to make this an instance method instead of a static method so you can just call `schema_max_lengths` instead of `SchemaHelpers.schema_max_lengths`, or rename the static method since we already know it's meant to be working on/with a schema by being a part of the `SchemaHelpers` module. Maybe something like `SchemaHelpers.override_max_lengths` since it does some data manipulation.

cilestin (changes requested):
Just a few small tweaks and it's good!

cilestin (dismissed):

ElleeB:
I'm just noticing that the 200996 max length pdf is not longer checking to make sure that `issue descriptions/text` _does not_ overlap with `area of disagreement`. That was the kick off for this whole max length party.

cilestin (approved):

PR #9394, opened 2022-03-18
Add mobile schedule appointment policy

aherzberg:
## Description of change

Adds policy for user authorization to use upcoming mobile feature, schedule appointments. 

## Original issue(s)
https://github.com/department-of-veterans-affairs/va-mobile-app/issues/2726

Policy checks two things
1. If user is loa3. This is somewhat redundant because all mobile users are loa3 but is a requirement for this feature in `vets-website` so if we were ever to have users that weren't loa3, this policy could still be used.
2. If user has registered facilities
- Pre-emptively added metric for policy if we were to ever want those stats.
- Moved policy to mobile module since it will only be used in mobile.


modules/mobile/spec/request/user_request_spec.rb
```diff
@@ -290,6 +293,49 @@
           )
         end
       end
+
+      context 'with a user who does not have access to schedule appointments' do
+        let(:user_request) do
+          iam_sign_in(FactoryBot.build(:iam_user, :no_participant_id))
```
kpethtel, line 299: Can you explain to me why this user does not have access to schedule appointments? The requirements for this policy are that the user have Level of Assurance 3 and have access to VA facilities. I don't really understand how levels of assurance work. Why does `no_participant_id` result in low LOA? 

Also, are low LOA and lack of facilities tied together? In other words, is it possible to meet one condition but not the other? The way we're testing here, we seem to assume that low LOA and lack of facilities go hand-in-hand. Is that the case? Can someone have LOA 3 but lack facilities, or visa versa?

modules/mobile/app/policies/schedule_appointment_policy.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+ScheduleAppointmentPolicy = Struct.new(:user, :schedule_appointment) do
+  def access?
+    accessible = user.loa3? && user.vha_facility_ids.any?
```
kpethtel, line 5: Does `user.vha_facility_ids.any?` differ meaningfully from the `user.va_patient?` method Patrick recommended? That method seems to be a bit more involved but the comments around the method seem to imply that it's essentially the same as what you have here.

kpethtel (commented):
Code looks good, but I would like to understand it better before approving. I'm mostly interested in better understanding LOA and how it's tied to a participant id. Any information you can share would be appreciated.

modules/mobile/app/policies/schedule_appointment_policy.rb
aherzberg: I didn't use `va_patients?` because it gets overridden in `app/models/iam_user.rb` to a method that doesn't actually correctly determine va patient status (according to comment above it.). Per your suggestion over slack, I'll just use the `va_treatment_facility_ids` to ensure non-valid facility ids are still getting filtered out.

modules/mobile/spec/request/user_request_spec.rb
```diff
@@ -290,6 +293,73 @@
...
+              user_request
+            end.to trigger_statsd_increment('mobile.schedule_appointment.policy.failure', times: 1)
+          end
+        end
+
+        context 'due to not being LOA3' do
+          let(:user_request) do
+            iam_sign_in(FactoryBot.build(:iam_user, :loa2))
+            VCR.use_cassette('payment_information/payment_information') do
+              VCR.use_cassette('user/get_facilities_no_ids', match_requests_on: %i[method uri]) do
```
kpethtel, line 324: Doesn't this mean both conditions are being met? If possible, I'd change this to the normal "get facilities" to ensure that it's failing for the right reason.

kpethtel (approved):
Looks good! Just one non-blocking question/comment. 🚀

modules/mobile/spec/request/user_request_spec.rb
aherzberg: `no_participant_id ` caused no facilities to be stored, so I utilized it to test the condition that user must have registered facilities get pass the policy. There may be other reasons no participant id could cause this though so I updated the spec to create a user that explicitly had the facilities taken out. Also created another spec with a loa2 user to test the other condition, though this shouldn't come up in production since all mobile users are loa3.

modules/mobile/spec/request/user_request_spec.rb
aherzberg: Looks like loa2 changes the facility behavior. no facility ids are given in the call when loa2.

aherzberg:
closing pull request to make some changes

kreek (approved):
LGTM 🚀

thilton-oddball (approved):

PR #9399, opened 2022-03-18
API-12495 Remove unused hlr files

ElleeB:
This PR removes a file that hasn't been used since we moved from a specific HLR pdf submit job, to a general pdf submit job that works for each appeal type.

This does not solve any problems related to our LoB testing in this [ticket](https://vajira.max.gov/browse/API-12495), but was discovered during the process.


stiehlrod (approved):
LGTM

stiehlrod (approved):

cilestin (approved):
👏 Thanks!

PR #9403, opened 2022-03-18
RRD: Refactor RapidReadyForDecision::Constants

yoomlam:
## Description of change

Followup to [comment](https://github.com/department-of-veterans-affairs/vets-api/pull/9352#discussion_r830141736) in #9352.
Refactor `RapidReadyForDecision::Constants` by passing around symbols instead of structs. Whenever we need the struct, just look it up in the `RapidReadyForDecision::Constants:: DISABILITIES` hash, which is keyed by a symbol.

## Original issue(s)
Slightly messy code in #9352

Only a refactor to clean up code. No change in behavior.


app/services/rapid_ready_for_decision/processor_selector.rb
```diff
@@ -6,38 +6,40 @@ def initialize(form526_submission)
       @form526_submission = form526_submission
     end
 
+    RrdConstants = RapidReadyForDecision::Constants
```
yoomlam, line 9: Used to shorten line lengths.

app/services/rapid_ready_for_decision/processor_selector.rb
```diff
@@ -6,38 +6,40 @@ def initialize(form526_submission)
...
-      # TODO: use a constant hash instead
-      case disability_type
-      when RapidReadyForDecision::Constants::DISABILITIES[:hypertension]
-        RapidReadyForDecision::DisabilityCompensationJob
-      when RapidReadyForDecision::Constants::DISABILITIES[:asthma]
-        RapidReadyForDecision::Form526AsthmaJob
-      end
+      return unless self.class.disability_increase?(form_disability, disability_struct)
+
+      DISABILITY_TO_PROCESSOR_CLASS[disability_symbol]
```
yoomlam, line 34: Addresses the `TODO`.

yoomlam:
@nanotone Low-priority PR.

app/services/rapid_ready_for_decision/processor_selector.rb
```diff
@@ -6,38 +6,40 @@ def initialize(form526_submission)
...
-      return unless disability_types.size == 1
+      return unless mapped_submission_disabilities.size == 1
 
-      disability_type = disability_types.first
+      disability_symbol = mapped_submission_disabilities.first
       # must be a disability supported by RRD
-      return unless disability_type
+      return unless disability_symbol
 
+      disability_struct = RrdConstants::DISABILITIES[disability_symbol]
```
nanotone, line 26: Given that `disability_symbol` is only used here to get the struct back, and below to get the class (which could also live in the struct) I don't really see the benefit of switching from struct to symbol. Maybe I missed something -- are we more worried about passing structs around, rather than symbols?

app/services/rapid_ready_for_decision/processor_selector.rb
```diff
@@ -6,38 +6,40 @@ def initialize(form526_submission)
       @form526_submission = form526_submission
     end
 
+    RrdConstants = RapidReadyForDecision::Constants
+
+    DISABILITY_TO_PROCESSOR_CLASS = {
+      hypertension: RapidReadyForDecision::DisabilityCompensationJob,
+      asthma: RapidReadyForDecision::Form526AsthmaJob
+    }.freeze
```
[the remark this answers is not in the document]
yoomlam: Since we're referencing an actual class, I prefer to reference the class so that in case we ever change the class name, the error will be obvious. However, I will concede to your suggestion to get this PR merged.

app/services/rapid_ready_for_decision/processor_selector.rb
yoomlam: I am somewhat worried about passing around a struct -- it seems like asking for trouble. Comparing symbols is straight-forward and fast; comparing structs is not.

yoomlam: I'm also planning to pass the disability symbols around to various classes (that are currently hypertension-specific) so that we can make them more general.

spec/services/rapid_ready_for_decision/constants_spec.rb
```diff
@@ -5,27 +5,33 @@
 RSpec.describe RapidReadyForDecision::Constants do
   let(:form526_submission) { create(:form526_submission, :hypertension_claim_for_increase) }
 
-  describe '.extract_disability_constant_list' do
-    subject { described_class.extract_disability_constant_list(form526_submission) }
+  describe 'processor_class' do
+    it 'resolves all processor_class values to classes' do
+      classes = RapidReadyForDecision::Constants::DISABILITIES.values.pluck(:processor_class).map(&:constantize)
+      expect(classes.any?(&:nil?)).to eq false
```
nanotone, line 11: This is so good.

Although I think a non-resolvable name will raise `NameError` on the previous line. Anyway, the test works.

nanotone (approved):
🚀

nanotone (approved):
Approving, since I had already approved this one and the base PR #9352 before. 👍

rileyanderson (approved):

PR #9406, opened 2022-03-18
Api 14219 nod pdf example

stiehlrod:
## Description of change
We should ensure we generate example PDFs that utilize the max character length for every data field so we can ensure there is no overlap.

The example PDFs should be available in the fixtures folder.

## Original issue(s)
[API-14219](https://vajira.max.gov/secure/RapidBoard.jspa?rapidView=1519&projectKey=API&view=detail&selectedIssue=API-14219)

* uses helper method to find and replace max lengths for fields in the NOD form.
* Removes max length test (form fill) in extra json file.


modules/appeals_api/spec/fixtures/v2/valid_10182_extra.json
```diff
@@ -19,7 +19,7 @@
           "phoneNumber": "8001111",
           "phoneNumberExt": "2"
         },
-        "email": "123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789_123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789_12345678901234567890@bobbytablesemail.com",
+        "email": "bob@bobbytablesemail.com",
```
cilestin, line 22: Since you made these changes to the extra json file, you'll almost certainly also need to regenerate the `expected_10182_extra.pdf` to get all the tests to pass.

ElleeB:
1. I'm not seeing any length checks in the max lengths pdf
2. I think we should remove the Ws from the extra pdf fixture and allow it to print what's in the data fixture since we'd expect to see the test for max length isolated from the test for extra data now

cilestin (approved):

PR #9408, opened 2022-03-18
Api 14218 hlr pdf example

stiehlrod:
## Description of change
We should ensure we generate example PDFs that utilize the max character length for every data field so we can ensure there is no overlap.

The example PDFs should be available in the fixtures folder.

## Original issue(s)
[API-14218](https://vajira.max.gov/browse/API-14218)

* related to [API-14219](https://vajira.max.gov/browse/API-14219) & [API-13721](https://vajira.max.gov/browse/API-13721)


modules/appeals_api/spec/fixtures/v2/valid_200996_extra.json
```diff
@@ -1,189 +1,189 @@
...
-          },
-          "phone": {
-            "countryCode": "34",
-            "areaCode": "555",
-            "phoneNumber": "8001111",
-            "phoneNumberExt": "2"
-          },
-          "email": "123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789_123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789_12345678901234567890@bobbytablesemail.com",
-          "timezone": "America/Chicago"
+  "data": {
```
cilestin, line 2: Since you've made changes to this file, you'll almost certainly need to regenerate the `expected_200996_extra.pdf` to get the existing extra test to pass as well :)

modules/appeals_api/spec/fixtures/v2/valid_200996_extra.json
```diff
@@ -1,189 +1,189 @@
...
-        "attributes": {
-          "issue": "cootiescootiescootiescootiescootiescootiescootiescootiescootiescootiescootiescootiescootiescootiescootiescooties",
-          "decisionDate": "1900-01-15",
-          "disagreementArea": "Service connectionService connectionService connectionService connectionService connection"
+      "socOptIn": true,
+      "informalConference": true,
+      "informalConferenceContact": "veteran",
+      "informalConferenceTime": "800-1200 ET",
+      "informalConferenceRep": {
+        "firstName": "Helen",
```
cilestin, line 50: The representative name looks pretty bad on the maxlength pdf. Let's put in a ticket to get that cleaned up.

cilestin:
A reminder that we want to load up the header schemas and adjust the veteran's name & other header data to test those maxwidths as well.

PR #9409, opened 2022-03-18
2808 mobile: handle nil providers in va video connect appointments

kreek:
## Description of change
When testing with a Cerner patient QA found a bug where the provider may be missing for VA video connect appointments. PR catches that case and returns nil for the appointment's location.

## Original issue(s)
department-of-veterans-affairs/va-mobile-app#2808


jperk51 (approved):
LGTM

PR #9417, opened 2022-03-21
RRD: Improve asthma offramp notification

nanotone:
## Description of change
This PR incrementally improves on the RRD Asthma offramp notification introduced in #9344, by
- adding the submission ID, and
- fixing the logic for counting number of medication requests.

* The original medication request counting code was buggy because I mocked out the data incorrectly. Switching from mocking to VCR helped surface the fix.


spec/jobs/rapid_ready_for_decision/form526_asthma_job_spec.rb
```diff
@@ -10,20 +10,30 @@
...
+    EVSS::DisabilityCompensationAuthHeaders.new(user).add_headers(EVSS::AuthHeaders.new(user).to_h)
+  end
+  let(:saved_claim) { FactoryBot.create(:va526ez) }
+  let(:submission) do
+    create(:form526_submission, :with_uploads,
+           user_uuid: user.uuid,
+           auth_headers_json: auth_headers.to_json,
+           saved_claim_id: saved_claim.id,
+           submitted_claim_id: '600130094')
+  end
```
nanotone, line 24: All this extra setup is to mimic the setup in Form526HypertensionJobSpec which allows the VCR cassette to just work.

I do think we're probably due for a VCR cassette refactor sometime in the near future.

yoomlam: Hmmm, interesting. Are you saying the FactoryBot for `form526_submission` isn't sufficient for enabling VCR? Why not?

yoomlam (approved):
Dropped a question for more clarity.

spec/jobs/rapid_ready_for_decision/form526_asthma_job_spec.rb
nanotone: Okay, it looks like I can get rid of the saved claim, but gotta keep the user (for their ICN). Pushing an update now.

yoomlam (approved):

PR #9419, opened 2022-03-21
Additional login type metrics mobile

jperk51:
## Description of change
We aren't seeing any login.gov authentications since it was rolled out at the end of last week. With the current metrics it's hard to determine if our metric capturing is the issue, if login.gov is never succeeding, or if we have no login.gov attempts. Adding this line to track all login types that make it through introspect. Logging only change.


jperk51:
We've confirmed the ability to login with login.gov users. It seems like we aren't getting any "login.gov" login methods back from iam introspect though. This log line should help us confirm that.

aherzberg (approved):
LGTM

rileyanderson (approved):

PR #9420, opened 2022-03-21
[38760] removes orch_search_service create_profile_message method

bramleyjl:
## Description of change

This change removes the `OrchSearchService` subclass of the MPI service class, folding its functionality back into the parent class. The OrchSearch child class was discovered to only be attempting to query MPI through the user's EDIPI; by integrating it into the normal MPI service `find_profile` method it will now also attempt to use a user's ICN or user attributes.

In addition, this PR also removes the `mvi.edipi_search` Setting that selectively prevented MPI queries by EDIPI. This is an outdated setting that is no longer needed.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#38760

orch_search specs have been updated, as well as find_profile specs that have switch from using user attributes to EDIPIS.


lib/mpi/service.rb
```diff
@@ -59,8 +59,10 @@ def add_person(user_identity)
     # @param user [UserIdentity] the user to query MVI for
     # @return [MPI::Responses::FindProfileResponse] the parsed response from MVI.
     # rubocop:disable Metrics/MethodLength
-    def find_profile(user_identity, search_type = MPI::Constants::CORRELATION_WITH_RELATIONSHIP_DATA)
-      profile_message = create_profile_message(user_identity, search_type: search_type)
+    def find_profile(user_identity,
+                     search_type = MPI::Constants::CORRELATION_WITH_RELATIONSHIP_DATA,
```
bosawt, line 63: I wonder if it'd be better to make all of these named params, so we're not mixing both types?
`def find_profile(user_identity: user_identity, search_type: MPI::Constants..., orch_search: false)`

lib/mpi/service.rb
```diff
@@ -165,12 +167,14 @@ def create_add_message(user_identity)
...
-    def create_profile_message(user_identity, search_type: MPI::Constants::CORRELATION_WITH_RELATIONSHIP_DATA)
+    def create_profile_message(user_identity,
+                               search_type: MPI::Constants::CORRELATION_WITH_RELATIONSHIP_DATA,
+                               orch_search: false)
       return message_icn(user_identity, search_type) if user_identity.mhv_icn.present?
       return message_edipi(user_identity, search_type) if user_identity.edipi.present? && Settings.mvi.edipi_search
       raise Common::Exceptions::ValidationErrors, user_identity unless user_identity.valid?
 
-      message_user_attributes(user_identity, search_type)
+      message_user_attributes(user_identity, search_type, orch_search: orch_search)
```
bosawt, line 177: I don't think this code will be reachable with orch_search, it'll fall into the `message_edipi` bucket instead, so I don't think `orch_search` will actually be called as expected

lib/mpi/service.rb
```diff
@@ -165,12 +167,14 @@ def create_add_message(user_identity)
       MPI::Messages::AddPersonMessage.new(user_identity).to_xml if user_identity.icn_with_aaid.present?
     end
 
-    def create_profile_message(user_identity, search_type: MPI::Constants::CORRELATION_WITH_RELATIONSHIP_DATA)
+    def create_profile_message(user_identity,
+                               search_type: MPI::Constants::CORRELATION_WITH_RELATIONSHIP_DATA,
+                               orch_search: false)
       return message_icn(user_identity, search_type) if user_identity.mhv_icn.present?
       return message_edipi(user_identity, search_type) if user_identity.edipi.present? && Settings.mvi.edipi_search
```
bosawt, line 174: Just as an aside, I think we can remove this `Settings.mvi.edipi_search` flag, don't really see a reason to have it toggleable..

bosawt (changes requested):
I think with the current implementation it won't work as expected. The problem is that `Settings.mvi.edipi_search` is set to `false` on localhost, and `true` everywhere else. We should prob just remove that altogether. But what'll happen on prod stacks is these add_person calls will fall into `message_edipi`, and then the `orch_search` parameter won't get properly passed down...

I think what I'd do is pass `orch_search` down through all three, `message_icn`, `message_edipi`, and `message_user_attributes`, and otherwise keep these functions the same as they were.

spec/lib/mpi/service_spec.rb
```diff
@@ -183,6 +183,52 @@
...
+            edipi: '1013590059'
+          }
+        end
+
+        it 'calls the find profile with an orchestrated search', run_at: 'Thu, 06 Feb 2020 23:59:36 GMT' do
+          allow(SecureRandom).to receive(:uuid).and_return('b4d9a901-8f2f-46c0-802f-3eeb99c51dfb')
+          allow(Socket).to receive(:ip_address_list).and_return([Addrinfo.ip('1.1.1.1')])
+
+          VCR.use_cassette('mpi/find_candidate/orch_search_with_attributes', VCR::MATCH_EVERYTHING) do
+            Settings.mvi.vba_orchestration = true
```
bosawt, line 207: Could you add a mock for this instead of actually changing the Settings values: `allow(Settings.mvi).to receive(:vba_orchestration).and_return(true)` ? 

I'm trying to remove singleton editing from specs in general, since it leads to unexpected behavior (for example, in this case it appears that you cleaned up properly, however, if we had actually expected `vba_orchestration` to be set to `true` for everything, then setting it to `false` at the end here would unexpectedly change downstream specs)

bosawt (changes requested):
Looking good! I only have one spec comment but otherwise looks good

bosawt (approved):
Looks good! Thanks for making the spec change

thilton-oddball (approved):

PR #9424, opened 2022-03-21
API-13516 wire up non-veteran claimant for nod v2

ElleeB:
While reviewing another [PR](https://github.com/department-of-veterans-affairs/vets-api/pull/9412), that adds claimant header documentation to our lower envs, I realized we'd never actually wired up Non-Veteran claimant (NVC) (via adding it to the schema and extra fixtures).

This PR ensures that NOD v2 handles and prints NVC data properly.

---

All three pdf fixtures required updating, which was expected for the standard and extra pdf. But, the expected minimum failed surprisingly, and needed updating, but as far as I could tell, the original expected and the newly generated were identical. This has happened a couple times before. Check me on this!


modules/appeals_api/spec/models/notice_of_disagreement_spec.rb
```diff
@@ -252,45 +252,45 @@
   end
 
   describe 'V2 methods' do
-    let(:notice_of_disagreement_v2) { create(:extra_notice_of_disagreement_v2, :board_review_hearing) }
+    let(:extra_notice_of_disagreement_v2) { create(:extra_notice_of_disagreement_v2, :board_review_hearing) }
```
ElleeB, line 255: Just to be more clear and differentiate from the other nods.

cilestin (approved):

PR #9426, opened 2022-03-21
Add missing additional properties key

ElleeB:
While tying up loose ends relating to NOD v2, I discovered that the safety, `"additional_properties": false`, had been removed from the schema at some point.

This keyword set to false _will not_ allow any other properties to be included in incoming payloads.

As our schema stands now, I could add anything and it will come through. If we aren't concerned, then we can delete this PR and remove the key/value pair from HLR schema since it's always been there.


cilestin:
As mentioned elsewhere, the Proposed API Styleguide's [payload contract ]( https://legendary-system-8b212937.pages.github.io/unified/payloads/#api-contract-description)does not want us to use `additionalProperties` in our schemas.

Closing.

PR #9428, opened 2022-03-22
API-14351: Update veteran identifiers validations, claims API

hinzed1127:
## Description of change

In diagnosing a bug, we realized that in the case of missing BIRLS/participant IDs, we were trying to add `@current_user` to MPI. Since this is not always the veteran, this changes that call to be made on the target veteran.

## Original issue(s)
https://vajira.max.gov/browse/API-14351


acovrig (approved):

jw81 (approved):

hinzed1127:
@acovrig the error would currently still stay the same, but just now for the correct person: it'd be a `422` saying they're missing EDIPI, although VSP is working to remove that constraint, so then it could be a `422` for other bad identifiers.

PR #9438, opened 2022-03-22
2826: do not include directDepositBenefitsUpdate in authorized servic…

kpethtel:
## Description of change
When EVSS error is raised due to unknown errors from the EVSS service, catch the error and exclude directDepositBenefitsUpdate from the user's authorized services.

## Original issue(s)
department-of-veterans-affairs/va-mobile-app#2826

This is a temporary patch for errors we're seeing from the EVSS service on staging. We don't have enough information about the errors to fix this in a better way, so we will simply catch them in the serializer (the only relevant code that's fully under our control) and it's blocking staging. We will create a secondary ticket to investigate further and develop a more durable fix.


aherzberg (dismissed):
LGTM

modules/mobile/app/serializers/mobile/v0/user_serializer.rb
```diff
@@ -94,12 +94,17 @@ def profile
...
-        end
+        auth_services.push(:directDepositBenefitsUpdate) if can_access_direct_deposit_benefits?(auth_services)
         auth_services
       end
 
+      def can_access_direct_deposit_benefits?(auth_services)
+        auth_services.include?(:directDepositBenefits) && user.authorize(:ppiu, :access_update?)
+      rescue EVSS::ErrorMiddleware::EVSSError
+        # this is a temporary patch for upstream issues
+        false
```
jperk51, line 105: I would add some logging here or a metric so we can see how often this is happening more easily

kpethtel: I thought about it but I assumed it would just output: `[{"key"=>"paymenthistory.partner.service.failed", "severity"=>"ERROR", "text"=>"Call to partner getPaymentHistory failed"}]`
I'm not certain though, so I'll add something and we'll see what happens.

jperk51:
We should wait to merge this until we get the go ahead from @travis-newby , @lexicalninja , or @kreek

PR #9442, opened 2022-03-23
updated logingov introspect and added mobile mapping

jperk51:
## Description of change
IAM introspect response has changed for login.gov since we integrated and tested with it. Updating the cassette to match and added mapping in our mobile session code to map new values to fediamassur_level so our existing logic still works. 

## Original issue(s)
department-of-veterans-affairs/va-mobile-app#2830

Tests were updated via the cassette change.


lib/iam_ssoe_oauth/session_manager.rb
```diff
@@ -62,6 +63,13 @@ def create_user_session
       raise e
     end
 
+    def add_missing_fediamassur_level(iam_profile)
+      if iam_profile[:fediamassur_level].blank?
+        iam_profile[:fediamassur_level] = iam_profile[:fediam_aal] == '2' && iam_profile[:fediam_ial] == '2' ? 3 : 0
```
kreek, line 68: I had to read it a couple of times before I realized the first part was a tenary argument:
```suggestion
        iam_profile[:fediamassur_level] = (iam_profile[:fediam_aal] == '2' && iam_profile[:fediam_ial] == '2') ? 3 : 0
```

kreek (approved):
One comment but LGTM 🚀

lib/iam_ssoe_oauth/session_manager.rb
rileyanderson, line 68: I agree that this is hard to read. Could you do something like this?
```suggestion
        iam_profile[:fediamassur_level] = if iam_profile[:fediam_aal] == '2' && iam_profile[:fediam_ial] == '2'
                                            3
                                          else
                                            0
                                          end
```

jperk51: done

rileyanderson (approved):

PR #9444, opened 2022-03-23
API-13065: Claims v2 error handling

acovrig:
## Description of change

Catch auth errors so we can force custom values as needed.

## Original issue(s)
[API-13065](https://vajira.max.gov/browse/API-13065)

Currently only catches `TokenValidationError`, but will easily allow us to catch more later.


modules/claims_api/lib/claims_api/error/error_handler.rb
```diff
@@ -0,0 +1,29 @@
...
+      def self.included(clazz)
+        clazz.class_eval do
+          rescue_from ::Common::Exceptions::TokenValidationError, with: :token_validation_error
+        end
+      end
+
+      private
+
+      def token_validation_error
+        json = {
```
jw81, line 15: Rather than manually building json structures here, I think we should be building custom exception classes.  Something like `ClaimsApi::Error::TokenValidationError`.
Similar to what is done in this [dir](https://github.com/department-of-veterans-affairs/vets-api/tree/master/lib/common/exceptions) in `vets-api`.

I imagine this handler module is going to grow a fair amount over time and we don't want it to be littered with tons of manual JSON structure building.

modules/claims_api/lib/claims_api/error/error_handler.rb
```diff
@@ -0,0 +1,29 @@
...
+          errors: [
+            {
+              title: 'Token Validation Error',
+              detail: 'Invalid token.',
+              code: '401',
```
jw81, line 20: I think for now at least, we should not be returning a `code` attribute.
I believe this is sort of like the "sub-code" stuff that Kayla mentioned.  
But we don't yet have a system of defined sub-codes or how to correlate them to meaningful details.
We also haven't observed a clear need for a sub-coding system yet.
So for now, I say we just don't provide a `code` attribute.
We can always add things to the structure later, but we can never remove something once it exists.  So let's not add anything that we aren't 100% certain is needed.

acovrig: I agree that we don't need a `code` field (or `status` for that matter), but that's what we're returning right now, can we remove it, or is a client using it?

modules/claims_api/spec/lib/claims_api/error/error_handler.rb
```diff
@@ -0,0 +1,71 @@
...
+      allow(JWT).to receive(:decode).and_raise(JWT::DecodeError)
+
+      get get_poa_path, headers: auth_header
+
+      parsed_body = JSON.parse(response.body)
+      expect(parsed_body['errors'][0]['detail']).to eq('Invalid token.')
+    end
+  end
+
+  it 'catches an expired token' do
```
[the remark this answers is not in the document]
acovrig: Dan and I tried that, but we could only get it to raise a `JWT::DecodeError` for some reason.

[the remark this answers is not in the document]
hinzed1127: The issue that we were running into was there are [three](https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/controllers/openid_application_controller.rb#L326) [different](https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/models/token.rb#L51) [times](https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/models/token.rb#L30) `JWT.decode` is called before [the code we want it to hit](https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/models/token.rb#L20) (i.e., where it catches and raises the `JWT::ExpiredSignature` error) and fail happens. It's a pretty brittle way of testing, so I'm totally in favor of a better approach that lets us still test a similar expectation. 

Austin had initially tried a different approach that used the `controller` keyword [within rspec](https://stackoverflow.com/questions/46986332/what-is-the-best-way-to-test-application-controller-rescue-status-codes-in-rails/47006072#47006072), but he was having different issues getting that to work. I went with the pattern I was more familiar with and then came up against the issue we solved for here. I'm all for trying to get the other technique working if it's not a time suck (seems much cleaner), but at the very least, adding some explanatory logic in a comment if this stays in.

modules/claims_api/lib/claims_api/error/error_handler.rb
jw81: So this stuff you're building is only going to be used in V2 (and versions beyond that) and there is no one currently consuming V2.  So we're free to build whatev we want here without fear of impacting current consumers.

modules/claims_api/spec/lib/claims_api/error/error_handler.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
```
jw81, line 1: Just noticed this, but the file name needs to end with `_spec.rb`.
So `error_handler_spec.rb`.

modules/claims_api/spec/lib/claims_api/error/error_handler.rb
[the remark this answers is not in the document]
jw81: Maybe I'm missing something, but this code works for me locally.

```
  it 'catches an expired token' do
    with_okta_user(scopes) do |auth_header|
      allow(JWT).to receive(:decode).and_raise(JWT::ExpiredSignature)

      get get_poa_path, headers: auth_header

      parsed_body = JSON.parse(response.body)
      expect(parsed_body['errors'][0]['detail']).to eq('Invalid token.')
    end
  end
```

modules/claims_api/lib/claims_api/error/error_handler.rb
acovrig: yea, depending on the library (assuming JS) theoretically `response.body.errors[0].code == response.body.errors[0].status == response.status_code` I feel like `code` should be reserved for a `subcodes` feature down the road Kayla mentioned.

modules/claims_api/lib/claims_api/error/error_handler.rb
acovrig: I moved it to it's own class. What do you think of the lambda bit, OK, or too much?

Thought: I don't really want to make a method for each type of error twice, so the lambda would take care of that.

modules/claims_api/spec/lib/claims_api/error/error_handler.rb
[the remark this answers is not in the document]
acovrig: Yes, that test should succeed with the error handling in place, but if you take out the error handling it still succeeds because the `ExpiredSignature` exception is getting converted to an `InvalidToken` somewhere. The test needs to throw an `ExpiredSignature` and needs to render the same as `InvalidToken`.

jw81 (changes requested):
We talked in Slack yesterday, but basically, we need to get this PR into a green and stable state.  Then I can look again and see if anything else needs to be addressed.

modules/claims_api/lib/claims_api/error/error_handler.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+require 'claims_api/common/exceptions/token_validation_error'
+
+module ClaimsApi
+  module Error
+    module ErrorHandler
+      def self.included(clazz)
+        clazz.class_eval do
+          # rescue_from ::Common::Exceptions::TokenValidationError, with: :token_validation_error
```
jw81, line 10: Let's drop this comment if we're gonna go with the `lambda` implementation.

modules/claims_api/lib/claims_api/error/error_handler.rb
```diff
@@ -0,0 +1,30 @@
...
+      private
+
+      def render_error(error)
+        render json: { errors: error.errors }, status: error.status_code
+      end
+
+      # def token_validation_error
+      #   err = ClaimsApi::Error::TokenValidationError.new
+      #   render_error(err)
+      # end
```
jw81, line 27: Let's drop this if we're gonna go with the `lambda` implementation.

spec/support/schemas/claims_api/v2/errors/default.json
```diff
@@ -0,0 +1,41 @@
...
+            "description": "HTTP error detail"
+          },
+          "code": {
+            "type": "string",
+            "description": "HTTP error code"
+          },
+          "status": {
+            "type": "string",
+            "description": "HTTP error code"
+          },
```
jw81, line 25: Let's remove these (`code` and `status`) from the schema if we're not planning to use them for now.

modules/claims_api/spec/lib/claims_api/error/error_handler_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+        i += 1
+        okta_jwt(scopes)
+      end
+
+      get get_poa_path, headers: auth_header
+
+      parsed_body = JSON.parse(response.body)
+      expect(parsed_body['errors'][0]['detail']).to eq('Invalid token.')
+    end
+  end
```
jw81, line 72: Let's remove the commented code from within this test.

I don't really love this test because it seems unnecessarily complex.
But we can revisit it once we get to focus on V2 again.

spec/support/schemas/claims_api/v2/errors/default.json
acovrig: Should I remove "code" and "status" from all other responses in v2?

jw81:
Since we're changing the error response structure, we'll need to re-run RSWAG so we can generate a new OpenApi Spec.

spec/support/schemas/claims_api/v2/errors/default.json
hinzed1127: @jw81 I think leaving them in is fine, for now. Since we're only intercepting the token validation error to start with, other errors should still have those properties, right? So having them here but not in `required` seems correct to me.

modules/claims_api/spec/lib/claims_api/error/error_handler_spec.rb
hinzed1127: [I was able to simplify this a lot](https://github.com/department-of-veterans-affairs/vets-api/pull/9444/commits/5dfb11e61f79393ab91ae30fe78a827e995b09fe) by just stubbing those 3 calls explicitly. It now fails where we want it to.

modules/claims_api/spec/lib/claims_api/error/error_handler_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+#       p request.headers
+#       # byebug
+#       get :test_invalid_token
+#       puts response.body
+#       expect(response).to have_http_status(401)
+#       parsed_body = JSON.parse(response.body)
+#       expect(parsed_body['detail']).to eq('Invalid token.')
+#     end
+#   end
+# end
```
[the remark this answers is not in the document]
hinzed1127: @acovrig I got that test working [here](https://github.com/department-of-veterans-affairs/vets-api/pull/9444/commits/5dfb11e61f79393ab91ae30fe78a827e995b09fe). If you like that implementation, just remove the other test checking the same thing in a different way and keep this one.

modules/claims_api/spec/lib/claims_api/error/error_handler_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+      get :test_invalid_token
+
+      expect(response).to have_http_status(:unauthorized)
+      parsed_body = JSON.parse(response.body)
+      expect(parsed_body['errors'].size).to eq(1)
+      expect(parsed_body['errors'][0]['title']).to eq('Token Validation Error')
+      expect(parsed_body['errors'][0]['detail']).to eq('Token Validation Error')
+    end
+  end
+end
```
hinzed1127, line 42: This works as intended now. I was unable to get a similar test working, because creating a `test_expired_token` route without skipping `authenticate` hits that named method [here](https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/controllers/concerns/authentication_and_sso_concerns.rb#L16) instead of [here](https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/controllers/openid_application_controller.rb#L22). If you can figure out how to get it to call the right inherited method, my updated approach below should work for that, too. Otherwise, it's solved below (see other comment)

modules/claims_api/spec/lib/claims_api/error/error_handler_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+  # it 'catches an invalid token' do
+  #   with_okta_user(scopes) do |auth_header|
+  #     allow(JWT).to receive(:decode).and_raise(JWT::DecodeError)
+
+  #     get get_poa_path, headers: auth_header
+
+  #     parsed_body = JSON.parse(response.body)
+  #     expect(parsed_body['errors'][0]['detail']).to eq('Invalid token.')
+  #   end
+  # end
```
hinzed1127, line 58: if you're satisfied with the working test above, delete this one.

modules/claims_api/spec/lib/claims_api/error/error_handler_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  it 'catches an expired token' do
+    with_okta_user(scopes) do |auth_header|
+      allow_any_instance_of(OpenidApplicationController).to receive(:jwt?).and_return(true)
+      allow_any_instance_of(Token).to receive(:valid_issuer?).and_return(true)
+      allow_any_instance_of(Token).to receive(:public_key).and_return(nil)
+      allow(JWT).to receive(:decode).and_raise(JWT::ExpiredSignature)
```
hinzed1127, line 65: @jw81 this simplifies the mocking logic while also getting the expired signature error to get rescued properly.

I think the reason that using just this last line doesn't work is because [`ExpiredSignature` is a child class of `DecodeError`](https://github.com/jwt/ruby-jwt/blob/a79bb0490f66985bcd3a72ba88d502f6615bf3c3/lib/jwt/error.rb#L9), so it gets rescued early in [`jwt?`](https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/controllers/openid_application_controller.rb#L325-L330)

hinzed1127 (changes requested):
Just need to remove commented code and re-run `rake rswag:claims_api:run` and then this should be good!

modules/claims_api/lib/claims_api/error/error_handler.rb
```diff
@@ -0,0 +1,33 @@
...
+                      with: lambda {
+                        render_error(ClaimsApi::Error::TokenValidationError.new)
+                      }
+          rescue_from ::Common::Exceptions::ResourceNotFound,
+                      ::Common::Exceptions::Forbidden,
+                      ::Common::Exceptions::Unauthorized,
+                      ::Common::Exceptions::ValidationErrorsBadRequest,
+                      ::Common::Exceptions::UnprocessableEntity do |err|
+                        render_error(err)
+                      end
```
jw81, line 20: This ticket has lingered on for too long, but I think we should come back at some point and follow the pattern shown on lines10-13.
Create custom exception classes for each of these and follow that pattern.
We're going to have to do this if we decide someday that we need to introduce some kind of error sub-code mechanism.

hinzed1127 (approved):

PR #2326, opened 2018-10-03
Form 526 reporting

kreek:
## Description of change
Adds a `with_tracking` block that all 526 jobs can use to get reporting and logging around their perform logic. Job results are upserted to a `disability_compensation_job_statuses ` table and linked to a `disability_compensation_submission` id. A bit redundant for the `Submit526Form` job but it allows the results to be stored in one spot. 

@abhijitvk haven't added the tracking to the 4142 job yet because _I think_ `handle_service_exception` needs to always raise an error if it's going to be within the `with_tracking` block and not misreport things.

## Testing done
Unit tests

## Testing planned
Staging integration tests

#### Unique to this PR
- [ ] `disability_compensation_job_statuses` migrations
- [ ] `JobStatus` concern that automates tracking, logging, and metrics
- [ ]  refactors job perform args to pass in `saved_claim_id` and `submission_id` to each job. 

#### Applies to all PRs

- [ ] Appropriate logging
- [ ] Swagger docs have been updated, if applicable
- [ ] Provide link to originating GitHub issue, or connected to it via ZenHub
- [ ] Does not contain any sensitive information (i.e. PII/credentials/internal URLs/etc., in logging, hardcoded, or in specs)
- [ ] Provide which alerts would indicate a problem with this functionality (if applicable)

PR #9450, opened 2022-03-23
Fix with_tracking to mark jobs as success when the block returns

nanotone:
## Description of change
This PR fixes a perceived defect in the way `JobTracker#with_tracking` behaves when its provided Ruby block returns or breaks from the block/method. Returning and breaking are non-exceptional ways to exit a block of code, but they were bypassing the after-yield logic which marks the job with status `success`.

This is a fairly subtle behavior of Ruby's yield-to-block, and quite surprising to developers expecting `with_tracking` to run clean-up logic like a "context manager" (purposefully borrowing language from Python, which does this the unsurprising way).

## Original issue(s)
- I discovered this quirk in a [recent PR](https://github.com/department-of-veterans-affairs/vets-api/pull/9344/files#r826368786)
- This quirk has existed since the code was [added 4 years ago](https://github.com/department-of-veterans-affairs/vets-api/pull/2326/files#diff-cb11fc3e41f1e7464b1f989941770884c4ea3b4f08d2c066e47e68dead3d84b4R13-R15)

## Testing
- I added 3 new RSpec scenarios to explicitly check the resulting status, when the block ends early via 1) `return` 2) `break` and 3) `raise`
- I reverted my [original change](https://github.com/department-of-veterans-affairs/vets-api/pull/9344/files#diff-379bfcdf70dbe6668e1063e38c811c0c842c3f8aec7ff87d23a54b19734970b7R34) to skirting the issue, which would have caused a corresponding test failure without the fix in this PR


lib/sidekiq/form526_job_status_tracker/job_tracker.rb
```diff
@@ -73,8 +73,14 @@ def with_tracking(job_title, saved_claim_id, submission_id, is_bdd = nil)
...
-        yield
-        job_success
+        begin
+          yield
+        rescue
+          exception_raised = true
+          raise
+        ensure
+          job_success unless exception_raised
+        end
```
nanotone, line 83: I haven't come across a better way to run some after-yield code when `return` or `break` is encountered, but _not_ when an exception is raised.

yoomlam, line 83: Did you try:
```suggestion
        begin
          yield
        rescue
          raise
        else
          job_success
        end
```
https://stackoverflow.com/questions/6279956/ruby-exceptions-why-else

nanotone: Yeah, I did -- returns and breaks also skip the `else` section, which doesn't have the no-matter-what urgency of `ensure`.

```ruby
def wrapper
  begin
    puts "before yield"
    yield
  rescue
    raise
  else
    puts "after yield"
  end
end
wrapper { break }
```
This just prints "before yield".

spec/lib/sidekiq/form526_job_status_tracker/job_tracker_spec.rb
```diff
@@ -3,7 +3,12 @@
 require 'rails_helper'
 
 describe Sidekiq::Form526JobStatusTracker::JobTracker do
-  let(:dummy_class) { Class.new { include Sidekiq::Form526JobStatusTracker::JobTracker } }
+  let(:dummy_class) do
```
yoomlam, line 6: More descriptive variable name:
```suggestion
  let(:worker_class) do
```

yoomlam (approved):
Left some non-blocking suggestions.

yoomlam (approved):

rileyanderson (approved):

PR #9451, opened 2022-03-23
API-12495 Fix HLR & SC businessLine mapping

cilestin:
## Description of change
Fixes the mapped `businessLine` we send to centralmail to meet their expectations (No OTH, only CMP)

## Original issue(s)
https://vajira.max.gov/browse/API-12495


stiehlrod (approved):
LGTM

PR #9453, opened 2022-03-24
[35662] inherited_proofing controller & authentication request to Login.gov

bramleyjl:
## Description of change

This PR is the first piece of an update to enable an inherited proofing account creation workflow to transfer MHV verified user accounts to Login.gov IAL2 accounts without forcing users to re-verify their identity.

The logic for this flow is being built into a new `InheritedProofingController` that calls upon a new `inherited_proofing` library to interact with the CSPs. Within that library there are `MHV` and `Logingov` service classes to initiate the calls and handle their responses.

#### Contained within this PR
1. The end user, through `vets-website`, initiates the inherited proofing process by calling `api.va.gov/inherited_proofing/auth`
2. The `InheritedProofingController` first verifies that the user is eligible by using the `MHV::InheritedProofingVerifier` (which calls the MHV service) to query the MHV API's account info endpoint, using the user's ICN as their identifier. MHV will return a hash of user info.
3. If MHV includes the `identityDocumentExist` field in the data hash in their response, the `vets-api` InheritedProofingVerifier service object will generate an `auth_code`, then save it, the entire MHV response object, and the user's uuid in a Redis cache, keyed on the auth code. It then returns the auth code to the controller. If not, no `auth_code` will be created and the workflow will end early.
4. The controller passes the `auth_code` to the Login.gov service, which makes an authentication call to Login.gov, including the `auth_code` in the `inherited_proofing_auth` URL param.

#### Contained in [following PRs](https://app.zenhub.com/workspaces/vsp-identity-5f5bab705a94c9001ba33734/issues/department-of-veterans-affairs/va.gov-team/39040)
5. As part of their account creation process Login.gov will make an API call to `api.va.gov/inherited_proofing/user_attributes`.
6. The `InheritedProofingController` will validate the request's authenticity by first decrypting the passed `inherited_proofing_auth` param using Login.gov's public key, then using the decrypted code to find the saved Redis cache containing the user's uuid. It will then use the User uuid to obtain the necessary user attributes for IAL2 proofing and store them in a JWT signed with vets-api's private key before returning them to Login.gov.
7. Login.gov will use vets-api's public key to decode the JWT, then complete the account creation process with the contained user attributes and forward the user back to `api.va.gov/inherited_proofing/login`. At this point, the user will be logged out and have their session cleared (since they are currently authenticated through MHV), then placed into the regular Login.gov authentication flow. Since they will still possess an active session for their new Login.gov account, they will immediately be authenticated and returned back to the `v1/sessions/saml_callback` as an authenticated Login.gov IAL2 user.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#35662

To test this PR:
* ensure you have the [vets-api-mockdata](https://github.com/department-of-veterans-affairs/vets-api-mockdata) enabled and updated to mock the MHV responses
* authenticate to vets-api with any user
* navigate to `localhost:3000/inherited_proofing/auth with the dev tools network tab open
* you will see the `inherited_proofing_auth` appended as a URL parameter to the vets-api call to `idp.int.identitysandbox.gov`

image


app/controllers/inherited_proofing_controller.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+require 'inherited_proofing/mhv/inherited_proofing_verifier'
+require 'inherited_proofing/logingov/service'
+
+class InheritedProofingController < ApplicationController
+  skip_before_action :verify_authenticity_token, :authenticate
+
+  def auth
+    load_user
```
bosawt, line 10: Since we shouldn't skip `authenticate` for this route, we don't need to manually `load_user`, and can assume `@current_user` is defined

thilton-oddball (dismissed):

app/controllers/inherited_proofing_controller.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+require 'inherited_proofing/mhv/inherited_proofing_verifier'
+require 'inherited_proofing/logingov/service'
+
+class InheritedProofingController < ApplicationController
+  skip_before_action :verify_authenticity_token
```
bosawt, line 7: Oh yeah, we can remove this line. This is a CSRF check that the frontend will send back since it's an authenticated route

config/settings.yml
```diff
@@ -14,6 +14,12 @@ logingov:
   client_cert_path: spec/fixtures/sign_in/oauth.crt
   client_key_path: spec/fixtures/sign_in/oauth.key
   oauth_url: https://idp.int.identitysandbox.gov
+  inherited_proofing:
```
bosawt, line 17: I still would make these completely separate, not in the login_gov namespace. So I guess just at the `inherited_proofing` level. The alternative could be to push the sign_in service logingov stuff into a namespace to look like:

logingov:
  sign_in:
    oauth_url....
    ...
  inherited_proofing:
    oauth_url....
    ...

I guess if we wanted to do that we could keep this as is and add a story to do that for the sign_in service settings

bosawt:
Confirmed this works when authenticating with a user and then navigating to `localhost:3000/inherited_proofing/auth`, it redirects to a logingov auth request with a code that matches the proper redis entry

bosawt (approved):

bosawt:
@thilton-oddball When you get a change to re-review, this should be good to go now

thilton-oddball (approved):

PR #9423, opened 2022-03-21
1095b

stephte:
## Description of change

This PR sets up the API for allowing Veterans to download their 1095B tax form from the VA. It includes a sidekiq script that pulls the tax form information from the AWS S3 bucket that it will be put in (by the Enrollment team); a Form1095B model/database table for storing the information; and a controller for downloading the PDF (plus a 1095 policy to determine form access).

## Original issue(s)
department-of-veterans-affairs/va.gov-team#35824

## Things to know about this PR
The form that the Veterans will be able to download will have PII on it, such as: last 4 of ssn; address; ICN; DOB.

There are rspec tests written, and we have tested all of the features locally, ensuring everything is working properly. We also plan on testing in staging and likely writing out some more rspec tests

PR #9457, opened 2022-03-24
created form 1095B table

stephte:
## Description of change
This PR is for the new database table that will be used for the 1095B PDF downloading.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#35824

other linked PRs:
https://github.com/department-of-veterans-affairs/vets-api/pull/9423
https://github.com/department-of-veterans-affairs/vets-api/pull/9430

Some PII stuff: this table stores user's name, address, last 4 of social, and ICN among other things


nprimak (approved):

db/migrate/20220217005515_create_form1095_bs.rb
```diff
@@ -0,0 +1,28 @@
...
+      t.string :last_name, null: false
+      t.date :birth_date
+      t.string :ssn
+      t.string :address, null: false
+      t.string :city
+      t.string :state
+      t.string :country
+      t.string :zip_code
+      t.string :province
+      t.string :foreign_zip
```
rileyanderson, line 17: Please try to avoid storing PII in the database. There are a few options:
- Store in Redis. Store the data in a single value field and parse it upon retrieval. (this would be faster too)
- If you must do this in the DB, you MUST cyphertext all PII.
- Reduce PII to only what you actually need in the DB (encrypted) and leave the rest in redis. We already store most of this information in a redis object.

rileyanderson (changes requested):

rileyanderson:
What's the estimated number of records you are wanting to store? Is it every veteran for every tax year?

rileyanderson:
> Yes, every veteran for every tax year (starting for 2021). We could get rid of the ICN, SSN, and maybe DOB if we had some kind of has_many relationship where it could access SSN and DOB data fields for that user, but I don't think we can have those relationships on the User redis model, can we? We use ICN to connect the 1095B data from the enrollment system to the User

I don't think we are comfortable storing any of this info in the DB. It should be avoided if possible. How are the files stored in S3? Is it one big file? Are you able to query the S3 file and generate the form on the fly based on the current_user?

stephte:
We need to have all the address data stored as well, since their address could change, and we don't want their address for 1095 forms for previous years changing. Also, with name, there are cases where a veteran's name could change (like if they get married or something), and in that case we also wouldn't want the 1095 for previous years changing

stephte:
actually, we are only worrying about the 'most recent' 1095B form for now. But I still have the same worry for the address/name. If a veteran moves in January 2022, their 1095B form should still show their 2021 address on it. Same for name changes (although I'm unsure how name changes work within the VA system as a whole). The name scenario is a pretty rare case probably, but the address scenario is definitely a reason to store the address with the other 1095B data

rileyanderson:
> We need to have all the address data stored as well, since their address could change, and we don't want their address for 1095 forms for previous years changing. Also, with name, there are cases where a veteran's name could change (like if they get married or something), and in that case we also wouldn't want the 1095 for previous years changing

Good points. In that case the form data will need to be encrypted. All of the form data (including non PII) should be put in a  `jsonb` or `text` column (`jsonb` would be my choice) called `form_data_ciphertext`. That way you would only have to encrypt one column and there would only be three columns (not including `timestamps`) in the table: `veteran_icn`, `tax_year`, and `form_data_cyphertext`

```ruby
class CreateForm1095Bs < ActiveRecord::Migration[6.1]
  def change
    create_table :form1095_bs, id: :uuid do |t|
      t.string :veteran_icn, null: false
      t.integer :tax_year, null: false
      t.jsonb :form_data_ciphertext, null: false # Or: t.text :form_data_ciphertext, null: false
      t.timestamps
    end
    add_index :form1095_bs, [:veteran_icn, :tax_year], unique: true
  end
end
```

Then in your model you would have: 
```ruby
class Form1095B < ApplicationRecord
  has_kms_key
  encrypts :form_data, key: :kms_key, **lockbox_options
end
```

More info on encrypted attributes [here](https://depo-platform-documentation.scrollhelp.site/developer-docs/How-to-use-encrypted-attributes-in-vets-api.1911390216.html).

stephte:
ok, pushed updates for the table, working to update the other PRs

stephte:
I'm getting this error when trying to save to the new table. I have put 
```
has_kms_key
encrypts :form_data, key: :kms_key, **lockbox_options
```
into the Form1095B model. 

Any idea why this error is happening with the encryption?

Screen Shot 2022-03-28 at 4 25 09 PM

rileyanderson:
> I'm getting this error when trying to save to the new table. I have put
> 
> ```
> has_kms_key
> encrypts :form_data, key: :kms_key, **lockbox_options
> ```
> 
> into the Form1095B model.
> 
> Any idea why this error is happening with the encryption?
> 
> Screen Shot 2022-03-28 at 4 25 09 PM

Ah yeah. I forgot you have to add the `encrypted_kms_key` column too. My bad. https://github.com/ankane/kms_encrypted#getting-started. Your migration would look like this then:

```ruby
class CreateForm1095Bs < ActiveRecord::Migration[6.1]
  def change
    create_table :form1095_bs, id: :uuid do |t|
      t.string :veteran_icn, null: false
      t.integer :tax_year, null: false
      t.jsonb :form_data_ciphertext, null: false # Or: t.text :form_data_ciphertext, null: false
      t.text :encrypted_kms_key
      t.timestamps
    end
    add_index :form1095_bs, [:veteran_icn, :tax_year], unique: true
  end
end
```

stephte:
ok, thank you

thilton-oddball (approved):

PR #9458, opened 2022-03-24
39044 Modifying confirm demographics request with stationNo as string

kanchanasuriya:
## Description of change

CHIP/confirm-demographics call from vets-api currently fails with `stationNo` being sent as string ([Sentry issue](http://sentry.vfs.va.gov/organizations/vsp/issues/92258/?environment=staging&project=3&query=is%3Aunresolved&statsPeriod=1h)).  PR changes the parameter to string as CHIP standardized all the endpoints to receive `stationNo` as string.  

## Original issue(s)
department-of-veterans-affairs/va.gov-team#39044

- [x] rspec
- [x] rubocop


modules/check_in/app/services/v2/chip/service.rb
```diff
@@ -195,7 +195,7 @@ def confirm_demographics_id_params
 
         {
           patientDfn: hashed_identifiers[:patientDFN],
-          stationNo: hashed_identifiers[:stationNo]
+          stationNo: hashed_identifiers[:stationNo].to_s
```
randomsync, line 198: minor: not 100% sure if it's more idiomatic to do `"#{hashed_identifiers[:stationNo]}"`, will leave it to you.

randomsync (approved):
minor comment, but otherwise lgtm

modules/check_in/app/services/v2/chip/service.rb
kanchanasuriya: Sure, will check when we consolidate these methods used in refresh & confirm-demographics

PR #9459, opened 2022-03-24
Mhv api service object 35657

jscanish:
## Description of change
The MHV inherited proofing api is a two step process. We will hit their API with a user's ICN and get back a correlation_id in response. We can skip this first step if we already have a correlation_id for the user in question.
We will then use that correlation_id to hit the MHV identity info endpoint. MHV will return a hash of identity document information. If the user has a valid MHV identity document they are considered
eligible for inherited proofing.

If they are eligible for inherited proofing we will cache the MHV identity information along with the user's UUID and an auth code we will generate upon success. That code will be relayed to Login.gov
so they may retrieve the relevant MHV indentity information from the cache.

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/35657

Tested with RSpec, and manually on a staging server.


lib/inherited_proofing/mhv/service.rb
```diff
@@ -0,0 +1,62 @@
...
+
+      def initialize(user)
+        @user = user
+      end
+
+      def identity_proof_data
+        return {} if correlation_id.blank?
+
+        data_hsh = mhv_api_request(verification_info_url)
+        data_hsh['identityDocumentExist'].present? ? data_hsh.merge(success_attributes) : data_hsh
```
bosawt, line 26: I think we want to either shove all relevant logic into this service (which means this service would deal with creating the redis object and would return only the code, or nil if identity document does not exist), or we want to remove any logic beyond just calling the service into a service object pattern.

My personal preference would be a service object pattern, as then we'd have a reusable service to mhv with clear inputs and outputs, and a simple service object that controlled the service calls and saving to the redis. It would look like this:

`MHV::Service` (Service) functions
* get_correlation_id (input: icn, output: correlation_id), self explanatory but some error handling about actual service failures or user not found
* get_verification_info (input: correlation_id, output: response hash parsed into attributes), some error handling about service failures or user not found, response hash parsed into a set of expected attributes that are easier to move around in Rails
 
`MHV::InheritedProofingVerifier` (Service Object) functions:
* new (input: user)
* perform (input: user, output: code), this would do the logic of determine if we need to extract correlation_id, whether or not the identity document exists on the verification response, saving to the redis, contextual error handling, etc

lib/inherited_proofing/mhv/configuration.rb
```diff
@@ -0,0 +1,43 @@
...
+      def valid_id_url
+        Settings.mhv.inherited_proofing.valid_id_url
+      end
+
+      def vacct_info_url
+        Settings.mhv.inherited_proofing.vacct_info_url
+      end
+
+      def app_token
+        Settings.mhv.rx.app_token
```
bosawt, line 17: Let's add a line for `mhv.inherited_proofing.app_token`, I know it'll be the same as `rx.app_token` but would be good to copy it in case we need to do separate upkeep

lib/inherited_proofing/mhv/configuration.rb
```diff
@@ -0,0 +1,43 @@
...
+      def connection
+        @connection ||= Faraday.new(
+          base_path,
+          headers: base_request_headers,
+          request: request_options
+        ) do |conn|
+          conn.use :breakers
+          conn.use Faraday::Response::RaiseError
+          conn.response :snakecase
+          conn.response :json, content_type: /\bjson$/
```
bosawt, line 37: We need to add a line to mock this: `conn.response :betamocks if Settings.mhv.inherited_proofing.mock`.

We will then need to add an entry to vets-api-mockdata:
https://github.com/department-of-veterans-affairs/vets-api-mockdata
If you look in that, there is an `mhv` folder, we can add the mocked response in there. I'll be out Monday but I can help on Tuesday if necessary

config/settings.yml
```diff
@@ -426,6 +426,10 @@ mhv:
     collection_caching_enabled: true
   account:
     mock: false
+  inherited_proofing:
+    base_path: https://essapi-sysb.myhealth.va.gov/mhv-api/patient/v1/multicorr
```
bosawt, line 430: We need to set these to fake values and `mock: true`

bosawt (changes requested):
Nice work! I do have a number of comments, some of them are code quality recommendations that I'd say only deal with them if you agree. 

The other stuff is around the Settings, since MHV is a beast to properly test, we need to mock it in a way that we can test it outside of Staging. I put some recommendations and we're going to have to leverage the mockdata repo. It's my bad for not warning about this earlier, I didn't think about the need to mock on localhost until now

config/settings.yml
```diff
@@ -426,6 +426,10 @@ mhv:
     collection_caching_enabled: true
   account:
     mock: false
+  inherited_proofing:
```
bramleyjl, line 429: Can we change these to be under an `mhv:` subheader?

lib/inherited_proofing/mhv/service.rb
```diff
@@ -0,0 +1,62 @@
...
+    # hash of identity data with an included code we will cache and send
+    # to login.gov. If the user is not eligible the result will be a hash
+    # of error information. If the entire service fails for whatever
+    # reason the result will be an empty hash.
+    class Service < Common::Client::Base
+      configuration InheritedProofing::MHV::Configuration
+
+      attr_reader :user
+
+      def initialize(user)
```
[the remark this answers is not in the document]
bramleyjl: I agree here, the way I interpreted these services was to make a one-time connection with the CSP for a specific user that wouldn't try to maintain any data about the user it's calling on behalf of.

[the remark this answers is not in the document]
jscanish: I don't have strong feelings about it either way. I've worked with people who strongly prefer one pattern or the other and tend to do whatever is more commonly accepted at the moment. I'll make the change.

lib/inherited_proofing/mhv/service.rb
jscanish: I actually forgot that I hadn't done the Redis part when I said this was good to review. I'm planning on doing more or less what you described, which is creating a different service that does the work of determining whether redis is necessary etc

lib/inherited_proofing/mhv/configuration.rb
```diff
@@ -0,0 +1,43 @@
...
+
+module InheritedProofing
+  module MHV
+    class Configuration < Common::Client::Configuration::REST
+      def valid_id_url
+        Settings.mhv.inherited_proofing.valid_id_url
+      end
+
+      def vacct_info_url
+        Settings.mhv.inherited_proofing.vacct_info_url
```
bosawt, line 13: why don't we have these urls just the endpoints, and build from the base url, that way we only have to update a single field in the settings for different stacks?

lib/inherited_proofing/mhv/inherited_proofing_verifier.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
```
bosawt, line 2: Need this at the top: require 'inherited_proofing/mhv/service'

config/settings.yml
```diff
@@ -426,6 +426,10 @@ mhv:
...
+    base_path: https://somethingfake.com
```
bosawt, line 430: Need to use this for some reason: `    base_path: https://mhv-api.example.com`

lib/inherited_proofing/mhv/service.rb
```diff
@@ -0,0 +1,45 @@
...
+        end
+
+        def correlation_id_api_request(icn)
+          perform(:get, correlation_id_url(icn), nil, headers)
+        rescue Common::Client::Errors::ClientError
+          {}
+        end
+
+        def verification_info_api_request(correlation_id)
+          perform(:get, verification_info_url(correlation_id), nil, headers)
```
bosawt, line 38: Need to parse the response so it returns a friendly hash

bramleyjl (approved):

thilton-oddball (approved):

PR #9460, opened 2022-03-24
RRD: add rescue within rrd_processor_failed_handler

yoomlam:
## Description of change

If an error is raised within `rrd_processor_failed_handler`, notify the RRD team and fall back to calling `start_evss_submission_job`.

## Original issue(s)
`rrd_processor_failed_handler` is called when a particular Sidekiq::Batch job fails all retries. However if an error is raised within `rrd_processor_failed_handler` there is no remedy.

No change in normal behavior except to add a rescue block.


app/models/form526_submission.rb
```diff
@@ -90,7 +90,14 @@ def rrd_processor_failed_handler(_status, options)
...
-
+    submission.start_evss_submission_job
+  rescue => e
+    message = <<~MESSAGE
+      RRD was skipped for submission #{submission.id} due to an error.<br/>
+      Sidekiq Job options: #{options}<br/>
+      Exception: #{e}<br/>
+    MESSAGE
+    submission.rrd_process_selector.send_rrd_alert(message)
     submission.start_evss_submission_job
```
yoomlam, line 101: I could use `ensure`:
```ruby
ensure
  submission.start_evss_submission_job
```
but I didn't like that it may be misinterpreted as always run `submission.start_evss_submission_job`, which is not the case when the `return submission.start_rrd_job(backup_processor_class)` line is executed.

spec/models/form526_submission_spec.rb
```diff
@@ -89,6 +89,22 @@
                                                             'submission_id' => form_for_hypertension.id,
                                                             'use_backup_processor' => false)
           end
+
+          context 'when an error is raised within rrd_processor_failed_handler' do
+            before do
+              allow_any_instance_of(RapidReadyForDecision::ProcessorSelector)
+                .to receive(:processor_class).and_raise('Any error')
```
yoomlam, line 96: Picked an arbitrary method call to raise an error.

nanotone (approved):

thilton-oddball (approved):

PR #9461, opened 2022-03-24
Mhv identity model 35659

jscanish:
## Description of change
This PR creates a redis-backed model to temporarily store MHV inherited proofing data. This will be used as part of the larger inherited proofing initiative which involves storing data retrieved via API from MHV and forwarding that data to Login.gov

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/35657

### DIFF
--- a/app/models/inherited_proofing/mhv_identity_data.rb
+++ b/app/models/inherited_proofing/mhv_identity_data.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+module InheritedProofing
+  class MHVIdentityData < Common::RedisStore
+    redis_store REDIS_CONFIG[:mhv_identity_data][:namespace]
+    redis_ttl REDIS_CONFIG[:mhv_identity_data][:each_ttl]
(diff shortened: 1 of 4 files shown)


app/models/inherited_proofing/mhv_identity_data.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module InheritedProofing
```
bosawt, line 3: What do you think about adding specs? I've heard some arguments against adding validator specs (I guess the argument is they're unnecessary), but what I do like is it forces you to create a factorybot object and just sets up the file which helps for specs in the future. 

Refer to this for an example: https://github.com/department-of-veterans-affairs/vets-api/pull/9364/files

bosawt (approved):
I think this looks good now! Thanks for adding the specs!

bramleyjl (approved):

rileyanderson (approved):

PR #9464, opened 2022-03-24
Fixing flaky spec in payment information request spec

bosawt:
## Description of change
This PR fixes a flaky spec. This particular test isn't testing the direct deposit email job, and the spec fails here often, so this PR mocks the call instead of calling on sidekiq here


modules/mobile/spec/request/payment_information_request_spec.rb
```diff
@@ -175,6 +175,7 @@
 
     context 'with a valid evss response' do
       it 'matches the ppiu schema' do
+        allow(DirectDepositEmailJob).to receive(:send_to_emails)
```
kpethtel, line 178: Trevor, thanks for fixing this. I wasn't aware it was flaky. I agree with your rationale and you could probably even stub it a level higher 
`allow(Mobile::V0::PaymentInformation::Proxy).to receive(:send_confirmation_email)`
But I'm curious why this is causing the test to be flaky. I assume it's some weird rspec issue in which the test is done but tested code is still running. Any insight I can learn from?

kpethtel (approved):
Looks good. Thanks for helping keep the CI suite running! 👍

modules/mobile/spec/request/payment_information_request_spec.rb
bosawt: @kpethtel Just as a follow up, I was seeing this error in many places even after this fix, so I think the root cause was actually unrelated to this spec. Looks like it was test pollution and not from my thoughts above: https://github.com/department-of-veterans-affairs/vets-api/pull/9533

PR #9227, opened 2022-02-25
Offramp from RRD if veteran has pending EP 020

nanotone:
## Description of change
This PR changes the DisabilityCompensationJob class to off-ramp in the event that the claim is attached to any pending EPs.

Additionally, there's a fix related to the use of the[ `return` vs `next` keywords](https://github.com/department-of-veterans-affairs/vets-api/pull/9344/files#r826368786).

## Original issue(s)

https://www.notion.so/Set-up-monitoring-of-off-ramped-EP400-claims-2269302a9a934743b4126f717cf23804

PR #9467, opened 2022-03-25
RRD: reword notification to 'Not processed'

yoomlam:
## Description of change

Change the notification text to be the more generic 'Not processed' and enumerate possible reasons.

## Original issue(s)
Our current RRD notification states "Insufficient Data" if a claim is not RRD processed. We now have another reason for not RRD-processing a claim due to #9227.

Only a change to the notification text.


app/mailers/views/hypertension_fast_track_pilot_mailer.erb.html
```diff
@@ -5,12 +5,12 @@
...
     <p>
+      A single-issue <%= @disability_code %> claim for increase was submitted on va.gov.<br/>
       <% if @rrd_claim_processed %>
-        A single-issue <%= @disability_code %> claim for increase was submitted on va.gov.<br/>
         A health summary PDF was generated and added to the claim's documentation.
       <% else %>
-        A single-issue <%= @disability_code %> claim for increase was submitted on va.gov.<br/>
-        There was not sufficient data to generate a health summary PDF associated with this claim. 
+        There was some issue (such as insufficient health data or pending EP) that prevents
+        a health summary PDF to be generated for this claim.
```
allthesignals, line 13: Is there no way to infer the exact cause here? I'm guessing not, just wondering if it's something we'd be able to determine later in the future

allthesignals (approved):

app/mailers/views/hypertension_fast_track_pilot_mailer.erb.html
yoomlam: We could run the same `pending_ep?` code but that's inefficient. Yang is working on storing the reason with the submission record so we can use for this emailer, as well as diagnostics.

rileyanderson (approved):

PR #9471, opened 2022-03-25
Api 14669 max length header tests

stiehlrod:
## Description of change
We've tested max widths for form_data, and now with this ticket, we have done the same for any fields that exist in the auth_headers. 

## Original issue(s)
[API-14669](https://vajira.max.gov/browse/API-14669)


modules/appeals_api/config/schemas/v2/200996_headers.json
```diff
@@ -17,6 +17,7 @@
         "X-VA-SSN": {
           "type": "string",
           "description": "Veteran's SSN",
+          "maxLength": 9,
```
cilestin, line 20: Since HLR v2 is in active production use we can't change schemas as readily as those we haven't released yet. Let's not worry about these maxLengths since the length is enforced by the pattern already.

cilestin (changes requested):
Small change request, but the rest looks good to go!

PR #9473, opened 2022-03-26
API-13229 HLRv2 & NOD Non-Veteran Claimant email support

cilestin:
## Description of change
Updates email event handler(s) to work with claimant data. Will only send the email if claimant email is present, and will skip sending an email entirely if other claimant data is present except for email (since it's optional on the forms).

⚠️ There's a related devops PR to add the template to our settings

## Original issue(s)
https://vajira.max.gov/browse/API-13229

* Properly hooked up for HLRv2 and NODv2
* Stubbed implementation for SC since we haven't completed our Claimant research yet.


stiehlrod (approved):
LGTM

PR #9474, opened 2022-03-26
API-14808 Fix error report skipping stale records

cilestin:
## Description of change
Ensures that the daily error report is sent even when the only thing to report are stale appeals.

## Original issue(s)
https://vajira.max.gov/browse/API-14808


stiehlrod (approved):
LGTM

PR #9478, opened 2022-03-28
RRD: rename hypertension classes

yoomlam:
## Description of change

For RRD classes, make the classes general by renaming them.

## Original issue(s)
RRD classes are named specifically with "Hypertension" when they should be more general.

No behavior change; just rename classes.


yoomlam:
Other things to be changed:
* https://github.com/department-of-veterans-affairs/vets-api/blob/1b391e26acb600683156295d04e82812aef1c2a4/app/services/rapid_ready_for_decision/fast_track_pdf_upload_manager.rb#L7
* https://github.com/department-of-veterans-affairs/vets-api/blob/ff956111f20c22c3db77279041caf631e96bd5d2/app/services/rapid_ready_for_decision/fast_track_pdf_generator.rb#L45-L46

To investigate:
* https://github.com/department-of-veterans-affairs/vets-api/blob/875b0ea1456ce4e9eb427b47e679091f0a2b65ab/app/models/evss_claim_document.rb#L26

app/services/rapid_ready_for_decision/fast_track_pdf_generator.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module RapidReadyForDecision
-  class HypertensionPdfGenerator
+  class FastTrackPdfGenerator
```
yoomlam, line 4: @allthesignals To customize for specific disabilities, new `HypertensionPdfGenerator` and `AsthmaPdfGenerator` classes that are subclasses of `FastTrackPdfGenerator` can be created. The new subclasses will probably have different implementations of `add_blood_pressure_outro`.

allthesignals: @yoomlam how might this class delegate the type of rrd? Maybe it's handed the type-specific class

app/services/rapid_ready_for_decision/lighthouse_medication_request_data.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module RapidReadyForDecision
-  class HypertensionMedicationRequestData
+  class LighthouseMedicationRequestData
```
allthesignals, line 4: Great

app/services/rapid_ready_for_decision/rrd_special_issue_manager.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module RapidReadyForDecision
-  class HypertensionSpecialIssueManager
+  class RrdSpecialIssueManager
```
allthesignals, line 4: Does `Rrd` need to be here? Why not `FastTrack`? or just `SpecialIssueManager`

allthesignals (approved):
I don't feel super strongly about the particulars of naming — I see a trend towards generic naming (away from hypertension-specific). These names signal something generic which serves the purpose. That said, I just left some thoughts/ideas on naming.

app/services/rapid_ready_for_decision/fast_track_pdf_generator.rb
yoomlam: You could pass the disability (`disability_symbol`) as an argument into the call to `new`.
Or create and use a method defined something like this:
```ruby
def single_disability_for(form526_submission)
  mapped_submission_disabilities = RapidReadyForDecision::Constants.extract_disability_symbol_list(form526_submission)
  disability_symbol = mapped_submission_disabilities.first
  disability_struct = RrdConstants::DISABILITIES[disability_symbol]
end
```
See https://github.com/department-of-veterans-affairs/vets-api/pull/9403#discussion_r833599970

app/services/rapid_ready_for_decision/fast_track_pdf_upload_manager.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module RapidReadyForDecision
-  class HypertensionUploadManager
+  class FastTrackPdfUploadManager
```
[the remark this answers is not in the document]
yoomlam: Feel free to rename the class (and associated spec and vcr cassettes) when the class is made more generic.

thilton-oddball (approved):

PR #9480, opened 2022-03-28
RRD: Update claim established message for pending EP offramp

nanotone:
## Description of change
This PR sets aside an `rrd_metadata` namespace for storing arbitrary RRD-related metadata in `Form526Submission#form_json`, and uses it to flag certain claims (offramped due to pending EP 020) to the HypertensionFastTrackPilotMailer.

## Original issue(s)
[Notion ticket](https://www.notion.so/Dashboard-RRD-Health-Evidence-24d77d9988a546c3b19a3c73e7ced29b?p=2269302a9a934743b4126f717cf23804)

- This PR moves form_json metadata from `rrd_med_stats` into the nested `rrd_metadata.med_stats`. This can cause an incompatibility between data written in the old scheme, and code expecting data to be in the new scheme; however, it will only affect non-mission-critical mailer notifications, so I decided not to do this in multiple PRs.

## Testing
- Updated Form526HypertensionJob spec to check that `rrd_metadata.pending_ep` gets correctly set
- Updated existing tests to use `rrd_metadata` in the form json
- Updated HypertensionFastTrackPilotMailer spec to check that pending_ep claims result in the correct mailer message


app/workers/rapid_ready_for_decision/disability_compensation_job.rb
```diff
@@ -21,12 +21,11 @@ def perform(form526_submission_id)
 
       begin
         with_tracking(self.class.name, form526_submission.saved_claim_id, form526_submission_id) do
-          # Prefer "next" keyword here so that after-yield code continues execution
-          next if RapidReadyForDecision::Form526BaseJob.pending_eps?(form526_submission.auth_headers)
+          return if RapidReadyForDecision::Form526BaseJob.pending_eps?(form526_submission)
```
nanotone, line 24: Reverted this to `return` (and removed the comment) now that #9450 is merged, fixing a bug where `return` would skip some important clean-up logic.

nanotone:
In the interest of keeping this PR smaller, here are things I've decided not to include:
- Pulling the descriptions and long descriptions out into a separate mapping
- Starting up an enumeration of offramp reasons (yes, even though insufficient data is Right There)
- Moving "rrd_metadata" code into a more general-purpose Concern
- Minimizing DB writes if business logic touches `form_json` multiple times in a chunk of code

app/mailers/hypertension_fast_track_pilot_mailer.rb
```diff
@@ -5,22 +5,21 @@ def build(submission)
...
-    subject =
-      if @rrd_claim_processed
-        'RRD claim - Processed'
-      else
-        'RRD claim - Insufficient Data'
-      end
-    subject += " - #{@disability_code}"
+    @description = if RapidReadyForDecision::Form526BaseJob.rrd_claim_processed?(submission)
+                     'Processed'
+                   elsif submission.form.dig('rrd_metadata', 'offramp_reason') == 'pending_ep'
```
allthesignals, line 12: This logic is similar to logic in `Form526BaseJob.rrd_claim_processed?` (spelunking through form_json) — I wonder if it should be another class method like `rrd_claim_processed?`?

app/workers/rapid_ready_for_decision/disability_compensation_job.rb
allthesignals: Great

app/workers/rapid_ready_for_decision/form526_base_job.rb
```diff
@@ -21,16 +21,18 @@ def self.rrd_claim_processed?(submission)
 
     # Fetch all claims from EVSS and return whether there are any open EP 020's.
     # This method could be moved into a Concern when ProcessorSelector adds new job classes.
-    def self.pending_eps?(auth_headers)
-      all_claims = EVSS::ClaimsService.new(auth_headers).all_claims.body
-      all_claims['open_claims'].any? { |claim| claim['base_end_product_code'] == '020' }
+    def self.pending_eps?(form526_submission)
+      all_claims = EVSS::ClaimsService.new(form526_submission.auth_headers).all_claims.body
+      pending = all_claims['open_claims'].any? { |claim| claim['base_end_product_code'] == '020' }
+      add_metadata(form526_submission, offramp_reason: 'pending_ep') if pending
```
allthesignals, line 27: Not that it matters here but out of curiosity how would this behave in the context of a Sidekiq job?

Tiny nit, but is there a style convention for referencing class methods inside the class in which they're defined? I would almost prefer `self.add_metadata(...)`...

app/mailers/hypertension_fast_track_pilot_mailer.rb
```diff
@@ -5,22 +5,21 @@ def build(submission)
...
+    @bp_readings_count = submission.form.dig('rrd_metadata', 'med_stats', 'bp_readings_count') || 'N/A'
 
-    subject =
-      if @rrd_claim_processed
-        'RRD claim - Processed'
-      else
-        'RRD claim - Insufficient Data'
-      end
-    subject += " - #{@disability_code}"
+    @description = if RapidReadyForDecision::Form526BaseJob.rrd_claim_processed?(submission)
```
allthesignals, line 10: I wonder if there could be a method on Form526BaseJob or something that could determine `rrd_status` and return symbols? Not necessary right this second but might be useful in the future, in case this needs to be centralized somewhere.

allthesignals (approved):

allthesignals:
> In the interest of keeping this PR smaller, here are things I've decided not to include:
> 
> * Pulling the descriptions and long descriptions out into a separate mapping
> * Starting up an enumeration of offramp reasons (yes, even though insufficient data is Right There)
> * Moving "rrd_metadata" code into a more general-purpose Concern
> * Minimizing DB writes if business logic touches `form_json` multiple times in a chunk of code

Ah woops, sorry, I saw this comment after reviewing.

app/mailers/views/hypertension_fast_track_pilot_mailer.erb.html
```diff
@@ -6,11 +6,13 @@
...
+      <% case @description
+         when 'Processed' %>
         A health summary PDF was generated and added to the claim's documentation.
-      <% else %>
-        There was some issue (such as insufficient health data or pending EP) that prevents
-        a health summary PDF to be generated for this claim.
+      <% when 'Pending EP' %>
+        There was already a pending EP 020 for the veteran associated with this claim.
+      <% when 'Insufficient Data' %>
+        There was not sufficient data to generate a health summary PDF associated with this claim.
```
yoomlam, line 15: Perhaps a cleaner version of this:
```ruby
      <% case @rrd_status
         when :processed %>
        A health summary PDF was generated and added to the claim's documentation.
      <% when :pending_ep %>
        There was already a pending EP 020 for the veteran associated with this claim.
      <% when :insufficient_data %>
        There was not sufficient data to generate a health summary PDF associated with this claim. 
```
where `@rrd_status` is  returns a symbol like [Matt suggested](https://github.com/department-of-veterans-affairs/vets-api/pull/9480/files?diff=split&w=0#r836777828)

yoomlam (dismissed):
👍 Only [one suggestion](https://github.com/department-of-veterans-affairs/vets-api/pull/9480/files#r836815123)

app/mailers/hypertension_fast_track_pilot_mailer.rb
```diff
@@ -5,22 +5,14 @@ def build(submission)
...
-    subject += " - #{@disability_code}"
+    @rrd_status = RapidReadyForDecision::Form526BaseJob.rrd_status(submission)
+    @bp_readings_count = submission.form.dig('rrd_metadata', 'med_stats', 'bp_readings_count') || 'N/A'
 
     template = File.read('app/mailers/views/hypertension_fast_track_pilot_mailer.erb.html')
 
     mail(
       to: Settings.rrd.event_tracking.recipients,
-      subject: subject,
+      subject: "RRD claim - #{@rrd_status.to_s.humanize} - #{@disability_code}",
```
nanotone, line 15: In line with my original list of things to not do in this PR, I'm relying on `String#humanize` here to easily boil down the status into a single symbol. This does mean `:pending_ep` will display as "Pending ep" 🤷

yoomlam (dismissed):
Just the lints remain.

yoomlam (approved):

rileyanderson (approved):

PR #9489, opened 2022-03-29
RRD: Move more existing Hypertension PDF into templates

allthesignals:
## Description of change

This PR continues the work of migrating hypertension PDF view logic into templates. These changes specifically allow for re-use of medications view logic for upcoming asthma PDF work.

This PR also adds a conditional after each block that renders for a given rspec example into the tmp folder. It can be enabled with the `_SAVE_RRD_PDF_FILE` env variable:

`_SAVE_RRD_PDF_FILE=true rspec ...`

Afterward it will print the paths:

image

## Original issue(s)
Related ticket: https://www.notion.so/Dashboard-RRD-Health-Evidence-24d77d9988a546c3b19a3c73e7ced29b?p=10f4de575ee3463fa08265761b45bd9a


app/services/rapid_ready_for_decision/hypertension_pdf_generator.rb
```diff
@@ -74,86 +80,21 @@ def add_intro
...
-      ]
-
-      bp_intro_lines.each do |line|
-        @pdf.text line, inline_format: true
-      end
-
-      return @pdf unless blood_pressure_data?
-
-      @pdf.text "\n", size: 10
-    end
```
allthesignals, line 99: This copy is now handled in the new template

app/services/rapid_ready_for_decision/hypertension_pdf_generator.rb
```diff
@@ -74,86 +80,21 @@ def add_intro
...
 
-      @medications.each do |medication|
-        @pdf.text medication['description'], size: 11, style: :bold
-        @pdf.text "Prescribed on: #{medication['authoredOn'][0, 10].to_date.strftime('%m/%d/%Y')}"
-        if medication['dosageInstructions'].present?
-          @pdf.text "Dosage instructions: #{medication['dosageInstructions'].join('; ')}"
-        end
-        @pdf.text "\n", size: 8
-      end
+      template = File.read('app/services/rapid_ready_for_decision/views/shared/medications.erb')
```
allthesignals, line 96: Note this is located in a "shared" view folder

yoomlam (changes requested):
Some questions and a suggestion: I think `pdf_helper.rb` could be implemented instead as a method in the pdf generator class. Create a test for that method and folks can call that method to create a pdf file locally.

spec/pdf_helper.rb
```diff
@@ -0,0 +1,17 @@
...
+  def lookatit
+    require 'tempfile'
+    f = Tempfile.new(%w[spec- .pdf])
+    f.close
+    compiled_pdf.render_file(f.path)
+    command = /darwin/ =~ RUBY_PLATFORM ? 'open' : 'xdg-open'
+    `#{command} #{f.path} &`
+  ensure
+    sleep 1
+    f.unlink
```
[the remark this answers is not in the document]
allthesignals: @yoomlam I'm not sure why but the docs encourage it. I wonder if it'd be nice to have a "render mode" that prints the PDF for every it block, naming the file after the `it` block... thoughts?

> When a [Tempfile](https://ruby-doc.org/stdlib-2.5.3/libdoc/tempfile/rdoc/Tempfile.html) object is garbage collected, or when the Ruby interpreter exits, its associated temporary file is automatically deleted. This means that's it's unnecessary to explicitly delete a [Tempfile](https://ruby-doc.org/stdlib-2.5.3/libdoc/tempfile/rdoc/Tempfile.html) after use, though it's good practice to do so: not explicitly deleting unused [Tempfile](https://ruby-doc.org/stdlib-2.5.3/libdoc/tempfile/rdoc/Tempfile.html)s can potentially leave behind large amounts of tempfiles on the filesystem until they're garbage collected. The existence of these temp files can make it harder to determine a new Tempfile filename.
> 
> Therefore, one should always call [unlink](https://ruby-doc.org/stdlib-2.5.3/libdoc/tempfile/rdoc/Tempfile.html#method-i-unlink) or close in an ensure block

https://ruby-doc.org/stdlib-2.5.3/libdoc/tempfile/rdoc/Tempfile.html#class-Tempfile-label-Explicit+close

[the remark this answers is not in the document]
yoomlam: AFAIK, the need to create a PDF file during testing is very limited scenario and typically for debugging or limited manual testing. Given that, I think a simple method in the pdf generator class is sufficient for now and keep the rspecs simple.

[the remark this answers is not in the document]
allthesignals: I went ahead and removed all of this and took a different approach that's a little less intrusive.

app/services/rapid_ready_for_decision/fast_track_pdf_generator.rb
```diff
@@ -74,86 +80,21 @@ def add_intro
...
-                  inline_format: true, size: 11
-        @pdf.text "Taken on: #{bp[:effectiveDateTime].to_date.strftime('%m/%d/%Y')} " \
-                  "at #{Time.iso8601(bp[:effectiveDateTime]).strftime('%H:%M %Z')}",
-                  size: 11
-        @pdf.text "Location: #{bp[:organization] || 'Unknown'}", size: 11
-        @pdf.text "\n", size: 8
-      end
-
-      @pdf.text "\n", size: 12
+      template = File.read('app/services/rapid_ready_for_decision/views/hypertension/blood_pressure_readings.erb')
```
yoomlam, line 84: Reminder: rename `hypertension` directory

yoomlam (approved):
Left a reminder to do in a separate PR.

app/services/rapid_ready_for_decision/fast_track_pdf_generator.rb
allthesignals: Aaah thank you! Good catch!

PR #9490, opened 2022-03-29
2732 staging tests

kpethtel:
## Description of change
This adds two scripts:
1) a script for automatically testing endpoints based on yaml inputs 
2) a script for fetching api tokens

## Original issue(s)
department-of-veterans-affairs/va.gov-team#2732

This is a somewhat limited implementation. I didn't want to go too deep down this rabbit hole until I know that this is something we truly want.

It's only designed to make GET requests at this time. In the long term, I would expect it to run full CRUD build/teardown sequences, but for an initial implementation I didn't want to add anything that would actually make changes on the server. 

I spent significant time looking at different options for fetching api tokens. One possible method (based on a defunct script we used at one time) was to bypass external services, create a token locally on the server, and add that token to redis. That could work on jenkins using a "review instance" created for the job, but that's not really what we wanted. 

I also looked at how Patrick Vinograd fetches tokens in the heroku cutter app. That approach involves automating the process of getting the token from our external auth service. The problem with this is that it requires having a token for that auth service. We probably don't want individual developers having that token, which would limit our ability to use this script. 

So I settled on scraping the api token from the heroku cutter web interface. It's a bit flaky and may need some refinement, but it can be run locally and on jenkins, so it will be useful both to developers and for automated testing.

Ideally, this code should not exist in our rails repo. It should be in a separate scripts repo. We can move it if we find a good location for it. 

I haven't written any specs for this yet because it's unclear if this is what we want. Tests would also be fairly limited due to the nature of these scripts.


modules/mobile/bin/health_checks/Gemfile
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
```
kpethtel, line 1: I haven't yet added a Gemfile.lock because that seems to be git ignored in this directory. As mentioned, I don't think this is the ideal location for this to live anyhow.

modules/mobile/bin/health_checks/endpoint_tester.rb
```diff
@@ -0,0 +1,125 @@
...
+    correct_status_received = expected_status == response.status
+
+    if correct_status_received
+      received_data = JSON.parse(response.body)['data']
+
+      count = expected_data.dig('response', 'count')
+      add_error('count', count, received_data.count) if count && count != received_data.count
+
+      expected_data = expected_data.dig('response', 'data')
+      # ensuring that both data sets are hashes to avoid having a special case when data is an array
```
kpethtel, line 65: I don't love that I have to convert these to objects, but it makes the `compare_data` method wants to have a key to use as a descriptor. I considered other options but nothing seemed any better.

modules/mobile/bin/health_checks/endpoint_tester.rb
```diff
@@ -0,0 +1,125 @@
...
+
+    puts
+    puts "Successful tests: #{successes.count}".green
+    if failures.any?
+      puts "Failed tests: #{failures.count}".red
+      failures.each do |name, messages|
+        puts "Test #{name} failed with errors:".red
+        messages.each { |message| puts "\t#{message.red}" }
+      end
+      abort
```
kpethtel, line 104: Using abort instead of raise because it results in a cleaner command line interface. It will end the script without outputting any stacktrace but will also have a non-successful exit status, which could be helpful to us in automation.

modules/mobile/app/services/mobile/v0/sessions/token_fetcher.rb
```diff
@@ -0,0 +1,55 @@
...
+  def initialize(user_name)
+    set_user(user_name)
+    Capybara.default_max_wait_time = 10
+    @session = Capybara::Session.new(:selenium)
+  end
+
+  def fetch_token
+    @session.visit TOKEN_URL
+    @session.click_link('Please Login')
+    sleep 3 # this sleep appears to be necessary for dealing with the redirect
```
jperk51, line 22: instead of having a sleep is it possible to hold until the 'Sign in with ID.me' or some other button from the next page is detected?

modules/mobile/bin/health_checks/token_fetcher.rb
```diff
@@ -0,0 +1,54 @@
...
+  def initialize(user_name)
+    set_user(user_name)
+    Capybara.default_max_wait_time = 10
+    @session = Capybara::Session.new(:selenium)
+  end
+
+  def fetch_token
+    @session.visit TOKEN_URL
+    @session.click_link('Please Login')
+    sleep 3 # this sleep appears to be necessary for dealing with the redirect
```
kpethtel, line 21: @jperk51 the file you commented on was in the wrong directory (i started working on this in another directory and failed to delete some of that work). Capybara is supposed to wait for page load by default, but I wasn't able to get it working without the sleep here. I assume it's because a number of redirects are made, but I'm actually not certain why. I'm definitely open to better solutions.

jperk51: Gotcha, I think this is probably fine overall, but could fail with slow connections is all. In the past I've set up webcrawlers with loops that are basically halting execution until a known element of the next page is detected. Not sure if that's really worth doing here or not.

kpethtel: It does sometimes fail on one of the "Accept" buttons not being there but it seems to work otherwise. We can definitely iterate on this. A loop might be better than a flat sleep. It's a good suggestion. I'll give it a try. But my overall understanding is that modern capybara should wait for elements to render. I just didn't see it working in this one spot.

modules/mobile/bin/health_checks/endpoint_tester.rb
```diff
@@ -0,0 +1,127 @@
...
+
+    @current_test.empty? ? print('.'.green) : print('.'.red)
+  end
+
+  def compare(expected, received)
+    expected.each do |k, v|
+      case v
+      when Hash
+        compare(v, received[k])
+      when Array
```
kpethtel, line 78: This has a limitation that I'm currently working on.

modules/mobile/bin/health_checks/request_data/appointments/get_appointments.yaml
```diff
@@ -0,0 +1,26 @@
+sequence:
```
kpethtel, line 1: These are obviously not intended to be full body checks. I'm just getting a few basic checks up to show that we can test status and some body spot checking.

modules/mobile/bin/health_checks/token_fetcher.rb
```diff
@@ -0,0 +1,55 @@
...
+    @session.click_button('Sign in with ID.me')
+    @session.click_button('Accept')
```
kpethtel, line 23: I sometimes get errors here, probably because clicking the sign in button failed. I'll need to improve it later.

aherzberg (commented):
Overall looks great. Not to be done in this PR but we could leverage this to get the test user tokens easily for manual testing instead of going through the heroku app ourselves. Noticed a couple things:

1. I had trouble setting up the staging_test_users.yml since I didn't realize it was using more then just the judy user. May be worth creating a yml template file that has a spot for each user expected then let the tester fill out the credentials somewhere outside the repo to ensure they don't accidentally push the credentials. 

2. Would making this a headless process be as easy giving it a headless browser binary to run from? If so, that may be worth doing. As this test suite grows, it'll be running for long periods and be less taxing on the system and be less of a distraction.

aherzberg (approved):
LGTM

modules/mobile/bin/health_checks/request_data/appointments/get_appointments.yaml
```diff
@@ -0,0 +1,25 @@
...
+  - case:
+      name: V0 get appointments judy
```
kpethtel, line 3: One of my thoughts was to have a services or upstream_dependencies field here that lists an array of upstream services based on [this chart](https://department-of-veterans-affairs.github.io/va-mobile-app/docs/Engineering/BackEnd/Architecture/Services). We could use those lists for failing requests to try to determine what's failing.

PR #9499, opened 2022-03-30
adding additional logging for appointments errors

jperk51:
## Description of change
Noticing some errors in staging that are hard to diagnose at the moment so we are adding logs for the error details when we catch them

logging only


aherzberg (approved):
LGTM

PR #9500, opened 2022-03-30
API-14634 - Add claims submission logging

acovrig:
## Description of change

Add a logging facility to log 526-EZ, PoA, and ITF events.

## Original issue(s)
[API-14634](https://vajira.max.gov/browse/API-14634)


modules/claims_api/lib/claims_api/claim_logger.rb
```diff
@@ -0,0 +1,50 @@
...
+      params.key?(:level) && params[:level].to_sym.in?(LEVELS) ? params[:level].to_sym : :info
+    end
+
+    def self.format_msg(tag, **params)
+      msg = ['ClaimsApi', tag]
+      case tag
+      when '526ez'
+        msg.append("Claim ID: #{params[:claim_id]}") if params[:claim_id].present?
+        msg.append("VMBS ID: #{params[:vbms_id]}") if params[:vbms_id].present?
+        msg.append("autoCrestPDFGenerationDisabled: #{params[:pdf_gen_dis]}") || false
```
jw81, line 31: Two things here.

1. It's `autoCestPDFGenerationDisabled` (not `autoCrestPDFGenerationDisabled`)
2. Isn't this always going to append `autoCrestPDFGenerationDisabled` to the log message?  Like, even when we don't care about logging the state of that attribute?

modules/claims_api/lib/claims_api/claim_logger.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+    def self.pick_level(**params)
+      params.key?(:level) && params[:level].to_sym.in?(LEVELS) ? params[:level].to_sym : :info
+    end
+
+    def self.format_msg(tag, **params)
+      msg = ['ClaimsApi', tag]
+      case tag
+      when '526ez'
```
jw81, line 28: Should consider making the tag values into constants.
So we can call the logger like :: `ClaimsApi::Logger.log(ClaimsApi::Logger::526, claim_id: 'id_here')`

Otherwise, we're gonna run into situations where we try to call the logger like :: `ClaimsApi::Logger.log('526', claim_id: 'id_here')` which won't find a match in the `case` and lead to a useless log message.

jw81: Or add a method that validates that the value given for `tag` is a recognized value.

modules/claims_api/lib/claims_api/claim_logger.rb
```diff
@@ -0,0 +1,50 @@
...
+        msg.append("Claim ID: #{params[:claim_id]}") if params[:claim_id].present?
+        msg.append("VMBS ID: #{params[:vbms_id]}") if params[:vbms_id].present?
+        msg.append("autoCrestPDFGenerationDisabled: #{params[:pdf_gen_dis]}") || false
+        msg.append("Attachment ID: #{params[:attachment_id]}") if params[:attachment_id].present?
+      when 'poa'
+        msg.append("PoA ID: #{params[:poa_id]}") if params[:poa_id].present?
+        msg.append("Status: #{params[:status]}") if params[:status].present?
+        msg.append("Error Code: #{params[:error]}") if params[:error].present?
+      when 'itf'
+        msg.append("ITF: #{params[:status]}") if params[:status].present?
```
jw81, line 38: I wouldn't use `status` here.
The term `status` is used in a bunch of different contexts not only in our API, but in the upstream dependencies we talk to. So I could absolutely guarantee a situation will arise one day where our logging shows a message that says an ITF was `submitted`, but then some actual response/record somewhere has a different value and leads to mass-confusion.

I would recommend changing this to something more like `detail: ITF uploaded to VBMS`.

modules/claims_api/lib/claims_api/claim_logger.rb
```diff
@@ -0,0 +1,50 @@
...
+      msg = ['ClaimsApi', tag]
+      case tag
+      when '526ez'
+        msg.append("Claim ID: #{params[:claim_id]}") if params[:claim_id].present?
+        msg.append("VMBS ID: #{params[:vbms_id]}") if params[:vbms_id].present?
+        msg.append("autoCrestPDFGenerationDisabled: #{params[:pdf_gen_dis]}") || false
+        msg.append("Attachment ID: #{params[:attachment_id]}") if params[:attachment_id].present?
+      when 'poa'
+        msg.append("PoA ID: #{params[:poa_id]}") if params[:poa_id].present?
+        msg.append("Status: #{params[:status]}") if params[:status].present?
```
jw81, line 35: Same recommendation as below.  STRONGLY recommend NOT using a term like `status`.
I would change this to something like, `detail: POA created in Lighthouse` and `detail: BIRLS success` and so on.

modules/claims_api/lib/claims_api/claim_logger.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+module ClaimsApi
+  # Logger for ClaimsApi, mostly used to format a print statement
+  # Uses the standard Rails logger
+  #
+  #   ClaimsApi::Logger.log('526ez', claim_id: 'id_here')
+  #   # => "ClaimsApi :: 526ez :: Claim ID: id_here"
```
jw81, line 8: Love the documentation and usage example.  👍 ❤️

jw81:
**REMINDER** :: We still need to add logging for the `POST /forms/526/:id/attachments` endpoint/background job.
The background job (`ClaimsApi::ClaimUploader`) is currently undergoing some logic tweaks over in this [PR](https://github.com/department-of-veterans-affairs/vets-api/pull/9488) of @hinzed1127 's.  Once Dan's changes are in Master, we'll need to go in there and add logging.

modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
```diff
@@ -44,6 +44,9 @@ def submit_form_526 # rubocop:disable Metrics/MethodLength
             source: source_name
           )
 
+          ClaimsApi::Logger.log('526', claim_id: params[:id], detail: 'Submitted',
```
jw81, line 47: Again, we don't really want to use simple terms like `submitted`.
We use the values `pending, submitted, updated, errored` as `status` values in varying resources within our API.

For instance, here in this POST, the claim is going to always receive a `status` value of `pending`. (see line# 39) So it's going to be confusing when we're looking at logs and we see a log that says it was "submitted", but has an actual `status` value of `pending`.

If you want to use the term "submitted" then it should have more context.  Something like, `Submitted to Lighthouse`.  That way, it'll be much easier to tell exactly what is happening in the logs and where that message is coming from in the code.

modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
```diff
@@ -77,6 +80,7 @@ def upload_form_526
             pending_claim.set_file_data!(documents.first, params[:doc_type])
             pending_claim.save!
 
+            ClaimsApi::Logger.log('526', claim_id: params[:id], detail: 'Uploaded')
```
jw81, line 83: We haven't actually uploaded anything to an upstream VA service at this point.  So the the detail of `Uploaded` is misleading.  All we've done at this point is store a PDF in Amazon S3.  We need a better detail value here.  Something like, `Uploaded PDF to S3`.

modules/claims_api/app/controllers/claims_api/v1/forms/intent_to_file_controller.rb
```diff
@@ -27,6 +27,7 @@ def submit_form_0966
           check_for_invalid_burial_submission! if form_type == 'burial'
 
           bgs_response = bgs_service.intent_to_file.insert_intent_to_file(intent_to_file_options)
+          ClaimsApi::Logger.log('itf', status: 'Submitted')
```
jw81, line 30: Your `case` for `itf` no longer works with a `status` attribute.
Instead, it's now a `detail` attribute.
The detail value should be something like, `submitted to BGS`

modules/claims_api/app/controllers/claims_api/v1/forms/power_of_attorney_controller.rb
```diff
@@ -46,6 +46,7 @@ def submit_form_2122 # rubocop:disable Metrics/MethodLength
             end
 
             power_of_attorney.save!
+            ClaimsApi::Logger.log('poa', poa_id: power_of_attorney.id, status: 'Created in Lighthouse')
```
jw81, line 49: Your `case` for `poa` no longer works with a `status` attribute.
Instead, it's now a `detail` attribute.

modules/claims_api/app/workers/claims_api/claim_uploader.rb
```diff
@@ -23,6 +24,7 @@ def perform(uuid)
         uploader.retrieve_from_store!(claim_object.file_data['filename'])
         file_body = uploader.read
         service(auth_headers).upload(file_body, claim_upload_document(claim_object))
+        ClaimsApi::Logger.log('526', claim_id: auto_claim.id, attachment_id: uuid)
```
jw81, line 27: We need to move this up just before the `service()` call.
That way, even if the service call blows up, we'll always get a log from right before the moment when it happened.

acovrig: Doesn't the service do the upload, I don't want to say that it's been uploaded if it hasn't yet; should I `begin service() ensure log() end` block?

jw81: > I don't want to say that it's been uploaded if it hasn't yet

But you're not doing that.  Right now you're just generating a log that will say, 
`ClaimsApi :: Claim ID: 123ABC :: Attachment ID: 987XYZ`

It doesn't mention anything about uploading anything.

modules/claims_api/lib/claims_api/claim_logger.rb
```diff
@@ -0,0 +1,53 @@
...
+      msg = ['ClaimsApi', tag]
+      case tag
+      when '526'
+        msg.append("Claim ID: #{params[:claim_id]}") if params[:claim_id].present?
+        msg.append("Detail: #{params[:detail]}") if params[:detail].present?
+        msg.append("VBMS ID: #{params[:vbms_id]}") if params[:vbms_id].present?
+        msg.append("autoCestPDFGenerationDisabled: #{params[:pdf_gen_dis]}") if params[:pdf_gen_dis].present?
+        msg.append("Attachment ID: #{params[:attachment_id]}") if params[:attachment_id].present?
+      when 'poa'
+        msg.append("PoA ID: #{params[:poa_id]}") if params[:poa_id].present?
```
hinzed1127, line 35:
```suggestion
        msg.append("POA ID: #{params[:poa_id]}") if params[:poa_id].present?
```
nitpick for consistency: it should always be styled as POA

modules/claims_api/lib/claims_api/claim_logger.rb
```diff
@@ -0,0 +1,53 @@
...
+        msg.append("Detail: #{params[:detail]}") if params[:detail].present?
+        msg.append("Error Code: #{params[:error]}") if params[:error].present?
+      when 'itf'
+        msg.append("ITF: #{params[:detail]}") if params[:detail].present?
+      else
+        msg.append(params.to_json)
+      end
+      called_from = caller_locations(1, 1).first
```
hinzed1127, line 43: [TIL](https://apidock.com/ruby/v2_5_5/Kernel/caller_locations). I like this touch.

modules/claims_api/lib/claims_api/claim_logger.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+module ClaimsApi
+  # Logger for ClaimsApi, mostly used to format a print statement
+  # Uses the standard Rails logger
+  #
+  #   ClaimsApi::Logger.log('526ez', claim_id: 'id_here')
+  #   # => "ClaimsApi :: 526ez :: Claim ID: id_here"
```
hinzed1127, line 8: nitpick to go off of some of Jeff's earlier comments: making this example use `526` instead of `526ez`, since we avoid making that specific distinction elsewhere

hinzed1127 (changes requested):
Just a few super small things, but this has shaped up nicely. Thanks for rolling with the punches. Nailing down a consistent way to log is important.

jw81 (approved):

hinzed1127 (approved):

PR #9509, opened 2022-03-30
38679 Custom cookie in check-in experience for low auth token

kanchanasuriya:
## Description of change

Currently, minimal authentication in session object gets reset with the user signing out of full authentication in the system.   This PR moves minimal auth token to custom cookie and avoids minimal authentication getting interrupted with full authentication. 

## Original issue(s)
department-of-veterans-affairs/va.gov-team#38679

- [x] rubocop
- [x] rspec


thilton-oddball (dismissed):

modules/check_in/app/controllers/check_in/application_controller.rb
```diff
@@ -25,6 +25,22 @@ def additional_logging?
       Flipper.enabled?('check_in_experience_logging_enabled')
     end
 
+    def minimal_auth_token
```
randomsync, line 28: we seem to be using the terms minimal auth and low auth interchangeably. should we standardize on one of them, maybe low auth?

randomsync (commented):
@kanchanasuriya are the test scenarios for this change listed somewhere?

modules/check_in/app/controllers/check_in/application_controller.rb
kanchanasuriya: Maybe, I will make it minimal auth for the feature flag.  If we change everything to low auth, then the cookies object having `min_session` will be out of sync.  

At the same time, not inlined in using low_auth_session in cookie object as it is visible to the client. Please let me know.

kanchanasuriya: As discussed offline, modified the method name and cookie name in the latest commit

thilton-oddball (approved):

randomsync (approved):
lgtm

PR #9511, opened 2022-03-31
Fix Redis.current deprecation warning

ericboehs:
Since Redis 4.6.0, using `Redis.current` to connect to Redis was deprecated. [Gem maintainer byroot recommends](https://github.com/redis/redis-rb/commit/9745e22db65ac294be51ed393b584c0f8b72ae98#commitcomment-65919666) using `$redis` or a connection pool.

While he suggests using a connection pool to prevent thread locking, that is out of scope of this PR as our existing solution does not use thread pooling and the intention of this PR is just to remove the thousands of lines of deprecation notices that is making Vets API work difficult.


spec/controllers/v0/profile/ch33_bank_accounts_controller_spec.rb
```diff
@@ -108,7 +108,7 @@ def send_successful_update
       it 'submits the update req and rerenders index' do
         send_successful_update
 
-        expect_find_ch33_dd_eft_res
+        # expect_find_ch33_dd_eft_res
```
thilton-oddball, line 111: Is there a reason why this is commented out? I tested locally with it uncommented and it passes.

Gemfile
```diff
@@ -112,12 +112,12 @@ gem 'puma'
 gem 'puma-plugin-statsd'
 gem 'pundit'
 gem 'rack'
-gem 'rack-attack'
+gem 'rack-attack', github: 'rack/rack-attack', branch: 'main' # FIXME: Pointing to GH just for testing
```
thilton-oddball, line 115: Does this need to be taken back out now or at a later date?

Gemfile
```diff
@@ -112,12 +112,12 @@ gem 'puma'
...
 gem 'rack-cors', require: 'rack/cors'
 gem 'rails_semantic_logger'
 gem 'rails-session_cookie'
 gem 'redis'
-gem 'redis-namespace'
+gem 'redis-namespace', github: 'resque/redis-namespace' # FIXME: Pointing to GH just for testing
```
thilton-oddball, line 120: Does this need to be taken back out now or at a later date?

ericboehs:
Resolved in another PR.

PR #9518, opened 2022-03-31
RRD: Allow worker subclass to skip release of generated PDF to EVSS

nanotone:
## Description of change
This PR adds an overridable `release_pdf?` method to Form526BaseJob, and overrides it to return false in Form526AsthmaJob. This allows us to generate and inspect RRD PDFs in production, while holding them from being released to RVSR users.

The reason this PR doesn't use a feature flag, is that there is already an `rrd_asthma_compensation` flag turned on for the asthma flow in production, but we want more fine-grained control in the code over which parts of the flow are fully live, and which parts are live but partially handled manually.

## Original issue(s)
[Notion ticket](https://www.notion.so/Dashboard-RRD-Health-Evidence-24d77d9988a546c3b19a3c73e7ced29b?p=11bc06ffb8cc4036bf04ac56469cd6d8)

* No feature flag; rationale is described above
* Updated the FastTrackPdfUploadManager spec to test the new pathway, and did a very minor refactor of that file for concision.


allthesignals (approved):

app/workers/rapid_ready_for_decision/form526_base_job.rb
```diff
@@ -135,10 +140,14 @@ def account(form526_submission)
...
     def upload_pdf(form526_submission, pdf)
-      RapidReadyForDecision::FastTrackPdfUploadManager.new(form526_submission).handle_attachment(pdf.render)
+      RapidReadyForDecision::FastTrackPdfUploadManager
+        .new(form526_submission)
+        .handle_attachment(pdf.render, add_to_submission: release_pdf?(form526_submission))
     end
 
     def set_special_issue(form526_submission)
+      return if release_pdf?(form526_submission)
+
```
yoomlam, line 150: The presumed logic for this is: if we don't want to release the pdf, then the claim shouldn't be fast-tracked (via the `RRD` special issue). Wondering if we should add a code comment to avoid anyone thinking this is a bug since generating the PDF and routing are arguably tangential (but related).

app/workers/rapid_ready_for_decision/form526_base_job.rb
```diff
@@ -72,6 +72,11 @@ def perform(form526_submission_id)
       end
     end
 
+    # Override this method to prevent the submission from getting the PDF and special issue
+    def release_pdf?(_form526_submission)
```
yoomlam, line 76: I'm tempted to turn `form526_submission` into an instance variable to avoid having to pass it around to these methods. Thoughts? My current thinking: this class is instantiated by `Sidekiq::Batch`; `perform(form526_submission_id)` is called -- we can set `@form526_submission` at the beginning of `perform` and unset at the end. The potential problem is if Sidekiq reuses the job instance on another `form526_submission_id` concurrently -- is this possible?
Another approach is to create another class that is instantiated with a `form526_submission` and move most of the methods there. I think I prefer this approach.

yoomlam (approved):

app/workers/rapid_ready_for_decision/form526_base_job.rb
nanotone: These are good topics for our weekly engineers' discussion!

Initial thoughts: agreed, even if I doubt Sidekiq would ever concurrently reuse job instances, it's better to not tie our implementation to that assumption. So maybe the `process_selector` can pass both a Form526Submission ID and a classname over to a whittled-down `#perform` that does the actual instantiation of the subclass?

app/workers/rapid_ready_for_decision/form526_base_job.rb
nanotone: Hm, it actually reads pretty straightforward to me, so I'm not inclined to add a comment unless you feel strongly.

I expect that we may eventually remove this mechanism anyway, unless in the asthma retro we decide that it's a pattern we want to keep and formalize for future contentions.

yoomlam: Not strongly. Go ahead!

thilton-oddball (approved):

PR #9521, opened 2022-03-31
2832: EVSS errors

kpethtel:
## Description of change
We've been having frequent issues with EVSS being down. The data being returned from EVSS is not critical to app function and we're only using it to determine whether the user has access to update their direct deposit benefits. This PR catches the error and just removes the ability to update direct deposit benefits, which will allow the user to use the rest of the app. 

## Original issue(s)
department-of-veterans-affairs/va-mobile-app#2832

If you're curious, the error appears to be caught and re-raised here `lib/evss/ppiu/service.rb#handle_error`


thilton-oddball (dismissed):

jperk51:
Since other service might be using the same authorization policy I moved the rescue into our serializer.

kreek (approved):
LGTM 🚀

PR #9523, opened 2022-03-31
Use single template for Hypertension PDF

allthesignals:
## Description of change
Moves further towards a template-driven approach towards PDF generation. This PR moves all hypertension-related templates into a single template. Implements a "partial" capability for template composability. 

## Original issue(s)
Step towards generalization to prep for https://www.notion.so/Set-up-test-users-5002d145ec6e4e2e8a4be96f5897e79c


app/services/rapid_ready_for_decision/fast_track_pdf_generator.rb
```diff
@@ -19,22 +19,20 @@ class FastTrackPdfGenerator
       }
     }.freeze
 
-    def initialize(patient_info, blood_pressure_data, medications)
+    def initialize(patient_info, assessed_data, disability_type)
```
allthesignals, line 22: Use more-generic `assessed_data`. Plan is to remove data-specific helpers (for things like `:medications`)...

app/services/rapid_ready_for_decision/fast_track_pdf_generator.rb
```diff
@@ -57,51 +55,24 @@ def blood_pressure_end_date
...
-        "<font size='11'>\n</font>"
-      ]
-
-      intro_lines.each do |line|
-        @pdf.text line, inline_format: true
-      end
+    def patient_name
+      first, middle, last, suffix = @patient_info.values_at(:first, :middle, :last, :suffix)
+      full_name = [first, middle, last].reject(&:blank?).join ' '
+      [full_name, suffix].reject(&:blank?).join ', '
```
allthesignals, line 61: This is vestigial from other parts of the code and needs some TLC

app/services/rapid_ready_for_decision/fast_track_pdf_generator.rb
```diff
@@ -57,51 +55,24 @@ def blood_pressure_end_date
...
     end
 
-    def add_blood_pressure_list
-      template = File.read('app/services/rapid_ready_for_decision/views/hypertension/blood_pressure_readings.erb')
-      @pdf.markup ERB.new(template).result(binding)
+    def birthdate
+      @patient_info[:birthdate]
     end
```
allthesignals, line 66: Probably not necessary and can be used directly in the template... `@patient_info[:birthdate]`

app/services/rapid_ready_for_decision/views/hypertension.erb
```diff
@@ -0,0 +1,5 @@
+<%= partial 'hypertension/intro' %>
+<%= partial 'hypertension/blood_pressure_readings' %>
+<%= partial 'hypertension/rating_schedule' %>
```
allthesignals, line 3: So, in order to enforce some encapsulation, maybe this partiail helper should accept arguments explicitly so that these partials don't have access to everything

app/services/rapid_ready_for_decision/fast_track_pdf_generator.rb
```diff
@@ -19,22 +19,20 @@ class FastTrackPdfGenerator
...
 
     # progressively builds a pdf and is sensitive to sequence
     def generate
-      add_intro
-      add_blood_pressure_list
-      add_blood_pressure_outro
-      add_medications_list
-      add_about
+      template = File.read('app/services/rapid_ready_for_decision/views/hypertension.erb')
```
yoomlam, line 34: You could use `@disability_type` here:
```suggestion
      template = File.read("app/services/rapid_ready_for_decision/views/#{@disability_type}.erb")
```

app/services/rapid_ready_for_decision/fast_track_pdf_generator.rb
yoomlam, line 61: Alternative to consider (not that much better):
```suggestion
      full_name = @patient_info.values_at(:first, :middle, :last).reject(&:blank?).join ' '
      [full_name, @patient_info[:suffix]].reject(&:blank?).join ', '
```

app/services/rapid_ready_for_decision/fast_track_pdf_generator.rb
yoomlam: Agree, only the template uses it.

app/services/rapid_ready_for_decision/fast_track_pdf_generator.rb
```diff
@@ -57,51 +55,24 @@ def blood_pressure_end_date
...
     end
 
-    def add_about
-      @pdf.start_new_page
-
-      template = File.read('app/services/rapid_ready_for_decision/views/hypertension/about.erb')
-      @pdf.markup ERB.new(template).result(binding)
+    def partial(erb_file_relative_path, *args)
+      erb_file_full_path = "app/services/rapid_ready_for_decision/views/#{erb_file_relative_path}.erb"
+      ERB.new(File.new(erb_file_full_path).read).result(binding)
```
yoomlam, line 75: From https://stackoverflow.com/questions/597488/how-to-do-a-safe-join-pathname-in-ruby; longer but safer?:
```suggestion
      erb_file_full_path = File.join("app/services/rapid_ready_for_decision/views", "#{erb_file_relative_path}.erb")
      ERB.new(File.new(erb_file_full_path).read).result(binding)
```

app/services/rapid_ready_for_decision/views/hypertension.erb
yoomlam: Opinion: I don't think you need to at this point since they all pertain to hypertension.
Passing in arguments to `shared/medications` on line 4 might be a good idea to reduce assumptions about what variables `shared/medications` has access to.

yoomlam (dismissed):
Looks good. Left some thoughts.

app/services/rapid_ready_for_decision/fast_track_pdf_generator.rb
allthesignals: Oh! Interesting! Wow good catch. Thanks.

app/workers/rapid_ready_for_decision/disability_compensation_job.rb
```diff
@@ -103,7 +103,11 @@ def upload_pdf_and_attach_special_issue(form526_submission, pdf)
     end
 
     def pdf(patient_info, bpreadings, medications)
-      RapidReadyForDecision::FastTrackPdfGenerator.new(patient_info, bpreadings, medications).generate
+      assessed_data = {
+        bp_readings: bpreadings,
+        medications: medications
+      }
+      RapidReadyForDecision::FastTrackPdfGenerator.new(patient_info, assessed_data, :hypertension).generate
```
leikkisa, line 110: I think it would be good to generalize this to put the disability type (it could be symbolized) instead of hypertension directly.

app/services/rapid_ready_for_decision/fast_track_pdf_generator.rb
```diff
@@ -19,22 +19,20 @@ class FastTrackPdfGenerator
...
     end
 
     # progressively builds a pdf and is sensitive to sequence
     def generate
-      add_intro
-      add_blood_pressure_list
-      add_blood_pressure_outro
-      add_medications_list
-      add_about
+      template = File.join('app/services/rapid_ready_for_decision/views', "#{@disability_type}.erb")
```
leikkisa, line 34: I like the idea of passing in the disability type (which I think is happening with having it in initialization and binding), and pointing towards a more generic base template that could use the disability type in logic for showing partials. I think something like this would be easier to scale than having a separate erb template per disability type.

app/services/rapid_ready_for_decision/fast_track_pdf_generator.rb
```diff
@@ -57,51 +55,24 @@ def blood_pressure_end_date
...
+      generated_time = Time.now.getlocal
+      "#{generated_time.strftime('%m/%d/%Y')} at #{generated_time.strftime('%l:%M %p %Z')}"
     end
 
-    def add_about
-      @pdf.start_new_page
-
-      template = File.read('app/services/rapid_ready_for_decision/views/hypertension/about.erb')
-      @pdf.markup ERB.new(template).result(binding)
+    def render_partial(erb_file_relative_path)
```
leikkisa, line 73: Love this!

app/services/rapid_ready_for_decision/views/hypertension.erb
```diff
@@ -0,0 +1,5 @@
+<%= render_partial 'hypertension/intro' %>
```
leikkisa, line 1: I think it would be cool to also make some of these partials more generic (looks like there's a shared folder). I got confirmation from Shannon that if the formatting is updated for the new contention, it's okay to apply it to the previous contention as well. She mentioned maybe wanting some differences with the medications though.

leikkisa (commented):
I like how this is looking so far! I left a couple of comments on more potential places to generalize, based on the description it sounds like that could be scoped to a future PR.

app/services/rapid_ready_for_decision/fast_track_pdf_generator.rb
allthesignals: @leikkisa  I think I like that, it would look something like this:

```erb
<%# views/base.erb %>

<% if test ==:hypertension  %>
  <%= render_partial 'shared/intro' %>
  <%= render_partial 'hypertension/blood_pressure_readings' %>
  <%= render_partial 'hypertension/rating_schedule' %>
  <%= render_partial 'shared/medications' %>
  <%= render_partial 'hypertension/about' %>
<% end %>

<% if test ==:asthma  %>
  <%= render_partial 'shared/intro' %>
  <%= render_partial 'asthma/rating_schedule' %>
  <%= render_partial 'shared/medications' %>
  <%= render_partial 'asthma/about' %>
<% end %>
```

What do you think? Is this what you were thinking of?

app/services/rapid_ready_for_decision/views/shared/medications.erb
```diff
@@ -9,13 +9,14 @@
...
-<% unless medications? %>
+<% unless @assessed_data[:medications].any? %>
+  <p><br/></p>
   <h6>No active medications were found in the last year</h6>
 <% end %>
 
 <p> <br /> </p>
 
-<% @medications.each do |medication| %>
+<% @assessed_data[:medications].each do |medication| %>
```
yoomlam, line 19: If `@assessed_data[:medications]` is nil, this will raise an error. A possible fix:
```suggestion
<% (@assessed_data[:medications] || []).each do |medication| %>
```

app/services/rapid_ready_for_decision/views/hypertension.erb
allthesignals: @leikkisa great. yeah, I was able to make intro into something generic (see new updates)... i could see BP even being pretty generic even though it's only used for this currently

app/services/rapid_ready_for_decision/views/shared/medications.erb
```diff
@@ -9,13 +9,14 @@
   <i>All VAMC locations using VistA/CAPRI were checked</i> <br />
 </p>
 
-<% unless medications? %>
+<% unless @assessed_data[:medications].any? %>
```
yoomlam, line 12: If `@assessed_data[:medications]` is nil, this will raise an error. I suggest creating a test so that we handle that situation.

yoomlam (changes requested):
Let's create a test to ensure we handle scenarios where `assessed_data[:some_key]` is `nil.

app/services/rapid_ready_for_decision/views/shared/medications.erb
allthesignals: Oof, thank you. Okay!

allthesignals: @yoomlam how about `present?`? It checks blank arrays and works on `nil`.

allthesignals: I'd need to update this template to something like... any preference?

```erb
<% if @assessed_data[:medications].present? %>
  <% @assessed_data[:medications].each do |medication| %>
    <p>
      <b><%= medication['description'] %></b> <br />
      Prescribed on: <%= medication['authoredOn'][0, 10].to_date.strftime('%m/%d/%Y') %> <br />
      <% if medication['dosageInstructions'].present? %>
        Dosage instructions: <%= medication['dosageInstructions'].join('; ') %><br />
      <% end %>
    </p>
  <% end %>
<% else %>
  <p><br/></p>
  <h6>No active medications were found in the last year</h6>
<% end %>
```

allthesignals: Gonna push up a change with a test. Let me look at how this might apply elsewhere in the code...

yoomlam: Looks good!

app/services/rapid_ready_for_decision/views/shared/intro.erb
```diff
@@ -0,0 +1,20 @@
+<%#
+  Note: This templating language is HTML-inspired. It does not always behave like true HTML.
+  To see limitations, check out the documentation: https://github.com/puzzle/prawn-markup.
+%>
+
+<%= @disability_type.capitalize %> Rapid Ready for Decision | Claim for Increase <br />
+<h1>
+  VHA <%= @disability_type.capitalize %> Data Summary<br/>
```
allthesignals, line 8: If for some reason there's a more descriptive term we should use, we can use the constants that yoom setup and lookup a proper label.

yoomlam: That was the purpose of the `label` key in the Constants hashes.

yoomlam (approved):

spec/services/rapid_ready_for_decision/fast_track_pdf_generator_spec.rb
```diff
@@ -81,5 +87,13 @@
         expect(subject).to include('Active Prescriptions')
       end
     end
+
+    context 'when medication data are nil' do
+      let(:parsed_medications_data) { nil }
```
yoomlam, line 92: Does it still work when `parsed_bp_data` is `nil`?

thilton-oddball (approved):

PR #9536, opened 2022-04-01
allow appointment location when no provider

aherzberg:
## Description of change

Creates appointment location object instead of setting location to nil when an appointment was at home and there was no provider.  

## Original issue(s)
https://github.com/department-of-veterans-affairs/va-mobile-app/issues/2898


jperk51 (approved):
LGTM

PR #9537, opened 2022-04-01
RRD: refactor to use new RRD concern

yoomlam:
## Description of change

Moves methods to new `Form526RapidReadyForDecisionConcern`, adds the concern to the `Form526Submission` model, and update callers of those moved methods.

## Original issue(s)
Several class methods in `Form526BaseJob` operate on the `Form526Submission` record. Those methods should be treated as a Rails `Concern` on `Form526Submission`.

Just refactoring and clean up code.


app/models/concerns/form526_rapid_ready_for_decision_concern.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module Form526RapidReadyForDecisionConcern
```
yoomlam, line 3: Methods are moved from `Form526BaseJob`, semantically unchanged.

app/mailers/rrd_completed_mailer.rb
```diff
@@ -5,7 +5,7 @@ def build(submission)
     @submission = submission
     @disability = submission.form.dig('form526', 'form526', 'disabilities')&.first
     @disability_code = @disability['diagnosticCode'] if @disability
-    @rrd_status = RapidReadyForDecision::Form526BaseJob.rrd_status(submission)
+    @rrd_status = submission.rrd_status
```
yoomlam, line 8: If desired, we could get rid of this instance variable and call `submission.rrd_status` directly.

app/workers/rapid_ready_for_decision/form526_base_job.rb
```diff
@@ -14,45 +14,12 @@ class Form526BaseJob
     # https://github.com/mperham/sidekiq/issues/2168#issuecomment-72079636
     sidekiq_options retry: 8
 
-    # @return if this claim submission was processed and fast-tracked by RRD
-    def self.rrd_claim_processed?(submission)
-      submission.form_json.include? RapidReadyForDecision::FastTrackPdfUploadManager::DOCUMENT_TITLE
-    end
-
-    # Fetch all claims from EVSS and return whether there are any open EP 020's.
-    # This method could be moved into a Concern when ProcessorSelector adds new job classes.
```
yoomlam, line 23: This PR is addressing the code comment here.

thilton-oddball (approved):

PR #9544, opened 2022-04-02
Normalize responses with the same example UUIDs and timestamps

cilestin:
## Description of change
Normalizes our rswag documentation examples to use the same UUIDs and timestamps regardless of when we generate the files. Keeps things consistent both for our examples and to keep our documentation diffs easier to manage.


ElleeB:
Did you ever know that you're my hero?

PR #9545, opened 2022-04-02
Add appeal created_at to PDFSubmitJob error reporting

cilestin:
## Description of change
Adds the appeal's `created_at` time to our job error reporting, since it's usually asked for when checking with CMP. Avoids needing to take an extra step or two to figure out when the record was actually created instead of when the job ran & failed.

[An example of the current error, without the created_at timestamp.](https://lighthouseva.slack.com/archives/C0182AHPURM/p1648583664101799)


stiehlrod (approved):
LGTM

PR #9550, opened 2022-04-04
[API-14930] - ClaimsApi - Add title10 'anticipatedSeparationDate' validation

jw81:
## Description of change

- Adds validation for the Form 526 `serviceInformation.reservesNationalGuardService.title10Activation.anticipatedSeparationDate` attribute.

## Original issue(s)
[API-14930](https://vajira.max.gov/browse/API-14930)


hinzed1127 (approved):
LGTM. Feeling like we're getting to the point where we could move all these 526 field validations into their own helper module and `require`-ing it. It'd be a helpful refactor IMO.

PR #9560, opened 2022-04-06
API-14928: Fix type error/error handling for 526 establishment

hinzed1127:
## Description of change

Two small fixes for 526 submissions:
1. `evss_response` is a `messages` array. We were attempting to reference that `messages` key, resulting in a "no implicit conversion of String into Integer" error
2. clearing out the `evss_response` value after successful claim establishment so we don't store outdated errors on the model

## Original issue(s)
https://vajira.max.gov/browse/API-14928


modules/claims_api/app/workers/claims_api/claim_establisher.rb
```diff
@@ -23,6 +23,7 @@ def perform(auto_claim_id) # rubocop:disable Metrics/MethodLength
                             vbms_id: response.claim_id)
       auto_claim.evss_id = response.claim_id
       auto_claim.status = ClaimsApi::AutoEstablishedClaim::ESTABLISHED
+      auto_claim.evss_response = nil
```
hinzed1127, line 26: clear out any previous errors if establishment is successful

modules/claims_api/app/controllers/claims_api/v1/claims_controller.rb
```diff
@@ -51,7 +51,7 @@ def show
 
       def fetch_errored(claim)
         if claim.evss_response&.any?
-          errors = format_evss_errors(claim.evss_response['messages'])
+          errors = format_evss_errors(claim.evss_response)
```
hinzed1127, line 54: `evss_response` is of type array. Trying to access a `messages` key was causing a type error because it's not present on `evss_response`

modules/claims_api/spec/requests/v1/claims_request_spec.rb
```diff
@@ -172,7 +172,7 @@
                  evss_id: 600_118_851,
                  id: 'd5536c5c-0465-4038-a368-1a9d9daf65c9',
                  status: 'errored',
-                 evss_response: { 'messages' => [{ 'key' => 'Error', 'severity' => 'FATAL', 'text' => 'Failed' }] })
+                 evss_response: [{ 'key' => 'Error', 'severity' => 'FATAL', 'text' => 'Failed' }])
```
hinzed1127, line 175: updating this to reflect how we actually set this value

acovrig (approved):

jw81 (approved):

PR #9565, opened 2022-04-06
Enable Puma /stats endpoint for healthchecking

ph-One:
## Description of change

- Enables the stats endpoint for puma.
- There is an issue where if puma threads are full, the instance is regarded as unhealthy by the load balancer and recycled by the autoscaling group.
- Having a health check that does not consume a thread allows us to keep puma alive while it's still performing work, as one would expect.
- This will be used for liveness checking in EKS, specifically.
- This endpoint / port will not be exposed publicly, and only allowed between the load balancer and the service itself.

Locally tested puma with the stats endpoint enabled and was able to hit said endpoint.


thilton-oddball (approved):

config/puma.rb
```diff
@@ -6,6 +6,9 @@
 threads_count = Integer(ENV.fetch('PUMA_THREADS', 16))
 threads(threads_count, threads_count)
 
+# used for a healthcheck endpoint that will not consume one of the threads
+activate_control_app 'tcp://0.0.0.0:9293', { no_token: true }
```
rmtolmach, line 10: @ph-One Port 🤔  I think I understand this from looking at some ruby docs https://www.rubydoc.info/gems/puma/Puma%2FDSL:activate_control_app + Puma docs https://github.com/puma/puma#controlstatus-server

But a question I have is, how/when/where does `activate_control_app` get called?

ph-One: Great question @rmtolmach , and here's my understanding... [others] please correct me (this is my first time touching Ruby in _many_ years and I bet I'm missing a piece here), but my understanding is that they use `foreman` which looks at this Procfile, https://github.com/department-of-veterans-affairs/vets-api/blob/master/Procfile, and within that Procfile you can see the processes that get kicked off, one of which is puma with `-C config/puma.rb`.

rmtolmach: The Procfile runs it... things are coming together now. Thanks!

PR #9566, opened 2022-04-06
Rename mobile appointments param to include pending

kreek:
## Description of change
Renames the appointments 'included' param to 'include' so that we correctly following the [JSON-API spec](https://jsonapi.org/format/#fetching-includes). Temporarily keep 'included' working to not disrupt QA testing. Will remove it in another PR once a new app build is out.

## Original issue(s)
department-of-veterans-affairs/va-mobile-app/2917


aherzberg (approved):
LGTM

PR #9570, opened 2022-04-06
RRD: Fix boolean bug preventing claims from adding special issue

nanotone:
## Description of change
Fixes a bug introduced in #9518 where claims with an RRD PDF attached would not get the corresponding RRD special issue.

Impact: a handful of RRD-processed hypertension claims (less than 10) _may_ not have been routed to the intended RO after being established in VBMS. It's possible the issue was manually corrected downstream -- ongoing investigation in this [Slack thread](https://dsva.slack.com/archives/C031LQ08N0K/p1649266696885619).

* Added a new test for the affected situation, which fails on the current buggy logic.


yoomlam (approved):
Do you need to add a new vcr cassette? If not, use an existing one by calling `VCR.use_cassette`.

app/workers/rapid_ready_for_decision/form526_base_job.rb
```diff
@@ -146,7 +146,7 @@ def upload_pdf(form526_submission, pdf)
     end
 
     def set_special_issue(form526_submission)
-      return if release_pdf?(form526_submission)
+      return unless release_pdf?(form526_submission)
```
yoomlam, line 149: FYI: Unless you do it here or suggest otherwise, I plan to move the condition check outside of this method, i.e., so line 65 will become:
```ruby
set_special_issue(form526_submission) if Flipper.enabled?(:rrd_add_special_issue) && release_pdf?(form526_submission)
```

nanotone:
@yoomlam Yep, there should already be one on this PR, but it might've gotten the auto-hide treatment from Github.

thilton-oddball (approved):

PR #9575, opened 2022-04-07
RRD: rename Constants' hash key to sidekiq_job

yoomlam:
## Description of change

Rename Constants' hash key from `processor_class` to `sidekiq_job`.
This is the first step to separate the RRD processing code from how the code is run (i.e., by a Sidekiq job), which will reduce passing around `form526_submission` to many methods by moving those methods into a class with instance variable `@form526_submission`.

## Original issue(s)


leikkisa (approved):
This looks good to me!

PR #9577, opened 2022-04-07
RRD: Extract RRD processor from Sidekiq job - part 1

yoomlam:
## Description of change

This PR is Part 1 (to keep PR sizes under the limit) of the follow-on step to PR #9575 and #9594 to separate the RRD processing code from how the code is run (i.e., by a Sidekiq job), which will reduce passing around `form526_submission` to many methods by moving those methods into a separate `RrdProcessor` class with instance variable `@form526_submission`.

This PR:
- extracts new `RrdProcessor` class from Sidekiq `Form526BaseJob`
- renames `Form526HypertensionJob` to `HypertensionProcessor` and make it a subclass of the new `RrdProcessor` class
- extract new `AsthmaProcessor` class from `Form526AsthmaJob` and make it a subclass of the new `RrdProcessor` class

## Original issue(s)

Part 2 of this effort in PR #9596 deletes `Form526HypertensionJob`, its spec, and vcr_cassettes.
Part 3 of this effort in PR #9597 moves specs originally in `Form526HypertensionJob` to `form526_base_job_spec.rb` and `hypertension_processor_spec.rb`.


app/services/rapid_ready_for_decision/constants.rb
```diff
@@ -7,7 +7,7 @@ class Constants
         code: 7101,
         label: 'hypertension',
         flipper_name: 'hypertension',
-        sidekiq_job: 'RapidReadyForDecision::Form526HypertensionJob',
+        sidekiq_job: 'RapidReadyForDecision::Form526BaseJob',
```
allthesignals, line 10: So is the idea here that if there is a disability the requires specific handling, it will get its own job class, but if it's generic _enough_ it will just use the base job?

spec/jobs/rapid_ready_for_decision/form526_base_job_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  let!(:user) { FactoryBot.create(:disabilities_compensation_user, icn: '2000163') }
+  let(:submission) { create(:form526_submission, :with_uploads, user: user, submitted_claim_id: '600130094') }
+
+  let(:mocked_observation_data) do
+    [{ effectiveDateTime: "#{Time.zone.today.year}-06-21T02:42:52Z",
+       practitioner: 'DR. THOMAS359 REYNOLDS206 PHD',
+       organization: 'LYONS VA MEDICAL CENTER',
+       systolic: { 'code' => '8480-6', 'display' => 'Systolic BP', 'value' => 115.0, 'unit' => 'mm[Hg]' },
+       diastolic: { 'code' => '8462-4', 'display' => 'Diastolic BP', 'value' => 87.0, 'unit' => 'mm[Hg]' } }]
+  end
```
allthesignals, line 18: Hmm I'm not seeing this used in this particular test. Is it for more tests in downstream PRs?

FWIW: other tests use the Lighthouse class + VCR to make this data available.

allthesignals (approved):

app/services/rapid_ready_for_decision/constants.rb
yoomlam: Pretty much. I would state it as: if the claim/disability requires special Sidekiq configurations that is different from `Form526BaseJob` (such as number of retries), then it can have it's own `sidekiq_job` class.

spec/jobs/rapid_ready_for_decision/form526_base_job_spec.rb
yoomlam: Yes, the tests that use this are restored/added in Part 3 PR #9597: https://github.com/department-of-veterans-affairs/vets-api/pull/9597/files#diff-08179f4e3d33c9ef93c77850c81934f28fc825ed339ea260804a04b10ca31287R48.

This code was copied from `disability_compensation_job_spec.rb:29` -- trying to minimize changes. We can make it more consistent in another PR.

app/workers/rapid_ready_for_decision/form526_base_job.rb
```diff
@@ -22,21 +22,14 @@ def perform(form526_submission_id)
...
-          return if assessed_data.nil?
-
-          add_medical_stats(form526_submission, assessed_data)
-
-          pdf = generate_pdf(form526_submission, assessed_data)
-          upload_pdf(form526_submission, pdf)
-
-          set_special_issue(form526_submission) if Flipper.enabled?(:rrd_add_special_issue)
+          processor = processor_class.new(form526_submission)
+          processor.run
```
leikkisa, line 32: This looks nice :-D

app/workers/rapid_ready_for_decision/form526_asthma_job.rb
```diff
@@ -8,6 +8,7 @@ class Form526AsthmaJob < Form526BaseJob
 
     sidekiq_options retry: 2
 
+    ## Todo later: the following will be removed in a separate PR to keep this PR small
```
leikkisa, line 11: And same for the a bunch of other stuff in here and in the hypertension job too

app/services/rapid_ready_for_decision/constants.rb
leikkisa: Ok, I see the hypertension job getting deleted in Part 2

leikkisa (approved):
This looks good to me, getting used to looking at the three parts. Overall appreciating the refactor.

spec/jobs/rapid_ready_for_decision/asthma_processor_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+  let(:saved_claim) { FactoryBot.create(:va526ez) }
+  let(:submission) do
+    create(:form526_submission, :with_uploads, :asthma_claim_for_increase,
+           user_uuid: user.uuid,
+           auth_headers_json: auth_headers.to_json,
+           saved_claim_id: saved_claim.id,
+           submitted_claim_id: '600130094')
+  end
+
+  let(:mocked_observation_data) do
```
leikkisa, line 24: Should this be medication data?

leikkisa (commented):
I think this looks good once the asthma spec is updated.

spec/jobs/rapid_ready_for_decision/asthma_processor_spec.rb
yoomlam: Updated.

leikkisa (approved):
👍

RachalCassity (approved):

PR #9580, opened 2022-04-07
RRD: Allow asthma claims with active medications to proceed to PDF generation

nanotone:
## Description of change
This PR adds logic to the RRD asthma flow to evaluate MedicationRequest data from the Lighthouse Veterans Health API, and, if there are any active medications, allow the claim to save a summary to `rrd_metadata` and proceed to the PDF generation step.

The actual PDF generation is still to come, but this PR sets the stage for it by adding a stub implementation (essentially just the generic header) and saving the PDF's upload GUID to `rrd_metadata`. This allows the team to look in S3 for the PDF, and prior work (see below) ensures that the PDF doesn't get released.

## Original issue(s)
- [Notion ticket](https://www.notion.so/Dashboard-RRD-Health-Evidence-24d77d9988a546c3b19a3c73e7ced29b?p=34eb5264471f48999a36c4f3063e21b7)
- Prior work:
  - #9518 
  - #9570


app/workers/rapid_ready_for_decision/form526_asthma_job.rb
```diff
@@ -9,27 +9,11 @@ class Form526AsthmaJob < Form526BaseJob
...
-      ActionMailer::Base.mail(
-        from: ApplicationMailer.default[:from],
-        to: Settings.rrd.event_tracking.recipients,
-        subject: 'RRD claim - Offramped',
-        body: body
-      ).deliver_now
-
-      # returning nil will short-circuit the PDF generation step
-      nil
+      assessed_data = query_and_assess_lighthouse(form526_submission)
```
nanotone, line 12: All the query/assess code in this class are heavily modeled after the analogous code in the hypertension job. It's so similar that some day soon we'll probably want to refactor it into the base class.

app/workers/rapid_ready_for_decision/form526_asthma_job.rb
```diff
@@ -38,22 +22,26 @@ def release_pdf?(_form526_submission)
...
     end
 
-    def meds_by_year_and_status_as_string(med_request_data)
-      medication_request_stats_by_year(med_request_data).map do |year, tallies|
-        "- #{year}: #{tallies}"
-      end.join("<br/>\n")
-    rescue
-      'Error formatting medications list!'
+    def med_stats_hash(_form526_submission, assessed_data)
+      { medications_count: assessed_data[:medications]&.size }
```
nanotone, line 38: Like number of blood pressure readings, this defines the summary data that we may want to see in the post-EVSS notification. Open to debate on what data we actually want to see.

app/workers/rapid_ready_for_decision/form526_base_job.rb
```diff
@@ -106,6 +106,10 @@ def account(form526_submission)
       Account.find_by(edipi: edipi) if edipi
     end
 
+    def patient_info(form526_submission)
+      form526_submission.full_name.merge(birthdate: form526_submission.auth_headers['va_eauth_birthdate'])
+    end
```
nanotone, line 111: Pulled this over unchanged from the hypertension job, since it's equally relevant to asthma.

spec/jobs/rapid_ready_for_decision/form526_asthma_job_spec.rb
```diff
@@ -11,14 +11,7 @@
...
   let(:user) { FactoryBot.create(:disabilities_compensation_user, icn: '2000163') }
-  let(:auth_headers) do
-    EVSS::DisabilityCompensationAuthHeaders.new(user).add_headers(EVSS::AuthHeaders.new(user).to_h)
-  end
-  let(:submission) do
-    create(:form526_submission, :with_uploads,
-           user_uuid: user.uuid,
-           auth_headers_json: auth_headers.to_json)
-  end
+  let(:submission) { create(:form526_submission, :with_uploads, user: user) }
```
nanotone, line 14: Unrelated change: this test setup simplification was suggested by @yoomlam a few weeks ago. Essentially, passing `user: user` to the factory takes care of all the rest.

app/workers/rapid_ready_for_decision/form526_base_job.rb
allthesignals: This may be something that could just go into the new RRD concern for the model but makes sense here

spec/jobs/rapid_ready_for_decision/form526_asthma_job_spec.rb
allthesignals: Nice! look at all those deleted lines

allthesignals (dismissed):
Awesome. I had a comment about the patient_info method but don't think it should block.

app/workers/rapid_ready_for_decision/form526_asthma_job.rb
yoomlam: To get a more precise count, we could store medications by year, like https://github.com/department-of-veterans-affairs/vets-api/pull/9580/files#diff-10b3b2429bd9e2e544c255257886ebbea6e0bb9ae0c16f6bf63aaf9499c74bfdL53

app/workers/rapid_ready_for_decision/form526_base_job.rb
yoomlam: Acknowledged. It can go into `RrdProcessor` https://github.com/department-of-veterans-affairs/vets-api/pull/9577/files#diff-10ef8ac3ff61abaac78e7986500c46a625c44c551a2eeee043939c72ad2408a0R4.
I still need to split the PR further into smaller PRs.

spec/jobs/rapid_ready_for_decision/form526_asthma_job_spec.rb
yoomlam: I wonder if you even need `:with_uploads`.

spec/jobs/rapid_ready_for_decision/form526_asthma_job_spec.rb
```diff
@@ -31,12 +24,6 @@
       it 'finishes successfully' do
         Sidekiq::Testing.inline! do
           expect { subject }.not_to raise_error
-          last_email = ActionMailer::Base.deliveries.last
-          expect(last_email.subject).to eq 'RRD claim - Offramped'
-          expect(last_email.body).to include submission.id
-          expect(last_email.body).to include 'API returned 24 medication requests'
-          expect(last_email.body).to include '2013: {"active"=>3}'
-          expect(last_email.body).to include '1998: {"active"=>4, "completed"=>2}'
         end
```
yoomlam, line 27: Suggest checking for artifacts that the asthma processing was successful, like https://github.com/department-of-veterans-affairs/vets-api/blob/9ed57531c0960c8840f4c52df82023c7c0322add/spec/jobs/rapid_ready_for_decision/form526_hypertension_job_spec.rb#L76

spec/jobs/rapid_ready_for_decision/form526_asthma_job_spec.rb
```diff
@@ -54,4 +41,14 @@
       end
     end
   end
+
+  describe '#assess_data', :vcr do
+    subject { RapidReadyForDecision::Form526AsthmaJob.new.assess_data(submission) }
```
yoomlam, line 46: This will facilitate class renaming (which is part of my PR):
```suggestion
    subject { described_class.new.assess_data(submission) }
```

yoomlam (approved):

leikkisa (dismissed):
All looks like good stuff to me!

app/workers/rapid_ready_for_decision/form526_asthma_job.rb
allthesignals: I'm realizing I'll need to flag certain medications for being "asthma-like" based on provided keywords (that we could store in constants). I wonder if that munging exercise should go in here somewhere.

I have an alternative way to do this in the pdf generator but seems off to me...

allthesignals (dismissed):

allthesignals (approved):

thilton-oddball (approved):

PR #9591, opened 2022-04-08
API-13926 Fix header text overlap

ElleeB:
## Description of change
The v2 version of the Notice of Disagreement form moved up some of the boiler plate text at the top of the pages. Some of our stamped text was overlapping the boiler plate.

A temporary fix had been previously implemented, but this PR revamps the stamping process so that ALL appeal top-of-page stamps move up.

A new Stamper service has been added to handle the stamps across all appeal types. Higher Level Review v1, however, stamps differently than the others, so it got special treatment.

To support the new design, #stamp_text has been moved from each individual form_data to each of the three appeal models - one benefit of multiple benefits is that the method occurs only three times, instead of 5 times currently, and then would need to be repeated every time we have a new form version.

## Original issue(s)
[Ticket](https://vajira.max.gov/browse/API-13926)

ALL pdf fixtures have been updated with the newly increased y coordinate stamps.
There are some temporary fixes that will be removed once HLR v1 is sunset.

## Things to consider
The size of the text is determined in the CentralMail::DateStampPdf class. We can control this if we want, but further changes don't seem necessary at this time.


modules/appeals_api/app/services/appeals_api/pdf_construction/higher_level_review/v1/structure.rb
```diff
@@ -90,7 +90,14 @@ def form_title
           '200996'
         end
 
-        def stamp(stamped_pdf_path)
+        def stamp(unstamped_path)
+          stamped_pdf_path = CentralMail::DatestampPdf.new(unstamped_path).run(
```
ElleeB, line 94: This stamp was previously handled in generator#stamp, which has been removed. Because HLR v1 stamping is handled differently than all other appeals, this portion of the stamping needed to be moved to this method, since it now handles all the stamping for HLR v1.

modules/appeals_api/spec/models/higher_level_review_spec.rb
```diff
@@ -73,6 +73,10 @@
     end
   end
 
+  describe '#stamp_text' do
```
ElleeB, line 76: One test for v1 and one for v2

modules/appeals_api/spec/models/notice_of_disagreement_spec.rb
```diff
@@ -149,6 +149,10 @@
     it { expect(notice_of_disagreement.board_review_option).to eq 'hearing' }
   end
 
+  describe '#stamp_text' do
```
ElleeB, line 152: One test for v1 and one for v2

stiehlrod (approved):
LGTM

PR #9592, opened 2022-04-08
Fix redis deprecations

rileyanderson:
## Description of change
Since Redis 4.6.0, using Redis.current to connect to Redis was deprecated. [Gem maintainer byroot recommends](https://github.com/redis/redis-rb/commit/9745e22db65ac294be51ed393b584c0f8b72ae98#commitcomment-65919666) using $redis or a connection pool.

While he suggests using a connection pool to prevent thread locking, that is out of scope of this PR as our existing solution does not use thread pooling and the intention of this PR is just to remove the thousands of lines of deprecation notices that is making Vets API work difficult.

Also updates `rack-attack` and `redis-namespace` gems to use the current master ref to silence these warnings until new releases are cut.


jperk51 (approved):
LGTM

thilton-oddball (approved):

PR #9593, opened 2022-04-08
API-14708: Fixes infinitely growing events issue by removing expired events from the event group

SlakrHakr:
## Description of change
Resolves issue regarding growing items in Redis.

Ticket removing the scheduled job when the issue was discovered:
https://github.com/department-of-veterans-affairs/vets-api/pull/9437

## Original issue(s)
https://vajira.max.gov/browse/API-14708


modules/apps_api/lib/apps_api/notification_service.rb
```diff
@@ -80,6 +80,8 @@ def event_already_handled?(parsed_hash)
       current_event = { 'email' => parsed_hash[:user_email], 'time' => parsed_hash[:options][:time] }
       members.each do |member|
         member_hash = Redis.current.hgetall(member)
+        Redis.current.srem('apps_notification_events', member) if member_hash.blank?
```
SlakrHakr, line 83: Looking at the 28 thousand odd items in the Redis group in production, all but 1 just had an empty hash.
It looks like an item isnt removed from a group when it expires, and instead it's content is just removed.

This change just removes any empty (expired) items from the "apps_notification_events" group

thilton-oddball (approved):

PR #9596, opened 2022-04-09
RRD: Extract RRD processor from Sidekiq job - part 2

yoomlam:
## Description of change

This PR is Part 2 (to keep PR sizes under the limit; Part 1 is #9577). This effort is a follow-on to PR #9575 and #9594 to separate the RRD processing code from how the code is run (i.e., by a Sidekiq job), which will reduce passing around `form526_submission` to many methods by moving those methods into a separate `RrdProcessor` class with instance variable `@form526_submission`.

This PR deletes `Form526HypertensionJob`, its spec, and vcr_cassettes.

## Original issue(s)
Within RRD's Sidekiq job, the code passes around `form526_submission` to many methods. Discussed in https://github.com/department-of-veterans-affairs/vets-api/pull/9518#discussion_r840024033 

## Things to know about this PR

Refactoring -- no change in behavior.

PR #9594, opened 2022-04-08
RRD: add RrdProcessor classes but don't enable them yet

yoomlam:
## Description of change

This PR is the follow-on step to PR #9575 to separate the RRD processing code from how the code is run (i.e., by a Sidekiq job), which will reduce passing around `form526_submission` to many methods by moving those methods into a separate `RrdProcessor` class with instance variable `@form526_submission`.

This PR add classes but doesn't yet use them:
- renames `ProcessorSelector` to `SidekiqJobSelector`
- adds a new `RrdProcessor` class, extracted from Sidekiq `Form526BaseJob`
- adds a new`HypertensionProcessor` based on `Form526HypertensionJob` and make it a subclass of the new `RrdProcessor` class
- adds a new `AsthmaProcessor` based on `Form526AsthmaJob` and make it a subclass of the new `RrdProcessor` class

## Original issue(s)
Within RRD's Sidekiq job, the code passes around `form526_submission` to many methods. Discussed in https://github.com/department-of-veterans-affairs/vets-api/pull/9518#discussion_r840024033 

## Things to know about this PR

Refactoring -- no change in behavior.

I acknowledge the changed LoC overage:
> This PR changes 290 LoC (not counting whitespace/newlines).

Due to the nature of the change, it's hard to stay under the 200 changed LoC limit. This PR will help make subsequent PRs (#9577, #9596, #9597) smaller.

PR #9597, opened 2022-04-09
RRD: Extract RRD processor from Sidekiq job - part 3

yoomlam:
## Description of change

This PR is Part 3 (to keep PR sizes under the limit; Part 1 is #9577 and Part 2 is #9596). This effort is a follow-on to PR #9575 and #9594 to separate the RRD processing code from how the code is run (i.e., by a Sidekiq job), which will reduce passing around `form526_submission` to many methods by moving those methods into a separate `RrdProcessor` class with instance variable `@form526_submission`.

This PR moves specs from `Form526HypertensionJob` (deleted in the Part 2 PR #9596) to `form526_base_job_spec.rb` and `hypertension_processor_spec.rb`, as well as removes unused code.

## Original issue(s)


app/workers/rapid_ready_for_decision/form526_asthma_job.rb
```diff
@@ -7,54 +7,5 @@ class Form526AsthmaJob < Form526BaseJob
     STATSD_KEY_PREFIX = 'worker.fast_track.form526_asthma_job'
 
     sidekiq_options retry: 2
-
```
leikkisa, line 10: Seems like this would be ready to switch over too (remove this and update in constants). Are we keeping this file for the different retry count? Or is it to break it up more (e.g. with tests)?

leikkisa (approved):
I have a question/comment, but non-blocking (could be a future part).

app/workers/rapid_ready_for_decision/form526_asthma_job.rb
yoomlam: We're keeping it for the retry count. Once the asthma-related code is stable, we can update `Constants` to use `Form526BaseJob` and delete `Form526AsthmaJob`.

leikkisa (dismissed):
Note: LoC changed will go down after Part 2 merged.

ericboehs (approved):

PR #9598, opened 2022-04-11
2863 mobile request new appointment endpoint

jperk51:
## Description of change
Creating a new endpoint for mobile that will allow users to request new appointments

## Original issue(s)
department-of-veterans-affairs/va-mobile-app#2863

Moved quite a bit of logic that would have needed to be duplicated from VAOS v2 to a lib so both of us can use it easily. 

Side note: Many of the files changed are schemas for the docs


jperk51:
Blocked waiting for VAOS team to release stable v2 before we move things into a lib

PR #9600, opened 2022-04-11
[39796] Unified Sign In `application` parameter validation

bramleyjl:
## Description of change

This PR removes the `skip_dupe` parameter that is saved into the SAML tracker and evaluated during the multiple ids validation with an `application` parameter that is used for the same purpose. If a unified sign in is made with an `mhv` or `myvahealth` application the validations that were downgraded to warnings for `skip_dupe` logins will behave in the same manner.

## Original issue(s)
department-of-veterans-affairs/va.gov-team#39796

Unit specs updated, manual logins with mocked data performed to test validations.


ianhundere:
from my minimal ruby on rails experience, this look good to me.

PR #9601, opened 2022-04-11
Flag asthma meds

allthesignals:
## Description of change
Part of main asthma release: highlighting meds that are potential asthma-related:

image

## Original issue(s)
https://www.notion.so/Generate-asthma-PDF-10f4de575ee3463fa08265761b45bd9a


Gemfile
```diff
@@ -109,7 +109,7 @@ gem 'pg_query'
 gem 'pg_search'
 gem 'pkce_challenge'
 gem 'prawn'
-gem 'prawn-markup'
+gem 'prawn-markup', github: 'allthesignals/prawn-markup', ref: 'fb17fa7fa72b8e5fe4a74d7da4e79af2e3216393'
```
allthesignals, line 112: Temporary until/if PR is merged https://github.com/puzzle/prawn-markup/pull/35

app/services/rapid_ready_for_decision/constants.rb
```diff
@@ -14,7 +14,33 @@ class Constants
...
+          Omalizumab
+          Prednisone
+          Reslizumab
+          Salmeterol
+          Theophylline
+          Zafirlukast
+          Zileuton
+          Asthma
+          Breath
+          Inhale
```
yoomlam, line 40: I believe this was intentionally shortened to capture other word forms like "inhalation"
```suggestion
          Inhal
```

app/services/rapid_ready_for_decision/fast_track_pdf_generator.rb
```diff
@@ -24,13 +25,27 @@ def self.extract_patient_name(patient_info)
       [full_name, patient_info[:suffix]].reject(&:blank?).join ', '
     end
 
+    # TODO: move this into the data class
+    def self.flag_with_keyword(keywords, medication_struct)
+      keywords.any? { |keyword| medication_struct.to_s.include?(keyword) }
+    end
```
yoomlam, line 31: You may have to split the keywords into 2 sets: medications and dosage instructions.
Also when you search for the keyword, ignore capitalizations.

allthesignals: > When you search for the keyword, ignore capitalizations: `medication_struct.to_s.downcase` and downcase the `keywords` in `Constants`.
> 
> ~You may have to split the keywords into 2 sets: medications and dosage instructions.~ Nevermind, I think this is acceptable -- we'll have to loop back with Researchers to confirm.

Yeah, I thought about that, and I think ultimately you're right, it needs to allow for that kind of specificity. Your point is another argument for having this behavior moved into the `LighthouseMedicationRequestData` class, where it could add a `flagged` key to the `medication_struct` if it meets certain criteria... still, will get this wrapped up asap...

app/services/rapid_ready_for_decision/fast_track_pdf_generator.rb
```diff
@@ -24,13 +25,27 @@ def self.extract_patient_name(patient_info)
       [full_name, patient_info[:suffix]].reject(&:blank?).join ', '
     end
 
+    # TODO: move this into the data class
+    def self.flag_with_keyword(keywords, medication_struct)
+      keywords.any? { |keyword| medication_struct.to_s.downcase.include?(keyword.downcase) }
```
yoomlam, line 30: Instead of `keyword.downcase`, I was thinking you can downcase the `keywords` in the `Constants` class.

app/services/rapid_ready_for_decision/constants.rb
```diff
@@ -14,7 +14,33 @@ class Constants
         code: 6602,
         label: 'asthma',
         flipper_name: 'asthma',
-        sidekiq_job: 'RapidReadyForDecision::Form526AsthmaJob'
+        sidekiq_job: 'RapidReadyForDecision::Form526AsthmaJob',
+        keywords: %w[
```
yoomlam, line 18: For future reference: https://www.notion.so/Asthma-meds-filter-for-PDF-da5f27c343f24e3daf5e08c05c2d9639

app/services/rapid_ready_for_decision/fast_track_pdf_generator.rb
yoomlam: You could build your own medication results struct that has the desired flag. You can start with something like this: https://github.com/department-of-veterans-affairs/vets-api/blob/90e350090d5afb98604e370a6d0cc732c6ab2e91/app/workers/rapid_ready_for_decision/form526_hypertension_job.rb#L25-L30

and then use it during PDF generation.
I would keep the code in the asthma processor/job since the meds are specific to asthma, and keep `LighthouseMedicationRequestData` general.

allthesignals: @yoomlam yes! even better! thanks

PR #9605, opened 2022-04-12
[API-15062] - ClaimsAPI - update 'servicebranch' mappings

jw81:
## Description of change

- EVSS now uses the new Lighthouse Benefits Reference Data service ([REF](https://github.com/department-of-veterans-affairs/lighthouse-benefits-reference-data/tree/master/benefits-reference-data/src/main/java/gov/va/api/lighthouse/benefitsreferencedata/servicebranches)) for validating the `serviceBranch` values they receive during a 526 submission.
- This PR updates our existing `serviceBranch` mapping logic to account for the new values.

## Original issue(s)
[API-15062](https://vajira.max.gov/browse/API-15062)


modules/claims_api/app/models/claims_api/auto_established_claim.rb
```diff
@@ -370,6 +370,9 @@ def transform_service_branch
 
       transformed_service_periods = received_service_periods.map do |period|
         name = period['serviceBranch']
+
```
jw81, line 373: Kayla asked that we provide a simple log that'll show what `serviceBranch` values we're seeing.  Then maybe we can get some insight into what values consumers are using more often/not at all.

modules/claims_api/spec/models/auto_establish_claim_spec.rb
```diff
@@ -307,200 +307,38 @@
       end
```
jw81, line 307: This file looks worse than it actually is.
All I've done in this file is remove all of the old specs that checked the old `serviceBranch` mappings.
All of the tests are now covered in the new `service_branch_mapper_spec.rb` class. (see above)

hinzed1127 (approved):

PR #9612, opened 2022-04-12
2792 update appointments to use mfs for facilities lookup

cadibemma:
## Description of change
There's been an ongoing bug where some appointment facilities can not be found. Through talking with the VAOS team and our own investigation we've determined [VAOS's MFS (Mobile Facility Service)](https://github.com/department-of-veterans-affairs/vets-api/blob/53bddd489890d28a7a3510064c7a74da5f979f90/modules/vaos/app/services/vaos/v2/mobile_facility_service.rb) finds some facilities than the current service that we use (Lighthouse).

## Original issue(s)
department-of-veterans-affairs/va-mobile-app#2792


modules/mobile/app/helpers/mobile/facilities_helper.rb
```diff
@@ -18,13 +18,26 @@ def facilities_service
     end
 
     def address_from_facility(facility)
-      address = facility.address['physical']
+      if facility.type == 'va_health_facility' # for MFS Facilities
```
kpethtel, line 21: So it seems like this PR really does two things: 
1) fixes the handling of VHAFacility because it isn't formatting correctly. Is it possible to fix this in the model? I looked briefly at this model and it's not clear to me why this one is different than the other sub-classes. IMO this is a bug in how our sub-classes are being handled. 
2) changing how we fetch facility data based on a feature flag. Is there a reason we're feature flagging this and not just fixing the bug?

modules/mobile/app/services/mobile/v0/appointments/proxy.rb
```diff
@@ -202,6 +224,18 @@ def fetch_appointment_requests
           }
         end
 
+        def ids(ids)
+          ids.is_a?(Array) ? ids.to_csv(row_sep: nil) : ids
+        end
+
+        def children
```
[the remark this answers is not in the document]
cadibemma: Ok I see.. The methods are removed

modules/mobile/app/services/mobile/v0/appointments/proxy.rb
```diff
@@ -202,6 +224,18 @@ def fetch_appointment_requests
           }
         end
 
+        def ids(ids)
```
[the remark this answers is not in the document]
cadibemma: Yup it was a convenient way of converting them to "id, id, id". I change name as suggested.

modules/mobile/app/services/mobile/v0/appointments/proxy.rb
```diff
@@ -131,7 +139,21 @@ def fetch_appointments(start_date, end_date)
           { va: va_response, cc: cc_response, requests: requests_response }
         end
 
-        def fetch_facilities(appointments)
+        def new_fetch_facilities(appointments)
+          facility_ids = appointments.map(&:id_for_address).uniq
+
+          ids = ids(facility_ids) # may not be needed
```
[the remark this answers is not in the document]
cadibemma: Since ids are formatted in Vets API facility controller as a csv first before using MFS's get_facilities, I had to check if it had to be set up this way as well within mobile. I forgot to remove the comment..

modules/mobile/app/helpers/mobile/facilities_helper.rb
cadibemma: It's a feature flag because we are trying to retrieve the facilities from using VAOS's Mobile Facility Service. The facilities are normally retrieved from Lighthouse, but there are some facilities not found in Lighthouse that can be found in MFS.

kpethtel: I looked more closely at the story and see now that the plan is to be able to switch back if the new approach causes more problems. Makes sense. Any idea about why that one facility sub-class is different than the others? I don't see us doing this same type checking elsewhere in the app, so I'm curious if it's truly necessary. Were you encountering issues when using strings instead of symbols?

modules/mobile/app/models/mobile/v0/adapters/va_facilities.rb
```diff
@@ -31,6 +31,15 @@ def map_appointments_to_facilities(appointments, facilities)
 
         private
 
+        def map_facilities_id(facilities)
+          facilities&.map do |facility|
+            next if facility.id.nil?
```
kpethtel, line 36: FYI, this will leave nils in the resulting array. You may want to compact the result.

modules/mobile/app/services/mobile/v0/appointments/proxy.rb
```diff
@@ -202,6 +224,10 @@ def fetch_appointment_requests
           }
         end
 
+        def format_facility_ids(ids)
```
kpethtel, line 227: Given that the inputs here come from appointments.map, it must be an array. I don't think you need handling for other types.

modules/mobile/app/services/mobile/v0/appointments/proxy.rb
```diff
@@ -131,7 +139,21 @@ def fetch_appointments(start_date, end_date)
           { va: va_response, cc: cc_response, requests: requests_response }
         end
 
-        def fetch_facilities(appointments)
+        def new_fetch_facilities(appointments)
+          facility_ids = appointments.map(&:id_for_address).uniq
+
+          ids = format_facility_ids(facility_ids)
+
+          return nil unless facility_ids.any?
```
kpethtel, line 147: You could move this guard to above where we convert the ids to a string.

modules/mobile/app/services/mobile/v0/appointments/proxy.rb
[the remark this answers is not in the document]
kpethtel: I looked in that controller and see what you mean. I have no idea why they did that. The typical way would be `ids.join(', ')`. They're working from user input params, so perhaps there's some strange case they're concerned about. Given that we know our inputs are an array of facility ids, I'd suggest we just change to the join method because it's more obvious and no one will have to look up the `row_sep` option to understand what's happening.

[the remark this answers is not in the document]
cadibemma: updated!

modules/mobile/app/helpers/mobile/facilities_helper.rb
cadibemma: Lighthouse and vaos sometimes used different key names for the values the method is looking for. So I thought the type check was necessary. And yes I was encountering issues when using strings instead of symbols

modules/mobile/app/models/mobile/v0/adapters/va_facilities.rb
cadibemma: compacting the result caused errors. it looks like the home appts won't have facilities ids (so the nils are needed)

modules/mobile/app/services/mobile/v0/appointments/proxy.rb
cadibemma: got it. I changed it to join as a string

modules/mobile/spec/support/vcr_cassettes/appointments/get_mfs_facilities.yml
```diff
@@ -0,0 +1,262 @@
+---
+http_interactions:
```
kpethtel, line 2: Is there a reason this vcr contains 3 requests to 3 different endpoints? In general it seems like we only do one request per cassette, which makes sense to me because it ensures that you know which request is being used. I could see maybe combining requests into a single file if they're closely coupled (i.e., MFS needs to make multiple calls that are always the same) but I don't think that's what's happening here because two of the calls return 403, which doesn't sound like a success case.

kpethtel (commented):
Aside from my questions about the specs, I'd also suggest adding some testing around error cases just to be safe since we've had a lot of bugs from upstream errors. We should ensure that upstream errors are handled appropriately.

modules/mobile/spec/workers/pre_cache_appointments_job_spec.rb
```diff
@@ -31,48 +31,50 @@
     context 'with no errors' do
       it 'caches the expected appointments' do
         VCR.use_cassette('appointments/get_facilities', match_requests_on: %i[method uri]) do
-          VCR.use_cassette('appointments/get_cc_appointments_default', match_requests_on: %i[method uri]) do
-            VCR.use_cassette('appointments/get_appointments_default', match_requests_on: %i[method uri]) do
-              expect(Mobile::V0::Appointment.get_cached(user)).to be_nil
-              subject.perform(user.uuid)
+          VCR.use_cassette('appointments/get_mfs_facilities', match_requests_on: %i[method uri]) do
```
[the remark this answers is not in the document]
cadibemma: Its surprisingly using it... the test was failing without the new VCR because it was also getting the new url and didn't know how to handle it.

modules/mobile/spec/support/vcr_cassettes/appointments/get_mfs_facilities.yml
cadibemma: It doesn't work without the other 2 endpoints.. (Maybe its to receive appointment data since the mfs endpoint is just facilities?). But I believe the 403 return was due to an issue found with the format of the facility id during the time of recording (when has been fixed)

modules/mobile/app/services/mobile/v0/appointments/proxy.rb
cadibemma: nope it doesn't so I removed it

modules/mobile/app/helpers/mobile/facilities_helper.rb
RachalCassity: There is an option to symbolize the keys. 

    def address_from_facility(facility)
      if facility.type == 'va_health_facility' # for MFS Facilities
        address = facility.physical_address
        street = address[:line].compact.join(', ')
        zip_code = address[:postal_code]
      else
        address = facility.address['physical']
        street = address.slice('address_1', 'address_2', 'address_3').values.compact.join(', ')
        zip_code = address['zip']
      end
      Mobile::V0::Address.new(
        street: street,
        city: address.symbolize_keys[:city],
        state: address.symbolize_keys[:state],
        zip_code: zip_code
      )
    end

RachalCassity: I feel as if there is a better fix to this. There are two places where this method is called. To avoid confusion in the future, it would be best for both calls to send the same params and the hash keys can be symbolized.

modules/mobile/app/helpers/mobile/facilities_helper.rb
```diff
@@ -39,12 +52,16 @@ def blank_location(appointment)
     end
 
     def phone_from_facility(facility)
-      phone = facility.phone['main']
+      phone = if facility.type == 'va_health_facility'
```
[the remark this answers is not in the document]
kpethtel: 👍

modules/mobile/spec/support/vcr_cassettes/appointments/get_mfs_facilities.yml
kpethtel: Interesting. Thanks for explaining. It's definitely a little atypical and I'd have to do some exploring to understand how those 403s aren't problematic, but I guess they're not.

modules/mobile/spec/workers/pre_cache_appointments_job_spec.rb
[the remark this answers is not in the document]
kpethtel: Any idea why it's being used without the feature flag being enabled? Seems like it shouldn't. Also, is the `'appointments/get_facilities'` VCR above still necessary? I don't understand why both would be called. Maybe I'm missing a piece and it's being called for reasons other than the appointments proxy.

kpethtel (approved):

kpethtel (changes requested):
Sorry, I didn't meant to click approve. I would like to understand that question about the spec feature flag before approving.

modules/mobile/spec/workers/pre_cache_appointments_job_spec.rb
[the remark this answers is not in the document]
kpethtel: Ok, I get it now. The issue is that all feature flags are on by default in specs. You disable them in the list specs but not here. IMO what you should do is have two versions of this test in this file, one with the feature flag on and the other with the feature flag off. This test as it's written does not actually use this vcr: 
`VCR.use_cassette('appointments/get_facilities', match_requests_on: %i[method uri]) do`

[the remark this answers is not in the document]
kpethtel: I would also suggest removing this from line 110:
`VCR.use_cassette('appointments/get_facilities', match_requests_on: %i[method uri]) do`
It doesn't use that cassette because the error happens before it fetches facilities. Now that we're using multiple facilities endpoints, it just adds confusion and will probably just become cruft once we permanently switch to the new approach.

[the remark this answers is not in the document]
cadibemma: Oh ok I see. So now theres one that tests using `'get_facilities'` and another that tests using `'get_mfs_facilities'`

modules/mobile/app/helpers/mobile/facilities_helper.rb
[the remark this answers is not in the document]
cadibemma: Much cleaner. thank you!

kpethtel (approved):
👍

RachalCassity:
I do not think it is wise to hard code a process into a method. In the future, if another place called "address_from_facility" and used  'va_health_facility' formatting, the method would fail and there would have to be another if cause. The params being sent to this method should be the same.

cadibemma:
> I do not think it is wise to hard code a process into a method. In the future, if another place called "address_from_facility" and used 'va_health_facility' formatting, the method would fail and there would have to be another if cause. The params being sent to this method should be the same.

Is there a suggestion on how I can make the params the same before going to the "address_from_facility" method?

RachalCassity:
> 
Let me look at the code. That can be a new PR just so this isn't broken still.

RachalCassity (approved):

PR #9614, opened 2022-04-12
DHP-93 confirm fitbit registration

ncbui:
## Description of change
This adds a controller and services to process requests sent to '/fitbit-callback' to connect fitbit devices.
Calls to the endpoint should redirect the client to the appropriate fitbit endpoint, then redirect back to '/health-care/connected-devices/?' with the status of the connection added as a url parameter. 

## Original issue(s)
Collaboration Cycle Kickoff: https://github.com/department-of-veterans-affairs/va.gov-team/issues/34264
IA Request from Digital Health Platform team: https://github.com/department-of-veterans-affairs/va.gov-team/issues/35290

* Are there additions to a `settings.yml` file? Do they vary by environment? No
* Is there a feature flag? What is it? Yes, feature is currently disabled in prod
* Is there some Sentry logging that was added? What alerts are relevant? No / NA
* Are there any Prometheus metrics being collected? What Grafana dashboard were they added do? No / NA
* Are there Swagger docs that were updated? Yes
* Is there any PII concerns or questions? No
* Added unit tests for services and functional tests for the controller that cover when users are not authenticated and when feature is toggled.


modules/dhp_connected_devices/app/services/fitbit_service.rb
```diff
@@ -0,0 +1,17 @@
...
+class FitbitService
+  def get_connection_status(data)
+    @fitbit_client = data[:fitbit_api]
+    data[:callback_params][:code] && exchange_code_for_token(data[:callback_params][:code]) ? 'success' : 'error'
+  end
+
+  private
+
+  def exchange_code_for_token(code)
+    response = @fitbit_client.get_token(code)
```
yonidoronpeters, line 14: This seems like an issue. We call `@fitbit_client` in this method because we assume that it was initialized in the previous method. This introduces a dependency between those 2 methods. So we must call `get_connection_status` before we call `exchange_code_for_token`. Otherwise it will error because `@fitbit_client` is not initialized.

modules/dhp_connected_devices/app/services/website_host_service.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class WebsiteHostService
```
yonidoronpeters, line 3: Awesome!

modules/dhp_connected_devices/app/services/website_host_service.rb
```diff
@@ -0,0 +1,23 @@
...
+  def get_redirect_url(data)
+    vendor = data.fetch(:vendor)
+    status = data.fetch(:status)
+    "#{WEBSITE_HOSTS[env]}/health-care/connected-devices/?#{vendor}=#{status}#_=_"
```
yonidoronpeters, line 7: I like the fact that you extracted the vendor here too :)

yonidoronpeters (commented):
Looks great - just one issue with the fitbit service potentially calling the client when it is not initialized

modules/dhp_connected_devices/app/services/fitbit_service.rb
```diff
@@ -4,14 +4,14 @@
...
   end
 
   private
 
-  def exchange_code_for_token(code)
-    response = @fitbit_client.get_token(code)
+  def exchange_code_for_token(data)
+    response = data[:client].get_token(data[:code])
     response.status == 200
   end
```
TravisGM, line 16: @yonidoronpeters what do you think of this change?

yonidoronpeters: Line 8 is pretty complex, and there's inconsistency between the fitbit client naming which is a bit confusing. It will work, but not super clean. 
Ideally, we would move the fitbit client initialization from the controller into this service. And then delegate to the service when we need to call fitbit with the client. But that will result in some changes to tests. What do you think?

yonidoronpeters (dismissed):

modules/dhp_connected_devices/app/services/fitbit_service.rb
TravisGM: @yonidoronpeters good point. We'll do that.

modules/dhp_connected_devices/app/services/fitbit_service.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'fitbit/client'
+
+class FitbitService
+  def self.fitbit_client
+    # code here
+  end
```
yonidoronpeters, line 8: Can we clean this up if it is not used?

yonidoronpeters (approved):

PR #9620, opened 2022-04-13
Allthesignals/asthma meds

allthesignals:
## Description of change
Moves asthma medication flagging logic into the asthma job class (out of the pdf generation class).

## Original issue(s)
https://www.notion.so/Generate-asthma-PDF-10f4de575ee3463fa08265761b45bd9a 

image

The "flagged" key is represented in the hash as a symbol. The mocked data in the pdf generator class needs to know this nuance in order to accurately capture the behavior. This shows a gap in coverage and if there's a way to do end-to-end verification with the job class, it'd be better...


app/services/rapid_ready_for_decision/views/shared/medications.erb
```diff
@@ -10,13 +10,13 @@
 </p>
 
 <% if @assessed_data[:medications].present? %>
+  <p> <br /> </p>
```
allthesignals, line 13: Unrelated to the main point of the PR but this fixes a small spacing issue that's been bugging me

spec/jobs/rapid_ready_for_decision/form526_asthma_job_spec.rb
```diff
@@ -51,6 +51,10 @@
       it 'returns the active medication requests' do
         expect(subject[:medications].count).to eq(19)
       end
+
+      it 'flags potential asthma-related medication' do
+        expect(subject[:medications].select { |med| med[:flagged] }.count).to eq(3)
```
leikkisa, line 56: This is for an above line, but saw that Yoom started adding the `asthma_claim_for_increase` trait when doing `create(:form526_submission, user: user)`

spec/services/rapid_ready_for_decision/fast_track_pdf_generator_spec.rb
```diff
@@ -108,6 +108,28 @@
...
+             'description' => 'Hydrocortisone 10 MG/ML Topical Cream',
+             'notes' => ['Hydrocortisone 10 MG/ML Topical Cream'],
+             'dosageInstructions' => ['Once per day.', 'As directed by physician.'],
+             :flagged => false },
+           { 'status' => 'active',
+             'authoredOn' => '1998-02-12T02:15:52Z',
+             'description' => 'Dose Inhaler',
+             'notes' => ['Dose Inhaler'],
+             'dosageInstructions' => ['Once per day.', 'As directed by physician.'],
+             :flagged => true }]
```
leikkisa, line 125: This threw me off when looking at the PDF, I think it would normally get sorted so the flagged is first. Non-blocking but would be nice to fix this.

app/services/rapid_ready_for_decision/views/shared/medications.erb
leikkisa: I think somewhere here we want to include an extra section for the asthma claims:

> Important! Medications that may have been prescribed for asthma are marked with ►arrows◀︎. Claims adjudicators should check the patient’s medical records to see what the medication was prescribed for. 
See “How we identify asthma prescriptions” at the end of this document for more details.

app/workers/rapid_ready_for_decision/form526_asthma_job.rb
```diff
@@ -32,7 +32,17 @@ def query_and_assess_lighthouse(form526_submission)
...
+      transformed_medications = RapidReadyForDecision::LighthouseMedicationRequestData.new(medications).transform
+      flagged_medications = transformed_medications.map do |medication|
+        {
+          **medication,
+          flagged: asthma_keywords.any? do |keyword|
+            medication.to_s.downcase.include?(keyword)
+          end
+        }
+      end
+      flagged_medications.sort_by { |medication| medication[:flagged] ? 0 : 1 }
```
leikkisa, line 45: This looks like it's working well! I wonder if we'd want to do any additional sorting (by name or prescription date) but that's not in the design, so just a side comment.

leikkisa (approved):
This is looking good to me. I see some formatting tweaks that are on the Asthma design not reflected, however they seem minor and could be easy to handle in another PR to keep this focused on the medications section. Otherwise just added a couple of minor comments.

spec/services/rapid_ready_for_decision/fast_track_pdf_generator_spec.rb
allthesignals: @leikkisa correct, yes, it does get sorted and should be first in the list... will add to follow up list of things. TY!

spec/jobs/rapid_ready_for_decision/form526_asthma_job_spec.rb
allthesignals: Would love to figure out the best way to use this those test features (this spec does not rely on typical Rails models)

app/workers/rapid_ready_for_decision/form526_asthma_job.rb
allthesignals: @leikkisa so there is additional generic sorting done in the medication data request class. Because this sorting is specific to asthma I put it here. Lemme know if you have any more concerns downstream...

app/workers/rapid_ready_for_decision/form526_asthma_job.rb
```diff
@@ -32,7 +32,17 @@ def query_and_assess_lighthouse(form526_submission)
     def assess_medications(medications)
       return [] if medications.blank?
 
-      RapidReadyForDecision::LighthouseMedicationRequestData.new(medications).transform
+      asthma_keywords = RapidReadyForDecision::Constants::DISABILITIES[:asthma][:keywords]
```
yoomlam, line 35: I'd change this into a constant: `ASTHMA_KEYWORDS`

app/workers/rapid_ready_for_decision/form526_asthma_job.rb
```diff
@@ -32,7 +32,17 @@ def query_and_assess_lighthouse(form526_submission)
...
+      transformed_medications = RapidReadyForDecision::LighthouseMedicationRequestData.new(medications).transform
+      flagged_medications = transformed_medications.map do |medication|
+        {
+          **medication,
+          flagged: asthma_keywords.any? do |keyword|
+            medication.to_s.downcase.include?(keyword)
+          end
```
yoomlam, line 42: This could be a one-liner:
```suggestion
          flagged: asthma_keywords.any? { |keyword| medication.to_s.downcase.include?(keyword) }
```

spec/services/rapid_ready_for_decision/fast_track_pdf_generator_spec.rb
yoomlam: Remember to add a test to check ordering.

yoomlam (commented):
Left nonblocking comments

app/workers/rapid_ready_for_decision/form526_asthma_job.rb
RachalCassity: I agree.

app/workers/rapid_ready_for_decision/form526_asthma_job.rb
allthesignals: @yoomlam interesting, that throws a syntax error: 

>  dynamic constant assignment

it's not dynamic in the way i'm thinking of the word...

yoomlam: Ok. Nevermind then. I guess a lookup in a hash (even though it's a frozen hash) is considered dynamic 🤷

leikkisa (approved):
This looks good, we're going to address additional feedback from design in a follow up PR.

RachalCassity (approved):

yoomlam (approved):

yoomlam (approved):

leikkisa (approved):
👍

app/workers/rapid_ready_for_decision/asthma_processor.rb
```diff
@@ -25,10 +25,19 @@ def query_and_assess_lighthouse
       { medications: medications }
     end
 
+    ASTHMA_KEYWORDS = RapidReadyForDecision::Constants::DISABILITIES[:asthma][:keywords]
```
allthesignals, line 28: @yoomlam oh, so I guess this isn't an issue when it's defined on the class level (I guess because it can't be statically evaluated)

RachalCassity (approved):

PR #9625, opened 2022-04-13
Api 15056 birth past validation

stiehlrod:
## Description of change
Adds `#birth_date_is_in_the_past` and `claimant_birth_date_is_in_the_past` validations.

## Original issue(s)
[API-15056](https://vajira.max.gov/browse/API-15056)


modules/appeals_api/spec/models/notice_of_disagreement_spec.rb
```diff
@@ -86,6 +86,17 @@
     end
   end
 
+  context 'birth date isn\'t in the past' do
```
ElleeB, line 89: Let's update this to 'Veteran birth date...' since we now have claimant as well :)

cilestin, line 89: I agree with @ElleeB - let's match this wording with the text you added for claimant: "veteran birth date is in the future"

ElleeB (approved):

cilestin (changes requested):

ElleeB:
I'm going to make a ticket for streamlining the wording of the tests. It's in multiple places where we have to add clarity for veteran and also streamline `is not in the past` vs `is in the future`

cilestin (approved):

modules/appeals_api/spec/models/notice_of_disagreement_spec.rb
```diff
@@ -86,6 +86,17 @@
     end
   end
 
+  context 'Veteran birth date is not in the future' do
```
ElleeB, line 89: `Veteran birth date is in the future`

modules/appeals_api/spec/models/notice_of_disagreement_spec.rb
```diff
@@ -86,6 +86,17 @@
...
+    it 'using a birth date /not/ in the past creates an invalid record' do
```
ElleeB, line 90: This can be shortened to `creates an invalid record` since the context takes care of the first part of the statement

modules/appeals_api/spec/models/notice_of_disagreement_spec.rb
```diff
@@ -262,6 +273,19 @@
       it { expect(subject.class).to eq AppealsApi::Appellant }
     end
 
+    context 'claimant birth date is not in the future' do
```
ElleeB, line 276: I think we should add `when` to the start of this statement

ElleeB (approved):

PR #9627, opened 2022-04-13
38796 demographics service object

tpharrison:
## Description of change
Built the service layer for returning demographics data from VA Profile API.  The endpoint returns a preferred_name and gender_identity.

Reference ticket [#38796](https://github.com/department-of-veterans-affairs/va.gov-team/issues/38796).

The VA Profile swagger docs for demographics API are located here:
https://qa.vaprofile.va.gov:7005/demographics/swagger-ui/index.html?configUrl=/demographics/v3/api-docs/swagger-config

The parameters passed are Organization ID, CSP ID, and Assigning Authority.  Example...
`/demographics/demographics/v1/2.16.840.1.113883.4.349/168447%5EPN%5E200VLGN%5EUSDVA`

Specs are in `spec/lib/va_profile/demographics/service_spec.rb`

* There is an addition to the `settings.yml` file to include demographics configuration (under `vet360`)
* There is no feature flag used since the service isn't hooked up to a controller yet (that work will be in another PR to keep the sizes manageable).
* Sentry logging that was added in the service class.


tpharrison:
If this PR is too large, let me know and I'll break it up into two smaller ones.

ericboehs (approved):

PR #9629, opened 2022-04-14
updating logging for better PII protection

jperk51:
## Description of change
Slight update to some logging


RachalCassity (commented):
Does the user name or email need to be recorded to figure out why the mobile is not working? 
Ex:
Rails.logger.info('BGS External Key Nil', @user.common_name.nil? ? "" : @user.common_name.nil, @user.email.nil? ? "" : @user.email) if external_key.blank?

jperk51:
Seeing the values shouldn't matter for this. We are more concerned with knowing if there are any users who have both as nil. Additionally we don't want to log both name and email next to each other for PII reasons

RachalCassity (approved):

PR #9633, opened 2022-04-14
RRD: refactor release_pdf method

yoomlam:
## Description of change

Refactor `release_pdf?` to `RrdProcessor` base class.
Add spec file for `RrdProcessor`.
Also add `Constants.processor` method to resolve processor classes where their class strings (e.g., `'RapidReadyForDecision::HypertensionProcessor'`) are defined.

## Original issue(s)
`release_pdf?` is specific to asthma. We'll want to use it for future new contentions.

Refactor. No change in behavior.


app/workers/rapid_ready_for_decision/rrd_processor.rb
```diff
@@ -33,7 +34,10 @@ def generate_pdf(_assessed_data)
 
     # Override this method to prevent the submission from getting the PDF and special issue
     def release_pdf?
-      true
+      flipper_symbol = "rrd_#{@disability_struct[:flipper_name].downcase}_release_pdf".to_sym
+      return true unless Flipper.exist?(flipper_symbol)
```
yoomlam, line 38: By default, if Flipper symbol doesn't exist (e.g., `rrd_hypertension_release_pdf`), return true. We want this default so we can remove the Flipper flag once we're done experimenting with the disability.

leikkisa (approved):
This looks good to me

RachalCassity (approved):

PR #9637, opened 2022-04-14
RRD: add pdf_guid in the RRD notification

yoomlam:
## Description of change

Add `pdf_guid` in the RRD notification.
When we generate an RRD PDF for a claim, include a path to the PDF in the Slack notification so we can look up the associated RRD PDF. 

## Original issue(s)
[Add asthma PDF path to notifications](https://www.notion.so/Dashboard-RRD-Health-Evidence-24d77d9988a546c3b19a3c73e7ced29b?p=d19b5c233efa4174ab09dac7dc821888)

The notification is for internal use.


spec/mailers/rrd_completed_mailer_spec.rb
```diff
@@ -28,7 +28,10 @@
                                                               form_json: form_json)
       # Set the bp_readings_count like `add_medical_stats` is expected to do
       form_json = JSON.parse(submission.form_json)
-      form_json['rrd_metadata'] = { med_stats: { bp_readings_count: bp_readings_count } }
+      form_json['rrd_metadata'] = {
+        med_stats: { bp_readings_count: bp_readings_count },
+        pdf_guid: 'a950ef07-9eaa-4784-b5af-bda8c50a83f9'
+      }
```
allthesignals, line 34: Can we verify this is the real structure? Even just spot-check since we're mocking it.

allthesignals (approved):

spec/mailers/rrd_completed_mailer_spec.rb
yoomlam: Yes, good question. This test ensures it's there: https://github.com/department-of-veterans-affairs/vets-api/blob/90e350090d5afb98604e370a6d0cc732c6ab2e91/spec/services/rapid_ready_for_decision/fast_track_pdf_upload_manager_spec.rb#L101-L108

If you're referring to an higher-level test, I didn't think this was important enough to warrant that.

allthesignals: Good enough for me! I agree.

thilton-oddball (approved):

PR #9639, opened 2022-04-14
Logging for mobile payment history filtering

aherzberg:
## Description of change

Getting a Type check error when filter payments by year. `invalid year (not numeric)`. Add logging to see what is coming in as non-numeric.
## Original issue(s)


jperk51:
Would it also be worth logging `available_years`?

aherzberg:
> Would it also be worth logging `available_years`?

Good idea. Updated

kpethtel (approved):
This should work. I still don't understand how we're getting a non-numeric. The Date#year docs say it returns an integer and we compact out any nils so 🤷.

PR #9642, opened 2022-04-15
API-15034: Update attribute properties for ClaimsApi::Veteran

hinzed1127:
## Description of change

Based on the [errors we're seeing](http://sentry.vfs.va.gov/organizations/vsp/issues/92733/events/58e1aa2fe7a1461aba8cabe424efa9b9/?environment=production&project=3&statsPeriod=24h), it's looks [like this function](https://github.com/department-of-veterans-affairs/vets-api/blob/master/app/models/mpi_data.rb#L170-L182) would have never previously worked with the `ClaimsApi::Veteran` model as it's presently constructed. Since we seem to just be getting new Veterans that need to be added to MPI first, we're running into this now.

## Original issue(s)
https://vajira.max.gov/browse/API-15034


modules/claims_api/app/models/claims_api/veteran.rb
```diff
@@ -17,6 +17,8 @@ class Veteran
        gender
        birls_file_number
        uuid
+       icn_with_aaid
+       search_token
```
hinzed1127, line 21: I initially tried to add these with
```ruby
delegate :icn_with_aaid, to: :mpi, allow_nil: true
delegate :search_token, to: :mpi, allow_nil: true
```
but I was still getting complaints that `icn_with_aaid=` and `search_token=` were undefined methods for `ClaimsApi::Veteran`. I'm not sure if I was doing it wrong trying this other way or if I'm missing something with Ruby.

SlakrHakr (approved):

PR #9654, opened 2022-04-18
Api 14342 fix overlapping pdf fields

stiehlrod:
## Description of change
Changes the `stamper.rb` file to adjust the size and position of the header stamps.
Adds default parameter to `datestamp_pdf.rb` to add the ability to adjust sizes based on forms.
Replaces fixtures/expected pdfs with new stamps.

## Original issue(s)
[API-14342](https://vajira.max.gov/browse/API-14342)

Adds default parameter to `datestamp_pdf.rb` to add the ability to adjust sizes based on forms.


modules/appeals_api/app/services/appeals_api/pdf_construction/stamper.rb
```diff
@@ -21,33 +21,45 @@ def date_stamp_path
...
-                       width: 215,
-                       height: 10
+                       width: 195,
+                       height: 8
         end
         veteran_stamp_path
       end
+
+      def generate_text(name_max_length, consumer_name)
+        leading_space = '  ' * (name_max_length - consumer_name.length).abs
```
ElleeB, line 56: I don't _think_ we need the .abs here. Please school me if I'm wrong lol!

modules/appeals_api/app/services/appeals_api/pdf_construction/stamper.rb
```diff
@@ -21,33 +21,45 @@ def date_stamp_path
...
+        if consumer_name.length > name_max_length
+          "Submitted by #{consumer_name.truncate(name_max_length)} via api.va.gov"
```
ElleeB, line 58: I think we need to get approval from Kelly before truncating the consumer name.

modules/appeals_api/app/services/appeals_api/pdf_construction/stamper.rb
stiehlrod: In one of our examples the name is so long that I got a negative number

RachalCassity (dismissed):

cilestin:
Talked this through and decided that:
* The priority is for intake specialists to be able to keep pages together, so the veteran's name & last-4 SSN stamp should afford the most space we can get away with.
* Last-4 SSN should always be displayed even if the veteran's name is long.
* Truncating the consumer name to 10 characters is fine.
* For now, no moving the header stamps into the footer or anything like that - we can revisit that separately if necessary.

ElleeB (approved):
Yay!

PR #9656, opened 2022-04-18
RDD: Warn if multiple Account records with different ICNs found

yoomlam:
## Description of change

There's no DB constraint that guarantees uniqueness of `Account.edipi`, so return an array if there are multiple Accounts found. and send a warning notification if multiple Account records found with different ICNs.

## Original issue(s)
There's no DB constraint that guarantees uniqueness of `Account.edipi`.
The original implementation returned the first Account found with a matching `edipi`, ignoring any other matching Account records.

This is not addressing an actual  problem in prod, but given the DB constraints, the scenario is possible.


app/workers/rapid_ready_for_decision/rrd_processor.rb
```diff
@@ -96,15 +97,33 @@ def icn
...
+    # There's no DB constraint that guarantees uniqueness of edipi, so return an array if there are multiple Accounts
+    def account_matching_edipi(edipi)
+      return nil unless edipi
+
+      accounts = Account.where(edipi: edipi).to_a
+      return nil if accounts.empty?
+
+      return accounts.first if accounts.size == 1
+
+      accounts
```
leikkisa, line 126: I think it's pretty confusing for this method to return either a single account or an array of accounts (here in this method as well as where it's used above).

app/workers/rapid_ready_for_decision/rrd_processor.rb
```diff
@@ -96,15 +97,33 @@ def icn
...
       raise AccountNotFoundError, "for user_uuid: #{form526_submission.user_uuid} or their edipi" unless account_record
 
-      account_record.icn.presence
+      return account_record.icn.presence unless account_record.is_a?(Array)
+
+      # Multiple Account records should have the same ICN
+      icns = account_record.pluck(:icn).uniq.compact
+      return icns.first if icns.size == 1
+
+      raise MultipleIcnsFound, "for the user '#{form526_submission.user_uuid}': #{icns}"
```
leikkisa, line 106: I might just not be understanding the problem with multiple ICNs and edipis or the consequences of using the wrong one (or when there are not consequences. I could imagine both cases, I could see it being better to go ahead with find_by instead of returning an error though.

What happens if the wrong account is used? Is there any possibility of it getting matched back to the correct thing through MPI or something?

leikkisa (commented):
Just commenting for now, I think that I can understand how this would be better but don't feel confident in it yet.

app/workers/rapid_ready_for_decision/rrd_processor.rb
yoomlam: > What happens if the wrong account is used?

I'm not sure. Right now, we have no indication when this happens.
I'd be willing to send an alert notification to Slack instead of an error.

app/workers/rapid_ready_for_decision/rrd_processor.rb
yoomlam: Yea. Any alternative ideas to make this less confusing?

app/workers/rapid_ready_for_decision/rrd_processor.rb
leikkisa: It might be good to do that to start, and look at some of them.

app/workers/rapid_ready_for_decision/rrd_processor.rb
yoomlam: Updated to return array of Accounts

app/workers/rapid_ready_for_decision/rrd_processor.rb
```diff
@@ -63,23 +63,6 @@ def add_medical_stats(assessed_data)
...
-        The error was: #{error_message}. The backtrace was:\n #{backtrace.join(",<br/>\n ")}
-      BODY
-      ActionMailer::Base.mail(
-        from: ApplicationMailer.default[:from],
-        to: Settings.rrd.alerts.recipients,
-        subject: 'Rapid Ready for Decision (RRD) Job Errored',
-        body: body
-      ).deliver_now
-    end
-
```
yoomlam, line 82: No longer used, in favor of `Form526RapidReadyForDecisionConcern.send_rrd_alert_email` instead.

app/workers/rapid_ready_for_decision/rrd_processor.rb
```diff
@@ -93,18 +76,36 @@ def patient_info
...
       edipi = form526_submission.auth_headers['va_eauth_dodedipnid'].presence
-      Account.find_by(edipi: edipi) if edipi
+      accounts_matching_edipi(edipi)
+    end
+
+    # There's no DB constraint that guarantees uniqueness of edipi, so return an array if there are multiple Accounts
+    def accounts_matching_edipi(edipi)
+      return [] unless edipi
+
+      Account.where(edipi: edipi).order(:id).to_a
```
yoomlam, line 108: `find_by` has no implied order, so this PR orders by the primary key for consistency.
See https://github.com/rubocop/rails-style-guide/issues/76.
https://apidock.com/rails/v4.0.2/ActiveRecord/FinderMethods/find_by source code:
```ruby
def find_by(*args)
   where(*args).take
end
```

https://apidock.com/rails/v4.0.2/ActiveRecord/FinderMethods/take
> Gives a record (or N records if a parameter is supplied) without any implied order.

app/workers/rapid_ready_for_decision/rrd_processor.rb
```diff
@@ -93,18 +76,36 @@ def patient_info
...
 
-      account_record.icn.presence
+      icns.first
     end
 
-    def account
+    def accounts
       account = Account.lookup_by_user_uuid(form526_submission.user_uuid)
-      return account if account
+      return [account] if account
```
leikkisa, line 98: I wonder if the ID.me or Login.gov find_by's generally return the same account or if they ever return different accounts.

app/workers/rapid_ready_for_decision/rrd_processor.rb
```diff
@@ -93,18 +76,36 @@ def patient_info
     end
 
     def icn
-      account_record = account
-      raise AccountNotFoundError, "for user_uuid: #{form526_submission.user_uuid} or their edipi" unless account_record
+      account_records = accounts
```
leikkisa, line 79: I don't understand why this variable is needed

app/workers/rapid_ready_for_decision/rrd_processor.rb
leikkisa: I wonder if there's a preferred order if there are multiple. I definitely like this change for consistency in the absence of one.

leikkisa (approved):
I like this, 👍  for the checking and improvements. I'll be curious to actually see some of these.

app/workers/rapid_ready_for_decision/rrd_processor.rb
yoomlam: They'll only ever return one Account since those fields must be unique: https://github.com/department-of-veterans-affairs/vets-api/blob/fd7236cc981a4fc6401d61800d250a6385be81d1/db/schema.rb#L51-L54

app/workers/rapid_ready_for_decision/rrd_processor.rb
yoomlam: I'm not memoizing `accounts` and I don't want to DB query for the account more than needed within this `icn` method.

app/workers/rapid_ready_for_decision/rrd_processor.rb
yoomlam: Once this PR is merged, we'll have some concrete examples to present to the Platform team, if this issue ever occurs. Without this PR, we'll never know.

RachalCassity (approved):

PR #9658, opened 2022-04-18
Revocation endpoint 37192

jscanish:
## Description of change
Creates an API endpoint that invalidates all refresh/access tokens for a given user.

## Original issue(s)
[department-of-veterans-affairs/va.gov-team#0000](https://github.com/department-of-veterans-affairs/va.gov-team/issues/37192)

Tested with RSpec and with user session.


app/services/sign_in/session_revoker.rb
```diff
@@ -0,0 +1,63 @@
...
+    def perform
+      anti_csrf_check if enable_anti_csrf
+      find_valid_oauth_session
+      detect_token_theft
+      delete_session!
+    end
+
+    private
+
+    def anti_csrf_check
```
bramleyjl, line 22: Since almost all of these private methods are duplicates of the session refresher ones I'm wondering if there's some way to consolidate them? Since they rely heavily on instance variables maybe they could have a shared parent class like `Session::Revoker` and `Session::Refresher` and those validation logic that takes up the bulk of the class could go there?

bramleyjl (changes requested):

spec/controllers/sign_in_controller_spec.rb
```diff
@@ -634,6 +634,87 @@
...
+
+    let(:refresh_token_param) { { refresh_token: refresh_token } }
+    let(:refresh_token) { 'example-refresh-token' }
+    let(:anti_csrf_token_param) { { anti_csrf_token: anti_csrf_token } }
+    let(:anti_csrf_token) { 'example-anti-csrf-token' }
+    let(:enable_anti_csrf) { false }
+
+    before { allow(Settings.sign_in).to receive(:enable_anti_csrf).and_return(enable_anti_csrf) }
+
+    context 'when refresh_token is a random string' do
```
bosawt, line 648: Can you add specs for the case when `enable_anti_csrf` is set to true? Luckily it should prob be basically a straight copy from the `refresh` route

spec/services/sign_in/session_revoker_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+  let(:session_revoker) do
+    SignIn::SessionRevoker.new(refresh_token: refresh_token,
+                               anti_csrf_token: input_anti_csrf_token,
+                               enable_anti_csrf: enable_anti_csrf)
+  end
+
+  describe '#perform' do
+    subject { session_revoker.perform }
+
+    context 'given a refresh token' do
```
bosawt, line 15: Also, can you add specs when `enable_anti_csrf` is passed in here? Again, think it should be a pretty straight copy from `session_refresher_specs`

bosawt (changes requested):
Looking good! Tested by signing in with a user, proving `/introspect` works, then using `/logout` endpoint and confirming that the refresh token can no longer be used for `/refresh` endpoint. `/introspect` still works with the access token, but it expires after 5 minutes, as expected.

Just have a couple quick comments about specs to add for the anti_csrf stuff. It isn't used for mobile currently but will likely be used with the website interface, so will be good to make sure we're testing it

bramleyjl (approved):
Authenticated and tested the revocation endpoint, then tried to re-refresh the revoked token. Looks good to me!

spec/controllers/sign_in_controller_spec.rb
bosawt: 👍

spec/services/sign_in/session_revoker_spec.rb
bosawt: 👍

bosawt (approved):

PR #9659, opened 2022-04-18
Pv my health module

patrickvinograd:
## Description of change
Refactors the MHV-related API controllers and serializers into a new module 'my_health', to enable cleaner ownership by the health apartment teams as they come aboard to develop on VA.gov. 

Functionality is equivalent to that of the main rails application, however it's available at a new route and I bumped the version to v1. Intent would be to find the couple remaining uses of the main app API routes, switch them to use this, and deprecate the v0 endpoints. 

This PR only includes the secure messaging endpoints. Health records and prescriptions are also intended to move to this module, but I will do that in separate PRs to keep the scope down. 

The upstream service integration code remains in lib/ and is shared with the flagship mobile app. A future PR will institute a CODEOWNERS entry that makes this shared responsibility explicit. 

## Original issue(s)
https://github.com/department-of-veterans-affairs/va.gov-team/issues/39364


LindseySaari (approved):
👍  I think this looks good to go out together. When you're ready to merge, we can do a dangerbot exception

PR #9666, opened 2022-04-18
Mobile Payment History Scheduled Payment Filtering 

aherzberg:
## Description of change

Future payments are not being filtered out until after the available years array, used to paginate, are created, this has an edge case of when there are only scheduled payments, the array of available years is empty causing the filtering to error out.


kpethtel (approved):
🚀

PR #9667, opened 2022-04-18
API-15140: Claims API - Update string property validation for 526 submissions

hinzed1127:
## Description of change

This adds in a length validation to the ~`treatments[].center.name`~ `treatments[].center.country` property. Previously it would allow an empty string, which passed our validations before failing against submissions to EVSS

## Original issue(s)
https://vajira.max.gov/browse/API-15140

- [API-15242](https://vajira.max.gov/browse/API-15242) is a complementary PR to this one that addresses a different set of failures we've been observing with 526 submissions to EVSS


modules/claims_api/config/schemas/526.json
```diff
@@ -1091,6 +1091,8 @@
               "name": {
```
jw81, line 1091: I'm confused on this because didn't Michael Harlow say that it was the `treatments[].center.country` that had the 1-50 characters requirement? ([REF](https://dsva.slack.com/archives/C8R3JS8BU/p1650304654062279?thread_ts=1650304364.435459&cid=C8R3JS8BU)]

modules/claims_api/spec/requests/v1/disability_compensation_request_spec.rb
```diff
@@ -239,6 +239,26 @@
           ]
         end
 
+        context "when 'treatments[].center.name' is an empty string'" do
```
jw81, line 242: Could also add a check here for making sure it behaves properly when the value is > 50 characters.

modules/claims_api/config/schemas/526.json
hinzed1127: ope! you're totally right. Remedying that in my next commit

modules/claims_api/spec/requests/v1/disability_compensation_request_spec.rb
hinzed1127: good idea, adding that in, as well

modules/claims_api/spec/requests/v1/disability_compensation_request_spec.rb
```diff
@@ -1017,6 +1058,7 @@ def obj.class
                     allow_any_instance_of(MPI::Service).to receive(:find_profile).and_return(mvi_profile_response)
 
                     post path, params: data, headers: auth_header
+                    expect(response.status).to eq(200)
```
hinzed1127, line 1061: Forgot to add this in to https://github.com/department-of-veterans-affairs/vets-api/pull/9642 . Basically just makes sure this is _actually succeeding_ . The test was throwing 422s and we were just missing it

jw81 (approved):

PR #9671, opened 2022-04-19
[API-15088] - ClaimsApi - Set "doctype" for 526 attachment uploads

jw81:
## Description of change

Sets the `docType` to `L023` for any 526 attachment being sent to EVSS.

**Background** :: When someone hits the `POST claims/v1/forms/526/:id/attachments` to add attachments to their claim, it first stores the docs in AWS S3 and then a background job will transfer the docs from S3 to EVSS.  EVSS requires that each doc has a `docType`.  For whatever reason, Benefits ClaimsApi appears to have always set `docType` to `nil`.  Which is not allowed by EVSS.  So (for now at least), always set the `docType` value to `L023`.  Apparently, in EVSS, `L023` maps to a value of `Other Correspondence`. ([REF](https://github.com/department-of-veterans-affairs/vets-api/blob/72546ee27484a47b2a9ef122eb6d4efd38c214e9/app/models/evss_claim_document.rb#L29)). Since literally any kind of document can be provided, this seems like the most likely "catch-all" value that can be used.

## Original issue(s)
[API-15088](https://vajira.max.gov/browse/API-15088)


modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
```diff
@@ -108,7 +109,7 @@ def upload_supporting_documents
 
           documents.each do |document|
             claim_document = claim.supporting_documents.build
-            claim_document.set_file_data!(document, params[:doc_type], params[:description])
+            claim_document.set_file_data!(document, EVSS_DOCUMENT_TYPE, params[:description])
```
SlakrHakr, line 112: why not still accept `doc_type` if provided but fall back to L023 if not provided?

SlakrHakr: ah, is `doc_type` not in the documented schema...?

jw81: ^^^ Exactly

SlakrHakr (approved):

modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
hinzed1127: I think this is fine for now, but I like the idea of adding an optional field and then switching to `EVSS_DOCUMENT_TYPE = params[:doc_type] || 'L023'` in the future

hinzed1127 (approved):

PR #9673, opened 2022-04-19
[API-15088] - ClaimsApi - Ensure Unique FileNames for 526 Attachments

jw81:
## Description of change
Ensures that binary PDF attachments submitted to the `POST /claims/v1/forms/526/:id/attachments` endpoint have unique file names.

**Background** :: When a consumer chooses to send us Binary PDF attachments, we have no control over the filenames for those attachments.  It was observed that if the consumer sends us 5 attachments named `attachment.pdf`, then we end up with only 1 doc named `attachment.pdf` stored in S3.  What happens is, when we iterate the 5 attachments, the first attachment is uploaded to S3 and is named `attachment.pdf`.  Then when the second attachment is uploaded to S3, it overwrites the file that was just stored in S3 because it too is named `attachment.pdf`.  This pattern repeats itself until in the end, you only have 1 doc stored in S3 named `attachment.pdf` and it's the very last attachment that was in the loop.  This new logic ensures that each attachment will have a unique filename so that nothing stored in S3 gets overwritten.

## Original issue(s)
[API-15088](https://vajira.max.gov/browse/API-15088)


modules/claims_api/app/controllers/claims_api/v1/forms/base.rb
```diff
@@ -61,7 +65,7 @@ def documents
         def decode_document(document)
           base64 = document.split(',').last
           decoded_data = Base64.decode64(base64)
-          filename = "temp_upload_#{Time.zone.now.to_i}.pdf"
+          filename = "temp_upload_#{SecureRandom.urlsafe_base64(8)}.pdf"
```
jw81, line 68: While locally running through everything, I noticed that if the processing is fast enough, you can end up with 2 (or more) Base64 encoded attachments that get the same timestamp in milliseconds.  Which would lead to the same overwriting problem in S3 that this ticket is addressing.

So let's prevent that scenario by generating a completely random string, rather than depending on the epoch.

modules/claims_api/app/controllers/claims_api/v1/forms/disability_compensation_controller.rb
```diff
@@ -78,7 +78,7 @@ def upload_form_526
           pending_claim = ClaimsApi::AutoEstablishedClaim.pending?(params[:id])
 
           if pending_claim && (pending_claim.form_data['autoCestPDFGenerationDisabled'] == true)
-            pending_claim.set_file_data!(documents.first, params[:doc_type])
+            pending_claim.set_file_data!(documents.first, EVSS_DOCUMENT_TYPE)
```
jw81, line 81: I forgot to add this in the [previous PR](https://github.com/department-of-veterans-affairs/vets-api/pull/9671).  
So I'm handling it here now.

acovrig (approved):

SlakrHakr (approved):

jw81:
**NOTE** :: I really tried for a couple of hours to find a way to add specs for this, but I just couldn't make it happen.  I think the that the doc logic really needs to be broken out into some kind of service object or something.  Just something that can be tested in isolation and much more cleanly.  I'm going to create a ticket to remind us to do that in V2.  Will link here once I've done that.

modules/claims_api/app/controllers/claims_api/v1/forms/base.rb
hinzed1127: 💯

hinzed1127 (approved):

jw81:
> **NOTE** :: I really tried for a couple of hours to find a way to add specs for this, but I just couldn't make it happen. I think the that the doc logic really needs to be broken out into some kind of service object or something. Just something that can be tested in isolation and much more cleanly. I'm going to create a ticket to remind us to do that in V2. Will link here once I've done that.

I said I'd create a ticket to remind us to do this better in V2 and [here is that ticket](https://vajira.max.gov/browse/API-15350).

PR #9679, opened 2022-04-20
Console UI Catalog File for the vets-api Rails App

jchapin:
## Description of change

The addition of this configuration file will allow the Console UI to give the vets-api a catalog entry and profile page in the Console UI. Additionally it will allow the Console UI to monitor and update all of the tools connected to the vets-api project.

## Original issue(s)

Related to:

https://github.com/department-of-veterans-affairs/platform-console-ui/issues/338

and

https://github.com/department-of-veterans-affairs/platform-console-ui/issues/275

Right now the Console UI has the platform-console-ui component and its dependencies recorded. Given that vets-api is a key dependency for the majority of vets-website VFS apps, we want to set it up in the Console UI as soon as practicable. It will also help test our auto-discovery of projects in the organization which should be referenced from the Console UI (338) and test our ability to identify projects which should be a part of the Console UI, but have not been onboarded yet.

1. No settings need to change.
2. No feature flags.
3. No new logging.
4. No new metrics.
5. This PR does not involve interacting with Swagger docs, but a subsequent PR may record their location for the Console UI.
6. No PII or PHI Concerns

This is simply a configuration file consumed by another system.

We have a similar, but more complex file in our own repository:

https://github.com/department-of-veterans-affairs/platform-console-ui/blob/main/catalog-info.yaml


catalog-info.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: vets-api
```
jchapin, line 4: Short name of the component. There will also be an API catalog entry that's similarly named.

catalog-info.yaml
```diff
@@ -0,0 +1,15 @@
...
+  description: This project provides common APIs for applications that live on VA.gov (formerly vets.gov APIs).
```
jchapin, line 5: I just used the description in the vets-api README for now.

catalog-info.yaml
```diff
@@ -0,0 +1,15 @@
...
+  annotations:
+    github.com/project-slug: department-of-veterans-affairs/vets-api
```
jchapin, line 7: GitHub slug... owner/repo

https://github.com/department-of-veterans-affairs/vets-api

catalog-info.yaml
```diff
@@ -0,0 +1,15 @@
...
+    backstage.io/techdocs-ref: dir:.
+  tags:
+    - ruby
+    - rails
```
jchapin, line 11: Two starter tags.

catalog-info.yaml
```diff
@@ -0,0 +1,15 @@
...
+spec:
+  type: website
```
jchapin: Called it a website, but maybe there's a better classification.

catalog-info.yaml
```diff
@@ -0,0 +1,15 @@
...
+  owner: vsp-tools-backend
```
jchapin, line 14: GitHub team that owns this product.

catalog-info.yaml
```diff
@@ -0,0 +1,15 @@
...
+  lifecycle: production
```
jchapin, line 15: Currently in production.

catalog-info.yaml
jchapin:
```
The software catalog accepts any type value, but an organization should take great care to establish a proper taxonomy for these. Tools including Backstage itself may read this field and behave differently depending on its value. For example, a website type component may present tooling in the Backstage interface that is specific to just websites.

The current set of well-known and common values for this field is:

service - a backend service, typically exposing an API
website - a website
library - a software library, such as an npm module or a Java library

```

service is better.

catalog-info.yaml
```diff
@@ -0,0 +1,15 @@
...
+  name: vets-api
+  description: This project provides common APIs for applications that live on VA.gov (formerly vets.gov APIs).
+  annotations:
+    github.com/project-slug: department-of-veterans-affairs/vets-api
+    backstage.io/techdocs-ref: dir:.
+  tags:
+    - ruby
+    - rails
+spec:
+  type: service
```
jchapin, line 13:
```
The current set of well-known and common values for this field is:

service - a backend service, typically exposing an API
website - a website
library - a software library, such as an npm module or a Java library
```

RachalCassity (approved):

catalog-info.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: backstage.io/v1alpha1
```
ericboehs, line 1: It would be nice if this could be in the `config/` directory or `.github/` instead of adding another root-level file.

catalog-info.yaml
ericboehs: This team name may change in the coming days/week.

ericboehs (approved):

PR #9680, opened 2022-04-20
Api 14961 add sc nvc schema

stiehlrod:
## Description of change
Adds information to the Supplemental Claims (SC) schema in order to take the first step towards enabling non-veteran claimants (NVC) to fill out a SC on behalf of a veteran.

## Original issue(s)
[API-14961](https://vajira.max.gov/browse/API-14961)


modules/appeals_api/config/schemas/v2/200995_headers.json
```diff
@@ -103,5 +128,19 @@
         { "$ref": "#/definitions/nonBlankString" }
       ]
     }
+  },
+  "additionalProperties": false,
```
ElleeB, line 132: I believe we stopped using the `additionalProperties` key in supplemental claims. @cilestin, can you confirm?

modules/appeals_api/config/schemas/v2/200995_headers.json
```diff
@@ -103,5 +128,19 @@
...
+  "required": [
+    "X-VA-SSN",
+    "X-VA-First-Name",
+    "X-VA-Last-Name",
+    "X-VA-Birth-Date"
+  ],
+  "dependencies": {
+    "X-VA-Claimant-SSN": ["X-VA-Claimant-First-Name", "X-VA-Claimant-Last-Name", "X-VA-Claimant-Birth-Date"],
```
ElleeB, line 140: I think we want to remove birth date from this list. Check me on this.

modules/appeals_api/spec/fixtures/v2/valid_200995.json
```diff
@@ -17,6 +17,20 @@
         "email": "josie@example.com",
         "timezone": "America/Chicago"
       },
+      "claimant": {
```
ElleeB, line 20: We may only want to put claimant in the extra fixture. I only say this because I recall that recently @cilestin made some changes to HLR and tests claimant only in the extra fixture. Do we want to still do that even though it reduces the claimant data testing coverage?

modules/appeals_api/spec/fixtures/v2/valid_200995_headers.json
```diff
@@ -7,5 +7,8 @@
   "X-VA-File-Number": "987654321",
   "X-VA-Insurance-Policy-Number": "987654321123456789",
   "X-Consumer-Username": "va.gov",
-  "X-Consumer-ID": "some-guid"
+  "X-Consumer-ID": "some-guid",
```
ElleeB, line 10: I _think_ we want these headers in the extra fixtures only. Maybe we need a _minimal_ 200995 fixture set and test??

modules/appeals_api/config/schemas/v2/200995_headers.json
cilestin: Confirmed, we should no longer be adding `additionalProperties` to our schemas.

cilestin (changes requested):
Couple small changes to make, but otherwise lookin' good!

ElleeB (approved):
🚀

PR #9684, opened 2022-04-20
API 15252 HLR NVC Feedback

stiehlrod:
## Description of change
Implements changes proposed by Joe to the API docs describing the new Non-Veteran Claimant support for HLRs.  These changes should be made in lower environments and Production.

## Original issue(s)
[API-15252](https://vajira.max.gov/browse/API-15252)


modules/appeals_api/spec/support/rswag_config.rb
```diff
@@ -137,6 +137,24 @@ def generic_schemas(ref_root)
           { '$ref': "#{ref_root}/date" }
         ]
       },
+      'X-VA-Claimant-First-Name': {
```
ElleeB, line 140: Should all of the claimant headers be down here? Birthdate and SSN included?

ElleeB (commented):
This looks great. One minor change, and then you'll need to run the rswag rake task to generate the new docs :)

modules/appeals_api/config/schemas/v2/200996.json
```diff
@@ -108,6 +108,7 @@
                 },
                 "claimant": {
                   "type": "object",
+                  "description": "If the claimant is a non-Veteran, claimant information is required. Without this information, we will be unable to identify the claimant. If you're a healthcare provider or an employee of a healthcare provider requesting review of a VA payment decision, you must identify the healthcare provider as the claimant and complete all relevant information in the claimant object.",
```
cilestin, line 111: I think we'd be safe to add this description to all our schemas that include claimant now, so that we don't forget to later.

modules/appeals_api/spec/support/rswag_config.rb
cilestin: Yeah, I think we should. The list of schemas doesn't mean it's used everywhere, just that it's one of many available across our endpoints.

cilestin (dismissed):
Marking as changes requested since some of the comments previously would need changes before being resolved.

ElleeB (approved):
🎈

PR #9693, opened 2022-04-21
RRD: remove unnecessary Sidekiq::Worker.clear_all

yoomlam:
## Description of change

In RRD tests, remove unnecessary `Sidekiq::Worker.clear_all` calls.
Also remove `type: :worker` from RSpecs for Processor classes.

## Original issue(s)
Unnecessarily calling `Sidekiq::Worker.clear_all`. It is already called for each test here: https://github.com/department-of-veterans-affairs/vets-api/blob/63926ac78e603c59c89edef48d54d6793e0bba03/spec/rails_helper.rb#L184

Test clean up PR.


RachalCassity:
@yoomlam  Should Sidekiq::Worker.clear_all be removed for all jobs as well?

yoomlam:
> @yoomlam Should Sidekiq::Worker.clear_all be removed for all jobs as well?

Probably, but I'm only doing this for tests that our team is responsible for.

RachalCassity (approved):

PR #9694, opened 2022-04-21
RRD: use metadata_hash so that we write to DB only once

yoomlam:
## Description of change

Pass around `metadata_hash` so that RRD processors write to the DB only once.

## Original issue(s)
Unnecessary frequent writing to the DB.

Optimization. No change to behavior.


allthesignals (approved):
My superficial understanding is that I'm observing fewer `#add_metadata` calls which means fewer writes to DB. Seems to be what we want here.

Any theories about potential unintended consequences? One left-field theory that comes it mind: a mailer makes a call to the model class method, which implies another lookup of the record. Given that there are fewer writes, if the mailer makes the call before the appropriate write, it'll have stale data.

yoomlam:
> My superficial understanding is that I'm observing fewer `#add_metadata` calls which means fewer writes to DB. Seems to be what we want here.

Yup!
 
> Any theories about potential unintended consequences? One left-field theory that comes it mind: a mailer makes a call to the model class method, which implies another lookup of the record. Given that there are fewer writes, if the mailer makes the call before the appropriate write, it'll have stale data.

Good thought! I considered the scenario where we may want the intermediate metadata values on the record itself. Since this is RRD metadata and the `metadata_hash` is available to any code running within an `RrdProcessor`, that RRD code should/could use the `metadata_hash` instead. But you're right if there is code running outside of an `RrdProcessor` that wants RRD metadata while `RrdProcessor` is running, we should write the metadata to the record -- I'm not aware of such code and I don't think there would be since that scenario would also have a race condition problem.

allthesignals (approved):

RachalCassity (approved):

PR #9695, opened 2022-04-21
RRD: remove unnecessary :vcr tag from tests

yoomlam:
## Description of change

Remove unnecessary `:vcr` tag from RRD tests.

## Original issue(s)
The `:vcr` tag appended to tests is use to create and locate VCR cassettes according to the class name. Since we're explicitly calling `VCR.use_cassette` with a specific directory path and we don't need to create new cassettes, we don't need `:vcr`. See https://www.rubyguides.com/2018/12/ruby-vcr-gem/

Cleans up tests.


allthesignals (approved):
fantastic

leikkisa (approved):
👍

RachalCassity (approved):

PR #9698, opened 2022-04-21
Swagger docs

stiehlrod:
## Description of change
Adds Swagger docs for dev and prod. Changes served components file to update change missed in previous PR.

## Original issue(s)
[API-15252](https://vajira.max.gov/browse/API-15252)


modules/appeals_api/app/swagger/appeals_api/v2/swagger.json
```diff
@@ -6745,6 +6746,26 @@
                                   "email": "bobsemail@bobbytablesemail.com",
                                   "timezone": "America/Chicago"
                                 },
+                                "claimant": {
```
cilestin, line 6749: Since SC NVCs are still in progress, we don't want to put anything into this production doc regarding SC NVCs. Let's remove this addition.

modules/appeals_api/app/swagger/appeals_api/v2/swagger.json
```diff
@@ -7026,6 +7047,26 @@
                           "email": "bobsemail@bobbytablesemail.com",
                           "timezone": "America/Chicago"
                         },
+                        "claimant": {
```
cilestin, line 7050: Same here. These are fine to put into our swagger_dev.json, but isn't quite ready for prod yet.

cilestin (changes requested):
So close! Just need a couple things removed from swagger.json that aren't ready to go out to prod yet and we're golden!

modules/appeals_api/app/swagger/appeals_api/v2/swagger_dev.json
```diff
@@ -4514,26 +4515,6 @@
                           "email": "bob@bobbytablesemail.com",
                           "timezone": "America/Chicago"
                         },
-                        "claimant": {
```
ElleeB, line 4517: This is really bizarre. Notice of Disagreement v2 is _**not**_ currently displayed on dev-developer-portal.

_the reason this object was removed was **because** it was not displayed on dev-developer-portal_

[Seen here](https://dev-developer.va.gov/explore/appeals/docs/decision_reviews?version=current)

So, the question is, was NOD v2 non-Veteran claimant supposed to be displayed on dev-developer-portal but wasn't displaying for a reason unrelated to us (again)?

ElleeB: @cilestin

ElleeB: I think we need to consider removing rswag. This is getting ridiculous.

modules/appeals_api/app/swagger/appeals_api/v2/swagger.json
ElleeB: 😡

ElleeB (approved):

PR #9699, opened 2022-04-21
2928 appointment request cc eligibility

aherzberg:
## Description of change

Adds Community Cares Eligibility endpoint used for appointment request scheduling

## Original issue(s)
https://github.com/department-of-veterans-affairs/va-mobile-app/issues/2928

- Do we want to put all appointment request endpoints (except the appointment request create) in a single controller with custom actions as well as putting all related routes under a single appointment request url?


modules/mobile/spec/request/cc_eligibility_request_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+          get "/mobile/v0/community_care/eligibility/#{service_type}", headers: iam_headers
+        end
+      end
+
+      it 'returns bad request response' do
+        expect(response).to have_http_status(:bad_request)
+      end
+
+      it 'returns expected schema' do
+        expect(JSON.parse(response.body)['errors'].first['detail']).to eq("Unknown service type: #{service_type}")
```
kpethtel, line 78: I'm not sure the description is accurate.

modules/mobile/config/routes.rb
```diff
@@ -42,6 +42,7 @@
     post '/user/phones', to: 'phones#create'
     put '/user/phones', to: 'phones#update'
     delete '/user/phones', to: 'phones#destroy'
+    get '/community_care/eligibility/:service_type', to: 'cc_eligibility#show'
```
kpethtel, line 45: Maybe I should rename 'specialty' to 'service type' in my PR. That's another place we should standardize naming.

modules/mobile/spec/request/cc_eligibility_request_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+        let(:service_type) { 'PrimaryCare' }
+
+        before do
+          VCR.use_cassette('cc_eligibility/get_eligibility_true', match_requests_on: %i[method uri]) do
+            get "/mobile/v0/community_care/eligibility/#{service_type}", headers: iam_headers
+          end
+        end
+
+        it 'returns successful response' do
+          expect(response).to have_http_status(:success)
```
kpethtel, line 37: Should we test for "true" in the body? Same with false below.

modules/mobile/app/serializers/mobile/v0/cc_eligibility_serializer.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require 'fast_jsonapi'
+
+module Mobile
+  module V0
+    class CCEligibilitySerializer
+      include FastJsonapi::ObjectSerializer
+
+      set_id do |object|
```
kpethtel, line 10: I'm going to try doing this in my PR to avoid creating a model.

kpethtel (commented):
This looks good. Just a couple of spec things. Let's keep talking about naming conventions in slack.

modules/mobile/config/routes.rb
kreek: What do we think about the information architecture of `/community_care` being a top level resource? Could we nest it under appointments? Or if it's specific to requests, maybe even further down?

modules/mobile/spec/request/cc_eligibility_request_spec.rb
aherzberg: haha yes, I forgot to test the most important part.

modules/mobile/spec/request/cc_eligibility_request_spec.rb
aherzberg: good point. fixed

modules/mobile/config/routes.rb
aherzberg: It would get used for both direct and request scheduling. I'm on the fence on how/if we should group the routes. I like the idea putting this one under `appointments/community_care/eligibility/:service_type` as community cares eligibility will always related to appointments. But I'd still want to keep this as its own controller rather than make two separate custom actions on the appointment and appointment request controllers.

aherzberg:
changed the naming of the new files to be Community Cares rather then CC to make it obvious what it was for people who don't know the acronyms.

aherzberg:
Just realized I forgot to update openapi docs. Will add those in.

modules/mobile/app/controllers/mobile/v0/community_cares_eligibility_controller.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Mobile
+  module V0
+    class CommunityCaresEligibilityController < ApplicationController
```
kpethtel, line 5: I would change this to `CommunityCare` singular because that's what we use elsewhere in the app.

kpethtel (commented):
One more small comment. Beyond that, if you feel comfortable with where we landed on Alastair's route question, I think this is good.

modules/mobile/docs/schemas/CommunityCaresEligibility.yml
```diff
@@ -0,0 +1,25 @@
...
+      id:
+        type: string
+        example: "600117255"
+      attributes:
+        type: object
+        required:
+          - eligible
+        properties:
+          eligible:
+            type: string
```
kpethtel, line 24: Isn't this type bool?

kpethtel (dismissed):
One very minor documentation question 👍

modules/mobile/docs/schemas/CommunityCaresEligibility.yml
aherzberg: yes it would. Updated.

PR #9701, opened 2022-04-21
2856 cc providers

kpethtel:
## Description of change
Adds a new endpoint for community care providers. It expects a "specialty" query param and optionally takes a facility id that is used to derive distance. 

## Original issue(s)
department-of-veterans-affairs/va-mobile-app#2856

This makes  a number of small changes from the ticket as described. 
1) The ticket describes using the user's device location. The front end team made it sound like that would be a substantial lift, so it's been left off for now. It's easy to add later if needed. 
4) The ticket also describes using a city to determine the lat/long. But the city options are pulled from facility locations, so I've just based it on the facility instead of the city. The front end can show the user city options based on facility location, then send the back end the facility id. This makes getting the lat/long easy.
3) The ticket also describes omitting the distance because it was unclear how that would be calculated. The calculation was actually free from the upstream service, so it's been included. 

This will be tested on the review instance.


modules/mobile/app/controllers/mobile/v0/community_care_providers_controller.rb
```diff
@@ -0,0 +1,100 @@
+# frozen_string_literal: true
+
+module Mobile
+  module V0
+    class CommunityCareProvidersController < ApplicationController
+      DEFAULT_PAGE_NUMBER = 1
+      DEFAULT_PAGE_SIZE = 10
```
kpethtel, line 7: The web app uses 15 and we may change to 15 to match the web app. I don't believe the web app actually allows pagination, so I think the first 15 are the only options available. The mocks don't show any pagination occurring on mobile either. Nonetheless, I've paginated this controller to stay consistent with the upstream service (which is paginated) and allow us to add pagination if we want to.

modules/mobile/app/controllers/mobile/v0/community_care_providers_controller.rb
```diff
@@ -0,0 +1,100 @@
...
+      RADIUS = 60 # default value used in web app
+      SPECIALTIES = {
+        primaryCare: %w[207QA0505X 363LP2300X 363LA2200X 261QP2300X],
```
kpethtel, line 10: I chose to make these keys (which are the values of query params) camel case. I could've let the front-end send them as strings like "primary care", but I wasn't sure that was any better on their side.

modules/mobile/app/controllers/mobile/v0/community_care_providers_controller.rb
```diff
@@ -0,0 +1,100 @@
...
+          page_size: params.dig(:page, :size) || DEFAULT_PAGE_SIZE
+        )
+      end
+
+      def paginate(records)
+        url = request.base_url + request.path
+        page_records, page_meta_data = Mobile::PaginationHelper.paginate(
+          list: records, validated_params: pagination_params, url: url
+        )
+        # this is temporary. this has come up multiple times and we should develop a better solution
```
kpethtel, line 87: The reason I didn't add a better fix for this is because I think it's unclear what we want out of the PaginationHelper and GetPaginatedList. I wrote the pagination helper because I saw the logic occurring in multiple places. But we don't want all params that are used in link formation to be in GetPaginatedList, so we keep doing this. There are a number of ways we could fix this, but I think we need to determine what we really want out of the GetPaginatedList first.

modules/mobile/app/controllers/mobile/v0/community_care_providers_controller.rb
```diff
@@ -0,0 +1,100 @@
+# frozen_string_literal: true
+
+module Mobile
+  module V0
+    class CommunityCareProvidersController < ApplicationController
```
kpethtel, line 5: I chose to go with the full name "CommunityCare" because I thought it was clearer. I know we use "cc" a lot in appointments, but I think there is value in being more explicit. The web controller is just called ccp_controller, which does not make it any easier to find if you're not familiar with it.

modules/mobile/app/controllers/mobile/v0/community_care_providers_controller.rb
```diff
@@ -0,0 +1,100 @@
...
+        end
+        page_records, page_meta_data = paginate(community_care_providers)
+        serialized = Mobile::V0::CommunityCareProvidersSerializer.new(page_records, page_meta_data)
+        render json: serialized, status: :ok
+      end
+
+      private
+
+      def ppms_api
+        FacilitiesApi::V1::PPMS::Client.new
```
jperk51, line 41: I know we recently updated some appointments logic to use `VAOS::V2::MobileFacilityService` is that also something we could do here?

modules/mobile/app/controllers/mobile/v0/community_care_providers_controller.rb
```diff
@@ -0,0 +1,100 @@
...
+          page: pagination_params[:page_number],
+          per_page: pagination_params[:page_size],
+          radius: RADIUS,
+          specialties: specialty_codes
+        }
+      end
+
+      def coordinates
+        if params[:facilityId]
+          facility = Mobile::FacilitiesHelper.get_facilities(Array(params[:facilityId])).first
```
jperk51, line 67: Similar note about `VAOS::V2::MobileFacilityService` since the helper is using the lighthouse API

modules/mobile/app/controllers/mobile/v0/community_care_providers_controller.rb
kpethtel: I'll look into that. Thanks for pointing it out.

modules/mobile/app/controllers/mobile/v0/community_care_providers_controller.rb
aherzberg: I agree this is the way to go. I changed my PR to match this.

modules/mobile/docs/schemas/CommunityCareProviders.yml
```diff
@@ -0,0 +1,9 @@
+type: object
```
aherzberg, line 1: docs look to be missing the attributes and type field.

aherzberg (commented):
All looks good. This one definitely was more involved then I originally anticipated. Glad we don't have to use a separate geolocation service. There may be some awkwardness with front end showing cities based on registered facilities than sending BE a single facility. If there are multiple cities in a city, which one will they send us? But the alternative is either to use a separate geolocation service to get a general lat long for a city or get providers with all facilities in a city. Both options add a lot of risks and complexity with little pay off. The default radius we use, 60, is that 60 miles? If so, that should captures whole cities even if the facility is on the edge of a city so it's a non-issue.

kpethtel:
> All looks good. This one definitely was more involved then I originally anticipated. Glad we don't have to use a separate geolocation service. There may be some awkwardness with front end showing cities based on registered facilities than sending BE a single facility. If there are multiple facilities in a city, which one will they send us? But the alternative is either to use a separate geolocation service to get a general lat long for a city or get providers with all facilities in a city. Both options add a lot of risks and complexity with little pay off. The default radius we use, 60, is that 60 miles? If so, that should captures whole cities even if the facility is on the edge of a city so it's a non-issue.

Yes, the 60 is miles. You raise an interesting question. I spent some time looking through the web app code and as best I can tell it uses the facility's lat/long, not the city's. I can't reasonably post the full stack trace here because there's a lot of it and I'm not sure I fully understand how the pieces fit together, but here are a few breadcrumbs. 

In src/applications/vaos/new-appointment/components/ClosestCityStatePage.jsx
we form a list of sites:
```
            enumNames: supportedParentSites.map(
              site => `${site.address.city}, ${site.address.state}`,
            ),
```

Which eventually calls `fetchCommunityCareSupportedSites` , which calls the configuration endpoint for a list of cc enabled facilities. 
```
  return apiRequestWithUrl(
    `/vaos/v2/scheduling/configurations?${locationIds
      .map(id => `facility_ids[]=${id}`)
      .join('&')}${ccEnabledParam}`,
    {
      method: 'GET',
    },
```

Also, the sort method is used in `requestProvidersList` (which is part of the method chain that requests the providers) is "distanceFromFacility", not "distanceFromCity":

```
export const FACILITY_SORT_METHODS = {
  distanceFromResidential: 'distanceFromResidentialAddress',
  distanceFromCurrentLocation: 'distanceFromCurrentLocation',
  distanceFromFacility: 'distanceFromFacility',
  alphabetical: 'alphabetical',
};
```

And when we request the providers list, we use facility.position (which is used for its lat/long): 

`dispatch(requestProvidersList(selectedCCFacility?.position));`

Sorry this is a really verbose response, but I had to do a lot of digging and it's hard to follow. I'd be happy to pair with you on this if you want to work with me to confirm how this works.

modules/mobile/docs/schemas/CommunityCareProviders.yml
kpethtel: Thank you for pointing this out. It led me to find some other small issues as well.

modules/mobile/app/controllers/mobile/v0/community_care_providers_controller.rb
```diff
@@ -0,0 +1,100 @@
...
+          [facility.lat, facility.long]
+        else
+          address = current_user.vet360_contact_info.residential_address
+          [address.latitude, address.longitude]
+        end
+      end
```
kreek, line 73: > [Else Considered Harmful](http://iq0.com/notes/deep.nesting.html#:~:text=Else%20Considered%20Harmful)
```suggestion
      def coordinates
        return facility_coordinates if params[:facilityId]
        address_coordinates
      end

      def facility_coordinates
        facility = Mobile::FacilitiesHelper.get_facilities(Array(params[:facilityId])).first
        [facility.lat, facility.long]
      end

      def address_coordinates
        address = current_user.vet360_contact_info.residential_address
        [address.latitude, address.longitude]
      end
```

aherzberg:
> > All looks good. This one definitely was more involved then I originally anticipated. Glad we don't have to use a separate geolocation service. There may be some awkwardness with front end showing cities based on registered facilities than sending BE a single facility. If there are multiple facilities in a city, which one will they send us? But the alternative is either to use a separate geolocation service to get a general lat long for a city or get providers with all facilities in a city. Both options add a lot of risks and complexity with little pay off. The default radius we use, 60, is that 60 miles? If so, that should captures whole cities even if the facility is on the edge of a city so it's a non-issue.
> 
> Yes, the 60 is miles. You raise an interesting question. I spent some time looking through the web app code and as best I can tell it uses the facility's lat/long, not the city's. I can't reasonably post the full stack trace here because there's a lot of it and I'm not sure I fully understand how the pieces fit together, but here are a few breadcrumbs.
> 
> In src/applications/vaos/new-appointment/components/ClosestCityStatePage.jsx we form a list of sites:
> 
> ```
>             enumNames: supportedParentSites.map(
>               site => `${site.address.city}, ${site.address.state}`,
>             ),
> ```
> 
> Which eventually calls `fetchCommunityCareSupportedSites` , which calls the configuration endpoint for a list of cc enabled facilities.
> 
> ```
>   return apiRequestWithUrl(
>     `/vaos/v2/scheduling/configurations?${locationIds
>       .map(id => `facility_ids[]=${id}`)
>       .join('&')}${ccEnabledParam}`,
>     {
>       method: 'GET',
>     },
> ```
> 
> I believe these facilities are what we use when we request the providers list. I believe facility.position is used for its lat/long:
> 
> `dispatch(requestProvidersList(selectedCCFacility?.position));`
> 
> Also, the sort method is used in `requestProvidersList` (which is part of the method chain that requests the providers) is "distanceFromFacility", not "distanceFromCity":
> 
> ```
> export const FACILITY_SORT_METHODS = {
>   distanceFromResidential: 'distanceFromResidentialAddress',
>   distanceFromCurrentLocation: 'distanceFromCurrentLocation',
>   distanceFromFacility: 'distanceFromFacility',
>   alphabetical: 'alphabetical',
> };
> ```
> 
> Sorry this is a really verbose response, but I had to do a lot of digging and it's hard to follow. I'd be happy to pair with you on this if you want to work with me to confirm how this works.

Looks like they use the parent facility which makes sense. There should only be one parent per city and is most like closer to city center. FE will have access to the parent facility as they'll be the registered facilities associated with the user object that FE will provide to the Facility info endpoint. so we may just need to add that to the openapi docs that to make the param `optional parent facility id`

modules/mobile/app/controllers/mobile/v0/community_care_providers_controller.rb
kpethtel: I don't think I can use that here. This endpoint isn't covered by that service.

modules/mobile/app/controllers/mobile/v0/community_care_providers_controller.rb
kpethtel: Here we could switch to the new version. However, it led to issues with the spec configuration that I wasn't able to fix easily. There must be some difference in the setup or user. This is an easy change to make later.

modules/mobile/app/controllers/mobile/v0/community_care_providers_controller.rb
kpethtel: We've agreed on 10.

aherzberg (approved):
LGTM

aherzberg (approved):
LGTM

modules/mobile/app/serializers/mobile/v0/community_care_provider_serializer.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module Mobile
+  module V0
+    class CommunityCareProviderSerializer
+      include FastJsonapi::ObjectSerializer
+
+      set_id :provider_identifier
+
+      attribute :name do |object|
```
kpethtel, line 10: This was taken from the CCP serializer. IMO this should be a model method rather than live in the serializer. That would prevent me from having to copy their business logic.

PR #9710, opened 2022-04-22
RRD: don't show "[production]" in RRD notification subjectline

yoomlam:
## Description of change

Only show the run environment in the subjectline when it is not "production".

## Original issue(s)
We don't need to show "[production]" in the subjectline. That will be the default assumption.
Also show the disability before the status in the subjectline.

Changes subjectline for internal notification.


spec/mailers/rrd_completed_mailer_spec.rb
```diff
@@ -42,8 +42,14 @@
       submission
     end
 
-    it 'has the expected subject' do
-      expect(email.subject).to include 'RRD claim - Processed'
+    it 'has the expected subject in production' do
+      Settings.vsp_environment = 'production'
```
allthesignals, line 46: Just a left-field observation: I'm reminded of that article you shared: because this is environmental, it introduces potential memory leak across randomized tests. While it probably isn't a huge deal here, is there a chance it could affect other things?

allthesignals (dismissed):
Nice! Thank you!

spec/mailers/rrd_completed_mailer_spec.rb
yoomlam: `Settings.vsp_environment` is `nil` in the test environment. I went ahead and restore it at the end of the test just in case.

allthesignals (approved):

RachalCassity (approved):

PR #9728, opened 2022-04-26
2862 appointment facility eligibility

aherzberg:
## Description of change

Create Facility Eligibility endpoint for Appointment Request feature. 

## Original issue(s)
https://github.com/department-of-veterans-affairs/va-mobile-app/issues/2862

Couple things to keep in mind
- Facility info endpoint will load all facilities at once, sorted, but call this eligibility endpoint in batches, paginated, as needed since each facility needs it's own call
- I was initially under the impression that doing multiple parallel calls would be costly but VAOS says they've seen their services called in parallel up to 10 calls at a time. I defaulted the page size to 3 but this can be upped if we find there's no performance implications.


modules/mobile/app/controllers/mobile/v0/facility_eligibility_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+
+module Mobile
+  module V0
+    class FacilityEligibilityController < ApplicationController
+      DEFAULT_PAGE_NUMBER = 1
+      DEFAULT_PAGE_SIZE = 3
+      SERVICE_TYPES = %w[primaryCare nutrition podiatry optometry audiology].freeze
+
+      def index
+        require_params
```
aherzberg, line 11: Is there a better way to do this?

modules/mobile/app/controllers/mobile/v0/facility_eligibility_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+        unless SERVICE_TYPES.include?(params[:serviceType])
```
aherzberg, line 12: Added this because the upstream services was returning 500 error when it was a bad service type

modules/mobile/app/controllers/mobile/v0/facility_eligibility_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+          raise Common::Exceptions::InvalidFieldValue.new('serviceType', params[:serviceType])
+        end
+
+        page_facilities, page_meta_data = paginate(params[:facilityIds])
```
aherzberg, line 16: Took @kpethtel pagination code used on the cc provider, will create a ticket that can be worked once all the appointment requests endpoints are complete to put all pagination under one library.

modules/mobile/app/serializers/mobile/v0/facility_eligibility_serializer.rb
```diff
@@ -0,0 +1,29 @@
...
+      include FastJsonapi::ObjectSerializer
+
+      set_type :FacilityEligibility
+      attributes :facility_id, :eligible, :reason
+
+      def initialize(facility_eligibilities, page_meta_data)
+        resource = facility_eligibilities.collect do |facility_eligibility|
+          FacilityEligibilityStruct.new(
+            facility_eligibility.facility_id,
+            facility_eligibility.facility_id,
```
aherzberg, line 17: Setting both the id to the facility id as well as the adding the facility id as it's own field. since we're not using our own model, what goes in the id field is not intuitive.

modules/mobile/app/controllers/mobile/v0/facility_eligibility_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+
+        results = Parallel.map(page_facilities, in_processes: page_facilities.size) do |facility|
```
kpethtel, line 18: I think this would be slightly clearer with facility renamed to facility_id.

modules/mobile/spec/request/facility_eligibility_request_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+          expect(response).to have_http_status(:ok)
+        end
+
+        it 'matches schema' do
+          expect(response.body).to match_json_schema('facility_eligibility')
+        end
+
+        it 'assumes facility exists but does not support requests' do
+          expect(response.parsed_body.dig('data', 0, 'attributes', 'reason'))
+            .to eq('appointment requests are disabled for the clinical service at the facility')
```
kpethtel, line 153: Interesting. Did you record the cassette for this by going to staging or a review instance?

modules/mobile/app/controllers/mobile/v0/facility_eligibility_controller.rb
kpethtel: One option I've seen is to create individual methods for the variables like:
```
def service_type
  params.require(:serviceType)
end
```
Then use those variables instead of directly calling the params. Alternatively, you could use a before_action, which is pretty much equivalent to what you're doing. I would probably lean more toward creating methods, unless you're doing expensive things before the methods would be called.

kpethtel (dismissed):
This looks good. Just a couple of small comments, but nothing blocking.

modules/mobile/app/controllers/mobile/v0/facility_eligibility_controller.rb
kpethtel: Created: https://app.zenhub.com/workspaces/va-mobile---backend-61d75d03a7505500147ec7f6/issues/department-of-veterans-affairs/va-mobile-app/3086

aherzberg: Thanks!

modules/mobile/app/controllers/mobile/v0/facility_eligibility_controller.rb
aherzberg: Yes, I like that better.

modules/mobile/app/controllers/mobile/v0/facility_eligibility_controller.rb
aherzberg: Yes, makes sense.

modules/mobile/spec/request/facility_eligibility_request_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+
+        it 'returns 200' do
+          expect(response).to have_http_status(:ok)
+        end
+
+        it 'matches schema' do
+          expect(response.body).to match_json_schema('facility_eligibility')
+        end
+
+        it 'forms expected meta links' do
```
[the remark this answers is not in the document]
aherzberg: good catch.

modules/mobile/spec/request/facility_eligibility_request_spec.rb
aherzberg: Went into staging and called the upstream service directly

kpethtel (approved):
The only thing that jumps out at me as missing is any error handling or testing around type. I'm not sure that's a big deal because the other use of the facility eligibility service object doesn't seem to do any special handling for this either.

PR #9730, opened 2022-04-26
API-12180 nod non-veteran claimant validation and error handling

ElleeB:
Adds validation and error handling for NOD v2 non-Veteran claimant.

Addresses the acceptance criteria of two tickets:

[Ticket 1](https://vajira.max.gov/browse/API-12180)
[Ticket 2](https://vajira.max.gov/browse/API-14857)

Still need to update/add tests


modules/appeals_api/spec/models/notice_of_disagreement_spec.rb
```diff
@@ -286,6 +287,76 @@
...
+                'issue' => 'right knee',
+                'decisionDate' => '1901-01-31'
+              }
+            }
+          ]
+        }
+      end
+
+      let(:notice_of_disagreement_v2) do
+        described_class.new(api_version: 'v2', form_data: form_data, auth_headers: auth_headers,
```
ElleeB, line 349: Example of manually creating and remembering to add the api_version. When forgotten, it might lead the tester to think the guard clauses need to check for api_version.

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
```diff
@@ -43,7 +43,10 @@ def self.date_from_string(string)
     # further validations:
     validate(
       :birth_date_is_in_the_past,
-      :claimant_birth_date_is_in_the_past
+      :claimant_birth_date_is_in_the_past,
+      :required_claimant_data_is_present,
+      :contestable_issue_dates_are_valid_dates,
```
cilestin, line 48: Since date format checking is taken care of by the schema, and we have another "is_in_the_past" method, let's rename this to "contestable_issue_dates_are_in_the_past" (or similar) - it's a bit more descriptive to what it's actually checking.

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
```diff
@@ -322,14 +329,50 @@ def birth_date_is_in_the_past
 
     # validation (header)
     def claimant_birth_date_is_in_the_past
-      return if api_version && api_version.upcase == 'V1'
```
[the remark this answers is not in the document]
cilestin: We set the api verison _now_, but didn't back before HLRv2 (and thus api_version) was a thing. We're probably safe to make it a validation on the model to require it at this point, though.

modules/appeals_api/spec/models/notice_of_disagreement_spec.rb
cilestin: This could be built via one of our factories so that the api version is included by default. Is there a reason we aren't using one here and need to use the `described_class`?

cilestin (changes requested):
Some small changes, and it might also be time to look over our error messages as a whole. We've added quite a few recently and I'd like to make sure they're staying reasonably consistent. I feel like every appeal type/version has a slightly different way of handling things and that could mean inconsistency in our error responses.

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
ElleeB: On it.

modules/appeals_api/app/models/appeals_api/notice_of_disagreement.rb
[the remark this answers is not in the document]
ElleeB: Ok. I'm going to wire it up.

modules/appeals_api/spec/models/notice_of_disagreement_spec.rb
ElleeB: Yes, because, because of the way the form data and auth headers are set up. I could make very specific factories for the different test scenarios, but it seems overkill.

I could build the existing factories and modify them, which seems like a fine solution, and I'm happy to refactor, but I still think it's a good example of potential confusing, as it may not be us in the future working on the code...

modules/appeals_api/config/schemas/v2/10182_headers.json
```diff
@@ -26,6 +26,12 @@
         "X-Consumer-Username":  { "$ref": "#/definitions/X-Consumer-Username"},
         "X-Consumer-ID":        { "$ref": "#/definitions/X-Consumer-ID"}
       },
+      "dependencies": {
```
ElleeB, line 29: Almost forgot. I reduced the amount of repetition in the list of dependencies (see contrast with current hlr v2 claimant dependencies).  With each attribute requiring all the rest, as it still is in hlr v2, if two attributes are missed, the error messages are cumbersome and repetitive. For example, on hlr v2 if you don't add claimant first name, but you add all the other attributes, it will repeat the 'x va claimant first name required' error/pointer 3 times. I hope this makes sense, and I think we should improve it on hlr and sc if necessary :)

stiehlrod (approved):
Looks like all comments have been resolved, and the changes look good to me.

modules/appeals_api/app/controllers/appeals_api/v2/decision_reviews/notice_of_disagreements_controller.rb
```diff
@@ -44,16 +44,13 @@ def validate
...
 
   def schema
-    # TODO: Return full schema after we've validated all Non-Veteran Claimant functionality
     response = AppealsApi::JsonSchemaToSwaggerConverter.remove_comments(
       AppealsApi::FormSchemas.new(
         SCHEMA_ERROR_TYPE,
         schema_version: 'v2'
       ).schema(self.class::FORM_NUMBER)
     )
-    response.tap do |s|
```
cilestin, line 54: I think we want to keep this filtering in, since we don't currently want our production schemas to include NVC information for NODv2.

cilestin (changes requested):

modules/appeals_api/app/controllers/appeals_api/v2/decision_reviews/notice_of_disagreements_controller.rb
ElleeB: You're so right. I'll get that back in!

cilestin (approved):

PR #9731, opened 2022-04-26
[API-15406] - ClaimsApi - 526 transform 'addressChangeType' case

jw81:
## Description of change

- Ensures the value of `form526.veteran.changeOfAddress.addressChangeType` is uppercased before sending to EVSS.

## Original issue(s)
[API-15406](https://vajira.max.gov/browse/API-15406)


modules/claims_api/app/models/claims_api/auto_established_claim.rb
```diff
@@ -399,5 +401,18 @@ def remove_special_issues_from_secondary_disabilities
         end
       end
     end
+
+    def change_of_address_provided?
+      form_data['veteran']['changeOfAddress'].present?
+    end
```
hinzed1127, line 407: not worth removing in this PR, but abstracting all these `.present?` checks into helper methods is starting to feel like unnecessary bloat to me

hinzed1127 (approved):

modules/claims_api/app/models/claims_api/auto_established_claim.rb
jw81: Fair enough.
I think the original intent was that I just hated the way it would read if it just said,
`form_data['veteran']['changeOfAddress'] = transform_change_of_address_type_case`
because that feels like it would always be changing something. When in reality, it's only going to change it if it exists.  So then I start adding the little check that would try to signal that, "_we're gonna change this thing, only if the condition is right_". 🤷

PR #9737, opened 2022-04-27
[API-15406] - ClaimsApi - GET claim by id, do not return 422 for errored claim

jw81:
## Description of change

- On the `GET /claims/:id` endpoint, do not return a 422 for a claim that has a `status` of `errored`.  Just return a 200 and ~let the `status` show that it is errored~ display the `status` as `pending`.  Kayla, Dan, and I decided to change this behavior to display the value as `pending` because this is a released API and we can't suddenly begin returning the value of `errored` without potentially breaking the workflow of current consumers.

## Original issue(s)
[API-15406](https://vajira.max.gov/browse/API-15406)

### DIFF
--- a/modules/claims_api/app/controllers/claims_api/v1/claims_controller.rb
+++ b/modules/claims_api/app/controllers/claims_api/v1/claims_controller.rb
@@ -21,11 +21,16 @@ module ClaimsApi
         raise ::Common::Exceptions::ResourceNotFound.new(detail: 'Claims not found')
       end
 
-      def show
+      def show # rubocop:disable Metrics/MethodLength
         claim = ClaimsApi::AutoEstablishedClaim.find_by(id: params[:id])
 
         if claim && claim.status == 'errored'
-          fetch_errored(claim)
+          # This is a released API and we can't start revealing 'errored' as a status,
+          #  without potentially breaking existing consumers.
+          #  So display the 'status' as 'pending'
+          #  Ref :: API-15406
+          claim.status = ClaimsApi::AutoEstablishedClaim::PENDING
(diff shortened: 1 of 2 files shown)


modules/claims_api/app/controllers/claims_api/v1/claims_controller.rb
```diff
@@ -24,9 +24,7 @@ def index
       def show
```
jw81, line 24: If we find in the future that consumers would like this endpoint to show details of what the error is when the claim is errored, we could consider doing something like adding a new serializer that would include the value of the `evss_errors` attribute.

Psuedo-Code Idea
```
if claim.status == 'errored
  render json: claim, serializer: ClaimsApi::AutoEstablishedClaimWithErrorsSerializer
elsif
  ...
end
```

hinzed1127: I think there's still value in conditionally showing any errors and just making it a `200` instead of a `4xx` status code

jw81: I don't know that we can do that because then we'll create the situation where a 200 response from this endpoint could have either a body that represents a claim, or a body that represents a set of errors.  Which I would think could definitely cause breaking changes for existing consumers.

jw81:
**UPDATE ::**
It was decided that maybe right now wasn't the time to attempt to remove this wonky 422 behavior.  So we've gone ahead and created a new ticket where we can maybe reconsider this down the road.  That ticket is [API-15526](https://vajira.max.gov/browse/API-15526).  Closing this PR now, but if we ever decide to pursue API-15526, then we can reference this PR as an example for the logical changes that need to happen.

PR #9741, opened 2022-04-27
[40530] Adding ability for User model to be reloaded as long as sign …

bosawt:
…in service session is still active

## Description of change
This PR reloads the User model if the Sign In Service Session accompanying it is still active, even if the User model has expired in the redis cache

## Original issue(s)
department-of-veterans-affairs/va.gov-team#40530

* To test this, authenticate through the Sign In Session as normal, until you are able to authenticate with an `access_token`
* Run `/introspect` and note the `uuid` that is returned 
* In Rails console, run `User.find(<uuid>).destroy` to destroy the User redis model, while preserving the session
* Run `/introspect` and confirm that user data is successfully returned (meaning User has been reloaded)


bramleyjl (approved):
LGTM, was able to destroy and reload the user.

PR #9742, opened 2022-04-28
RRD: increase Sidekiq job retries for RRD Asthma job

yoomlam:
## Description of change

Increase Sidekiq job retries for RRD Asthma job

## Original issue(s)
RRD Asthma processor is sufficiently stable to keep retrying in case of temporary external service failure.


app/workers/rapid_ready_for_decision/form526_asthma_job.rb
```diff
@@ -6,6 +6,6 @@ module RapidReadyForDecision
   class Form526AsthmaJob < Form526BaseJob
     STATSD_KEY_PREFIX = 'worker.fast_track.form526_asthma_job'
 
-    sidekiq_options retry: 2
+    sidekiq_options retry: 11
```
ericboehs, line 9: Why 11? Do we know how long this stay in the retry queue if it continually fails?

yoomlam: 11 is what is use in the `Form526BaseJob`:
https://github.com/department-of-veterans-affairs/vets-api/blob/7f75177d668ca4e1bfb7b3bd8d8d19b7fee8fbd1/app/workers/rapid_ready_for_decision/form526_base_job.rb#L16

(So we could actually remove this line completely, but we'll do that later. I'd like to keep it for now to make it explicit.)

According to https://github.com/mperham/sidekiq/wiki/Error-Handling#automatic-job-retry, 11 retries equates to about 0d  7h 21m 28s.

ericboehs (approved):

PR #9746, opened 2022-04-28
2732 mobile staging test tokens

kreek:
## Original issue(s)


patrickvinograd:
You might look at this PR I put up, but haven't gotten around to merging - same idea but for VA.gov login - instantiate a session based on a yaml file of user data. I was going to write a script to spit out these yaml files from mockdata, but didn't get there yet. But we could consolidate these two approaches somewhat.

kreek:
@patrickvinograd thanks for that. @kpethtel and I have been brainstorming on ways to automate/reduce the friction in getting API tokens for testing. Specifically, in staging, so that our QA team can automate API integration tests against that environment. 

For this solution, I’m not crazy about having alternate auth code in there, even if it is only for staging, that works around login using one JWT secret for all users.

We have another ChromeDriver solution that uses the full auth flow. ~~60%~~ 95% of the time it works every time with the redirects/IAM staging flakiness. However, we can limit the browser driving with the solutions (IAM, new identity login) that implement refresh tokens.

Ideally the auth flow would have an API only way to log users in. We asked for this is in our feedback on the identity team’s new solution. Since that would also work for regular va.gov users, it may be worth talking to the identity team about that.

PR #9747, opened 2022-04-28
Removes never-launched /v0/vso_appointments API

patrickvinograd:
## Description of change
Removes the /v0/vso_appointments API controller, and the associated model and service code. 

This API was built as part of an early platform pilot and was never launched to production. 

Also establishes `docs/HISTORY.md` as a way of tracking removed/deprecated API resources.

## Original issue(s)
(none)

* There is no database scheme for this model, hence no migration needed to clean up data. 
* This API submitted data to the central mail upstream service, so existing central mail configuration remains in place for use by other APIs.


patrickvinograd:
IDK why this is failing the test run - looks like all tests pass but coverage fails, maybe somehow related to parallel specs? I feel like this was happening a while back...

ericboehs (approved):

ericboehs:
Girl Happy GIF

ewilson-adhoc (approved):
Good stuff!

PR #9751, opened 2022-04-29
2982 precache appointments flag

kpethtel:
## Description of change
Adding a feature flag for turning off appointments pre-caching when the user logs in. We're doing this because we apparently brought down the upstream service when we turned on the appointment request index feature. 
## Original issue(s)
department-of-veterans-affairs/va-mobile-team#2982

I've also added my personal email address to the settings.yml because I already have a production id.me account via that email. You can only have one production account, so I can't create another one. I need that access to be able to turn on this feature in production. 

I didn't want to give the feature flag a negative name like, "disable_appointments_precache", so I will need to turn on this feature in production and staging once the code is merged.


modules/mobile/app/controllers/mobile/v0/users_controller.rb
```diff
@@ -30,7 +30,9 @@ def options
       end
 
       def pre_cache_resources
-        Mobile::V0::PreCacheAppointmentsJob.perform_async(@current_user.uuid)
+        if Flipper.enabled?(:mobile_precache_appointments)
```
kpethtel, line 33: IMO the flag needs to be in the job itself to ensure that there's not a way around the flag, but I also added it in the controllers because I don't want to queue up a bunch of jobs for no reason and add more strain to sidekiq.

modules/mobile/spec/workers/pre_cache_appointments_job_spec.rb
```diff
@@ -78,96 +78,107 @@
           end
         end
       end
+    end
```
kpethtel, line 81: I re-structured this test file slightly because I noticed that the 'with errors' context was embedded within the 'without errors' context. This is the problem with embedding too much. It becomes visually difficult to parse and leads to this type of error. So I decided to flatten the file into more top-level contexts.

config/settings.yml
```diff
@@ -885,6 +885,7 @@ flipper:
     - kevin.musiorski@gcio.com # Kevin Musiorski (GCIO), Engineer
     - kmircovich@governmentcio.com
     - kris.pethtel@adhocteam.us
+    - kris.pethtel@gmail.com # email tied to production id.me account
```
kpethtel, line 888: As mentioned in the description, I needed to add my personal email in order to get access to feature flags in production.

config/features.yml
```diff
@@ -416,6 +416,10 @@ features:
     actor_type: user
     description: Finds facilities using VAOS's Mobile Facility Service.
     enable_in_development: false
+  mobile_precache_appointments:
```
aherzberg, line 419: How does this work for staging? Will we be able to test the flipper in staging?

kpethtel: I believe it will default to off and I will have to turn the flag on after this code is deployed.

ericboehs (approved):

kreek (approved):
LGTM 🚀

